PR #20593, opened 2022-03-11
fix: use decodeURI in loader and add tests

ZachJW34:
Use `decodeURI` so as to allow specs with `[]` and spaces to be loaded and executed.


JessicaSachs (changes requested):
Can we get some unicode tests and a test with the brackets? e.g. `サイプレス.cy.js` (Cypress) and `[...all].cy.js`

lmiller1990:
I'll add those to the main PR @JessicaSachs.

PR #20598, opened 2022-03-12
feat: Refetch cloud query based on notification

estrada9166:
- Closes  [UNIFY-1353](https://cypress-io.atlassian.net/jira/software/projects/UNIFY/boards/20?assignee=5ea883deddd45f0b94f5d98d&selectedIssue=UNIFY-1353)

### User facing changelog

Use subscriptions to know when the organization has been created, update the data without the need for polling

[Cloud work](https://github.com/cypress-io/cypress-services/pull/4193)

Screen Shot 2022-03-21 at 1 29 35 PM
Screen Shot 2022-03-21 at 1 29 45 PM

### PR Tasks


packages/app/src/runs/modals/CreateCloudOrgModal.vue
```diff
@@ -34,13 +34,6 @@
           </template>
           {{ t('runs.connect.modal.createOrg.waitingButton') }}
         </Button>
-        <Button
```
tgriesser, line 37: Is there any benefit to keeping this button around to just refetch the query (rather than poll) just incase there's something that goes wrong with the subscription/notification?

marktnoonan (changes requested):
I think we should add a test for the refresh button as well, so that something would fail if it is removed.

Testing manually, I got a bit stuck in the "waiting for you to create an organization" state. This might have been because I had to set up a new Cy account in staging with no organizations, verify my email, come back, and created the org in a totally different tab.

It did eventually flip back to the state that allows me to refresh the list. Maybe just because the polling 60s expired

packages/app/src/runs/modals/CreateCloudOrgModal.vue
```diff
@@ -101,28 +94,19 @@ const props = defineProps<{
...
+const refetch = useDebounceFn(() => query.executeQuery(), 1000)
 
-const { pause, resume } = useIntervalFn(() => {
-  if (props?.gql?.organizationControl?.nodes?.length || 0 > 0) {
-    pause()
-  } else {
-    query.executeQuery()
-  }
-}, 5000)
+const polling = ref(false)
```
marktnoonan, line 104: Is `polling` still the best name for this? Should is be `waitingForOrg`, `startWaitingForOrg()` or something like that?

marktnoonan:
@estrada9166 actually, manual testing probably won't work at all until the cloud work is merged, would that be right.

estrada9166:
@marktnoonan you can test it manually making a request to `http://localhost:4444/cloud-notification?operationName=orgCreated` in the browser tab

But with the cloud merge is gonna be automatically

marktnoonan (approved):
Nice!

packages/app/src/runs/modals/CreateCloudOrgModal.vue
```diff
@@ -101,29 +94,20 @@ const props = defineProps<{
...
 
-function startPolling () {
-  if (!polling.value) {
-    resume()
-    polling.value = true
-  }
+function startWaitingOrgToBeCreated () {
+  waitingOrgToBeCreated.value = true
 
   setTimeout(() => {
```
lmiller1990, line 109: Do we need to clean this `setTimeout` up (in `onUnmount`, or using the `useTimeout` function maybe?)

packages/app/src/runs/modals/CreateCloudOrgModal.vue
```diff
@@ -101,29 +94,20 @@ const props = defineProps<{
 }>()
 
 const query = useQuery({
-  query: CheckCloudOrganizationsDocument,
+  query: CloudOrganizationsCheckDocument,
   requestPolicy: 'network-only',
+  pause: true,
```
lmiller1990, line 99: Why is this `pause: true`? Is that because we want to trigger it manually, but not automatically?

packages/data-context/src/sources/CloudDataSource.ts
```diff
@@ -129,7 +129,7 @@ export class CloudDataSource {
       return dfd.promise
     }
 
-    return pipe(executingQuery, toPromise).then((data) => {
+    return pipe(executingQuery, take(1), toPromise).then((data) => {
```
lmiller1990, line 132: Why do we do `take(1)` here?

packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
```diff
@@ -155,13 +155,17 @@ async function makeE2ETasks () {
       sinon.stub(ctx.actions.electron, 'openExternal')
       sinon.stub(ctx.actions.electron, 'showItemInFolder')
 
+      const operationCount: Record<string, number> = {}
```
lmiller1990, line 158: Could also use `Map` - `new Map<string, string>`. Not sure which is more idiomatic - just an optional suggestion.

lmiller1990: Do we ever need to reset the call counts?

lmiller1990 (commented):
Just some questions

packages/data-context/src/sources/CloudDataSource.ts
tgriesser: @estrada9166 can you add a comment w/ a link to where you found that this was needed?

packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
estrada9166: it is going to reset before each test

estrada9166:
@lmiller1990 I was using staging to create/delete org, and because the cloud side was not merged (now it is) I make requests to `http://localhost:4444/cloud-notification?operationName=orgCreated` so, it's gonna trigger the subscription

lmiller1990 (approved):

lmiller1990:
Looks good - I did not test this one locally, I'll give it a try once the cloud stuff is merged.

tgriesser (approved):

PR #6459, opened 2020-02-14
User-provided options immutability.

sainthkh:
- Closes #3171

Breaking down #5762 into 3 PRs.

### User facing changelog

* Clone `options` object so that they're not mutated by the `cy` commands.

### PR Tasks

- [x] Have tests been added/updated?
- [ ] Has the original issue been tagged with a release in ZenHub? 
- [ ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [ ] Have API changes been updated in the [`type definitions`](cli/types/index.d.ts)?
- [ ] Have new configuration options been added to the [`cypress.schema.json`](cli/schema/cypress.schema.json)?

PR #20377, opened 2022-02-27
fix: types for Cypress.Commands.add (#20376)

atarek12:
- Closes https://github.com/cypress-io/cypress/issues/20376

### User facing changelog
Fix `Cypress.Command.add()` TypeScript types. Fixes [#20376](https://github.com/cypress-io/cypress/issues/20376)
 - Infer the subject type to be an `element` while using `prevSubject: true` option
 - Adding property `selector` to the subject while using `prevSubject: true | "element"` option

### Additional details
This change fixes typescript errors while using the `Cypress.Command.add()` with { prevSubject: true } or { prevSubject: "element" }

### How has the user experience changed?
No typescript error while trying to access the selector.
```ts
Cypress.Commands.add("newCommand", { prevSubject: "element" }, (element) => {
  element.selector // no typescript error
});
```

### PR Tasks
- [X] Have tests been added/updated?
- [x] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [ ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [ ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [ ] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

PR #20601, opened 2022-03-14
chore: remove command type todos

sainthkh:
- Related with #18172

### User facing changelog

N/A. It's an internal cleanup. I removed command type `TODO`s.

### Additional details
- Why was this change necessary? => I had to add a lot of `TODO`s on commands when trying to remove `// @ts-ignore` comments. They had to be cleaned up.
- What is affected by this change? => Nothing.

Long ago, I tried to name internally-extended `options` with `opts` in #6459. `opts` and `options` are a bit confusing and this idea was turned down. We decided to use `options` and `userOptions`. 

But the problem is that TypeScript doesn't like extending objects. To do that, we need to use general types like `any`, `Record`, etc. 

To solve this problem, we need a way to name this internally-extended `options`. I thought about a lot of name candidates like `iOptions`, `internalOptions`, etc. But decided to use `_options`. 

Traditionally, private class members were named with initial `_` [(link)](https://stackoverflow.com/a/27853642/1038927). I think it's simple name for our purpose (`options` with internal _private_ members).

N/A.

### PR Tasks


cli/types/cypress.d.ts
```diff
@@ -2500,7 +2500,7 @@ declare namespace Cypress {
     action: 'select' | 'drag-drop'
   }
 
-  interface BlurOptions extends Loggable, Forceable { }
+  interface BlurOptions extends Loggable, Timeoutable, Forceable { }
```
sainthkh, line 2503: [`cy.blur`](https://docs.cypress.io/api/commands/blur#Arguments) supports `timeout` but it was missing in types. I added it.

packages/driver/src/cy/commands/files.ts
```diff
@@ -109,16 +117,15 @@ export default (Commands, Cypress, cy, state) => {
       return verifyAssertions()
     },
 
-    // TODO: change the type of `any` to `Partial<Cypress.WriteFileOPtions & Cypress.Timeoutable>`
-    writeFile (fileName, contents, encoding, options: any = {}) {
+    writeFile (fileName, contents, encoding, options: Partial<Cypress.WriteFileOptions & Cypress.Timeoutable> = {}) {
```
emilyrohrbough, line 120: It would be easier to review this if we renamed the parameter instead of changing `options` to `_options`. The majority of the changes would be adding the types then.

packages/driver/cypress/integration/commands/screenshot_spec.js
```diff
@@ -761,9 +761,9 @@ describe('src/cy/commands/screenshot', () => {
         cy.get('.short-element').within(() => {
           cy.screenshot({ capture: 'runner' })
         }).then(() => {
-          // the runner was captured
-          expect(Cypress.action.withArgs('cy:before:screenshot').args[0][1].appOnly).to.be.true
-          expect(Cypress.automation.withArgs('take:screenshot').args[0][1].capture).to.equal('viewport')
+          // the runner was captured ("appOnly === true" means to hide the runner UI)
+          expect(Cypress.action.withArgs('cy:before:screenshot').args[0][1].appOnly).to.be.false
+          expect(Cypress.automation.withArgs('take:screenshot').args[0][1].capture).to.equal('runner')
```
sainthkh, line 766: According to the definition of `appOnly` below,

https://github.com/cypress-io/cypress/blob/b0fe8bf0dd90d302e51b8d75782edb3b19d5af3e/packages/driver/src/cy/commands/screenshot.ts#L269-L272

it means to hide the runner UI. In other words, it should be `false` to capture the runner UI.  

This weird test failure happened because `appOnly` exists only for testing. 

And the original code was trying to check the options given with `options` after `userOptions` alias is created. 

https://github.com/cypress-io/cypress/blob/b0fe8bf0dd90d302e51b8d75782edb3b19d5af3e/packages/driver/src/cy/commands/screenshot.ts#L444-L455

The code above, when `name` is `{ capture: 'runner' }` and `options` is `undefined`. `options.capture !== 'runner'` is always `true`.

packages/driver/src/cy/commands/files.ts
sainthkh: I changed `options` to `userOptions` and `_options` to `options`. 

I was trying not to change the parameter name, because I thought it would be great to generate type definitions automatically from the source code. 

But I realized that it is almost impossible for us because the first `subject` parameter should be specified on our commands but they don't exist in our API. 

So, it doesn't matter to change those names. 

This PR focuses on fixing the type `TODO`s. 

In the next PR, other un-typed `option`s should be renamed to `userOptions` and types should be specified.

cli/types/cypress.d.ts
flotwig: hmm, what does `.timeout` actually do in `cy.blur`? it gets passed to Cypress.log, but shouldn't this be a synchronous operation? maybe we need to remove this if it's a no-op.

packages/driver/src/cy/commands/files.ts
```diff
@@ -3,18 +3,24 @@ import { basename } from 'path'
 
 import $errUtils from '../../cypress/error_utils'
 
+interface InternalReadFileOptions extends Partial<Cypress.Loggable & Cypress.Timeoutable> {
+  _log?: any
```
flotwig, line 7: all of these `_log: any`s... these should ideally be `Cypress.Log`, is there a reason they're not?

flotwig (changes requested):
nice, i love a good TS improvement PR. just a couple of questions

cli/types/cypress.d.ts
sainthkh: I also think it's a synchronous operation and harder to finish blur() over 4 seconds. 

But it might be here because we need it for `retry` and `verifyAssertions`. 

And it's also in the documentation. Removing it is a breaking change.

packages/driver/src/cy/commands/files.ts
sainthkh: Their exact type is this `Log` class. I fixed them to use that.

cli/types/cypress.d.ts
flotwig: IDK, I feel like if it's a no-op currently, removing the option isn't really breaking. But it's ok, don't need to worry about it here.

flotwig (approved):

packages/driver/src/cy/commands/navigation.ts
```diff
@@ -973,13 +974,13 @@ export default (Commands, Cypress, cy, state, config) => {
           }
 
           if (options.log) {
```
tbiethman, line 976: Might be worth changing the check at this point since we're fairly removed from the initial setting of _log.
```suggestion
          if (options._log) {
```

tbiethman (approved):
Just had the one minor comment

packages/driver/src/cy/commands/navigation.ts
sainthkh: `log` and `_log` can be a bit different. 

* `log` is a user-provided or default option. 
* `_log` can be `undefined` if some conditions are met. 

I personally think changing `log` to `_log` doesn't change much in this case, but I believe it's safer to leave it as-is.

sainthkh:
Failure caused by #20377. Fix at #21106.

packages/driver/src/cy/keyboard.ts
```diff
@@ -12,6 +12,7 @@ import type { HTMLTextLikeElement } from '../dom/elements'
 import $selection from '../dom/selection'
 import $utils from '../cypress/utils'
 import $window from '../dom/window'
+import type { Log } from '../cypress/log'
```
emilyrohrbough, line 15: These should pull from the existing Log Types

```suggestion
- import type { Log } from '../cypress/log'
```

packages/driver/src/cy/keyboard.ts
```diff
@@ -680,7 +681,7 @@ export interface typeOptions {
   force?: boolean
   simulated?: boolean
   release?: boolean
-  _log?: any
+  _log?: Log
```
emilyrohrbough, line 684:
```suggestion
  _log?: Cypress.Log
```

packages/driver/src/cy/commands/navigation.ts
emilyrohrbough: +1. Agree with @tbiethman, at this point _log is referencing the Log instance created by the command when the user options are `log = true`

packages/driver/src/cy/commands/actions/submit.ts
```diff
@@ -5,14 +5,17 @@ import $dom from '../../../dom'
...
+}
 
 export default (Commands, Cypress, cy) => {
   Commands.addAll({ prevSubject: 'element' }, {
-    // TODO: any -> Partial<Cypress.Loggable & Cypress.Timeoutable>
-    submit (subject, options: any = {}) {
-      const userOptions = options
-
-      options = _.defaults({}, userOptions, {
+    submit (subject, userOptions: Partial<Cypress.Loggable & Cypress.Timeoutable> = {}) {
```
emilyrohrbough, line 17:
```suggestion
    submit (subject:JQuery<HTMLFormElement>, userOptions: Partial<Cypress.Loggable & Cypress.Timeoutable> = {}) {
```

packages/driver/src/cy/commands/angular.ts
```diff
@@ -89,10 +94,7 @@ export default (Commands, Cypress, cy, state) => {
   }
 
   Commands.addAll({
-    // TODO: Change the options type from `any` to `Partial<Cypress.Loggable & Cypress.Timeoutable>`.
-    ng (type, selector, options: any = {}) {
-      const userOptions = options
-
+    ng (type, selector, userOptions: Partial<Cypress.Loggable & Cypress.Timeoutable> = {}) {
```
emilyrohrbough, line 97:
```suggestion
    ng (type:string, selector:string, userOptions: Partial<Cypress.Loggable & Cypress.Timeoutable> = {}) {
```

packages/driver/src/cy/commands/cookies.ts
```diff
@@ -78,6 +79,30 @@ function cookieValidatesSecurePrefix (options) {
...
 }
 
+interface InternalGetCookieOptions extends Partial<Cypress.Loggable & Cypress.Timeoutable> {
+  _log?: Log
+  cookie?: any
+}
+
+interface InternalGetCookiesOptions extends Partial<Cypress.Loggable & Cypress.Timeoutable> {
+  _log?: Log
+  cookies?: any
```
emilyrohrbough, line 89: This should be an array of cookies.

packages/driver/src/cy/commands/cookies.ts
```diff
@@ -78,6 +79,30 @@ function cookieValidatesSecurePrefix (options) {
...
+}
+
+interface InternalSetCookieOptions extends Partial<Cypress.SetCookieOptions> {
+  _log?: Log
+  name: string
+  cookie?: any
```
emilyrohrbough, line 95:
```suggestion
  cookie?: Cypress.Cookie
```

packages/driver/src/cy/commands/files.ts
```diff
@@ -2,19 +2,26 @@ import _ from 'lodash'
 import { basename } from 'path'
 
 import $errUtils from '../../cypress/error_utils'
+import type { Log } from '../../cypress/log'
+
+interface InternalReadFileOptions extends Partial<Cypress.Loggable & Cypress.Timeoutable> {
+  _log?: Log
```
emilyrohrbough, line 8:
```suggestion
  _log?: Cypress.Log
```

packages/driver/src/cy/commands/window.ts
```diff
@@ -35,6 +36,24 @@ type CurrentViewport = Pick<Cypress.Config, 'viewportWidth' | 'viewportHeight'>
...
+interface InternalTitleOptions extends Partial<Cypress.Loggable & Cypress.Timeoutable> {
+  _log?: Log
+}
+
+interface InternalWindowOptions extends Partial<Cypress.Loggable & Cypress.Timeoutable> {
+  _log?: Log
+  error?: any
+}
+
+interface InternalDocumentOptions extends Partial<Cypress.Loggable & Cypress.Timeoutable> {
```
emilyrohrbough, line 48: Can we reuse this interface?

patches/@types+jquery.scrollto+1.4.29.dev.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@types/jquery.scrollto/index.d.ts b/node_modules/@types/jquery.scrollto/index.d.ts
+index 0a00c69..1707682 100755
+--- a/node_modules/@types/jquery.scrollto/index.d.ts
++++ b/node_modules/@types/jquery.scrollto/index.d.ts
+@@ -6,7 +6,7 @@
+ 
+ /// <reference types="jquery"/>
+ 
+-interface ScrollToOptions {
++interface ScrollToOptions extends JQuery.EffectsOptions<any> {
```
emilyrohrbough, line 10: There wasn't a dep update. Did we need to update this patch?

emilyrohrbough (changes requested):

emilyrohrbough:
@sainthkh FYI - there are a BUNCH of log types coming this week! You'll appreciate them 😄  https://github.com/cypress-io/cypress/pull/18075/files#diff-a2cd504f16ad316e4675f37a2b3d9d7032d6cc6e3b591934a7b453b16a7ad924R10

packages/driver/src/cy/commands/navigation.ts
sainthkh: I changed `log` to `_log` when we're trying to manipulate the `_log`.

packages/driver/src/cy/keyboard.ts
sainthkh: It's not possible because the `Log` class under `driver/src/cypress/log.ts` has some internal methods.

Here are some examples:

https://github.com/cypress-io/cypress/blob/68173be4aa3a9d6c64cc447aa68d8f374bc3c066/packages/driver/src/cy/commands/actions/click.ts#L138-L140

https://github.com/cypress-io/cypress/blob/68173be4aa3a9d6c64cc447aa68d8f374bc3c066/packages/driver/src/cy/assertions.ts#L465-L468

packages/driver/src/cy/keyboard.ts
sainthkh: Please check the comment above.

packages/driver/src/cy/commands/actions/submit.ts
sainthkh: Done.

packages/driver/src/cy/commands/angular.ts
sainthkh: Done.

packages/driver/src/cy/commands/cookies.ts
sainthkh: Done.

packages/driver/src/cy/commands/files.ts
sainthkh: Please check the comment above.

packages/driver/src/cy/commands/window.ts
sainthkh: I've chose this way because they're for different commands. They can grow differently. And it's only 2 of them (it's not 3 strike out yet.)

patches/@types+jquery.scrollto+1.4.29.dev.patch
sainthkh: It's because of the code below:

https://github.com/cypress-io/cypress/blob/68173be4aa3a9d6c64cc447aa68d8f374bc3c066/packages/driver/src/cy/commands/actions/scroll.ts#L121-L143

`ScrollToOptions` uses `Jquery.EffectsOptions` methods like `done()`, `always()`, but they don't exist without extending it.

PR #20610, opened 2022-03-14
refactor: remove for-of loop with await, replace with Promise.all + map

tgriesser:
Perf improvement/generally the pattern we should be following for loop iteration & promises, the `for / of` with an `await` in the body will execute in series, while the `Promise.all` loop executes in parallel.

Minor thing, but just a good idea to an eye out for these & replace anytime it won't affect the outcome.

### DIFF
--- a/packages/data-context/src/codegen/code-generator.ts
+++ b/packages/data-context/src/codegen/code-generator.ts
@@ -36,7 +36,7 @@ export async function codeGenerator (
   const templateFiles = await allFilesInDir(action.templateDir)
   const codeGenResults: CodeGenResults = { files: [], failed: [] }
 
-  for (const file of templateFiles) {
+  const scaffoldResults = await Promise.all(templateFiles.map(async (file) => {
     const isBinary = await isBinaryFile(file)
     const parsedFile = path.parse(file)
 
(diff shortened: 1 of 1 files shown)


packages/data-context/src/codegen/code-generator.ts
```diff
@@ -116,20 +125,16 @@ function computePath (
...
-      res.push(child)
-    } else {
-      res = [...res, ...(await allFilesInDir(child))]
-    }
-  }
+    return isDir ? await allFilesInDir(child) : child
+  }))
 
-  return res
+  return _.flatten(result)
```
emilyrohrbough, line 137: Are there unit tests that can be associated with this change?

emilyrohrbough (approved):

tbiethman (approved):

packages/data-context/src/codegen/code-generator.ts
tgriesser: There shouldn't be a functional change here, it's just doing what the previous behavior was: 
```
res = [...res, ...(await allFilesInDir(child))]
```
and that should be covered in the unit test suite for `code-generator.spec.ts` which is passing

ZachJW34 (approved):

PR #20629, opened 2022-03-15
fix: remove drag from toast notifications in App and Launchpad

marktnoonan:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1141

### User facing changelog
These toasts are not shown to users in production mode, but in development, the default behavior of the vue-toastification library makes it difficult to copy and share the errors that might appear here. This PR configures the toast notifications so they will no longer close on click, or be closed with drag. This allows text to be copied from the toast and it can still be closed with the X button.

### How to test
Edit a graphql field resolver to throw an error, for example the [`needsRefresh` one here](https://github.com/cypress-io/cypress/blob/4626f7481c9904fec484aa167a02e0197a3095c4/packages/graphql/src/schemaTypes/objectTypes/gql-DevState.ts#L36) then start Cypress and you should get an error in the launchpad right away.

### Automated tests
Currently we don't have any automated tests around the toasts since they should never appear as part of the user facing application. If we decide we want them, we could test this behavior too, but it seemed a bit much for this PR to add those tests.

### PR Tasks


flotwig (approved):
nice!

ZachJW34 (approved):

PR #20643, opened 2022-03-16
fix: remove --config-file false references and update types

ZachJW34:
- Closes [UNIFY-1277](https://cypress-io.atlassian.net/browse/UNIFY-1277)

### User facing changelog
Remove `--config-file=false` references and throw error if it is passed

### Additional details
There are a few places we can throw this error, but I deemed it best to throw early so I could clean up some of the code that had special handling for `config-file=false` in the cli and server packages and to not spin up any further processes needlessly.

Latest Cypress docs for unification have already been updated to remove config-file=false references.

Screen Shot 2022-03-16 at 10 36 03 AM

### PR Tasks


packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -1126,10 +1122,6 @@ export class ProjectLifecycleManager {
       hasCypressEnvFile: fs.existsSync(this._pathToFile('cypress.env.json')),
     }
 
-    if (configFile === false) {
-      return metaState
-    }
-
```
emilyrohrbough, line 1132: below on line 1151, do we need the `if (typeof configFile === 'string') {` check?

emilyrohrbough (approved):

packages/data-context/src/data/ProjectLifecycleManager.ts
ZachJW34: Good catch, we can just change that to check for a truthy value. It can be null here since this is where the default gets detected if none is specified.

ZachJW34: Done in [cd40998](https://github.com/cypress-io/cypress/pull/20643/commits/cd40998ed0648f89da31ba9cf652684efde28713)

emilyrohrbough (approved):

cli/lib/exec/open.js
```diff
@@ -67,7 +69,17 @@ const processOpenOptions = (options = {}) => {
...
+
+    try {
+      args = processOpenOptions(options)
+    } catch (err) {
+      if (err.details) {
+        return exitWithError(err.details)()
+      }
+
+      throw err
+    }
```
lmiller1990, line 82: Seems `args` should be scoped to `open`, the only place it's used?

```js
function open () {
  try {
    const args = processOpenOptions(options)
    return spawn.start(args, {
      dev: options.dev,
      detached: Boolean(options.detached),
      stdio: 'inherit',
    })
  } catch (err) {
    if (err.details) {
      return exitWithError(err.details)()
    }
    throw err
  }
}
```

Optional, but I feel this is an improvement.

packages/driver/src/cypress/error_messages.ts
```diff
@@ -25,11 +25,7 @@ function removeLeadingSlash (str: string) {
   return str.substring(li + 1)
 }
 
-const formatConfigFile = (projectRoot: string, configFile: string | false) => {
-  if (configFile === false) {
-    return '`cypress.config.{ts|js}` (currently disabled by --config-file=false)'
-  }
-
+const formatConfigFile = (projectRoot: string, configFile: string) => {
```
lmiller1990, line 28: Potentially could use `Cypress.Config['configFile']` for a single source of truth for the type of `configFile`, not sure though. I wonder if we want to start working towards this?

Eg - you wouldn't have needed this change if we'd been doing this in the first place, it would be free since you'd be just changing the type for `Cypress.Config['configFile']`.

packages/server/lib/project-base.ts
```diff
@@ -234,7 +234,7 @@ export class ProjectBase<TServer extends Server> extends EE {
     await this.saveState(stateToSave)
 
     await Promise.all([
-      checkSupportFile({ configFile: cfg.configFile, supportFile: cfg.supportFile }),
+      checkSupportFile({ supportFile: cfg.supportFile }),
```
lmiller1990, line 237: Do we need `Promise.all` now it's just one value?

packages/server/lib/project_utils.ts
```diff
@@ -38,10 +38,8 @@ export const getSpecUrl = ({
 
 export const checkSupportFile = async ({
   supportFile,
```
lmiller1990, line 40: Prob can just be a positional arg, not much point having an object arg for one value.

packages/types/src/modeOptions.ts
```diff
@@ -2,7 +2,7 @@ export interface CommonModeOptions {
   invokedFromCli: boolean
   userNodePath?: string
   userNodeVersion?: string
-  configFile?: false | string | null
+  configFile?: string | null
```
lmiller1990, line 5: How come this can be `null`? I wonder if we can do `configFile?: string` here. Not your change specifically but always good to question and improve code quality.

lmiller1990 (commented):
Cool PR, functionality seems fine, I left a bunch of comments - I think we have some quick wins in terms of code quality here, some related to your PR, some not. If you don't think this is the time and place to do them, all good, let me know and I can give you an approval.

Main comment I'd like to see at least considered/responded to: https://github.com/cypress-io/cypress/pull/20643/files#diff-aa4d6453f80b1103477a0cc93c94f94c5cfb3866d029b1b70708e77be023600bR82

packages/types/src/modeOptions.ts
ZachJW34: It can be null here since you can opt to not pass in any value for `configFile` and if none is passed, we will assume you are using the default and populate it with the detected file. So after ProjectLifeCycleManager runs, `configFile` will be a string.

packages/server/lib/project_utils.ts
emilyrohrbough: @lmiller1990 You arn't wrong...I do have a sneaky suspicion this will be needed to update this logic to 
```
const found = await fs.pathExists(path.join(configFile.projectPath, supportFile))
```
to handle custom supportFile paths. I'm pretty sure I logged a Jira for this. If not, I'll get it logged.

emilyrohrbough: Nope! I'm wrong. I was hitting this issue at one point, but in the latest binary this isn't an issue. 

+1 to updating to a single arg.

packages/driver/src/cypress/error_messages.ts
ZachJW34: Done in [84bfdcc](https://github.com/cypress-io/cypress/pull/20643/commits/84bfdcc97c31c9acd6077486f9976bb6b18ddf73), also converted the `projectRoot` definition as it's the same deal. I agree, we should use the predefined types when and where we can.

packages/server/lib/project-base.ts
ZachJW34: Done in [84bfdcc](https://github.com/cypress-io/cypress/pull/20643/commits/84bfdcc97c31c9acd6077486f9976bb6b18ddf73)

packages/server/lib/project_utils.ts
ZachJW34: Done in [84bfdcc](https://github.com/cypress-io/cypress/pull/20643/commits/84bfdcc97c31c9acd6077486f9976bb6b18ddf73)

emilyrohrbough (approved):

cli/lib/exec/open.js
ZachJW34: I also made the same change to `run` so they are symmetrical.

lmiller1990 (approved):

PR #20652, opened 2022-03-16
fix: renaming specs on migration account for numbers

estrada9166:
- Closes  [UNIFY--196](https://cypress-io.atlassian.net/browse/UNIFY-1196)

### User facing changelog

When renaming specs this logic needs to handle situations where there is a "collision" on rename such as this situation where a potentially renamed file would match an existing renamed file. In this case we swap out the extension but leave the basename intact.

app_spec.js -> app.cy.js
app_spec2.js -> app_spec2.cy.js
app2_spec.js -> app2.cy.js
app-spec2.js -> app-spec2.cy.js

image

### PR Tasks


tbiethman:
For the collision example, it feels odd to inject indexes for cases where using the basename would already be unique.

```
app.spec.js -> app.cy.js
app_spec.js -> app2_spec.cy.js
app-spec.js -> app3-spec.cy.js
```

Could we not just do:

```
app.spec.js -> app.cy.js
app_spec.js -> app_spec.cy.js
app-spec.js -> app-spec.cy.js
```

or even:

```
app.spec.js -> app.cy.js
app_spec.js -> app-1.cy.js
app-spec.js -> app-2.cy.js
```

brian-mann:
@tbiethman I based my algorithm off the desire to *always* remove the `spec` as a first priority so basically this:

- rename `spec` -> `cy`
- append indexes on collisions after renaming

Your algorithm appears to be:

- rename `spec` -> `cy` if there's no collision
- append indexes on collisions without renaming

brian-mann:
@tbiethman okay i see why you're confused - you're thinking the 2nd set of only two examples is the algorithm. It's not - the text above it are using those 2 as an example for when their ideal renamed state ALREADY MATCHES an existing spec. I did not list those existing matched specs - I just described the logic.

If it's easier - forget the bottom 2. The TOP 4 are really the situation. Work off of the top 4 and see if you can solve it in a better way using a different algorithm.

packages/launchpad/cypress/e2e/migration.cy.ts
```diff
@@ -743,6 +743,61 @@ describe('Full migration flow for each project', { retries: { openMode: 2, runMo
...
+    cy.contains('cypress/integration/app_spec.js')
+    cy.contains('cypress/integration/app-spec.js')
+
+    // after auto migration
+    cy.contains('cypress/e2e/app-spec2.cy.js')
+    cy.contains('cypress/e2e/app_spec2.cy.js')
+    cy.contains('cypress/e2e/app.cy.js')
+    cy.contains('cypress/e2e/app2.cy.js')
+    cy.contains('cypress/e2e/app_spec.cy.js')
+    cy.contains('cypress/e2e/app-spec.cy.js')
```
tbiethman, line 767: For these assertions, it's worth asserting the render order to ensure that the before and after names are mapped appropriately. Otherwise, I don't think your sorted extension prioritization has coverage.

```js
// these test ids don't currently exist
cy.get('[data-cy="before-file"]').eq(0).should('contain', 'cypress/integration/app-spec2.js')
cy.get('[data-cy="after-file"]').eq(0).should('contain', 'cypress/e2e/app-spec2.cy.js')
```

tbiethman (changes requested):
Great stuff @estrada9166, everything aligns with what we talked about last week. Just had the one comment around enhancing the test assertions.

tbiethman (approved):

packages/data-context/src/sources/MigrationDataSource.ts
```diff
@@ -189,4 +190,65 @@ export class MigrationDataSource {
   get configFileNameAfterMigration () {
     return this.ctx.lifecycleManager.legacyConfigFile.replace('.json', `.config.${this.ctx.lifecycleManager.fileExtensionToUse}`)
   }
+
+  private checkAndUpdateDuplicatedSpecs (specs: MigrationFile[]) {
+    const updatedSpecs: MigrationFile[] = []
+
+    const sortedSpecs = this.sortSpecsByExtension(specs)
+
+    sortedSpecs.forEach((spec) => {
```
lmiller1990, line 199: Been thinking about this function for a few days. I still think it's a good candidate for a unit test, the system tests are not exactly the best for debugging complex logic.

I think we could have done this with `reduce` and a `Set` to see if a spec already has been updated instead of the nested `_.find` in each loop, it'd be a little more efficient. This is okay though, but good to make small optimizations here and there.

Neither of these is a blocker on merging, though.

:heavy_check_mark:

lmiller1990 (approved):

PR #20654, opened 2022-03-16
feat: create default spec file name based on specPattern

estrada9166:
- Closes  [UNIFY-1344](https://cypress-io.atlassian.net/browse/UNIFY-1344)

### PR Tasks


packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -210,6 +211,57 @@ export class ProjectDataSource {
...
+
+      const parsedGlob = parseGlob(specPatternSet)
+
+      let dirname = parsedGlob.path.dirname
+
+      if (dirname.startsWith('**')) {
+        dirname = dirname.replace('**', 'cypress')
+      }
+
+      const splittedDirname = dirname.split('/').filter((s) => s !== '**').map((x) => this.formatGlob(x, 'e2e')).join('/')
```
lmiller1990, line 251: Do we need to use `path.sep` here? I wonder if this will be a problem for windows.

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -210,6 +211,57 @@ export class ProjectDataSource {
...
+
+    try {
+      if (!this.ctx.currentProject || !this.ctx.coreData.currentTestingType) {
+        return null
+      }
+
+      let specPatternSet: string | undefined
+      const { specPattern = [] } = await this.ctx.project.specPatternsForTestingType(this.ctx.currentProject, this.ctx.coreData.currentTestingType)
+
+      if (Array.isArray(specPattern)) {
```
lmiller1990, line 235: This method is so complicated, without unit tests I would be scared to touch this. I think this is a good candidate for some unit tests. There is a lot of conditionals, I feel like we cannot possibly be capturing all the branches in the one e2e test you added.

What if we extract a pure, stateless method? Something like:

```js
// not inside `ProjectDataSource` class directly, maybe declared outside?

export async function getDefaultSpecFileName (currentProject: string, currentTestingType: TestingType) : Promise<string> {
  let specPatternSet: string | undefined
  // ... all the things ...
  return splittedDirname + fileName + extname + ext
}
```

Unit test this, than wrap it in `ProjectDataSource#defaultSpecFileName`.

```js
async defaultSpecFileName () {
  const defaultFileName = 'cypress/e2e/filename.cy.js'

  if (!this.ctx.currentProject || !this.ctx.coreData.currentTestingType) {
    return null
  }

  try {
    return getDefaultSpecFileName(this.ctx.currentProject, this.ctx.coreData.currentTestingType)
  } catch (e) {
    return defaultFileName
  }
}
```

What do you think? We definitely lack a lot of coverage in `sources`, this seems like a good place to start adding some.

lmiller1990 (commented):
Seems the code functions but I feel we are missing quite a bit of coverage - or, at least, it's only covered in E2E tests, which makes it a bit hard to track down bugs, as well as figure out what `defaultSpecFileName` actually does.

What do you think about spending a bit of time adding some unit tests, and maybe a comment explaining what this method is doing? See my idea below.

packages/data-context/src/sources/ProjectDataSource.ts
lmiller1990: Right yes, I think this is correct, we always expect `/` in config files.

packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
```diff
@@ -189,3 +189,92 @@ describe('findSpecs', () => {
...
+
+      expect(defaultFileName).to.eq('cypress/e2e/filename.cy.ts')
+    })
+
+    it('if project use TS, set TS as extension if it exists in the options of extensions', () => {
+      const defaultFileName = getDefaultSpecFileName('cypress/e2e/filename.cy.{js,ts,tsx}', 'ts')
+
+      expect(defaultFileName).to.eq('cypress/e2e/filename.cy.ts')
+    })
+
```
lmiller1990, line 267: Nice job on these tests, super useful and makes the actual implementation much clearer.

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -101,6 +102,40 @@ export function transformSpec ({
   }
 }
 
+export function getDefaultSpecFileName (specPattern: string, fileExtensionToUse?: 'js' | 'ts') {
+  function formatGlob (s: string, fallback: string) {
+    let [cleanedVal] = s.replace('{', '').replace('}', '').split(',')
```
lmiller1990, line 107: I'm kind of surprised this works - how come we don't need either `replaceAll` or `replace('{', //g)`? I see you have a test with mutiple `{` and `}` and it works just fine, but I don't see anywhere that handles replacing multiple `{` in a single glob. Did I miss something?

lmiller1990: Test I was referring to with multiple `{`: https://github.com/cypress-io/cypress/pull/20654/files#diff-59bf8023d3f96e2d91bc479be5372ebde0317cf6040e5e514ceb276ac71c1342R221

lmiller1990 (dismissed):
Code and tests seem fine - just have a question!

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -210,6 +245,31 @@ export class ProjectDataSource {
...
+    try {
+      if (!this.ctx.currentProject || !this.ctx.coreData.currentTestingType) {
+        return null
+      }
+
+      let specPatternSet: string | undefined
+      const { specPattern = [] } = await this.ctx.project.specPatternsForTestingType(this.ctx.currentProject, this.ctx.coreData.currentTestingType)
+
+      if (Array.isArray(specPattern)) {
+        specPatternSet = specPattern[0]
```
emilyrohrbough, line 260: We should make a notes in the documentation that the first pattern provided in the specPattern configuration will be used to generate a spec file from the Cypress UI.

cc/ @cowboy

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -210,6 +245,31 @@ export class ProjectDataSource {
     this._specWatcher.on('unlink', onSpecsChanged)
   }
 
+  async defaultSpecFileName () {
+    const defaultFileName = 'cypress/e2e/filename.cy.js'
```
emilyrohrbough, line 249: Is this fallback needed? Won't our default specPattern be parsed to determine this value from `getDefaultSpecFileName`?

packages/app/src/specs/CreateSpecModal.vue
```diff
@@ -99,7 +100,9 @@ const helpLink = computed(() => {
 const specFileName = computed(() => {
   const extension = props.gql.currentProject?.fileExtensionToUse ?? 'js'
 
-  return getPathForPlatform(`cypress/e2e/filename.cy.${extension}`)
+  const fileName = props.gql.currentProject?.defaultSpecFileName ?? `cypress/e2e/filename.cy.${extension}`
+
```
emilyrohrbough, line 104: Would it be better to provide a warning message instead of scaffolding a new spec that didn't match the project's specPattern if `props.gql.currentProject?.defaultSpecFileName` is undefined?

packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
```diff
@@ -189,3 +189,92 @@ describe('findSpecs', () => {
...
+      expect(defaultFileName).to.eq('cypress/foo.spec.ts')
+    })
+
+    it('replace ** for cypress if it starts with **', () => {
+      const defaultFileName = getDefaultSpecFileName('**/e2e/foo.spec.ts')
+
+      expect(defaultFileName).to.eq('cypress/e2e/foo.spec.ts')
+    })
+
+    it('replace ** for cypress if it starts with ** and omit extra **', () => {
```
emilyrohrbough, line 214: Can we add a test for something like `e2e/**/*.cy.{js,jsx,ts,tsx}`

packages/app/cypress/e2e/specs.cy.ts
```diff
@@ -383,6 +383,67 @@ describe('App: Index', () => {
...
+              req.on('before:response', (res) => {
+                res.body.data.matchesSpecPattern = true
+              })
+            }
+          })
+
+          cy.get('@enterSpecInput').type('x')
+          cy.contains(defaultMessages.createSpec.e2e.importEmptySpec.specExtensionWarning)
+          cy.percySnapshot('Non-recommended spec pattern warning')
+          cy.contains('span', '{filename}.cy.jx')
```
emilyrohrbough, line 429: Is this the correct message? Wondering if the test should have been `cypress/e2e/MyTest.cy.js`
Screen Shot 2022-03-22 at 9 51 25 AM

Also looking while running looking and looking at this test, the spec file name generate (`cypress/e2e/*.spec.{js,jsx]}`) does not match the custom spec pattern (`src/**/*.cy.{js,jsx}`), however the test is passing.
Screen Shot 2022-03-22 at 9 52 27 AM

emilyrohrbough (changes requested):

packages/data-context/src/sources/ProjectDataSource.ts
cowboy: Is there an explanation of how the pattern will be used to generate a spec file name?

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -101,6 +102,40 @@ export function transformSpec ({
   }
 }
 
+export function getDefaultSpecFileName (specPattern: string, fileExtensionToUse?: 'js' | 'ts') {
```
cowboy, line 105: Here are a few globs where `getDefaultSpecFileName()` doesn't return what I'd expect. I expect that there are more, that I haven't thought of. Hopefully we're considering all the patterns that the glob library we're using supports and looking at actual user glob patterns when we're writing this code and tests.

```
tests/*.js
src/*-test.js
src/*.foo.bar.js
src/prefix.*.test.js
src/*/*.test.js
src-*/**/*.test.js
src/*.test.(js|jsx)
(src|components)/**/*.test.js
```

estrada9166: all these cases should be covered now - happy to add more cases if you have in mind!

https://github.com/cypress-io/cypress/pull/20654/commits/5deb317f0cf72ae584ef385345a92c43e8bf80ef#diff-59bf8023d3f96e2d91bc479be5372ebde0317cf6040e5e514ceb276ac71c1342R281-R339

packages/data-context/src/sources/ProjectDataSource.ts
estrada9166: It's being used if we are not able to generate the fileName, or there's an unexpected error!

packages/app/src/specs/CreateSpecModal.vue
estrada9166: The warning message can be the someone as this one https://github.com/cypress-io/cypress/pull/20654#discussion_r832633978 where the spec name to create do not match - what do you think?

packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
estrada9166: https://github.com/cypress-io/cypress/commit/5deb317f0cf72ae584ef385345a92c43e8bf80ef#diff-59bf8023d3f96e2d91bc479be5372ebde0317cf6040e5e514ceb276ac71c1342R334-R338

lmiller1990 (approved):

packages/app/src/specs/CreateSpecModal.vue
emilyrohrbough: +1 I think a warning message like you linked would better-align with what users would expect.

system-tests/projects/no-specs-custom-pattern/cypress.config.js
```diff
@@ -11,6 +11,6 @@ module.exports = {
   },
   e2e: {
     supportFile: false,
-    specPattern: 'src/**/*.cy.{js,jsx}',
+    specPattern: 'src/**/*.{cy,spec}.{js,jsx}',
```
emilyrohrbough, line 14: Did you intent to check this in?

emilyrohrbough (dismissed):

system-tests/projects/no-specs-custom-pattern/cypress.config.js
estrada9166: Yup, so, the new test makes sense - waiting the CI to check if it breaks something - but the current spec file passed

emilyrohrbough (approved):

PR #20656, opened 2022-03-17
chore: [Multi-domain] Clean up stability code.

mjhenkes:
- Closes 

### User facing changelog

n/a

### Additional details

On further inspection, we should not get into a state where the AUT loads before the spec bridge is active due to our delaying logic. Given this we can remove some code trying to sync a load even that happened if the page loaded before the specbridge was active.

### PR Tasks

### DIFF
--- a/packages/driver/cypress/integration/e2e/multi-domain/navigation_spec.ts
+++ b/packages/driver/cypress/integration/e2e/multi-domain/navigation_spec.ts
@@ -246,12 +246,8 @@ describe('sub domain window loads before multi-domain command', { experimentalSe
   })
 
   it('establish a foobar spec bridge', () => {
-    return new Promise<void>((resolve) => {
-      setTimeout(() => {
-        cy.switchToDomain('http://foobar.com:3500', () => {
-          cy.get('a[data-cy="multi-domain-page"]')
-        }).then(resolve)
-      }, 1000)
+    cy.switchToDomain('http://foobar.com:3500', () => {
+      cy.get('a[data-cy="multi-domain-page"]')
     })
   })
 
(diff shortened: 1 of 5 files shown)


packages/driver/cypress/integration/e2e/multi-domain/navigation_spec.ts
```diff
@@ -237,3 +237,31 @@ describe('delayed navigation', { experimentalSessionSupport: true, defaultComman
...
+        cy.switchToDomain('http://foobar.com:3500', () => {
+          cy.get('a[data-cy="multi-domain-page"]')
+        }).then(resolve)
+      }, 1000)
+    })
+  })
+
+  it('page load before switchToDomain command when spec bridge is already established', () => {
+    return new Promise<void>((resolve) => {
+      setTimeout(() => {
```
mjhenkes, line 256: Since i cannot force the page to load faster, i'm delaying switchToDomain. Originally i had a test like this above, but when the spec bridge does not exist, we delay loading cross origin requests for 2000 ms to ensure the bridge is created before the request loads, so forcing a delay seemed invalid.

mjhenkes: Actually this got me curious in that we should delay for 2000 ms regardless of the state of the spec bridge.. With this PR i think i'm fixing a symptom, not the cause. It appears that the config tests are running so quickly we're getting messages mixed up, we send ready for domain and by the time we release the domain to load we've already moved on to the next test.

packages/driver/src/multi-domain/domain_fn.ts
```diff
@@ -10,7 +10,7 @@ interface RunDomainFnOptions {
   env: Cypress.ObjectLike
   fn: string
   skipConfigValidation: boolean
-  state: {}
+  state: { isStable: boolean | undefined }
```
mjhenkes, line 13:
```suggestion
  state: {}
```

packages/driver/cypress/integration/e2e/multi-domain/navigation_spec.ts
mjhenkes: The actual problem was elsewhere in another branch.

mschile (approved):

chrisbreiding (approved):

PR #20657, opened 2022-03-17
feat: add subscription for specs changing

ryanthemanuel:
- Closes [UNIFY-1347](https://cypress-io.atlassian.net/browse/UNIFY-1347) and [UNIFY-1275](https://cypress-io.atlassian.net/browse/UNIFY-1275)

### User facing changelog

Ensure that when the spec pattern changes inside of the cypress config, we start watching the new spec pattern in the file system for changes

### Additional details

In addition to fixing the spec pattern watching logic, we are also converting specs to use the gql subscription system. In the process of working on this, I realized that we had a couple of race conditions involving the subscription async iterator, where we might get events while we're not actually awaiting them yet and thus they get lost. Also, when stepping through the `graphql-ws` code, I realized that when we are "returning" out of the async iterator, so I am fixing that as well by ensuring we always resolve the promise we are waiting on in the subscription.

### PR Tasks


packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -458,7 +452,13 @@ export class ProjectActions {
 
     this.ctx.actions.project.setSpecs(specs)
 
-    return { specs, specPattern, excludeSpecPattern, additionalIgnorePattern }
+    this.ctx.project.startSpecWatcher(path, testingType, specPattern, excludeSpecPattern, additionalIgnorePattern)
+
+    if (testingType === 'component') {
+      this.api.getDevServer().updateSpecs(specs)
```
ZachJW34, line 458: Seems we have a pattern whenever we add/remove a spec:
* call `setSpecs()`
* call `devServer.updateSpecs()`
* trigger emitter

Should we move all this into `setSpecs`?

packages/app/cypress/e2e/subscriptions/specChange-subscription.cy.ts
```diff
@@ -0,0 +1,357 @@
...
+        cy.withCtx(async (ctx, o) => {
+          await ctx.actions.file.writeFileInProject(o.path, '')
+        }, { path: getPathForPlatform('cypress/e2e/new-file.spec.js') })
+
+        cy.get('[data-cy="spec-item-link"]')
+        .should('have.length', 1)
+        .should('contain', 'new-file.spec.js')
+      })
+
+      it('handles removing the last file', () => {
```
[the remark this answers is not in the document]
ryanthemanuel: @ZachJW34, the test before this one accomplishes that.

packages/data-context/src/actions/ProjectActions.ts
ryanthemanuel: Good call I moved this logic.

packages/data-context/src/actions/DataEmitterActions.ts
```diff
@@ -62,34 +71,59 @@ export class DataEmitterActions extends DataEmitterEvents {
...
+        }
+
+        return pending.shift()
+      },
+      throw: async (error: Error) => {
+        throw error
       },
       return: async () => {
         this.pub.off(evt, subscribed)
+        if (dfd) {
```
ryanthemanuel, line 117:
```suggestion
        // If we are currently waiting on a deferred promise, we need to resolve it and signify we're done to ensure that the async loop terminates
        if (dfd) {
```

packages/data-context/src/actions/DataEmitterActions.ts
```diff
@@ -62,34 +71,59 @@ export class DataEmitterActions extends DataEmitterEvents {
...
     const iterator = {
       async next () {
+        if (done) {
+          return { done: true, value: undefined }
+        }
+
         if (!hasSentInitial && sendInitial) {
           hasSentInitial = true
 
           return { done: false, value: {} }
```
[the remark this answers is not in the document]
tgriesser: The value here is what will flow into the first argument of the `resolve` for the subscription. So if we want to have a more targeted subscription where we're not looking to merge with existing types, like notifying where the type is dependent on an individual file change (where there is no initial value), we can set `sendInitial` to false, and just use the values that are sent into the emitter to determine the resolved value. Here's a rough example of what I mean:

```graphql
type SpecEventType {
  ADD
  CHANGE
  REMOVE
}

type SpecEvent {
  event: SpecEventType
  filePath: String
  fileParts: FileParts
}

type Subscription {
  """Fired whenever a spec changes"""
  specEvent: SpecEvent
}
```


```ts
subscribe: (source, args, ctx) => ctx.subscribeTo('specEvt', false)
resolve: (specEvt, args, ctx) => {
  return {...specEvt, fileParts: {absolute: specEvt.filePath}}
}
```

```ts
on.all('evt', (event, filePath) => {
  ctx.emitter.specEvt({
     event,
     filePath
  })
}
```

guides/graphql-subscriptions.md
```diff
@@ -81,7 +81,8 @@ useSubscription({ query: OnSpecChangeDocument }, (prev, next) => {
 - [API Docs for useSubscription](https://formidable.com/open-source/urql/docs/api/vue/#usesubscription)
 - [Subscriptions overview](https://formidable.com/open-source/urql/docs/advanced/subscriptions/)
 
- One thing to be aware of, is the subscription is only mounted/responded to when the containing component is mounted on the page.
+One thing to be aware of, is the subscription is only mounted/responded to when the containing component is mounted on the page. Therefore, one rule of thumb when using subscriptions is to ensure that they are declared in the subscription dependent view that is highest in the hierarchy so that the subscription can be active when it needs to be. This location is often alongside a query rather than a fragment.
+
```
tgriesser, line 85: Great addition here!

packages/data-context/src/actions/DataEmitterActions.ts
```diff
@@ -62,34 +71,60 @@ export class DataEmitterActions extends DataEmitterEvents {
...
+          dfd = pDefer()
 
-        return { done: false, value: await dfd.promise }
+          return await dfd.promise
+        }
+
+        return pending.shift()
+      },
+      throw: async (error: Error) => {
+        throw error
```
tgriesser, line 113: Does adding `throw` here make a difference? Or was this more useful while debugging the timing issue with the deferreds?

packages/server/lib/server-ct.ts
```diff
@@ -21,7 +21,7 @@ export class ServerCt extends ServerBase<SocketCt> {
       this.server.on('connect', this.onConnect.bind(this))
       this.server.on('upgrade', (req, socket, head) => this.onUpgrade(req, socket, head, socketIoRoute))
 
-      graphqlWS(this.server, `${socketIoRoute}-graphql`)
+      this._graphqlWS = graphqlWS(this.server, `${socketIoRoute}-graphql`)
```
tgriesser, line 24: Nice catch!

packages/data-context/src/actions/DataEmitterActions.ts
ryanthemanuel: This was to complete the expected type for the async generator. From what I could tell in the graphql-ws an async generator seemed to be what they were expecting if you were going to expose `return` like we are:

https://github.com/enisdenjo/graphql-ws/blob/7649ad739291a6bfb031bdc4b3487291568fe8a8/src/server.ts#L827-L828

Declaring it as an async generator like I am now, allows me to iterate like I am in the test i wrote but it also necessitates adding the throw function.

ZachJW34 (approved):

packages/app/cypress/e2e/subscriptions/specChange-subscription.cy.ts
```diff
@@ -0,0 +1,357 @@
+import defaultMessages from '@packages/frontend-shared/src/locales/en-US.json'
+
+function getPathForPlatform (posixPath: string) {
+  if (Cypress.platform === 'win32') return posixPath.replaceAll('/', '\\')
+
+  return posixPath
+}
```
tbiethman, line 7: This helper has been propagated a few times already, but you could import it from here instead: https://github.com/cypress-io/cypress/blob/10.0-release/packages/app/src/paths.ts

packages/app/cypress/e2e/subscriptions/specChange-subscription.cy.ts
```diff
@@ -0,0 +1,357 @@
...
+    devServerConfig: {
+      webpackFilename: 'webpack.config.js',
+    },
+  },
+  e2e: {
+    specPattern: 'cypress/e2e/**/dom-cont*.spec.{js,ts}',
+    supportFile: false,
+  },
+}`)
+        }, { path: getPathForPlatform('cypress/e2e/dom-list.spec.js') })
```
tbiethman, line 250: Doesn't look like the path is being used here.

tbiethman (approved):
Works good! My comments are small.

Not sure if you've done much testing on windows, but I'm curious if this might have an impact on https://cypress-io.atlassian.net/browse/UNIFY-1221 as well.

packages/app/cypress/e2e/subscriptions/specChange-subscription.cy.ts
ryanthemanuel: [`574a705` (#20657)](https://github.com/cypress-io/cypress/pull/20657/commits/574a705437ef800269ff00682da405dfe5c858c2)

tgriesser (approved):

tbiethman (approved):

PR #20474, opened 2022-03-03
feat(Unification): add internal log group api and consume in .within()

emilyrohrbough:
- Enhances # 20433 but a follow up ticket will be logged to determine/capture remaining work. This PR may be closed and merged with #20465 

### Summary
- Added log group api to be consumed internally by Cypress commands.
- Updated `.within()` to consume log group api
- Update Reporter's Command Group to be open by default for parent or child commands
- Fixed session's 'warn' state to be 'warning' to align with other 'warning' states.

### User facing changelog
Enhanced `.within()` to provide visual indication of nested commands and logs.

### How has the user experience changed?

Before | After
-- | --
Screen Shot 2022-03-04 at 4 36 44 PM| Screen Shot 2022-03-04 at 4 33 17 PM

### PR Tasks

- [x] Have tests been added/updated?
- [x] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [n/a] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [n/a] 	Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [n/a] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

PR #17069, opened 2021-06-23
chore(runner): render spec list inline (behind feature flag)

lmiller1990:
After lots of shuffling of code in https://github.com/cypress-io/cypress/pull/16866 and https://github.com/cypress-io/cypress/pull/16981, it's now much easier for the two runners to share features and components. This PR adds an inline **SpecList** from CT in the **E2E runner**. It's behind a **feature flag**, `CypressInternal_UseInlineSpecList`, which is turned **off by default**. So it is **not user facing** yet (nor will it be, for a while).

If you'd like to try it out, you can enable it like this:

```json
{
  "env": {
    "CypressInternal_UseInlineSpecList": true
  }
}
```

The main problem I see moving forward is the need to settle on a shared layout. CT uses `<SplitPane>`, e2e uses a jQuery + mobx state solution, both have some problems. Ideally, we should be able to find something that doesn't require us to add new variables every time we'd like to modify the UI.

image

Demo (including resizing):

https://user-images.githubusercontent.com/19196536/123042891-d617ae80-d43a-11eb-9e07-3ed42c276e34.mov

PR #20663, opened 2022-03-17
chore: delete unsed code in runner, runner-shared, runner-ct

lmiller1990:
Deleting some unused code, round 1. I did not delete any of the E2E tests in `runner` yet, since we are still auditing those and porting them to the new `packages/app` runner. The eventual goal will be deleting all of `runner`, `runner-shared` and `runner-ct`, which seems like it should be done in the near future. Apart of: https://cypress-io.atlassian.net/browse/UNIFY-898 

Percy is showing the diff against develop for some reason - not sure what's going on here.


lmiller1990:
@BlueWinds @emilyrohrbough This should be good to review now. I tried deleting a bunch of scss too, but some of the ones in `runner` seem to be somehow used in `reporter` (intention or not, not sure). 

I'll be making a few more of these "delete all the things" PRs in the next week or two.

emilyrohrbough:
@lmiller1990 Theres still quite a bit of code left in both `runner-ct`, `runner` and `runner-shared`. I assume some day this will merge into one, just not today. Since you just went through this, mind sharing a quick status update of the current high-level state of where these are at? Looking at each read-me's it says "todo: delete this" so I'm curious what's left.

lmiller1990:
@emilyrohrbough, updated the READMEs a bit and deleted some more code.

The blocker on eliminating the three runner packages all together is the fact that we cannot import `@cypress/driver` into Vite. Webpack 4 + our webpack configuration is super customized to handle many things like builtin node modules, various aliases and transforms of JS + CSS + experimental decorators, which Vite does not understand.

I spiked into it a bit to see what happened, here's the various things I encountered:

- in driver at the top we do `import 'setimmediate`. This seems ES module incompatible. I fixed it by removing the import so I could progress.
- old versions of sinon and sinonjs/fake-timers. Fixed by combo of updating + patch package.
- error with `dom.js` using inline webpack loaders to load SCSS. Fixed by removing that code (well, just so I could progress and see what would be next)
- various errors with node dependencies expecting `global` and `process` to exist. Webpack polyfills these, but Vite does not. I could process by using various combinations with Vite's `define` config: [docs](https://vitejs.dev/config/#define)
- `Buffer` is not defined  - we are using the cjs build of mocha, which assumes `Buffer` exists.
- `process` does not exist - Polyfill it...

Basically this is leading to rebuilding our webpack config in Vite, which is not good. The idea of Vite is one true native module system (ES modules). We need to audit to core of `driver` and make it a true ES module, that does not rely on webpack's various polyfills or hacks to run in a browser. All the libs we use (mocha, sinon etc) have modern. ES builds, so it should be doable, but based on my initial audit and previous attempts this on the level of the webpack 5 migration in terms of complexity and surface area.

packages/reporter/src/main.tsx
```diff
@@ -18,7 +18,7 @@ import shortcuts from './lib/shortcuts'
 import Header, { ReporterHeaderProps } from './header/header'
 import Runnables from './runnables/runnables'
 import TestingPreferences from './preferences/testing-preferences'
-import { BaseStore } from '@packages/runner-shared/src/store'
+import type { MobxRunnerStore } from '@packages/app/src/store/mobx-runner-store'
```
emilyrohrbough, line 21: Does this create a circular dependency?

packages/app/src/runner/event-manager.ts
```diff
@@ -131,11 +130,6 @@ export class EventManager {
     })
 
     this.ws.on('specs:changed', ({ specs, testingType }) => {
-      // do not emit the event if e2e runner is not displaying an inline spec list.
-      if (testingType === 'e2e' && state.useInlineSpecList === false) {
-        return
-      }
-
```
emilyrohrbough, line 138: should this have been maintained? I'm not seeing a test around this, is there an existing one?

packages/reporter/src/main.tsx
```diff
@@ -89,7 +89,7 @@ class Reporter extends Component<SingleReporterProps> {
   // this hook will only trigger if we switch spec file at runtime
   // it never happens in normal e2e but can happen in component-testing mode
   componentDidUpdate (newProps: BaseReporterProps) {
-    this.props.runnablesStore.setRunningSpec(this.props.runnerStore.spec.relative)
+    this.props.runnablesStore.setRunningSpec(this.props.runnerStore.spec?.relative ?? null)
```
emilyrohrbough, line 92: Is it possible to update the setRunnerSpec method to accept undefined/null?

packages/reporter/src/runnables/runnables-store.ts
```diff
@@ -200,7 +200,7 @@ export class RunnablesStore {
   }
 
   @action
-  setRunningSpec (specPath: string) {
+  setRunningSpec (specPath: string | null) {
```
emilyrohrbough, line 203: when the specPath is null, should we be storing the history?

packages/runner-shared/README.md
```diff
@@ -1,20 +1,6 @@
...
-The components are imported to the [`runner`](../runner/README.md#Developing)  and `runner-ct` packages respectively. Please see their instructions for develoment.
-
-## Testing
-
-### Unit Tests
-
-```bash
-yarn workspace @packages/runner-shared test
-```
+1. Contains `dom.js`, which uses propietary webpack loaders and cannot easily be imported with Vite (dev server in `@packages/app`). This is bundled via webpack in either `@packages/runner` or `@packages/runner-ct`. Once `dom.js` is free of webpack-specific loader code, we should move it to `@packages/app`.
```
emilyrohrbough, line 5:
```suggestion
1. Contains `dom.js`, which uses proprietary webpack loaders and cannot easily be imported with Vite (dev server in `@packages/app`). This is bundled via webpack in either `@packages/runner` or `@packages/runner-ct`. Once `dom.js` is free of webpack-specific loader code, we should move it to `@packages/app`.
```

packages/runner/src/lib/state.js
```diff
@@ -1,5 +1,6 @@
 import { action, computed, observable } from 'mobx'
-import { automation, BaseStore } from '@packages/runner-shared'
+import { MobxRunnerStore } from '@packages/app/src/store/mobx-runner-store'
```
emilyrohrbough, line 2: Does this create a circular dependency?

packages/server/lib/html/non_proxied_error.html
```diff
@@ -6,7 +6,46 @@
 
     <link href="/__cypress/runner/favicon.ico?v2" rel="icon">
 
-    <link rel="stylesheet" href="/__cypress/runner/cypress_runner.css">
+    <style>
```
emilyrohrbough, line 9: It looks like these styles might have been updated , thus the updated screenshot. Was that intentional?

packages/runner-ct/webpack.config.ts
```diff
@@ -80,10 +80,6 @@ const config: webpack.Configuration = {
 config.plugins = [
   // @ts-ignore
   ...config.plugins,
-  new HtmlWebpackPlugin({
```
emilyrohrbough, line 83: We no longer need an entry point to load ct tests?

packages/reporter/src/main.tsx
lmiller1990: It does not - `@packages/app` does not import from `@packages/reporter`. Even if it did, this would not be a problem since we are doing `import type` which just grabs the definition - it won't import an actual JS module. `import type` is a really nice addition to TS.

packages/app/src/runner/event-manager.ts
lmiller1990: Good question... here's the historical context. Before it was decided to go all in on the Vite/Vue rewrite, I wanted to just rework the existing app without touching the stack (React, Webpack, event emitter instead of GraphQL). I actually added an inline spec list to E2E as part of that, which was hidden behind a feature flag that never saw the light of day. The idea was `runner-shared` would have 99% of the code, the `runner` and `runner-ct` would have been the 1% of code differing between the two.

You can see where that was added and a sweet demo here: https://github.com/cypress-io/cypress/pull/17069. There is a test that isn't running, since we are not running the `runner` tests on CI (since we are not using that code anymore). For interest, the test is here: https://github.com/cypress-io/cypress/blob/develop/packages/runner/cypress/integration/runner.ui.spec.js#L409.

I'm still wondering if reworking the existing infrastructure with React/Webpack would have been the better move. :man_shrugging:.

To answer your question: this code is now obsolete, since we will always be showing an inline spec list in the runner, so that's why I deleted it.

packages/reporter/src/main.tsx
lmiller1990: It is, but this leads to needing to change a bunch of things in reporter that expect `null`. This change leads to the least amount of code changes.

This was not nullable before, since the type was overrode here: https://github.com/cypress-io/cypress/pull/20663/files#diff-8ec8396907f7c8abf3999a7ec920f7210d6730c10f9a30c9df87253b0b28c125L37 like this:

```ts
runnerStore: BaseStore & {spec: Cypress.Spec, specRunId: string}
```

I changed this to avoid having multiple types for the store spread over the app. If `spec` is really not allowed to be null here, I think it would make sense to pass it as a separate prop, rather than requiring an entire store with several type changes, so we don't end up in a situation where the `BaseStore` is augmented all over the place.

It's not entirely clear to me that we can actually guarantee `spec` is not null, and I didn't want to change too much here, so this seemed like the past of least resistance/lowest risk.

packages/reporter/src/runnables/runnables-store.ts
lmiller1990: See comment: https://github.com/cypress-io/cypress/pull/20663/files/be5b95a3d923368e80fae240c7a2e0740be4da07#r832425950. I've reverted this change since I don't know the answer, but my understanding is spec can never be null (not sure how strictly this type safety is actually exercised in the entire system, but the current code works, so it's probably safest not to touch this, after all).

packages/runner/src/lib/state.js
lmiller1990: I realized this file is also not used, so I'm just going to delete it. Can't have a circular dep without this file :100:

lmiller1990: https://github.com/cypress-io/cypress/pull/20663/commits/02f75439e3fb8dfb640e254aa9b4d1f505f23f29

packages/server/lib/html/non_proxied_error.html
lmiller1990: It was - I deleted a bunch of styles as part of this PR. The page originally contained all of `cypress_runner.css`, which included all the styles for the entire runner and reporter, which are really not necessary for this simple page.

Instead of going over literally thousands of lines of CSS to figure out the exact thing that was messing up the screenshot, I just committed a new one, that looks almost identical, but the text is a tiny bit smaller. I think this is a good compromise - we remove thousands of lines of otherwise-unused CSS, and the only visual difference is a page is almost never seen, which just shows a basic warning.

packages/runner-ct/webpack.config.ts
lmiller1990: Nope - this is only used when developing `runner-ct`, which we do not do anymore. We now get the index file here, for app: https://github.com/cypress-io/cypress/blob/10.0-release/packages/data-context/src/sources/HtmlDataSource.ts#L60

emilyrohrbough: Just curious, does this mean we no longer need the `HtmlWebpackPlugin` for the runner package either?

packages/app/src/runner/event-manager.ts
emilyrohrbough: huh! thanks for all the info! Sounds like this was the first step toward unification's shared spec list! I would LOVE if we could run e2e & component tests within the same frame. Maybe some day 😄

packages/reporter/src/main.tsx
```diff
@@ -89,7 +89,11 @@ class Reporter extends Component<SingleReporterProps> {
   // this hook will only trigger if we switch spec file at runtime
   // it never happens in normal e2e but can happen in component-testing mode
   componentDidUpdate (newProps: BaseReporterProps) {
-    this.props.runnablesStore.setRunningSpec(this.props.runnerStore.spec?.relative ?? null)
+    if (!this.props.runnerStore.spec) {
+      throw Error(`Expected runnerStore.spec and runnerStore.specRunId not to be null.`)
```
emilyrohrbough, line 93:
```suggestion
      throw Error(`Expected runnerStore.spec not to be null.`)
```

packages/reporter/src/main.tsx
```diff
@@ -103,6 +107,10 @@ class Reporter extends Component<SingleReporterProps> {
   componentDidMount () {
     const { appState, runnablesStore, runner, scroller, statsStore, autoScrollingEnabled, isSpecsListOpen, runnerStore } = this.props
 
+    if (!runnerStore.spec) {
+      throw Error(`Expected runnerStore.spec and runnerStore.specRunId not to be null.`)
```
emilyrohrbough, line 111:
```suggestion
      throw Error(`Expected runnerStore.spec not to be null.`)
```

emilyrohrbough (approved):
Looks good to me! Huge thank you for cleaning this up! It feels amazing 🤩

packages/app/src/runner/event-manager.ts
lmiller1990: Not sure on the technical complexity/value but it sure would be nice. Obvious problem is we'd need to start the dev server regardless, since we wouldn't know if you are using E2E or CT. Another battle for another day.

packages/runner-ct/webpack.config.ts
lmiller1990: I'd say so. I deleted it.

lmiller1990:
One of the Percy screenshots looks wrong, but it's actually fine when I run locally.

image

BlueWinds (approved):

emilyrohrbough:
@lmiller1990 I'm super confused why you are seeing this. This is a reflection of the work I've been doing on #20474....

emilyrohrbough (approved):

PR #20677, opened 2022-03-17
fix: scope config to current testing type

estrada9166:
- Closes  [UNIFY-1369](https://cypress-io.atlassian.net/browse/UNIFY-1369)

### PR Tasks


packages/config/src/options.ts
```diff
@@ -98,6 +98,11 @@ const isValidConfig = (key: string, config: any) => {
   return true
 }
 
+export const defaultSpecPattern = {
```
emilyrohrbough, line 101: Can we update the e2e & component configuration values to return these?

packages/app/cypress/e2e/settings.cy.ts
```diff
@@ -209,8 +209,8 @@ describe('App: Settings', () => {
       cy.get('[data-cy="config-code"]').within(() => {
         cy.get('.bg-teal-100').contains('tests/_fixtures')
         cy.get('.bg-teal-100').contains('abc123')
-        cy.get('.bg-teal-100').contains('specFilePattern')
-        cy.get('.bg-teal-100').contains('supportFile')
+        cy.get('.bg-teal-100').contains('tests/**/*')
+        cy.get('.bg-teal-100').contains('tests/_support/spec_helper.js')
```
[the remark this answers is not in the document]
estrada9166: The reason is that now we are not returning the values as a nested property of the `e2e` or `component` and bc it was set as from `config` the whole object was highlighted

packages/config/src/options.ts
```diff
@@ -500,6 +505,12 @@ const runtimeOptions: Array<RuntimeConfigOption> = [
     validation: validate.isString,
     isInternal: true,
     canUpdateDuringTestTime: false,
+  }, {
+    name: 'additionalIgnorePattern',
```
emilyrohrbough, line 509: Looks like a new config value I had missed. Will add to the changelog. Are others able to add to this or is this internal to Cypress? I'd expect the excludeSpecPattern to be leverage instead of needing to add another option as well.

Can we scope this to the component config? it appears to be specific to component testing.

emilyrohrbough: Looking below, do we need this at all? Seems like we can just pull the specPattern and it'll resolve to either the default e2e specPttern or the custom specPattern provided: https://github.com/cypress-io/cypress/pull/20677/files#diff-0349e7b425446de9ccaf2edbb623f43cffd420688dd312850417888318b9f609R162

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -217,7 +217,7 @@ export class ProjectDataSource {
 
     const MINIMATCH_OPTIONS = { dot: true, matchBase: true }
 
-    const { specPattern = [], excludeSpecPattern = [] } = await this.ctx.project.specPatternsForTestingType(this.ctx.currentProject, this.ctx.coreData.currentTestingType)
+    const { specPattern = [], excludeSpecPattern = [] } = await this.ctx.project.specPatterns()
```
emilyrohrbough, line 220: If we don't remove the ignoreSpecPattern config option, shouldn't it be taken into account here?

packages/server/lib/config.ts
```diff
@@ -203,6 +211,11 @@ export function mergeDefaults (
     throw errors.get(err, ...args)
   })
 
+  delete config['e2e']
```
emilyrohrbough, line 214: Can you add a comment on why this is being done? I doesn't feel obvious if I were to read this later.

packages/config/src/options.ts
estrada9166: It needs to be exported separately because it replaces [this](https://github.com/cypress-io/cypress/pull/20677/files#diff-635c05fe0a204b142186818a9bc2c31711cacf2059208cb8421b4eb264872c29) file that was removed - also we can't get the values from the resolved config bc there's no gonna be nested properties based on the testing type - just the top-level ones, so we do not now the default value if it's set on the config file and also we may not be able to know the default value of the testing type that is not being selected

packages/config/src/options.ts
estrada9166: This value is only used internally when we are getting the specs for the CT we want to ignore as well the specPattern that is set on the e2e property, the reason why we can just pull the specPattern is bc know that we are removing the nested properties, we are not gonna be able to know what was the value set on the e2e, the one that is going to be on the top level is the one of the CT - so we may end up ignoring the specPattern that we are looking for

emilyrohrbough (approved):

tgriesser (approved):

PR #20558, opened 2022-03-10
feat: Add initial structure to support GraphQL subscriptions

tgriesser:
Closes: UNIFY-885

Broken out from #20488 this is the core work needed to support subscriptions at the protocol level, with documentation & reference implementation / guide for testing.

- Adds `graphql-ws`
- Adds `authChange` & `devChange` subscriptions
- Adds [documentation](https://github.com/cypress-io/cypress/blob/28577ebd7fbba51f20375eea3e34a1ac5e454c67/guides/graphql-subscriptions.md) for subscriptions

Bonus:
- Removes `currentTestingType` from the root query field (still exists on `CurrentProject`)
---
Todo: 
- [x] Remove ping subscription
- [x] Add basic subscriptions for devEvents
- [x] Tests

PR #20678, opened 2022-03-18
chore: migrate mochaEvents integration tests from runner to app

lmiller1990:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1153

### User facing changelog

I migrated the `runner.mochaEvents` and `retries.mochaEvents` that use the `snapshotMocha` function to the new app. The only remaining `mochaEvents` tests are the Cypress Studio ones, which we cannot migrate since we don't support studio yet.

I ran into some complexity here, I'll explain what I ended up doing.

### Additional details

I tried migration our [custom snapshot plugin](https://github.com/cypress-io/cypress/tree/develop/packages/runner/cypress/plugins/snapshot) but had some problems. For some reason `cy.task('getSnapshot')` would not work as expected. I suspect it's running in the **inner** Cypress process, not the outer one. I spend a lot of time debugging this but couldn't quite figure it out.

Instead, I just turned the `__snapshots__` into a regular JS module, and run the assertion in client. I used the [disparity](https://www.npmjs.com/package/disparity) library to generate meaningful diffs:

image

The snapshots themselves are unchanged - no regressions.

Instead of stubbing the inner `Cypress.runner` (mocha), I just captured the events by emitting them under a `cypress:in:cypress` namespace. I'm not sure the stub-mocha strategy is ideal, now we've moved away from the `isolatedCypress` to a real Cypress in Cypress - relying less on stubbing seems like a good thing to do. 

This diagram visualizes what's going on - the inner Cypress emits runner events, which are caught by the outer Cypress. When the final event is emitted (`mocha:end`), we run the assertion and compare the snapshots.

image

I'll do some more investigation around why `cy.task('getSnapshot')` isn't working as expected in the future. 

Tests work as before - I simply copy-pasted the snapshots to `packages/app` and changed them into a TS module.

### PR Tasks


packages/driver/src/cypress.ts
```diff
@@ -397,6 +411,8 @@ class $Cypress {
           return
         }
 
+        this.maybeEmitCypressInCypress('mocha', 'start', args[0])
+
```
lmiller1990, line 415: I don't know why we do `if (this.config('isTextTerminal')) {` before each mocha event is emitted. This indicates we should only emit these events when running in headless run mode, eg on CI, but I don't know why that would be the case.

packages/driver/src/cypress.ts
```diff
@@ -397,6 +411,8 @@ class $Cypress {
...
         if (this.config('isTextTerminal')) {
           return this.emit('mocha', 'start', args[0])
```
lmiller1990, line 417: I don't know why we do `if (this.config('isTextTerminal')) {` before each mocha event is emitted. This indicates we should only emit these events when running in headless run mode, eg on CI, but I don't know why that would be the case.

packages/runner/__snapshots__/runner.mochaEvents.cy.js
```diff
@@ -1,2093 +1,3 @@
-exports['src/cypress/runner tests finish with correct state hook failures fail in [before] #1'] = [
```
lmiller1990, line 1: moved to packages/app

system-tests/projects/runner-e2e-specs/cypress/e2e/runner/can-retry-from-afterEach.retries.mochaEvents.cy.js
```diff
@@ -0,0 +1,37 @@
+let i = 0
```
lmiller1990, line 1: I could not get these to work with the `gen-cypress-test` functions we have, so I just wrote these by hand. @tbiethman might have some ideas what I'm missing. I do find these a bit more readable though.

packages/app/cypress/e2e/runner/retries.mochaEvents.snapshots.ts
```diff
@@ -0,0 +1,7120 @@
+export const snapshots = {
```
lmiller1990, line 1: Copy paste from the snapshots previously existing in `packages/runner/__snapshots__`.

packages/app/cypress/e2e/runner/support/mochaEventsUtils.ts
```diff
@@ -0,0 +1,189 @@
...
+      })
+    }
+
+    return sanitizedLifecycleData
+  },
+}
+
+const keysToEliminate = ['codeFrame', '_testConfig'] as const
+
+function removeUnusedKeysForTestSnapshot<T> (obj: T): T {
```
lmiller1990, line 111: Cleans the snapshot and removed non-deterministic keys (same as we did before).

packages/app/cypress/e2e/runner/support/mochaEventsUtils.ts
```diff
@@ -0,0 +1,189 @@
...
+}
+
+class SnapshotError extends Error {
+  constructor (message: string) {
+    super()
+    this.message = `\n${message}`
+  }
+}
+
+export function runCypressInCypressMochaEventsTest<T> (snapshots: T, snapToCompare: keyof T, done: Mocha.Done) {
```
lmiller1990, line 146: This is the function that maps to the Excalidraw diagram in the PR description. Basically, the "inner" Cypress emits events, and the "outer" Cypress catches them via an `EventEmitter`. Once we encounter the final event, `mocha:end`, we run the assertion.

packages/driver/src/cypress.ts
```diff
@@ -415,6 +431,8 @@ class $Cypress {
         // test:after:run events ourselves
         this.emit('run:end')
 
+        this.maybeEmitCypressInCypress('mocha', 'end', args[0])
```
lmiller1990, line 434: Needed to capture the inner Cypress events in the outer Cypress. Not super ideal, I wonder if there's a cleaner way to do this. I tried spying on the inner Mocha but I could not get this to work.

packages/app/cypress/e2e/runner/support/spec-loader.ts
```diff
@@ -78,3 +78,23 @@ export function loadSpec (options: LoadSpecOptions): void {
...
+  cy.openProject('runner-e2e-specs')
+  cy.startAppServer()
+
+  cy.__incorrectlyVisitAppWithIntercept()
+
+  cy.findByLabelText('Search Specs').type(fileName)
+  // wait for virtualized spec list to update, there is a chance
+  // of disconnection otherwise
+  cy.wait(500)
+  cy.contains('[data-cy=spec-item]', fileName).click()
```
tbiethman, line 93: From what I can tell, #20558 corrected the issue with the runner hanging when directly visiting a spec. So you should be able to do something like:

```js
// we'd need the relative filePath rather than fileName ('runner/file.cy.ts' vs. 'file.cy.ts') to
// build the url correctly
export function runSpec ({ filePath }: { filePath: string }) {
  cy.scaffoldProject('runner-e2e-specs')
  cy.openProject('runner-e2e-specs')
  cy.startAppServer()

  // Need to normalize the spec path for windows usage
  cy.visitApp(getRunnerHref(`cypress/e2e/${filePath}`))

  // Assert runner has loaded before continuing, eventManager isn't initialized otherwise
  // There's probably a better way to assert this?
  cy.contains('.runnable-header', Cypress._.last(filePath.split('/')) as string).should('be.visible')

  return cy.window()
}
```

Doing so would shave quite a bit of time off those test runs. I'm actively working on updating the loadSpec function above.

system-tests/projects/runner-e2e-specs/cypress/e2e/runner/can-retry-from-afterEach.retries.mochaEvents.cy.js
tbiethman: I ported the `generateMochaTestsForWin` helper and its usage as-is to minimize potential changes to the existing tests. It's flexible, but to the detriment of its readability, I agree. 

If it doesn't suit your needs, I think writing these out is preferable to further extending the existing APIs.

packages/app/cypress/e2e/runner/support/mochaEventsUtils.ts
```diff
@@ -0,0 +1,189 @@
...
+  const bus = new EventEmitter()
+  const outerRunner = window.top!.window
+
+  outerRunner.bus = bus
+
+  // TODO: Can we automate writing the snapshots to disk?
+  // For some reason `cy.task('getSnapshot')` has problems when executed in
+  // "cypress in cypress"
```
tbiethman, line 154: Let's log a ticket for this and reference here if we have something concrete to spike on. I don't think any other specs use that task at this point; when the old runner packages get removed, we will lose what little context we currently have around this.

packages/driver/src/cypress.ts
lmiller1990: I'm finding out if we even need this at all. https://github.com/cypress-io/cypress/compare/lmiller1990/try-remove-text-terminal?expand=1

packages/app/cypress/e2e/runner/support/mochaEventsUtils.ts
lmiller1990: Great idea, done: https://cypress-io.atlassian.net/browse/UNIFY-1383. Let me know if you think this needs more details, is missing info, etc.

packages/app/cypress/e2e/runner/support/spec-loader.ts
lmiller1990: What does your `getRunnerHref` look like? I might just wait for your refactor to land and grab it (or if this merges first, make the change after).

tbiethman: We have an [existing implementation](https://github.com/cypress-io/cypress/blob/10.0-release/packages/app/cypress/e2e/specs.cy.ts#L4-L10) that we could generalize, but let's wait on this. I'll make sure this gets updated along with the other implementation once this merges.

tbiethman (approved):
Maintaining this coverage and proving there are no regressions is **awesome** 💯 . Great work with this.

ZachJW34 (approved):
Non-blocking question: Do these snapshots (`{retries,runner}.mochaEvents.snapshots.ts`) have to be manually updated? What is the process for fixing these tests if a change is made that breaks them.

packages/app/cypress/e2e/runner/support/mochaEventsUtils.ts
ryanthemanuel: Do we still need to add the ticket to the comment here?

lmiller1990:
@ZachJW34 yes they do - same as before. It was a little more seamless with the snapshot plugin, but these tests seem to be "write only" - they should not really change, unless mocha changes internals. If we break them accidentally, we wouldn't be updating them, but getting things to pass.

The only time I'd expect these to be changed is when we update Mocha and if they change their internals, (not often).

PR #20689, opened 2022-03-18
fix(driver): readFile - verify null contents to Buffer.from

### LINKED ISSUE #20683: Adding null as an encoding argument to cy.readFile command results in TypeError
### Current behavior

Adding null as an encoding argument to `cy.readFile` command results in `TypeError: First argument must be a string, Buffer, ArrayBuffer, Array, or array-like object`.

It seems that `null` option somehow bypasses the timeout and waiting for file to be downloaded, because as soon as I add 1 sec wait, the command works as expected, and file is read.

### Desired behavior

`cy.readFile` command (with null as encoding argument) should fall back to default timeout of 4 seconds and retry until file is found.

### Test code to reproduce

... trigger a download of a file
```js
cy.readFile(pathToFile, null);
```

### Cypress Version

9.5.1

### Other

_No response_

emilyrohrbough:
- Closes #20683
When `null` encoding is used, and a file cannot be found, we are passing `null` to Buffer.from which is throwing `TypeError: First argument must be a string, Buffer, ArrayBuffer, Array, or array-like object`. This caused the command to bail instead of retrying and providing a meaningful error message to the users.

### User facing changelog
- Fixed an  issue in `cy.readFile()` where the command would not retry or provide a meaningful error message if `null` encoding was used and the file was not found. Fixed [#20683](https://github.com/cypress-io/cypress/issues/20683).

| before | after|
| --- | ---|
| Screen Shot 2022-03-18 at 9 34 39 AM | Screen Shot 2022-03-18 at 9 34 11 AM |
### PR Tasks

### DIFF
--- a/packages/driver/src/cy/commands/files.ts
+++ b/packages/driver/src/cy/commands/files.ts
@@ -72,10 +72,11 @@ export default (Commands, Cypress, cy, state) => {
           }
         }).then(({ filePath, contents }) => {
           // https://github.com/cypress-io/cypress/issues/1558
+          // https://github.com/cypress-io/cypress/issues/20683
           // We invoke Buffer.from() in order to transform this from an ArrayBuffer -
-          // which socket.io uses to transfer the file over the websocket - into a
-          // `Buffer`, which webpack polyfills in the browser.
-          if (options.encoding === null) {
+          // which socket.io uses to transfer the file over the websocket - into a `Buffer`.
+          if (options.encoding === null && contents !== null) {
+          // if (options.encoding === null && contents !== null) {
             contents = Buffer.from(contents)
           }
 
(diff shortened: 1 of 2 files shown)


packages/driver/src/cy/commands/files.ts
```diff
@@ -72,10 +72,11 @@ export default (Commands, Cypress, cy, state) => {
...
         }).then(({ filePath, contents }) => {
           // https://github.com/cypress-io/cypress/issues/1558
+          // https://github.com/cypress-io/cypress/issues/20683
           // We invoke Buffer.from() in order to transform this from an ArrayBuffer -
-          // which socket.io uses to transfer the file over the websocket - into a
-          // `Buffer`, which webpack polyfills in the browser.
-          if (options.encoding === null) {
+          // which socket.io uses to transfer the file over the websocket - into a `Buffer`.
+          if (options.encoding === null && contents !== null) {
+          // if (options.encoding === null && contents !== null) {
```
davidmunechika, line 79:
```suggestion
```

davidmunechika (approved):

packages/driver/src/cy/commands/files.ts
emilyrohrbough: whoops.!

davidmunechika (approved):

BlueWinds (approved):

PR #20690, opened 2022-03-18
fix: Updating link detection in markdown content

tbiethman:
- Closes UNIFY-1063

Some of the links in our markdown content are styled with inner `<code>` elements, which causes our link detection logic to fail. In these instances, the click target is the `<code>` element, not the link itself.

I updated the logic to search for parent links to prevent this. I also added component test coverage for the `useMarkdown` composable and e2e coverage for the links in Settings.

### PR Tasks

### DIFF
--- a/packages/frontend-shared/src/composables/useMarkdown.ts
+++ b/packages/frontend-shared/src/composables/useMarkdown.ts
@@ -102,8 +102,15 @@ export const useMarkdown = (target: Ref<HTMLElement>, text: MaybeRef<string>, op
 
     whenever(target, () => {
       useEventListener(target, 'click', (e: MouseEvent) => {
+        const link = (e.target as HTMLElement).closest('a[href]')
+
+        if (!link) {
+          return
+        }
+
         e.preventDefault()
-        const url = (e.target as HTMLElement).getAttribute('href')
+
+        const url = link.getAttribute('href')
 
         if (url) {
           open(url)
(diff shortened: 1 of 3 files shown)


ZachJW34 (approved):

packages/frontend-shared/src/composables/useMarkdown.ts
```diff
@@ -102,8 +102,15 @@ export const useMarkdown = (target: Ref<HTMLElement>, text: MaybeRef<string>, op
 
     whenever(target, () => {
       useEventListener(target, 'click', (e: MouseEvent) => {
+        const link = (e.target as HTMLElement).closest('a[href]')
```
flotwig, line 105: I was not familiar with `.closest` until reviewing this PR, and wow, that is kind of a misnomer. I assumed it searched siblings too. I wish they had called it something like `.closestParent`.

flotwig (approved):

packages/frontend-shared/src/composables/useMarkdown.ts
tbiethman: I agree with you there. One thing that also isn't immediately clear is that closest() is inclusive of the element that is calling it. So if the target is a link, it'll find itself first. I added tests to cover that case as well.

jaffrepaul: Agree it could be a little wonky but AFAIK it's just the [jQuery  implementation](https://api.jquery.com/closest/). So in theory, has event delegation in mind while being inclusive of the element that's calling it and working up the tree.

PR #20692, opened 2022-03-18
fix: add automation UI

ZachJW34:
- Closes [UNIFY-1341](https://cypress-io.atlassian.net/browse/UNIFY-1341)

### User facing changelog
Adds UI for when the automation gets disconnected or is missing

### Additional details
Adds back in the stubbed components for AutomationDisconnected and AutomationMissing. All of the events were already wired up, we just needed to surface them when the events were fired

Automation:Disconnected

https://user-images.githubusercontent.com/25158820/160470303-9a69d77c-177a-4c20-8423-56aa40b734a5.mov

Automation:Missing

https://user-images.githubusercontent.com/25158820/160470332-7affc34d-f236-4cc2-8ba4-657d0a83c06a.mov

UI based on https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/Cypress-App?node-id=139%3A1221

Had to make a few tweaks to some existing components. Notably, I had to change the `Alert` to allow no overflow. Since the `missing` state has the vertical browser list, with overflow-auto (even though it's an absolutely positioned element) the popup was being clipped by the container and forcing you to scroll to see the list of browsers.

### PR Tasks


packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
```diff
@@ -145,4 +145,67 @@ describe('Cypress In Cypress CT', { viewportWidth: 1500, defaultCommandTimeout:
...
+        }
+
+        originalEmit.call(win.ws, ...args)
+      })
+    })
+
+    cy.get('[data-cy="spec-item"]').first().click()
+    // Let runner stabilize
+    cy.get('[data-cy="reporter-panel"]').should('be.visible').then(() => {
+      connectedCallback()
```
lmiller1990, line 193: Took me a bit to realize that by default, calling this without an argument will set the automation to disconnected. What about making it more explicit?

```js
connectedCallback(false)
```

packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
```diff
@@ -145,4 +145,67 @@ describe('Cypress In Cypress CT', { viewportWidth: 1500, defaultCommandTimeout:
...
+    })
+  })
+
+  it('shows automation missing warning', () => {
+    let connectedCallback: any
+
+    cy.visitApp()
+
+    cy.window().then((win) => {
+      const originalEmit: Function = win.ws?.emit || function () {}
```
lmiller1990, line 178: Neat trick. Since `win.ws.emit` must be defined for this test to pass, should we just throw an error if it's not defined? If we ever introduce a race condition that makes `win.ws` not defined, it'll be a lot easier to track down an explicit error than the classic "cannot call emit of undefined".

packages/app/cypress/e2e/cypress-in-cypress-e2e.cy.ts
```diff
@@ -154,4 +154,73 @@ describe('Cypress In Cypress E2E', { viewportWidth: 1500, defaultCommandTimeout:
     cy.get('[data-model-state="failed"]').should('contain', 'renders the blank page')
     cy.percySnapshot()
   })
+
+  it('shows automation disconnected warning', () => {
```
lmiller1990, line 158: I wonder if we want to have these two very similar tests in the same file? We could even use the same test for both CT and E2E, with something like:

```js
;['component, 'e2e'].forEach(testingType => {
  it(`handles automation disconnects in ${testingType}`, () => {
    cy.scaffoldProject('cypress-in-cypress')
    cy.findBrowsers()
    cy.openProject('cypress-in-cypress')
    cy.startAppServer(testingType)
  })
})
```

I somewhat foresee more "it should work the same in CT and E2E"-like tests in the future, so we might want a pattern for this, to reduce duplication and decrease the chance of a testing type specific bug.

lmiller1990 (commented):
Feature seems fine, just some code quality comments/suggestions.

ZachJW34:
Thanks for the review @lmiller1990. I broke Firefox though (even on 10.0-release, the `is:automation:client:connected` event never receives `isConnected = true`), so looking into what's going on

packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
ZachJW34: Addressed in e1ac0624350e26daac737eace6e2fecde8bdaa85

ZachJW34:
Fixed Firefox. The automation element was using the wrong id (it was `${window.__CYPRESS_CONFIG__.namespace}` when it should have been `${window.__CYPRESS_CONFIG__.namespace}-string` which is what the [packages/extension](https://github.com/cypress-io/cypress/blob/655090330f2a94bd1e041f444d6829fbce368190/packages/extension/app/background.js#L227) expects. Not sure how chrome/electron was receiving `isConnected = true` but now all browsers are rigged up properly.

lmiller1990:
> Fixed Firefox. The automation element was using the wrong id (it was `${window.__CYPRESS_CONFIG__.namespace}` when it should have been `${window.__CYPRESS_CONFIG__.namespace}-string` which is what the [packages/extension](https://github.com/cypress-io/cypress/blob/655090330f2a94bd1e041f444d6829fbce368190/packages/extension/app/background.js#L227) expects. Not sure how chrome/electron was receiving `isConnected = true` but now all browsers are rigged up properly.

If we can move the extension package to TS, we can catch this with the `${string}-string` type. 

I think we should have as many tests as possible running against e2e and CT - only using e2e and CT for cases that are truly unique to those testing types (should not be many).

✅ . Don't forget to approve the new Percy snapshots before merging. Let's get 1 more :eyes: before merging.

packages/app/src/runner/automation/AutInfo.vue
```diff
@@ -0,0 +1,18 @@
...
+    class="flex flex-col p-32px gap-16px items-center"
+    :style="{width: `${windowWidth}px`}"
+  >
+    <img
+      class="h-64px w-64px"
+      src="../../../../frontend-shared/src/assets/logos/cypress-dark.png"
+    >
+    <slot />
+  </div>
+</template>
```
tbiethman, line 12: Seems like the alert overflows this container at smaller sizes; is it worth adding another wrapper element with a min-width/horizontal overflow? 

Screen Shot 2022-03-29 at 6 15 34 PM

tbiethman (approved):
Works great, left one comment regarding some edge-case styling that might be worth tackling now, or perhaps more holistically in the future.

packages/app/src/runner/automation/AutInfo.vue
ZachJW34: Addressed in [14becb3](https://github.com/cypress-io/cypress/pull/20692/commits/14becb328f52a9237aba69a257c61fa3be8a5a6e). I was setting the min-width for the Alerts on the `Automation{Disconnected,Missing}.vue` components so it was easy to add.
Screen Shot 2022-03-30 at 9 30 53 AM

ZachJW34:
Going to merge once green as the last push was a style improvement and I've had two ✅

PR #20717, opened 2022-03-21
fix(unify): command log error code frame should show icon

mapsandapps:
- Closes [UNIFY-1021]

Before | After
--- | ---
Screen Shot 2022-03-23 at 11 08 35 AM | Screen Shot 2022-03-23 at 11 08 02 AM

### PR Tasks

[UNIFY-1021]: https://cypress-io.atlassian.net/browse/UNIFY-1021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/reporter/src/errors/error-code-frame.tsx
```diff
@@ -24,7 +24,7 @@ class ErrorCodeFrame extends Component<Props> {
 
     return (
       <div className='test-err-code-frame'>
-        <FileNameOpener className="runnable-err-file-path" fileDetails={this.props.codeFrame} />
+        <FileNameOpener className="runnable-err-file-path" fileDetails={{ ...this.props.codeFrame, ...{ displayFile: this.props.codeFrame.relativeFile } }} hasIcon />
```
flotwig, line 27: even simpler:

```suggestion
        <FileNameOpener className="runnable-err-file-path" fileDetails={{ ...this.props.codeFrame, displayFile: this.props.codeFrame.relativeFile }} hasIcon />
```

flotwig (dismissed):

ZachJW34 (commented):
What conditions was causing the path to not show as relative? Your picture shows the addition of the icon but the path is the same between the two and I'm not sure what I need to do to reproduce what the picture in the Jira issue shows e.g a file path with `../../path-to-test.cy.js`

mapsandapps:
well @ZachJW34 , you ask a good question... i thought i had fixed this, since i was seeing the bug before my change and not after it, but it turns out that my change didn't fully fix it (actually, i'm not sure it helped at all). i'm currently trying to figure out if there's a way i can consistently reproduce it to track down what's actually wrong.

emilyrohrbough (approved):

ZachJW34 (approved):

PR #20725, opened 2022-03-22
chore: remove un-necessary conditional

lmiller1990:
- Closes N/A

### User facing changelog

I removed some conditionals that don't seem to be necessary in driver.

When working on migration some specs from `runner` to `app`, eg [this PR](https://github.com/cypress-io/cypress/pull/20678), I dived into the driver to intercept the Mocha events. I noticed they all have an `if (config('isTextTerminal')) {` condition, that doesn't seem to do anything, so I removed them. I don't understand why we'd only emit these events in run mode, not open mode. If there's some specific reason, it'd be great to know so we can comment and document it.

**One** of the `if (config('isTextTerminal')) {` conditionals seems necessary. That's this one: https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cypress.ts#L418-L420. 

If I remove this conditional, one test fails - [this test](https://github.com/cypress-io/cypress/blob/develop/system-tests/test/session_spec.ts#L141) in `sessions_spec.ts`. [Failing on CI](https://app.circleci.com/pipelines/github/cypress-io/cypress/34501/workflows/4924d5c8-d306-4a39-b75c-4d1f544f5d70/jobs/1367930/tests#failed-test-0). 

That system test ultimately runs [this test](https://github.com/cypress-io/cypress/blob/develop/system-tests/projects/e2e/cypress/integration/session_persist_spec_1.js#L16) which does:

```js
// this simulates interactive/open mode
// so that the run does not complete until after reload
Cypress.config().isTextTerminal = false
```

I wonder if this is enough to truly simulate run mode. Interestingly enough, I removed that line from the 10.0-release branch and it's fine! I am not sure if this is a regression or the original code didn't make sense. I hope someone who has more context on driver can shed some light on this.

### PR Tasks

### DIFF
--- a/packages/driver/src/cypress.ts
+++ b/packages/driver/src/cypress.ts
@@ -397,11 +397,7 @@ class $Cypress {
           return
         }
 
-        if (this.config('isTextTerminal')) {
-          return this.emit('mocha', 'start', args[0])
-        }
-
-        break
+        return this.emit('mocha', 'start', args[0])
 
       case 'runner:end':
         // mocha runner has finished running the tests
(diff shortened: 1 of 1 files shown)


lmiller1990:
Going to deprioritise this and revisit one day.

PR #20730, opened 2022-03-22
chore: fix data-context clean script, scaffoldProjectNodeModule in open mode

tgriesser:
- Fix "clean" script, noticed this was broken while pairing w/ @emilyrohrbough last week
- Scaffold `package.json` deps if they exist in a system-test directory in open-mode, as we do in regular system tests

### DIFF
--- a/packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
+++ b/packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
@@ -66,6 +66,7 @@ interface FixturesShape {
   scaffold (): void
   scaffoldProject (project: string): void
   scaffoldCommonNodeModules(): Promise<void>
+  scaffoldProjectNodeModules(project: string): Promise<void>
   scaffoldWatch (): void
   remove (): void
   removeProject (name): void
@@ -116,6 +117,8 @@ async function makeE2ETasks () {
     await Fixtures.scaffoldProject(projectName)
 
     await Fixtures.scaffoldCommonNodeModules()
+
+    await Fixtures.scaffoldProjectNodeModules(projectName)
 
     scaffoldedProjects.add(projectName)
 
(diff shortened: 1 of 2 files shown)


packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
```diff
@@ -117,6 +118,8 @@ async function makeE2ETasks () {
 
     await Fixtures.scaffoldCommonNodeModules()
 
+    await Fixtures.scaffoldProjectNodeModules(projectName)
```
emilyrohrbough, line 121: oof def needed but feels like this could add setup times to our tests

emilyrohrbough (approved):

packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
tgriesser: It's a noop if there's no `package.json` in the system test, so it should only add time for the ones where it's warranted. Looking to see how it does in CI though

emilyrohrbough: Yeah, we just re-scaffold these projects for many tests so if a test uses a project that installs node_modules it'd ibe re-install for each tests. Hard to say how many are really using deps though. A before hook to install the projects that are used might save time if this is a huge hit.

tgriesser: I believe there's also already some pretty good optimizations around this that @ZachJW34 did which should make it pretty quick: https://github.com/cypress-io/cypress/blob/f2100a8bbaf946a01b18aa6935e32291712cdbe2/system-tests/lib/dep-installer/yarn.ts#L22-L34

estrada9166 (approved):

PR #20737, opened 2022-03-22
chore: consistent mocha types

elevatebart:
I was annoyed that too often types would fail in mocha & cypress tests in our codebase.

I found that making this consistency push removed some of the flakes.


jennifer-shehane:
@elevatebart Should this change be made against develop and backmerged?

system-tests/package.json
```diff
@@ -30,7 +30,7 @@
     "@packages/ts": "0.0.0-development",
     "@storybook/testing-vue3": "0.0.1",
     "@types/chai": "4.2.15",
-    "@types/mocha": "9.1.0",
+    "@types/mocha": "8.0.3",
```
lmiller1990, line 33: The version of `mocha` in this `package.json` is `7.x`.

I'm not too sure on the best way to do this - should we standardize a single version of Mocha in the entire code base? Maybe 8.x?

lmiller1990 (commented):
I have not seen the flake you mention - can you describe it more?

I'm also a little concerned that some package.json files have different version of mocha and @types/mocha specified - this seems a little weird.

packages/errors/tsconfig.json
```diff
@@ -2,7 +2,6 @@
   "extends": "../ts/tsconfig.json",
   "include": [
     "src",
-    "test",
```
flotwig, line 5: don't the tests need type-checking? if they require a different config, you could add it in `packages/error/test/tsconfig.json`

flotwig (changes requested):
It's not needed for this PR, but eventually I would like to have a `@tooling/test-utils` package. Inside that package, we could maintain:

* codebase-wide versions of mocha, chai, sinon
* scripts that uniformly run/watch tests, so every package has the same command structure for tests
* common fixtures or test utilities that are used everywhere (probably move some of the fixtures out of `packages/server` for this)

@elevatebart could you cherry-pick this to `develop` when it merges please?

packages/errors/tsconfig.json
elevatebart: They do need type checking. 

Tests are ignored in most other packages. Therefore, I removed them from this package as well. 

I wanted this PR to stay small, so I removed it here too.

flotwig: This seems like a regression in test/type coverage if you had to remove type checking from `@packages/errors/test`. I don't follow the reasoning of "I wanted this PR to stay small, so I removed it here too.", I don't think.

lmiller1990:
This seems like a bug can of worms - I agree we need more consistency, should we make this PR a little bigger and either go for 

1) single version of mocha everywhere (suggesting 8.x)
2) proper package to manage this like @flotwig recommended?

Also unsure how valuable this is right now as a "quick win" - seems like there is a few more things involved? I definitely do not want to regress in type checking.

elevatebart:
Thank you all for your input. This needs more work that I initially thought. We need a ticket

PR #20739, opened 2022-03-22
fix: [Multi-domain] Preserve switchToDomain command logs between top domain changes.

### LINKED ISSUE #20621: [Multi-domain]: The command queue disappears when switching top domain.
### What would you like?

The command queue for previous commands is lost when switching top.

### Other

mjhenkes:
- Closes #20621

### User facing changelog

n/a

### Additional details

There were two things wrong that caused the disappearing logs when switching the top domain. 

1. The logs in the spec bridges always started at zero so we were getting duplicate log ids.
2. The logs in the switchToDomain were not getting attached to any commands in the primary domain. This lead to the commands getting lost when state was preserved and then rehydrated during a top domain switch event.

To resolve these problem, this PR passes the current log count into the secondary domain to reset its count on each `run:domain:fn` event. This may result in skipped number sequences per domain, but it will ensure unique id's across all domains, regardless of the situation. (Having a spec bridge that turns into top and back again comes to mind)

Secondly we attach the command logs to the current command by injecting the command log earier in the flow. Where previously we were calling directly to the runner, we now create a full fledge command log and attach it to the current command. For updates, we keep a reference of cross origin logs and update them as updates roll in, subsequently triggering the command updated action in the primary domain as well.

This has the side effect of firing the log:added event in the primary domain when the even was logged in a cross-origin frame. I think this is ok because the log truly is being added and updated in the primary domain, as that is where it is displayed from.

Before:
Screen Shot 2022-03-22 at 3 45 29 PM

After:
Screen Shot 2022-03-22 at 3 44 42 PM

### PR Tasks

### DIFF
--- a/packages/driver/cypress/integration/e2e/multi-domain/commands/multi_domain_misc.spec.ts
+++ b/packages/driver/cypress/integration/e2e/multi-domain/commands/multi_domain_misc.spec.ts
@@ -35,19 +35,6 @@ context('multi-domain misc', { experimentalSessionSupport: true }, () => {
     })
   })
 
-  it('.log()', () => {
-    cy.switchToDomain('http://foobar.com:3500', () => {
-      const afterLogAdded = new Promise<void>((resolve) => {
-        cy.once('log:added', () => {
-          resolve()
-        })
-      })
-
-      cy.log('test log in multi-domain')
-      cy.wrap(afterLogAdded)
-    })
-  })
-
   it('.pause()', () => {
     cy.switchToDomain('http://foobar.com:3500', () => {
       const afterPaused = new Promise<void>((resolve) => {
--- a/packages/driver/cypress/integration/e2e/multi-domain/multi_domain_yield_spec.ts
+++ b/packages/driver/cypress/integration/e2e/multi-domain/multi_domain_yield_spec.ts
@@ -83,8 +83,8 @@ describe('multi-domain yields', { experimentalSessionSupport: true }, () => {
 
   it('throws if subject cannot be serialized and is accessed synchronously', (done) => {
     cy.on('fail', (err) => {
-      assertLogLength(logs, 6)
-      expect(logs[5].get('error')).to.eq(err)
+      assertLogLength(logs, 7)
+      expect(logs[6].get('error')).to.eq(err)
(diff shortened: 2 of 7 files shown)


packages/runner-shared/src/event-manager.js
```diff
@@ -582,12 +582,21 @@ export const eventManager = {
...
     Cypress.multiDomainCommunicator.on('after:screenshot', handleAfterScreenshot)
 
+    const crossOriginLogs = {}
+
     Cypress.multiDomainCommunicator.on('log:added', (attrs) => {
-      reporterBus.emit('reporter:log:add', attrs)
+      // Create a new local log representation of the cross origin log.
+      // It will be attached to the current command.
+      // We also keep a reference to it to update it in the future.
+      crossOriginLogs[attrs.id] = Cypress.log(attrs)
```
mjhenkes, line 591: Store the logs created by cross-origins. I could not find a good way to 'find' a log by id another way as i think they are spread out across many commands.

packages/driver/cypress/integration/e2e/multi-domain/multi_domain_yield_spec.ts
```diff
@@ -83,8 +83,8 @@ describe('multi-domain yields', { experimentalSessionSupport: true }, () => {
 
   it('throws if subject cannot be serialized and is accessed synchronously', (done) => {
     cy.on('fail', (err) => {
-      assertLogLength(logs, 6)
-      expect(logs[5].get('error')).to.eq(err)
+      assertLogLength(logs, 7)
```
mjhenkes, line 86: These tests changed because by attaching the log to the current command the log:added event is now fired for events happening in cross-origins.

packages/driver/cypress/integration/e2e/multi-domain/commands/multi_domain_misc.spec.ts
```diff
@@ -35,19 +35,6 @@ context('multi-domain misc', { experimentalSessionSupport: true }, () => {
     })
   })
 
-  it('.log()', () => {
```
mjhenkes, line 38: moved to it's own file.

chrisbreiding (approved):

PR #20744, opened 2022-03-23
fix(app): show correct git status for newly created/modified files

lmiller1990:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-863

### User facing changelog

- Show correct git status for untracked, staged and modified files
- Align UI with Figma mocks (Should not show git status on folders, just files)
- Fix bug where git status is not appearing on linux (git command not working in `/bin/sh`; default to `$SHELL` or `/bin/bash` for unix-like systems).
- Add proper tests (create an actual git repo; not mocking the response)

image

### Additional details

N/A

### PR Tasks


packages/data-context/test/unit/sources/GitDataSource.spec.ts
```diff
@@ -0,0 +1,59 @@
+import { expect } from 'chai'
+import path from 'path'
+import os from 'os'
+import simpleGit from 'simple-git'
+import fs from 'fs-extra'
+import { createTestDataContext, scaffoldMigrationProject } from '../helper'
+
+describe('GitDataSource', () => {
+  it(`gets correct status for files on ${os.platform()}`, async () => {
```
lmiller1990, line 9: Making sure it runs on windows/linux/macos.

marktnoonan (changes requested):
I've done a bit of testing and seems to work great overall. There are two things to look at. First, when creating a spec from Cypress, it doesn't show the git status in the table, see `filename.cy.ts` here:

Screen Shot 2022-03-24 at 1 44 27 PM

Second, in the mocks, before files are committed, the designs show "created" and "modified" without the name attached. Should we match that here?

Love the new tests!

packages/types/src/git.ts
```diff
@@ -0,0 +1 @@
+export const gitStatusType = ['modified', 'created', 'unmodified'] as const
```
elevatebart, line 1: It's almost like we should have this model of files for everything in the `constants.ts` file.

I like this though.

elevatebart (approved):

lmiller1990:
@marktnoonan I have

- changed UI not to show username for created/modified
- fixed bugs where UI is not updating when you create a new spec from the UI
- other small bugs I found

I put a video demo here to show the new UI and it working as expected: https://cypress-io.atlassian.net/browse/UNIFY-863?focusedCommentId=17183

This was a surprisingly complex ticket. Our virtualized spec list is a bit weird (well, the API with `useSpecList`). We should revisit this - it was the cause of the UI not updating correctly, it was doing some caching, preventing the UI from updating.

marktnoonan:
@lmiller1990 I'm confused by this Percy [snapshot](https://percy.io/cypress-io/cypress/builds/16817567/changed/945431778?browser=chrome&browser_ids=20&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=800&widths=400%2C800%2C1000%2C1024), and [this one](https://percy.io/cypress-io/cypress/builds/16818194/changed/945471300?browser=chrome&browser_ids=20&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=800&widths=400%2C800%2C1000%2C1024).

The first seems to show results where we should be on the empty state, and both show `file.git.null` in the UI. I am OOO the rest of the day and haven't looked any deeper into this. I've dismissed my change request since I won't be able to re-review until Monday.

I'd agree about revisiting the specs list, @ZachJW34 and I have talked about this as well. There are a couple of related problems it would be good to talk about together.

ryanthemanuel:
Should foo.cy.js and blah.cy.js have been checked in?  It looks like they're generated as part of a test

lmiller1990:
@marktnoonan thanks for catching this. I'd forgotten to update the stubs, in [this file](https://github.com/cypress-io/cypress/blob/10.0-release/packages/frontend-shared/cypress/support/mock-graphql/testStubSpecs.ts).

It's fixed now. I hard coded "modified 1 minute ago" since I think it makes it easier to search for. Now Percy should no longer show `git.info.null`. We will need to re-approve the status. I think we should move away from the highly dynamic fixtures we have, it's quite hard to debug and write assertions against, since you need to run the test to see what Faker will generate.

Happy to wait for your re-review if you are out, or get someone else to :eyes:.

@ryanthemanuel thanks, I removed those two junk testing files.

@elevatebart Can you please re-review?

ryanthemanuel (approved):

PR #20758, opened 2022-03-23
feat: skip dependency installation

estrada9166:
- Closes  [UNIFY-1113](https://cypress-io.atlassian.net/browse/UNIFY-1113)

### PR Tasks

### DIFF
--- a/packages/launchpad/src/setup/InstallDependencies.vue
+++ b/packages/launchpad/src/setup/InstallDependencies.vue
@@ -6,6 +6,7 @@
     :next-fn="confirmInstalled"
     class="max-w-640px relative"
     :main-button-variant="canNavigateForward ? 'primary' : 'pending'"
+    :skip-fn="canNavigateForward ? null : confirmInstalled"
   >
     <ManualInstall
       :gql="props.gql"
--- a/packages/launchpad/src/setup/WizardLayout.vue
+++ b/packages/launchpad/src/setup/WizardLayout.vue
@@ -19,6 +19,8 @@
       :back="backLabel"
       :alt="alt"
       :main-variant="mainButtonVariant"
+      :skip="skipLabel"
+      :skip-fn="skipFn"
     >
       <slot name="button-bar" />
     </ButtonBar>
@@ -37,6 +39,7 @@ const props = withDefaults(
   defineProps<{
     next?: string
     back?: string
+    skip?: string
     alt?: string
     showNext?: boolean
     canNavigateForward?: boolean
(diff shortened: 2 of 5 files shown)


packages/launchpad/src/setup/WizardLayout.vue
```diff
@@ -37,29 +39,34 @@ const props = withDefaults(
...
     nextFn: undefined,
     backFn: undefined,
+    skipFn: undefined,
     mainButtonVariant: 'primary',
   },
 )
 
 const nextLabel = computed(() => props.next || t('setupPage.step.next'))
 const backLabel = computed(() => props.back || t('setupPage.step.back'))
+const skipLabel = computed(() => props.back || t('setupPage.step.skip'))
```
tbiethman, line 70: Should be  ```props.skip```

Can we add component tests for the new prop permutations?

packages/launchpad/src/setup/InstallDependencies.vue
```diff
@@ -6,6 +6,7 @@
     :next-fn="confirmInstalled"
     class="max-w-640px relative"
     :main-button-variant="canNavigateForward ? 'primary' : 'pending'"
+    :skip-fn="!canNavigateForward && intervalQueryTrigger.isActive.value ? confirmInstalled : undefined"
```
lmiller1990, line 9: Shouldn't we just let them skip straight away? Why do we wait 18s (that's really long). Or am I reading this logic incorrectly?

estrada9166: It shows the skip immediately - but is gonna be hidden after `18s`

image

image

tbiethman: Looks like it's 180 seconds... but given that we're waiting for them to install dependencies, more time there is better. 

If we're giving them the ability to skip now, I'm not sure why we need to stop the interval at all until they take some sort of action (they installed the dependencies, they go back, or they skip). I guess if someone were to park at this view, it's nice that we stop eventually as a sanity check?

But if that is the case, it also feels weird to have the "Waiting..." button look active when it's technically not doing anything until Refresh is selected. It seems like we could repurpose the "Waiting" button, enable it when paused and present a different message, and thus keep "Skip" always accessible. This is definitely out of scope for this PR, but might be worth some thought at some point.

tbiethman (dismissed):

packages/launchpad/src/setup/InstallDependencies.vue
lmiller1990: This feature feels wrong - a skip button that goes away after 18 seconds? What kind of button randomly changes after some interval?

I don't see "refresh" in the designs. Where is this requirement specified?

Code seems fine but unless I missed something the feature seems incorrect.

lmiller1990 (commented):
Code seems fine but feature seem wrong - comment: https://github.com/cypress-io/cypress/pull/20758#discussion_r834819309

packages/launchpad/src/setup/InstallDependencies.vue
```diff
@@ -6,20 +6,12 @@
     :next-fn="confirmInstalled"
     class="max-w-640px relative"
     :main-button-variant="canNavigateForward ? 'primary' : 'pending'"
+    :skip-fn="!canNavigateForward ? confirmInstalled : undefined"
```
lmiller1990, line 9: Should we always show this? Is there ever a case where this will not be shown?

lmiller1990 (commented):
Great - one last comment/question.

jennifer-shehane:
@estrada9166 Can you add the details of the issue to the PR description outside of Jira.

flotwig:
@estrada9166 can you please fill out the PR description? Without a clear description of the changes, `git blame` for this change will not be very useful.

packages/launchpad/src/setup/InstallDependencies.vue
estrada9166: Currently it's not being shown when the user can continue - otherwise it always shows the skip, bc, can be confusing clicking skip and actually doing what continue does

lmiller1990 (approved):
Looks good, thanks for all the back-and-forth on this one.

tbiethman (approved):

PR #20760, opened 2022-03-23
fix: cache last-used browser by project, name, and channel

flotwig:
- Closes UNIFY-565

### User facing changelog

n/a

### Additional details

* Caches the last-used browser per-project and pre-selects it on next startup.
* Previously, if a `cliBrowser` was passed that did not exist, there would be no activebrowser. It now falls back to the regular selection algorithm (last-used, and if none is found, fall back to first found browser)
* Deleted some unused code.
* Also renamed currentBrowser, chosenBrowser to "activeBrowser" everywhere.

Previously, the first browser detected would be preselected in launchpad.

Now, the last used browser is stored per-project by name+channel and auto-selected in launchpad.

### PR Tasks


elevatebart:
I know this is draft but I could not resist :D

@flotwig headsup @tbiethman is working on the same features. 

I would avoid renaming `currentBrowser` into `activeBrowser` without a very good reason... ^^

flotwig:
@elevatebart his PR shouldn't conflict too much. I'm going to rename it for consistency's sake, we shouldn't have 2 names for the same ole thing.

packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -208,28 +208,20 @@ export class ProjectActions {
...
     let activeSpec: FoundSpec | undefined
 
     if (specPath) {
       activeSpec = this.ctx.project.getCurrentSpecByAbsolute(specPath)
     }
 
-    // Ensure that we have loaded browsers to choose from
-    if (this.ctx.appData.refreshingBrowsers) {
-      await this.ctx.appData.refreshingBrowsers
-    }
```
flotwig, line 225: this was never being set

packages/data-context/src/actions/AppActions.ts
```diff
@@ -17,19 +17,24 @@ export interface ApplicationDataApiShape {
 export class AppActions {
   constructor (private ctx: DataContext) {}
 
-  setActiveBrowser (browser: FoundBrowser) {
-    this.ctx.coreData.chosenBrowser = browser
+  async setActiveBrowser (browser: FoundBrowser) {
+    this.ctx.coreData.activeBrowser = browser
+
+    await this.ctx._apis.projectApi.insertProjectPreferencesToCache(this.ctx.lifecycleManager.projectTitle, {
```
elevatebart, line 23: Reaching so deep in the context object is suspicious.

It could mean that this function is better fitted for the `ProjectActions.ts`

packages/frontend-shared/cypress/e2e/support/e2eSupport.ts
```diff
@@ -266,6 +266,9 @@ function startAppServer (mode: 'component' | 'e2e' = 'e2e') {
 
         await isInitialized.promise
 
+        if (!ctx.lifecycleManager.browsers?.[0]) throw new Error('No browsers available in startAppServer')
```
elevatebart, line 269: This is a more usual way of doing this check. 

```suggestion
        if (!ctx.lifecycleManager.browsers?.length) throw new Error('No browsers available in startAppServer')
```

packages/graphql/src/schemaTypes/objectTypes/gql-ProjectPreferences.ts
```diff
@@ -7,9 +7,5 @@ export const ProjectPreferences = objectType({
     t.string('testingType', {
       description: 'The preferred testing type to start in',
     })
-
-    t.string('browserPath', {
-      description: 'The preferred browser to launch',
-    })
```
elevatebart, line 13: Would be cool to have a ticket to add this back somewhere. The number of times I forgot to click to start a browser is non-negligeable.

elevatebart (commented):
Some non-blocking comments

packages/graphql/src/schemaTypes/objectTypes/gql-ProjectPreferences.ts
flotwig: There is UNIFY-1059 and UNIFY-1243 though I think UNIFY-1243 may be a duplicate. Does that/do those cover this ask?

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -272,6 +270,43 @@ export class ProjectLifecycleManager {
...
+      if (this.ctx.coreData.activeBrowser) return
+    }
+
+    // lastBrowser is cached per-project.
+    const prefs = await this.ctx.project.getProjectPreferences(path.basename(this.projectRoot))
+    const browsers = await this.ctx.browser.machineBrowsers()
+
+    if (!browsers[0]) throw new Error('No browsers available in setInitialActiveBrowser, cannot set initial active browser')
+
+    this.ctx.coreData.activeBrowser = (prefs?.lastBrowser && browsers.find((b) => {
```
ZachJW34, line 293: Small, but all mutations of `coreData` should be done via `ctx.update`

ZachJW34 (approved):
Small non-blocking comment, works great!

packages/data-context/src/data/ProjectLifecycleManager.ts
flotwig: Noted, I was adapting the exist code. What does `ctx.update` do? The comment isn't super clear: https://github.com/cypress-io/cypress/blob/25e44ab8e6f531fe847949023f2bd9ff83ca592f/packages/data-context/src/DataContext.ts#L276-L281

Maybe we can add a lint rule to enforce this - I see lots of places where this rule is violated.

packages/graphql/src/schemaTypes/objectTypes/gql-ProjectPreferences.ts
elevatebart: UNIFY-1536

packages/data-context/src/data/ProjectLifecycleManager.ts
ZachJW34: I guess it doesn't do much right now. I believe Tim had an idea of using Immer for time travel debugging, or something along those lines. I'll check to see if an issue exists and make one if it does not. Probably shouldn't enforce a lint rule until we hook up something that provides real value.

PR #20764, opened 2022-03-23
chore: Removing usage of __incorrectlyVisitAppWithIntercept

tbiethman:
- Part of UNIFY-1151

The __incorrectlyVisitAppWithIntercept utility was being used throughout our app e2e tests to allow us to intercept GraphQL operations over `fetch`. We shouldn't need to do that, as we can stub out any necessary actions using withCtx and get a much more representative test.

The existing usages were removed and replaced with the appropriate stubbing. Doing so corrected a few specs where GraphQL errors were present (though they had no impact to the spec's assertions).  I also removed the __incorrectlyVisitAppWithIntercept API outright, as we shouldn't need to use it, nor do we want to use it going forward.

I left the underlying functionality in place so we can still toggle this behavior from the GUI for debugging purposes. But we can remove that, too, if it makes sense to do so.

### PR Tasks


packages/app/cypress/e2e/specs.cy.ts
```diff
@@ -416,19 +424,11 @@ describe('App: Index', () => {
...
         .as('enterSpecInput')
 
-        cy.get('@enterSpecInput').clear().type(getPathForPlatform('src/e2e/MyTest.spec.j'))
-        cy.intercept('mutation-EmptyGenerator_MatchSpecFile', (req) => {
-          if (req.body.variables.specFile === getPathForPlatform('src/e2e/MyTest.spec.jx')) {
-            req.on('before:response', (res) => {
-              res.body.data.matchesSpecPattern = true
-            })
-          }
-        })
```
tbiethman, line 426: This intercept was really indirect. I updated the spec to just input a spec that matches the pattern so that the expected warning is presented.

packages/app/cypress/e2e/specs.cy.ts
```diff
@@ -1012,10 +1020,8 @@ describe('App: Index', () => {
         cy.percySnapshot('Component Generator')
         checkCodeGenCandidates(['cypress.config.js', 'App.cy.jsx', 'App.jsx', 'index.jsx', 'support.js', 'Button.jsx', 'Button.stories.jsx'])
 
-        cy.intercept('query-ComponentGeneratorStepOne').as('code-gen-candidates')
         cy.findByTestId('file-match-button').click()
         cy.findByPlaceholderText(defaultMessages.components.fileSearch.byExtensionInput).clear().type('App.*')
-        cy.wait('@code-gen-candidates')
```
tbiethman, line 1018: This interception wasn't adding much; we validate the query results below.

packages/app/cypress/e2e/top-nav.cy.ts
```diff
@@ -341,12 +353,8 @@ describe('App Top Nav Workflows', () => {
           })
         })
 
-        cy.intercept('mutation-Auth_Logout').as('logout')
-
         cy.findByRole('button', { name: 'Log Out' }).click()
 
-        cy.wait('@logout')
```
tbiethman, line 348: This interception was unnecessary, as we validate the effects of it below.

packages/launchpad/cypress/e2e/open-mode.cy.ts
```diff
@@ -113,7 +114,6 @@ describe('Launchpad: Open Mode', () => {
 
     it('opens using finder', () => {
       cy.withCtx(async (ctx, o) => {
-        ctx.actions.electron.showItemInFolder = o.sinon.stub()
```
tbiethman, line 116: This is stubbed by default by the helper

packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
```diff
@@ -191,7 +191,7 @@ async function makeE2ETasks () {
                   query,
                   result,
                   callCount: operationCount[operationName ?? 'unknown'],
-                })
+                }, testState)
```
tbiethman, line 194: I added testState to the remoteGraphQLIntercept callback so that I could assert that [an operation occurred](https://github.com/cypress-io/cypress/pull/20764/files#diff-18d3d39b03b089d6c1a5c08ea987b6ae379d4dc05307619b14a18fc7f02ef8d3R237). This might be controversial? Suggestions welcome for an alternative approach.

packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
```diff
@@ -118,20 +119,30 @@ describe('Cypress In Cypress CT', { viewportWidth: 1500, defaultCommandTimeout:
...
 
     cy.get('[data-cy="select-browser"]')
     .click()
 
     cy.contains('Firefox')
     .click()
 
-    cy.wait('@setBrowser').then(({ request }) => {
-      expect(request.body.variables.specPath).to.contain('/cypress-in-cypress/src/TestComponent.spec.jsx')
+    cy.withCtx((ctx, o) => {
```
tgriesser, line 137: One general comment - if you're seeing any flake where there's race conditions between the action being fired in response to an event, and the `withCtx` body evaluating on the server, there's a `withRetryableCtx` which can be used in situations where the `withCtx` body is idempotent and just needs to pass with the assertions once.

packages/frontend-shared/cypress/e2e/support/e2eSupport.ts
```diff
@@ -513,7 +489,6 @@ Cypress.Commands.add('scaffoldProject', scaffoldProject)
 Cypress.Commands.add('addProject', addProject)
 Cypress.Commands.add('openGlobalMode', openGlobalMode)
 Cypress.Commands.add('visitApp', visitApp)
-Cypress.Commands.add('__incorrectlyVisitAppWithIntercept', __incorrectlyVisitAppWithIntercept)
```
tgriesser, line 516: 🎉

tgriesser (approved):
Nice work! Just added a few comments, otherwise this looks great.

packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
BlueWinds: This seems similar to `operationCount`, which was merged in yesterday - https://github.com/cypress-io/cypress/pull/20598/files#r831748428. Though operationCount is desgined for a different use, it seems like it could be used here.

I'd have been inclined towards your implementation, but that one landed first. We should probably settle on just one of them :woman_shrugging:

BlueWinds:
> I left the underlying functionality in place so we can still toggle this behavior from the GUI for debugging purposes. But we can remove that, too, if it makes sense to do so.

I'd be inclined to remove it - this seems like the sort of confusing functionality that we'll forget about, and then in 18 months have to go back digging for why it's there. If we don't have an active use for it, let's axe the code.

packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
tbiethman: I had a similar thought when I initially merged the conflicts, but I settled on leaving both as they don't conflict with each negatively. The operation count could be reimplemented using testState on a per-test basis, but it seems fair to track that internally and offer it first-class for the tests that need it. The operationCount doesn't give me the ability to assert that an operation was called, which was what I was wanting for these tests.

BlueWinds: operationCount is used precisely once (the second usage is in error, and I'm removing it in an upcoming PR) - I think implementing it per-test isn't a terrible thing.  But with that said, looks good and it doesn't bother me enough to block merging and getting on with our days. :)

PR #20770, opened 2022-03-24
fix: allow migration of pluginsFile using `env` properties

elevatebart:
- Closes UNIFY-1392

### User facing changelog

Allow migrating a project that uses `config.env` by adding a default env property to the config object.

### Details

In some pre-10.X plugins, the config object is modified to add variables to `config.env`. In the migration process, we run the `pluginsFile` function. During this execution, plugins expect the config object contain certain values like "env" since they are provided as default values in 9.X.

EDIT 1: I do not think it is a good idea to copy all the default values from 9 into the migration process. This would mean a lot of code to maintain. I preferred a more pragmatic approach: Check that most common plugins work and fix only the values that are actually used in those. 

EDIT 2: Since I had to diff the output anyway, I finally opted for copying the full defaults from 9.X. I don't expect them to change much and it will be helpful for users. Migration now mirrors as well as possible the way the config is resolved in 9.X.


lmiller1990 (dismissed):

BlueWinds:
@elevatebart - Can you add details about what this is fixing and why to the PR description?

system-tests/projects/migration-e2e-plugins-modify-config/cypress/plugins/index.js
```diff
@@ -1,6 +1,8 @@
 module.exports = (on, config) => {
   config.integrationFolder = 'tests/e2e'
   config.testFiles = '**/*.spec.js'
+  // In the plugin @cypress/code-coverage this pattern is used
+  config.env.CYPRESS_RECORD_KEY_DRAGGING = 'true'
```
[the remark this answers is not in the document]
elevatebart: It does not.
I will indeed add a more detailed description of the changes.

BlueWinds (approved):

lmiller1990:
@elevatebart looks good - can you get CI green so we can merge? Anything left?

elevatebart:
@lmiller1990 The PR seems ready for review.

I will wait until CI is greeen to send you any slack message and re-request your review.

lmiller1990 (approved):

lmiller1990:
✅

packages/data-context/src/sources/migration/legacyOptions.ts
```diff
@@ -0,0 +1,305 @@
...
+}
+
+// NOTE:
+// If you add/remove/change a config value, make sure to update the following
+// - cli/types/index.d.ts (including allowed config options on TestOptions)
+// - cypress.schema.json
+//
+// Add options in alphabetical order for better readability
+
+// TODO - add boolean attribute to indicate read-only / static vs mutable options
```
ryanthemanuel, line 29: Do we have a followup ticket for this?

ryanthemanuel (approved):

lmiller1990:
Looks good @elevatebart, once we have an answer on the question above let's merge!!

packages/data-context/src/sources/migration/legacyOptions.ts
elevatebart: no we don't. this file is a pure copy paste of the 9.X `config/lib/options.ts`. I should clean it up and remove all the not so useful fluff from it.

I could also add where it comes from at the top of the file with a warning to use it only to migrate 9.X projects.

PR #20772, opened 2022-03-24
chore: Sort orgs and projects when connecting new project

BlueWinds:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1309

### User facing changelog
This matches the existing 9.x behavior, where projects and organizations are sorted by name when connecting a new project. It also matches existing behavior by automatically selecting a project with the same name if one is available.

### Additional details
Adds additional assertions and tests around these components and features.

Organizations are now sorted by name (matching 9.x behavior. Previously in 10.x this was unsorted):
image

Projects are now sorted by name (matching 9.x behavior. Previously in 10.x this was unsorted)
image

If an existing project under an organization matches the current project name, it is automatically selected (matching 9.x behavior. Previously in 10.x this would always show "Pick project"):
image

### PR Tasks


packages/app/src/runs/modals/CreateCloudOrgModal.vue
```diff
@@ -13,10 +13,9 @@
         <ExternalLink
           class="border rounded mx-auto border-gray-100 py-4px px-16px text-indigo-500 inline-block"
           :href="createOrgUrl"
-          :prefix-icon="OrganizationIcon"
           :include-graphql-port="true"
-          prefix-icon-class="icon-light-transparent icon-dark-white"
         >
+          <i-cy-office-building_x16 class="inline-block icon-dark-gray-500" />
```
tbiethman, line 18: It looks like there's some overall differences in style between what's here and what's [in Figma](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/Cypress-App?node-id=4597%3A31174). Seeing the icon is an improvement, but it looks like we'd need to transition to a button to match the design.

tbiethman (approved):
Sorting/selection behavior works great, tests look great. Might be worth logging a follow-up for the create org button styles rather than update anything here?

packages/app/src/runs/modals/CreateCloudOrgModal.vue
BlueWinds: I was trying to look at the figma yesterday, but didn't have access until just now. I've updated the button style to match the design.

BlueWinds: I think a link is more accessible / semantic. it only takes a few styles to make it look like a button anyway.

tbiethman (approved):

estrada9166 (approved):

PR #20775, opened 2022-03-25
fix: refactor config lifecycle and fix issues that occur while loading the config

ryanthemanuel:
- Closes [UNIFY-1235](https://cypress-io.atlassian.net/browse/UNIFY-1235)

### User facing changelog

* Don't watch the config and config-required files until the config successfully loads. Users will have to click 'Try Again' in order to refresh the config if they experience on error on config loading
* If there's an error loading the legacy config during the migration process, we will properly kick off the migration process again when the user clicks 'Try Again'
* When a file involved with the configuration changes, we will now *always* reload the config rather than trying to only kick off specific aspects of the config loading process

### Additional details

The goal of this PR is to simplify and clean up the configuration loading lifecycle. The main design decisions made here are:

* Move all IPC logic to be self contained inside of ProjectConfigIPC
* Move config loading lifecycle logic to ProjectConfigManager
* Extract out self contained modules where possible (e.g. env file loading and node event management)

In addition, we made functional improvements/fixes. The user facing changes are mentioned above. In addition, the following changes have been made:

* The config file itself is not watched twice
* We will now always reload the full config on a file changing, rather than trying to be smarter about it and only setting up node events, etc. in some cases
* setCurrentProjectAndTestingTypeForTestSetup will no longer trigger the server to launch (fixing some flake that would show up in some of our tests)

### PR Tasks


packages/server/lib/makeDataContext.ts
```diff
@@ -62,9 +62,6 @@ export function makeDataContext (options: MakeDataContextOptions): DataContext {
       validateConfig: configUtils.validate,
       updateWithPluginValues: config.updateWithPluginValues,
       setupFullConfigWithDefaults: config.setupFullConfigWithDefaults,
-      validateRootConfigBreakingChanges: configUtils.validateNoBreakingConfigRoot,
```
emilyrohrbough, line 65: Can we remove everything passed in from `@package/config` and directly consume it in the DataContext package?

Also, is there any need to keep the config validation logic in the server or can it all be moved to the config package and consume it from the Data Context page so all config logic is in one place?

packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -0,0 +1,790 @@
...
+  private _watchers = new Set<chokidar.FSWatcher>()
+  private _registeredEvents: Record<string, Function> = {}
+  private _registeredEventsTarget: TestingType | undefined
+  private _initializedProject: unknown | undefined
+  private _testingType: TestingType | undefined
+  private _state: ConfigManagerState = 'pending'
+  private _loadConfigPromise: Promise<LoadConfigReply> | undefined
+  private _cachedLoadConfig: LoadConfigReply | undefined
+
+  #cypressEnv: CypressEnv
```
emilyrohrbough, line 67: Just curious, why prefix these names with `#`?

packages/data-context/src/data/index.ts
```diff
@@ -1,7 +1,9 @@
 /* eslint-disable padding-line-between-statements */
 // created by autobarrel, do not modify directly
 
+export * from './CypressEnv'
```
emilyrohrbough, line 4: Just curious, do we need to expose these?

emilyrohrbough:
This PR makes me happy! Love the clean up 😄

packages/server/lib/makeDataContext.ts
```diff
@@ -62,9 +62,6 @@ export function makeDataContext (options: MakeDataContextOptions): DataContext {
       validateConfig: configUtils.validate,
```
emilyrohrbough, line 62: I will say, we are passing in `cypressVersion` above, but I'm not sure we need to. We are pulling `pkg.version` directly in the DataContext package in several places.

tgriesser (commented):
Nice - I think this is starting to take a good shape

packages/server/lib/makeDataContext.ts
tgriesser: Yeah, we can probably just do `require('@packages/root')` here

packages/data-context/src/data/ProjectConfigManager.ts
tgriesser: This is the new ES syntax for private class member: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

For some reason I thought TS compiled it to just use a symbol, but i guess it uses a WeakSet to keep track of the class instance 🤔 https://www.typescriptlang.org/play#code/MYGwhgzhAEBiD29oG8BQ1oGIC2BTbARrgE4AUAlCuhtMbgC4CuxAdtAIzUC+qqXQA

packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -0,0 +1,790 @@
...
+    this._registeredEventsTarget = undefined
+    this.closeWatchers()
+  }
+}
+
+function pDeferFulfilled<T> (): pDefer.DeferredPromise<T> & {settled: boolean} {
+  const dfd = pDefer<T>()
+  let settled = false
+  const { promise, resolve, reject } = dfd
+
```
tgriesser, line 769: From when I was looking into this previously, I think once we rework things a bit more, we should be able to get rid of the need for this.

packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -0,0 +1,790 @@
...
+    const finalConfig = this._cachedFullConfig = this.options.updateWithPluginValues(fullConfig, result.setupConfig ?? {})
+
+    // This happens automatically with openProjectCreate in run mode
+    if (!this.options.isRunMode) {
+      if (!this._initializedProject) {
+        this._initializedProject = await this.options.initializeActiveProject({})
+      } else {
+        // TODO: modify the _initializedProject
+      }
+    }
```
tgriesser, line 270: I think this will probably live outside of the config manager, assuming that we're able to handle changes to the config without destroying the `initializedProject`, since that's what holds the server reference.

packages/data-context/src/data/index.ts
ryanthemanuel: These get generated automatically

packages/data-context/src/data/ProjectConfigIpc.ts
```diff
@@ -52,11 +70,11 @@ export class ProjectConfigIpc extends EventEmitter {
   send(event: 'setupTestingType', testingType: TestingType, options: Cypress.PluginConfigOptions): boolean
   send(event: 'loadConfig'): boolean
   send (event: string, ...args: any[]) {
-    if (this.childProcess.killed) {
+    if (this._childProcess.killed || !this._childProcess.connected) {
```
tgriesser, line 73: Nice, maybe this handles the `ERR_IPC_CHANNEL_CLOSED` issue we were seeing?

packages/launchpad/cypress/e2e/migration.cy.ts
```diff
@@ -1,11 +1,22 @@
...
 const renameAutoStep = `[data-cy="migration-step renameAuto"]`
 const renameManualStep = `[data-cy="migration-step renameManual"]`
 const renameSupportStep = `[data-cy="migration-step renameSupport"]`
 const configFileStep = `[data-cy="migration-step configFile"]`
 const setupComponentStep = `[data-cy="migration-step setupComponent"]`
 
+function getPathForPlatform (posixPath: string) {
+  // @ts-ignore
+  const cy = window.Cypress
+  const platform = cy?.platform || JSON.parse(decodeBase64Unicode(window.__CYPRESS_CONFIG__.base64Config)).platform
```
tgriesser, line 13: Minor, but is this something we can normalize in `writeFileInProject`?

packages/server/test/unit/config_spec.js
```diff
@@ -134,11 +136,11 @@ describe('lib/config', () => {
     context('validation', () => {
       beforeEach(function () {
         this.expectValidationPasses = () => {
-          return config.get(this.projectRoot) // shouldn't throw
+          return this.ctx.lifecycleManager.getFullInitialConfig() // shouldn't throw
```
tgriesser, line 139: Does this remove all uses of `config.get` / can we remove it?

tgriesser (approved):
Awesome, left a few comments but given that the tests are green, seems like a better foundation than what we have now to iterate on

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -274,59 +332,60 @@ export class ProjectLifecycleManager {
...
+
+    const packageManagerUsed = this.getPackageManagerUsed(projectRoot)
+
     this.ctx.update((s) => {
       s.currentProject = projectRoot
       s.packageManager = packageManagerUsed
     })
 
+    if (this.readyToInitialize(this._projectRoot)) {
+      this._configManager.initializeConfig().catch(this.onLoadError)
```
elevatebart, line 351: Why is this statement not awaited?
It could be because `setCurrentProject()` needs to be called in the constructor (and the contructor cannot be async).

If yes, I would suggest we split it off in a second function called after creation.

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -274,59 +332,60 @@ export class ProjectLifecycleManager {
...
+    }
+  }
+
+  readyToInitialize (projectRoot: string): boolean {
```
elevatebart, line 355: Would you please add a comment on this function to say what it checks?

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -274,59 +332,60 @@ export class ProjectLifecycleManager {
...
-    // NOTE: If we watch the files while initializing,
-    // the config will be loaded before the migration is complete.
-    // The migration screen will disappear see `Main.vue` & `MigrationAction.ts`
-    if (!needsCypressJsonMigration) {
-      this.initializeConfigWatchers()
+      this.ctx.emitter.toLaunchpad()
+    } catch (error) {
+      this.onLoadError(error)
     }
   }
```
elevatebart, line 391: Why not have used a setter function for `setRunModeExitEarly()` ?
`set runModeExitEarly()` would have been fine, right?

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -1368,19 +636,19 @@ export class ProjectLifecycleManager {
...
+      if (!this.metaState.hasValidConfigFile) {
+        return this.ctx.onError(getError('NO_DEFAULT_CONFIG_FILE_FOUND', this.projectRoot))
+      }
 
-    if (!this.metaState.hasValidConfigFile) {
-      return this.ctx.onError(getError('NO_DEFAULT_CONFIG_FILE_FOUND', this.projectRoot))
+      if (testingType) {
+        this.setCurrentTestingType(testingType)
+      } else {
+        this.setCurrentTestingType('e2e')
```
elevatebart, line 650: Should we make `setCurrentTestingType()` async now and await it where it's used?

elevatebart (approved):
Amazing work really good step forward. I have a few non-blocking remarks.

I could quickly do the work I suggest if you want.

packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -0,0 +1,490 @@
...
+
+  constructor (private options: ProjectConfigManagerOptions) {
+    this._cypressEnv = new CypressEnv({
+      envFilePath: this.envFilePath,
+      validateConfigFile: (filePath, config) => {
+        this.validateConfigFile(filePath, config)
+      },
+    })
+
+    return autoBindDebug(this)
```
elevatebart, line 54: Also @tgriesser, what does this function do exactly?

If feels like it is helping debug getters, setters and method calls of the current object.
Why don't we add it on every class? what could be the downside?

elevatebart: A short comments would be invaluable.

packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -0,0 +1,490 @@
...
+  loadTestingType () {
+    // If we have set a testingType, and it's not the "target" of the
+    // registeredEvents (switching testing mode), we need to get a fresh
+    // config IPC & re-execute the setupTestingType
+    if (this._registeredEventsTarget && this._testingType !== this._registeredEventsTarget) {
+      this.reloadConfig().catch(this.onLoadError)
+    } else if (this._eventsIpc && !this._registeredEventsTarget && this._cachedLoadConfig) {
+      this.setupNodeEvents(this._cachedLoadConfig).catch(this.onLoadError)
+    }
+  }
```
elevatebart, line 143: This function is async and would be useful to await.

When this is called, it usually is a good idea to know when it's ended.

```suggestion
  async loadTestingType () {
    // If we have set a testingType, and it's not the "target" of the
    // registeredEvents (switching testing mode), we need to get a fresh
    // config IPC & re-execute the setupTestingType
    if (this._registeredEventsTarget && this._testingType !== this._registeredEventsTarget) {
      await this.reloadConfig().catch(this.onLoadError)
    } else if (this._eventsIpc && !this._registeredEventsTarget && this._cachedLoadConfig) {
      await this.setupNodeEvents(this._cachedLoadConfig).catch(this.onLoadError)
    }
  }
```

packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -0,0 +1,490 @@
...
+
+    return result
+  }
+
+  async reloadConfig () {
+    this._loadConfigPromise = undefined
+    this._registeredEventsTarget = undefined
+
+    await this.initializeConfig()
+    this.loadTestingType()
```
elevatebart, line 238: This should be awaited since it has only promise returns
```suggestion
    await this.loadTestingType()
```

packages/data-context/src/data/ProjectLifecycleManager.ts
ryanthemanuel: +1. This was existing code that I copied, but I can update it.

packages/data-context/src/data/ProjectConfigManager.ts
ryanthemanuel: Adding here: https://github.com/cypress-io/cypress/pull/20775/commits/1e12aa86126f153a44c0660f36e4bcdc9c6d0867

packages/data-context/src/data/ProjectLifecycleManager.ts
ryanthemanuel: Added here: [`6bace98` (#20775)](https://github.com/cypress-io/cypress/pull/20775/commits/6bace98a24172d9dcaa72b133eee180e7edba875)

packages/data-context/src/data/ProjectLifecycleManager.ts
ryanthemanuel: Added here: [`6bace98` (#20775)](https://github.com/cypress-io/cypress/pull/20775/commits/6bace98a24172d9dcaa72b133eee180e7edba875)

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -361,6 +361,13 @@ export class ProjectLifecycleManager {
     }
   }
 
+  /**
+   * Handles pre-initialization checks. These will display warnings or throw with errors if catastrophic.
+   * Returns yes, if we're not ready to initialize due to needing to migrate
```
elevatebart, line 366: Returns false?

```suggestion
   * Returns false, if we're not ready to initialize due to needing to migrate
```

elevatebart (dismissed):

elevatebart (approved):

packages/server/lib/makeDataContext.ts
ryanthemanuel: Logged a future issue here: https://cypress-io.atlassian.net/browse/UNIFY-1431

tbiethman (approved):

PR #20786, opened 2022-03-25
refactor: allow ts-node to be scoped

elevatebart:
Advances UNIFY-1216

### TLDR

- This PR allows, in development, to register ts-node on a reduced scope.

### Details

- We want the child process for running `cypress.config.js` to be written in TypeScript.
- Some of the TypeScript code run in this process might not be in Cypress since it's the users.
- We need for ts-node to register once for Cypress's code and once for User's code so that if the 2 configs are different they do not conflict
- This PR does not change the way everything is run in production. We only use ts-node there for the User's code.
- This PR allows, in development, to register ts-node on a reduced scope.
- Then users can use their own `tsconfig.json` for their own files


packages/ts/registerDir.js
```diff
@@ -0,0 +1,40 @@
...
+const debug = require('debug')('cypress:ts')
+const path = require('path')
+
+// in development we should have TypeScript hook installed
+// in production or staging we are likely to be running
+// built Electron app without ts-node hook. Assume the
+// build has been done correctly
+module.exports = function (scopeDir) {
+  try {
+  // Prevent double-compiling if we're testing the app and already have ts-node hook installed
```
ryanthemanuel, line 11: Comment spacing off here

elevatebart: What do you mean?

packages/ts/registerDir.js
```diff
@@ -0,0 +1,40 @@
...
+        scopeDir,
+      })
+    } else {
+      debug('skipping ts-node registration while testing the app')
+    }
+
+  // do we need to prevent any other TypeScript hooks?
+  } catch (e) {
+  // continue running without TypeScript require hook
+    debug('Running without ts-node hook in environment "%s"', process.env.CYPRESS_INTERNAL_ENV)
```
ryanthemanuel, line 38: Is it worth debug logging the error here?

elevatebart: If you are debugging ts, it might be helpful to know if tsconfig is not compatible with ts-node, for instance.

packages/ts/register.js
```diff
@@ -1,34 +1 @@
...
-      preferTsExts: true, // Helps when the files are compiled locally, resolves the TS file
-    })
-  }
-
-  // do we need to prevent any other TypeScript hooks?
-} catch (e) {
-  // continue running without TypeScript require hook
-  log('Running without ts-node hook in environment "%s"', process.env.CYPRESS_INTERNAL_ENV)
-}
+require('./registerDir')()
```
flotwig, line 1: this could probably use a comment for why it's needed, it wasn't immediately clear how this differed from registerDir to me. is this comment correct?

```suggestion
// This file can be used as a shortcut to invoke `registerDir` without a scope.
// This is useful in Mocha tests with the `-r` option (`-r @packages/ts/register`)
require('./registerDir')()
```

packages/config/index.js
```diff
@@ -1,5 +1,5 @@
 if (process.env.CYPRESS_INTERNAL_ENV !== 'production') {
-  require('@packages/ts/register')
+  require('@packages/ts/registerDir')(__dirname)
```
flotwig, line 2: doesn't this same change need to be made everywhere we use `require('@packages/ts/register')` for full benefits?

flotwig (changes requested):
nice find in ts-node!

packages/config/index.js
elevatebart: Created ticket UNIFY-1548 to apply this to all packages.

flotwig (approved):

lmiller1990 (approved):

PR #20792, opened 2022-03-28
fix(launchpad): show more specific error when default supportFile is missing

lmiller1990:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1087

### User facing changelog

Improves error message for missing `supportFile` when `supportFile` is not explicitly set in `cypress.config.js` and the default supportFile is not found.

### Additional details

I added a specific error message for the use case of `supportFile` is **not** set, **and** the default supportFile is not found.

There as a `TODO` saying we should show the supportFile in the error message relative to the project root, so I fixed that, since it was on the same path as the code I was touching.

I also added a test ensuring custom supportFiles are respected. According to [this comment](https://cypress-io.atlassian.net/browse/UNIFY-1087?focusedCommentId=16969), they are not, but the system test for `ts-proj-custom-names`, which has a custom supportFile adding custom commands, works fine.

Note: we show the relative path for the supportFile in the error. Previously, it showed the absolute, with a `TODO: show relative path`.

image

### PR Tasks


emilyrohrbough (approved):
nice update 👍🏻

packages/server/test/unit/config_spec.js
```diff
@@ -2120,11 +2115,18 @@ describe('lib/config', () => {
...
+            from: 'default',
+          },
+        },
       })
 
-      return config.setSupportFileAndFolder(obj, mockSupportDefaults)
+      return config.setSupportFileAndFolder(obj)
       .catch((err) => {
-        expect(stripAnsi(err.message)).to.include('Your supportFile is missing or invalid:')
+        console.log('Error', err)
```
ZachJW34, line 2128: Remove?

ZachJW34 (approved):
Looks good!

lmiller1990:
Since receiving two approvals, I made a single commit removing a `console.log` from a spec: https://github.com/cypress-io/cypress/pull/20792/commits/dd9a6b23ba589ca3c9b1d73f68133840d64b3419

I'll merge this once CI is green - I don't think this change it worth re-requesting two reviews over.

PR #20798, opened 2022-03-28
feat: add maxExecution directive for controlling field resolution 

tgriesser:
- Closes UNIFY-1206, UNIFY-1396

### User facing changelog

Allows specifying guards against execution of a given GraphQL field taking longer than a certain amount of time, with an optional subscription field to "trigger" when the race fails and the value completes. 

Went through a few iterations / exploration here and found this was the best first-pass option for now. `@defer` fields seem promising, and are probably what we want for local data like the `gitInfo` / `branch` fields guarded here. 

Trying to guard against remote fields at all times makes it difficult to handle when there are situations we don't mind waiting for the delay (runs list page).

### Additional details

Any user on a slow network connection should not be blocked from using Cypess due to slow resolving fields.

Also bumped several urql deps trying to get to the root of an issue, which ended up being caused by our memoization of the specFiles. Need to check with @ZachJW34 to better understand the original need there.

### PR Tasks

- [x] (N/A) Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [x] (N/A) Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [x] (N/A) Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [x] (N/A) Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?


packages/data-context/src/sources/VersionsDataSource.ts
```diff
@@ -25,14 +25,20 @@ const NPM_CYPRESS_REGISTRY = 'https://registry.npmjs.org/cypress'
...
     this._currentTestingType = this.ctx.coreData.currentTestingType
-    this._latestVersion = this.getLatestVersion()
-    this._npmMetadata = this.getVersionMetadata()
+    this.#ensureData()
+  }
+
+  #ensureData () {
+    this.ctx.update((d) => {
+      d.versionData.latestVersion ??= this.getLatestVersion().catch((e) => undefined)
+      d.versionData.npmMetadata ??= this.getVersionMetadata().catch((e) => undefined)
```
tgriesser, line 38: Changed these to be stored in the `coreData` so they are cleared between e2e tests, since we currently reuse the same `ctx` class across all tests.

Would probably be useful to look into creating a new "DataContext" class per-spec and just copying over the server details

packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
```diff
@@ -159,70 +161,77 @@ async function makeE2ETasks () {
...
-                  result,
-                  callCount: operationCount[operationName ?? 'unknown'],
-                }, testState)
-              } catch (e) {
-                const err = e as Error
-
-                result = { data: null, extensions: [], errors: [new GraphQLError(err.message, undefined, undefined, undefined, undefined, err)] }
-              }
-            }
+      sinon.stub(ctx.util, 'fetch').callsFake(async (url: RequestInfo, init?: RequestInit) => {
```
tgriesser, line 164: Was having trouble following when/how this was mocked in different situations when working through the e2e test for ensuring the app starts up without delay. Making it a method made it easier to track / ensured the getter wasn't accessed & cached.

packages/graphql/src/schemaTypes/objectTypes/gql-CachedViewer.ts
```diff
@@ -0,0 +1,19 @@
+import dedent from 'dedent'
+import { objectType } from 'nexus'
+
+export const CachedViewer = objectType({
+  name: 'CachedViewer',
```
tgriesser, line 5: Currently not in use, was thinking it'd be useful as an eager response of the currently logged in user data to prevent a jump in the header if the `cloudViewer` is delayed due to network / if the user is offline.

packages/frontend-shared/cypress/e2e/support/e2eSupport.ts
```diff
@@ -126,6 +126,10 @@ declare global {
        * Simulates a user logged-out from the cypress app
        */
       // logoutUser: typeof logoutUser
+      /**
+       * Sets the delay for all remote fetches
+       */
+      setFetchDelay: typeof setFetchDelay
```
tgriesser, line 132: Could probably name this better, `setRemoteFetchDelay`?

packages/graphql/src/plugins/nexusMaxExecutionPlugin.ts
```diff
@@ -0,0 +1,74 @@
...
+
+// Added here because the field is defined automatically by the merged schema
+const CLOUD_EXECUTION: Record<string, Record<string, MaxExecutionConfig>> = {
+  Query: {
+    cloudViewer: {
+      duration: 1000,
+      triggerOnResult: 'cloudViewerChange',
+    },
+  },
+}
```
flotwig, line 26: hmm, shouldn't this go where the query itself is defined?

packages/graphql/src/plugins/nexusMaxExecutionPlugin.ts
```diff
@@ -0,0 +1,74 @@
...
+      if (info.operation.operation !== 'query' || info.operation.directives?.some((d) => d.name.value === 'skipMaxExecutionGuard')) {
+        return next(source, args, ctx, info)
+      }
+
+      const { duration, triggerOnResult } = maxExecutionConfig
+
+      let timeoutHit = false
+      const RACE_RESULT = {} as const
+
+      const result = await Promise.race([
```
flotwig, line 49: I don't think this closes UNIFY-1396 - there is still a 1 second delay when the launchpad initially loads if the user is on a slow network connection. Slow network should not block rendering:

> Desired behavior: All cloud operations happen asynchronously and do not block the launchpad rendering. Tests are added to simulate this condition

I'd expect a solution to this problem to asynchronously "fill in" the missing CloudData fields and never block rendering. Effectively, the `duration` here should always be 0. From what I understand, you have the subscriptions for "filling in" already set up, so would this be a simple change?

packages/graphql/src/schemaTypes/objectTypes/gql-CachedViewer.ts
flotwig: :grimacing: there is already a lot of unused code, it makes it difficult to search through the codebase, any way we can not introduce this file?

packages/launchpad/cypress/e2e/maxExecution-launchpad.cy.ts
```diff
@@ -0,0 +1,24 @@
+describe('maxExecution: launchpad', () => {
+  beforeEach(() => {
+    cy.scaffoldProject('todos')
+    cy.openProject('todos')
+  })
+
+  it('loads through to the browser screen when the network is slow', () => {
+    cy.setFetchDelay(10000)
```
flotwig, line 8: dang, i  wish we could use `cy.clock()` here. hello slow test!

flotwig (changes requested):

packages/launchpad/cypress/e2e/maxExecution-launchpad.cy.ts
tgriesser: Haha, just fixed this locally, keeping track of the timers & added a method to clear them `cy.clearFetchDelay`

packages/graphql/src/plugins/nexusMaxExecutionPlugin.ts
tgriesser: It will typically go on where the field is defined, but this field is defined dynamically because we're stitching in the GraphQL schema from the dashboard automatically.

packages/app/cypress.config.ts
```diff
@@ -44,6 +44,9 @@ export default defineConfig({
         throw new Error('HTTP_PROXY_TARGET_FOR_ORIGIN_REQUESTS is missing. Close Cypress and run tests using the `yarn cypress:*` commands from the `packages/app` directory')
       }
 
+      // Cy-in-cy currently only works in Chrome
+      config.browsers = config.browsers.filter((b) => b.name.includes('chrom'))
```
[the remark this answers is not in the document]
tgriesser: so many more characters 😆

packages/graphql/src/plugins/nexusMaxExecutionPlugin.ts
tgriesser: > Desired behavior: All cloud operations happen asynchronously and do not block the launchpad rendering. Tests are added to simulate this condition

Somewhat, I had it at close to zero and it felt janky, was looking to see if we could strike a balance between time to render and not having it feel jumpy. There's also the difference in that the launchpad is pre-fetching the data for the `cloudViewer` so it's not normally that bad. Maybe 500ms would be a decent middle-ground?

packages/graphql/src/schemaTypes/objectTypes/gql-CachedViewer.ts
tgriesser: If we go with the zero delay then we'll probably want to use this. Will revisit before it gets merged.

packages/app/cypress.config.ts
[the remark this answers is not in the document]
flotwig: BUT it will not be mistaken for a typo!

ZachJW34:
Seeing a lot of requests on the Specs List page (and in the terminal). Is there some polling/loop that this PR introduces?

https://user-images.githubusercontent.com/25158820/160492389-4248e926-43d2-447a-86f7-d3fb2c003ae4.mov

packages/app/src/pages/Settings.vue
```diff
@@ -17,4 +17,21 @@ query Settings {
 
 const query = useQuery({ query: SettingsDocument })
 
+gql`
+subscription Settings_cloudProjectUpdate {
```
ZachJW34, line 21: To use this feature, is it required that the field be a subscription? Whatever wiring is required, I think we should add as much documentation as we can for what these new directives do and when/why you'd implement them

packages/app/src/pages/Settings.vue
```diff
@@ -17,4 +17,21 @@ query Settings {
...
+  cloudProjectChange {
+    id
+    ...CloudSettings_CloudProject
+  }
+}
+`
+
+useSubscription({ query: Settings_CloudProjectUpdateDocument }, (current, next) => {
+  // Need to investigate why this is needed. When the subscription comes in the cache should update & automatically
```
ZachJW34, line 30: Is this something you are looking into before this PR is merged? If not, would be great to get an issue logged and linked.

packages/app/cypress.config.ts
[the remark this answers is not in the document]
tgriesser: Very good point

packages/app/src/pages/Settings.vue
tgriesser: Yes, still looking into this. Will log an issue if not

tgriesser:
> Seeing a lot of requests on the Specs List page (and in the terminal). Is there some polling/loop that this PR introduces?

@ZachJW34 Need to double check, but I think it may be due to the lack of batching in the `gitInfo` notification. Had a fix for that but didn't push it up yet.

packages/app/src/pages/Settings.vue
lmiller1990: I thought I had a similar bug, but the problem was we were caching data in the app layer (in Vue using `ref`). Did you check where this goes and if it's saved/cached anywhere along the line?

lmiller1990: My fix was around invalidating the spec list caching, landed here: https://github.com/cypress-io/cypress/pull/20744/files#diff-4227699afbcb444e02cc568cf473652e333a9a2847d0164d00e627061772ddd1R191

We need to look into our spec list, it's pretty messy. It was created when a bunch of the team was learning Vue, we know more and can likely do a better job there.

packages/data-context/src/sources/VersionsDataSource.ts
lmiller1990: Just to confirm, you mean per-spec for our internal tests, not per-spec for when the user executes specs? Or both?

packages/frontend-shared/src/gql-components/OpenFileInIDE.vue
```diff
@@ -38,7 +38,7 @@ const props = defineProps<{
   filePath: string
 }>()
 
-const query = useQuery({ query: OpenFileInIdeDocument, requestPolicy: 'network-only' })
+const query = useQuery({ query: OpenFileInIdeDocument, requestPolicy: 'cache-and-network' })
```
lmiller1990, line 41: Was this necessary? I think it might be `network-only` for a reason (could be wrong).

packages/frontend-shared/src/utils/spec-utils.ts
```diff
@@ -125,7 +126,7 @@ export function useCachedSpecs (specs) {
     const specsAreDifferent =
         currentSpecs.length !== prevSpecs.length ||
         currentSpecs.some(
-          (spec, idx) => spec.absolute !== prevSpecs[idx]?.absolute,
+          (spec, idx) => !_.isEqual(spec, prevSpecs[idx]),
```
lmiller1990, line 129: Ah, this is probably far better than my solution, which was passing a custom function to compare the lists: https://github.com/cypress-io/cypress/pull/20744/files#diff-4227699afbcb444e02cc568cf473652e333a9a2847d0164d00e627061772ddd1R191

When you rebase you can probably eliminate what I did. :facepalm:

packages/graphql/src/plugins/nexusMaxExecutionPlugin.ts
lmiller1990: I tried adding 

```js
const w = () => {
  return new Promise(res => setTimeout(res, 100000))
}
await w()
```

To simulate a request that never returns (I added to `VersionsDataSource`) but I couldn't even tell I was waiting for 1s. Did you give this a try locally @flotwig? I had the same instinct - thinking 1s is a really huge delay - but I didn't really notice it when I tried it locally.

packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
lmiller1990: This part of the code is pretty complex - it looks like it mostly exists for the purpose of just 2-3 tests, I wonder if we can change those tests so we don't even need this anymore.

lmiller1990 (commented):
I tried it out locally and it seems to work as advertised - I made a bunch of requests time out by awaiting a promise that would not resolve, and the launchpad/app feels pretty good still. I could barely notice a difference, actually, even with the 1s artificial delay.

I agree we need good docs around how to use and when to use the `maxExecution` directive - it's obvious we should attach it to anything that relies on a non deterministic request, but I'm not sure how to make this visible (unless you review this PR, it's unlikely you'll know about this new feature).

packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
tgriesser: > This part of the code is pretty complex - it looks like it mostly exists for the purpose of just 2-3 tests, I wonder if we can change those tests so we don't even need this anymore.

Not sure what "it" is referring to here. This code exists as the default handler for all remote requests in the tests. We've talked about creating a server module to handle this, but that'd probably be a separate task we'd want to design the requirements around.

packages/data-context/src/sources/VersionsDataSource.ts
tgriesser: Correct, per-spec for our internal tests

packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
tgriesser: Oh, realizing you're probably referring to the remote fetch stuff specifically. Yeah, agreed, seeing if I can extract that to just the tests that need it.

lmiller1990: Was referring the logic around stubbing `fetch` and specifically [this part](https://github.com/cypress-io/cypress/blob/10.0-release/packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts#L169) where we count the number of operations - I thought that was what you were referring to w.r.t a lot of complexity to grok.

It looks like only a handful of tests (cloud page) cares about the operation count - it's quite a bit of code for 1-2 tests. Now I'm thinking I may have misunderstood your comment.

Maybe this stub is doing far more than I'm realizing.

tgriesser: Oh my original comment was about the fact that I changed from `get` to `callsFake` which caused this big indent diff

packages/frontend-shared/src/utils/spec-utils.ts
tgriesser: This could probably be simplified further to just `_.isEqual(currentSpecs, prevSpecs)`... lodash's isEqual is always going to take the optional path (will start with `currentSpecs.length !== prevSpecs.length`, etc)

packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
lmiller1990: Yeah, the code is really different from when I originally commented. I'm just going to mark this as resolved.

packages/graphql/src/plugins/nexusMaxExecutionPlugin.ts
flotwig: Yeah, the delay was noticeable for me locally. @tgriesser and I met and talked about this, and decided to go ahead and effectively set the `duration` to 0. Two changes will reduce the "jumpiness" that Tim mentioned:

1. Caching the current cloud user (`CachedViewer`) and pre-hydrating it so it's ready at launchpad render time
2. Populate the Git Status information and cache it in-memory so it's ready when the app renders

I think he's going to create a task for (2) and do (1) in this PR, correct me if I'm wrong @tgriesser

packages/app/src/pages/Runs.vue
```diff
@@ -25,6 +26,10 @@ query Runs {
 
 const query = useQuery({ query: RunsDocument, requestPolicy: 'network-only' })
 
+onMounted(() => {
+  query.executeQuery()
+})
+
```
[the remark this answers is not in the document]
tgriesser: I've been digging in to understand better, I think there's potentially a bug in urql but haven't been able to fully isolate yet

tgriesser:
Close it in favor of a simpler approach in https://github.com/cypress-io/cypress/pull/21020 and https://github.com/cypress-io/cypress/pull/20997

PR #20806, opened 2022-03-28
chore: Only send honeycomb events for system tests run against binary

BlueWinds:
## Draft only, not ready for review.

- Closes https://cypress-io.atlassian.net/browse/UNIFY-1126

### User facing changelog
No user-facing changes.

### Additional details
This PR modifies our circle build to only report honeycomb events for system tests run against the final binary, which is more reflective of release performance.

### PR Tasks

- [N/A] Have tests been added/updated?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,9 @@ packages/proxy/lib/**/*.js
 # npm packages
 npm/**/cypress/screenshots
 
+# from app
+packages/app/cypress/videos
+
 # from example
 packages/example/app
 packages/example/build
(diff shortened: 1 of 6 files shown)


flotwig (dismissed):
> which is more reflective of release performance.

So this change isn't motivated by "the test runs aren't consistent in unbuilt Cypress", it's motivated by "test runs against built Cypress are more representative of real-world performance"?

If so, it seems like it would still be useful to capture the other `system-tests-*` jobs' performance metrics, since they can still be compared b/t commits.

PR #20817, opened 2022-03-29
fix: cy.root respect timeout option.

### LINKED ISSUE #19985: `cy.root` ignores `timeout` option
### Current behavior

Passing a `timeout` option to `cy.root` has no effect.

### Desired behavior

Like `get` and al., it should retry for the given timeout duration.

Note: there's a separate bug with `root` [not retrying *at all* when called from `within`](https://github.com/cypress-io/cypress/issues/19404). This seems different but may be related.

### Test code to reproduce

From [this demo](https://github.com/cypress-io/cypress-test-tiny/pull/72):

```js
describe("page", () => {
  beforeEach(() => {
    cy.visit("https://example.cypress.io");
    delayedHelloAfter(7);
  });

  it("`root` retries using the default so this test fails (good)", () => {
    cy.root().should("contain", "Hello");
  });

  it("`root` does not respect the `timeout` options so this test fails too (bad)", () => {
    cy.root({ timeout: 10000 }).should("contain", "Hello");
  });
  it("get('html') does respect the `timeout` options so this test passes (good)", () => {
    cy.get("html", { timeout: 10000 }).should("contain", "Hello");
  });
});

function delayedHelloAfter(seconds) {
  cy.window().then((w) => {
    w.eval(`setTimeout(() => {
        document.querySelector('.banner .container').innerHTML = '<p> Hello, world!</p>'
      }, ${seconds} * 1000)
      console.log("ok")
      `);
  });
}
```

### Cypress Version

9.3.1 & 9.4.0

### Other

_No response_

sainthkh:
- Closes #19985

### User facing changelog

`cy.root` respects `timeout` option.

### Additional details
- Why was this change necessary? => `timeout` option exists in `cy.root`, but it does not respect it. (It's always 4000ms.)

Code below works correctly.

```js
cy.root({ timeout: 4300 }).should('contain', 'root world')
```

### PR Tasks


packages/driver/cypress/integration/commands/querying/root_spec.js
```diff
@@ -35,6 +35,23 @@ describe('src/cy/commands/querying', () => {
       cy.root().should('have.class', 'foo').and('have.class', 'bar')
     })
 
+    // https://github.com/cypress-io/cypress/issues/19985
```
BlueWinds, line 38: Can we use a timeout of 150ms here (and setTimeout of 50ms), rather than 4200 + 4300?

BlueWinds (changes requested):

packages/driver/cypress/integration/commands/querying/root_spec.js
sainthkh: I rewrote the test like below:

```js
cy.on('fail', (err) => {
  expect(err.message).to.not.contain('4000ms')
  done()
})

cy.root({ timeout: 50 }).should('contain', 'root world')
```

I wrote the first test like that because the default timeout is 4000ms, so I needed to test that `cy.root` can wait for more than 4000ms. If it's shorter than that, we cannot be sure that `cy.root({ timeout: 150})` worked because it's fixed or 150ms is simply lower than 4000ms. 

But I realized that I can test it faster like the new code.

davidmunechika (approved):

PR #20818, opened 2022-03-29
chore(launchpad): work on infra for scaffold tests

lmiller1990:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1305

### User facing changelog

Add some code to assert the correct files are created for new projects. It makes it quickly and easy to add new projects and assert the scaffolded code is what we expect.

### Additional details

See this file for more details on my idea: https://github.com/cypress-io/cypress/pull/20818/files#diff-4b0c978082d6a9bf6dd7dfb2211caa0bf93f93a20da215a9907bd3e3c840137f

Basically, I added a function in `TestActions` that is kind of like snapshots for the scaffolded files. If a project hasn't got any expected files (in the `cypress-expected` directory) we just copy the scaffolded ones there, like a snapshot, for next time the test runs. 

If the files do exist, we assert the scaffolded files are the same as the expected ones.

When successful, a test looks like this:
simage

A failure shows the diff:

image

### PR Tasks


packages/launchpad/cypress/e2e/scaffold-project.cy.ts
```diff
@@ -0,0 +1,109 @@
...
+      return result
+    }
+
+    throw new Error(result.message)
+  }).then((res) => {
+    cy.log(`✅ ${res.message}`)
+  })
+}
+
+describe('scaffolding new projects', () => {
```
lmiller1990, line 89: We need more to express the scaffolded CT `cypress.config.js`, but since that's changing soon with the CT Arch project, I'll add those when we actually implement the changes.

packages/data-context/src/actions/TestActions.ts
```diff
@@ -0,0 +1,129 @@
+import path from 'path'
+import globby from 'globby'
+import assert from 'assert'
+import fs from 'fs-extra'
+import disparity from 'disparity'
+import tempDir from 'temp-dir'
+
+// Also defined in lib/fixtures, but we do not want to
+// import from system-tests into code we ship in production.
+const cyTmpDir = path.join(tempDir, 'cy-projects')
```
flotwig, line 10: why are we shipping test code in production? this seems very wrong

packages/launchpad/cypress/e2e/scaffold-project.cy.ts
```diff
@@ -0,0 +1,109 @@
...
+// in a directory named `expected-cypress-{lang}-{testingType}`.
+// For example, when configuring the `pristine` project using
+// plain JS for E2E testing, it will make
+// a new directory in `pristine` named `expected-cypress-js-e2e`
+// containing the scaffolded files.
+//
+// Each subsequent run will compare the scaffolded files in the
+// `expected-cypress-js-e2e` directory to the newly created ones.
+//
+// If there is a descrepancy, the test will fail and show the diff in the command log.
```
flotwig, line 19:
```suggestion
// If there is a discrepancy, the test will fail and show the diff in the command log.
```

packages/data-context/package.json
```diff
@@ -27,6 +27,7 @@
     "dataloader": "^2.0.0",
     "dayjs": "^1.9.3",
     "dedent": "^0.7.0",
+    "disparity": "^3.0.0",
```
flotwig, line 30: this and `temp-dir` should be in `devDependencies` since this is test code

flotwig (changes requested):
very cool!

system-tests/projects/pristine/expected-cypress-ts-component/cypress.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from 'cypress'
```
BlueWinds, line 1: So this is unrelated to the PR, but I think it would be a good idea to have a comment about what `defineConfig` actually does and why people should use it. As far as I understand it, its only job is to provide typing information, and is recommended but not required?

BlueWinds: We have a lot of comments in the other generated files, and it feels strange that the most important one doesn't explain what's going on in here at all.

BlueWinds (changes requested):

packages/data-context/src/actions/TestActions.ts
```diff
@@ -0,0 +1,129 @@
...
+
+interface FileToDiff {
+  actual: string
+  expected: string
+}
+
+export class TestActions {
+  constructor (private ctx: DataContext) {}
+
+  async #snapshotScaffoldedFiles (expectedScaffoldDir: string, filesToDiff: FileToDiff[]): Promise<SnapshotScaffoldTestResult> {
```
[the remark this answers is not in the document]
flotwig: It's a new shortcut for doing `private methodName() {`... I had to look it up too. Honestly I prefer `private` just because it's more visually distinct, but absent of a coding styleguide to back me up, I was gonna keep it to myself :smile:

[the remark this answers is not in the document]
BlueWinds: I think JS lends itself poorly to OOP / classes, and the less of it we do the better our code is going to come out.

packages/data-context/src/actions/TestActions.ts
lmiller1990: Yeah not super ideal - I will move this into a `cy.task`.

packages/data-context/src/actions/TestActions.ts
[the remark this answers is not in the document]
lmiller1990: Sure, we can just use `private`. My reasoning here was it makes more sense to use a native JS feature, which has runtime privacy guarantees, instead of a compile time only, proprietary TS feature where possible, but I don't think it makes too big a difference either way.

I don't agree private is more visually distinct though - `#` is so ugly it's impossible to not see it 😆

system-tests/projects/pristine/expected-cypress-ts-component/cypress.config.ts
lmiller1990: Yep, it's just for type interface - it's a no-op.

It might be good to explain what this does with a comment in the generated code, but changing the actual content of the scaffolded files needs to go through DX and friends. Luckily, if we do, it's trivial to update these tests with the infrastructure introduced here.

packages/data-context/src/actions/TestActions.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    const projectDir = this.ctx.currentProject.replace(cyTmpDir, systemTestsDir)
+    const projectName = projectDir.replace(systemTestsDir, '').slice(1)
+
+    const expectedScaffoldDir = path.join(projectDir, `expected-cypress-${this.ctx.coreData.wizard.chosenLanguage}-${this.ctx.coreData.currentTestingType}`)
+
+    const files = (
+      await Promise.all([
+        globby(path.join(this.ctx.currentProject, 'cypress'), { onlyFiles: true }),
+        globby(path.join(this.ctx.currentProject, 'cypress.config.*'), { onlyFiles: true }),
```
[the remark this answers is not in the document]
lmiller1990: Done. Please see https://github.com/cypress-io/cypress/pull/20818/files#diff-205e537e85adf50907257ece5985a0576754cefa95b0f691cc68bb8c11b7c3d0R98-R118

Shouldn't we just use `path.posix.join` all the time? Any case to not use it?

packages/data-context/package.json
lmiller1990: I moved these from `data-context` to `launchpad` when moving the test code to a task, and also moved them into `devDependencies`, so they are no longer in `dependencies`.

packages/data-context/src/actions/TestActions.ts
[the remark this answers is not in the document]
lmiller1990: I ended up moving everything into a task: https://github.com/cypress-io/cypress/pull/20818/files#diff-205e537e85adf50907257ece5985a0576754cefa95b0f691cc68bb8c11b7c3d0 which is just a function, no class, so this problem is gone.

I do think it's better to use native JS runtime features where we can over TS compile time features, though, but we can discuss that at some point.

lmiller1990:
@BlueWinds @flotwig I addressed all the feedback, can you please re-review? Thanks!

lmiller1990:
Ok, it's green again! :pray:

packages/graphql/tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "script"
   ],
   "compilerOptions": {
+    "lib": ["es2020"],
```
ZachJW34, line 12: Was this changes to accommodate the `#myVar` syntax? Should this be reverted?

packages/launchpad/cypress/e2e/scaffold-project.cy.ts
```diff
@@ -0,0 +1,116 @@
...
+  })
+
+  it('scaffolds E2E for a TS project', () => {
+    scaffoldAndOpenE2EProject('pristine', 'ts')
+    assertScaffoldedFilesAreCorrect('ts', 'e2e')
+  })
+
+  it('scaffolds CT for a JS project', () => {
+    scaffoldAndOpenCTProject('pristine', 'js', 'Create React App (v5)')
+    assertScaffoldedFilesAreCorrect('js', 'component')
```
ZachJW34, line 109: This will create a folder called `expected-cypress-js-component`, even though it is a `create-react-app` specific scaffold. I don't see this being used more than 4 types per project (2 for testing type, two for language). Should the framework be included in the scaffolded directory or does it make sense to have this per system-test project (where the actual system-test project provides the context of what selection will be chosen)?

ZachJW34 (dismissed):
Looks great! Could this system be used for some of the existing system-test projects we have  like `vue-cli-{configured,unconfigured}`? Had a few minor comments but going to approve.

packages/launchpad/cypress/tasks/snapshotsScaffold.ts
```diff
@@ -5,7 +5,7 @@ import disparity from 'disparity'
 import dedent from 'dedent'
 import { cyTmpDir } from '@tooling/system-tests/lib/fixtures'
 
-const systemTestsDir = path.join(__dirname, '..', '..', '..', '..', 'system-tests', 'projects')
+const systemTestsDir = path.posix.join(__dirname, '..', '..', '..', '..', 'system-tests', 'projects')
```
flotwig, line 8: on Windows, `__dirname` will have backslashes in it, and path.posix.join will not do what is expected - it will treat the entire `__dirname` as a single filename: 
image

I'd just keep this as a native `path.join` and convert to `posix` paths at only the places necessary.

flotwig (changes requested):

packages/launchpad/cypress/e2e/scaffold-project.cy.ts
lmiller1990: I see what you are saying. I'm not entirely sure of the best way to handle component setup, since there's lots of different combinations. Tim is also working on some testing infra as part of the CT architecture project - I wonder if we rethink this (the CT specific scaffolding) again once that lands?

My initial idea was we'd use different projects for each CT scaffold test - eg instead of "pristine", we might use "vuecli-unconfigured", so it wouldn't be necessary to have the framework in the name, but now I'm thinking we just use a single project and include the CT framework in the folder too, like you are suggesting. I think I'll make that change now.

packages/launchpad/cypress/tasks/snapshotsScaffold.ts
lmiller1990: Right, I see - thanks.

packages/launchpad/cypress/e2e/scaffold-project.cy.ts
lmiller1990: I've made this change @ZachJW34. https://github.com/cypress-io/cypress/pull/20818/commits/1af1e6c311781ec16a528fc2db2886878454cee8

packages/launchpad/cypress/tasks/snapshotsScaffold.ts
lmiller1990: Ok, I think I've done this right now. We only use posix for the globby globs now. Can you please double check? Thanks.

packages/data-context/src/actions/TestActions.ts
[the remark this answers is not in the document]
lmiller1990: Sounds good.

packages/launchpad/cypress/e2e/scaffold-project.cy.ts
ZachJW34: I've seen some of the work Tim is setting up for testing infra. We've got a few systems in place that work well (I like the changes you made to accommodate the CT permutations), curious to see how we can leverage them together with Tim's work.

ZachJW34 (approved):

packages/launchpad/cypress/e2e/scaffold-project.cy.ts
lmiller1990: Yep, let's see what improvements we can make overall post CT arch changes landing.

packages/graphql/tsconfig.json
lmiller1990: Yes, not sure we need to revert or not - seems nice to use the latest `es` version in general?

lmiller1990:
One more set of 👀 @flotwig 🙏

BlueWinds (approved):

lmiller1990:
Still managed to mess up the path stuff and broke windows. Fixing... https://github.com/cypress-io/cypress/pull/20914

PR #20829, opened 2022-03-29
chore: [Multi-domain]: Support spec bridges with same domain

mjhenkes:
- Closes 

### User facing changelog

n/a

### Additional details

This pr modifies the spec bridge and the communicator to be keyed off origin policy instead of domain to accommodate spec bridges that have the same domain but different protocols or ports. 

### PR Tasks


packages/driver/cypress/integration/e2e/multi-domain/multi_domain_spec.ts
```diff
@@ -275,4 +275,20 @@ describe('domain validation', { experimentalSessionSupport: true }, () => {
     .invoke('text')
     .should('equal', 'Welcome TJohnson')
   })
+
+  it('creates a spec bridge for https://example.com', () => {
+    cy.visit('/fixtures/auth/index.html') // Establishes Primary Domain
+    cy.switchToDomain('example.com', () => {
+      cy.visit('https://example.com')
```
mschile, line 282: Suggest using one of the hosts in the `cypress.json` file instead of loading a real page.

packages/driver/src/multi-domain/communicator.ts
```diff
@@ -169,14 +169,14 @@ export class SpecBridgeDomainCommunicator extends EventEmitter {
    * @param {Cypress.ObjectLike} data - any meta data to be sent with the event.
    */
   toPrimary (event: string, data?: Cypress.ObjectLike, options: { syncGlobals: boolean } = { syncGlobals: false }) {
-    debug('<= to Primary ', event, data, document.domain)
+    debug('<= to Primary ', event, data, location.origin)
```
mschile, line 172:
```suggestion
    debug('<= to Primary ', event, data, location.originPolicy)
```

packages/driver/src/multi-domain/communicator.ts
```diff
@@ -169,14 +169,14 @@ export class SpecBridgeDomainCommunicator extends EventEmitter {
...
     if (options.syncGlobals) this.syncGlobalsToPrimary()
 
     this.handleSubjectAndErr(data, (data: Cypress.ObjectLike) => {
       window.top?.postMessage({
         event: `${CROSS_DOMAIN_PREFIX}${event}`,
         data,
-        domain: document.domain,
+        originPolicy: location.origin,
```
mschile, line 179:
```suggestion
        originPolicy: location.originPolicy,
```

mschile (changes requested):

packages/driver/cypress/integration/e2e/multi-domain/multi_domain_spec.ts
mjhenkes: can do, i didn't realize you needed a different port for https. Will update!

packages/driver/src/multi-domain/communicator.ts
mjhenkes: location.origin === location.originPolicy in this instance because the location is from the spec bridge, but to reduce confusion I'll grab the origin policy anyway.

packages/driver/cypress/integration/e2e/multi-domain/multi_domain_spec.ts
```diff
@@ -275,4 +275,20 @@ describe('domain validation', { experimentalSessionSupport: true }, () => {
     .invoke('text')
     .should('equal', 'Welcome TJohnson')
   })
+
+  it('creates a spec bridge for https://idp.com:3502', () => {
```
mschile, line 279: Should these tests be in `multi_domain_validation_specs` where we are validating the src on the spec bridges?

mjhenkes: 🤷‍♂️  sure, I thought that one was more about validating command params, but i can move this whole describe block over there.

mschile (approved):

AtofStryker (approved):

PR #20830, opened 2022-03-29
fix: Ensuring current browser is synchronized between app and launchpad

tbiethman:
- Closes UNIFY-1395

When a new browser is selected from the app's top navigation, the current browser instance is destroyed and the new browser is opened. As part of this lifecycle, the currentBrowser field is updated, and the browserStatus field is changed ('closed' -> 'opening' -> 'open'), with each change being broadcast through the `browserStatusChange` subscription event.

The launchpad remains open throughout this process and should be updated by its subscription to the 'browserStatusChange` event. However, because we are not using the currentBrowser field in the subscription definition, we are not getting notified that its value has changed and thus continue to show the previous browser as selected/open.

I updated the subscription to include the currentBrowser so that any changes to browserStatus will include any relevant corresponding changes to the currentBrowser. This ensures that the currentBrowser is always synchronized properly.

I also cleaned up a few things around this logic and updated tests where necessary.

### User facing changelog

* Fixes an issue where changing the browser from the app does not reflect in the launchpad

https://user-images.githubusercontent.com/1711637/160693180-ff459067-36d0-4029-9d63-6bd6a1debc77.mov

### PR Tasks


packages/app/src/runner/SpecRunnerHeaderOpenMode.vue
```diff
@@ -171,26 +171,6 @@ fragment SpecRunnerHeader_Browser on Browser {
...
-      id
-      isSelected
-    }
-  }
-  launchOpenProject(specPath: $specPath) {
-    id
-  }
-}
-`
-
```
tbiethman, line 193: This mutation is no longer used.

packages/launchpad/src/setup/OpenBrowserList.vue
```diff
@@ -55,7 +55,7 @@
           </div>
           <div
             class="font-medium pt-2 text-indigo-600 text-18px leading-28px"
-            :class="{ 'text-jade-600': browser.isSelected, 'text-gray-500': browser.disabled || !browser.isVersionSupported }"
+            :class="{ 'text-jade-600': browser.id === selectedBrowserId, 'text-gray-500': browser.disabled || !browser.isVersionSupported }"
```
tbiethman, line 58: Updated this to ensure we're always reflecting the currentBrowser value in our selections.

packages/launchpad/src/setup/OpenBrowserList.vue
```diff
@@ -238,7 +236,11 @@ const browsers = computed(() => {
 const setBrowser = useMutation(OpenBrowserList_SetBrowserDocument)
 
 const selectedBrowserId = computed({
-  get: () => props.gql.currentBrowser?.id || props.gql.browsers?.find((browser) => browser.displayName === 'Electron')?.id,
+  get: () => {
+    // NOTE: The currentBrowser is set to the first detected browser
+    // found during project initialization. It should always be defined.
+    return props.gql.currentBrowser?.id
+  },
```
tbiethman, line 243: I don't think the defaulting value logic was ever executed outside of component tests, as we always default the currentBrowser to the first browser detected on the user's system, which would always at least include Electron. So I removed that bit.

I also updated the markup so that if we end up in a scenario where we don't have a currentBrowser, we do not show the action buttons and allow the user to open a browser when nothing is actually selected.

packages/launchpad/src/setup/OpenBrowserList.vue
```diff
@@ -182,21 +183,14 @@ fragment OpenBrowserList on CurrentProject {
...
-    name
-    family
     disabled
-    isSelected
     isVersionSupported
-    channel
+    name
     displayName
-    path
-    version
```
tbiethman, line 199: I removed fields that weren't being used anymore

packages/launchpad/src/setup/OpenBrowserList.vue
ZachJW34: Could we not have used the `checked` slot prop? We've got so many conditionals in this template 😅

ZachJW34 (approved):
Awesome PR description, tested and it works great!

packages/launchpad/src/setup/OpenBrowserList.vue
tbiethman: Ah good call, that works great. I'll have to do some reading on how the slot props work.

ZachJW34 (approved):

flotwig (approved):

PR #20839, opened 2022-03-30
fix: `cy.type()` + `{enter}` or ` ` should work with Firefox 98+

### LINKED ISSUE #20562: `{enter}` not working in Firefox 98
### Current behavior

When sending `{enter}`, it doesn't work in Firefox 98, while it did work in Firefox 97.

### Desired behavior

Sending `{enter}` has the same behavior in Firefox 98 as in older Firefoxes and other browsers.

### Test code to reproduce

```javascript
describe('Hitting enter', () => {
  it('works', () => {
    cy.visit('https://www.duckduckgo.com/');
    cy.get('input#search_form_input_homepage').type('hello{enter}');
    cy.get('div#links_wrapper').should('exist');
  });
});
```

### Cypress Version

9.5.1 (reproduced in 7.5.0)

### Other

_No response_

sainthkh:
- Closes #20562

### User facing changelog

`cy.type()` + `{enter}` or ` ` should work with Firefox 98+

### Additional details
- Why was this change necessary? => Some `click` event related behaviors of Firefox has been changed with 98. This breaks `cy.type()`.

Before Firefox 98, it automatically sent `click` event to all button-like elements. But after 98, it only does that with `<button>` element and does nothing with other button-like `<input>` elements. 

In addition, it does not change the state of `<input type="checkbox">` `<input type="radio">`. I fixed that, too. 

`cy.type()` works well with Firefox 98+

### IMPORTANT NOTE

Currently Cypress CI uses Firefox 93. We couldn't detect this problem because of that. And the current CI cannot test this scenario. This should be tested manually.

### PR Tasks

### DIFF
--- a/packages/driver/src/cy/commands/actions/type.ts
+++ b/packages/driver/src/cy/commands/actions/type.ts
@@ -334,9 +334,16 @@ export default function (Commands, Cypress, cy, state, config) {
           }
 
           if (
-            // Firefox sends a click event when the Space key is pressed.
-            // We don't want to send it twice.
-            !Cypress.isBrowser('firefox') &&
+            (
+              // Before Firefox 98,
+              // Firefox sends a click event when the Space key is pressed.
+              // We don't want to send it twice.
+              !Cypress.isBrowser('firefox') ||
+              // After Firefox 98,
+              // it sends a click event automatically if the element is a <button>
+              // it does not if the element is an <input>
+              (Cypress.isBrowser('firefox') && Cypress.browserMajorVersion() >= 98 && $elements.isInput(event.target))
+            ) &&
             // Click event is sent after keyup event with space key.
             event.type === 'keyup' && event.code === 'Space' &&
             // When event is prevented, the click event should not be emitted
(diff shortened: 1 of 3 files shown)


jennifer-shehane:
Thanks again @sainthkh! We’ll review as soon as possible.

packages/driver/src/cy/commands/actions/type.ts
```diff
@@ -404,8 +421,13 @@ export default function (Commands, Cypress, cy, state, config) {
           // https://github.com/cypress-io/cypress/issues/19541
           // Send click event on type('{enter}')
           if (sendClickEvent) {
-            // Firefox sends a click event automatically.
-            if (!Cypress.isBrowser('firefox')) {
+            if (
```
BlueWinds, line 424: I find this conditional fairly hard to read, split as it is by comments.
```
// Earlier in file
isFirefoxBefore98 = () => (Cypress.isBrowser('firefox') && Cypress.browserMajorVersion() < 98)

// Firefox versions before 98 always sends a click event automatically.
// Firefox 98+ only send click events automatically if the element is a <button>, but not for <input>s.
if (sendClickEvent && (!isFirefoxBefore98() || $elements.isInput(el))) {
  fireClickEvent(el)
}
```

BlueWinds (approved):

packages/driver/src/cy/commands/actions/type.ts
sainthkh: Done. 

But I defined `isFirefoxBefore98` to be a `const` because its browser version doesn't change on runtime.

tbiethman (approved):
Affected tests pass locally with FF98. There are other unrelated FF98 failures that we will need to address when we bump the CI image used.

BlueWinds (approved):

PR #20849, opened 2022-03-30
fix: viewport from CLI on CT

estrada9166:
- Closes  [UNIFY-1366](https://cypress-io.atlassian.net/browse/UNIFY-1366)

### Additional details

Fix viewport config params pass through CLI for CT - currently the usage of `--config viewportWidth=333,viewportHeight=333` was not working for CT because it was being overridden in a config update; it was working fine for e2e tests

### PR Tasks


packages/config/src/options.ts
```diff
@@ -384,12 +384,12 @@ const resolvedOptions: Array<ResolvedConfigOption> = [
...
   }, {
     name: 'viewportHeight',
-    defaultValue: 660,
+    defaultValue: (options: Record<string, any> = {}) => options.testingType === 'component' ? 500 : 660,
     validation: validate.isNumber,
     canUpdateDuringTestTime: true,
   }, {
     name: 'viewportWidth',
-    defaultValue: 1000,
+    defaultValue: (options: Record<string, any> = {}) => options.testingType === 'component' ? 500 : 1000,
```
lmiller1990, line 392: Should this be `options: Cypress.Config`?

lmiller1990 (commented):
One comment around types

packages/config/src/options.ts
estrada9166: I'm not sure if it should be `Cypress.Config` because it comes from [here](https://github.com/cypress-io/cypress/blob/10.0-release/packages/server/lib/config.ts#L117) which is when we are building the config in runtime

lmiller1990 (approved):

packages/config/src/options.ts
lmiller1990: Ok, seems fine then. 👍

lmiller1990:
cc @davidmunechika

davidmunechika (approved):

PR #20852, opened 2022-03-30
fix(desktop-gui): use native file picker on non-GTK desktop environments

flotwig:
### User facing changelog

- Fixed an issue on Linux where the GTK file picker would be used in non-GTK desktop environments.

### Additional details

- Switched to KDE from GNOME recently and noticed that Cypress doesn't respect the DE's file picker
- See https://tristan.partin.io/blog/2021/04/01/electron-linux-and-your-file-chooser/ for the reasoning and below for screenshots

#### Before (on KDE)

image

#### After (on KDE)

image

#### After (on GNOME, or when a `xdg-desktop-portal` package is not present)

image

### PR Tasks


cli/lib/exec/spawn.js
```diff
@@ -85,7 +85,6 @@ module.exports = {
 
     _.defaults(options, {
       dev: false,
-      env: process.env,
```
emilyrohrbough, line 88: why remove the default here?

cli/lib/exec/open.js
```diff
@@ -48,11 +49,21 @@ module.exports = {
   start (options = {}) {
     const args = processOpenOptions(options)
 
+    function getOpenModeEnv () {
+      if (os.platform() !== 'linux') return
+
+      // on non-GTK desktop environments (KDE, XFCE, ...), native file pickers will not be used in Electron unless
+      // we explicitly opt-in the the GTK portal functionality
+      // @see https://tristan.partin.io/blog/2021/04/01/electron-linux-and-your-file-chooser/
+      return { GTK_USE_PORTAL: '1' }
```
emilyrohrbough, line 58: Just curious, is this something that could be validated and set in `packages/electron` instead of needing to setup & pass this ENV around?

cli/lib/exec/open.js
```diff
@@ -48,11 +49,21 @@ module.exports = {
   start (options = {}) {
     const args = processOpenOptions(options)
 
+    function getOpenModeEnv () {
```
emilyrohrbough, line 52: nit: feels like this name should be something about `getNativeFilePickerEnvs`

BlueWinds:
Hm. Possibly because I don't have the appropriate gtk modules installed, I'm seeing your "After" filepicker across all branches and versions - it doesn't seem to make any difference to me. Which goes to show, I suppose, that it's at the very least a harmless change.

cli/__snapshots__/verify_spec.js
```diff
@@ -154,6 +154,7 @@ STRIPPED
 
 
 
+
```
BlueWinds, line 157: This newline seems to be causing the unit tests to fail in CI.

BlueWinds (changes requested):

BlueWinds (approved):

BlueWinds:
@flotwig - I believe we discussed this in slack, and you decided the issue was a relic of your setup. Still planning on coming back to this PR?

flotwig:
@BlueWinds I'll close this for now, if I see it again I'll reopen.

PR #20854, opened 2022-03-30
fix: windows build

estrada9166:
- Closes  [UNIFY-1377](https://cypress-io.atlassian.net/jira/software/projects/UNIFY/boards/20?selectedIssue=UNIFY-1377)

### Additional details

Currently, the build on windows machines is not working as expected because there's a bash script that is not supported unless extra packages are installed on the machine.

Also, `rm` is not supported at all, so there's the need to update the usage of it for `rimraf` which is fully supported!

Update failing migration test where the file path is not matching on windows machines

### PR Tasks


elevatebart:
I believe bash was already required for `run-if-not-ci.sh`

elevatebart:
I like TypeScript as it is easier to debug than bash.

packages/icons/scripts/build.ts
```diff
@@ -0,0 +1,26 @@
...
+  const assetsPath = path.join(__dirname, '../assets')
+  const iconsPath = path.join(distPath, 'icons')
+  const iconsetPath = path.join(distPath, 'cypress.iconset')
+
+  await fs.remove(distPath)
+  await fs.mkdir(iconsPath, { recursive: true })
+
+  await exec('yarn tsc -p ./tsconfig.build.json')
+
+  if (os.platform() === 'darwin') {
```
ZachJW34, line 17: Not related to this PR, but does this platform specific build have any effect on what the binary expects? When we build the Cypress binary for windows, will it have `dist/icons/cypress.icns` available and is it needed? Maybe @elevatebart has an idea

elevatebart (dismissed):

package.json
```diff
@@ -40,7 +40,7 @@
     "effective:circle:config": "circleci config process circle.yml | sed /^#/d",
```
tbiethman, line 40: The 'clean-deps' command here at root is still using `rm`

packages/config/package.json
```diff
@@ -8,8 +8,8 @@
   "scripts": {
     "build-prod": "tsc || echo 'built, with errors'",
     "check-ts": "tsc --noEmit",
-    "clean-deps": "rm -rf node_modules",
-    "clean": "rm -f ./src/*.js ./src/**/*.js ./src/**/**/*.js ./test/**/*.js || echo 'cleaned'",
+    "clean-deps": "rimraf node_modules",
+    "clean": "rimraf ./src/*.js ./src/**/*.js ./src/**/**/*.js ./test/**/*.js || echo 'cleaned'",
```
tbiethman, line 12: These clean commands that delete implementation files are wild; are they actively used?

package.json
estrada9166: Yeah! I didn't changed bc it was changed 2 years ago and seems not to be an issue - but may worth it changing to be consistent across all the packages

image

packages/config/package.json
estrada9166: It seems those are actively used - the screenshot is running `yarn` in `10.0-release` branch

image

packages/icons/scripts/build.ts
elevatebart: @ZachJW34, the answer is no and it does not need it. the only moment where we "need" the icns file is when we package the electron app for mac. `icns` is the mac specific icon format that contains multiple definitions. Since this electron app can only be built and signed on Mac, I am not afraid of this happening.

packages/config/package.json
elevatebart: Yeah, we don't use dist folders at cypress to avoid navigating to type files when changing packages. source maps would be a solution, but it does not seem to be one we considered for dev. we prefer using ts-node.

elevatebart: When we run build-prod, on local, it will create a bunch of files that need to be cleaned up.

elevatebart (approved):

elevatebart (approved):

lmiller1990 (approved):
Code seems fine but how did you test this on windows? Are you using windows? Should we enable windows for this PR using `circle.yml` to see if it actually all works?

elevatebart:
I tested it on windows.

lmiller1990:
Ok - so once this merges, and windows runs on the 10.0-release it should be green? Are there any outstanding windows problems?

I'll leave merging this to @estrada9166 :pray:

marktnoonan:
I'm going to make a branch off of this now that runs the windows tests, so we can see them run in CI before merging to 10.0-release without modifying this branch's `circle.yml`, I will update with results when they have completed.

scripts/run-postInstall.js
```diff
@@ -0,0 +1,12 @@
+const os = require('os')
+const { execSync } = require('child_process')
+
+if (os.platform() === 'win32' && !process.env.CI) {
```
marktnoonan, line 4: Excluding CI for this caused the `postinstall` step in the `windows-node-modules-install` job to fail on Windows in CI due to the non-windows version of this code running. Can we remove the CI check safely, or do we need another condition for windows on CI?

Here is a copy [run based on a copy of this branch with windows tests enabled](https://app.circleci.com/pipelines/github/cypress-io/cypress/35384/workflows/5275720e-e097-4a07-a8af-860b05258000/jobs/1408820) in CI. We see this error:

```
Error: Command failed: patch-package && ./scripts/run-if-not-ci.sh yarn-deduplicate --strategy=highest && yarn clean && gulp postinstall && ./scripts/run-if-not-ci.sh yarn build
```

marktnoonan (changes requested):
Looks like we need a little more logic in `scripts/run-postInstall.js` - or just to remove the CI check, I haven't looked in detail yet, just wanted to make sure we don't merge yet. I'll look into it a bit more and try some changes on the `win-fix-windows-build-ci` branch. If that starts passing I'll make a PR into this with the changes.

tbiethman (approved):

marktnoonan (approved):
Approving this now as the windows build succeeds, I expect a few windows tests may continue to fail in CI after this PR based on letting CI run already in a different branch, but that's fine, this PR is not intended to fix all outstanding windows test issues.

PR #20856, opened 2022-03-30
feat: bundling react, vue, and mount-utils with cypress binary

JessicaSachs:
- Closes [UNIFY-1386](https://cypress-io.atlassian.net/browse/UNIFY-1386)

### User facing changelog
Cypress users no longer have to install the Vue or React mounting libraries separately when mounting components. If you are using Vue 3 or React, you can remove those dependencies from your `package.json` entirely and replace the imports to the newly supported `cypress/vue` and `cypress/react` namespaces.

For Vue 2 users, a separate pull request will add support for importing mount from `cypress/vue-2`.

#### Before: Externally Install Mounting Libs
Mounting libraries are always explicitly imported from the Cypress npm organization namespace.

```js
// React
import { mount } from '@cypress/react'

// Vue 3
import { mount } from '@cypress/vue' // v3.x, pinned in your package.json

// Vue 2
import { mount } from '@cypress/vue' // v2.x, pinned in your package.json
```

#### After: Bundling Mounting Libs
Mounting libraries are shipped with Cypress, and are exposed as subpaths form the Cypress namespace.

This API is supported on a bundler-by-bundler basis and relies on **webpack's** and **vite's** `resolve.alias`.

If you do not want to rely on this bundler-specific feature (or if you're using a bundler that doesn't support aliases) you can import the mounting library explicitly by adding it to your **package.json** via `npm install @cypress/react -D` or `npm install @cypress/vue -D`

```js
// React

// Vue 3
import { mount } from 'cypress/vue'

// Vue 2, still have to npm install the pinned version of @cypress/vue for now
import { mount } from '@cypress/vue'
```


tgriesser (changes requested):
Nice - added a few thoughts

cli/package.json
```diff
@@ -101,6 +101,9 @@
     "bin",
     "lib",
     "index.js",
+    "vue/**",
```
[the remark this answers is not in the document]
tgriesser: Technically there's nothing actually exported from this module, it's just the types that live here. We're resolving via webpack & vite aliases to the code bundled with Cypress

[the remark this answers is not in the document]
JessicaSachs: We do, but it's patched via `resolve.alias`, it's not a real file.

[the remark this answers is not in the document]
JessicaSachs: I'm not particularly happy about this approach, honestly, but it seems like prettier + eslint don't through errors so long as there is a `d.ts` file there and this was the approach that Tim spiked last week, and I don't have a reason to change it.

lmiller1990:
> For Vue 2 users, a separate pull request will add support for importing mount from cypress/vue-2.

Any ticket to track this? So we don't forget.

npm/vite-dev-server/src/resolveConfig.ts
```diff
@@ -31,10 +32,48 @@ export const createConfig = async ({ options, viteConfig: viteOverrides = {} }:
...
+        //
+        // If `resolve.alias` did not map to a cypressBinaryRoot outside of the user's own project,
+        // then this line would not be necessary.
+        allow: [
+          searchForWorkspaceRoot(process.cwd()),
+          binaryRoot,
+        ],
+      },
+    },
+    resolve: {
```
lmiller1990, line 61: Will there be any problem with someone implementing a third party dev server - will they need to do a similar alias, or is this only needed for dev-servers bundled with the binary?

packages/config/src/options.ts
```diff
@@ -431,6 +432,11 @@ const runtimeOptions: Array<RuntimeConfigOption> = [
     // so we don't consider it a "public" option
     isInternal: true,
     canUpdateDuringTestTime: false,
+  }, {
+    name: 'cypressBinaryRoot',
+    defaultValue: path.join(__dirname, '..', '..', '..'),
+    validation: validate.isString,
+    isInternal: true,
```
lmiller1990, line 439: Should we add the same as line 434: `    canUpdateDuringTestTime: false,`?

lmiller1990 (commented):
Few minor comments/questions!

I did not test this by hand yet.

npm/vite-dev-server/src/resolveConfig.ts
tgriesser: Yes, if someone implements a third-party dev server they'd need to do this as well, or they'd need to use the `@cypress/` bundled packages. 

I guess that's another potential benefit of exporting the actual rollup dist files from the CLI rather than relying on this approach.

npm/vite-dev-server/src/resolveConfig.ts
```diff
@@ -31,10 +32,48 @@ export const createConfig = async ({ options, viteConfig: viteOverrides = {} }:
...
+        // Cypress will load npm/* projects as external modules from the binary's location.
+        // Because the Cypress binary is not located within the user's node_modules,
+        // Vite needs to be explicitly allowed to traverse the cypressBinaryRoot directory.
+        //
+        // If `resolve.alias` did not map to a cypressBinaryRoot outside of the user's own project,
+        // then this line would not be necessary.
+        allow: [
+          searchForWorkspaceRoot(process.cwd()),
+          binaryRoot,
+        ],
```
tgriesser, line 58:
```suggestion
        allow: [
          ...(viteOverrides?.server?.allow ?? []),
          searchForWorkspaceRoot(process.cwd()),
          binaryRoot,
        ],
        ...(viteOverrides?.server ?? {})
```

system-tests/projects/create-react-app-configured-bundled-mount/README.md
```diff
@@ -0,0 +1,70 @@
+# Getting Started with Create React App
```
BlueWinds, line 1: This looks like the README that CRA creates - can we remove it for this system test, or better yet, include a brief explanation of what this project is testing? a'la

https://github.com/cypress-io/cypress/tree/10.0-release/system-tests/projects/react-a11y
or
https://github.com/cypress-io/cypress/tree/10.0-release/system-tests/projects/migration-e2e-no-plugins-support-file.

BlueWinds: Same question with a lot of files in this repo - can we remove the react logo, cut down the dependency list (eg, remove testing-library), remove App.css and otherwise cut this down to show what's trying to be tested more precisely.

JessicaSachs: You're spot-on, but there's already two big PRs in progress to clean up the duplication of system tests by Tim and Lachlan... so I'm not touching that infrastructure right now.

npm/vite-dev-server/src/resolveConfig.ts
JessicaSachs: The reason I'm not doing this everywhere is because of Vite's own `mergeConfig` which manages all of the deeply nested merging.

system-tests/projects/create-react-app-configured-bundled-mount/README.md
BlueWinds: I don't mean the infrastructure, I mean the new files added directly here in this PR.

npm/vite-dev-server/src/resolveConfig.ts
lmiller1990: Hm, so are we going with the proposed approach here? Or do we want to explore alternatives? Or a combination? I think first class third party support is pretty important - it's going to be much more likely an ecosystem emerges if we have seamlessly third party integration.

lmiller1990:
I fixed some problems.

1. Bug in upstream (Test Utils). Commit: https://github.com/cypress-io/cypress/pull/20856/commits/7a207063e19cf64dda62c23e0f5c59d6c95f8906

Basically

```js
mount(Card, {
  slots: {
      header: '<h1>HEADER</h1>',
      footer: '<div>FOOTER</div>',
    },
  })
)
```

Is not working. [Upstream issue](https://github.com/vuejs/test-utils/issues/1166). I used the work-around (see the commit).

2. Multiple Vue versions were imported and conflicting

Fixed using Vite's `dedupe` feature: https://github.com/cypress-io/cypress/pull/20856/commits/98d0674e914ea9eadddcde07329662c45a9d7943. I am not sure exactly how multiple Vue versions were conflicting, but they were. Potentially related to the various things we do with cached `node_modules` and symlinks for CI - could not reproduce outside of system tests.

JessicaSachs:
As is, this PR doesn't work when the build binary is used against a basic Vite + Vue project. The copy of `@cypress/vue` fails to resolve user dependencies. I'm trying an alternative approach where we just ship the dist'd files underneath Cypress.

JessicaSachs:
Closing this PR because in favor of the real-files approach using the exports map in #20930.

PR #20858, opened 2022-03-31
fix: UNIFY-1389 proper absolute pathing in stack_utils

tgriesser:
- Closes UNIFY-1389

### User facing changelog

We were using `path.join` rather than `path.resolve` for constructing the absolute path in a stack frame, where the stack trace may contain absolute paths.

Also accounts for `webpack:///` custom protocol, previously it was stripping all trailing slashes rather than just the first three, if the path is absolute.

Also fixes the `cwd` not being reset after each spec execution in Cypress open-mode tests, which affects the behavior of webpack bunding, manifesting as showing absolute rather than relative paths in error traces.

### PR Tasks


chrisbreiding (approved):

mapsandapps:
i hate to be the bearer of bad news, but i can still reproduce this bug (same steps as my video in the ticket) on this branch 😕

tgriesser:
@mapsandapps ah, yep I see. We need to properly stop the tests when navigating away from a spec.

tgriesser:
@mapsandapps so this PR fixes the ability to open the absolute path in the IDE, if it does exist, since it fixes the part of your report with `see that the 'absoluteFile' getting passed in duplicates part of the path`.

The issue here isn't a user facing bug in Cypress as much as a side-effect of how we're testing w/ cypress-in-cypress, specifically that we're changing the current working directory during the test execution. The `after` isn't called when we're navigating away from a spec, which is a separate issue/will be the actual fix here. Logged UNIFY-1521 to take care of that.

Would like to get this merged since it helps with the quality-of-life of being able to click into a spec, even if it appears as an absolute path, and then the issue of test cleanup separately, if that sounds good.

chrisbreiding (approved):

mapsandapps (approved):

PR #20861, opened 2022-03-31
feat: add new webpack-dev-server package based on object API & bundling

tgriesser:
- Closes UNIFY-1267

### User facing changelog
Adds support for the new object syntax for the `webpack-dev-server`. Adds a new `webpack-dev-server-fresh` package to manage this support, it will replace the existing `webpack-dev-server` once all frameworks are supported.

Ported unit tests from `webpack-dev-server`, and added additional unit tests covering new functionality. Adds "open mode" tests for the matrix of supported webpack / webpack-dev-server versions, including recovering from failures, and adds system-tests.

Adds the following **system test improvements**:

- Adds the concept of `project-fixtures` for better testing the same project against multiple dependency versions [documentation added here](https://github.com/cypress-io/cypress/blob/7fd7a6aa2e85776e81d10da81d1b966c121dc7fb/system-tests/README.md#reusable-project-fixtures)
- For open-mode system tests, automatically tries updating the `yarn.lock` in the project directory if the `--frozen-lockfile` install fails, and then `rm node_modules` in the project directory, and retry once, so you just have to bump the `package.json` version and it'll work without needing to do all that manually: [link](https://github.com/cypress-io/cypress/blob/7fd7a6aa2e85776e81d10da81d1b966c121dc7fb/packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts#L124-L135) (will update to use `fs.remove`)

Also marked the `webpack-dev-server-fresh` `package.json` with `internal: true` for now to ensure it doesn't accidentally get published. 

### PR Tasks

- [x] N/A Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [X] N/A Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

### DIFF
--- a/packages/driver/cypress/e2e/cypress/stack_utils.cy.js
+++ b/packages/driver/cypress/e2e/cypress/stack_utils.cy.js
@@ -262,6 +262,54 @@ Error: spec iframe stack
       ])
     })
 
+    it('strips webpack protocol and maintains absolute path', () => {
+      $sourceMapUtils.getSourcePosition.returns({
+        file: 'cypress:////root/absolute/path/some_other_file.ts',
+        line: 2,
+        column: 1,
+      })
+
+      $sourceMapUtils.getSourcePosition.onCall(1).returns({
+        file: 'webpack:////root/absolute/path/cypress/integration/features/source_map_spec.coffee',
+        line: 4,
+        column: 3,
+      })
+
+      const sourceStack = $stackUtils.getSourceStack(generatedStack, projectRoot)
+
+      expect(sourceStack.sourceMapped).to.equal(`Error: spec iframe stack
+    at foo.bar (cypress:////root/absolute/path/some_other_file.ts:2:2)
+    at Context.<anonymous> (webpack:////root/absolute/path/cypress/integration/features/source_map_spec.coffee:4:4)\
+`)
+
+      expect(sourceStack.parsed).to.eql([
+        {
+          message: 'Error: spec iframe stack',
+          whitespace: '',
+        },
+        {
+          function: 'foo.bar',
+          fileUrl: 'http://localhost:1234/source_map_spec.js',
+          originalFile: 'cypress:////root/absolute/path/some_other_file.ts',
+          relativeFile: '/root/absolute/path/some_other_file.ts',
+          absoluteFile: '/root/absolute/path/some_other_file.ts',
+          line: 2,
+          column: 2,
+          whitespace: '    ',
+        },
+        {
+          function: 'Context.<anonymous>',
+          fileUrl: 'http://localhost:1234/tests?p=cypress/integration/features/source_map_spec.js',
+          originalFile: 'webpack:////root/absolute/path/cypress/integration/features/source_map_spec.coffee',
+          relativeFile: '/root/absolute/path/cypress/integration/features/source_map_spec.coffee',
+          absoluteFile: '/root/absolute/path/cypress/integration/features/source_map_spec.coffee',
+          line: 4,
+          column: 4,
+          whitespace: '    ',
+        },
+      ])
+    })
+
     it('returns empty object if there\'s no stack', () => {
       expect($stackUtils.getSourceStack()).to.eql({})
     })
@@ -363,6 +411,27 @@ Error: spec iframe stack
       expect(details.originalFile).to.equal('webpack:///cypress/integration/spec%with space &^$ emoji👍_你好.js')
       expect(details.relativeFile).to.equal('cypress/integration/spec%with space &^$ emoji👍_你好.js')
       expect(details.absoluteFile).to.equal(`${projectRoot}/cypress/integration/spec%with space &^$ emoji👍_你好.js`)
+    })
+
+    it('maintains absolute path when provided', () => {
+      cy.stub($sourceMapUtils, 'getSourcePosition').returns({
+        file: '/root/path/cypress/integration/spec.js',
+        line: 1,
+        column: 0,
+      })
+
+      // stack is fairly irrelevant in this test - testing transforming getSourcePosition response
+      const stack = stripIndent`
+        Error
+          at Object../cypress/integration/spec.js (http://localhost:50129/__cypress/tests?p=/root/path/cypress/integration/spec.js:99:1)
+      `
+
+      const projectRoot = '/Users/gleb/git/cypress-example-todomvc'
+      const details = $stackUtils.getSourceDetailsForFirstLine(stack, projectRoot)
+
+      expect(details.originalFile).to.equal('/root/path/cypress/integration/spec.js')
+      expect(details.relativeFile).to.equal('/root/path/cypress/integration/spec.js')
+      expect(details.absoluteFile).to.equal(`/root/path/cypress/integration/spec.js`)
     })
   })
 
--- a/packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
+++ b/packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts
@@ -107,6 +107,8 @@ async function makeE2ETasks () {
   let remoteGraphQLIntercept: RemoteGraphQLInterceptor | undefined
   let scaffoldedProjects = new Set<string>()
 
+  const cachedCwd = process.cwd()
+
   clearCtx()
   ctx = setCtx(makeDataContext({ mode: 'open', modeOptions: { cwd: process.cwd() } }))
 
@@ -148,8 +150,19 @@ async function makeE2ETasks () {
     async __internal__before () {
       Fixtures.remove()
       scaffoldedProjects = new Set()
+      process.chdir(cachedCwd)
 
       return { launchpadPort }
+    },
+
+    /**
+     * Force a reset to the correct CWD after all tests have completed, just incase this
+     * was modified by any code under test.
+     */
+    __internal__after () {
+      process.chdir(cachedCwd)
+
+      return null
     },
 
     /**
--- a/packages/frontend-shared/cypress/e2e/support/e2eSupport.ts
+++ b/packages/frontend-shared/cypress/e2e/support/e2eSupport.ts
@@ -175,6 +175,10 @@ beforeEach(() => {
   taskInternal('__internal__beforeEach', undefined)
 })
 
+after(() => {
+  taskInternal('__internal__after', undefined)
+})
+
 function scaffoldProject (projectName: ProjectFixture, options: { timeout?: number } = { timeout: SIXTY_SECONDS }) {
   return logInternal({ name: 'scaffoldProject', message: projectName }, () => {
     return taskInternal('__internal_scaffoldProject', projectName, options)


packages/config/src/options.ts
```diff
@@ -458,7 +465,7 @@ const runtimeOptions: Array<RuntimeConfigOption> = [
     canUpdateDuringTestTime: false,
   }, {
     name: 'morgan',
-    defaultValue: true,
+    defaultValue: false,
```
tgriesser, line 468: Will revert this

packages/driver/src/cypress/stack_utils.ts
```diff
@@ -12,7 +12,7 @@ import { getStackLines, replacedStack, stackWithoutMessage, splitStack, unsplitS
 
 const whitespaceRegex = /^(\s*)*/
 const stackLineRegex = /^\s*(at )?.*@?\(?.*\:\d+\:\d+\)?$/
-const customProtocolRegex = /^[^:\/]+:\/+/
+const customProtocolRegex = /^[^:\/]+:\/{1,3}/
```
tgriesser, line 15: Will open a separate PR for this

tgriesser: https://github.com/cypress-io/cypress/pull/20858

npm/webpack-dev-server-fresh/src/devServer.ts
```diff
@@ -0,0 +1,152 @@
...
+    switch (devServerConfig.framework) {
+      case 'create-react-app':
+        // frameworkConfig = createReactApp()
+        break
+      case 'react':
+        break
+      case 'nuxt':
+        break
+      default:
+        throw new Error(`Unexpected framework ${devServerConfig.framework}, expected one of ${ALL_FRAMEWORKS.join(', ')}`)
```
lmiller1990, line 133: Following my above comment - does this module really belong in `npm`? Or at least this part of it (the error handling, expecting some specific string arguments) - it doesn't seem like something we would not publish on npm, it's really opinionated and fairly useless outside of the binary.

Or is the separation more like:

- object API - internal only
- function API - external

It feels like these might be better separated as packages to avoid someone accidentally coupling them together (eg, if one is meant for third parties to build on, and one is meant for internal use, it seems like they should be in separate packages, one for npm publishing, one for bundling w/ the binary).

Could be missing something, a final synchronous code review after the initial comments have been addressed might be good.

npm/webpack-dev-server-fresh/src/helpers/sourceRelativeWebpackModules.ts
```diff
@@ -0,0 +1,213 @@
...
+const originalModuleLoad = (Module as ModuleClass)._load
+const originalModuleResolveFilename = (Module as ModuleClass)._resolveFilename
+
+/**
+ * Based on the current project config, we look for the closest webpack,
+ * webpack-dev-server, and html-webpack-plugin for a user's project
+ *
+ * @internal
+ */
+export function sourceRelativeWebpackModules (config: WebpackDevServerConfig) {
```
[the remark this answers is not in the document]
lmiller1990: We do something similar in the launchpad when we do the initial setup and find your deps, [see here](https://github.com/cypress-io/cypress/blob/10.0-release/packages/data-context/src/sources/WizardDataSource.ts#L91), using [resolve-from](https://www.npmjs.com/package/resolve-from).

Would `resolve-from` be a candidate for finding the nearest module here, or is this a bit too custom/different?

npm/webpack-dev-server-fresh/src/helpers/sourceRelativeWebpackModules.ts
```diff
@@ -0,0 +1,213 @@
...
+  htmlWebpackPlugin: {
+    importPath: string
+    module: unknown
+    packageJson: PackageJson
+    majorVersion: 4 | 5
+  }
+}
+
+const originalModuleLoad = (Module as ModuleClass)._load
+const originalModuleResolveFilename = (Module as ModuleClass)._resolveFilename
```
lmiller1990, line 53: Are we also going to bundle html-webpack-plugin v4 and v5, like we will do with `webpack-dev-server`? (are we still doing that? that's my understanding).

tgriesser: Yes, this is done in the `package.json`:

```
"dependencies": {
  "html-webpack-plugin-4": "npm:html-webpack-plugin@^4",
  "html-webpack-plugin-5": "npm:html-webpack-plugin@^5",
```

npm/webpack-dev-server/package.json
```diff
@@ -26,7 +26,7 @@
     "speed-measure-webpack-plugin": "1.4.2",
     "typescript": "^4.2.3",
     "webpack": "^4.44.2",
-    "webpack-dev-server": "^4.0.0"
+    "webpack-dev-server": "3.11.0"
```
ZachJW34, line 29: Doesn't matter too much but why did we need this change?

system-tests/project-fixtures/react/cypress.config.ts
```diff
@@ -0,0 +1,26 @@
+import { defineConfig } from 'cypress'
```
ZachJW34, line 1: Can we make this as minimal as we can? This is a javascript project so we could go with a js config and drop the types.

npm/webpack-dev-server-fresh/src/devServer.ts
tgriesser: Yeah I think where we landed is that it's more that the object API construction is internal only, function API is external - the function API contains the ability to call with a `framework`. 

We don't have to publish it independently to start out because the aim is to have most folks using the Object API directly, but when we do, it should be possible to import this & utilize similar to how we're using internally.

npm/webpack-dev-server-fresh/src/helpers/sourceRelativeWebpackModules.ts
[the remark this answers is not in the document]
tgriesser: It looks like `resolve-from` is [providing behavior](https://github.com/sindresorhus/resolve-from/blob/4f02820afe9b4e9c379f4bb2745debdb49087d4f/index.js#L29-L33) that is now provided in Node's `require.resolve` with the `paths` option: https://github.com/sindresorhus/resolve-from/issues/8 since v8.7

I think most other places in the codebase where we do scoped requires, we utilize this pattern, rather than depending on a package, so I'd lead toward explicitness here.

npm/webpack-dev-server/package.json
tgriesser: I'm not sure, Might have been a result of me debugging some things, let me revert and see if anything breaks.

npm/webpack-dev-server-fresh/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Testing
+
+Unit tests can be run with `npm test`. Integration tests can be run with `yarn cypress:run`
+
+This module should be primarily covered by system-tests / open-mode tests. All system-tests directories should be created using the notation:
+
+`webpack${major}_wds${devServerMajor}-$framework{-$variant}`
+
+- webpack4_wds3-react
```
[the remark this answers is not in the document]
tgriesser: Yeah I had this to distinguish between the webpack variants & frameworks. Don't mind if we do trailing `_`

npm/webpack-dev-server-fresh/src/CypressCTWebpackPlugin.ts
```diff
@@ -0,0 +1,159 @@
+import type { Compiler, Compilation } from 'webpack'
```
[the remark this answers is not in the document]
tgriesser: I'm actually not sure if this will track properly, since the files are being added/removed in different commits. But honestly now that we have this, working & tested, I'm wondering if it'd be much work to just replace the `webpack-dev-server` with the new implementation & provide a shim for the existing API use. We can discuss when we meet later.

npm/webpack-dev-server-fresh/src/helpers/sourceRelativeWebpackModules.ts
```diff
@@ -0,0 +1,213 @@
...
+      paths: [searchRoot],
+    })
+  } catch (e) {
+    if ((e as {code?: string}).code !== 'MODULE_NOT_FOUND') {
+      throw e
+    }
+
+    webpackJsonPath = require.resolve('webpack/package.json', {
+      paths: [
+        require.resolve('@cypress/webpack-batteries-included-preprocessor', {
```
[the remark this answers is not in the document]
tgriesser: > It should be deduped and would remove a link between these packages that could cause issues down the road.

The reason I didn't do that is that we don't have good machinery right now for ensuring that these aren't deduped, and  if we publish this independently we don't want it to come with an extra webpack when it should be consuming as a peer dependency. This is just being explicit about where we know/want that peer-dependency to be sourced from. 

I'm not too concerned issues down the road with this one, if they do come up, we have tests and we can make a decision based on the issue that arises, rather trying to predict a hypothetical.

npm/webpack-dev-server-fresh/src/devServer.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  let frameworkConfig: object | undefined
+
+  // If we have a framework specified, source the associated config
+  if (typeof devServerConfig.framework === 'string') {
+    switch (devServerConfig.framework) {
+      case 'create-react-app':
+        // frameworkConfig = createReactApp()
+        break
+      case 'react':
```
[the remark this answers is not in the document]
tgriesser: Yeah I was figuring this would be part of the logic for frameworks.

system-tests/project-fixtures/react/cypress.config.ts
tgriesser: I had been using this as a place to try out the `DefineDevServerConfig`, also since users can define either it shouldn't make a difference. Was also thinking we might want to add `ts` / `tsx` files into this package at some point.

npm/webpack-dev-server-fresh/src/CypressCTWebpackPlugin.ts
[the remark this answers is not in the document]
tgriesser: Nevermind, in talking w/ Zach think it should probably track fine.

npm/webpack-dev-server/package.json
ZachJW34: Stuff broke 😆

ZachJW34 (approved):
Tested it and it worked great!

system-tests/lib/fixtures.ts
```diff
@@ -49,12 +53,32 @@ export function scaffold () {
...
+    const packageJson = require(`${to}/package.json`)
+    const fixtureDir = packageJson.projectFixtureDirectory
+
+    if (fixtureDir) {
+      if (!projectFixtureDirs.includes(fixtureDir)) {
+        throw new Error(`Invalid project fixture directory: ${fixtureDir}, expected one of ${projectFixtureDirs}`)
+      }
+
+      await fs.copy(_path.join(projectFixtures, fixtureDir), to)
+    }
```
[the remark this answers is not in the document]
tbiethman: Fixtures will be an awesome feature for our system tests 💯

system-tests/lib/fixtures.ts
```diff
@@ -49,12 +53,32 @@ export function scaffold () {
...
+  try {
+    const packageJson = require(`${to}/package.json`)
+    const fixtureDir = packageJson.projectFixtureDirectory
+
+    if (fixtureDir) {
+      if (!projectFixtureDirs.includes(fixtureDir)) {
+        throw new Error(`Invalid project fixture directory: ${fixtureDir}, expected one of ${projectFixtureDirs}`)
+      }
+
+      await fs.copy(_path.join(projectFixtures, fixtureDir), to)
```
tbiethman, line 73: Does this overwrite or merge with existing directories in the project dir? Merging might help keep the fixtures from similarly ballooning in number by allowing additive tweaks at the project level. Or it might just make code tracing more complicated 🤷‍♂️

tgriesser: No, it just writes the files in the chosen `project-fixtures` within a given individual directory after scaffolding. The intent is that the project we're merging into will only have a `package.json` / `yarn.lock` file, and we're copying the fixture project we want to test against different dependency versions into it.

See the blurb about it here: https://github.com/cypress-io/cypress/blob/7fd7a6aa2e85776e81d10da81d1b966c121dc7fb/system-tests/README.md#reusable-project-fixtures

tgriesser: That is to say, the main intent, at least to start, is that the additive tweaks are limited to different package versions. In the future we can get fancier with it as the use cases arise.

tbiethman (dismissed):

npm/webpack-dev-server-fresh/src/helpers/sourceRelativeWebpackModules.ts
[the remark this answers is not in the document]
lmiller1990: Oh. Maybe we should not bother with `resolve-from` and just use `require.resolve` in the other places, too.

lmiller1990 (approved):
I was not able to attend the sync w/ @tgriesser and @ZachJW34 but seems like everything is :ok_hand: , none of my comments are blockers.

Let's get this :heavy_check_mark: and resolve conflicts and merged :100:

ZachJW34 (approved):
Once were green LGTM!

ZachJW34 (dismissed):

tbiethman (approved):

ZachJW34 (approved):

PR #20869, opened 2022-03-31
fix(unify): Cypress version link should point to Cypress doc's changelog

davidmunechika:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1372?atlOrigin=eyJpIjoiZGMwNjc4ZGZjZjNmNGVjNzg5NDNjM2EwM2IzYTViNTIiLCJwIjoiaiJ9

### User facing changelog
The Cypress version link should point to the Cypress doc's changelog instead of the Git tag release notes.

### Additional details

image

### PR Tasks


packages/frontend-shared/src/gql-components/topnav/TopNav.vue
```diff
@@ -22,7 +22,7 @@
     >
       <div class="font-medium">
         <ExternalLink
-          :href="`${releasesUrl}/tag/v${versions.latest.version}`"
+          :href="`https://on.cypress.io/changelog#${versions.latest.version.replaceAll('.', '-')}`"
```
marktnoonan, line 25: We could make a function in the script block to compose this string, taking `latest` or `current` as an argument, especially since we're doing a bit more work in there now.

packages/frontend-shared/src/gql-components/topnav/TopNav.vue
```diff
@@ -97,7 +97,7 @@
       class="h-16px w-16px group-hocus:icon-dark-indigo-500 group-hocus:icon-light-indigo-50 icon-dark-gray-500 icon-light-gray-100"
     />
     <span>
-      v{{ versions.latest.version }}
+      v{{ versions.current.version }}
```
marktnoonan, line 100: Switching to `current` does feel slightly safer, even though this is only shown if we are not running an old version - so `current` and `latest` should be the same.

marktnoonan (approved):
Looks good. Couple small comments but nothing blocking.

packages/frontend-shared/src/gql-components/topnav/TopNav.vue
davidmunechika: good point! should be updated now

packages/frontend-shared/src/gql-components/topnav/TopNav.vue
```diff
@@ -352,4 +350,8 @@ onKeyStroke(['Enter', ' ', 'Escape'], (event) => {
   resetPrompt(event)
 })
 
+const changelogLink = (version) => {
+  return `https://on.cypress.io/changelog#${version.replaceAll('.', '-')}`
```
emilyrohrbough, line 354: 👍🏻 Nice changes all and all. Only question -- if we are in a pre-release binary version, does this cause any issues? Do we care?

emilyrohrbough (approved):

PR #20870, opened 2022-03-31
refactor: lift indexHtmlFile up to component, add validation

ZachJW34:
- Closes [UNIFY-1251](https://cypress-io.atlassian.net/browse/UNIFY-1251)

### User facing changelog
Lift `indexHtmlFile` up to `component` config and validate the property is only valid on the component config.

### Additional details
The PR is broken up into two commits, the first contains the logic for lifting the property and adding validation and the second adds the missing `component-index.html` file to projects that were relying on the default (hopefully makes it easier to review). There is no longer a fallback template for `@cypress/webpack-dev-server` and `@cypress/vite-dev-server` since we are scaffolding this file to the users project at `cypress/support/component-index.html` which will be the default value of `component.indexHtmlFile`.

There is validation of the `indexHtmlFile` property to ensure that it is not added at the root level of the config or in the e2e config. All devServers have been updated to pull this value from the cypress config rather than passing the option explicitly.

I did not implement the `getTemplate` API outlined in the issue as that is an implementation detail that should be implemented after the [Object API](https://github.com/cypress-io/cypress/pull/20861) is merged, I will update the ticket and create another one to implement it. We already have this logic inside of [WizardActions](https://github.com/cypress-io/cypress/blob/7636287eee6b8d89031f6d5f2904ba2c591ce9e3/packages/data-context/src/actions/WizardActions.ts#L286), it's just not ideal but we can move forward without it.

The removal of the `template` property from `@cypress/webpack-dev-server` is a breaking change but we already plan on making breaking changes to the dev-servers. These changes won't be released until 10.0 and there is an [issue](https://cypress-io.atlassian.net/browse/UNIFY-1255) for ensuring 9.x users don't use the 10.0 supported version. This property was mostly an internal property and not documented.

User no longer needs to specify `indexHtmlFile` in their config as `config.component.indexHtmlFile` will have a default value of `cypress/support/component-index.html` which is scaffolded for them during project setup.

Before:
```
module.exports = {
  component: {
  	devServer,
	devServerConfig: {
	  indexHtmlFile: 'cypress/support/component-index.html',
	},
  },
} 
```

After:
```
module.exports = {
  component: {
  	devServer,
  },
}
```

### PR Tasks

- [NA] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [NA] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? [Notion](https://www.notion.so/cypressdx/Changes-for-10-0-0-1dca213837b84d6ea297c2d12c615f3a#8d191a4f7694412f95d50049bc476fcc) is up to date
- [NA] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?


cli/types/cypress.d.ts
```diff
@@ -2858,7 +2858,7 @@ declare namespace Cypress {
      * Override default config options for E2E Testing runner.
      * @default {}
      */
-    e2e: CoreConfigOptions
+    e2e: Omit<CoreConfigOptions, 'indexHtmlFile'>
```
tgriesser, line 2861: Is there a reason we want this on `CoreConfigOptions` rather than `ComponentConfigOptions`?

tgriesser (changes requested):

cli/types/cypress.d.ts
tgriesser: I think for the `devServer` we're generally going to be accessing the properties off of the un-merged config anyway, because the merging happens in the parent process and we can't serialize/unserialize a user's webpackConfig over the wire (ran into that when refactoring the object API)

> The options in ComponentConfigOptions aren't typed at the root level,

Might not be sure what you're referring to - I was just saying to add the type to this interface:

```
interface ComponentConfigOptions<ComponentDevServerOpts = any> extends Omit<CoreConfigOptions, 'baseUrl'> {
    devServer: DevServerFn<ComponentDevServerOpts>
}
```

And then use it from that to pass into `startServer` directly

lmiller1990 (approved):
@ZachJW34 nice - this all seems pretty straight forward, since `getTemplateHtml` isn't implemented, and we are just lifting the `indexHtmlFile` up

If we are going to do the `getTemplateHtml` changes separately, I think we should:

1. Update the Jira ticket to reflect the actual work done here (right now based on Jira, it looks like everything should be here, which it's not)
2. Create a new ticket with the remaining work to be done

What do you think?

tgriesser (approved):

ZachJW34:
@lmiller1990 I'll update the issue now, was just going to wait a bit to see how the PR shakes out.

PR #20872, opened 2022-03-31
chore: [Multi-domain] Renaming variable and functions to reduce confusion between usage of origin, origin policy and domain Part 2

mjhenkes:
- Closes 

### User facing changelog

n/a

### Additional details

This PR focuses on `communicator.ts`. Other files are included as well but the changes there are more ancillary than comprehensive. There will be more changes coming in future PRs.

### PR Tasks

### DIFF
--- a/packages/driver/src/cy/commands/navigation.ts
+++ b/packages/driver/src/cy/commands/navigation.ts
@@ -50,24 +50,24 @@ const isValidVisitMethod = (method) => {
 const timedOutWaitingForPageLoad = (ms, log) => {
   debug('timedOutWaitingForPageLoad')
 
-  const anticipatedCrossOriginHref = cy.state('anticipatingMultiDomain')
+  const anticipatedCrossOriginHref = cy.state('anticipatingCrossOriginResponse')?.href
 
   // Were we anticipating a cross domain page when we timed out?
   if (anticipatedCrossOriginHref) {
     // We remain in an anticipating state until either a load even happens or a timeout.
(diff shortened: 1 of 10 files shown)


packages/driver/src/cy/commands/navigation.ts
```diff
@@ -588,7 +588,7 @@ export default (Commands, Cypress, cy, state, config) => {
     } catch (e) {} // eslint-disable-line no-empty
   })
 
-  Cypress.multiDomainCommunicator.on('visit:url', ({ url }) => {
+  Cypress.multiOriginCommunicator.on('visit:url', ({ url }) => {
```
mschile, line 591: Would `crossOriginCommunicator` be more accurate or maybe we call this one `primaryOriginCommunicator` to go along with `specBridgeCommunicator`?

mschile (changes requested):

packages/driver/src/cy/commands/navigation.ts
mjhenkes: I started with crossOriginCommunicator but walked it back because that makes it sound like it's only used for cross origin when we also use it for intra origin communcation too.

I like primaryOriginCommunicator but that kicks the can down the road for a name when we merge the two.

mjhenkes: you convinced me to go with primaryOriginCommunicator

mschile (approved):

AtofStryker (approved):
6bchjy

chrisbreiding (approved):

PR #20873, opened 2022-03-31
test: UNIFY-1310 add tests to runner 

rachelruderman:
- Closes UNIFY-1310

### User facing changelog
n/a

### Additional details
1. Adds test to verify that  text changes when the Request Access button is clicked
1. Adds test to verify `projectId` is inserted into config file when a new project is created
1. Removes try/catch that was silencing an error message
 
n/a

### PR Tasks


packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
```diff
@@ -585,11 +585,7 @@ export const mutation = mutationType({
         projectId: nonNull(stringArg()),
       },
       resolve: async (_, args, ctx) => {
-        try {
-          await ctx.actions.project.setProjectIdInConfigFile(args.projectId)
-        } catch {
-          // ignore error as not useful for end user to see
```
rachelruderman, line 591: 🤣 😭 this ignored error was painful for me to track down. I think there's value in having it:

Even if it's not immediately useful for the end user to see, it becomes very useful if/when the end user reports a bug. Without this error, the "Create project" button just fails silently.

packages/app/src/runs/modals/SelectCloudProjectModal.vue
```diff
@@ -269,7 +269,9 @@ const setProjectIdMutation = useMutation(SelectCloudProjectModal_SetProjectIdDoc
 async function createOrConnectProject () {
   let projectId: string | undefined
 
-  if (newProject.value && pickedOrganization.value) {
+  const isNewProject = Boolean(newProject.value && pickedOrganization.value)
+
+  if (isNewProject) {
```
rachelruderman, line 274: Just threw in an intermediate variable for clarity

packages/app/cypress/e2e/runs.cy.ts
```diff
@@ -171,6 +171,44 @@ describe('App: Runs', { viewportWidth: 1200 }, () => {
...
+      cy.scaffoldProject('launchpad')
+      cy.openProject('launchpad')
+      cy.startAppServer('e2e')
+      cy.loginUser()
+      cy.visitApp()
+
+      cy.withCtx(async (ctx) => {
+        const config = await ctx.project.getConfig()
+
+        expect(config.projectId).to.not.equal('newProjectId')
```
emilyrohrbough, line 196: would this value be undefined?

packages/app/cypress/e2e/runs.cy.ts
```diff
@@ -171,6 +171,44 @@ describe('App: Runs', { viewportWidth: 1200 }, () => {
     })
   })
 
+  context('Runs - Create Project', () => {
```
emilyrohrbough, line 174: Not specific to your changes, but with name of this test file I wasn't expecting these tests to be specific to the dashboard.

packages/app/cypress/e2e/runs.cy.ts
rachelruderman: At this point it's `null`. I checked `to.not.equal('newProjectId')` instead of `to.equal(null)` so that if at some poit in the future a projectId is added to the scaffolded project, this test won't break 😃 happy to change if you prefer

packages/app/cypress/e2e/runs.cy.ts
rachelruderman: I see that some of the specs in this directory have the `cypress-in-cypress` prefix, i.e. `cypress-in-cypress-run-mode.cy.ts`. Should I update to something like `cypress-in-cypress-open-mode-runs.cy.ts`?

marktnoonan: We should probably retired the "cypress-in-cypress" convention, it refers to solving a specific problem, which was getting the spec runner to work in Cypress tests.

I do think "runs" is an OK name, since the the app contexts, runs and the "runs page" are all about test runs recorded to the dashboard. But `runs_page` is probably clearer file name if we want one, to be more clear that this is checking the display of runs info from the cloud, not anything at all about the behavior of the runner.

I even got confused writing that so we probably do need a better name.

Not that there are getting to be more files in here, it might be a good time to step back and organize this folder with more of pattern... some files test pages, some test multi-page features like the top-nav, and the names of the cypress-in-cypress tests are quite unclear.

emilyrohrbough: > I do think "runs" is an OK name, since the the app contexts, runs and the "runs page" are all about test runs recorded to the dashboard.

Given I haven't worked in the dashbaord, I alway associate runs to a spec run or test run within a spec. Def an overloaded term.!

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
```diff
@@ -585,11 +585,7 @@ export const mutation = mutationType({
...
-        }
+        await ctx.actions.project.setProjectIdInConfigFile(args.projectId)
```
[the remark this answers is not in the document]
emilyrohrbough: hmmm is a way to catch this and provide an useful error message to the end-user so the error isn't hidden while giving the user information on where to find help or how to solve it?

packages/app/cypress/e2e/runs.cy.ts
emilyrohrbough: > so that if at some point in the future a projectId is added to the scaffolded project

I'm not as familiar with the dashboard space, but if a new scaffolded project had a projectId, are users able to create a new project & override it from the UI?

packages/app/cypress/e2e/runs.cy.ts
```diff
@@ -280,6 +318,30 @@ describe('App: Runs', { viewportWidth: 1200 }, () => {
...
+            hasRequestedAccess: true,
+          }
+        }
+
+        return obj.result
+      })
+
+      cy.visitApp('/runs')
+      cy.findByText(defaultMessages.runs.errors.unauthorized.button).click()
+      cy.findByText(defaultMessages.runs.errors.unauthorizedRequested.button).should('exist')
```
emilyrohrbough, line 343: Just curious, this the context of this test, but is the following describe block of `'Runs - Unauthorized Project Requested'` testing something different compared to the current describe block of `Runs - Unauthorized Project`? 
 Wondering if we just need to enhance one of these?

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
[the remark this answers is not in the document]
marktnoonan: We end up showing two different errors different places, one of which is the launchpad, which does not seem ideal. @tgriesser is this expected? I only found this error when I went looking for the little blue toast to restart Cypress.

Screen Shot 2022-04-05 at 4 19 01 PM

Screen Shot 2022-04-05 at 4 17 06 PM

marktnoonan (approved):
Provokes a lot of thoughts about the code _around_ the PR, but the changes look good and it adds the test specified in the ticket. Happy to approve again if you make more changes.

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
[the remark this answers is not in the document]
tgriesser: Yeah, we probably need to sync on this and figure out how to make this actionable as part of the response of the mutation, rather than having this be unhandled and displayed like this.

Any error in a mutation should be caught/handled, otherwise it shows up as the "An unexpected internal error occurred" as above, which we don't really want. Maybe we want to log the error out to the console or something? 

I think the way the error case handled is a bit hard to follow in this flow and could be improved, it's almost like we want the return type of the mutation to be a union of `Query | ErrorWrapper`

packages/app/cypress/e2e/runs.cy.ts
rachelruderman: > if a new scaffolded project had a projectId, are users able to create a new project & override it from the UI
I believe so

packages/app/cypress/e2e/runs.cy.ts
rachelruderman: AFAIK in `'Runs - Unauthorized Project Requested'`, we're validating the correct UI appears for users who have previously requested access, whereas in `Runs - Unauthorized Project`, we're testing the request access flow itself

PR #20876, opened 2022-03-31
fix: remove create spec from story

elevatebart:
- Closes UNIFY-1412

### User facing changelog

Remove the disabled create button from story

#### Before

Screen Shot 2022-03-31 at 5 14 11 PM

Screen Shot 2022-03-31 at 5 15 25 PM

#### After

Screen Shot 2022-03-31 at 5 04 59 PM
Screen Shot 2022-03-31 at 5 05 31 PM

### PR Tasks


ZachJW34 (commented):
Looks like there are a few more e2e tests to fix up that are relying on Storybook codegen.

I remember some convos around wanting to disable the Storybook codegen but keep a "Coming Soon" message. Is that no longer a requirement?

Also, we have a lot of Storybook codegen stuff on the backend, should we create an issue for removing that? It's doing no harm but the removal on the frontend vs disabling has me thinking we should get the backend code cleaned up as well.

Looks good otherwise, I'll approve once CI is green!

ZachJW34 (dismissed):
Couple more files that can be remove:
* packages/data-context/test/unit/codegen/files/react/Button.stories.jsx
* packages/data-context/test/unit/codegen/files/vue/Button.stories.ts

Going to approve, code changes look good 👍

packages/app/cypress/e2e/specs.cy.ts
```diff
@@ -455,177 +455,6 @@ describe('App: Index', () => {
     viewportHeight: 768,
     viewportWidth: 1024,
   }, () => {
-    context('project with storybook', () => {
```
emilyrohrbough, line 458: Arn't we adding story spec generation in a future release? IF so, should we skip these tests with a link tot he followup work and just update the UI to remove this spec option?

emilyrohrbough: If not a soon followup, remove this code LGTM.

tbiethman (dismissed):

lmiller1990 (commented):
@elevatebart I second @emilyrohrbough - should we be deleting this feature, or just removing the UI stuff and skipping the tests? If we will bring it back soon, it's much easier to leave it here than have to go back through git history.

We could even feature flag it, like `CYPRESS_INTERNAL_STORYBOOK`, so we could continue running the tests, etc. I'm not sure we should remove all this code that works, and we know we want to add back soon.

elevatebart:
If it works why are we removing it?

jennifer-shehane:
I would always prefer removing dead code. We always have the git commits to refer to from the PR of the removal of code. Create a Jira story to reimplement the feature, link to the PR that removed it, it won't get lost from there.

emilyrohrbough (approved):

ZachJW34 (approved):

Issue #20745, opened 2022-03-23
pause() should not be ignored with cypress run --headed --no-exit


totomaze:
### Current behavior

With version 9.5.2 pause() is ignored even with the 2 flags --headed --no-exit.

I downgraded to 8.6.0 and this feature works fine. A regression must have occured

### Desired behavior

pause() should not be ignored with cypress run --headed --no-exit

I would suggest to make it work without the --no-exit flag since this flag will block all the following tests to run even if the window is close.

### Test code to reproduce

Add a cy.pause() in a test case and run it with cypress run --headed --no-exit (the pause() will be ignored)

If you run the test with cypress open it will work

### Cypress Version

from 9.2.0 to 9.5.2 pause() is ignored

### Other

The last version where it s working is 9.1.1, after this version using --headed --no-exit no longer pause during a test run (cypress run --headed --no-exit)

I believe it s caused by this update in 9.2.0:
Cypress will throw an error when a user attempts to update a configuration
value at run-time using either the Test Configuration or using
Cypress.config() that is a
[readonly option](https://docs.cypress.io/guides/references/configuration#Test-Configuration).
Addresses https://github.com/cypress-io/cypress/issues/6407 and
https://github.com/cypress-io/cypress/issues/19001.

Is there any workaround to get the test to pause when running it by cypress run ?

Here is the original pull request that allow a test run to pause when started by cypress run:
https://github.com/cypress-io/cypress/pull/18358

emilyrohrbough:
@totomaze This indeed does appear to be a regression. Is this something you'd be willing to make a contribution to fix?

totomaze:
> @totomaze This indeed does appear to be a regression. Is this something you'd be willing to make a contribution to fix?

If I find some free time I might try to find a solution , but for now I found a workaround that suits me and work for 9.5.2 :

```
if (Cypress.config('browser').isHeaded){
  window.top.__cySkipValidateConfig = true
  Cypress.config('isInteractive',true)
}
```


Another solution will be to allow the following to work
cypress run -s test --config isInteractive=true

If someone have the dev environment maybe try to remove the config('exit') to see if it works :
packages/driver/src/cy/commands/debugging.ts Line 50
```
pause (subject, options: any = {}) {
      // bail if we're in run mode, unless --headed and --no-exit flags are passed
      if (!config('isInteractive') && (!config('browser').isHeaded || config('exit'))) {
        return subject
      }
```

PR #18896, opened 2021-11-12
fix: throw when writing to 'read only' properties of `config`

davidmunechika:
- Closes https://github.com/cypress-io/cypress/issues/6407
- Closes https://github.com/cypress-io/cypress/issues/19001

### User facing changelog
Cypress will throw an error when a user attempts to set these 'non-writable' config values using `Cypress.config`.

### Additional details
The following were identified as properties that could be updated during test time:
- animationDistanceThreshold
- baseUrl
- blockHosts
- defaultCommandTimeout
- env
- execTimeout
- experimentalSessionSupport
- includeShadowDom
- keystrokeDelay
- numTestsKeptInMemory
- pageLoadTimeout
- redirectionLimit
- requestTimeout
- responseTimeout
- retries
- screenshotOnRunFailure
- scrollBehavior
- slowTestThreshold
- taskTimeout
- viewportHeight
- viewportWidth
- waitForAnimations

The following were identified as read-only properties for per test config:
- autoOpen (internal)
- browsers
- chromeWebSecurity
- clientCertificates
- clientRoute (internal)
- component
- componentFolder
- configFile (internal)
- devServerPublicPathRoute (internal)
- downloadsFolder
- e2e
- exit
- experimentalFetchPolyfill
- experimentalInteractiveRunEvents
- experimentalSourceRewriting
- experimentalStudio
- fileServerFolder
- fixturesFolder
- hosts
- ignoreTestFiles
- integrationFolder
- isInteractive
- isTextTerminal
- modifyObstructiveCode
- morgan (internal)
- namespace (internal)
- nodeVersion
- pluginsFile
- port
- projectId
- reporter
- reporterOptions
- reporterRoute (internal)
- resolvedNodePath
- resolvedNodeVersion
- socketId (internal)
- socketIoCookie (internal)
- socketIoRoute (internal)
- supportFile
- testFiles
- trashAssetsBeforeRuns
- userAgent
- video
- videoCompression
- videoUploadOnPasses
- videosFolder
- watchForFileChanges
- xhrRoute (internal)

## User experience

Trying to set read-only config in test configuration:

Screenshot 2021-11-23 at 8 57 09 AM

Trying to set read-only config within test with Cypress.config():

Screenshot 2021-11-23 at 8 57 43 AM

### PR Tasks

- [x] Have tests been added/updated? (Added config package unit tests, need to add integration test still)
- [x] Has the original issue or this PR been tagged with a release in ZenHub? 
[...]

PR #18358, opened 2021-10-05
fix(driver): `cy.pause()` should not be ignored with `cypress run --headed --no-exit`

davidmunechika:
- Closes https://github.com/cypress-io/cypress/issues/4044

### User facing changelog
Users will be able to use `cy.pause()` and observe that a test paused when using `cypress run --headed --no-exit`.

### Additional details
Previous behavior: `cy.pause()` was ignored when using `cypress run` even if `--headed` and `--no-exit` flags were passed.

New behavior: `cy.pause()` works as expected (i.e. the same as with `cypress open`) when `--headed` and `--no-exit` flags are passed to `cypress run`.

### PR Tasks

- [x] Have tests been added/updated?
- [ ] Has the original issue or this PR been tagged with a release in ZenHub? 
- [ ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [ ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [ ] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

PR #20877, opened 2022-04-01
fix(regression): cy.pause() should not be ignored with `cypress run --headed --no-exit`

davidmunechika:
- Closes https://github.com/cypress-io/cypress/issues/20745

### User facing changelog
Fixes a regression that was introduced by [this PR](https://github.com/cypress-io/cypress/pull/18896) which prevented cy.pause() from correctly executing when run with `cypress run --headed --no-exit`.

### Additional details
Original implementation PR: https://github.com/cypress-io/cypress/pull/18358

The cause of this regression was the addition of `exit` to `ResolvedConfigOptions` in [this PR](https://github.com/cypress-io/cypress/pull/18896). As a result, when the args got merged with the config, `config('exit')` would always be `true` even when `--no-exit` was passed. When this change was made, we should have removed [this line](https://github.com/cypress-io/cypress/commit/1d08280f4a85fc9eaf9326c88c2ea0e6d9d099d7#diff-7ec134b05d5f4b3f8dac0940baefe0ecd2b6bd331acf11ab4332cfe87fec8096R75)

However, it was decided that `exit` should be an arg that is not included as part of config. Therefore, this PR removes `exit` from the `ResolvedConfigOptions` and correctly passes `exit` through the options so no conflicts will occur with the config and `--no-exit` will still work as expected.

### PR Tasks


packages/config/lib/options.ts
```diff
@@ -140,11 +140,6 @@ const resolvedOptions: Array<ResolvedConfigOption> = [
     defaultValue: 60000,
     validation: validate.isNumber,
     canUpdateDuringTestTime: true,
-  }, {
-    name: 'exit',
-    defaultValue: true,
-    validation: validate.isBoolean,
-    canUpdateDuringTestTime: false,
```
lmiller1990, line 147: Why does removing this fix the issue? :thinking: 

Do you mind explaining a little more about this fixed the regression (not entirely clear what part of the regression PR linked broke this in the first place).

davidmunechika: @lmiller1990 I've updated this PR description with more details about the regression. The expected behavior is that Cypress will pause if it is in open mode _or_ if it in run mode with `--headed` and `--no-exit` both passed as args. Therefore, if cypress is in run mode aka not interactive (`!config('isInteractive')`) **and** either `--headed` or `--no-exit` was not passed, then we **do not** want to pause. Hence, the logic in the snippet above

lmiller1990: Right, I see! I hope we can have `config('mode')` in the future - `config('mode') === 'run` would be way more readable. Nice job hunting down this regression.

system-tests/test/pause_headed_exit_spec.ts
```diff
@@ -0,0 +1,13 @@
+import systemTests from '../lib/system-tests'
+
+describe('cy.pause() works with --headed and --no-exit', () => {
+  systemTests.setup()
+
+  systemTests.it('passes', {
+    spec: 'pause_simple_spec.js',
+    snapshot: false,
+    headed: true,
+    noExit: true,
```
lmiller1990, line 10: Somewhat surprised a test can have an exit code of 0 **and** have `noExit` as true - seems like these would conflict, but seems this isn't the case?

lmiller1990 (dismissed):
Seems fine - I had my comment clarified, also left one final question but happy to drop the :heavy_check_mark:  since the regression is fixed w/ a test.

system-tests/test/pause_headed_exit_spec.ts
lmiller1990: At least locally this seems to hang forever (on develop, was verifying it fails there)

image

davidmunechika: Yes this test is still a wip- it will hang forever because `--no-exit` has been passed

davidmunechika: Currently working on about adding a callback to ExecOptions (something like onSpawn) that yields the ChildProcess so we can listen to the stdout and kill the ChildProcess within the system test itself when `not exiting due to options.exit being false` gets logged.

lmiller1990: Gotcha, can you mark it as draft? I can re-review when you are happy with it - I thought this was ready for review.

image

system-tests/test/pause_headed_exit_spec.ts
```diff
@@ -0,0 +1,51 @@
+import systemTests from '../lib/system-tests'
+
+describe('cy.pause() works with --headed and --no-exit', () => {
```
emilyrohrbough, line 3:
```suggestion
describe('cy.pause() in run mode', () => {
```
Might be worth generalizing the describe & test file name given this testing these various permutations.

emilyrohrbough (approved):
🎉 Nice!! Glad you got this figured out. At this point are the existing driver tests around this behavior helpful and/or should those test cases be updated as well?

system-tests/lib/system-tests.ts
```diff
@@ -135,6 +135,10 @@ type ExecOptions = {
    * Pass a function to assert on and/or modify the stdout before snapshotting.
    */
   onStdout?: (stdout: string) => string | void
+  /**
+   * Pass a function to yield the ChildProcess
+   */
+  onSpawn?: (cp: any) => void
```
flotwig, line 141: types are good, to be generic to Docker processes:

```suggestion
   * Pass a function to receive the spawned process as an argument.
   */
  onSpawn?: (sp: SpawnerResult) => void
```

system-tests/projects/e2e/cypress/integration/no_pause_spec.js
```diff
@@ -0,0 +1,27 @@
...
+        Cypress.emit('run:end')
+        Cypress.emit('resume:all')
+      })
+
+      Cypress.emit('resume:next')
+      Cypress.emit('run:end')
+    })
+
+    cy.pause().wrap({}).should('deep.eq', {}).then(function () {
+      expect(didPause).to.be.false
```
flotwig, line 22: personally, I would combine `no_pause_spec` and `pause_spec` and switch on behavior with `Cypress.browser.isHeadless` or by passing `env` in, but since it's a test it's fine to just copy it too

system-tests/lib/system-tests.ts
```diff
@@ -265,6 +269,7 @@ export type SpawnerResult = {
   stderr: stream.Readable
   on(event: 'error', cb: (err: Error) => void): void
   on(event: 'exit', cb: (exitCode: number) => void): void
+  kill?: ChildProcess['kill']
```
flotwig, line 272: nit: i'm guessing this is optional here because `DockerProcess#kill` doesn't exist, and without this, type checking fails. i get it if you don't want to implement `DockerProcess#kill`, but IMO this would be slightly cleaner if you added a stub `kill` method to alert the dev the reason `DockerProcess#kill` isn't working. something like this:

```
kill() {
  throw new Error('.kill not implemented for DockerProcess.')
}
```

system-tests/test/pause_headed_exit_spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      })
+    },
+  })
+
+  systemTests.it('does not pause if headless', {
+    spec: 'no_pause_spec.js',
+    snapshot: true,
+    headed: false,
+    noExit: true,
+    expectedExitCode: null,
```
flotwig, line 26: what's the exit code? is it...

```suggestion
    expectedExitCode: 1,
```

...? we should get the same exit code every time, so might as well assert on it

flotwig (changes requested):
Nice work on the clean fix for this, and on the new test coverage for `--no-exit`. I was really surprised that we have no coverage considering it's part of our public API.

system-tests/test/pause_headed_exit_spec.ts
davidmunechika: @flotwig The exit code is always null because it kills the process before returning an exit code. Is there a way to get around this? I tried sending an exit code before/after killing the process but it continued to be null every time.

system-tests/projects/e2e/cypress/integration/no_pause_spec.js
davidmunechika: Wouldn't I need to factor in `noExit` as well to switch on behavior? Since if `--headed` was passed but `--no-exit` was not, we _do not_ want to pause?

system-tests/test/pause_headed_exit_spec.ts
flotwig: oh yeah, I forgot about this. Node.js actually passes 2 arguments to the `.on('exit')` handler: `exitCode`, AND `signal`. if the process exited from a signal, `exitCode` is null: https://stackoverflow.com/a/39169784/3474615

this is in contrast to how POSIX-compatible shells work, which assign an exit code > 128 if a process exited due to an unhandled signal: https://unix.stackexchange.com/a/99134/305482

so yeah, nothing to do here, this is correct :smile: sorry

system-tests/lib/docker.ts
```diff
@@ -104,6 +104,12 @@ class DockerProcess extends EventEmitter implements SpawnerResult {
       },
     )
   }
+
+  kill () {
+    log('.kill not implemented for DockerProcess.')
+
+    return false
```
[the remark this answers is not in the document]
davidmunechika: How do I fix the type error when doing this though because .kill should return a boolean, not void

flotwig (approved):

emilyrohrbough (approved):

PR #20884, opened 2022-04-01
fix: add index.mjs to the published files of cli

elevatebart:
This one is an oops.

We moved the entry point of cypress with ESM but forgot to publish the new file.

Here is to publishing `index.mjs` 🍻

Screen Shot 2022-04-01 at 9 30 05 AM


JessicaSachs (approved):
I'm very concerned that the mjs system-tests I wrote didn't catch this. Oh, actually, it makes sense that they didn't. The files existed locally.

Issue #18822, opened 2021-11-08
Internal: Fix flaky tests in circle CI


mjhenkes:
### Current behavior

We currently have several flaky tests that need to be retried several times to pass.

### Desired behavior

Cypress tests should not be flakey and offer 

### Test code to reproduce

Submit a PR to cypress and watch the circle ci build.

We can identify the flaky tests to focus on by using the cypress flak report: https://dashboard.cypress.io/projects/ypt4pf/analytics/flaky-tests

and circle CI's report: https://app.circleci.com/insights/github/cypress-io/cypress/workflows/linux/tests?branch=develop&reporting-window=last-30-days

### Cypress Version

n/a

mjhenkes:
settings spec flaky test fix: #18979

davidmunechika:
PR to fix flaky net stubbing, websocket, proxy logging, and Firefox system tests: https://github.com/cypress-io/cypress/pull/19159

ryanthemanuel:
PR for system-tests-firefox screenshot fix: #20260 
PR for server-performance-tests screenshot fix: #20271

rachelruderman:
Closing since we're taking care of flaky tests as part of Firewatch duties 🔥

PR #20382, opened 2022-02-28
chore: refactor log function to class

sainthkh:
- Related with [this comment](https://github.com/cypress-io/cypress/pull/20169#discussion_r813042620)

### User facing changelog

N/A. It's an internal refactoring

### Additional details
- Why was this change necessary? => Modernize the code.
- What is affected by this change? => N/A
- Any implementation details to explain? => Change the function/object to class.

### Summary of changes.

As it's a refactoring, the code change is hard to read. So, I left this.

* I organized the code into 3 parts: `LogUtils`, `LogManager`, `Log`. 
  * `LogUtils` are functions used in other files. They were exported with `export default`. I could remove them.
  * `LogManager`: Inside `create` function, `logs` variable exists and it manages logs created. But it makes us create `fireChangeEvent` function inside `create` function and attach it to `Log`. So, I created `LogManager` class and `Log` class reference it.
  * `Log`: It was a function that returns an object. I changed it to a class.
* I left some comments about other logic changes below.

### How has the user experience changed?

N/A

### PR Tasks

- [na] Have tests been added/updated?
- [na] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [na] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [na] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [na] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

PR #20886, opened 2022-04-01
fix: fix flaky report test

emilyrohrbough:
- Apart of #18822 

https://dashboard.cypress.io/projects/ypt4pf/runs/31638/overview/0296e264-b0b3-4ebc-997e-a0e473f8fdfa
Screen Shot 2022-04-01 at 11 34 27 AM
Screen Shot 2022-04-01 at 11 35 05 AM


tbiethman (dismissed):

davidmunechika (dismissed):
lgtm

packages/driver/src/cypress.ts
```diff
@@ -555,7 +555,9 @@ class $Cypress {
         return this.emit('log:added', ...args)
 
       case 'command:log:changed':
-        this.runner.addLog(args[0], this.config('isInteractive'))
+        // Cypress logs will only trigger an update every 4 seconds so there is a
```
tgriesser, line 558:
```suggestion
        // Cypress logs will only trigger an update every 4 milliseconds so there is a
```

packages/runner-shared/src/event-manager.js
```diff
@@ -431,11 +431,15 @@ export const eventManager = {
     })
 
     Cypress.on('log:changed', (log) => {
-      const displayProps = Cypress.runner.getDisplayPropsForLog(log)
+      // Cypress logs will only trigger an update every 4 seconds so there is a
```
tgriesser, line 434:
```suggestion
      // Cypress logs will only trigger an update every 4 milliseconds so there is a
```

tgriesser:
Was looking into this to understand a bit better, and I think this flake actually might actually be due to a legitimate regression in the `chore: refactor log function to class` merge here: https://github.com/cypress-io/cypress/pull/20382 

Specifically the fact that the original behavior seemed to create a single debounced `fireChangeEvent` per Log class.

https://github.com/cypress-io/cypress/blob/047aefc4ba0b2eeb60b869e8ac2fab1322034fa8/packages/driver/src/cypress/log.ts#L552-L556

whereas the new behavior:

https://github.com/sainthkh/cypress/blob/ce95894a35bedbee51920ae7073ee48a241195a4/packages/driver/src/cypress/log.ts#L556-L566

appears to create a new debounced fn, which is immediately invoked, per event fired. Might be missing something non-obvious since there's quite a few `fn` binds, etc. going on here, but that's just how it appears at first glance.

/cc @sainthkh

tgriesser (changes requested):

brian-mann:
Yes this is the problem per @tgriesser comment:

https://github.com/cypress-io/cypress/pull/20886#issuecomment-1088120615

The debounce cannot be on each log instance - it needs to be a singleton and all log calls need to go through that function.

sainthkh:
I admit that I refactored it wrong way. I didn't read the `_.debounce()` carefully. [It says that](https://lodash.com/docs/4.17.15#debounce):

```
Subsequent calls to the debounced function return the result of the last func invocation.
```

I guess creating debounce function every call changed the behavior. 

----

As `fireChangeEvent` is an internal function only used inside `log.ts` file. I think it's OK to give it `any` type. 

As I don't have the authority to push commits to this PR. I left my code suggestion at #20908.

tgriesser (approved):

davidmunechika (approved):

PR #20889, opened 2022-04-01
fix: build mjs in the cli

elevatebart:
closes UNIFY-1438

### Build/index.mjs missing

When publishing the Cypress CLI, we move all we need to a `build` folder. We forgot to move `index.mjs` to that directory.
For that reason, cypress was still breaking even after adding `index.mjs` to the `files` property of the `package.json`.

Issues should be resolved.

NOTE: This might create an unexpected behavior when using the `preserveValueImports` flag as it is removed. There is a good chance that this flag would have no effect in `commonjs` though.

NOTE2: ts-node recommend to use `"module": "commonjs"` in the `tsconfig.json`. Their `esnext` support is still experimental. Plus, if we set ts-node to sue ESM, CJS will not be understood anymore. 

### Invalid TypeScript config

We used to get the following error when running cypress against `npm init vue` with `TypeScript`.

This PR also changes the `tsconfig.json` used for reading the config file with ts-node. Since this updated config is not used for building anything, the modified parameter will not impact production.

https://www.typescriptlang.org/tsconfig#preserveValueImports

TSError: ⨯ Unable to compile TypeScript:
error TS5095: Option 'preserveValueImports' can only be used when 'module' is set to 'es2015' or later.

Screen Shot 2022-04-01 at 3 14 37 PM


system-tests/projects/ts-proj-4-5/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "@vue/tsconfig/tsconfig.web.json"
```
lmiller1990, line 2: Any way to provide a more minimal tsconfig here? It's not clear what this config contains/why it's needed here.

lmiller1990 (commented):
The PR description an example specifies the tsconfig.json shipped by the Vue CLI as problematic, but really the problem is `[preserveValueImports](https://www.typescriptlang.org/tsconfig#preserveValueImports)`, right?

Can we make the example more minimal and focused on that? Could the `tsconfig.json` in the system test project just be

```json
{
  preserveValueImports: true,
}
```

Or something like this? I feel it's much easier to understand what bug a system test addresses if it's as minimal and focused on the problem as possible. What do you think?

system-tests/projects/ts-proj-4-5/tsconfig.json
elevatebart: yes absolutely

lmiller1990 (dismissed):
My comments have been addressed, LGTM!

tgriesser (changes requested):

packages/server/lib/plugins/child/ts_node.js
```diff
@@ -3,18 +3,41 @@ const debugLib = require('debug')
...
 const tsnode = require('ts-node')
 const resolve = require('../../util/resolve')
+const semver = require('semver')
 
 const debug = debugLib('cypress:server:ts-node')
 
 const getTsNodeOptions = (tsPath, registeredFile) => {
+  const compiler = process.env.TS_NODE_COMPILER || tsPath
+
+  const version = require(compiler).version
```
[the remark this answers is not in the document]
elevatebart: What version of typescript does `typescript-cached-transpile` use? do you know?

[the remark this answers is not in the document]
elevatebart: Maybe I should simply ignore TS_NODE_COMPILER and do it using `tsPath` all the time. 

I want to make sure I use the version of typescript installed in the targetted repo, not the one where cypress is installed. Otherwise, if we use global mode, it will crash opening an `npm init vue` project

tgriesser (approved):

packages/server/lib/plugins/child/ts_node.js
```diff
@@ -3,18 +3,39 @@ const debugLib = require('debug')
...
+   * If we pass an unknown compiler option to ts-node,
+   * it ignores all options passed.
+   * If we want `module: "commonjs"` to always be used,
+   * we need to only set options that are supported in this version of TypeScript.
+   */
+  const compilerOptions = {
+    module: 'commonjs',
+    ...(semver.satisfies(version, '>=4.5.0')
+      // Only adding this option for TS >= 4.5.0
+      ? { preserveValueImports: false }
```
flotwig, line 28: i'm glad you fixed this, but why do we have TS versions < 4.5.0 in development hitting this anyways?

cli/scripts/start-build.js
```diff
@@ -24,3 +24,4 @@ includeTypes.forEach((folder) => {
 
 shell.exec('babel lib -d build/lib')
 shell.exec('babel index.js -o build/index.js')
+shell.cp('index.mjs', 'build/index.mjs')
```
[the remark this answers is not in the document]
flotwig: ahh now i see, we pack out of `build`

packages/server/lib/plugins/child/ts_node.js
flotwig, line 28: @elevatebart why do we need to pass this option in child processes?

cli/scripts/start-build.js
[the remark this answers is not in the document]
elevatebart: we do both

packages/server/lib/plugins/child/ts_node.js
elevatebart: this needs to be passed every where we compile ts-code from a user.
It can have a different version of TypeScript and a different `tsconfig.json` than what we have internally.

flotwig: i'm still not clear why this is needed

elevatebart: If the user has TypeScript beyond 4.5 and has this tsconfig

```json
{
  "compilerOptions": {
    "preserveValueImports": true,
    "module": "es2015"
  }
}
```

To have it work as close as possible of their config we will use their version of TypeScript and their tsconfig.json
We need to make it work with ts-node though, so we will change the "module" to be `commonjs`. 
Unfortunately this configuration makes TypeScript throw an error:

`preserveValueImports` can only be set with `module` at least `es2015`.

To circumvent this error (described in the ticket) we force the value to false.

BUT... ts-node does not accept any unknown values in the config so if we pass it all the time, it will fail in older versions of typescript that do not know this option. 

We protect it by only adding the option when we know it exists.

flotwig (approved):
I think there are some large unknowns here, but since this is required for unblocking some ecosystem support I'll approve it. I think we may see some really weird bug reports from this one so we'll have to keep an eye on it.

PR #20893, opened 2022-04-03
feat(launchpad): update CT setup and config scaffolding

lmiller1990:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1402

### User facing changelog

WIP

### PR Tasks


packages/scaffold-config/test/unit/detect.spec.ts
```diff
@@ -1,100 +1,199 @@
...
+  version: string
+}
+
+interface DevDepToFake {
+  devDependency: string
+  version: string
+}
+
+/**
+ * The way we detect dependencies is by using resolve-from (https://www.npmjs.com/package/resolve-from).
```
lmiller1990, line 27: I wonder if we can just do `require.resolve`? We already were using `resolve-from`, but I don't think it seems to do anything special `require.resolve` doesn't?

In the README it says

> Resolve the path of a module like [require.resolve()](https://nodejs.org/api/globals.html#globals_require_resolve) but from a given pathResolve the path of a module like [require.resolve()](https://nodejs.org/api/globals.html#globals_require_resolve) but from a given path

I think we can just do `require.resolve('react-scripts', { paths: [this.ctx.currentProject] })`.

@elevatebart I see you added the `resolve-from` module a while ago, did you know about `require.resolve(m, { paths: ... })`? Or did I miss something about what `resolve-from` does?

elevatebart: If tests pass with `require.resolve()` you should use that instead. Less dependencies is always better.
At the time, I had trouble with refreshing the cached values while running the app. A dependency that would not be there when the process started would never be detected as added when installed.

Do you want me to try and remove `resolve-from` in a separate PR?

packages/launchpad/src/setup/ManualInstall.vue
```diff
@@ -22,12 +22,23 @@
...
+          </ExternalLink>
+          <span class="rounded-md border border-gray-400 ml-6px">
+            <span class="p-4px">{{ dep.minVersion }}</span>
+            <span
+              v-if="!dep.satisfied && dep.detectedVersion"
+              class="border-l border-gray-400 bg-gray-100 text-gray-600"
+            >
+              Detected {{ dep.detectedVersion }}
+            </span>
+          </span>
```
ZachJW34, line 40:
```suggestion
          <span class="border rounded-md font-semibold border-gray-100 text-xs ml-6px text-gray-700">
            <span class="p-4px">{{ dep.minVersion }}</span>
            <span
              v-if="!dep.satisfied && dep.detectedVersion"
              class="border-l font-medium bg-gray-50 border-gray-100 px-4px text-gray-600"
            >
              detected {{ dep.detectedVersion }}
            </span>
          </span>
```

To more closely match Figma

ZachJW34 (changes requested):

ZachJW34:
We should sync with DX after this is merged so we can update the CT testing setup docs based on the required deps.

packages/scaffold-config/src/frameworks.ts
```diff
@@ -1,689 +1,243 @@
...
-        version: '^4.0.0',
-      },
-    ],
-    config: {
-      js: () => {
-        return dedent`
-      const { devServer } = require('@cypress/react/plugins/react-scripts')
+        export default defineConfig({
+          component: {
+            framework: '${config.framework}',
```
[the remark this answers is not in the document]
lmiller1990: Right, that's what's going on there. I thought it was a bug due to no object API. Thanks.

packages/data-context/src/sources/migration/codegen.ts
```diff
@@ -147,7 +147,7 @@ async function getPluginRelativePath (cfg: LegacyCypressConfigJson, projectRoot:
 // project's node_modules, we don't want to include
 // defineConfig(/***/) in their cypress.config.js,
 // since it won't exist.
-function defineConfigAvailable (projectRoot: string) {
+export function defineConfigAvailable (projectRoot: string) {
```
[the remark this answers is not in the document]
lmiller1990: Global mode and weird monorepos! https://cypress-io.atlassian.net/browse/UNIFY-1531

Not that common, but I ran into it during testing. I wanted to update a 9.x project and I was using global mode, so that's how I found it.

packages/scaffold-config/src/frameworks.ts
```diff
@@ -1,689 +1,243 @@
...
+    type: 'reactscripts',
+    configFramework: 'create-react-app',
+    category: 'template',
+    name: 'Create React App',
+    supportedBundlers: [dependencies.WIZARD_DEPENDENCY_WEBPACK],
+    detectors: [dependencies.WIZARD_DEPENDENCY_REACT_SCRIPTS],
+    storybookDep: dependencies.WIZARD_DEPENDENCY_STORYBOOK_REACT,
+    dependencies: (bundler: WizardBundler, projectPath: string): DependencyToInstall[] => {
+      return [
+        inPkgJson(dependencies.WIZARD_DEPENDENCY_WEBPACK, projectPath),
```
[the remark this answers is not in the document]
lmiller1990: I thought we landed on showing all the required dependencies we are **not** shipping, and they will always show up as "installed". More like a "confirming you've got the required dependencies".

Either we should include webpack **and** react - like you said, those are in the same boat (you should have it, and we are just making sure) or neither. It is not entirely clear from the tech brief or ticket if webpack should be here, or not.

Please confirm and let me know. Maybe we need an additional column [in the matrix](https://docs.google.com/spreadsheets/d/10bBADYvNiTztEkU5DGkbpUxOoMABLmU1-0nyLj1dXOM/edit#gid=645743321) showing "assumed dependencies" which would be react and webpack for `react-scripts`?

packages/scaffold-config/src/frameworks.ts
```diff
@@ -1,689 +1,243 @@
...
-        throw Error(`No config defined for ${bundler}`)
-      },
+    configFramework: 'vue',
+    category: 'library',
+    name: 'Vue.js 2',
+    detectors: [dependencies.WIZARD_DEPENDENCY_VUE_2],
+    supportedBundlers: [dependencies.WIZARD_DEPENDENCY_WEBPACK, dependencies.WIZARD_DEPENDENCY_VITE],
+    storybookDep: dependencies.WIZARD_DEPENDENCY_STORYBOOK_VUE_2,
+    dependencies: (bundler: WizardBundler, projectPath: string): DependencyToInstall[] => {
+      return [
```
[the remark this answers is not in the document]
lmiller1990: Yeeeeeees. On it.

packages/launchpad/src/setup/ManualInstall.vue
lmiller1990: I'm the worst at styling, thanks.

packages/scaffold-config/test/unit/detect.spec.ts
lmiller1990: Sure, how about after this one lands? Then you'll have the test coverage to refactor with more confidence.

packages/scaffold-config/src/frameworks.ts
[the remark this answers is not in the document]
lmiller1990: a5311084565136dea3d07b58ad36eb1b734b49c2

packages/scaffold-config/src/frameworks.ts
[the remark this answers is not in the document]
lmiller1990: 72168c0a773df2a2e4466b53cb800330b8a25294

packages/scaffold-config/src/frameworks.ts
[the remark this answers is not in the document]
lmiller1990: Added them all for good measure for now: 72168c0a773df2a2e4466b53cb800330b8a25294. Can easily change if needed.

lmiller1990:
> We should sync with DX after this is merged so we can update the CT testing setup docs based on the required deps.

@ZachJW34 for sure. We should not merge this until Object API & associated dev-servers are done - otherwise we will be generating config files that do not work, but we can still do all the review and docs prior to that.

packages/scaffold-config/test/unit/detect.spec.ts
```diff
@@ -1,100 +1,199 @@
...
+ *
+ * {
+ *   "version": "5.0.0",
+ *   "main": "index.js"
+ * }
+ *
+ * We have some real e2e tests that actually run `npm install`.
+ * Those are in launchpad/cypress/e2e/scaffold-component-testing.cy.ts.
+ */
+function fakeDepsInNodeModules (cwd: string, deps: Array<DepToFake | DevDepToFake>) {
```
tgriesser, line 41: Do we even still need to fake the deps? Shouldn't the system tests infra take care of this for us?

tgriesser (commented):
Looks good, added a comment. Also does this PR automatically skip the "dependencies" page if we've detected all of the dependencies exit?

packages/data-context/test/unit/helper.ts
```diff
@@ -16,7 +17,13 @@ export function getSystemTestProject<T extends typeof e2eProjectDirs[number]> (p
   return path.join(__dirname, '..', '..', '..', '..', 'system-tests', 'projects', project) as SystemTestProjectPath<T>
 }
 
-export async function scaffoldMigrationProject (project: typeof e2eProjectDirs[number]) {
+export function removeCommonNodeModules () {
+  fs.rmSync(path.join(Fixtures.cyTmpDir, 'node_modules'), { recursive: true, force: true })
+}
+
+fs.rmSync(path.join(Fixtures.cyTmpDir, 'node_modules'), { recursive: true, force: true })
```
tbiethman, line 24: Doesn't look like we want/need to call this here?

packages/graphql/src/schemaTypes/objectTypes/gql-WizardFrontendFramework.ts
```diff
@@ -9,26 +9,26 @@ export const WizardFrontendFramework = objectType({
   definition (t) {
     t.nonNull.field('type', {
       type: FrontendFrameworkEnum,
-      description: 'The name of the framework',
+      description: 'The unique identifier for a framework of library',
```
tbiethman, line 12:
```suggestion
      description: 'The unique identifier for a framework or library',
```

lmiller1990:
> Looks good, added a comment. Also does this PR automatically skip the "dependencies" page if we've detected all of the dependencies exit?

I didn't implement that. Instead, it just shows and confirms all the dependencies, and the user can click continue. I made it more explicit in the tests that use the system-tests infrastructure to install the node modules. See: https://github.com/cypress-io/cypress/pull/20893/commits/b487860b4ed1aedd80388c1df3d1957d2f3a7ec2

Here's how it looks:


image

packages/scaffold-config/test/unit/detect.spec.ts
lmiller1990: We've got real ones with real deps here: https://github.com/cypress-io/cypress/blob/lmiller1990/UNIFY-1402/packages/launchpad/cypress/e2e/scaffold-component-testing.cy.ts. They are kind of slow.

We also have more more here testing the correct install deps string is generated: https://github.com/cypress-io/cypress/blob/lmiller1990/UNIFY-1402/packages/data-context/test/unit/sources/WizardDataSource.spec.ts#L13

The reason I added these is so I could easily test many cases with less common configurations like 

```json
{
  "dependencies": {
    "vite": "version: '2.5.0-alpha.4"
  }
}
```

Without the overhead of actually running a yarn install, since this is supposed to be a unit test for the `detect` function. I see some value in more "unit-y" tests, so we can easily express less common cases without having to spin up an entire launchpad workflow test, eg:

```js
it(`React with Vite using pre-release version`, async () => {
    const projectPath = await scaffoldMigrationProject('react-vite-ts-unconfigured')

    fakeDepsInNodeModules(projectPath, [
      { devDependency: 'vite', version: '2.5.0-alpha.4' },
      { dependency: 'react', version: '17.0.0' },
    ])

    const actual = detect(projectPath)

    expect(actual.framework.type).to.eq('react')
    expect(actual.bundler.type).to.eq('vite')
  })
```

I could go either way, but to answer your original question:

- yes, we have e2e tests using the system tests infrastructure
- no, we don't strictly need to fake the deps, but I am not sure system-tests infra is the right tool for testing `detect`, which has quite a lot of conditionals that are more concisely expressed in the unit tests here.

I don't have a super strong opinion here, but these unit tests really helped with getting all the edge cases nailed, like `package.json` with less common configs, like `1.0.0-alpha.10`. What do you think?

packages/graphql/src/schemaTypes/objectTypes/gql-WizardFrontendFramework.ts
lmiller1990: https://github.com/cypress-io/cypress/pull/20893/commits/f5a85371cdf01433765e5844864bf129d68f0965

lmiller1990:
@tgriesser @ZachJW34 please re-review, I merged in 10.0-release and responded to all comments.

We can merge this last once the other tickets for the CT stuff are done, but still good to review.

packages/scaffold-config/src/frameworks.ts
```diff
@@ -1,689 +1,253 @@
...
+        }
       }`
-      },
-      ts: () => {
-        return dedent`
-      import { defineConfig } from 'cypress'
-      import { devServer } from '@cypress/react/plugins/react-scripts'
+  }
+
+  return dedent`
```
ZachJW34, line 94: The `cypress.config.js` should also use `defineConfig` if it is available.

ZachJW34 (commented):
Went through the deps, this is what I think we should land on for the required deps permutations:

```
Enumerated options:

| framework        | bundler | required deps                   |
| ---------------- | ------- | ------------------------------- |
| react            | webpack | react >= 16, webpack >= 4       |
| create-react-app | webpack | react >= 16, react-scripts >=4  |
| next             | webpack | next >= 11, react >= 16         |
| vue              | webpack | vue >= 2, webpack >= 4          |
| vue-cli          | webpack | vue >= 2, @vue/cli-service >= 4 |
| nuxt             | webpack | vue = 2, nuxt = 2               |
| react            | vite    | react >=16, vite >= 2           |
| vue              | vite    | vue >= 2, vite >= 2             |

> Note: `typescript` is a required dependency (for all projects) if the user chooses typescript during project setup
```

My reasoning behind the list is that we should only ask them to install deps we know we need for ct to work. For `create-react-app`, all we need is `react` and `react-scripts`. We don't need `webpack` since we can source it from `react-scripts`. Same for `vue-cli`.

I think we should get all the presets merged and then come to a final decision on this list. I'd like to update the tech-brief with this list as well.

packages/scaffold-config/src/frameworks.ts
lmiller1990: Did not realise. I will make this change at the end once all the presets are merged, along with updating the dependencies based on whatever decision we arrive at.

packages/data-context/src/actions/WizardActions.ts
```diff
@@ -83,44 +88,44 @@ export class WizardActions {
...
-    })
+    this.resetWizard()
 
     await this.detectLanguage()
     debug('detectedLanguage %s', this.data.detectedLanguage)
     this.data.chosenLanguage = this.data.detectedLanguage || 'js'
 
     try {
-      const detected = detect(await fs.readJson(path.join(this.ctx.currentProject, 'package.json')))
+      const detected = detect(this.ctx.currentProject)
```
tbiethman, line 114: Can we remove the try/catch now that we're not awaiting? Nothing else seems like it should (reasonably) throw.

lmiller1990: Done

lmiller1990:
Ok, Object API seems in a working state - I'll give this one a bit more of a test and get it merged.

tgriesser (approved):
It's all coming together!

ZachJW34 (approved):

PR #20906, opened 2022-04-05
fix: initialization lifecycle after project setup

ryanthemanuel:
- Closes [UNIFY-1443](https://cypress-io.atlassian.net/browse/UNIFY-1443)

### User facing changelog

Going through the project set up will now always show a list of browsers on the "Choose a Browser" page upon completion

### Additional details

On https://github.com/cypress-io/cypress/pull/20775, we made a change to be more explicit about the config loading lifecycle by having the callee be deliberate with loading the different phases of the config. In the process, we missed a spot with setting up new projects where we were counting on initializing the config to immediately proceed with setting and loading the testing type. This PR fixes that scenario.

It's worth noting that we weren't completely testing this scenario previously. While we were ensuring we got to the "Choose a Browser" page as part of project setup, we didn't check that the browsers were actually loading.

### PR Tasks


tgriesser (approved):
Thanks for the fix @ryanthemanuel!

PR #20916, opened 2022-04-05
fix: Resend *.enable commands on CDP reconnect

### LINKED ISSUE #20618: Leaving Cypress open for a long time or putting computer to sleep results in missing command logs
Leaving Cypress open for a long time or putting one's computer to sleep will result in command logs missing when using a Chromium-based browser.

Certain command logs like download and network events will no longer appear after running tests without closing and re-opening the browser

In production Cypress, this is only a visual/informational loss, but in the multi-domain feature work, this results in tests failing (after a page load timeout, which is 60s by default, so it seems like the test is hanging).

chrisbreiding:
- Closes #20618

### User facing changelog

- Command logs for downloads and network requests will correctly display in Chrome after Cypress reconnects to the DevTools Protocol

### Additional details

After detecting a disconnect from the CDP, we recreate the connection and resubscribe to event listeners. However, CDP won't send events unless the corresponding `*.enable` command is sent first. Since we didn't resend the `*.enable` commands, no events were received.

This fixes that by keeping track of any `*.enable` commands, similar to how we keep track of event subscriptions, and resending them when we reconnect to the CDP.

Previously, after a reconnect, command logs like download events would not appear. Now, they will be displayed as expected.

### PR Tasks


packages/server/lib/browsers/cri-client.ts
```diff
@@ -175,13 +186,20 @@ export const create = async (target: websocketUrl, onAsynchronousError: Function
   }
 
   const connect = async () => {
-    cri?.close()
+    await cri?.close()
```
flotwig, line 189: nice catch!

packages/server/lib/browsers/cri-client.ts
```diff
@@ -226,10 +244,20 @@ export const create = async (target: websocketUrl, onAsynchronousError: Function
         })
       }
 
+      // Keep track of '*.enable' commands so they can be resent when
+      // reconnecting
+      if (command.endsWith('enable')) {
```
flotwig, line 249: nit: i am bothered more than i should be that this is `'enable'` and not `'.enable'`, even though it literally does not matter at all with the snakeCase commands

```suggestion
      if (command.endsWith('.enable')) {
```

flotwig (approved):
really nice work fixing this @chrisbreiding, I feel like this is one of those bugs that users definitely run into and have no idea what's happening or how to report it

flotwig (approved):

mschile (approved):

mjhenkes (approved):

PR #20920, opened 2022-04-05
fix(unify): null `binary` field on `editor` blocks launchpad from opening

davidmunechika:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1376?atlOrigin=eyJpIjoiMzdkYWI5MDgzNjA3NDRjNGJlOGMyNzRiNDg5NWE5YjAiLCJwIjoiaiJ9

### User facing changelog
It's possible for users to create an Editor with no `binary` property and to have already done this unintentionally in previous versions of Cypress. This field should be nullable and the case should be handled for when binary is not present.

### PR Tasks


tbiethman (approved):
Tested 'er out, no more errors! We're also handling cleared values gracefully through the v10 UI, no issues unsetting a previously set value.

ZachJW34 (approved):

PR #20921, opened 2022-04-05
feat: Align Select Preferred Editor modal with Figma design

estrada9166:
- Closes  [UNIFY-1240](https://cypress-io.atlassian.net/browse/UNIFY-1240)

### User facing changelog

Update preferred editor modal with Figma design

## Before
image

## After
image

## Settings
image

### PR Tasks


packages/frontend-shared/src/gql-components/ChooseExternalEditorModal.cy.tsx
```diff
@@ -13,14 +13,20 @@ describe('ChooseExternalEditor', { viewportHeight: 400, viewportWidth: 300 }, ()
...
       },
       render: (gql) => {
-        return <div class="p-16px"><ChooseExternalEditor gql={gql} /></div>
+        return <div class="p-16px"><ChooseExternalEditorModal gql={gql} open /></div>
       },
     })
 
+    cy.contains(defaultMessages.globalPage.externalEditorPreferences)
+    cy.contains(defaultMessages.globalPage.externalEditorPreferencesDescription1)
+    cy.contains('You can change this later in the Settings page.')
```
rachelruderman, line 22: More of a question for me than for you -- when coming across hardcoded strings that are not in `defaultMessages`, is the expectation to leave as is or to update `defaultMessages`?

rachelruderman: Ah I scrolled on further down and I see 
```
    "externalEditorPreferencesDescription2": "You can change this later in the {0} page.",
```
Is it not usable in the test b/c of the `{0}` variable?

rachelruderman:
Perhaps out of scope for this ticket, but should the UI in Settings be updated to match the new modal design? Currently it mirrors the old one:

Screen Shot 2022-04-06 at 9 25 32 AM

mapsandapps:
I'm checking with Ryan about what icons we should use for these two form elements.

packages/frontend-shared/src/gql-components/ChooseExternalEditorModal.cy.tsx
estrada9166: I think I may need to update it, found another test where we are using something like `defaultMessages.globalPage.externalEditorPreferencesDescription1.replace('{0}', 'Settings')` and it should work! good catch!

mapsandapps:
@estrada9166 Ryan updated the [figma](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/Cypress-App?node-id=4088%3A81314) with the correct icons. All your other changes are correct. The presentation of the two form inputs should be the same on the Settings page & the modal.

mapsandapps (approved):
Thanks for updating it! Looks great 👍🏻

packages/launchpad/cypress/e2e/open-mode.cy.ts
```diff
@@ -105,10 +105,17 @@ describe('Launchpad: Open Mode', () => {
...
+
+      cy.findByTestId('project-card')
+      cy.get('[aria-label="Project Actions"]').click()
+      cy.get('button').contains('Open In IDE').click()
+
       cy.intercept('POST', 'mutation-ChooseExternalEditorModal_SetPreferredEditorBinary').as('SetPreferred')
       cy.get('@modal').contains('Choose your editor...').click()
       cy.get('@modal').contains('Well known editor').click()
-      cy.get('@modal').contains('Done').click()
+      cy.get('@modal').contains('Save Changes').click()
```
rachelruderman, line 118:
```suggestion
      cy.get('@modal').contains(defaultMessages.globalPage.saveChanges).click()
```

rachelruderman (approved):

jennifer-shehane:
I approved the Percy snapshots.

PR #20922, opened 2022-04-05
chore: Unskip test that now passes

chrisbreiding:
### User facing changelog

N/A

### Additional details

- Unskips a test that now passes after recent work.
- Throws an error when using `Cypress.session.clearAllSavedSessions()` in the `switchToDomain` callback, which was missed when handling non-support of `cy.session()`.

### PR Tasks


mschile (approved):

packages/driver/cypress/integration/e2e/multi-domain/multi_domain_cypress_api.spec.ts
```diff
@@ -240,5 +231,17 @@ describe('multi-domain Cypress API', { experimentalSessionSupport: true }, () =>
         Cypress.Cookies.preserveOnce({})
       })
     })
+
+    it('throws an error when a user attempts to call Cypress.session.clearAllSavedSessions() inside of multi-domain', (done) => {
```
mschile, line 235: Maybe move to the `multi-domain-unsupported-commands` spec?

mschile (approved):

packages/driver/cypress/integration/e2e/multi-domain/multi_domain_cypress_api.spec.ts
chrisbreiding: I had the same thought before, but the other `Cypress.*` unsupported methods are in here and this isn't a command...

packages/driver/src/cypress/error_messages.ts
```diff
@@ -1807,6 +1807,10 @@ export default {
         message: `${cmd('session')} use is not supported in the ${cmd('switchToDomain')} callback. Consider using it outside of the callback instead. Otherwise, please 👍 the following issue and leave a comment with your use-case:`,
         docsUrl: 'https://on.cypress.io/github-issue/20721',
       },
+      Cypress_session: {
```
emilyrohrbough, line 1810: Just curious, what are the reasons this can't be supported if we are expecting users to cy.session with cy.switchToDomain?

chrisbreiding: It's not that it necessarily can't be supported. For the experimental release, we're choosing not the support it. We also don't know of any use-cases where it needs to be supported within the switchToDomain callback that won't work if it's used outside of the callback instead.

mjhenkes (approved):

chrisbreiding:
The test I unskipped is legitimately failing, so I'm looking into that. It passes when using `cypress open`, but fails with `cypress run` for some reason.

PR #20925, opened 2022-04-05
feat: add new vite-dev-server package based on object API & bundling

tgriesser:
- Closes UNIFY-1401

### User facing changelog
Adds support for the new object syntax for the `vite-dev-server`. Adds a new `vite-dev-server-fresh` package to manage this support, it will replace the existing `vite-dev-server` once all frameworks are supported.

Adds "open mode" tests for vite + react in `2.8.1` & `2.9.6`, including recovering from failures, and adds system-tests.

Also marked the `vite-dev-server-fresh` `package.json` with `internal: true` for now to ensure it doesn't accidentally get published. 

### PR Tasks


lmiller1990 (dismissed):
Cool, seems fine - took me a while to review since there's a bit of copy+paste with some subtle changes to the vite dev server, but I like the changes.

I left a few questions, I don't see them as blockers, though.

npm/vite-dev-server-fresh/cypress/e2e/react.cy.ts
```diff
@@ -0,0 +1,68 @@
...
+type ProjectDirs = typeof e2eProjectDirs
+
+const VITE_REACT: ProjectDirs[number][] = ['vite2.8.6-react', 'vite2.9.1-react']
+
+// Add to this list to focus on a particular permutation
+const ONLY_PROJECTS: ProjectDirs[number][] = []
+
+afterEach(() => {
+  // Visiting about.blank to cleanup any sockets left over between each test,
+  // such as __vite_ping
```
brian-mann, line 14: this should happen in a `beforeEach`, not `afterEach`, otherwise the page is wiped each time.

Also its unclear why this is necessary - won't this be a problem for all users if we have to do this?

system-tests/test/vite_dev_server_fresh_spec.ts
```diff
@@ -0,0 +1,25 @@
+import systemTests from '../lib/system-tests'
+
+describe('@cypress/vite-dev-server-fresh', function () {
+  systemTests.setup()
+
+  describe('react', () => {
+    systemTests.it('runs specs in vite 2.8.6', {
+      project: 'vite2.8.6-react',
```
brian-mann, line 8: why aren't we setting `snapshot: true` here? is it because of non-deterministic `stdout` coming from the vite bundler? could we pass an environment variable or an `options: { ... }` into vite to silence this or log it to `stderr` ?

system-tests/projects/vite2.8.6-react/package.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
tgriesser: It’s not empty it’s renamed which GitHub shows as zero lines changed

system-tests/test/vite_dev_server_fresh_spec.ts
tgriesser: Added snapshots

packages/frontend-shared/cypress/e2e/support/e2eSupport.ts
```diff
@@ -245,8 +245,20 @@ function startAppServer (mode: 'component' | 'e2e' = 'e2e') {
           initializeActiveProjectStub.restore()
         }
 
+        function formatError (e: Error & {messageMarkdown?: string, originalError?: Error}) {
+          if (e.messageMarkdown) {
+            e.message = e.messageMarkdown
```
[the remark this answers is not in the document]
tgriesser: Yeah, will add a comment to clarify but previously an error looked like this: 
image

and now it looks like this:

image

Basically, the error we're receiving is normally expected to be formatted for the terminal or the UI, but not the reporter. The cy-in-cy tests should be the only place we need to do this

npm/vite-dev-server-fresh/cypress/e2e/react.cy.ts
```diff
@@ -0,0 +1,68 @@
+/// <reference types="cypress" />
+/// <reference path="../support/e2e.ts" />
+import type { e2eProjectDirs } from '@packages/frontend-shared/cypress/e2e/support/e2eProjectDirs'
+
+type ProjectDirs = typeof e2eProjectDirs
+
+const VITE_REACT: ProjectDirs[number][] = ['vite2.8.6-react', 'vite2.9.1-react']
+
+// Add to this list to focus on a particular permutation
+const ONLY_PROJECTS: ProjectDirs[number][] = []
```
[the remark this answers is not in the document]
tgriesser: Yeah, we had this in the tr-node spike and it was super useful

npm/vite-dev-server-fresh/src/plugins/cypress.ts
```diff
@@ -0,0 +1,157 @@
...
+      indexHtmlContent.substring(endOfBody)
+    }`
+    },
+    configureServer: async (server: ViteDevServer) => {
+      server.middlewares.use(`${base}index.html`, async (req, res) => {
+        let transformedIndexHtml = await server.transformIndexHtml(base, '')
+        const viteImport = `<script type="module" src="${options.cypressConfig.devServerPublicPathRoute}/@vite/client"></script>`
+
+        // If we're doing cy-in-cy, we need to be able to access the Cypress instance from the parent frame.
+        if (process.env.CYPRESS_INTERNAL_VITE_OPEN_MODE_TESTING) {
```
tgriesser, line 74: This change was needed to get cypress-in-cypress working

brian-mann: this shouldn't be necessary - the reason this is happening is because `top` is setting `document.domain` for `e2e` mode. If top stops doing that for component testing, its unnecessary to inject `document.domain` into the iframe.

This likely applies to webpack too - or any other situation where top needs to communicate with the iframe.

brian-mann: if you have an frame that sets document.domain = '...' and even if you embed an iframe on the same domain it also needs to set that. If neither set it (and are on the same domain) they do not need to set document.domain to talk to one another.

npm/vite-dev-server-fresh/cypress/e2e/react.cy.ts
tgriesser: Good to know

npm/vite-dev-server-fresh/src/plugins/cypress.ts
tgriesser: Got it, makes sense

packages/frontend-shared/cypress/e2e/support/e2eSupport.ts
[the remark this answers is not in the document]
lmiller1990: :+1:

lmiller1990 (approved):
Re-approving - my comments were addressed, looks like there might be a few more from Brian that need to be addressed.

npm/vite-dev-server-fresh/cypress/e2e/react.cy.ts
tgriesser: Added `experimentalSessionSupport` to the cy config

tgriesser:
> They can override this via the viteConfig but should we explicitly set this too false?

Yeah, we should prob set this to false

> More broadly, what options should we lock down (like server related options) and have we tested that our overrides that we need to make are preserved after merge?

Not sure, but now that we have a solid test foundation and bundling in-house, I think we'll be able to address concerns as they arise.

PR #20928, opened 2022-04-05
backmerge develop into 10.0-release

marktnoonan:
Straigthforward merge, two conflicts:

- `CRIWrapper` vs `CRI` in `packages/server/lib/browsers/cri-client.ts` - adjusted to keep `CRIWrapper` in 10.0-release
- small conflict around `criStub` in `packages/server/test/unit/browsers/cri-client_spec.ts`

### DIFF
--- a/packages/server/lib/browsers/cri-client.ts
+++ b/packages/server/lib/browsers/cri-client.ts
@@ -17,6 +17,9 @@ const WEBSOCKET_NOT_OPEN_RE = /^WebSocket is (?:not open|already in CLOSING or C
  */
 export namespace CRIWrapper {
   export type Command =
+    'Page.enable' |
+    'Network.enable' |
+    'Console.enable' |
     'Browser.getVersion' |
     'Page.bringToFront' |
     'Page.captureScreenshot' |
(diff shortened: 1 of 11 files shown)


packages/server/lib/browsers/cri-client.ts
```diff
@@ -155,6 +166,13 @@ export const create = async (target: string, onAsynchronousError: Function, host
       port,
       target,
       local: true,
+      // Minor optimization. chrome-remote-interface creates a DSL based on
+      // this so you can call methods instead of using the event emitter
+      // (e.g. cri.Network.enable() instead of cri.send('Network.enable'))
+      // We only use the event emitter, so if we pass in an empty protcol,
+      // it will keep c-r-i from looping through it and needlessly creating
+      // the DSL
+      protocol: { domains: [] },
```
marktnoonan, line 175: @chrisbreiding after merging change with 10.0-release, I am getting two kinds of type errors that cause CI to fail:

1. It seems the `protocol` object expects a `version` key with a value like `{ major: number, minor: number }`
2. With the `protocol` object correctly formed, when it is present I get a second ts error warning that `host` is not allowed. This is more confusing.

Is this optimization worth sorting out now as part of the merge, or should I remove this and make a ticket to re-add?

chrisbreiding: It's okay to remove it. I wouldn't worry about making a ticket to re-add it. I doubt it would be worth the effort.

emilyrohrbough (approved):

elevatebart (approved):

PR #20732, opened 2022-03-22
spike: UNIFY-1354 consolidating devServer config & client modules

tgriesser:
Code related to: UNIFY-1354, UNIFY-1362

PR #20930, opened 2022-04-05
feat: embedding mount into the cypress binary (real dependency)

JessicaSachs:
- Closes [UNIFY-1386](https://cypress-io.atlassian.net/browse/UNIFY-1386)

### Approach
This PR essentially copies the built libraries from **npm/vue**, **npm/react**, and **npm/mount-utils** and keeps the **cli/package.json**'s exports map up to date.

This avoids making changes to specific dev-servers or relying on specific bundler features like `resolve.alias`.

### Resources
- Original (dev-server specific, alias-based approach) https://github.com/cypress-io/cypress/pull/20732
- Tech Brief (contains all CT changes) https://github.com/cypress-io/tech-briefs/pull/18

### User facing changelog
Cypress users no longer have to install the Vue or React mounting libraries separately when mounting components. If you are using Vue 3 or React, you can remove those dependencies from your `package.json` entirely and replace the imports to the newly supported `cypress/vue` and `cypress/react` namespaces.

For Vue 2 users, a separate pull request will add support for importing mount from `cypress/vue-2`.

#### Before: Externally Install Mounting Libs
Mounting libraries are always explicitly imported from the Cypress npm organization namespace.

```js
// React
import { mount } from '@cypress/react'

// Vue 3
import { mount } from '@cypress/vue' // v3.x, pinned in your package.json

// Vue 2
import { mount } from '@cypress/vue' // v2.x, pinned in your package.json
```

#### After: Bundling Mounting Libs
Mounting libraries are shipped with Cypress, and are exposed as subpaths form the Cypress namespace. This is done as part of the cli build process.

This API is supported via the `cypress` package's `exports` map as defined in the **package.json**.

```js
// React

// Vue 3
import { mount } from 'cypress/vue'

// Vue 2, still have to npm install the pinned version of @cypress/vue for now
import { mount } from '@cypress/vue'
```


scripts/sync-exported-npm-with-cli.js
```diff
@@ -0,0 +1,78 @@
...
+ */
+const shell = require('shelljs')
+const path = require('path')
+const packlist = require('npm-packlist')
+const fs = require('fs')
+
+shell.set('-v') // verbose
+shell.set('-e') // any error is fatal
+
+// 2. This script will be run in a postbuild task for each npm package
```
JessicaSachs, line 16:
```suggestion
// This script will be run in a postbuild task for each npm package
```

cli/.gitignore
```diff
@@ -1,3 +1,6 @@
+# You must keep cli/.gitignore in sync with cli/.npmignore
+# becuase .npmignore is preferred when publishing
+# See: https://stackoverflow.com/a/24942436
```
JessicaSachs, line 3:
```suggestion
```

scripts/sync-exported-npm-with-cli.js
```diff
@@ -0,0 +1,78 @@
...
+    // ./react/dist/cypress-react-esm.js, etc
+    subPackageExports.import = `./${exportName}/${esmEntry}`
+  }
+
+  if (!isModule) {
+    // ./react/dist/cypress-react-cjs.js, etc
+    subPackageExports.require = `./${exportName}/${currentPackageConfig.main}`
+  }
+
+  if (cliPackageConfig.files.includes(exportName)) {
```
JessicaSachs, line 68:
```suggestion
  if (!cliPackageConfig.files.includes(exportName)) {
```

JessicaSachs:
It works!!!! Going to remove the circle changes.

tgriesser (approved):
Nice! We can probably look into symlinking the dirs for local development in a follow up PR

cli/package.json
```diff
@@ -117,9 +120,20 @@
...
     "./package.json": {
       "import": "./package.json",
       "require": "./package.json"
+    },
+    "./react": {
+      "import": "./react/dist/cypress-react.esm-bundler.js",
+      "require": "./react/dist/cypress-react.cjs.js"
+    },
+    "./mount-utils": {
+      "require": "./mount-utils/dist/index.js"
```
BlueWinds, line 136: No import entry for mount-utils?

JessicaSachs: No it wasn't built for ESM.

Because only cypress-react and cypress-vue consume it and **they** have an esm export it's not a problem.

We have tech debt to convert all of our npm/* packages to be esm-first and share the same rollup + typescript compilation settings.

scripts/sync-exported-npm-with-cli.js
```diff
@@ -0,0 +1,78 @@
...
+    cliPackageConfig.files.push(exportName)
+  }
+
+  const output = `${JSON.stringify(cliPackageConfig, null, 2) }\n`
+
+  // eslint-disable-next-line no-console
+  console.log('Writing to CLI package.json for', exportName)
+
+  fs.writeFileSync(path.join(cliPath, 'package.json'), output, 'utf-8')
+})
```
lmiller1990, line 78: The tiniest thing but this seems like pretty important logic to test - do we need some kind of basic test to ensure that the `package.json` updates correctly, and that `import { mount } from 'cypress/vue` is actually working? We don't have any coverage for this right now, from what I can see.

I wonder if we can update all internal code to use the deep import syntax in the future. Is this something we plan to do?

lmiller1990 (commented):
Code seems fine, just have a comment around how we can test this and the end goal (we will migrate to `import ... from 'cypress/vue'` internally? How do we QA and verify this works (no tests for it so far).

cli/scripts/start-build.js
```diff
@@ -22,6 +22,10 @@ includeTypes.forEach((folder) => {
   shell.cp('-R', source, 'build/types')
 })
 
+// TODO: Add a typescript or rollup build step
```
BlueWinds, line 25: Is there a github issue for this we can link? Have been trying to do `TODO: blah, https://github.com/blahblah`.

scripts/sync-exported-npm-with-cli.js
lmiller1990: We can test it by adding code here https://github.com/cypress-io/cypress-example-kitchensink/pull/528

I'll do that. It'll break all other pipelines until this is merged, maybe I'll create a new branch for it (in kitchensink).

BlueWinds: I was also wondering if we can convert a couple of the existing system tests to `import { mount } from 'cypress/vue'` rather than `import { mount } from '@cypress/vue'`.

JessicaSachs: Oh, here https://github.com/cypress-io/cypress-example-kitchensink/pull/547/files

BlueWinds: That probably only works in `test-binary`, where we have tests that run in a docker container against the compiled binary. But still worth looking into. Just an extremely simple 'can import' test there.

JessicaSachs: Yeah in general I feel uncomfortable that we don't have a lot of post-build infrastructure for testing the output of Circle CI's build scripts. The kitchen sink is the right place. I put a PR up above that will at least fail if those packages aren't declared in the exports map.

cli/scripts/start-build.js
JessicaSachs: There isn't right now.

BlueWinds:
Looks good to me. I'd love to see a binary system test to cover this, but if you want to hold off for a separate PR on that, I'll approve this one as-is.

BlueWinds (approved):

PR #20932, opened 2022-04-06
fix: remove extra bundled electron in 10.0 binary

tgriesser:
- Closes UNIFY-1388

### User facing changelog

Fixes the giant electron that was added to the binary accidentally:

image

### Additional details

Fixes the places where we're eagerly loading code which needs to execute after electron is loaded. Moves electron to the binary root, since it is depended on by multiple modules. It's technically a "virtual" dependency, in the sense that it doesn't get bundled when the app is shipped, so having it as a root `devDependency` seems correct.

### PR Tasks


packages/electron/package.json
```diff
@@ -23,7 +23,6 @@
     "minimist": "1.2.6"
   },
   "devDependencies": {
-    "electron": "15.3.5",
```
flotwig, line 26: we are back to relying on the magic of hoisting to make our dependencies resolve :cry:

tgriesser: I noted this in the PR comment:

> Moves electron to the binary root, since it is depended on by multiple modules. It's technically a "virtual" dependency, in the sense that it doesn't get bundled when the app is shipped, so having it as a root devDependency seems correct.

It's not magic, it's the node module resolution. At least in this case it's explicitly hoisted at the root

flotwig: It seems more correct to me to put the electron dependency in @packages/electron since the purpose of that package is to expose it, and not put it in root because it's more convenient, have some separation of concerns. Yarn also explicitly discourages adding deps to the root workspace. This is just being done this way because it's convenient 😛

flotwig (dismissed):

packages/data-context/package.json
```diff
@@ -27,7 +27,6 @@
     "dayjs": "^1.9.3",
     "dedent": "^0.7.0",
     "ejs": "^3.1.6",
-    "electron": "14.1.0",
```
lmiller1990, line 30: Should we not leave it as a `devDependency` here, since it's imported in `data-context`? (if only for types with `import type`)?

packages/electron/package.json
lmiller1990: Just looking, isn't it weird that we listed `electron` as a `devDependency` in `packages/electron` in the first place - since we are not actually requiring it anywhere in there?

image

Edit: I suppose it was due to

```js
// ensure we have an electronVersion set in package.json
if (!(electronVersion = pkg.devDependencies.electron)) {
  throw new Error('Missing \'electron\' devDependency in ./package.json')
}
```

or the fact we have `electron-package` as a dependency.

I don't fully understand why it's weird to put it at the top level, considering it's a hard dependency across multiple packages. What exactly makes workplace root dependencies a bad practice? I haven't worked with complex monorepos before Cypress, so I don't fully understand the implications - asking for my own understanding, as opposed to a review of this part of the PR.

lmiller1990 (commented):
I left some questions (not really a review as such). I'm glad we are not bundling a huge electron anymore.

Also, couldn't we guard against accidentally blowing out the size of the binary with this check: https://github.com/cypress-io/cypress/blob/10.0-release/scripts/binary/zip.js#L78-L81. I wonder if we can decrease it now we've removed this?

packages/electron/package.json
tgriesser: > I don't fully understand why it's weird to put it at the top level, considering it's a hard dependency across multiple packages.

I personally think it's fine to have it at the root, it actually avoids hoisting, where hoisting is defined as a peer `packages/` dependency having a dependency, which gets installed at the root `node_modules` rather than `packages/$pkg/node_modules` for reasons that should be an implementation detail of the package manager. 

By putting it in the root as a devDependency, you're basically saying "this is needed in multiple places in the repo during development, so keep a single copy here". I actually kind of wish we did that with more dev only things like `mocha`, `chai`, etc. like we do in services repo, instead we have like 20 different versions of each library that get installed everywhere during development.

lmiller1990: Yeah was thinking the same thing, we definitely need to have a centralised mocha/chai for consistency across the mono repo.

lmiller1990 (approved):

lmiller1990:
Opinions on monorepos and dependencies aside, I think we can all agree that removing this huge, duplicated dependency is a win :100:

flotwig (approved):

packages/data-context/package.json
lmiller1990: I think this might have broke the `ELECTRON_RUN_AS_NODE` feature. docs: https://docs.cypress.io/guides/continuous-integration/introduction#Running-headless-tests-without-Xvfb

https://github.com/cypress-io/cypress/issues/23636

PR #20938, opened 2022-04-06
feat: allow passing --inspect / --inspect-brk in disted binary

tgriesser:
### User facing changelog

N/A

### Additional details

Rather than keying off the `--dev` flag to determine whether we can `--inspect` / `--inspect-brk`, which makes it only possible in local development, this instead checks for `--inspect` / `--inspect-brk` in the `process.argv` to add them to the available flags to trigger debugging. This way, the commands don't show up in `--help` / aren't advertised as a public contract, but they're available if you really need them.

### PR Tasks

- [x] N/A Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [x] N/A Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [x] N/A Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [x] N/A Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -29,7 +29,7 @@ mainBuildFilters: &mainBuildFilters
       only:
         - develop
         - 10.0-release
-        - marktnoonan/merg-develop-2022-4-5
+        - tgriesser/feat/always-add-inspect-flag
 
 # usually we don't build Mac app - it takes a long time
 # but sometimes we want to really confirm we are doing the right thing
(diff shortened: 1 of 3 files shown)


flotwig (commented):
@tgriesser you are knocking it out of the park with making Cypress more debuggable. PR looks good, I just pushed a commit so I can test this locally with a pre-release build. Once I've tested LGTM.

flotwig:
@tgriesser I'm not having success running `cypress open --inspect-brk` on Linux, does the `darwin` build work for you? 

`cypress open --inspect-brk` launches fine, but it never breaks for debugging.

tgriesser:
Looked into why this isn't working, pretty sure i know what it is, but in looking at it realizing the CLI tests are all messed up with unhandled errors that aren’t propagating and such (unrelated to this change), so I'd like to do a bit more investigation work before fixing. Will revisit when I get a chance, marking as draft

BlueWinds:
@tgriesser - Mind if we close this, and reopen when you're ready to work on it again?

Issue #18947, opened 2021-11-16
Windows 10 npm vesion throws Passthrough is not supported, GL is disabled


flowt-au:
### Current behavior

I haven't used Cypress for a couple f months then this error started showing up:
`[1720:1117/073201.722:ERROR:gpu_init.cc(441)] Passthrough is not supported, GL is disabled`

I was on version 8.4.0. So I did a `cypress clear cache`, then installed v8.7.0 (interestingly v9.0.0 is available but npm elected to install 8.7.0)
I got the `It looks like this is your first time using Cypress: 8.7.0` then 
`✔  Verified Cypress!`  then
`Opening Cypress...` then, after about a minute,
`[1720:1117/073201.722:ERROR:gpu_init.cc(441)] Passthrough is not supported, GL is disabled`

### Desired behavior

It should open.

### Test code to reproduce

n.a. See above

### Cypress Version

v8.7.0

### Other

I recently added a big monitor and am using the DisplayPort interface and this is the first time I am running Cypress since then. Given the error is about 'gpu' could there be some hardware setup / compatibility issue? Just a thought.

vincentixprice:
The same here, except for slightly different code and message: 

"[16080:1122/151752.429:ERROR:gpu_init.cc(453)] Passthrough is not supported, GL is disabled, ANGLE is"

Cypress v9.0.0

slemme1:
On Windows 10 I get 
"[7556:1129/115427.042:ERROR:gpu_init.cc(441)] Passthrough is not supported, GL is disabled" 
in console when running "npx cypress open" but the IDE and tests still run.  
The only thing I have changed recently is Cypress version to 8.7 (though corporate Windows/.NET/Chrome patches may have been applied).  If I find any more details (going back to previous or newer Cypress version) I will let you know.

flowt-au:
Thanks. Unfortunately, neither the IDE nor the tests run for me. Glad to know I am not alone here in having the issue, though.
Cheers,
Murray

flowt-au:
Clarification: the Passthrough error only comes up with the IDE version of Cypress. 
My CLI tests were failing for an entirely different (self-inflicted) reason.

flowt-au:
And on my OSX machine, running same version of Cypress 8.7.0 IDE against the same tests and Cypress settings, plugins, etc - works fine. So, it seems it is a Windows only issue.

flowt-au:
[...]

PR #20939, opened 2022-04-06
fix: silence dbus errors

### LINKED ISSUE #19299: "Failed to connect to the bus: Address does not contain a colon" throws a SIGSEGV error
### Current behavior

Our tests intermittently crash in our Docker runs with the error below. We get the `Failed to connect to the bus: Address does not contain a colon` error often, but it doesn't crash it.

But, when it *does* crash, that error is always the last one before the SIGSEGV.

I think we're going to downgrade to 8.5.0 to fix this as it wasn't happening back then. However, before we do, is there some kind of diagnostic log that will help you, help me, help you? :)

Thanks, and I know this is vague, not sure where to even start here.

```
[15:24:18]W:		 [Docker Run] [111:1207/202418.627106:ERROR:bus.cc(392)] Failed to connect to the bus: Address does not contain a colon
[15:24:19] :		 [Docker Run] The Test Runner unexpectedly exited via a exit event with signal SIGSEGV
[15:24:19] :		 [Docker Run] 
[15:24:19] :		 [Docker Run] Please search Cypress documentation for possible solutions:
[15:24:19] :		 [Docker Run] 
[15:24:19] :		 [Docker Run] https://on.cypress.io
[15:24:19] :		 [Docker Run] 
[15:24:19] :		 [Docker Run] Check if there is a GitHub issue describing this crash:
[15:24:19] :		 [Docker Run] 
[15:24:19] :		 [Docker Run] https://github.com/cypress-io/cypress/issues
[15:24:19] :		 [Docker Run] 
[15:24:19] :		 [Docker Run] Consider opening a new issue.
[15:24:19] :		 [Docker Run] 
[15:24:19] :		 [Docker Run] ----------
[15:24:19] :		 [Docker Run] 
[15:24:19] :		 [Docker Run] Platform: linux-x64 (Debian - 10.10)
[15:24:19] :		 [Docker Run] Cypress Version: 9.1.1
[15:24:19] :		 [Docker Run] info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
[15:24:19]W:		 [Docker Run] error Command failed with exit code 1.
[15:24:19]W:		 [Docker Run] Error: Command failed with exit code 1: yarn test:cypress --reporter teamcity --config video=false,defaultCommandTimeout=10000 --record false
[15:24:19]W:		 [Docker Run]     at makeError (/workspace/Sites/node_modules/start-server-and-test/node_modules/execa/lib/error.js:60:11)
[15:24:19]W:		 [Docker Run]     at handlePromise (/workspace/Sites/node_modules/start-server-and-test/node_modules/execa/index.js:118:26)
[15:24:19]W:		 [Docker Run]     at processTicksAndRejections (node:internal/process/task_queues:96:5) {
[15:24:19]W:		 [Docker Run]   shortMessage: 'Command failed with exit code 1: yarn test:cypress --reporter teamcity --config video=false,defaultCommandTimeout=10000 --record false',
[15:24:19]W:		 [Docker Run]   command: 'yarn test:cypress --reporter teamcity --config video=false,defaultCommandTimeout=10000 --record false',
[15:24:19]W:		 [Docker Run]   escapedCommand: '"yarn test:cypress --reporter teamcity --config video=false,defaultCommandTimeout=10000 --record false"',
[15:24:19]W:		 [Docker Run]   exitCode: 1,
[15:24:19]W:		 [Docker Run]   signal: undefined,
[15:24:19]W:		 [Docker Run]   signalDescription: undefined,
[15:24:19]W:		 [Docker Run]   stdout: undefined,
[15:24:19]W:		 [Docker Run]   stderr: undefined,
[15:24:19]W:		 [Docker Run]   failed: true,
[15:24:19]W:		 [Docker Run]   timedOut: false,
[15:24:19]W:		 [Docker Run]   isCanceled: false,
[15:24:19]W:		 [Docker Run]   killed: false
[15:24:19]W:		 [Docker Run] }
[15:24:19]W:		 [Docker Run] error Command failed with exit code 1.
[15:24:19] :		 [Docker Run] info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### Desired behavior

Do not throw the SIGSEGV.

### Test code to reproduce

It can happen on any test, at any time, in our 850+ test test suite.

### Cypress Version

9.1.1

### Other

It doesn't seem to happen if we run on Windows, and we have a parallel set of Component tests (about 100) that have yet to show the issue and they're run exactly the same way in the same Docker container, just as component tests.

BlueWinds:
- Closes #19299

### User facing changelog
Silences electron warnings about being unable to connect to dbus. These errors are normal and expected, and do not result in test failures. They are present when running electron inside docker containers 100% of the time.

### Additional details
Chromium (which Electron uses) always makes several attempts to connect to the system dbus. This works fine in most desktop environments, but in a docker container, there is no dbus service and Chromium emits several error lines, similar to these:

```
[1957:0406/160550.146820:ERROR:bus.cc(392)] Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
[1957:0406/160550.147994:ERROR:bus.cc(392)] Failed to connect to the bus: Address does not contain a colon
[1957:0406/160550.147994:ERROR:bus.cc(392)] Failed to connect to the bus: Address does not contain a colon
[1957:0406/160550.147994:ERROR:bus.cc(392)] Failed to connect to the bus: Address does not contain a colon
```

These warnings are absolutely harmless. Failure to connect to dbus means that electron won't be able to access the user's credential wallet (none exists in a docker container) and won't show up in the system tray (again, none exists). Failure to connect is expected and normal here, but users frequently misidentify these errors as the cause of their problems.

On a more technical level, chromium calls [dbus_bus_get()](https://dbus.freedesktop.org/doc/api/html/group__DBusBus.html#ga77ba5250adb84620f16007e1b023cf26), which is provided by libdbus. This call returns an error if dbus cannot be found or connected to for whatever reason. Chromium handles these errors correctly, internally ceasing to attempt to use the service, but it re-prints the errors to stderr while doing so.

Basically: It tries to connect, gets an error, and aborts. All very normal behavior, which is why no one has tried to 'fix' the problem. Because there is no problem.

Users will no longer see errors in the log when running Cypress inside docker. Instead of:

image

they'll get

image

### PR Tasks

### DIFF
--- a/cli/lib/exec/spawn.js
+++ b/cli/lib/exec/spawn.js
@@ -16,7 +16,21 @@ const isXlibOrLibudevRe = /^(?:Xlib|libudev)/
 const isHighSierraWarningRe = /\*\*\* WARNING/
 const isRenderWorkerRe = /\.RenderWorker-/
 
-const GARBAGE_WARNINGS = [isXlibOrLibudevRe, isHighSierraWarningRe, isRenderWorkerRe]
+// Chromium (which Electron uses) always makes several attempts to connect to the system dbus.
+// This works fine in most desktop environments, but in a docker container, there is no dbus service
+// and Chromium emits several error lines, similar to these:
+
+// [1957:0406/160550.146820:ERROR:bus.cc(392)] Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
+// [1957:0406/160550.147994:ERROR:bus.cc(392)] Failed to connect to the bus: Address does not contain a colon
+
+// These warnings are absolutely harmless. Failure to connect to dbus means that electron won't be able to access the user's
+// credential wallet (none exists in a docker container) and won't show up in the system tray (again, none exists).
+// Failure to connect is expected and normal here, but users frequently misidentify these errors as the cause of their problems.
+
+// https://github.com/cypress-io/cypress/issues/19299
+const isDbusWarning = /Failed to connect to the bus:/
+
+const GARBAGE_WARNINGS = [isXlibOrLibudevRe, isHighSierraWarningRe, isRenderWorkerRe, isDbusWarning]
 
 const isGarbageLineWarning = (str) => {
   return _.some(GARBAGE_WARNINGS, (re) => {
--- a/cli/test/lib/exec/spawn_spec.js
+++ b/cli/test/lib/exec/spawn_spec.js
@@ -65,6 +65,9 @@ describe('lib/exec/spawn', function () {
         [46454:0702/140217.292555:ERROR:gles2_cmd_decoder.cc(4439)] [.RenderWorker-0x7f8bc5815a00.GpuRasterization]GL ERROR :GL_INVALID_FRAMEBUFFER_OPERATION : glDrawElements: framebuffer incomplete
         [46454:0702/140217.292584:ERROR:gles2_cmd_decoder.cc(4439)] [.RenderWorker-0x7f8bc5815a00.GpuRasterization]GL ERROR :GL_INVALID_FRAMEBUFFER_OPERATION : glClear: framebuffer incomplete
         [46454:0702/140217.292612:ERROR:gles2_cmd_decoder.cc(4439)] [.RenderWorker-0x7f8bc5815a00.GpuRasterization]GL ERROR :GL_INVALID_FRAMEBUFFER_OPERATION : glDrawElements: framebuffer incomplete'
+
+        [1957:0406/160550.146820:ERROR:bus.cc(392)] Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
+        [1957:0406/160550.147994:ERROR:bus.cc(392)] Failed to connect to the bus: Address does not contain a colon
       `
 
       const lines = _


cli/lib/exec/spawn.js
```diff
@@ -16,7 +16,21 @@ const isXlibOrLibudevRe = /^(?:Xlib|libudev)/
...
+
+// [1957:0406/160550.146820:ERROR:bus.cc(392)] Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
+// [1957:0406/160550.147994:ERROR:bus.cc(392)] Failed to connect to the bus: Address does not contain a colon
+
+// These warnings are absolutely harmless. Failure to connect to dbus means that electron won't be able to access the user's
+// credential wallet (none exists in a docker container) and won't show up in the system tray (again, none exists).
+// Failure to connect is expected and normal here, but users frequently misidentify these errors as the cause of their problems.
+
+// https://github.com/cypress-io/cypress/issues/19299
+const isDbusWarning = /Failed to connect to the bus:/
```
jennifer-shehane, line 31: Do you think this is an adequate Regex? I guess it's highly unlikely someone would have "Failed to connect to the bus:" as some other output that would be stripped. Feels a bit less specific than the other ones stripped though.

BlueWinds: This regex is applied pretty narrowly: to the stdout/stderr from electron itself. This doesn't filter any user-generated messages or things coming from our server. I settled on this based on the chromium source code:

```
  if (!connection_) {
    LOG(ERROR) << "Failed to connect to the bus: "
               << (dbus_error_is_set(error.get()) ? error.message() : "");
    return false;
  }
```
https://chromium.googlesource.com/chromium/src/+/524f7d22d706ccf6264dbe0ea90a5b469f2c3379/dbus/bus.cc#392

I think it's about as specific as is reasonable here - ideally we won't need to update it as we change electron versions, so the 'technical' stuff up front (`[1957:0406/160550.147994:ERROR:bus.cc(392)]`) doesn't seem like a good candidate for inclusion.

flotwig (commented):
@BlueWinds When testing `yarn cypress:run` locally with `yarn docker`, there is a new Chromium stderr that looks like it will mislead folks in a similar way: 

`[371:0406/175035.061253:ERROR:gpu_init.cc(453)] Passthrough is not supported, GL is swiftshader, ANGLE is `

Can reproduce by doing `yarn docker` in the monorepo and running `yarn cypress:run --project packages/example`.

---

Also, when I run the latest prod `cypress open`, I get two log lines:

```
[343988:0406/135427.824393:WARNING:vaapi_wrapper.cc(586)] VAAPI video acceleration not available for swiftshader
[343988:0406/135427.825018:ERROR:gpu_init.cc(453)] Passthrough is not supported, GL is swiftshader, ANGLE is 
```

Sounds like these could all be related to video acceleration? https://stackoverflow.com/questions/67501093/passthrough-is-not-supported-gl-is-disabled

Feels like these are important to address too, @BlueWinds thoughts?

emilyrohrbough:
@BlueWinds I am good with the changes, but I'm wondering if this change will satisfy closing the issue it's linked to? Quite a few people have reported issues with crashes & SIGSEGV errors immediately after seeing these dbus errors and this isn't addressing SIGSEGV errors at all.

BlueWinds:
> @BlueWinds I am good with the changes, but I'm wondering if this change will satisfy closing the issue it's linked to? Quite a few people have reported issues with crashes & SIGSEGV errors immediately after seeing these dbus errors and this isn't addressing SIGSEGV errors at all.

I am 100% sure that these errors are not causing the crashes they're seeing.

The ticket linked to here contains people reporting half a dozen different bugs - but none of them are actually related to dbus. This is just the last output line, so they assume their problems must be related, and put it all in the same ticket. The hope is, that by hiding these misleading lines, they will open new, more focused, more helpful tickets. Or at the very least, they'll clump their error reports in more useful ways (systems, browsers, versions).

It won't address the crashes people are seeing - but a bit 'catch all' ticket ensures that we'll never be able to help them at all.

BlueWinds:
> @BlueWinds When testing `yarn cypress:run` locally with `yarn docker`, there is a new Chromium stderr that looks like it will mislead folks in a similar way:
> 
> `[371:0406/175035.061253:ERROR:gpu_init.cc(453)] Passthrough is not supported, GL is swiftshader, ANGLE is `
> 
> Can reproduce by doing `yarn docker` in the monorepo and running `yarn cypress:run --project packages/example`.
>
> Also, when I run the latest prod `cypress open`, I get two log lines:
> 
> ```
> [343988:0406/135427.824393:WARNING:vaapi_wrapper.cc(586)] VAAPI video acceleration not available for swiftshader
> [343988:0406/135427.825018:ERROR:gpu_init.cc(453)] Passthrough is not supported, GL is swiftshader, ANGLE is 
> ```
> 
> Sounds like these could all be related to video acceleration? https://stackoverflow.com/questions/67501093/passthrough-is-not-supported-gl-is-disabled
> 
> Feels like these are important to address too, @BlueWinds thoughts?

Yes, I've seen these as well. I don't want to hide them because I don't understand them - I'd want to do another deep-dive into those log lines to see what they mean, if we can fix them, if they're genuinely harmless, etc. before deciding what to do about them. So far I've only done this with the dbus errors, not the GL ones.

Would prefer to do that in a separate effort though.

emilyrohrbough:
100% agree with your thoughts and reasoning here. Just pointing out if we link this PR as closing this issue, our bot will indicate the issue has been which will be mis-leading since your intentions are close this issue with too many different issues reported and ask people to open new issues with details after upgrading with these error messages filtered out. Might be worth linking this issue, commenting your intentions on the thread when this merges and locking the issue to push towards new issue creation.

emilyrohrbough (approved):

cubanx:
So I originally posted that message. Sorry it caused all the confusion, my original bug report was about `SIGSEV` and not about the issue this is fixing, I just pointed it out as I thought it *might* be related, but it turns out it isn't :)

You might want to point out to folks that I solved the `SIGSEV` issue by switching to Chrome from Electron. Our runs have been rock solid (well, except when we break them ;)) since.

Thanks for deep diving and hiding the errors, and tell @chrisbreiding I said 👋🏼

BlueWinds:
> So I originally posted that message. Sorry it caused all the confusion, my original bug report was about `SIGSEV` and not about the issue this is fixing, I just pointed it out as I thought it _might_ be related, but it turns out it isn't :)
> 
> You might want to point out to folks that I solved the `SIGSEV` issue by switching to Chrome from Electron. Our runs have been rock solid (well, except when we break them ;)) since.
> 
> Thanks for deep diving and hiding the errors, and tell @chrisbreiding I said 👋🏼

Yeah, absolutely nothing wrong with your original report! It just got tangled up in a dogpile of other issues, and is now the top (and almost only) result when people google the dbus error, which leads to them to pile in there and connect the two (error + SIGSEV).

flotwig (approved):
> Yes, I've seen these as well. I don't want to hide them because I don't understand them - I'd want to do another deep-dive into those log lines to see what they mean, if we can fix them, if they're genuinely harmless, etc. before deciding what to do about them. So far I've only done this with the dbus errors, not the GL ones.
> 
> Would prefer to do that in a separate effort though.

Totally understand. Is there a follow-up issue or task created for this? I don't want this to get lost, if we don't follow up with that next fix, users will be back in the same confusing place.

BlueWinds:
> > Yes, I've seen these as well. I don't want to hide them because I don't understand them - I'd want to do another deep-dive into those log lines to see what they mean, if we can fix them, if they're genuinely harmless, etc. before deciding what to do about them. So far I've only done this with the dbus errors, not the GL ones.
> > Would prefer to do that in a separate effort though.
> 
> Totally understand. Is there a follow-up issue or task created for this? I don't want this to get lost, if we don't follow up with that next fix, users will be back in the same confusing place.

https://github.com/cypress-io/cypress/issues/18947. It is marked as 'solved' in the title, but it's an issue specifically related to those errors which users are also finding and identifying as the source of their problems. Not quite as popular as the dbus one, but I'll bring it up for prioritization next.

Issue #2102, opened 2018-07-05
Screenshots taken during 'cypress run' do not respect the viewport size set


egucciar:
### Current behavior:

Observe these two images:

Headless / Electron: 
![e2e spec -- goes through profile flow -- 01 - select old home size](https://user-images.githubusercontent.com/2848185/42338970-93c830d2-8059-11e8-85f2-e8b4a7823568.png)

Headed / Chrome:
![e2e spec -- goes through profile flow -- 01 - select old home size](https://user-images.githubusercontent.com/2848185/42338985-a0babc06-8059-11e8-8139-2195598027d8.png)

Notice the spacing to the Left of our Logo is much different than the spacing to left of the Logo in headed and headless, indicating that our 1500px viewport setting was not respected in headless.

### Desired behavior:

Ideally, the Viewport width should be the same for both browsers, headless and headed.

### Steps to reproduce:

Try using a Website with a viewport of 1500 pixels wide and confirm if the viewport taken in headless vs headed is the same

### Versions

Cypress 3.0.2 and MacOS

### Motivation

I am working towards pegging down the Visual Diffing workflow for our company. Truth is, having comparable screenshots in headless vs headed runs is a desirable feature. When it comes to scaling and cropping, there is no issue on my side to do some POC work to scale/crop images and set thresholds such that differences in the CI / Headless enviroment can be accounted for and ignored. However I cannot ignore the fact that headless screenshots do not honor the viewport width I explicitly set in the Cypress.json. Ideally, if the viewport is 1500px wide headed, the headless screenshots should reflect the same.

Bkucera:
Possibly related to #1857

egucciar:
@Bkucera I agree that this needs more information. I tried to output the "innerWidth" of window in both scenarios and Headless also said 1500px. Maybe it's just a difference between how High DPI screens render UIs verses the headlesS? maybe this is not a bug then, it just is confusing to me why there would be such different margins in one vs the other.

brian-mann:
I believe the reason this is happening is because in headless mode we set XVFB to render at 1280x720.

Normally Cypress scales the screen so that running tests makes no difference to the screen size. However when taking screenshots, we no longer scale the image, which forces it to display in full width + height.

Think of the 1280x720 as the literal screen size. If you've set your viewportWidth to be 1500px then it physically cannot fit inside of 1280. Your application is responsive so that it all "works" without a scrollbar, but its not the same image anymore.

[...]

PR #20951, opened 2022-04-06
feat: swap the #__cy_root id selector to become data-cy-root for component mounting

JessicaSachs:
- Closes [UNIFY-1405](https://cypress-io.atlassian.net/browse/UNIFY-1405)

### User facing changelog

For component tests, we're changing the default selector for where the component should attach onto.

#### Currently: using an ID for the mounting point
The html scaffolded for component testing requires a mounting point. This is currently a **`ROOT_ID`**. The downside to having an **`id`** is that there can only be one. For users whose stylesheets use a root node with id to segment their global app styles, this means that you'll need to create an inner-div or other DOM fixture so that the stylesheet rules apply.

```html
<div id="__cy_root">
  <!--
    your component here, but unless you refactor your styles...
    it might render in Times New Roman
  -->
</div>
```

#### After: using a **`data-*`** attribute for the mounting point
Now, we expose a **`ROOT_SELECTOR`** which is a **`data-*`** attribute and so now the user can assign an **`id`** instead of refactoring their stylesheets.

```html
<!-- support/component-index.html -->

<div data-cy-root id="your-app">
  <!-- your component here, but now with your global app styles! -->
</div>
```


lmiller1990 (approved):
Some tests are skipped for known reasons

- vuecli5vue3: blocked by https://cypress-io.atlassian.net/browse/UNIFY-1565
- vue 2 tests: blocked by https://github.com/cypress-io/cypress/pull/2102

Fine with this for now, let's ship it.

system-tests/projects/vueclivue2-configured/package.json
```diff
@@ -10,8 +10,7 @@
     "vue": "^2.6.11"
   },
   "devDependencies": {
-    "@cypress/vue": "^2.0.0",
-    "@cypress/webpack-dev-server": "^1.0.0",
+    "@cypress/webpack-dev-server": "file:../../../npm/webpack-dev-server",
```
BlueWinds, line 13: Not `-fresh` for this and those below, like `system-tests/projects/vuecli5vue3-configured/package.json` above?

JessicaSachs: Those didn't exist at the time and we're going to convert everything to object API anyway in a half-second.

tgriesser (approved):

PR #20967, opened 2022-04-07
chore(build): check max path length during Windows build

flotwig:
- Partially addresses UNIFY-1449

### User facing changelog

n/a

### Additional details

- Uses a heuristic to try and prevent shipping paths that will exceed the Windows max path length. It's not going to catch everything, but it should prevent egregious violations of the max path length.
- See code and comments for details.

### PR Tasks


flotwig:
@marktnoonan curious your thoughts on this approach - it's still not testing an extra-long pre-release cache directory path, but it seems like this should help block this type of regression from reaching prod.

marktnoonan:
I think you're right. I also like the error message being clear that about what is happening and what to do about it. This seems like a good guard to have in place for now.

scripts/binary/build.ts
```diff
@@ -33,6 +37,38 @@ interface BuildCypressAppOpts {
...
+  const allRelPaths = (await globAsync('**/*', { cwd: distDir, absolute: true }))
+  .map((p) => p.slice(distDir.length))
+
+  if (!allRelPaths.length) throw new Error('No binary paths found in checkMaxPathLength')
+
+  const violations = allRelPaths.filter((p) => p.length > maxRelPathLength)
+
+  if (violations.length) {
+    throw new Error([
+      `${violations.length} paths in the built binary were too long for Windows. Either hoist these files or remove them from the build.`,
```
emilyrohrbough, line 64: Does this place restrictions on the binary path length user should provide via the ENV override or this is specific to only building the binary?

flotwig: Only to building the binary, to prevent us from releasing a .zip with the most common broken scenario.

To wrap up UNIFY-1449 I intend to add a better error to the CLI when this kind of failure to unzip occurs with instructions on how to mitigate.

emilyrohrbough (approved):

ryanthemanuel (approved):

PR #20971, opened 2022-04-07
chore: [Multi-domain] Fix uncaught error

mjhenkes:
- Closes 

### User facing changelog

n/a

### Additional details

Quick fix for an uncaught error.

In the event that the `run:origin:fn` event fails a secondary error was thrown because we weren't releasing the request.

I didn't add a test because it's hard to test a negative, This wasn't caught in our existing tests because it's a second error and we already handled the first error.

### PR Tasks

### DIFF
--- a/packages/driver/src/cy/multi-domain/index.ts
+++ b/packages/driver/src/cy/multi-domain/index.ts
@@ -129,7 +129,7 @@ export function addCommands (Commands, Cypress: Cypress.Cypress, cy: Cypress.cy,
           resolve(unserializableSubjectType ? createUnserializableSubjectProxy(unserializableSubjectType) : subject)
         }
 
-        const _reject = (err, cleanupOptions = {}) => {
+        const _reject = (err, cleanupOptions: {readyForOriginFailed?: boolean} = {}) => {
           cleanup(cleanupOptions)
           log?.error(err)
           reject(err)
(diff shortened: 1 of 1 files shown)


packages/driver/src/cy/multi-domain/index.ts
```diff
@@ -231,6 +231,9 @@ export function addCommands (Commands, Cypress: Cypress.Cypress, cy: Cypress.cy,
                 logCounter: LogUtils.getCounter(),
               })
             } catch (err: any) {
+              // Release the request if 'run:origin:fn' fails
+              Cypress.backend('ready:for:origin', { originPolicy: location.originPolicy })
```
mschile, line 235: I wonder if we should consider this a failure scenario and pass `failed: true`? If so, we'd need to conditionally call `cross:origin:finished`.

mschile (changes requested):

packages/driver/src/cy/multi-domain/index.ts
mjhenkes: That seems more correct. I was missing that you shouldn't call `cross:origin:finished` before

AtofStryker (approved):

packages/driver/src/cy/multi-domain/index.ts
```diff
@@ -126,8 +129,8 @@ export function addCommands (Commands, Cypress: Cypress.Cypress, cy: Cypress.cy,
           resolve(unserializableSubjectType ? createUnserializableSubjectProxy(unserializableSubjectType) : subject)
         }
 
-        const _reject = (err) => {
-          cleanup()
+        const _reject = (err, cleanupOptions = {}) => {
```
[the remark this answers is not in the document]
mjhenkes: Done!

chrisbreiding (approved):

mschile (approved):

PR #20997, opened 2022-04-08
refactor: change git info handling to happen eagerly / asynchronously

tgriesser:
- Part of UNIFY-1206

### User facing changelog
Improves loading of the test runner, by moving git info loading from blocking the GraphQL resolution, to loading when the project loads & refreshing when specs are updated.

### Additional details
Previously, we fetched the git branch & spec list git info as the GraphQL query resolved. This can be a slow operation, particularly on Big Sur, which suffers from an electron bug causing slow child_process execution. Regardless of the bug, we shouldn't block the rendering of the UI on the fetching of git info.

Moves the `GitDataSource` to be more self-contained, keeps a map of resolved spec `gitInfo` which can be used to quickly respond to the `gitInfo` field on the `Spec` type. This info is refreshed when the specs list updates, and we put a watcher on the `.git/HEAD` file to know when the branch has changed. We also refresh the git info every 60 seconds while in open mode otherwise, to make sure the UI stays up to date.

### PR Tasks


packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -373,6 +377,19 @@ export class ProjectLifecycleManager {
 
     this.ctx.update((s) => {
       s.currentProject = projectRoot
+      s.currentProjectGitInfo?.destroy()
+      s.currentProjectGitInfo = new GitDataSource({
```
lmiller1990, line 381: I wonder if `DataSource` is the right suffix here, now we've changed it's lifecycle and where it's assigned? 

It feels like this is a little different to the other `Sources` we have now - all the others are created and cached on the `ctx.sources`, but not this one. What do you think?

packages/data-context/test/unit/sources/GitDataSource.spec.ts
```diff
@@ -40,10 +63,18 @@ describe('GitDataSource', () => {
...
+
+    await dfd.promise
+
     const [created, unmodified, modified] = await Promise.all([
-      ctx.git.gitInfo(fooSpec),
-      ctx.git.gitInfo(aRecordSpec),
-      ctx.git.gitInfo(xhrSpec),
+      gitInfo.gitInfoFor(fooSpec),
+      gitInfo.gitInfoFor(aRecordSpec),
+      gitInfo.gitInfoFor(xhrSpec),
```
lmiller1990, line 77: I'm glad these tests were useful during this refactor - you hardly changed them which is a good sign, I think in general we are doing a good job of testing behaviours, not implementation details.

lmiller1990 (approved):
Two questions, not really blockers.

Also just to be on the safe side, I tested this out by hand on Mac, Ubuntu AND Windows 11, and they all work fine. ✅

packages/data-context/src/data/ProjectLifecycleManager.ts
tgriesser: I don't think we need to change it right now, in that the `Source` is indicating it's "Reading" data, and `Action` indicates it's writing data. This is sort of the direction that we had discussed wanting to move, having clearer contracts for each class where we're injecting the minimum amount of state into the class for better isolation / testability.

tbiethman (approved):

packages/data-context/src/actions/DataEmitterActions.ts
```diff
@@ -109,7 +123,7 @@ export class DataEmitterActions extends DataEmitterEvents {
         if (!hasSentInitial && sendInitial) {
           hasSentInitial = true
 
-          return { done: false, value: {} }
+          return { done: false, value: undefined }
```
ryanthemanuel, line 126: +1 here. I think I noticed this at one point, but forgot to ask about it/change it.  It was weird to me that we did {} in this scenario, but undefined in others

packages/data-context/src/sources/BrowserDataSource.ts
```diff
@@ -1,15 +1,19 @@
...
-} catch {
-  // Powershell is unavailable
+// Only need to worry about checking for PowerShell in windows,
+// doing it asynchronously so to not block startup
+if (os.platform() === 'win32') {
+  execa(`[void] ''`, { shell: 'powershell' }).then(() => {
+    isPowerShellAvailable = true
+  }).catch(() => {
+    // Powershell is unavailable
+  })
```
tgriesser, line 16: @ryanthemanuel just wanted to confirm this change was ok / there wasn't a reason this needed to be done synchronously during script loading?

packages/data-context/src/sources/BrowserDataSource.ts
```diff
@@ -1,15 +1,19 @@
...
 
-try {
-  execSync(`[void] ''`, { shell: 'powershell' })
-  isPowerShellAvailable = true
-} catch {
-  // Powershell is unavailable
+// Only need to worry about checking for PowerShell in windows,
+// doing it asynchronously so to not block startup
+if (os.platform() === 'win32') {
+  execa(`[void] ''`, { shell: 'powershell' }).then(() => {
```
ryanthemanuel, line 12: Good call here.

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -373,6 +377,19 @@ export class ProjectLifecycleManager {
...
+        isRunMode: this.ctx.isRunMode,
+        projectRoot,
+        onError: this.ctx.onError,
+        onBranchChange: () => {
+          this.ctx.emitter.branchChange()
+        },
+        onGitInfoChange: (specPaths) => {
+          this.ctx.emitter.gitInfoChange(specPaths)
+        },
```
ryanthemanuel, line 390: Can we do?

```suggestion
        onGitInfoChange: this.ctx.emitter.gitInfoChange,
```

tgriesser: We'd need to add a bind or change `gitInfoChange (` to `gitInfoChange = ()` where it's defined.

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -373,6 +377,19 @@ export class ProjectLifecycleManager {
...
     this.ctx.update((s) => {
       s.currentProject = projectRoot
+      s.currentProjectGitInfo?.destroy()
+      s.currentProjectGitInfo = new GitDataSource({
+        isRunMode: this.ctx.isRunMode,
+        projectRoot,
+        onError: this.ctx.onError,
+        onBranchChange: () => {
+          this.ctx.emitter.branchChange()
+        },
```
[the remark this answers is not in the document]
tgriesser: Same as below, we'd need to change to bind the fn `this.ctx.emitter.branchChange.bind(this.ctx.emitter)`

packages/data-context/src/sources/BrowserDataSource.ts
ryanthemanuel: I believe this should be ok. I think the weird scenario would be if this takes longer to resolve than when we get to the OpenBrowserList page:

https://github.com/cypress-io/cypress/blob/10.0-release/packages/launchpad/src/setup/OpenBrowserList.vue#L113

i guess we could guard around that by ensuring to wait for the result here:

https://github.com/cypress-io/cypress/pull/20997/files#diff-ccfea29f817d69316d07e1e2ad570f26a2125e956e357df65244757a12331cd3R75

ryanthemanuel (approved):

Issue #16130, opened 2021-04-22
Make a 1 second (1000ms) delay to show Cypress Command Log configurable globally and for commands


t00:
Currently Cypress Command Log is shown if the command did not finish after 1 second.

Most DOM commands should execute within this time but remote requests or DOM objects depending on such requests often exceed this limit and Command Log shows up, slowing down tests considerably.

There are several timeouts available for different command types and several options for requests but none allows adjusting this 1 second show log trigger..

Reasoning behind configurable show Command Log delay is that often tests require resetting the environment/database to isolate tests from each other, in line with cypress recommendations, but this process takes time and leads to slower tests overall as needs to be performed at the very beginning of each test.

My proposal is pretty simple, in the order of importance as I can see it:

1. Add logTimeout option to the cypress.json options with a default value of 1000 (ms).
2. Add cy.request() option logTimeout which when set will override cypress.json logTimeout allowing to specify a custom delay after which Command Log will be shown for this request
3. Add logTimeout to all other functions with already have a timeout option (cy.get(), cy.click() etc) allowing to handle slow dom-triggered requests.

Having only globally configurable logTimeout (1) should be enough in most cases and since it is a global setting there is a high probability it will be easy to implement so please consider it a priority even if (2) and (3) would be a nice addition.

jennifer-shehane:
>Command Log shows up, slowing down tests considerably.

You want this feature to increase performance because you believe the rendering of the Command Log is slowing down your tests? Just want to be clear on the motivation for hiding the Command Log. 

There's an option to hide the command log altogether documented here: https://docs.cypress.io/guides/references/troubleshooting#Disable-the-Command-Log

Essentially I would say this is similar to https://github.com/cypress-io/cypress/issues/6270, allowing more flexibility in collapsing/expanding commands. But if there were no performance issues you probably wouldn't even request this feature, correct?

t00:
Rendering of the Command Log surely slows down tests, especially if test takes a few minutes to run - as I understand adding more elements to the log slows down rendering of both log and page being tested - a design choice making this unavoidable.

The CYPRESS_NO_COMMAND_LOG=1 option is great to improve overall speed but does not help when tests fail as test recording does not contain essential information of what operations led to the error.

Indeed keepCommandLogSectionsClosed would help to replace CYPRESS_NO_COMMAND_LOG=1 option and performance improvement with Command Log is considerable as you know already, but it would still make post-analysis of a failed test more difficult.

[...]

Issue #15880, opened 2021-04-08
Reporter slows down due to creating an unreasonable number of DOM nodes


agg23:
### Current behavior
Cypress slows down as more entries are added to the Reporter's Command Log. This results in exponential growth of total test time whenever the Command Log is visible (though luckily it's a small exponent). Further investigation showed the `Command` component renders 27+ DOM nodes (there are also React Portals, so even more than 27). [Google Lighthouse](https://web.dev/dom-size/) recommends that the total count of DOM elements should remain below 1,500. Larger tests can easily reach 15,000+ DOM nodes, which is the likely source of the slowdown.

### Desired behavior
Cypress does not slow down based on the number of rendered `Command`s. Additionally, the DOM footprint of the application should be minimized.

jennifer-shehane:
Yes, we've seen this reported numerous times from users. @chrisbreiding spent A LOT of time trying to improve the rendering if you want to discuss with him any options he's already explored. 

Some performance issues involving many DOM nodes with reproducible examples:

- https://github.com/cypress-io/cypress/issues/5987
- https://github.com/cypress-io/cypress/issues/6967
- https://github.com/cypress-io/cypress/issues/6783
- https://github.com/cypress-io/cypress/issues/19906

BlueWinds:
Just doing some cleanup as we begin looking into this internally; closing in favor of https://github.com/cypress-io/cypress/issues/4164, which contains links to a bunch of different examples.

Definitely a real issue, and closing doesn't mean we're not paying attention; just consolidating a bit as we start looking into this more seriously.

Issue #8658, opened 2020-09-24
Test slows down when commands are executed multiple times


Konstruktour:
### Current behavior
When executing teststeps (commands) in a loop, the duration increases with each iteration. 
Eg The evolving durations of a testrun with about 500 iterations:
![image](https://user-images.githubusercontent.com/7323845/94137272-761db880-fe66-11ea-950c-026ce45f5559.png)

### Desired behavior
The duration should be constant within each iteration.

### Test code to reproduce
loop.spec.ts
```javascript
describe('Performance - loop', () => {
    it('visit dummy site', () => {
        cy.visit('http://127.0.0.1:5500/index.html');
        for (let index = 0; index < 100; index++) {
            let start: number;
            cy.then(() => (start = Date.now()));
            cy.get('#testinput').clear().type('hello world');
            cy.then(() => {
                const end: number = Date.now();
                const duration: number = end - start;
                cy.task('log', `duration for ${index}: ${duration}`);
            });
        }
    });
});
```
index.html
```html
<!DOCTYPE html>
<html lang="en">
    <head> </head>
    <body>
        <input id="testinput" />
    </body>
</html>

```

### Versions
Cypress 5.2.0
Windows 10
Electron/Chromium (headless)

jennifer-shehane:
I can reproduce this in 6.1.0 running `cypress run` locally on my Mac on Chrome 87. 

`index.html`

```html
<html>
<body>
  <input id="testinput" />
</body>
</html>
```

`spec.js`

```js
it('visit dummy site', () => {
  cy.visit('index.html');
  for (let i = 0; i < 300; i++) {
    let start;
    cy.then(() => (start = Date.now()));
    cy.get('#testinput').clear().type('hello world');
    cy.then(() => {
      const end = Date.now();
      const duration = end - start;
      cy.task('log', `duration for ${i}: ${duration}`);
    });
  }
});
```

`plugins/index.js`

```js
module.exports = (on, config) => {
  on('task', {
    log(message) {
      console.log(message)

      return null
    }
  })
}
```

```
duration for 0: 390
duration for 1: 279
duration for 2: 283
duration for 3: 279
duration for 4: 285
duration for 5: 285
duration for 6: 274
duration for 7: 281
duration for 8: 289
duration for 9: 296
duration for 10: 296
duration for 11: 307
duration for 12: 304
duration for 13: 310
duration for 14: 312
duration for 15: 310
duration for 16: 312
duration for 17: 308
[...]

Issue #6388, opened 2020-02-10
Cypress 4.0.1 - Electron is 1.5x slower in a test spec


sandrina-p:
# Current behavior:

In our project, we are updating Cypress from 3.5 to 4.0.1. See PR https://github.com/okTurtles/group-income-simple/pull/829

The same test spec "Large Group" takes 2:29m to run instead of 1:36m. The other specs take the same time. See [run 284 - @3.5](https://dashboard.cypress.io/projects/q6whky/runs/284/specs) vs [run 285 - @4.01](https://dashboard.cypress.io/projects/q6whky/runs/285/specs). 

### Desired behavior:

All specs should take the same time.

### Test code to reproduce

1. Clone our project and [checkout the PR](https://github.com/okTurtles/group-income-simple/pull/829) (or master to see v 3.5.0)
2. Install dependencies `npm install`
3. Run the project `grunt dev`
4. In another terminal run cypress `npm run cy:open`
5. Run the test step "Group Large" on Electron.

The test in the spec taking too long is located at `test/cypress/integration/group-large.spec.js`. See [file here](https://github.com/okTurtles/group-income-simple/blob/820cdb2511a4c2c0016426a23cc1b7ce05c64d66/test/cypress/integration/group-large.spec.js#L22)

```js
  it(`A group with ${groupLength} members shows correctly the pledging month overview widget`, () => {
    for (let i = 2; i <= groupLength; i++) {
      // The member joins the group through an invitation...
      cy.giAcceptGroupInvite(invitationLinks.anyone, {
        username: `user${i}-${userId}`,
        groupName,
        bypassUI: true,
        actionBeforeLogout: () => {
          if (i > 3) {
            cy.getByDT('contributionsLink').click()
          }
          cy.giAddRandomIncome()
        }
      })
    }

    cy.giLogin(`user1-${userId}`)
    cy.giAddRandomIncome()
    cy.get('.graph-bar')
      .should('have.length', groupLength)
    cy.giLogout()
  })
```

If anything else is missing, please let me know.

### Versions

- Cypress 3.5 and 4.0.1
- Browser Electron 72 and 78 (maybe that's the reason?)
- OS Ubuntu 14.04

Olgagr:
We have the same issue. However, we noticed the speed tests degradation from version 3.8.2 to 3.8.3 (15 minutes slower). We thought that maybe the speed improves in version 4 but it is not.

**Versions**
* Cypress 3.8.2 to 4.0.1
[...]

PR #21007, opened 2022-04-08
fix: Fix perf issue with command log

andrewmtam:
- Closes (I tagged a bunch of performance issues that I think this will address) 
    - https://github.com/cypress-io/cypress/issues/16130
    - https://github.com/cypress-io/cypress/issues/8658
    - https://github.com/cypress-io/cypress/issues/6270
    - https://github.com/cypress-io/cypress/issues/6388
    - https://github.com/cypress-io/cypress/issues/15880
    - https://github.com/cypress-io/cypress/issues/5987
### User facing changelog

Added a new configuration option, `numCommandsToShow`, which can be used to yield performance improvements for long running tests.

This will limit the number to commands that show in the Command Log (the left sidebar) of a cypress test, instead of allowing it to grow indefinitely.

### Additional details

We already have a parameter, `CYPRESS_NO_COMMAND_LOG`, which hides the Command Log altogether. However, this is not sufficient for both debugging and performance.  `numCommandsToShow` seeks to provide a solution that satisfies both performance and debugging concerns.

While this does restrict the user from clicking on snapshots past the specified `numCommandsToShow` value (and only if they opted into this), I don't think this is typically needed (when debugging, we only need the last couple of commands around when the test failed)

Related docs PR:
https://github.com/cypress-io/cypress-documentation/pull/4431

I believe this simple change / configuration can yield huge huge performance gains for users, with little to no risk of regression at all.

I ran a cypress test w/o this change, and repeated the same series of steps in a test 1000 times (I just took screenshots because the video was too big, and I also didn't bother waiting to get to 1000).

The test was to:
* Go to a public todo page
* Add a todo item
* Delete the todo item
* Repeat

From the screenshots, we can see that this step started off taking 400ms, and very quickly we found ourselves at the 1200ms mark, a 3x increment from when the test started, and after only 100 iterations.

#### Before applying this PR
Screen Shot 2022-04-08 at 6 09 32 PM

Screen Shot 2022-04-08 at 6 10 44 PM

#### After applying this PR (using a `numCommandsToShow` of 25)

Next, I applied this PR, and adjusted the new configuration option, numCommandsToShow,  to 25.

For the duration of the test, the duration of the same sequence of actions did not linearly increase with the number of commands getting run.  Also, note that at the 100 iteration mark, this test took /50%/ of the time as the previous test, 42s vs 87s.

Screen Shot 2022-04-08 at 6 13 36 PM

Screen Shot 2022-04-08 at 6 14 07 PM

### PR Tasks

- [x] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?


cli/types/cypress.d.ts
```diff
@@ -2632,6 +2632,11 @@ declare namespace Cypress {
      * @default 50
      */
     numTestsKeptInMemory: number
+    /**
+     * The number of tests to show in the command log. Reduce this number if you are experiencing slowness with long running tests
```
rachelruderman, line 2636:
```suggestion
     * The number of commands to show in the command log. Reduce this number if you are experiencing slowness with long running tests
```

cli/schema/cypress.schema.json
```diff
@@ -30,6 +30,11 @@
           "default": 50,
           "description": "The number of tests for which snapshots and command data are kept in memory. Reduce this number if you are experiencing high memory consumption in your browser during a test run."
         },
+        "numCommandsToShow": {
+          "type": "number",
+          "default": null,
+          "description": "The number of tests to show in the command log. Reduce this number if you are experiencing slowness with long running tests"
```
andrewmtam, line 36:
```suggestion
          "description": "The number of commands to show in the command log. Reduce this number if you are experiencing slowness with long running tests"
```

cli/types/cypress.d.ts
andrewmtam, line 2636:
```suggestion
     * The number of commands to show in the command log. Reduce this number if you are experiencing slowness with long running tests
```

jennifer-shehane:
@andrewmtam We appreciate the work put into this PR, however exposing a configuration option that would limit the number of commands shown is not a feature that we want to add. 

We would prefer a solution that addresses the performance issues without degrading the experience of using the entirety of the Cypress Test Runner at the time it is run. Not being able to see x commands degrades the experience of the runner significantly as information is essentially lost from that run. 

Ideally a rendering solution that would show the relevant commands and relevant data upon scroll. So far our investigations into a rendering solution like this that would improve performance have not given us any clear solutions.

andrewmtam:
Thanks @jennifer-shehane , appreciate the feedback!  I've got two follow up questions if you've got a moment:

1. As we already have a `CYPRESS_NO_COMMAND_LOG={1|0}` environment variable which does something similar, how do you feel about adding a `CYPRESS_LIMIT_COMMAND_LOG={number}` environment variable (instead of it being a configuration option). I believe this environment variable could provide more benefits over the one that hides the command log altogether, as it still surfaces information surrounding failures, /and/ it speeds up performance.  I know this just shifts your argument about:

> We would prefer a solution that addresses the performance issues without degrading the experience of using the entirety of the Cypress Test Runner at the time it is run 

out of a configuration option and into an environment variable, but seeing that we already have an environment variable that hides the command log altogether, I just wanted to double check whether or not doing the same thing for limiting the command log is also off the table.  My main push for this point is that, if consumers set `CYPRESS_NO_COMMAND_LOG` and find that there is a significant performance boost, there is not really a next step for them (while it fixes their perf issues, it cannot be run in production because it does not surface any error messaging), and so providing an alternate `CYPRESS_LIMIT_COMMAND_LOG` may be a way to bridge that gap ([some requests here regarding that](https://github.com/cypress-io/cypress/issues/6270)). 

2. If that doesn't sound good, another alternative could be to look into virtualization of some sort (since we know the command log can grow indefinitely, it sounds like we'd want to curb that, as opposed to continuing to render everything, but faster).  If that's an approach you are aligned with, I'd be happy to poke around there next time I have a few cycles

andrewmtam:
Ooo one more idea came to mind:

3. By default, limit the command log to show X items (lets say 25), but surface "Show more" and/or "Show all" button at the top of the command list for that section that a user can interact with if they need to see more commands. This sounds like it could be a good happy medium:
* It provides a nice performance boost out of the box
* If users need to see more commands, they can opt-in by clicking "Show more" within the same instance of the test run

Something kinda like this:
Screen Shot 2022-04-11 at 11 25 11 AM

Clicking "Show More" each time could increase the amount shown by 25. So on first click, 50 commands would show, then 75, 100, etc etc.  But if the user really does want to see everything, they just click "Show All", and it'll behave as-if th command log was never limited

andrewmtam:
Hey @jennifer-shehane , I wanted to give one more go at this and put up a PR here:

https://github.com/cypress-io/cypress/pull/21036

This does not implement the ideal solution you hinted at, but it does solve this problem:
> Not being able to see x commands degrades the experience of the runner significantly as information is essentially lost from that run.

by providing a way to surface the entire command log list and allow users to click through snapshots from the whole test run. Curious to hear your thoughts on this approach if you've got a couple minutes!

PR #21011, opened 2022-04-09
fix: Plugin error when sending on disconnected IPC channel

### LINKED ISSUE #21010: Interrupted `file:preprocessor` child process with disconnected IPC channel during `preprocessor:close` causes unclear plugin error
### Current behavior

When interrupting the Cypress process with `SIGINT` during the execution of a test the following error is displayed:

```bash
We stopped running your tests because a plugin crashed.

Your pluginsFile threw an error from: /home/tom/cypress-test/src/plugin.ts

Error [ERR_IPC_CHANNEL_CLOSED]: Channel closed
    at new NodeError (node:internal/errors:370:5)
    at ChildProcess.target.send (node:internal/child_process:724:16)
    at Object.send (/home/tom/.cache/Cypress/9.5.3/Cypress/resources/app/packages/server/lib/plugins/util.js:32:25)
    at EventEmitter.<anonymous> (/home/tom/.cache/Cypress/9.5.3/Cypress/resources/app/packages/server/lib/plugins/preprocessor.js:45:16)
    at EventEmitter.emit (node:events:406:35)
    at Object.removeFile (/home/tom/.cache/Cypress/9.5.3/Cypress/resources/app/packages/server/lib/plugins/preprocessor.js:119:17)
    at Object.options.onBrowserClose (/home/tom/.cache/Cypress/9.5.3/Cypress/resources/app/packages/server/lib/open_project.js:135:40)
    at EventEmitter.<anonymous> (/home/tom/.cache/Cypress/9.5.3/Cypress/resources/app/packages/server/lib/browsers/index.js:212:19)
    at Object.onceWrapper (node:events:513:28)
    at EventEmitter.emit (node:events:394:28)
    at BrowserWindow.<anonymous> (/home/tom/.cache/Cypress/9.5.3/Cypress/resources/app/packages/server/lib/browsers/electron.js:431:25)
    at Object.onceWrapper (node:events:514:26)
    at BrowserWindow.emit (node:events:406:35)
```

Based on the text above the stack trace I would assume I made a mistake in my `plugin.ts` `pluginsFile`, however, that is actually empty. Without a custom `pluginFile`, the error will default to the 'standard' Cypress plugin file (`plugin.js`).

### Desired behavior

I would expect that Cypress checks whether or not the `file:preprocessor` child process is still connected or not terminated before trying to send something on its IPC channel. If disconnected or terminated, Cypress should not try to send something to the child process.

In other words, I would expect that I do not get an error.

### Test code to reproduce

`package.json`:

```json
{
  "name": "cypress-interrupt-test",
  "version": "1.0.0",
  "main": "index.ts",
  "scripts": {
    "start": "node --loader ts-node/esm index.ts"
  },
  "devDependencies": {
    "cypress": "^9.5.3",
    "ts-node": "^10.7.0",
    "typescript": "^4.6.3"
  }
}
```

`index.ts`:

```typescript
import cypress from "cypress";

cypress.run({
    config: {
        baseUrl: "http://localhost:2000",
        supportFile: "src/overwrite.ts",
        pluginsFile: "src/plugin.ts",
        integrationFolder: "cypress/integration",
        testFiles: ["**/*.*"],
    },
    spec: null,
    quiet: true,
});
```

`src/overwrite.ts`:

```typescript
// is empty
```

`src/plugin.ts`:

```typescript
/**
 * @type {Cypress.PluginConfig}
 */
module.exports = (on, config) => {}
```

`cypress/integration/interrupt.spec.ts`:

```typescript
describe('an elaborate test', () => {
    it('can wait', () => {
        cy.wait(10000); // wait to allow us to interrupt the Cypress process
    });
});
```

The SUT is just a simple web server with two pages, `/` containing a link to `/page2.html`

Execute `npm start` or `node --loader ts-node/esm index.ts`.

### Cypress Version

9.5.3

### Other

Tested on Ubuntu 20.04 and 21.10, with node versions v16.5.0, v17.4.0, and v17.9.0. Also tested with Cypress 9.5.0.

The following check is performed before trying to send something, however, it purely checks whether or not the process is killed. The process might already have been disconnected at this point (or even terminated).
https://github.com/cypress-io/cypress/blob/933e0eb3d427c5e1dac0d6c34d0fb9c01ab30bf7/packages/server/lib/plugins/util.js#L28-L30

`aProcess` looks like this when it passes the `if` above:

```js
{
    ...,
    connected: false,
    signalCode: 'SIGINT',
    exitCode: null,
    killed: false,
    ...,
}
```

To at least make sure that the process is not disconnected this additional check should be added:
```diff
diff --git a/packages/server/lib/plugins/util.js b/packages/server/lib/plugins/util.js
index 8d0e823847..b0b6e19076 100644
--- a/packages/server/lib/plugins/util.js
+++ b/packages/server/lib/plugins/util.js
@@ -25,7 +25,7 @@ module.exports = {
 
     return {
       send (event, ...args) {
-        if (aProcess.killed) {
+        if (aProcess.killed || !aProcess.connected) {
           return
         }
```

You could go even further by checking the termination status through `.exitCode`, but I do not think that would be more helpful in this situation.

tomudding:
- Closes #21010

### User facing changelog

Fixed an issue where an ambiguous error message was incorrectly shown to the user when Cypress was interrupted during an active test. Fixed [#21010](https://github.com/cypress-io/cypress/issues/21010).

### Additional details

Interrupting Cypress during an active test would cause the `file:preprocessor` child process' IPC channel to already be disconnected while handling `preprocessor:close`. This would lead to an error as it is no longer possibly to send message. That could cause confusion due to the ambiguous error message produced.

This fix check that the process is still connected before sending messages on the IPC channel.

No more ambiguous error when interrupting the Cypress process.

### PR Tasks


tomudding:
I have added a test case for the new behaviour (similar to the test for killed processes). However, `lib/plugins/util #wrapIpc #send sends event through the process` still fails. I assume this is because the `connected` property is not defined in the `theProcess` object, but I am not sure how it would have worked before then as `killed` is also not defined.

Let me know what you think.

jennifer-shehane (changes requested):
@tomudding There's a failing test in the test file you touched, could you look at the failure to see if this is related? Thanks.

tomudding:
> @tomudding There's a failing test in the test file you touched, could you look at the failure to see if this is related? Thanks.

Yes I am aware, the reason is that the `connected` is undefined in the test `theProcess` object. Hence it fails for the generic test (as `!undefined == true`). Should I just add this property to the `theProcess` object in `beforeEach`?

BlueWinds:
> > @tomudding There's a failing test in the test file you touched, could you look at the failure to see if this is related? Thanks.
> 
> Yes I am aware, the reason is that the `connected` is undefined in the test `theProcess` object. Hence it fails for the generic test (as `!undefined == true`). Should I just add this property to the `theProcess` object in `beforeEach`?

Yeah, adding it to the mocked process object seems correct. Looks good to me, just rerunning the flaky test so we can get everything green.

BlueWinds (approved):

BlueWinds:
These bogus error messages when cancelling tests have bothered me, but never enough to take the time to fix it - thanks for taking the time!

emilyrohrbough (approved):

Issue #20995, opened 2022-04-08
Issues in testing REST API using Cypress in corporate env


anithiya:
### Current behavior

I am working in a corporate environment. I need to test certain APIs using Cypress. The major challenge I have is I need to call APIs which require Client Certificates to be passed. However, the Client Certificates are not being detected by Cypress.

I followed the steps laid out to for configuring proxy as explained in: https://docs.cypress.io/guides/references/proxy-configuration#Set-a-proxy-on-Linux-or-macOS

I am able to test the UI using Cypress in our corporate env.

However, when testing with REST APIs, i followed the steps here to capture the ClientCertificates as our API use certs for auth.

https://docs.cypress.io/guides/references/client-certificates

But I get a timeout error:
`CypressError
cy.request() failed trying to load:

https://apiendpoint1.com:1234/v1/resourceName

We attempted to make an http request to this URL but the request failed without a response.

We received this error at the network level:

Error: Error establishing proxy connection. Response from server was: HTTP/1.1 503 Service Unavailable
Cache-Control: no-cache
Pragma: no-cache
X-XSS-Protection: 1
Content-Type: text/html; charset=utf-8
Connection: close
Content-Length: 935`

CYPRESS.JSON
{ "viewportHeight":1080, "viewportWidth":1920, "testFiles": "**/*.{feature,features,js}", "defaultCommandTimeout": 60000, "responseTimeout":120000, "supportFile": "cypress/support/index.js", "reporter":"mochawesome", "reporterOptions": { "reportDir": "cypress/results", "overwrite": false, "html": true, "json": false }, "pageLoadTimeout": 120000, "env": { "testEnv": "qa", "apiendpoints":{ "apiendpoint1": "https://apiendpoint1.com:1234", "apiendpoint2": "https://apiendpoint2.com:1234" } },"clientCertificates": [ { "url": "https://apiendpoint1.com:1234", "ca": [], "certs": [ { "pfx": "cypress/certs/certs.jks", "passphrase": "cypress/certs/pass_phrase.txt" } ] }, { "url": "https://apiendpoint2.com:1234", "ca": [], "certs": [ { "pfx": "cypress/certs/certs.jks", "passphrase": "cypress/certs/pass_phrase.txt" } ] } ] }

SPEC.JS

`describe("SampleTest", () => {
it('gives a response matching a fixture object', () => {
let requestBody = "{ 'name':'nameValue','address':'addressVal'}";
cy.request({
method: 'POST',
[...]

PR #21016, opened 2022-04-11
fix: cy.request follows RFC7230 5.4

### LINKED ISSUE #20289: cy.request causing 400 status due to missing Host header
### Current behavior

I am using `cy.request` to simulate user behaviour.

It appears that Cypress does not send the [Host](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host) header, which makes my (cloudflare-gated) api reject the request.

```bash
https://my_apicall.com/

The response we received from your web server was:

  > 400: Bad Request

This was considered a failure because the status code was not `2xx` or `3xx`.

If you do not want status codes to cause failures pass the option: `failOnStatusCode: false`

-----------------------------------------------------------

The request we sent was:

Method: GET
URL: https://my_apicall.com/
Headers: {
  "Connection": "keep-alive",
  "user-agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Cypress/9.4.1 Chrome/94.0.4606.81 Electron/15.3.4 Safari/537.36",
  "accept": "*/*",
  "accept-encoding": "gzip, deflate",
  "content-type": "application/json",
  "content-length": 83,
  "referer": "https://my_apicall.com/"
}

Redirects: [
  "307: https://my_apicall.com/expected_redirect"
]

-----------------------------------------------------------

The response we got was:

Status: 400 - Bad Request
Headers: {
  "date": "Mon, 21 Feb 2022 14:14:55 GMT",
  "content-type": "text/html; charset=UTF-8",
  "transfer-encoding": "chunked",
  "connection": "keep-alive",
  "referrer-policy": "no-referrer",
  "accept-ranges": "bytes",
  "x-served-by": "cache-iad-kiad7000064-IAD",
  "x-cache": "MISS",
  "x-cache-hits": "0",
  "x-timer": "S1645452895.319054,VS0,VE0",
  "cf-cache-status": "DYNAMIC",
  "expect-ct": "max-age=604800, report-uri=\"https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct\"",
  "report-to": "{\"endpoints\":[{\"url\":\"https:\\/\\/a.nel.cloudflare.com\\/report\\/v3?s=wYuVS5epYzpLY8Q4hM8ZId0we2tHSErWBDPwW7tLSOmWHAfExOyK8y27u9bPocmpg5oYp2xAcKgOvSMfjo1nT2vcU077sCIfdYD70PO5cePeF%2BDjvSIcB0sz4OqC7rg%3D\"}],\"group\":\"cf-nel\",\"max_age\":604800}",
  "nel": "{\"success_fraction\":0,\"report_to\":\"cf-nel\",\"max_age\":604800}",
  "server": "cloudflare",
  "cf-ray": "6e109573bdfe5a46-IAD",
  "alt-svc": "h3=\":443\"; ma=86400, h3-29=\":443\"; ma=86400"
}
Body: <!DOCTYPE html>
```

### Desired behavior

Send the `Host` header so that they are valid according to `rfc7230 5.4`.

### Test code to reproduce

Attempt to call an API that forces strict RFC compliance: fail.

Manually add Host using `cy.request( { headers: { Host: ... } } )`: success.

### Cypress Version

9.5.0

### Other

_No response_

sainthkh:
- Closes #20289

### User facing changelog

`cy.request()` now always sends `host` header field.

### Additional details
- Why was this change necessary? => [RFC7230 5.4](https://datatracker.ietf.org/doc/html/rfc7230#section-5.4) states that all requests must send "Host" header field, but `cy.request` doesn't if it is not specified by the user. 
- Any implementation details to explain? => I decided to add this on the server side, because it would break all `request_spec.js` tests.

**Before:** Some strict servers don't allow empty "host" field and `cy.request` fails.
**After:** It passes for those servers.

### PR Tasks

### DIFF
--- a/packages/driver/cypress/integration/commands/request_spec.js
+++ b/packages/driver/cypress/integration/commands/request_spec.js
@@ -1387,4 +1387,16 @@ describe('src/cy/commands/request', () => {
       })
     })
   })
+
+  context('#request tests without Cypress.backend() stub', () => {
+    it('host field is added to the headers', (done) => {
+      cy.on('fail', (err) => {
+        expect(err.message).to.include('"host": "localhost:3500"')
+
+        done()
+      })
+
+      cy.request('/')
+    })
+  })
 })
--- a/packages/server/lib/server-base.ts
+++ b/packages/server/lib/server-base.ts
@@ -374,6 +374,21 @@ export abstract class ServerBase<TSocket extends SocketE2E | SocketCt> {
   }
 
   _onRequest (headers, automationRequest, options) {
+    // RFC7230 5.4 specifies that:
+    // A client MUST send a Host header field in all HTTP/1.1 request messages.
+    // @see https://datatracker.ietf.org/doc/html/rfc7230#section-5.4
+    //
+    // If Host header field is not defined in options.headers object,
+    // we define it here.
+
+    if (!options.headers) {
+      options.headers = {}
+    }
+
+    let url = new URL(options.url)
+
+    options.headers['host'] = url.host
+
     // @ts-ignore
     return this.request.sendPromise(headers, automationRequest, options)
   }
--- a/system-tests/__snapshots__/request_spec.ts.js
+++ b/system-tests/__snapshots__/request_spec.ts.js
@@ -227,6 +227,7 @@ Method: GET
 URL: http://localhost:2294/statusCode?code=503
 Headers: {
   "Connection": "keep-alive",
+  "host": "localhost:2294",
   "user-agent": "foo",
   "accept": "*/*",
   "accept-encoding": "gzip, deflate"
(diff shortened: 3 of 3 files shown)


jennifer-shehane:
Have to have a discussion on whether this should be considered a breaking change. @sainthkh thoughts?

flotwig:
@sainthkh I'm having trouble reproducing the issue as described.

```ts
it('', () => {
    cy.request('http://requestbin.net/r/axmmsvi7')
})
```

bins a request with a valid Host header: 
image

I thought maybe this fix had something to do with relative URLs, but

```ts
it('', () => {
    cy.visit('http://requestbin.net')
    cy.request('/r/axmmsvi7')
})
```

still bins a request with a Host header: 
image

What exactly needs to be done to reproduce this?

---

Possibly related? https://github.com/cypress-io/cypress/issues/20995

sainthkh:
I simply read the error message and thought "host" header isn't included. 

I tested it with WireShark and learned that "host" is sent correctly. I'll close this PR and ask for reproducible example.

PR #20604, opened 2022-03-14
fix: e2e percy snapshots

brian-mann:
- Closes UNIFY-1149 / https://github.com/cypress-io/cypress/issues/23435

### Details
Allows static vue assets to be accessible without going through the proxy. 

Percy generates local snapshots outside of Cypress in their own browser without configuring the proxy server. Our internal proxy server normally rejects requests made outside of the browser (in the proxy mode). This change allows static assets for vite to be served.

PR #21019, opened 2022-04-11
fix: aut centering and height calculation

marktnoonan:
- Closes UNIFY-1380, UNIFY-1340 and UNIFY-1446

### User facing changelog

- The AUT will be at the expected scale when the runner launches, as the reporter and specs list width used to calculate the initial scale 
- The height of the header in the AUT area will be included when calculating scale - items wrapping in the header, message, or opening the reporter will all be correctly accounted for the so AUT remains fully visible. 
- A 1 pixel vertical border that was appearing on the left hand side of screenshots created with `cy.screenshot()` has been fixed.

👆All of this brings us into parity with 9.x (though screenshots from CT in 9.x have a similar extra pixel on the left, I don't see any issues about this but still seems better to fix than to leave it wrong, especially since the extra border would now impact all screenshots including e2e).

### Additional details

We were not initializing the Specs List and Reporter width values from saved preferences correctly when calculating the initial scale of the AUT. This meant we calculated the initial scale based on the default values for the reporter and specs lists sizes, even though the UI showed the correct actual size of the Reporter and Specs List. Setting those values from GQL in the store before initializing the style for the AUT fixes this.

Since I was I already in the scale math and was aware of UNIFY-1340, I added the fix for that as well, to get scaling 100% correct.

The "extra 1 px border on the left" screenshot issue doesn't have a ticket, but it caught my attention while dialing in the screenshot behavior. The screenshot behavior needed to be rechecked for this PR because changing the transform origin of the AUT and adding the left margin from 9.x  would break many screenshot captures if we didn't make those changes go away during screenshotting.

Walking through the PR, these are the changes:

- add "scales the AUT correctly" tests to cypress-in-cypress
- start watching the height of the AUT header so that we can recalculate scale when it changes
- clean up some imports in SpecRunnerHeaderOpenMode a bit
- only add the 1px left hand boarder to the main-pane to avoid it showing up as an extra pixel on the left of screenshots
- fix origin of scaling transforms, and bring over an explicit margin-left that was used in 9.x to keep things centered when scaling.
- Reorganize how things happen in the SpecRunnerOpenMode (and run mode) script section, so that the we update the store with the computed values of `specsListWidthPreferences` and `reporterWidthPreferences` _before_ calling `useRunnerStyle()` to give us the initial scale. This was the source of both the "too large for window" and "not taking enough space" category of bugs.
- Simplified `useRunnerStyle.ts` a little bit, it exports some values and accepts some params that we just not need any more.
- Replaced the old reference screenshots and fixed a typo in the folder name.
- Almost got the screenshots into Percy, but created https://cypress-io.atlassian.net/browse/UNIFY-1566 for that instead when it seemed to have some nuance.

#### Old resizing behavior
Shows all three bugs - starting out scaled to small, header wrapping causing bottom of AUT to disappear, and starting out scaled too large, cutting off right side of AUT):

https://user-images.githubusercontent.com/8340719/163070501-d44a3cbd-c135-416a-89ca-a09ccf6149f4.mov

#### New resizing behavior
Sows that all 3 bugs cannot be recreated in the same ways, full AUT iframe is always visible, and when scaled, the margins and centering are correct.

https://user-images.githubusercontent.com/8340719/163069583-71bc4749-1bf1-4191-85c0-f17ed9c66f5e.mov

#### Old vs new screenshots
No longer has that  1px border on the left.

old vs new screenshots

9.x screenshots from its versions of the spec had the same extra pixel, caused by same issue of a border hanging around:

Screen Shot 2022-04-13 at 8 43 19 AM

### Testing

This PR adds percy snapshots for various sizes of the runner and confirms that we use the value from GQL for the initial scale calculation. These snapshots won't add value until #20604 is merged, since Cy-in-Cy tests do not render in Percy.

To manually test, follow the example in the videos. Resize the reporter to much larger or much smaller than the default. And refresh the page. Prior to this PR, scaling should start out wrong, after this PR scaling should start out correct.

### PR Tasks


packages/app/src/runner/SpecRunnerOpenMode.vue
```diff
@@ -179,21 +172,32 @@ const {
...
 
+const specsListWidthPreferences = computed(() => {
+  return props.gql.localSettings.preferences.specListWidth ?? runnerUiStore.specListWidth
+})
+
+const reporterWidthPreferences = computed(() => {
+  return props.gql.localSettings.preferences.reporterWidth ?? runnerUiStore.reporterWidth
+})
+
+// we must update pr before calling userRunnerStyle, to make sure that values from GQL
```
ZachJW34, line 183: Comment seems incomplete

ZachJW34 (approved):
When I run the new test you added in open mode, the `scales the AUT correctly in e2e` always fails, though it passes when run in isolation. Are you seeing this?

Other than that looks good, not sure if my machine is being weird and CI is green so I'm going to ✔️

packages/app/src/runner/SpecRunnerOpenMode.vue
marktnoonan: Dang I did not see that test failure, gonna check it out and fix that comment!

marktnoonan: I was able to see it fail but only once. I think though, we might have run afoul of some gql updates when trying to close the specs list so soon after the doing the last resize earlier in the test - from what I saw the reporter ended up too wide, which could be explained if closing the specs list happened to resolve before updating the width. 

I adjusted the test a bit get better separation between those things. Would be good to know if you can still reproduce? Though I think it represents something no user can every be fast enough to do.

ZachJW34:
@marktnoonan tested the changes, tests are passing locally for me now

packages/app/src/runner/SpecRunnerHeaderRunMode.vue
```diff
@@ -70,6 +72,14 @@ const displayScale = computed(() => {
 
 const autStore = useAutStore()
 
+const AUTHeaderEl = ref<HTMLDivElement>()
```
lmiller1990, line 75: I was going to suggest sharing this via composable

```ts
function useAutHeader () {
  const autHeaderEl = ref<HTMLDivElement>()
  const { height } = useElementSize(AUTHeaderEl)
  
  watch(height, () => {
    autStore.setSpecRunnerHeaderHeight(height.value)
  })

  return {
    autHeaderEl
  }
}

// usage

const { autHeaderEl } = useAutHeader()
```

But now I think we don't even need to duplicate this. In run mode, the AUT header should never resize, right, since the user cannot drag/drop during run mode - so I don't think you need this code at all here, only in `SpecRunnerHeaderOpenMode.vue`.

packages/app/src/runner/SpecRunnerOpenMode.vue
```diff
@@ -1,7 +1,8 @@
 <template>
   <RemovePositioningDuringScreenshot
     id="main-pane"
-    class="flex border-gray-900 border-l-1"
+    class="flex border-gray-900"
+    :class="screenshotStore.isScreenshotting ? '' : 'border-l-1'"
```
lmiller1990, line 5: `RemovePositioningDuringScreenshot.vue` already has conditional logic on i`Screenshotting`:

```js
// ...

const style = computed(() => {
  if (screenshotStore.isScreenshotting) {
    return {
      left: `0px`,
      width: `100%`,
    }
  }

  return {
    width: `calc(100% - ${navbarWidth})`,
  }
})
```

Can we just add this class inside there? The idea of `<XXXDuringScreenshot>` is to keep all the screenshot-specific conditional styles in a single place, rather than it leaking out into the rest of the app.

packages/app/src/runner/SpecRunnerRunMode.vue
```diff
@@ -1,7 +1,8 @@
 <template>
   <RemovePositioningDuringScreenshot
     id="main-pane"
-    class="flex border-gray-900 border-l-1"
+    class="flex border-gray-900"
+    :class="screenshotStore.isScreenshotting ? '' : 'border-l-1'"
```
lmiller1990, line 5: Can avoid this duplication by putting the logic in `RemovePositioningDuringScreenshot` :+1:

packages/app/src/runner/SpecRunnerRunMode.vue
```diff
@@ -57,7 +58,8 @@
           <div
             v-show="!autStore.scriptError"
             :id="RUNNER_ID"
-            class="origin-top-left viewport"
+            class="viewport"
+            :class="screenshotStore.isScreenshotting ? 'origin-top-left' : 'origin-top'"
```
lmiller1990, line 62: How come this is slightly different to https://github.com/cypress-io/cypress/pull/21019/files#diff-664b062627eb76f574368ed4f4a7d4e76a18393da8663e7ff64b46fe6020a433L78 ?

packages/app/src/runner/useRunnerStyle.ts
```diff
@@ -67,10 +53,18 @@ export const useRunnerStyle = ({
...
-      transform: scale(${scale.value});`
+    let style = `
+    width: ${autStore.viewportDimensions.width}px;
+    height: ${autStore.viewportDimensions.height}px;
+    transform: scale(${scale.value});
+    `
+
+    if (!screenshotStore.isScreenshotting) {
+      style += `
+      margin-left: ${(containerWidth.value / 2) - (autStore.viewportDimensions.width / 2) }px`
```
lmiller1990, line 64: How did you arrive at this logic? I wonder if this is worth commenting - it reads like black magic, haha. This is for centering the AUT, correct?

I'm really surprised we can center using `margin` at all - wouldn't this usually just be something like `flex justify-center` on the parent element? I'm guessing there's a good reason for this, but I don't know what it is (thus a comment might be good).

packages/app/src/store/aut-store.ts
```diff
@@ -78,9 +80,12 @@ export const useAutStore = defineStore({
     setScriptError (err: ScriptError) {
       this.scriptError = err
     },
-    setScale (scale) {
+    setScale (scale: number) {
```
lmiller1990, line 83: We should add a `no-implicit-any` rule to `tsconfig.json`.

lmiller1990 (commented):
Looks great! I have a ton of comments/questions, mainly around 1) how the heck some of this works and 2) maintainability (mainly around code duplication, I see this quickly becoming a point of regressions between run vs open mode).

PS I am so glad that 1px is gone, it was bothering me a lot, good to see we are in polish mode now.

packages/app/cypress/e2e/cypress-in-cypress.cy.ts
```diff
@@ -84,5 +92,72 @@ describe('Cypress in Cypress', { viewportWidth: 1500, defaultCommandTimeout: 100
...
+
+      // resize the reporter using each of the dragPositions and take Percy snapshots
+      dragPositions.forEach((position, index) => {
+        dragHandleToClientX('panel2', position).then(() => {
+          const expectedScale = testingTypeExpectedScales[testingType][index]
+
+          if (expectedScale) {
+            cy.contains(expectedScale).should('be.visible')
+          } else {
+            cy.contains('%)').should('not.exist')
```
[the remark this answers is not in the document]
marktnoonan: Due to the different default viewport sizes of CT and E2E, as we make the reporter smaller, we hit the 100% state earlier in CT. `testingTypeExpectedScales.component` has a length of 2.

I wondered about whether this approach is truly ideal, vs having this in two separate specs, because of the different outcomes. For now, I'll add a couple of comments to keep things clear... it's a lot of comments, but it's pretty hard to work backwards from the code and understand why the feature works this way.

packages/app/src/runner/SpecRunnerHeaderRunMode.vue
marktnoonan: The header can resize in runmode sometimes, based on the size of the elements in it - a long URL will cause wrapping inside the header, and (less likely of course) a long custom browser name would do the same. 

We have a [ticket](https://cypress-io.atlassian.net/browse/UNIFY-1510) to correct this, but I don't think it's a truly solved problem yet, though maybe, for runmode, the behavior should be different. The wrapping seems way more important/useful when you are resizing the window in the GUI.

Good point about the composable, I think we should do that for now, and I'll add some comments in UNIFY-1510 about pros/cons around how we currently handle wrapping inside the header.

packages/app/src/runner/SpecRunnerOpenMode.vue
marktnoonan: I didn't initially because I didn't look at all the places we are using `RemovePositioningDuringScreenshot` - maybe it's just in this though, in which case cool, I know we have a "remove classes" helper too.

But yeah, point taken about the leaking, let's add it here.

packages/app/src/runner/SpecRunnerRunMode.vue
marktnoonan: Great catch, I will double check open mode screenshots, it seems like they would probably have the same problem this is solving, so these should match.

packages/app/src/runner/useRunnerStyle.ts
marktnoonan: I can put a comment. Short version is: I arrived at it by looking at `develop`, we already do it there.

There was a centering issue @ZachJW34 pointed out, particularly obvious with portrait viewports. When we have a perfectly centered element (with traditional means) and do a CSS transform to scale it, it the transform doesn't change the actual space the element takes up in the document flow, so margin hacks like this end up being pretty common. I wasn't surprised when I went looking to see how we solved it in 9.x and we were programmatically resetting margin with JS.

tbiethman:
The functionality works great. I'll wait for the suggested refactors to land before giving a final review.

packages/app/cypress/e2e/cypress-in-cypress.cy.ts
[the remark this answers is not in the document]
lmiller1990: Ok, seems fine for now :+1:

packages/app/src/runner/SpecRunnerHeaderOpenMode.vue
```diff
@@ -180,6 +181,14 @@ const props = defineProps<{
 
 const showAlert = ref(false)
 
+const AUTHeaderEl = ref<HTMLDivElement>()
```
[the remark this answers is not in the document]
lmiller1990: @marktnoonan any update on this?

lmiller1990:
@marktnoonan can you re-ping me when this is ready to re-review? Thanks.

packages/app/src/runner/SpecRunnerRunMode.vue
marktnoonan: Turned out both are working fine and this conditional setting of origin was able to be removed.

packages/app/src/runner/SpecRunnerHeaderOpenMode.vue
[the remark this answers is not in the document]
marktnoonan: updated 👍

packages/app/src/runner/useAutHeader.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const autStore = useAutStore()
+
+export function useAutHeader () {
+  const autHeaderEl = ref<HTMLDivElement>()
+  const { height } = useElementSize(autHeaderEl)
+
+  watch(height, () => {
+    autStore.setSpecRunnerHeaderHeight(height.value)
+  })
```
lmiller1990, line 12: Should it be `immediate: true`?

lmiller1990 (commented):
One last comment

packages/app/src/runner/useAutHeader.ts
marktnoonan: Yeah that seems good 👍

marktnoonan:
Ok, updates have been made for @lmiller1990's feedback, ready for you @tbiethman :)

packages/app/src/runner/useAutHeader.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  watch(height, () => {
+    autStore.setSpecRunnerHeaderHeight(height.value)
+  }, {
+    immediate: true,
+  })
+
+  return {
+    autHeaderEl,
+  }
```
tbiethman, line 18: We should consider building a full-on AutHeader component at some point (not this PR). There's enough shared logic now that is would clean up the header containers quite a bit, even if we have to disable most of the functionality in run mode.

tbiethman (approved):
Try as I might, I can't break the new scaling logic 🎉  new test coverage looks good, lots of good reuse, no complaints from me

packages/app/src/runner/useAutHeader.ts
lmiller1990: Agreed

lmiller1990:
Let's ship it

PR #21020, opened 2022-04-11
fix: Guard against slow requests in GraphQL Resolution (part 2)

tgriesser:
- Closes UNIFY-1206

### User facing changelog

Part 2 of the changes related to UNIFY-1206, this addresses requests dealing with fetching remote data. The main change is the addition of a `nexusCloudRequestPlugin`, and `graphqlRequestInfo` associated with a DataContext run through a `Proxy`, scoped to each request. 

We assume that any remote fields should return `null` immediately, unless we already have their info in the server-side cache, we will then trigger the client to refetch. By adding the `graphqlRequestInfo`, we are able to know what the current operation is, and target the client and tell it specifically which operations to refetch, rather than refetching them all. In general, this adds better handling for this specific situation which will be more and more common as we stitch in the remote schema, without needing to define any extra logic via directives client-side, or add additional duplication in subscriptions for each possible case.

### Additional details

- Simplifies the data layer by removing the SSR "prefetching" of GraphQL data. That was mostly added as a band-aid to cover up the slowness of field-resolution, but adds a lot of unnecessary complexity for the value it provides. If the queries are fast by default, there's no real value to having this layer.

### PR Tasks


packages/data-context/src/sources/VersionsDataSource.ts
```diff
@@ -51,20 +57,24 @@ export class VersionsDataSource {
    * }
    */
   async versionData (): Promise<VersionData> {
-    const [latestVersion, npmMetadata] = await Promise.all([this._latestVersion, this._npmMetadata])
+    const versionData = this.#ensureData()
+    const [latestVersion = pkg.version, npmMetadata] = await Promise.all([
```
ryanthemanuel, line 61: Why do we need to default this to pkg.version. Doesn't `latestVersion` always return a string

tgriesser: I think this was a type issue, but I will update this

ryanthemanuel: Ah I guess since it's getting it from ensureData as `this.ctx.coreData.versionData` there may not be a guarantee they're populated.

packages/data-context/src/sources/VersionsDataSource.ts
```diff
@@ -25,14 +25,20 @@ const NPM_CYPRESS_REGISTRY = 'https://registry.npmjs.org/cypress'
...
     this._initialLaunch = true
     this._currentTestingType = this.ctx.coreData.currentTestingType
-    this._latestVersion = this.getLatestVersion()
-    this._npmMetadata = this.getVersionMetadata()
+    this.#ensureData()
+  }
+
+  #ensureData () {
+    this.ctx.update((d) => {
+      d.versionData.latestVersion ??= this.getLatestVersion().catch((e) => undefined)
```
ryanthemanuel, line 37: What's the value of storing these on core data rather than on the data source?

packages/graphql/src/plugins/nexusCloudRequestPlugin.ts
```diff
@@ -0,0 +1,117 @@
...
+const NO_RESULT = {}
+// 2ms should be enough time to resolve from the local cache of the
+// cloudUrqlClient in CloudDataSource
+const RACE_MAX_EXECUTION_MS = 2
+const IS_DEVELOPMENT = process.env.CYPRESS_INTERNAL_ENV !== 'production'
+const debug = debugLib('cypress:graphql:nexusCloudRequestPlugin')
+
+/**
+ * This plugin taps into each of the requests and checks for the existence
+ * of a "Cloud" prefixed type. When we see these, we know that we're dealing
```
ryanthemanuel, line 15:
```suggestion
 * of a "Cloud" prefixed type or if it's marked as remote. When we see these, we know that we're dealing
```

packages/graphql/src/schemaTypes/objectTypes/gql-Query.ts
```diff
@@ -45,6 +51,7 @@ export const Query = objectType({
     })
 
     t.field('versions', {
+      remote: true,
```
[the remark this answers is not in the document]
ryanthemanuel: Hmmm I might like `autoDeferred`. Other thoughts:

* deferOnInitialLoad
* deferIfNotLoaded
* deferIfNull

packages/data-context/src/sources/VersionsDataSource.ts
tgriesser: It was b/c the `VersionsDataSource` is cached on the data context, and since we keep the same data context globally, it was hard to reset the state for test.

packages/data-context/src/sources/VersionsDataSource.ts
tgriesser: Yeah, to make it correctly typesafe without cheating, the alternative was uglier I think:

```ts
  #ensureData () {
    this.ctx.update((d) => {
      d.versionData.latestVersion ??= this.getLatestVersion().catch((e) => pkg.version)
      d.versionData.npmMetadata ??= this.getVersionMetadata().catch((e) => undefined)
    })

    return this.ctx.coreData.versionData
  }
```
becomes

```ts
  #ensureData () {
    this.ctx.update((d) => {
      d.versionData.latestVersion ??= this.getLatestVersion().catch((e) => pkg.version)
      d.versionData.npmMetadata ??= this.getVersionMetadata().catch((e) => undefined)
    })

    const { latestVersion, npmMetadata } = this.ctx.coreData.versionData

    assert(latestVersion)

    return { latestVersion, npmMetadata }
  }
```

Up to you if you think we should change. We could also look at a different pattern of doing this, similar to `GitDataSource` where we pass in what we need and cache that in the `versionData`, making it possible to reset without many changes here

estrada9166 (approved):

packages/data-context/src/DataContext.ts
```diff
@@ -441,10 +420,12 @@ export class DataContext {
       this.actions.dev.watchForRelaunch()
     }
 
+    // We want to fetch the user immediately, but we don't need to block the UI on this
+    this.actions.auth.getUser().catch(() => {
+      // This error is handled internally by getUser, and isn't actionable here
```
flotwig, line 425: an empty `catch` block makes this code smelly. either this error should *really* be handled internally by getUser, and not throw, or we should at least log that there is an error being thrown here. otherwise we could be swallowing real errors form getUser.

packages/data-context/src/actions/WizardActions.ts
```diff
@@ -146,23 +146,30 @@ export class WizardActions {
...
         this.ctx.actions.project.setForceReconfigureProjectByTestingType({ forceReconfigureProject: false, testingType: 'e2e' })
 
-        return chosenLanguage
+        break
       }
       case 'component': {
+        const { chosenBundler, chosenFramework } = this.ctx.wizard
+
+        if (!chosenBundler || !chosenFramework) {
+          return
```
flotwig, line 165: `break` or `return`? pick one to use to exit this function, you have a mix in here :smile: my vote is on `return` since it is clearer that the function is exiting

packages/data-context/src/data/coreDataShape.ts
```diff
@@ -138,6 +138,10 @@ export interface CoreDataShape {
   packageManager: typeof PACKAGE_MANAGERS[number]
   forceReconfigureProject: ForceReconfigureProjectDataShape | null
   cancelActiveLogin: (() => void) | null
+  versionData: {
+    latestVersion?: Promise<string | undefined>
+    npmMetadata?: Promise<Record<string, string> | undefined>
```
flotwig, line 143: nit: I really don't like these tri-states, it's a decent amount of mental load when `latestVersion` can either be `undefined`, `Promise<undefined>`, or `Promise<string>`. I try to refactor them whenever I encounter them. Can you think of a way to clean this up?

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -152,9 +168,14 @@ fragment HeaderBar_HeaderBarContent on Query {
   projectRootFromCI
   ...TopNav
   ...Auth
+  ...HeaderBarContent_Auth
 }
 `
 
+const userData = computed(() => {
+  return props.gql.cloudViewer ?? props.gql.cachedUser
+})
```
flotwig, line 177: what do you think about doing this on the server-side instead? IMO the HeaderBarContent component shouldn't have to understand the inner workings of how we cache user data

packages/graphql/src/plugins/nexusDeferIfNotLoadedPlugin.ts
```diff
@@ -0,0 +1,117 @@
...
+export const nexusDeferIfNotLoadedPlugin = plugin({
+  name: 'nexusDeferIfNotLoadedPlugin',
+  fieldDefTypes: 'deferIfNotLoaded?: true',
+  onCreateFieldResolver (def) {
+    const { name: parentTypeName } = def.parentTypeConfig
+
+    // Don't ever need to do this on Subscription / Mutation fields,
+    // or on the Cloud types themselves, since these don't actually need to resolve themselves,
+    // they're resolved from the remote request
+    if (parentTypeName === 'Mutation' || parentTypeName === 'Subscription' || parentTypeName.startsWith('Cloud')) {
```
flotwig, line 28: switching behavior off of name prefixes feels dirty. is there a better way to accomplish this?

flotwig (changes requested):

flotwig:
@tgriesser Launchpad feels so fast now! Am I at McDonalds? Cuz I'm Loving It.

packages/data-context/src/sources/VersionsDataSource.ts
ryanthemanuel: Yeah that's a little messier. I think I'm ok with the way you have it and it makes sense. It just stood out to me a little bit was the main thing

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
tgriesser: I would definitely opt for that if there was a good way. With how I've written it, it currently can't happen b/c the `cloudViewer` is resolved directly from the cloud, so it returns `null` immediately when the type isn't fulfilled with the full shape (and there is more info that doesn't exist in the `CachedUser`). The `cachedUser` type we expose in the GraphQL schema is to avoid the flash that happens when navigating pages.

The component is only aware so far as what info we provide via the graphql schema, so it's technically user-facing in that sense, but I agree it's still not great.

I'll think a bit about whether there's a better solution here, e.g. by having a way to provide an alternate result on cache miss for `deferIfNotLoaded` though 🤔

packages/data-context/src/data/coreDataShape.ts
tgriesser: We could probably get it to: 
```
latestVersion?: Promise<string>,
npmMetadata?: Promise<Record<string, string>>
```
let me give it a shot

packages/data-context/src/actions/WizardActions.ts
tgriesser: Yeah I'll make this consistent

packages/data-context/src/DataContext.ts
tgriesser: Yeah good point. It doesn't throw, it's just that b/c it's a promise the linting yells b/c it's an unhandled promise. I'll add a `ctx.log(err)` or something here for good measure

packages/app/cypress/e2e/top-nav.cy.ts
```diff
@@ -1,15 +1,18 @@
 import type { SinonStub } from 'sinon'
 import defaultMessages from '@packages/frontend-shared/src/locales/en-US.json'
+import type Sinon from 'sinon'
 
 const pkg = require('@packages/root')
 
 const loginText = defaultMessages.topNav.login
 
+beforeEach(() => {
+  cy.clock(Date.UTC(2021, 9, 30), ['Date'])
```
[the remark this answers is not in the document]
tgriesser: There's 2 describes in this file I think and that was the issue. The whole approach here needs to change, the client should not be concerned with the logic of the timing it's doing, that should be modeled in the GraphQL layer

packages/data-context/src/actions/WizardActions.ts
```diff
@@ -146,23 +146,30 @@ export class WizardActions {
   async scaffoldTestingType () {
     const { currentTestingType, wizard: { chosenLanguage } } = this.ctx.coreData
 
-    assert(currentTestingType)
-    assert(chosenLanguage)
+    if (!currentTestingType || !chosenLanguage) {
+      return
+    }
```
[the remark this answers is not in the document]
tgriesser: Commented on the Jira ticket. I agree it is weakened but a lot of this stuff is wrong in general and need to be cleaned up in better places than is in scope for getting this working.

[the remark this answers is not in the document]
tgriesser: I think we're overhauling this anyway with a lot of the other work that's going on in Lachlan's PR. I'll add a note/TODO here.

packages/frontend-shared/src/gql-components/topnav/UserAvatar.vue
```diff
@@ -11,7 +11,7 @@ import gravatar from 'gravatar'
 import { computed } from 'vue'
 
 const props = defineProps<{
-  email?: string
+  email?: string | null
```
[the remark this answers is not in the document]
tgriesser: I kept it b/c there was a test for:

```
it('renders when no email address is passed', () => {
    cy.mount(() => <UserAvatar class="h-50px w-50px"/>)
    validateUserAvatar()
})
```

but that seems like a weird use-case, we should change that test to say when `null` email is passed

packages/graphql/src/plugins/nexusDeferIfNotLoadedPlugin.ts
tgriesser: Types in the `cloud.graphql` schema are consistently named with `Cloud*` so I don't think it's terrible, we could check against things defined in that schema, but then we'd need to have a separate list of things that can be used in both, like `PageInfo` & `Node`. I think this is probably simplest for now.

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
tgriesser: Yeah, I think this is probably the simplest solution for now honestly. Anything else would add a lot more complexity for what is effectively an edge case... again, it's exposed in the GraphQL schema, so that contract is as far as the client knows about anything there

packages/frontend-shared/src/gql-components/topnav/UserAvatar.vue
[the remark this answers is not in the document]
tgriesser: Nevermind, undid this - it was simpler to just have it be a `Maybe<T>`, which I don't see as being a terrible pattern if it's typed properly

packages/graphql/src/plugins/nexusDeferIfNotLoadedPlugin.ts
flotwig: I was thinking something like adding a `noDefer` field... problem is, someone will add something without the `Cloud` prefix eventually, and this line will stop doing what it's supposed to do. Moving this to the definition of the type itself would make it a lot clearer to future devs what's happening.

packages/data-context/src/actions/WizardActions.ts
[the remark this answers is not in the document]
flotwig: Is this change necessary though? I'd rather see it undone here while we're talking about it than in a future PR that may or may not catch this. 'Cause otherwise we're dealing with a function that decides if itself is called until a future date where these cleanups are a priority.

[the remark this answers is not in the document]
tgriesser: Yes, the changes here making things work properly exposed legit bugs that I spent way too much time debugging before deciding it wasn’t worth the effort to fix because I’m ripping this all up in my next PR anyway

packages/graphql/src/plugins/nexusDeferIfNotLoadedPlugin.ts
tgriesser: > someone will add something without the Cloud prefix eventually

That’s a process issue we can/will address on the cloud side. We already have pretty good static checks in place

tgriesser: Unless you’re actually saying someone will add it in the local typedefs. I can also add something that guards on that

packages/data-context/src/actions/WizardActions.ts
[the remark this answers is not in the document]
tgriesser: @flotwig if it helps, I added an explicit comment in the ticket where I'll address this: https://cypress-io.atlassian.net/browse/UNIFY-1256?focusedCommentId=17678

packages/graphql/src/plugins/nexusDeferIfNotLoadedPlugin.ts
tgriesser: Updated - it's a bit more explicit now

packages/frontend-shared/src/graphql/urqlClient.ts
```diff
@@ -75,13 +62,16 @@ interface AppUrqlClientConfig {
...
-export function makeUrqlClient (config: UrqlClientConfig): Client {
+export async function makeUrqlClient (config: UrqlClientConfig): Promise<Client> {
   let hasError = false
 
+  const dfd = pDefer()
   const exchanges: Exchange[] = [dedupExchange]
 
   const io = window.ws ?? getPubSubSource(config)
 
+  io.on('connect', dfd.resolve)
```
ryanthemanuel, line 73: Are there any errors we need to be listening to and rejecting here?

ryanthemanuel, line 73: Rather than doing a pDefer,  could we say:

```suggestion
  const connectPromise = new Promise((resolve, reject) => { io.on('connect', resolve) })
```

And then await that promise on line 148.

packages/graphql/src/makeGraphQLServer.ts
```diff
@@ -195,60 +190,100 @@ export const graphqlWS = (httpServer: Server, targetRoute: string) => {
...
+  variables: Record<string, unknown> | null
+}
+
 /**
- * Adds runtime validations during development to ensure patterns of access are enforced
- * on the DataContext
+ * Since the DataContext is considered a singleton throughout the electron app process,
+ * we create a Proxy object for it, adding metadata associated each GraphQL operation.
+ * This is used in middleware, such as the `nexusDeferIfNotLoadedPlugin`, to associate
+ * remote requests to operations needing to be refetch on the client.
```
ryanthemanuel, line 247:
```suggestion
 * remote requests to operations needing to be refetched on the client.
```

packages/graphql/src/schemaTypes/objectTypes/gql-ErrorWrapper.ts
```diff
@@ -34,7 +34,7 @@ export const ErrorWrapper = objectType({
     t.nonNull.string('errorMessage', {
       description: 'The markdown formatted content associated with the ErrorTypeEnum',
       resolve (source) {
-        return source.cypressError.messageMarkdown
+        return source.cypressError.messageMarkdown ?? source.cypressError.message
```
ryanthemanuel, line 37: Yay. This has bit me a bunch in the past.

packages/frontend-shared/src/graphql/urqlClient.ts
tgriesser: I don't believe so the `connect_error` might be an option, but that retries, so in theory the `connect` should eventually hit. If it doesn't, it means the electron socket server has been terminated, in which the browser shouldn't even be open. Something else bigger would be wrong, and there wouldn't be a way to recover.

tgriesser: Updated!

packages/graphql/src/makeGraphQLServer.ts
tgriesser: Updated!

packages/data-context/src/data/coreDataShape.ts
flotwig: @tgriesser no luck here?

packages/data-context/src/actions/WizardActions.ts
flotwig: @tgriesser

PR #21026, opened 2022-04-11
feat: Add vue2 package from npm/vue/v2 branch

BlueWinds:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1253

### User facing changelog

- bundles `npm/vue2` in the binary, accessible via `cypress/vue2`.

### Additional details

Part of the component testing rearchitecture, to allow us to easily bring updates to users w/o requiring them to install anything other than `cypress`.

### PR Tasks


tgriesser (changes requested):
2 things - I thought we had landed on `vue-2` for the package name

Also, we should skip copying over all of the Cypress tests & `examples` dir from the existing Vue project

BlueWinds:
> 2 things - I thought we had landed on `vue-2` for the package name
> 
> Also, we should skip copying over all of the Cypress tests & `examples` dir from the existing Vue project

The only discussion I saw was in the Jira ticket, where Lachlan commented:

> Can we name it vue2 instead of vue-2? I think it looks way nicer. Also there is some existing literature, like jest27-vue2, etc.

and it seemed reasonable to me. If there was any further discussion, I was unaware, vue2@1.0.0 looks quite nice to me, vs vue-2@1.0.0. The dash feels weird in contexts where we don't see *only* the module name. But I'm not married to the idea if other people have strong preferences / there was discussion I missed.

tgriesser:
> and it seemed reasonable to me. If there was any further discussion, I was unaware, vue2@1.0.0 looks quite nice to me, vs vue-2@1.0.0. The dash feels weird in contexts where we don't see only the module name. But I'm not married to the idea if other people have strong preferences / there was discussion I missed.

That's fine - I was just going based on the Jira ticket title/body, didn't see that comment. @brian-mann or @ZachJW34 do you have a strong opinion on naming here

brian-mann:
`vue2` is fine

lmiller1990:
Regarding names, I see @brian-mann's point now - in the project I maintain, `vue-jest`, we actually have `vue2-jest27` and `vue3-jest27`, so there's no ambiguity. We aren't doing that here, though, so I'm not sure if `vue2` or `vue-2` is better.

lmiller1990 (approved):

tgriesser (approved):

PR #21035, opened 2022-04-12
fix: launching global mode passing in a testing type and cleaning up git change watchers on process exit

ryanthemanuel:
- Closes UNIFY-1561 and UNIFY-1560

### User facing changelog

Launching global mode from the command line passing a testing type no longer crashes. It does *not* actually persist this parameter after selecting a project

### Additional details

This is a regression due to #20775. This PR is fixing the behavior so that the app no longer errors out. It does *not* actually persist the testing type after selecting a project; however, since this was not working previously, we are not addressing it with this PR. Long term we should either try and default to the passed in testing type or not allow a testing type to be passed in in global mode without a project passed in.

In the process of fixing this, I noticed a regression caused by #20997 and am fixing that with this PR as well. Specifically, we were not cleaning up the file watching in the git data source on process exit. This was causing the app to crash with a message similar to this when you kill the process from the command line:

image

### PR Tasks


tgriesser (approved):
Nice, thanks for fixing the `gitInfo` piece too!

tbiethman (approved):

PR #21036, opened 2022-04-12
fix: Command Log performance v2

andrewmtam:
- Closes 

### Additional details

This is a followup to https://github.com/cypress-io/cypress/pull/21007.  In this iteration, we limit the amount of commands that show up by default, but we also expose a toggle which when clicked, shows all of the commands.

This is not the ideal solution of "virtualizing" the list to show only what is necessary as the user scrolls, but it is sort of a half way between providing some control over what commands show vs providing no control in the sense that no data is loss, and the user can control whether or not to show all or some of the commands.

Some notes to point out:
* By default, we only show 50 commands, so that local testing and CI runs can benefit from the performance improvements of a short command list
    * If we reversed this and started off showing the full list, and made the truncated list opt-in, than runs in CI would not be able to reap the performance benefits of the feature showcased in this PR
* When the test is running, I think a user is typically not clicking through snapshots as this could interfere with the test itself, so it seems ok to default to showing a small amount of commands, and that showing all commands requires an interaction
* If the user needs to see the rest of the commands and click through the snapshots, they can opt in by clicking "Show all commands"
    * In the use-cases I'm familiar with, it's only when I stop the test, or the test fails, that I might want to do this to further my investigation of a test run

Some gifs to illustrate this behavior, as well as illustrate the performance difference between showing all the commands vs a truncated list:

#### Toggling Between "show 50" and "show all commands"

The gif reveals that as soon as Show All is clicked (to show all the commands), there are immediate performance implications.

Once they click "Show last 50 commands" to re-hide the commands, the performance becomes smooth again

cypress-fast

#### llustrating that a user can go back and "show all" to interact with snapshots of any command they want

Even though we show only 50 by default, the user can reveal all the commands, and click on any of them to view the snapshot at that moment in time, so no data loss occurs

cypress-debug

### PR Tasks

### DIFF
--- a/packages/reporter/src/hooks/hooks.scss
+++ b/packages/reporter/src/hooks/hooks.scss
@@ -21,6 +21,7 @@
           color: #6a6b6c;
         }
 
+        .hook-show-all-commands,
         .hook-open-in-ide {
           display: block;
         }
@@ -49,7 +50,9 @@
         display: none;
       }
 
+      .hook-show-all-commands,
       .hook-open-in-ide {
+        cursor: pointer;
         align-items: center;
         color: #6a6b6c;
         display: none;
--- a/packages/reporter/src/hooks/hooks.tsx
+++ b/packages/reporter/src/hooks/hooks.tsx
@@ -31,6 +31,33 @@ const HookOpenInIDE = ({ invocationDetails }: HookOpenInIDEProps) => (
   </FileOpener>
 )
 
+const HookShowAllCommands = ({
+  numCommandsToShow,
+  setNumCommandsToShow,
+}: {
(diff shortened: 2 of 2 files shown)


bobbyrenwick:
This is the perfect balance that's needed for solving the problem of performance locally as well as being able to run the same setup during CI. We use ~1 million CPU minutes a month on Cypress and any improvement in performance would be much appreciated from many perspectitves. Cost, Developer Iteration Speed and Environmental. Thanks for implementing [andrewmtam](https://github.com/andrewmtam).

@jennifer-shehane would the Cypress team consider approving this change?

PR #21042, opened 2022-04-12
fix: cy.type('{enter}') on <input> elements submits the form correctly after Firefox 98.

### LINKED ISSUE #21033: `{enter}` not working in Firefox 99 and 100
### Current behavior

When sending `{enter}`, this has no effect in Cypress 9.5.4 in Firefox 100, while it did work in Firefox 97.

This had been already reported in #20562 and has been claimed to be fixed in v9.5.4, but it seems that is not the case.

Tested with Firefox Dev 99.0b8 and 100.0b4 (probably same problem with Firefox 98 as well).

### Desired behavior

Sending `{enter}` should work as it did in Firefox 97 and older.

### Test code to reproduce

```js
describe('Hitting enter', () => {
  it('works', () => {
    cy.visit('https://www.duckduckgo.com/');
    cy.get('input#search_form_input_homepage').type('hello{enter}');
    cy.get('div#links_wrapper').should('exist');
  });
});
```

### Cypress Version

9.5.4

### Other

_No response_

sainthkh:
- Closes #21033

### User facing changelog

cy.type('{enter}') submits the form correctly after Firefox 98.

### Additional details
- Why was this change necessary? => `Enter` event behavior change in Firefox 98 caused `cy.type('{enter}')` to not submit the `<form>`s.

`<form>` is submitted correctly with `cy.type('{enter}')`.

### PR Tasks

### DIFF
--- a/packages/driver/src/cy/commands/actions/type.ts
+++ b/packages/driver/src/cy/commands/actions/type.ts
@@ -233,12 +233,11 @@ export default function (Commands, Cypress, cy, state, config) {
           return
         }
 
-        // In Firefox, submit event is automatically fired
+        // Before Firefox 98, submit event is automatically fired
         // when we send {Enter} KeyboardEvent to the input fields.
         // Because of that, we don't have to click the submit buttons.
         // Otherwise, we trigger submit events twice.
-        // But after Firefox 98, submit event isn't automatically fired.
-        if (!isFirefoxBefore98 || !Cypress.isBrowser('firefox')) {
+        if (!isFirefoxBefore98) {
           // issue the click event to the 'default button' of the form
           // we need this to be synchronous so not going through our
           // own click command
--- a/packages/driver/src/cy/commands/actions/type.ts
+++ b/packages/driver/src/cy/commands/actions/type.ts
@@ -174,6 +174,8 @@ export default function (Commands, Cypress, cy, state, config) {
     }
 
     const type = function () {
+      const isFirefoxBefore98 = Cypress.isBrowser('firefox') && Cypress.browserMajorVersion() < 98
+
       const simulateSubmitHandler = function () {
         const form = options.$el.parents('form')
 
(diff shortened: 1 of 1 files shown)


sainthkh:
## Note

Our CI uses Firefox 93. This PR should be tested manually.

jennifer-shehane:
@sainthkh Thanks for looking into this again!

packages/driver/src/cy/commands/actions/type.ts
```diff
@@ -235,7 +237,8 @@ export default function (Commands, Cypress, cy, state, config) {
         // when we send {Enter} KeyboardEvent to the input fields.
         // Because of that, we don't have to click the submit buttons.
         // Otherwise, we trigger submit events twice.
-        if (!Cypress.isBrowser('firefox')) {
+        // But after Firefox 98, submit event isn't automatically fired.
+        if (!isFirefoxBefore98 || !Cypress.isBrowser('firefox')) {
```
BlueWinds, line 241: The second part of this conditional is redundant. Three cases:
FF 98: (!true || !true) == false
FF 99: (!false || !true) == true
Chromium: (!false || !false) == true

In every case, `!Cypress.isBrowser('firefox')` has no effect on the outcome (because it's already part of the "is FF before 98" check).

BlueWinds (changes requested):

tbiethman (approved):
Tested against local FF98, the related tests pass after changes :+1:

packages/driver/src/cy/commands/actions/type.ts
sainthkh: You're right. 

`isFirefoxBefore98` is `isFirefox && version < 98`.

`!isFirefoxBefore98 || !isFirefox` is `!isFirefox || !(version < 98) || !isFirefox`. 

I fixed it.

BlueWinds (approved):

PR #21047, opened 2022-04-12
fix(cli): show additional mitigation steps for max path length error

flotwig:
- Closes UNIFY-1449

### User facing changelog

- Improved the error message Cypress throws when max file length is exceeded on Windows platforms and added mitigation steps.
- This is pretty easy to check - the syscall is always `realpath` and the code is always `ENOENT` when this throws.
- Could have done an even more excessive check like in [155742b](https://github.com/cypress-io/cypress/pull/21047/commits/155742b0131fd5166d61820ef87c57a3fce3f650) but I think this is not necessary.
- Also opened a docs PR for the mitigation steps: https://github.com/cypress-io/cypress-documentation/pull/4437
- And an onlink PR: https://github.com/cypress-io/cypress-services/pull/4441

### PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4437

### DIFF
--- a/cli/__snapshots__/unzip_spec.js
+++ b/cli/__snapshots__/unzip_spec.js
@@ -1,4 +1,4 @@
-exports['unzip error 1'] = `
+exports['lib/tasks/unzip throws when cannot unzip 1'] = `
 Error: The Cypress App could not be unzipped.
 
 Search for an existing issue or open a GitHub issue at
@@ -12,6 +12,24 @@ Error: end of central directory record signature not found
 ----------
 
 Platform: darwin-x64 (Foo-OsVersion)
+Cypress Version: 1.2.3
+
+`
+
+exports['lib/tasks/unzip throws max path length error when cannot unzip due to realpath ENOENT on windows 1'] = `
+Error: The Cypress App could not be unzipped.
+
+This is most likely because the maximum path length is being exceeded on your system.
+
+Read here for solutions to this problem: https://on.cypress.io/error-messages#win-max-path-length
+
+----------
+
+Error: failed
+
+----------
+
+Platform: win32-x64 (Foo-OsVersion)
 Cypress Version: 1.2.3
 
 `
--- a/cli/__snapshots__/unzip_spec.js
+++ b/cli/__snapshots__/unzip_spec.js
@@ -21,7 +21,7 @@ Error: The Cypress App could not be unzipped.
 
 This is most likely because the maximum path length is being exceeded on your system.
 
-Read here for solutions to this problem: https://on.cypress.io/error-messages#win-max-path-length
+Read here for solutions to this problem: https://on.cypress.io/win-max-path-length-error
 
 ----------
 
--- a/cli/__snapshots__/errors_spec.js
+++ b/cli/__snapshots__/errors_spec.js
@@ -32,6 +32,7 @@ exports['errors individual has the following errors 1'] = [
   "childProcessKilled",
   "failedDownload",
   "failedUnzip",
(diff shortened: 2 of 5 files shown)


cli/__snapshots__/unzip_spec.js
```diff
@@ -15,3 +15,21 @@ Platform: darwin-x64 (Foo-OsVersion)
 Cypress Version: 1.2.3
 
 `
+
+exports['lib/tasks/unzip throws max path length error when cannot unzip due to realpath ENOENT on windows 1'] = `
+Error: The Cypress App could not be unzipped.
+
+This is most likely because the maximum path length is being exceeded on your system.
+
+Read here for solutions to this problem: https://on.cypress.io/error-messages#win-max-path-length
```
cowboy, line 24: The hash doesn't seem to be preserved when the on link system redirects.

```
❯ wget https://on.cypress.io/error-messages#win-max-path-length 2>&1 | grep Location:
Location: https://docs.cypress.io/guides/references/error-messages.html [following]
Location: https://docs.cypress.io/guides/references/error-messages [following]
```

You should probably add an entry here and reference that link in the app. https://github.com/cypress-io/cypress-services/blob/develop/packages/on/data/links.yml#L78

cowboy (changes requested):

cli/__snapshots__/unzip_spec.js
flotwig: It works for me in Chrome and FF. I think the browser is responsible for keeping the hash portion constant across redirects, so `wget` is probably just not emulating this behavior.

flotwig: @cowboy found it, it's actually encoded in an RFC: https://httpwg.org/specs/rfc7231.html#rfc.section.7.1.2

> If the Location value provided in a [3xx (Redirection)](https://httpwg.org/specs/rfc7231.html#status.3xx) response does not have a fragment component, a user agent MUST process the redirection as if the value inherits the fragment component of the URI reference used to generate the request target (i.e., the redirection inherits the original reference's fragment, if any).
> 
> For example, a GET request generated for the URI reference "http://www.example.org/~tim" might result in a [303 (See Other)](https://httpwg.org/specs/rfc7231.html#status.303) response containing the header field:
> 
>    `Location: /People.html#tim`
> 
> which suggests that the user agent redirect to "http://www.example.org/People.html#tim"
> 
> Likewise, a GET request generated for the URI reference "http://www.example.org/index.html#larry" might result in a [301 (Moved Permanently)](https://httpwg.org/specs/rfc7231.html#status.301) response containing the header field:
> 
>    `Location: http://www.example.net/index.html`
> 
> which suggests that the user agent redirect to "http://www.example.net/index.html#larry", preserving the original fragment identifier.
> 
> There are circumstances in which a fragment identifier in a Location value would not be appropriate. For example, the Location header field in a [201 (Created)](https://httpwg.org/specs/rfc7231.html#status.201) response is supposed to provide a URI that is specific to the created resource.

flotwig: Good point, I created a PR for the onlink: https://github.com/cypress-io/cypress-services/pull/4441

tbiethman (approved):
New error looks good and tests are solid :+1:

marktnoonan (approved):
Changes look good, really happy to have this guidance for users about what's otherwise a really unintuitive error.

PR #21055, opened 2022-04-12
chore: fix tests that fail on windows

lmiller1990:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1418
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1417

### User facing changelog

Fixes some flaky tests.

### Additional details

- fixes some runner mocha event tests w/ a slightly longer timeout
- fixes an issue where git info wasn't working correctly due to `/` vs `\` on different OSs
- fixes race condition where could not connect to CDP when launching with `cypress --browser chrome` which primarily manifested on windows

Tests pass on windows.

### PR Tasks


packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -315,20 +318,25 @@ export class GitDataSource {
...
+    const result = await execa(cmd, { shell: true, cwd: this.config.projectRoot })
 
-    const split = result.stdout
-    .split('\r\n') // windows uses CRLF for carriage returns
-    .filter((str) => !str.includes('git log')) // windows stdout contains [cmd,output]. So we remove the code containing the executed command, `git log`
+    const stdout = normalize(result.stdout).split('\r\n') // windows uses CRLF for carriage returns
 
     // windows returns a leading carriage return, remove it
-    const [, ...stdout] = split
+    let output: string[] = []
```
lmiller1990, line 326: More reliable way to pass shell command output.

packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -315,20 +318,25 @@ export class GitDataSource {
   async #getInfoWindows (absolutePaths: readonly string[]) {
     const paths = absolutePaths.map((x) => path.resolve(x)).join(',')
     const cmd = `FOR %x in (${paths}) DO (${GIT_LOG_COMMAND} %x)`
-    const result = await execa(cmd, { shell: true })
+    const result = await execa(cmd, { shell: true, cwd: this.config.projectRoot })
 
-    const split = result.stdout
-    .split('\r\n') // windows uses CRLF for carriage returns
-    .filter((str) => !str.includes('git log')) // windows stdout contains [cmd,output]. So we remove the code containing the executed command, `git log`
+    const stdout = normalize(result.stdout).split('\r\n') // windows uses CRLF for carriage returns
```
lmiller1990, line 323: This was the problem, `/` vs `\`. 

Def need to standardize how we handle this internally. At least w/ this PR we have more coverage and confidence across all OSes, for when we do try to standardize this (if we need to make changes, maybe we don't).

packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -35,6 +35,9 @@ const GIT_LOG_COMMAND = `git log -1 --pretty="format:%ci %ar %an"`
 const GIT_ROOT_DIR_COMMAND = '--show-toplevel'
 const SIXTY_SECONDS = 60 * 1000
 
+function normalize (file: string) {
+  return file.replace(/\\/g, '/') // normalize \ to /
+}
```
flotwig, line 40: i prefer more specific names for utility functions...

```suggestion
function ensurePosixPathSeparators (text: string) {
  return text.replace(/\\/g, '/') // normalize \ to /
}
```

flotwig: Also, since we are on a recent version of Node, you can use `String.prototype.replaceAll`:

```suggestion
function ensurePosixPathSeparators (path: string) {
  return path.replaceAll('\\', '/') // normalize \ to /
}
```

packages/data-context/src/sources/GitDataSource.ts
flotwig, line 326:
```suggestion
    const output: string[] = []
```

packages/data-context/src/sources/GitDataSource.ts
flotwig: Yeah... maybe a `guides/` would be good. Maybe a guide all about platform differences, not just paths?

In general, I think it should work like this:

1. We should always use the native path if it is being displayed to the user. Unless it's displayed in a place where it may be shared cross-platform in a programmatic way (like `require(path)`, `import path`, or in the URL bar)
2. We should always use the native path internally unless the library does not support it (like in `globby`, where `\` acts as an escape character.

What do you think?

packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -315,20 +318,25 @@ export class GitDataSource {
...
     const cmd = `FOR %x in (${paths}) DO (${GIT_LOG_COMMAND} %x)`
-    const result = await execa(cmd, { shell: true })
+    const result = await execa(cmd, { shell: true, cwd: this.config.projectRoot })
 
-    const split = result.stdout
-    .split('\r\n') // windows uses CRLF for carriage returns
-    .filter((str) => !str.includes('git log')) // windows stdout contains [cmd,output]. So we remove the code containing the executed command, `git log`
+    const stdout = normalize(result.stdout).split('\r\n') // windows uses CRLF for carriage returns
 
     // windows returns a leading carriage return, remove it
```
flotwig, line 325: we aren't doing this anymore right?

```suggestion
```

flotwig (changes requested):

packages/app/cypress/e2e/specs_list_actual_git_repo.cy.ts
```diff
@@ -13,21 +13,33 @@ describe('Spec List - Git Status', () => {
...
+      ctx.fs.writeFileSync(
+        ctx.path.join(ctx.currentProject!, 'cypress', 'e2e', 'dom-container.spec.js'),
+        '// modifying the spec.',
+      )
+    })
+
+    // should update via GraphQL subscription, now the status is modified.
+    cy.get('[data-cy-row="dom-container.spec.js"]')
+    .contains('Modified')
+    .get('[data-cy="git-status-modified"]')
```
tbiethman, line 43: Good call on the additional coverage :+1:

packages/data-context/src/sources/GitDataSource.ts
tbiethman: Guides + trusted windows tests would be a good place to start. Shared utils/libs to handle the translation when necessary could help too.

packages/app/cypress/e2e/runner/retries.mochaEvents.cy.ts
```diff
@@ -2,7 +2,7 @@ import { runSpec } from './support/spec-loader'
 import { runCypressInCypressMochaEventsTest } from './support/mochaEventsUtils'
 import { snapshots } from './retries.mochaEvents.snapshots'
 
-describe('src/cypress/runner retries mochaEvents', { retries: 0 }, () => {
+describe('src/cypress/runner retries mochaEvents', { retries: 0, defaultCommandTimeout: 7500 }, () => {
```
jennifer-shehane, line 5: @lmiller1990 Can we add comments for all these timeout increases, that they are for windows tests so people in the future will understand why this is increased. Ideally I’d like to understand why these are even necessary though.

lmiller1990: Ok - we can probably make these conditional potentially, and add a comment, until we find out what (if there is an actual reason) makes these slow in windows. My feeling is general windows slowness, and since these use cypress-in-cypress, which is a lot more resource intensive than the original `runIsolatedCypress`, the compounding is just slower on windows (everything is, even the unit tests in 9.x are almost 2x slower in windows).

I'll log an issue for this to be looked into.

lmiller1990: https://github.com/cypress-io/cypress/pull/21055/commits/167a43627b5989c19dd635df851940c34b82d0e6

Also made issue to track in Jira, I sent it in our internal chat.

lmiller1990:
@flotwig made a start on the platform differences guide: https://github.com/cypress-io/cypress/pull/21055/commits/61cdff4941150b61e9c3bb38a62ae64545e17faf

You might have some more ideas, let me know and we can add them (either now, or later date - this will be an ongoing effort).

circle.yml
```diff
@@ -1189,6 +1189,8 @@ jobs:
     parallelism: 1
     steps:
       - restore_cached_workspace
+      - run: mkdir -p ~/.ssh
+      - run: ssh-keyscan github.com >> ~/.ssh/known_hosts
```
ryanthemanuel, line 1193:
```suggestion
```

guides/remaining-platform-agnostic.md
```diff
@@ -0,0 +1,85 @@
+# Remaining Platform Agnostic
```
flotwig, line 1: "Writing Cross-Platform JavaScript" might be a clearer title, or something along those lines? Thanks for adding this, now we have something to point to.

guides/remaining-platform-agnostic.md
```diff
@@ -0,0 +1,85 @@
...
+
+```json
+{ 
+  "scripts": {
+    "clean": "rm -rf **/*.js"
+  }
+}
+```
+
+This will stop developers on windows from running `yarn clean`. Instead, opt for a Node.js script where possible, or use a cross-platform Node.js module. In this case, we could use the [`rimraf`](https://www.npmjs.com/package/rimraf) module:
```
flotwig, line 57:
```suggestion
This will stop developers on Windows from running `yarn clean` unless they are specifically using a POSIX shell (like Git Bash). Instead, opt for a Node.js script where possible, or use a cross-platform Node.js module. In this case, we could use the [`rimraf`](https://www.npmjs.com/package/rimraf) module:
```

packages/app/src/specs/SpecsList.vue
```diff
@@ -67,7 +67,7 @@
           :id="getIdIfDirectory(row)"
           :key="row.index"
           :data-cy="row.data.isLeaf ? 'spec-list-file' : 'spec-list-directory'"
-          :data-cy-row="row.data.data?.relative"
+          :data-cy-row="row.data.data?.baseName"
```
tbiethman, line 70: Just wanted to note that using baseName here doesn't guarantee uniqueness like relative did. Doesn't seem like it impacts any current tests, though.

packages/server/lib/browsers/browser-cri-client.ts
```diff
@@ -105,19 +111,20 @@ export class BrowserCriClient {
    * @returns the chrome remote interface wrapper for the target
    */
   attachToTargetUrl = async (url: string): Promise<CRIWrapper.Client> => {
-    debug('Attaching to target url %s', url)
-    const { targetInfos: targets } = await this.browserClient.send('Target.getTargets')
+    return retryWithIncreasingDelay(async () => {
```
tbiethman, line 114: Can we add some comments here to document why the retries may be necessary?

packages/server/lib/browsers/browser-cri-client.ts
```diff
@@ -105,19 +111,20 @@ export class BrowserCriClient {
...
+      debug('Attaching to target url %s', url)
+      const { targetInfos: targets } = await this.browserClient.send('Target.getTargets')
 
-    const target = targets.find((target) => target.url === url)
+      const target = targets.find((target) => target.url === url)
 
-    // TODO: Move this into packages/error post merge: https://github.com/cypress-io/cypress/pull/20072
```
tbiethman, line 113: Is this the TODO worth persisting/doing?

packages/app/src/specs/SpecsList.vue
lmiller1990: Yep - I think it's probably fine for now, actually updating the paths is a bit of scope creep, and changing the paths dynamically for the assertion seems like more technical debt. This seems like a reasonable trade-off.

guides/remaining-platform-agnostic.md
lmiller1990: Updated!

lmiller1990:
I have responded to the comments

- updated guide
- updated guide name
- added some comments to explain what code does

guides/writing-cross-platform-javascript.md
```diff
@@ -0,0 +1,85 @@
+# Writing Cross-Platform JavaScript
+
+Cypress works on Linux, macOS and Windows. This includes both installing from npm, as well as for local development. Code should be written in a platform agnostic style.
+
+## Handling File Paths
+
+Throughout the code base, we access the file system in various ways, and need be concious of how we do so to ensure Cypress can be used and developed seemlessly on multiple platforms. One thing to keep in mind is file paths and file separators. macOS and Linux systems that use `/`, and Windows which uses `\`.
```
ryanthemanuel, line 7:
```suggestion
Throughout the code base, we access the file system in various ways, and need to be conscious of how we do so to ensure Cypress can be used and developed seamlessly on multiple platforms. One thing to keep in mind is file paths and file separators. macOS and Linux systems use `/`, and Windows uses `\`.
```

guides/writing-cross-platform-javascript.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+As a general rule, we want to use **native paths** where possible. There are a few reasons for this. Where ever we display a file path, we want to use the native file separator, since that is what the user will expect on their platform. In general, we can use the Node.js `path` module to handle this:
```
ryanthemanuel, line 10:
```suggestion

As a general rule, we want to use **native paths** where possible. There are a few reasons for this. Whereever we display a file path, we want to use the native file separator, since that is what the user will expect on their platform. In general, we can use the Node.js `path` module to handle this:
```

packages/server/lib/browsers/browser-cri-client.ts
ryanthemanuel: This fix actually negates the rationale for the TODO (which is nice). This scenarios is now handled by this catch block:

https://github.com/cypress-io/cypress/pull/21055/files#diff-b5b66d85423fa714add2e866448d3411b0f1cdf2cb4b4a95aa80394fa3edd1b3R57

which properly displays the appropriate error.

tbiethman (approved):

PR #21057, opened 2022-04-13
fix: resolves correctly specPattern

estrada9166:
- Closes  [UNIFY-1461](https://cypress-io.atlassian.net/browse/UNIFY-1461)

### User facing changelog

The default `specPattern` was being resolved as it was coming from the user config file which was incorrect; the reason why it was being resolved from the config file was that when trying to search for the default value, it was never found because it was nested in the testing type - and we are merging the testing type into the top-level config, creating a mismatch between the defaults values and the config itself

### Before
image

### After
image

### PR Tasks


packages/app/cypress/e2e/settings.cy.ts
```diff
@@ -232,6 +232,35 @@ describe('App: Settings', () => {
...
+      cy.findByText('Project Settings').click()
+      cy.get('[data-cy="config-legend"]').within(() => {
+        cy.get('.bg-gray-50').contains('default')
+        cy.get('.bg-teal-100').contains('config')
+        cy.get('.bg-yellow-100').contains('env')
+        cy.get('.bg-red-50').contains('cli')
+      })
+
+      cy.get('[data-cy="config-code"]').within(() => {
+        cy.get('.bg-gray-50').contains('cypress/e2e/**/*.cy.{js,jsx,ts,tsx}')
```
lmiller1990, line 253: I wonder if these should be data selectors, like `[data-cy-config="resolved"]`. I do not like using CSS much for test selectors, it's super hard to read and tends to be brittle. What do you think?

packages/server/lib/config.ts
```diff
@@ -138,7 +138,11 @@ export function mergeDefaults (
     config.baseUrl = url.replace(/\/\/+$/, '/')
   }
 
-  const defaultsForRuntime = configUtils.getDefaultValues(options)
+  let defaultsForRuntime = configUtils.getDefaultValues(options)
+
+  // As we normalize the config based on the selected testing type, we need
+  // to do the same with the default values to resolve those correctly
+  defaultsForRuntime = { ...defaultsForRuntime, ...defaultsForRuntime[options.testingType] }
```
lmiller1990, line 145: Interesting - kind of surprised we need to do this, I thought this was a UI only bug - so was the bug more deeper in the system, too?

The config stuff is so spread out all over the place, probably why we have some many bugs around it :(

lmiller1990 (commented):
One recommendation and one general question

packages/server/lib/config.ts
emilyrohrbough: +1 all configuration things should prob be happening in the config package to keep this logic in one place and so we can make it easily testable.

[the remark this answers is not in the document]
emilyrohrbough: Below we spread these values with the ignore type. Was the config overriding the default resolved values? Are we able to add a unit test for this?

lmiller1990 (approved):
Thanks for the quick follow up... :heavy_check_mark:

packages/server/lib/config.ts
```diff
@@ -138,7 +138,14 @@ export function mergeDefaults (
...
 
-  const defaultsForRuntime = configUtils.getDefaultValues(options)
+  let defaultsForRuntime = configUtils.getDefaultValues(options)
+
+  // As we normalize the config based on the selected testing type, we need
+  // to do the same with the default values to resolve those correctly
+  defaultsForRuntime = {
+    ...defaultsForRuntime,
+    ...defaultsForRuntime[options.testingType],
+  }
```
flotwig, line 148: imo, this logic would be more suitably moved to `packages/config/src/index.ts` where `getDefaultValues` is defined. this is the only usage of that function, so IMO it would make sense to confine this logic to the `config` package. what do you think @estrada9166 ?

flotwig (changes requested):

packages/server/lib/config.ts
estrada9166: Sounds good! yeah, I'm gonna update it!

flotwig (approved):

PR #21062, opened 2022-04-13
refactor: cleanup for ported vue2 module

tgriesser:
### User facing changelog

n/a

### Additional details
- Removes example projects / Cypress tests from vue2, these aren't being used (the Cypress tests still have a `cypress.json`) & we have decided all tests should be run in the `system-tests` folder
- Renames `cypress-vue` -> `cypress-vue2` for the bundled files for clarity: ec2a259be54b78aed2ea20ce242eba8934ffd7bc
- Fixes the ordering of CI build steps which currently seems to be broken in 10.0: 423b78144fac64eeb2607078c4cdedcf213d5b51
- Removes unused code based on the new patterns: b169c9e05b39c9277d31e81c8cabb294b5ee33ea

### PR Tasks


npm/vue2/cypress.json
```diff
@@ -1,9 +0,0 @@
-{
```
BlueWinds, line 1: Are we sure we don't want to retain at least some tests over this module? I agree it makes sense not to duplicate everything (especially the examples), but deleting the entire test setup feels a bit much. This is still going to be published as a separate package on NPM that we have to maintain, and I'm not sure we have test coverage of it other than the deleted files.

PR #21069, opened 2022-04-13
feat: add next preset to webpack-dev-server-fresh

ZachJW34:
- Closes [UNIFY-1400](https://cypress-io.atlassian.net/browse/UNIFY-1400)

### User facing changelog
Add next support for new Object API in `webpack-dev-server-fresh` 

### Additional details
Mostly a lift and shift from `npm/react/plugins` for next specific webpack code. Since Next ships it's own version of webpack, I had to change how we source webpack deps to account for this. The most notable change is the removal of Next v10 support which is two major versions behind the current Next v12. I also changed how we detect if `swc` is being used. It was accounting for the possibility of being used on `9.x` but we don't have that requirement for the new packages.

https://user-images.githubusercontent.com/25158820/163234632-a11e5a96-dc25-4e3b-8d47-9c855035b194.mov

### PR Tasks


ZachJW34:
Need to update the fixture with the new `data-cy-root` selector.

npm/webpack-dev-server-fresh/src/helpers/nextHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+ */
+export function findPagesDir (projectRoot: string) {
+  // prioritize ./pages over ./src/pages
+  let pagesDir = path.join(projectRoot, 'pages')
+
+  if (existsSync(pagesDir)) {
+    return pagesDir
+  }
+
+  pagesDir = path.join(projectRoot, 'src/pages')
```
lmiller1990, line 117: Should it be `path.join(projectRoot, 'src', 'pages')` to use the correct separator on windows?

npm/webpack-dev-server-fresh/src/helpers/sourceRelativeWebpackModules.ts
```diff
@@ -94,30 +98,67 @@ export function sourceRelativeWebpackModules (config: WebpackDevServerConfig) {
...
-        require.resolve('@cypress/webpack-batteries-included-preprocessor', {
-          paths: [__dirname],
-        }),
-      ],
-    })
-  }
+    let webpack5 = true
+    const importPath = path.dirname(webpackJsonPath)
+    const webpackModule = require(path.join(importPath, 'webpack.js'))
+
```
lmiller1990, line 113: This entire webpack module file is getting to be really complex really fast, I wonder if we can find some cleaner way to handle this eventually...

npm/webpack-dev-server-fresh/src/helpers/nextHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+import type { Configuration } from 'webpack'
+import * as fs from 'fs'
+import * as path from 'path'
+
+type PresetHandler = Omit<CreateFinalWebpackConfig, 'frameworkConfig'>
+
+const debug = debugLib('cypress:webpack-dev-server-fresh:nextHandler')
+
+export async function nextHandler ({ devServerConfig, sourceWebpackModulesResult }: PresetHandler) {
+  const webpackConfig = await loadWebpackConfig({ devServerConfig, sourceWebpackModulesResult })
```
lmiller1990, line 12: This look really thorough, nice job -- it's a little challenging to fully understand what case each branch handles, might be a good candidate for some fairly verbose comments?

lmiller1990 (approved):
Code seems fine and tests pass - I think some comments might be useful, considering how many different branches there are for the next.js handler.

npm/webpack-dev-server-fresh/src/helpers/sourceRelativeWebpackModules.ts
ZachJW34: I agree, @tgriesser might have some good ideas on cleaning this up. I think we should allow the handlers to source the information so we don't have the framework bloat in this file, and `sourceRelativeWebpackModules.ts` could be helper functions that the handlers consume, but not going to refactor in this PR.

npm/webpack-dev-server-fresh/src/helpers/nextHandler.ts
ZachJW34: Added some more comments throughout

npm/webpack-dev-server-fresh/src/devServer.ts
```diff
@@ -137,6 +138,10 @@ devServer.create = async function (devServerConfig: WebpackDevServerConfig) {
       case 'vue-cli':
         frameworkConfig = vueCliHandler({ devServerConfig, sourceWebpackModulesResult })
         break
+
```
BlueWinds, line 141: Just a style thing, with big switch statements like this, I always prefer having them out in their own function, so you can just `return` from each branch, a'la
```
getConfig = async (framework) => {
  switch (framework) {
    case 'react':
      return
    case 'nuxt':
      return await nuxtHandler({ devServerConfig, sourceWebpackModulesResult })
    ...etc...
  }
}

let frameworkConfig = await getConfig(devServerConfig.framework)
```

npm/webpack-dev-server-fresh/src/helpers/nextHandler.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    debug('found options next.js watchOptions.ignored %O', webpackConfig.watchOptions.ignored)
+  }
+
+  return webpackConfig
+}
+
+/**
+ * Acquire the modules needed to load the Next webpack config
```
BlueWinds, line 33: I'm not really following this comment. Acquire them how, for what, why, from where?

npm/webpack-dev-server-fresh/src/helpers/sourceRelativeWebpackModules.ts
BlueWinds: Agreed, this stuff is all super ugly. It makes me want to have a `next-dev-server` separate from our `webpack-dev-server`, since it's requiring so much custom configuration / patching.

BlueWinds: Or at the very least its own file / function in `webpack-dev-server`. `webpackConfig = framework === 'nextjs') ? getNextWebpackConfig() : getWebpackConfig()`, and they can share utility functions as needed.

BlueWinds: The more I look at it the more it feels like something we should decide now rather than later, pull it out in some fashion rather than kick it down the road.

BlueWinds (changes requested):

npm/webpack-dev-server-fresh/src/helpers/sourceRelativeWebpackModules.ts
ZachJW34: I'm going to refactor this! I'm sidetracked getting the dev-servers to work with the built binary but going to return

tgriesser: Yeah I think the next stuff should be handled separately than the regular one. 

The first attempt here was intentionally procedural, until we knew what we needed to reuse for next, it didn’t make sense to abstract.

ZachJW34: Refactored, broke the functions up so they could be reused and altered the handler contract so that each handler is responsible for providing the sourced deps.

npm/webpack-dev-server-fresh/src/devServer.ts
```diff
@@ -111,36 +112,39 @@ export function devServer (devServerConfig: WebpackDevServerConfig): Promise<Cyp
...
+async function getPreset (devServerConfig: WebpackDevServerConfig): Promise<PresetHandlerResult> {
+  switch (devServerConfig.framework) {
+    case 'create-react-app':
+      return createReactAppHandler(devServerConfig)
+    case 'nuxt':
+      return await nuxtHandler(devServerConfig)
+
+    case 'vue-cli':
+      return await vueCliHandler(devServerConfig)
```
lmiller1990, line 125: No need `await here.

lmiller1990 (approved):
One comment around unnecessary await.

Any other blocker? Let's get this merged CT Arch and this PR should be #1 priority.

npm/webpack-dev-server-fresh/src/devServer.ts
tgriesser, line 125:
```suggestion
      return vueCliHandler(devServerConfig)
```

npm/webpack-dev-server-fresh/src/devServer.ts
ZachJW34: Addressed in ac29f02aa37dc2f16e412f546093b9f48e0cf710

npm/webpack-dev-server-fresh/src/helpers/nextHandler.ts
ZachJW34: Addressed in ac29f02aa37dc2f16e412f546093b9f48e0cf710

BlueWinds (approved):

PR #21070, opened 2022-04-13
fix: scaffold examples with correct ext based if project has TS

estrada9166:
- Closes  [UNIFY-1451](https://cypress-io.atlassian.net/browse/UNIFY-1451)

### Additional details

Scaffold the examples using the correct extension, checking if the project has TS, and use `.ts` extension to be consistent with the project.

In this case, it's using the same logic as when creating an empty spec from the app, we identify if the project has TS and we suggest the extension for the files.

### PR Tasks


lmiller1990 (approved):

flotwig (changes requested):
I feel like this change needs a little bit more thought than just renaming `.js` to `.ts`. `cypress-example-kitchensink` isn't written with types in mind nor is it typechecked at all, and this PR does not add typechecking for the converted files... so it's going to be really easy for us to accidentally be generating broken `.ts` without knowing it.

I think to fix this, there are ~~two~~ three options:

1. Convert `cypress-example-kitchensink` to typescript and build a JS version from that
	* This might not be possible to do without TS injecting boilerplate, which would make this a non-starter
2. Convert `cypress-example-kitchensink` to typescript, and update this PR to rename `cy.ts` to `cy.js` (reversed from how it is now)
	* Only problem here is that TS has syntax that is invalid JS, so you'd have to also lint that no TS-specific constructs are used (type annotations, `type` keywords...)
3. Add a test in this PR that typechecks the converted `.ts` files against the `cypress` typedefs.

(3) would probably be easiest. (2) would need to be a bigger decision involving DX I think.

lmiller1990:
> I feel like this change needs a little bit more thought than just renaming `.js` to `.ts`. `cypress-example-kitchensink` isn't written with types in mind nor is it typechecked at all, and this PR does not add typechecking for the converted files... so it's going to be really easy for us to accidentally be generating broken `.ts` without knowing it.
> 
> I think to fix this, there are ~two~ three options:
> 
>     1. Convert `cypress-example-kitchensink` to typescript and build a JS version from that
>        
>        * This might not be possible to do without TS injecting boilerplate, which would make this a non-starter
> 
>     2. Convert `cypress-example-kitchensink` to typescript, and update this PR to rename `cy.ts` to `cy.js` (reversed from how it is now)
>        
>        * Only problem here is that TS has syntax that is invalid JS, so you'd have to also lint that no TS-specific constructs are used (type annotations, `type` keywords...)
> 
>     3. Add a test in this PR that typechecks the converted `.ts` files against the `cypress` typedefs.
> 
> 
> (3) would probably be easiest. (2) would need to be a bigger decision involving DX I think.

I agree (1) and (2) probably are not ideal and/or a lot of work for not (that) much value. For (3), we could just throw a `/// <reference types="cypress" />` at the top and run `tsc --noEmit` at the tail end of this test to ensure the generated code is valid.

estrada9166:
Closing this one - we can revisit if needed

PR #21063, opened 2022-04-13
fix(unify): Remove 'Reconfigure' dropdown from Testing Type chooser

davidmunechika:
- Closes [UNIFY-1533](https://cypress-io.atlassian.net/browse/UNIFY-1533?atlOrigin=eyJpIjoiYjg0YjJkZjYzNmUyNDUwOThhYmJiZGM0NzNkNjJmNTYiLCJwIjoiaiJ9)

### User facing changelog
Remove the ‘Reconfigure’ dropdown from testing type chooser.
Update: this also removes the 'Choose a Browser' dropdown from testing type chooser since the functionality of this doesn't work as intended.

### Additional details

Previous version:

Screenshot 2022-04-13 at 10 49 35 AM

Updated version:

Screenshot 2022-04-13 at 10 42 55 AM

### PR Tasks

- [ ] Have tests been added/updated?
- [ ] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [n/a] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [n/a] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [n/a] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

PR #21077, opened 2022-04-13
feat: replace reconfigure button on settings page with link to config doc

rachelruderman:
- Closes UNIFY-1532

### User facing changelog
n/a

### Additional details
Since we have removed reconfiguring the testing type from the 10.0 release, the `Reconfigure Project` button in the Settings page should link to the `Configuration Guide` in the Cypress docs instead.

**Before:**
Screen Shot 2022-04-14 at 9 08 13 AM

**After:** 
Screen Shot 2022-04-14 at 9 09 09 AM

### PR Tasks


rachelruderman:
Note: I haven't marked this for a release since it is not technically user-facing. This UI change is not considered user-facing since the UI it changes has not been released yet. Plz let me know if you think differently 😃

davidmunechika (approved):
Looks good- i think just update percy snapshots? https://percy.io/cypress-io/cypress/builds/17367227/changed/974720034?browser=chrome&browser_ids=23&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=900&widths=900%2C1000

packages/app/src/settings/SettingsContainer.cy.tsx
```diff
@@ -50,4 +50,11 @@ describe('<SettingsContainer />', { viewportHeight: 800, viewportWidth: 900 }, (
 
     cy.findByText(defaultMessages.settingsPage.projectId.title).should('not.exist')
   })
+
+  it('renders footer with CTA button', () => {
```
rachelruderman, line 54: I used a component test instead of an e2e test to get some experience with CT 😄 in general, when both approaches are equal, do we have a preference for one or the other?

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -410,8 +410,9 @@
       }
     },
     "footer": {
-      "text": "You can reconfigure the settings for this project if you're experiencing issues with your Cypress configuration.",
-      "button": "Reconfigure Project"
+      "text": "You can reconfigure the component testing settings for this project if you're experiencing issues with your Cypress configuration.",
```
ZachJW34, line 413:
```suggestion
      "text": "You can reconfigure the {testingType} testing settings for this project if you're experiencing issues with your Cypress configuration.",
```

packages/app/src/settings/SettingsContainer.vue
```diff
@@ -43,7 +43,7 @@
       variant="outline"
```
ZachJW34, line 43: Above for the `t('settingsPage.footer.text')` should be modified to include the testing type since it's currently wired to only show component. Something like:
```
const footerText = computed(() => t('settingsPage.footer.text', { testingType: props.gql.currentProject?.currentTestingType === 'component' ? 'component' : 'E2E' }))
```

Then you can render `{{ footerText }}`

ZachJW34 (changes requested):
There are two PRs removing the ability to reconfigure a project, this one and https://github.com/cypress-io/cypress/pull/21063 @davidmunechika. There is some backend code that should be cleaned up as well since it's now dead code, I think it should be removed in one of these PRs (whichever gets merged last) or a ticket should be created and a fast follow.

packages/app/src/settings/SettingsContainer.cy.tsx
ZachJW34: For verifying text content I prefer to keep it in the component tests.

packages/app/src/settings/SettingsContainer.vue
rachelruderman: Great callout!

packages/app/src/settings/SettingsContainer.cy.tsx
```diff
@@ -50,4 +50,11 @@ describe('<SettingsContainer />', { viewportHeight: 800, viewportWidth: 900 }, (
...
+    mountSettingsContainer()
+    cy.contains('p', defaultMessages.settingsPage.footer.text)
```
ZachJW34, line 56:
```suggestion
    cy.contains('p', defaultMessages.settingsPage.footer.text.replace('{testingType}', 'E2E'))
```

ZachJW34 (approved):
Nice work!

PR #21081, opened 2022-04-14
chore: making the npm deps for vue, react, and vue2 use 0.0.0-dev

JessicaSachs:
I changed the dependency on mount-utils from **`*`** to **`file:../mount-utils`** to fix what I thought was an issue with pulling in the right version of mount-utils.

This was originally introduced in a PR earlier today. For whatever reason, CI wasn't having these issues. https://github.com/cypress-io/cypress/commit/0e7b555f93fb403f431c5de4a07ae7ad6ac89ba2

I'm going to try reverting the change back to **`"@cypress/mount-utils": "*"`**.

image


lmiller1990:
No dice

tgriesser:
Pretty sure we actually need to be using `0.0.0-development` for local deps like this in the workspace. Updated to this and it looks like things are passing

BlueWinds (approved):

PR #21087, opened 2022-04-14
fix: running a new test after already having run tests

ryanthemanuel:
- Closes UNIFY-1520 and UNIFY-1462

### User facing changelog

Users can now execute a test in open mode, go back to the specs page, add a new test, and run that newly added test without error

### Additional details

Previously, we had a watcher that was only watching the routing changes in order to set the active spec. The problem is that this was caching the specs list so if we had a stale specs list it wasn't going to be able to find the newly added spec. We fix this by ensuring that any active spec management is done inside of a watchEffect so that we are watching changes to all things relevant.

### PR Tasks


packages/app/src/runner/SpecRunnerContainerOpenMode.vue
```diff
@@ -22,24 +22,21 @@ const specStore = useSpecStore()
 const router = useRouter()
 const route = useRoute()
 
-const { initialized, watchSpec } = useUnifiedRunner()
-
 const specs = computed(() => {
   return props.gql.currentProject?.specs ?? []
 })
 
-watchSpec(specs)
```
lmiller1990, line 31: Not a strong opinion but I like to functions that call `watch` or `watchEffect` in the actual component file to give visibility, or you need to dig through a bunch of other files to find the watcher. It's like when you are trying to track down a `useState` or `useEffect` nested deeply in a hook in React.

I don't mind too much, but something about `watchSpec(specs)` is more readable to me than `useUnifiedRunner(specs)`. The former is more clear that something reactive will happen. Happy to field opinions and not really a big deal either way.

packages/app/src/runner/unifiedRunner.ts
```diff
@@ -20,45 +19,39 @@ export function useUnifiedRunner () {
...
+    if (activeSpecInSpecsList && specStore.activeSpec?.relative !== activeSpecInSpecsList.relative) {
+      specStore.setActiveSpec(activeSpecInSpecsList)
+    } else if (!activeSpecInSpecsList) {
+      specStore.setActiveSpec(null)
+    }
+  })
 
-        if (selectorPlaygroundStore.show) {
-          const autIframe = getAutIframeModel()
+  watch(() => getPathForPlatform(route.query.file as string), (newQueryFile) => {
```
lmiller1990, line 44: We have two things watching `route.query.file`, line 26 and line 44. I wonder if we should combine them so all the watching of the same variable is in the same function? There's an argument of "one watch per effect" so again I don't mind either way, but wanted to point this out since you might not have noticed the double watcher.

I'm guessing the perf overhead is negligible so just wanted to call this out to see if it was intentional or not.

I tried the single watcher and it seems to work just fine, too: https://github.com/cypress-io/cypress/pull/21112

lmiller1990 (commented):
Left two comments, neither are really blockers but wanted to get some opinions - the readability of this part of the code base is really important since there's so much going on.

marktnoonan (commented):
Nice, having tested a bit this works well. The only drawback to the `watchEffect` that I can think of is that it runs more times than strictly necessary, since it sets values that it also watches. But there don't seem to be any negative side effects of that, and the simplicity of gathering this all into the composable is nice. I just left one comment about an unused param.

As for the problem itself:

> Previously, we had a watcher that was only watching the routing changes in order to set the active spec. The problem is that this was caching the specs list so if we had a stale specs list it wasn't going to be able to find the newly added spec. 

I was super curious about this because I couldn't grok why it would be happening, so I looked around a little bit. It seems that the route watcher's callback was firing twice - once with a reference to the _correct_ specs list, exactly as we'd want, and then again with the _out of date_ specs list, and it's that second run that triggers the redirect + error.

The watcher runs twice because when you first visit the spec runner calls `getPathForPlatform` for the new and old values of `route.query.file`. In running the watcher's callback for the _old_ value (which is just `undefined`, then `null`), Vue is using the old value of the specs list, but somehow that callback runs second. So, before this PR, we briefly passed through a correct "spec found" state before it gets overwritten by a "not found" state and bounces us to the specs page. That's probably an oversimplification but I'm trying to avoid going down a rabbit hole. 

We could fix the issue by making sure the watcher doesn't run twice with a coupe of checks on the new/old value parameters, I tried it out and it works fine, but I still feel the changes in this PR make it all more readable (to @lmiller1990's point).

I'd be fine with `useUnifiedRunnerAndWatchers` or something as the name to address the discoverability of these watchers.

I'm sure we can tighten this up even further at some point but it's probably not very valuable to do that unless we see perf issues.

packages/app/src/runner/unifiedRunner.ts
ryanthemanuel: Yeah I played around with combining them at one point and ran into what @marktnoonan is talking about. Specifically I think it's these tests failing on #21112: https://app.circleci.com/pipelines/github/cypress-io/cypress/36466/workflows/828e7886-b042-41de-8753-e693c6f9a292/jobs/1460784

There may be a way to tweak the logic to fix that, but ultimately, I'm ok with keeping them separate since I do like the idea of "one watch per effect"

tbiethman: "one watch per effect" seems like a reasonable approach to take until it becomes a performance issue. I prefer the sensible encapsulation over worrying about the lifecycle too much.

packages/app/src/runner/SpecRunnerContainerOpenMode.vue
tbiethman: I'd prefer adding 'watcher' to the composable's name in some way if more discoverability is desired. Putting the responsibility of activating the watchers on the consuming component just leaves a logic hole where there doesn't need to be one.

tbiethman (approved):
Works great! Any of the proposed naming/API solutions would work for me, though I stated my preferences :P

packages/app/src/runner/unifiedRunner.ts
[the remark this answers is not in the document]
ryanthemanuel: Removed: [cypress-io/cypress@`98c813d` (#21087)](https://github.com/cypress-io/cypress/pull/21087/commits/98c813d91db2b69e395b6dff5962475f91f72f49)

marktnoonan:
@ryanthemanuel It looks like moving it into the `watchSpecs` function has introduced a problem - unless this is actually flaky, which seems unlikely for a brand new test: https://app.circleci.com/pipelines/github/cypress-io/cypress/36484/workflows/3f24cac4-46d2-424a-ae8c-b9db7a492f12/jobs/1461338

Video shows us doing the right behavior, but the order of things has changed.

lmiller1990 (approved):

tbiethman (approved):

PR #21090, opened 2022-04-14
chore: Move component-index generation to scaffold-config package

BlueWinds:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1433

### User facing changelog
None. Internal change only.

### Additional details
We wanted to move the template for component-index.html to live with the rest of the framework-specific code. This is general code cleanup related to the ct-archecture changes.

Only effect is a small fix to remove a couple of blank lines from the generated `component-index.hmtl`.

### PR Tasks


packages/scaffold-config/src/frameworks.ts
```diff
@@ -122,6 +123,7 @@ export const WIZARD_FRAMEWORKS = [
     codeGenFramework: 'react',
     glob: '*.{js,jsx,tsx}',
     mountModule: 'cypress/react',
+    componentIndexHtml: componentIndexHtmlGenerator(),
```
lmiller1990, line 126: I like this approach. It seems like there's duplication (there is) but I think it's actually somewhat ideal.  That said, if you forgot to add `componentIndexHtml`, the only way we'd know is via a type error.

One thing we could do is add a test that loops each of these, asserting the output, like

```
for (const framework of WIZARD_FRAMEWORKS) {
  const actual = framework.componentIndexHtmlGenerator()
  if (framework === 'nextjs') {
    // specific assertion
  } else {
    expect(actual).to.eq(<generic_template>)
  }
}
```

Do you think this would be useful to have, or the types are enough to keep us safe and confident? I could go either way.

packages/scaffold-config/src/component-index-template.ts
```diff
@@ -0,0 +1,28 @@
...
+      <body>
+        ${bodyModifier}
+        <div data-cy-root></div>
+      </body>
+    </html>
+  `
+
+    // If the framework returns an empty string for either of the modifiers,
+    // strip out the empty lines
+    return template.replace(/\n {4}\n/g, '\n')
```
lmiller1990, line 24: I wonder if we should be using https://nodejs.org/api/os.html#oseol

```sh
os.EOL // =>
    \n on POSIX
    \r\n on Windows
```

I think this would fail on windows?

lmiller1990 (commented):
Minor comment, mainly around keeping the tests cross platform - I think we'd need to use `os.EOL` to make the test you've added run on windows, although many of our tests fail on windows, it mightn't be a big deal.

packages/scaffold-config/src/component-index-template.ts
```diff
@@ -0,0 +1,28 @@
+import dedent from 'dedent'
+
+const componentIndexHtmlGenerator = (headModifier: string = '', bodyModifier: string = '') => {
```
ZachJW34, line 3: Let's remove the bodyModifier, no point in having it if we never use it.

packages/scaffold-config/src/frameworks.ts
```diff
@@ -181,6 +185,7 @@ export const WIZARD_FRAMEWORKS = [
...
```
ZachJW34, line 188:
```suggestion
    componentIndexHtml: componentIndexHtmlGenerator('<div id="__next_css__DO_NOT_USE__"></div>')
```

We have the `react` and `next` templates swapped.

packages/scaffold-config/src/frameworks.ts
```diff
@@ -257,5 +265,6 @@ export const WIZARD_FRAMEWORKS = [
     codeGenFramework: 'react',
     glob: '*.{js,jsx,tsx}',
     mountModule: 'cypress/react',
+    componentIndexHtml: componentIndexHtmlGenerator('<div id="__next_css__DO_NOT_USE__"></div>'),
```
ZachJW34, line 268:
```suggestion
    componentIndexHtml: componentIndexHtmlGenerator(),
```

ZachJW34 (changes requested):

packages/scaffold-config/src/frameworks.ts
BlueWinds: I think that relying on types is fine - if they're not here to catch issues like this, then why do we even have them? It's also implicitly covered by the system tests (though I wouldn't want to rely on them to cover this feature, they are a backstop against the types as well).

packages/scaffold-config/src/component-index-template.ts
BlueWinds: Hm. I don't think this is the right place for that - if we scaffold files using OS EOL separators, then that would be done in the `scaffoldFile` method in `data-context/src/actions/WizardActions.ts`. Then it would be in one place, applying to all files we scaffold to disk.

I don't think we should do that, though. Several other methods in that same file (`scaffoldFixtures` and `scaffoldConfig` for example) use `\n` as well. It'd be a bit of a change.

packages/scaffold-config/src/component-index-template.ts
BlueWinds: Done. Super easy to add  back later if need be.

packages/scaffold-config/src/frameworks.ts
BlueWinds: Whoops. Fixed in upcoming commit.

ZachJW34: This was added to `nuxt` instead `next`!!

BlueWinds: Fix for realzies this time! ^^;;

ZachJW34 (approved):

lmiller1990 (approved):
Looks good, let's ship!

PR #21091, opened 2022-04-14
chore: add dev-servers as deps to server to be included in the binary

ZachJW34:
### User facing changelog
Bundle `@cypress/webpack-dev-server-fresh` and `@cypress/vite-dev-server-fresh` in the binary.

### Additional details
Add `@cypress/webpack-dev-server-fresh` and `@cypress/vite-dev-server-fresh` to dependencies of `packages/server/package.json`. This ensures they are bundled in the binary.

Refactored how `vite` is imported in `@cypress/vite-dev-server-fresh`. Since this package is now part of the binary, we need to `require.resolve` from the projectRoot to emulate how the package was previously resolved.

Fixed an issue where the disted files for `@cypress/webpack-dev-server-fresh` weren't being copied (thanks @tgriesser)

New Object API now works when testing with the binary.

### PR Tasks


npm/vite-dev-server-fresh/src/getVite.ts
```diff
@@ -0,0 +1,23 @@
+import debugFn from 'debug'
+import type { ViteDevServerConfig } from './devServer'
+
+const debug = debugFn('cypress:vite-dev-server:getVite')
+
+let vite: typeof import('vite')
```
tgriesser, line 6: I think this binding might technically work, but having this side-effect thing in the module feels super weird, can we just have `initVite` be named `getVite`

```ts
return require(viteImportPath)
```

and just pass the result `vite` module returned into the functions that use it like we do w/ webpack

tgriesser (changes requested):

npm/vite-dev-server-fresh/src/getVite.ts
ZachJW34: Yeah I can do that, thought the side-effect was cheeky but I'll update it, would be nice to have consistency between the dev-servers

tgriesser: Yeah I probably shoulda said "clever" instead of "super weird" haha, we've already overspent the "clever/weird" budget in the `require.resolve` stuff, gotta keep things simple somewhere 😛

ZachJW34: Updated in [bee4eb4](https://github.com/cypress-io/cypress/pull/21091/commits/bee4eb4609237702f2781025ce5c3e7e313644b1)

tgriesser (approved):

PR #21092, opened 2022-04-14
chore: remove unused dependencies

flotwig:
- Closes 

### Additional details

- Used depcheck to check some of our packages for unused packages.

### PR Tasks


packages/ui-components/package.json
```diff
@@ -28,15 +28,13 @@
...
     "mobx": "5.15.4",
     "mobx-react": "6.1.7",
     "prop-types": "15.7.2",
     "react": "16.8.6",
     "react-dom": "16.8.6",
-    "webpack": "4.41.6",
-    "webpack-cli": "3.3.11"
+    "webpack": "4.41.6"
   }
 }
```
emilyrohrbough, line 40:
```suggestion
}

```

packages/driver/cypress/fixtures/dropdown.html
```diff
@@ -1,44 +0,0 @@
-<!DOCTYPE html>
```
emilyrohrbough, line 1: Did we lose test coverage at some point ? did you look into when these became obsolete?

lmiller1990 (commented):
Cool. How did you figure these were unused? Did you use some kind of tool to do the analysis?

Seems like something we should be doing on a schedule, or possibly automatically.

flotwig:
@lmiller1990 It's a npm module called `depcheck`, it works okay. I think eventually we should add it to our `lint` step in CI so we can really button this up. It also notices when dependencies are referenced but not defined in `package.json`.

packages/driver/cypress/fixtures/dropdown.html
flotwig: Maybe my Git-fu is weak, but I was unable to tell when these files were ever being used for testing. They are over 5 years old.

ryanthemanuel (commented):
Nice work! Always nice to clean things up.

ryanthemanuel (approved):

emilyrohrbough (approved):
I spot checked quite a few of these and everything appears to be jusssst fine. Thanks for doing this! I'm always happy to remove dependencies 😄

PR #21096, opened 2022-04-15
chore: Refactor cy-in-cy mocha snapshots

BlueWinds:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1383

### User facing changelog
No user facing changes. Internal tests only.

### Additional details
When we migrated some tests from packages/runner to packages/app using cypress-in-cypress, our existing snapshot plugin broke. @lmiller1990 ported them into the `app` tests as modules, to preserve the snapshots for later consideration.

This is that later consideration. These snapshots were not providing much value, so rather than maintain tens of thousands of lines of snapshots, I've reduced them to assert only on the names and orders of the mocha events, which is what we really care about for these purposes anyway - the event names alone reveal whether the test was behaving as expected (with one exception, to which I added additional assertions).

I also created some minor infrastructure, to make it easy to update these snapshots as necessary, or add additional tests in the same format. They almost never change though (count on one hand the number of times in the history of the project), so a more thorough solution didn't seem necessary.

No user-facing changes.

### PR Tasks


lmiller1990 (approved):
Seems fine to me - obviously we lose some granularity, but I'm not sure how valuable snapshotting mocha internals really was.

brian-mann:
> I also created some minor infrastructure, to make it easy to update these snapshots as necessary, or add additional tests in the same format. They almost never change though (count on one hand the number of times in the history of the project), so a more thorough solution didn't seem necessary.

There was a very intentional reason to snapshot the internals. We serialize those values and send them over the wire and the snapshots were the lowest level guarantees that nothing unintentionally changed. They'd be an early warning system enabling you too see exactly what was being translated and sent, in response to making big changes like adding test retries.

Even though they didn't change that much, I can tell you from personal experience that over the lifetime of the project these area *did* change considerably, and there were other solutions in place that did similar things over time. This particular implementation was actually created in response to test retries, which came much later in its evolution - and the developer working on that feature utilized these extensively while building out that subsystem.

I'm a little remiss to delete these - the number of lines of code doesn't really matter IMO because the intention here is that these tests were never flaky or difficult to manage, and they'd reveal at the lowest and most sensitive level *anything* that changed the payloads.

I do understand better now that typescript types may give us the same kinds of data guarantees in the long run over these snapshots, but I'd suggest keeping them for now unless there are strong type guarantees on the payloads.

brian-mann (dismissed):
See my comment here: https://github.com/cypress-io/cypress/pull/21096#issuecomment-1100131278

Additionally I'm not sure if you were looking at the history of the *renamed* file or the original files. If you look at the history of the 3 original files you'll see that they have in fact changed as a result of new features, as they fixed bugs, and were updated in response to features that changed their payloads.

The 3 original files were this:

- https://github.com/cypress-io/cypress/blob/develop/packages/runner/__snapshots__/runner.mochaEvents.spec.js
- https://github.com/cypress-io/cypress/blob/develop/packages/runner/__snapshots__/retries.mochaEvents.spec.js
- https://github.com/cypress-io/cypress/blob/develop/packages/runner/__snapshots__/studio.mochaEvents.spec.js

And you can see the `git blame` for each file to get an idea of what was updated with each successive feature change or bugfix:

- https://github.com/cypress-io/cypress/blame/b7002bfd52b8e9db5b761ec571bba0a0524f3555/packages/runner/__snapshots__/runner.mochaEvents.spec.js
- https://github.com/cypress-io/cypress/blame/b7002bfd52b8e9db5b761ec571bba0a0524f3555/packages/runner/__snapshots__/retries.mochaEvents.spec.js
- https://github.com/cypress-io/cypress/blame/b7002bfd52b8e9db5b761ec571bba0a0524f3555/packages/runner/__snapshots__/studio.mochaEvents.spec.js

jennifer-shehane:
I would have to agree with brian on this one. These snapshots played an important part when refactoring larger parts of the mocha internals, especially during test retries. It is relatively low effort to keep them with the amount of time these are touched and high value for catching a critical bug when any portion of this is rewritten. 

I do think that snapshots are cumbersome to generate and update locally, for sure, and would be open to hearing more ways that we could improve upon this if possible.

brian-mann:
@jennifer-shehane the previous developer working on this actually built a custom button in the UI that would allow you to "check/uncheck" a box that would persist/update/save the snapshots when working to make the snapshot updating experience better. He'd be able to work on things locally and then click a button whenever he wanted the snapshot to persist which was very useful and clever. Simple ergonomic improvements like that went a long way to managing these.

lmiller1990:
Based on the above, maybe we just leave this as-is? It might be nice to get the "auto update snapshot" feature working again, but having some manual friction (copy and paste the new snapshot) when updating some critical files like these probably isn't the worst thing, either - it really forces you to think "do I really want to update these?"

BlueWinds:
> I would have to agree with brian on this one. These snapshots played an important part when refactoring larger parts of the mocha internals, especially during test retries. It is relatively low effort to keep them with the amount of time these are touched and high value for catching a critical bug when any portion of this is rewritten.
> 
> I do think that snapshots are cumbersome to generate and update locally, for sure, and would be open to hearing more ways that we could improve upon this if possible.

So I tend to really dislike these sorts of extremely granular snapshots. Assertions should, as far as practical, be on the desired behavior, rather than implementation details.

The more I can read a tests and see what the developer intended, the more confident I can be that my new version will retain the same behavior. When I look at the existing snapshots, it's hard to tell _what behavior it is testing_ - which parts of it are important, what in this giant snapshot ensures that `it('can retry from [afterEach]')`?

The new versions of these tests assert precisely on what they say they do - `it('can retry from [afterEach]')` asserts that tests can retry from `afterEach`. Relying on snapshots in `retry` tests is an unsafe way to gain that coverage.

That's what I've tried to do with this PR - the retries tests now assert, directly, that test retries occurred. They won't break from the addition or removal of features unrelated to retries.  If there are areas left uncovered, safety that the snapshots provided that we're now missing, we should add additional tests in those areas that explain exactly what they're asserting on.

I'm very much interested and willing to improve the assertions, or add additional tests - but I think large snapshots communicate poorly with future developers looking at them. Tests about testing retries failing because of features unrelated to test retries is a test-code smell.

BlueWinds:
(waiting on fixing test failures for further discussion if this is a direction we want to go; since PR may end up cancelled, not putting more work in right this sec)

lmiller1990:
I'd agree `it('can retry from [afterEach]')` might not be a good test description. It's more like `it('captures all internal mocha events and data when running retry from [afterEach]')`.

lmiller1990:
Should we close this for now? We can revisit it in the future, but I'm not sure it's moving forward right now. We could keep the branch alive, just close off the PR.

Issue #9029, opened 2020-10-29
Brotli files fail to load in Cypress 5.5


chasemgray:
### Current behavior
We've been trying to upgrade to Cypress 5.5 but it will not work with our website due to the fact that we serve javascript assets with brotli encoding and do not provide a gzip version. It appears this code is what is causing the issue:
https://github.com/cypress-io/cypress/blob/64a60f4bc1b0d055431400e81c55ba525ebda6ea/packages/proxy/lib/http/request-middleware.ts#L69-L82

I'm not sure why this is only an issue in 5.5 but everything has been working for us fine with brotli encoding up until now.

### Desired behavior
We should be able to load a javascript file for our website using brotli encoding.

### Test code to reproduce
Download any brotli file from within the opened Chrome window during a cypress run.

### Versions

Cypress 5.5

bahmutov:
Hi Chase, did it use to work previously? And if yes in which Cypress version?

chasemgray:
Yes, it works on all versions prior to 5.5. I tested 5.0, 5.1, 5.2, 5.3, and 5.4 today to verify.

jennifer-shehane:
The file referenced was not touched during the 5.5.0 release, so I'm not sure that's the source of the issue you're describing. 

Can you provide a reproducible example? That will help in getting this issue addressed more quickly. Thanks!

chasemgray:
It's possible it's not that file, but my suspicion is something using those functions, not the file itself.

Anyways, if you open a cypress controlled browser, and request the following brotli encoded file, it'll fail. It will succeed on previous versions of Cypress. This file is being served directly off S3 with the correct brotli content-encoding headers, etc.

https://aurora-v2.s3.amazonaws.com/vendors~ChartJS.0.111.0.js

chasemgray:
@jennifer-shehane is there any update on this? Were you able to reproduce it with the link I sent?

Issue #6197, opened 2020-01-20
Brotli encoding request fails (ERR_CONTENT_DECODING_FAILED)


backfromreality:
Requests with encoding content by Brotli compression algorithm always fails.

### Current behavior:

Request header:
Accept-Encoding: gzip, deflate, br

Response header: 
content-encoding: br

Result:
Request has status 200 OK, but failed with an error ERR_CONTENT_DECODING_FAILED.

### Desired behavior:

Request header:
Accept-Encoding: gzip, deflate, br

Response header: 
content-encoding: br

Result:
Request 200 OK. Files successfully decoded.

### Versions
Requests began to fail, when the encoding of our files was changed (2 month ago)
Bug exists in the last version of Cypress too (3.8.2)

jennifer-shehane:
@backfromreality Can you provide a repo with Brotli encoded files used in a request through Cypress tests? Providing a way for us to get this running quickly will be the fastest way to get your issue addressed.

backfromreality:
@jennifer-shehane I tried to start own app on localhost with brotli encoded files and visit localhost by cypress but got an error like this [#5446 ](https://github.com/cypress-io/cypress/issues/5446)
so I'll wait for fix

also found one more unclear thing for me
if we are visiting [google.co.in](https://www.google.co.in) the html file and some JS files have `content-encoding: br`
![Chrome](https://user-images.githubusercontent.com/48588120/73075252-2a280780-3ec4-11ea-9588-faf414311ec3.PNG)
when we are trying to visit [google.co.in](https://www.google.co.in) with cypress, these files have `content-encoding: gzip`
![Cypress](https://user-images.githubusercontent.com/48588120/73075253-2a280780-3ec4-11ea-9c84-bdfa2ff9635e.PNG)


[repo](https://github.com/backfromreality/brotli)

flotwig:
@backfromreality Cypress only currently supports injecting into non-compressed or gzip-compressed content. But Brotli could probably be added easily too since IIRC it's part of Node's standard lib.

backfromreality:
UPD:
(ERR_CONTENT_DECODING_FAILED) an error occurred due to incorrect server-side environment settings

You can close this issue, or rename and bind it to this comment https://github.com/cypress-io/cypress/issues/6197#issuecomment-578145654

egemon:
Same issue for me. It passes accept-encoding brotle but can't really decode it.

egemon:
[...]

Issue #535, opened 2017-06-14
Brotli encoded requests fail, strip unsupported encoding request headers


brian-mann:
Recently a competitor to `gzip` has emerged and webservers which send these response headers will fail when going through the Cypress proxy.

```
Content-Encoding: br
```

The quick (and temporary) fix for this is to simply strip out `br` from the `Accept-Encoding` request header. In later version we can fully support `br` encoding.

brian-mann:
Fixed in [0.19.3](https://github.com/cypress-io/cypress/wiki/Changelog#0193-06142017).

PR #21101, opened 2022-04-15
Feat/brotli deflate support

itsjfx:
- Closes the following issues:
  - https://github.com/cypress-io/cypress/issues/6197
  - https://github.com/cypress-io/cypress/issues/9029
 - and implements `Brotli` as suggested in https://github.com/cypress-io/cypress/issues/535

### User facing changelog

Does this count as user facing? If so...

Support `Brotli` and `deflate` as content-encoding methods for increased compatibility and performance.

### Additional details

<https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding>
<https://en.wikipedia.org/wiki/Brotli>

- Why was this change necessary?
  - The `cypress` interception proxy currently only supports `gzip` and plaintext content encoding. This change adds `Brotli` and `deflate` support
  - Faced issues end to end testing a website with compressed `Brotli` assets
  - Brotli support has been added over the years to web browsers, with 94% of worldwide users using a browser that supports the format
- What is affected by this change?
  - Cypress `proxy` module
- Any implementation details to explain?
  - The change is implemented a similar way to the `gzip` support, but as it now supports multiple content-encodings there's a need to loop through the `content-encoding` header to decompress and compress in the order they were applied instead of explicitly checking for the encoding method
  - There are no tests for `gzip` encoding so I did not add any tests for the new encodings

`Brotli` and `deflate` requests will no longer return no content and instead return desired content.

### PR Tasks

### DIFF
--- a/packages/proxy/lib/http/request-middleware.ts
+++ b/packages/proxy/lib/http/request-middleware.ts
@@ -3,6 +3,7 @@ import { blocked, cors } from '@packages/network'
 import { InterceptRequest } from '@packages/net-stubbing'
 import debugModule from 'debug'
 import type { HttpMiddleware } from './'
+// eslint-disable-next-line no-duplicate-imports
 import { AllowedContentEncodings } from './'
 
 export type RequestMiddleware = HttpMiddleware<{
--- a/packages/proxy/lib/http/request-middleware.ts
+++ b/packages/proxy/lib/http/request-middleware.ts
@@ -3,6 +3,7 @@ import { blocked, cors } from '@packages/network'
 import { InterceptRequest } from '@packages/net-stubbing'
 import debugModule from 'debug'
 import type { HttpMiddleware } from './'
+import { AllowedContentEncodings } from './'
 
 export type RequestMiddleware = HttpMiddleware<{
   outgoingReq: any
@@ -111,15 +112,18 @@ const EndRequestsToBlockedHosts: RequestMiddleware = function () {
 }
 
 const StripUnsupportedAcceptEncoding: RequestMiddleware = function () {
-  // Cypress can only support plaintext or gzip, so make sure we don't request anything else
-  const acceptEncoding = this.req.headers['accept-encoding']
+  // ignore as header is a string
+  // @ts-ignore
+  const acceptEncoding: string = this.req.headers['accept-encoding'] || ''
+  const acceptEncodingArray: string[] = acceptEncoding
+  .split(',')
+  .filter((str) => AllowedContentEncodings.includes(str))
+  .map((str) => str.trim())
 
-  if (acceptEncoding) {
-    if (acceptEncoding.includes('gzip')) {
-      this.req.headers['accept-encoding'] = 'gzip'
-    } else {
-      delete this.req.headers['accept-encoding']
-    }
+  if (acceptEncodingArray.length) {
+    this.req.headers['accept-encoding'] = acceptEncodingArray.join(', ')
+  } else {
+    delete this.req.headers['accept-encoding']
   }
 
   this.next()
--- a/packages/proxy/lib/http/response-middleware.ts
+++ b/packages/proxy/lib/http/response-middleware.ts
@@ -5,6 +5,7 @@ import { cors, concatStream, httpUtils } from '@packages/network'
 import type { CypressIncomingRequest, CypressOutgoingResponse } from '@packages/proxy'
 import debugModule from 'debug'
 import type { HttpMiddleware } from '.'
+// eslint-disable-next-line no-duplicate-imports
 import { AllowedContentEncodings } from '.'
 import iconv from 'iconv-lite'
 import type { IncomingMessage, IncomingHttpHeaders } from 'http'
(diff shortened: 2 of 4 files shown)


jennifer-shehane:
@itsjfx Thanks for the contribution! We'll likely want some more tests around this behavior. We'll have an engineer take a look at the PR and request any changes necessary as soon as we can.

packages/proxy/lib/http/request-middleware.ts
```diff
@@ -111,15 +113,18 @@ const EndRequestsToBlockedHosts: RequestMiddleware = function () {
...
-    } else {
-      delete this.req.headers['accept-encoding']
-    }
+  // ignore as header is a string
+  // @ts-ignore
+  const acceptEncoding: string = this.req.headers['accept-encoding'] || ''
+  const acceptEncodingArray: string[] = acceptEncoding
+  .split(',')
+  .filter((str) => AllowedContentEncodings.includes(str))
+  .map((str) => str.trim())
```
flotwig, line 122: this filtering is not 100% correct, see the MDN doc for accept-encoding: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding

the quality value (`q`) is ignored in this parsing. `q` should probably be preserved, so filtering on `/^(gzip|br|deflate)/.test(str)` or something would be more correct. also, lines 121 and 122 ought to be reversed, right now `gzip ` (note the space) would be filtered out by this logic.

packages/proxy/lib/http/response-middleware.ts
```diff
@@ -5,6 +5,8 @@ import { cors, concatStream, httpUtils } from '@packages/network'
 import type { CypressIncomingRequest, CypressOutgoingResponse } from '@packages/proxy'
 import debugModule from 'debug'
 import type { HttpMiddleware } from '.'
+// eslint-disable-next-line no-duplicate-imports
```
flotwig, line 8: deduplicate import

packages/proxy/lib/http/response-middleware.ts
```diff
@@ -137,18 +135,54 @@ const LogResponse: ResponseMiddleware = function () {
...
-      this.incomingResStream = this.incomingResStream.pipe(gunzip).on('error', this.onError)
+      debug('received encodings: %o', encodings)
+      // first pass, make sure we support all the encodings
+      for (let encoding of encodings) {
+        if (!AllowedContentEncodings.includes(encoding)) {
+          debug('received unsupported content encoding: %s, not uncompressing body', encoding)
+          this.next()
 
-      this.isGunzipped = true
+          return
```
flotwig, line 162: `this.next()` probably isn't right here - middlewares call `this.makeResStreamPlainText()` to guarantee that `this.resStream` is a plaintext stream. also, `this.next()` signals a middleware's completion, so this could end up being called twice.

instead, I think we should throw an error here and kill the request/response. we've already sent a valid `accept-encoding` - if the server is disregarding that and sending back unknown `content-encoding`, then that's an error, right?

packages/proxy/lib/http/response-middleware.ts
```diff
@@ -441,10 +475,34 @@ const MaybeRemoveSecurity: ResponseMiddleware = function () {
...
+const ReCompressBody: ResponseMiddleware = function () {
+  if (this.isUncompressed) {
+    const encodings: string[] = resGetContentEncodings(this.incomingRes)
+
+    debug('encodings to recompress: %o', encodings)
+    for (let encoding of encodings) {
+      let recompressor: zlib.Gzip | zlib.BrotliCompress | undefined
+
+      if (encoding === 'gzip') {
+        debug('recompressing gzip encoding')
```
flotwig, line 487: i know it's not the current pattern in this file, but can you change the debug statements you've added to `this.debug` instead of `debug`? `this.debug` will log some additional info about the current request

flotwig (changes requested):
@itsjfx really nice work here, thank you for opening this PR!

> There are no tests for gzip encoding so I did not add any tests for the new encodings

There are actually tests for the `gzip` and `accept-encoding` functionality, but they exist in the `packages/server` integration tests: https://github.com/cypress-io/cypress/blob/890bde6a053525dfaaac9abcdf041c98753d50c6/packages/server/test/integration/http_requests_spec.js#L1341-L1519

You can run these tests via `yarn test-integration http_requests` inside of `packages/server`.

We definitely want new tests to cover the `br` and `deflate` support, as well as tests that cover multi-layered decompression/recompression as you've implemented it.

You can add these tests to the existing `http_requests_spec`, or add unit/integration tests in `packages/proxy/test`, or a mix of both. 

Let me know if I can be of further assistance, would love to get this merged in.

BlueWinds:
As @flotwig mentioned, this is a pretty nice PR, and a feature well worth having. It's been a couple months though, so closing for inactivity.

Feel free to open a new PR or ping us and we can reopen this one if you do come back and address some of the comments, we'd be happy to take another look, @itsjfx.

PR #21111, opened 2022-04-18
feat: add alpha/beta labels in launchpad

lmiller1990:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1534

### User facing changelog

Add "beta" label for component testing and "alpha" label for Next.js in launchpad.

## Beta label

image

## Alpha label

Note: This screenshot is from a test with mock data, using Create React App as a fixture. It's not really in alpha. You can see in `frameworks.ts`-  [see here](https://github.com/cypress-io/cypress/pull/21111/files#diff-dce5144c067a639808029e38cef9e4e90a64dec50aeec74e6064f3cb31a56981R1870) - which frameworks have the alpha label (just Next.js right now).

image

### PR Tasks


packages/frontend-shared/src/components/Card.vue
```diff
@@ -9,6 +9,12 @@
     data-cy="card"
     @click="emits('click')"
   >
+    <div
+      v-if="title === t('testingType.component.name')"
+      class="ribbon absolute top-0 right-0 text-teal-600"
```
marktnoonan, line 14: Let's put `aria-hidden` on this so that it isn't announced by screen readers at this point in the DOM - it is encountered after the first card and at the start of the second card, but there's nothing to tell the user which it refers to or what is in beta.

Probably, as part the card title we could add something like this at the end:

```html
<span class='visually-hidden'>Beta</span>
```

Which would seem to be the clearest way to reproduce the effect of the ribbon when listening to the page.

packages/launchpad/src/setup/AlphaLabel.vue
```diff
@@ -0,0 +1,11 @@
+<template>
+  <span class="text-warning-600 bg-warning-50 border-warning-300 border-1 border-transparent rounded p-2px ml-4px pr-3px">
```
marktnoonan, line 2: Similarly, here we could have a `.visually-hidden` piece of text that clarifies the work "alpha" refers to Cypress support of the framework, since it is read aloud as something like "nuxt alpha". We could prefix "alpha" with "support is in " or something similar.

marktnoonan (changes requested):
Works well, I have couple small suggestions on using our existing `visually-hidden` class to add context to these labels when announced by screen readers.

packages/graphql/src/schemaTypes/objectTypes/gql-WizardFrontendFramework.ts
```diff
@@ -31,6 +31,10 @@ export const WizardFrontendFramework = objectType({
       description: 'The display name of the framework',
     })
 
+    t.nonNull.boolean('alpha', {
+      description: 'Whether the framework is in alpha or not',
+    })
```
tbiethman, line 36: Just curious if it's worth codifying this as `supportStatus` at this point. Do we anticipate marking future frameworks as "beta" when component testing overall leaves beta?

packages/launchpad/src/setup/TestingTypeCards.cy.tsx
```diff
@@ -82,4 +82,14 @@ describe('TestingTypeCards', () => {
...
   })
+
+  it('renders beta label for component testing', () => {
+    cy.mountFragment(TestingTypeCardsFragmentDoc, {
+      render: (gqlVal) => {
+        return <TestingTypeCards gql={gqlVal} />
+      },
+    })
+
+    cy.get('[data-cy-testingtype="component"]').contains('Beta')
```
tbiethman, line 93: Let's get a Percy snapshot here.

tbiethman (approved):
New styles look great :+1:

packages/graphql/src/schemaTypes/objectTypes/gql-WizardFrontendFramework.ts
lmiller1990: I'm expecting so - Jess has talked about support for Svelte and is giving a presentation on it soon, and Angular is another one we "support" (inside `npm/angular`, and part of our "official" adapters) but certainly isn't production ready.

packages/launchpad/src/setup/TestingTypeCards.cy.tsx
lmiller1990: Will do, and will re-ping for a review in around 14 hours from the time of this post.

packages/frontend-shared/src/components/Card.vue
lmiller1990: Nice idea, will do.

We should do a screen test run through of the entire app.

lmiller1990:
Fixed and added requested a11y from Mark.

packages/graphql/src/schemaTypes/objectTypes/gql-WizardFrontendFramework.ts
lmiller1990: Done.

PR #21117, opened 2022-04-18
chore: [Multi-origin] Correctly find the spec bridge if chromeWebSecurity set to false.

mjhenkes:
- Closes 

### User facing changelog

n/a

### Additional details

If chromeWebSecurity is set to false we can't rely on a security error to identify the correct spec bridge.

This PR adds some check to verify we are on the correct origin policy for the spec bridge.

### PR Tasks

### DIFF
--- a/packages/driver/cypress/integration/e2e/multi-domain/multi_domain_validation_specs.ts
+++ b/packages/driver/cypress/integration/e2e/multi-domain/multi_domain_validation_specs.ts
@@ -198,6 +198,15 @@ describe('cy.origin', () => {
       .should('equal', 'Welcome TJohnson')
     })
 
+    it('finds the correct spec bridge even if a previous spec bridge host is a subset of the current host', () => {
+      // Establish a spec bridge with a 'bar.com' host prior to loading 'foobar.com'
+      cy.origin('http://www.bar.com:3500', () => undefined)
+
+      cy.origin('http://www.app.foobar.com:3500', () => {
+        cy.visit('/fixtures/multi-domain.html')
+      })
+    })
+
     it('uses cy.origin twice', () => {
       cy.visit('/fixtures/auth/index.html')
       cy.get('[data-cy="login-idp"]').click()
--- a/packages/runner/injection/multi-domain.js
+++ b/packages/runner/injection/multi-domain.js
@@ -16,7 +16,7 @@ const findCypress = () => {
     try {
       // If Cypress is defined and we haven't gotten a cross origin error we have found the correct bridge.
       if (frame.Cypress) {
-        // If the ending $ is not included in the template string, it breaks transpilation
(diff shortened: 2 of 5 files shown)


packages/runner/injection/multi-domain.js
```diff
@@ -16,7 +16,18 @@ const findCypress = () => {
     try {
       // If Cypress is defined and we haven't gotten a cross origin error we have found the correct bridge.
       if (frame.Cypress) {
-        return frame.Cypress
+        // If the ending $ is not included in the template string, it breaks transpilation
```
chrisbreiding, line 19:
```suggestion
        // If the ending $ is included in the template string, it breaks transpilation
```

It breaks it if it _is_ included, right? Does it still break if you escape the $ (e.g. `\$`)?

chrisbreiding (approved):

packages/driver/cypress/integration/e2e/multi-domain/multi_domain_validation_specs.ts
```diff
@@ -198,6 +198,15 @@ describe('cy.origin', () => {
       .should('equal', 'Welcome TJohnson')
     })
 
+    it('finds the correct spec bridge even if a previous spec bridge host is a subset of the current host', () => {
```
emilyrohrbough, line 201: Given the PR title, I expect this test or a few tests to explicitly set chromeWebSecurity: false. Given we can't update that mid-test, it seems like a system-test should have been added to validate this??

packages/runner/injection/multi-domain.js
mjhenkes: My wording is confusing 🤪 .

This string breaks:
```
`(^|\\.)${ frame.location.host.replaceAll('.', '\\.') }$`
```

This string works: 
```
`(^|\\.)${ frame.location.host.replaceAll('.', '\\.') }` + '$'
```

packages/driver/cypress/integration/e2e/multi-domain/multi_domain_validation_specs.ts
mjhenkes: The 'positive' case of the code is tested for each origin. We verify the port, protocol, and that the host ends with the host from the spec bridge. You are correct that the 'negative' is not tested. I made the call that testing this case  did not nessitate the cost of adding a system test (System tests seem expensive to me 🤷‍♂️ ). I can add a system test if the group disagrees with that reasoning.

mjhenkes: I've updated an existing system test for chromeWebSecurity to also test this case.

emilyrohrbough (approved):

AtofStryker (commented):
was a little late 😆

PR #21122, opened 2022-04-18
fix: label text should be clickable to toggle snapshot highlight

marktnoonan:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1384

### User facing changelog

Fixes switch label to be clickable, not just the switch itself. Switch is also labelled correctly in the Accessibility Tree after this PR. @lmiller1990 noticed this in a PR review on a related part of the code, but it was merged before this could be updated.

### Additional details

Updated test to use `cy.findByLabelText(highlightsLabel)` instead of `cy.get('[id="toggle-highlights"]')` to catch regressions.

Previously, text was not clickable to active the switch:

https://user-images.githubusercontent.com/8340719/163885723-a556b235-ae1e-4ab6-8548-704c57a198c6.mov

Now, it is:

https://user-images.githubusercontent.com/8340719/163885752-901bc288-f605-4a4c-9aaf-78a1f757d863.mov

### PR Tasks


rachelruderman:
Now that the label is clickable, whaddya think about adding `cursor: pointer` on hover?

marktnoonan:
@rachelruderman good point. On a general form field it seems like too much, but here I can see it making it easier for users to discover how it works and not having to be so precise.

lmiller1990 (approved):

lmiller1990:
Nice, this has been a pain point for me for a while. Thanks for following up on this.

marktnoonan:
@rachelruderman done!

rachelruderman (approved):

PR #21139, opened 2022-04-19
feat: Spec List: Move git status details to a dynamic file icon [CLOUD-576]

MuazOthman:
- Closes [CLOUD-576](https://cypress-io.atlassian.net/browse/CLOUD-576) & [CLOUD-604](https://cypress-io.atlassian.net/browse/CLOUD-604)

### User facing changelog

In the spec list, replaced the column for Git Status with icons to the left of the file name indicating if the file was created/modified/unchanged

Old Screen: 
image

New Screen:

Screen Shot 2022-05-09 at 3 29 03 PM
Screen Shot 2022-05-09 at 3 29 11 PM

### PR Tasks


packages/frontend-shared/src/components/Tooltip.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <Tooltip
+    v-if="!skipTooltip"
     :popper-class="{'no-arrow': hideArrow}"
     :popper-hide-triggers="['hover']"
```
rachelruderman, line 5: Should `popper-hide-triggers` also include focus for keyboard accessibility?

packages/app/cypress/e2e/specs_list_e2e.cy.ts
```diff
@@ -59,8 +61,8 @@ describe('App: Spec List (E2E)', () => {
   })
 
   it('shows a git status for each spec', () => {
-    cy.get('[data-cy="spec-list-file"]').each((row) => {
-      cy.wrap(row).contains('.git-info-row', 'Test Author')
+    cy.get('.git-info-row').each((row) => {
```
rachelruderman, line 64: Would be great to use a data selector instead of a class, but I realize this was already-existing

packages/app/src/specs/SpecListGitInfo.vue
```diff
@@ -1,24 +1,34 @@
...
+      class="h-full grid items-center"
+      :skip-tooltip="!classes.showTooltip"
     >
-      <span
-        v-if="props.gql.statusType !== 'unmodified'"
-        class="rounded-full border-1 min-w-6px min-h-6px max-w-6px max-h-6px"
-        :class="classes?.indicator"
+      <component
+        :is="classes.icon"
+        :class="classes?.iconClasses"
```
rachelruderman, line 12: Is optional chaining necessary? From the definition of `classes` on line 163, it looks like it will always be an object?

rachelruderman (approved):
Just some minor comments, looks good!

packages/frontend-shared/src/components/Tooltip.vue
```diff
@@ -16,9 +18,11 @@ import { Tooltip } from 'floating-vue'
 withDefaults(defineProps<{
   color?: string
   hideArrow?: boolean
+  skipTooltip?: boolean
```
tbiethman, line 21: Nit: thoughts on calling this prop `disabled` to align with the standard semantics?

My initial thought is "Why are we rendering a tooltip if we don't want to show a tooltip?" But I assume it's because the vue templates aren't flexible enough to reuse/dry up complex child markup. We could move to a render function to better enable that, but that doesn't sound ideal either.

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -120,11 +120,10 @@
     "newSpecButton": "New Spec",
     "searchPlaceholder": "Search Specs",
     "componentSpecsHeader": "Component Specs",
-    "e2eSpecsHeader": "E2E Specs",
-    "gitStatusHeader": "Git Status",
+    "e2eSpecsHeader": "E2E specs",
```
tbiethman, line 123: Are we intentionally capitalizing the leading word only? Figma still seems to show the previous capitalization, but I might have missed an update somewhere.

rachelruderman: Yeah, we're meant to be using sentence case although Figma has a mix of sentence and title case. There's a JIRA card for it: https://cypress-io.atlassian.net/browse/UNIFY-1635

tbiethman: Ah, I see. Thanks @rachelruderman!

packages/frontend-shared/src/components/Tooltip.vue
MuazOthman: Name updated

packages/app/src/specs/SpecListGitInfo.vue
```diff
@@ -1,24 +1,35 @@
...
+            {{ props.gql.subject }}
+          </p>
+          <p class="text-xs">
+            {{ gitTooltipSubtext }}
+          </p>
+        </div>
+      </template>
+    </Tooltip>
     <div
       class="overflow-hidden truncate"
```
tbiethman, line 30: Looks like we're missing the old gitText classes that were previously applied to the face-up text: 

```js
'text-gray-600 group-hocus:text-indigo-500'
```

I don't think we want the hocus styles anymore based on the current designs, but text-gray-700 (the value now shown in Figma) should be applied here.

packages/app/src/specs/SpecListGitInfo.vue
```diff
@@ -1,24 +1,35 @@
...
+        <div>
+          <p class="max-w-sm text-sm truncate overflow-hidden">
+            {{ props.gql.subject }}
+          </p>
+          <p class="text-xs">
+            {{ gitTooltipSubtext }}
+          </p>
+        </div>
+      </template>
+    </Tooltip>
```
tbiethman, line 28: I think changing this is out of scope in the near term, but we need to come up with a strategy for making this and future tooltip content available to users that can't or don't want to use mouse input. 

The tooltip target seems to catch focus and open the tooltip when it receives focus, but that conflicts with our desire to update the Specs List to use standard arrow keys for directory navigation: https://cypress-io.atlassian.net/browse/UNIFY-864. Given the scope of the additions being planned, though, we may need to change our desired approach.

packages/app/src/specs/SpecListGitInfo.vue
MuazOthman: Fixed

packages/app/src/specs/SpecListGitInfo.vue
MuazOthman: Interesting. I guess we need to address this separately.

tbiethman (approved):
Looks good! The failing tests look flaky and unrelated.

PR #21143, opened 2022-04-19
fix: stop ResizeObserver error within the SelectorPlayground from throwing error

JessicaSachs:
- Closes [UNIFY-1425](https://cypress-io.atlassian.net/browse/UNIFY-1425)

There was a bug where a `ResizeObserver` error was being thrown when opening the SelectorPlayground. It is no longer thrown.

To fix it, we used `ch` and CSS to calculate width and offset instead of using JavaScript to measure the absolute pixel width.

### Changelog

There is no change to functionality, this bug has never released to the wild.

Shouldn't have changed. Only changed the implementation of the styles to prevent the error

### DIFF
--- a/packages/app/src/runner/selector-playground/SelectorPlayground.vue
+++ b/packages/app/src/runner/selector-playground/SelectorPlayground.vue
@@ -59,6 +59,7 @@
           v-model="selector"
           autocapitalize="none"
           autocorrect="off"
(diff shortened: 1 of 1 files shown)


jennifer-shehane (changes requested):
@JessicaSachs Can you please include a full description of what this PR does outside of the Jira link?

PR #21152, opened 2022-04-20
feat: Auto-merge config when adding new testing type

tgriesser:
- Closes UNIFY-1256

### User facing changelog

When setting up a new testing type, if they already have a project, instead of asking the user to merge the config manually, we can do it automatically in the large majority of cases (basically unless the user is doing something extremely custom, or using non-babel compatible TS syntax). If we encounter an error while attempting to add, we will then show the existing "merge" UI.

### Additional details

- Uses Babel to reliably merge with the config, using ASTs to detect the most common patterns:

```ts
import { defineConfig } from 'cypress' // or const { defineConfig } = require('cypress')

defineConfig({ ...
```

```ts
import { defineConfig as myDefineConfig } from 'cypress' // or const { defineConfig: myDefineConfig } = require('cypress')

myDefineConfig({ ...
```

```ts
import cy from 'cypress' // (or import * as cy from 'cypress')

cy.defineConfig({ ...
```

```ts
export default { ... // (or module.exports = { ...)
```

- If we don't detect one of the common patterns, but see something like:

```ts
module.exports = callSomeFn() // (or export default callSomeFn())
```

We add via a spread operator:

```ts
module.exports = {
  ...callSomeFn(),
  e2e: {}
}
```

- If there is already an object property matching the given testing type, we do not add to the file:

```ts
export default {
  e2e: {}  // we will not scaffold if we are requested to add "e2e" for some reason
}
```

- If the file is empty for whatever reason, we will recover & scaffold it anyway (this is currently an unhandled edge case)
- Uses [babel-plugin-tester](https://github.com/babel-utils/babel-plugin-tester) to unit test the different situations we want to be able to merge, via the `__babel_fixtures__` directories.
- Removes prettier as a dependency, opting to source the prettier if it's defined in the user's project

### PR Tasks


packages/config/src/ast-utils/addToCypressConfig.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    module.exports = defineConfig({
+      
+    })
+  `
+}
+
+function maybeFormatWithPrettier (code: string, filePath: string) {
+  try {
+    const prettierImportPath = require.resolve('prettier', { paths: [path.dirname(filePath)] })
```
lmiller1990, line 161: Curious why we do the require syntax here - we ship our own version of prettier in the bundle from the looks of things, should we just use that? My understanding is we'd like to use their local version, if they've got one, but if they don't, should we fallback to the one we are shipping w/ the binary?

lmiller1990 (commented):
I tried it out and it works as expected. It does mess with the formatting of the config file though - I added CT to `driver`, and it does this:

## Before

```js
import { defineConfig } from 'cypress'

export default defineConfig({
  'projectId': 'ypt4pf',
  'hosts': {
    '*.foobar.com': '127.0.0.1',
  },
  'reporter': 'cypress-multi-reporters',
  'reporterOptions': {
    'configFile': '../../mocha-reporter-config.json',
  },
  'e2e': {
    'setupNodeEvents': require('./cypress/plugins'),
    'baseUrl': 'http://localhost:3500',
  },
})
```

## After (adds carriage returns)

```js
import { defineConfig } from 'cypress'

export default defineConfig({
  'projectId': 'ypt4pf',

  'hosts': {
    '*.foobar.com': '127.0.0.1',
  },

  'reporter': 'cypress-multi-reporters',

  'reporterOptions': {
    'configFile': '../../mocha-reporter-config.json',
  },
  'e2e': {
    'setupNodeEvents': require('./cypress/plugins'),
    'baseUrl': 'http://localhost:3500',
  },
})
```

packages/config/src/ast-utils/addToCypressConfig.ts
tgriesser: > we ship our own version of prettier in the bundle from the looks of things, should we just use that?

We’re trying to avoid that moving forward, it’s like 30mb

lmiller1990:
Only comment I have is around how the spacing/indentation: https://github.com/cypress-io/cypress/pull/21152#pullrequestreview-949347422.

Outside of that, I didn't find any other missing edge cases. I'll leave a :heavy_check_mark:. Is there anything else to do around the white-space or are we just going to leave it as-is for now? We still need to some work if we want to completely remove `prettier` from our deps.

lmiller1990 (approved):
Left a comment around the formatting, but code seems to work as expected - gave it a test run, feels great to not need to modify `cypress.config.js` manually :100:

packages/config/src/ast-utils/addToCypressConfig.ts
```diff
@@ -0,0 +1,171 @@
...
+
+import { addToCypressConfigPlugin } from './addToCypressConfigPlugin'
+import { addComponentDefinition, addE2EDefinition, ASTComponentDefinitionConfig } from './astConfigHelpers'
+
+const debug = debugLib('cypress:config:addToCypressConfig')
+
+/**
+ * Adds to the Cypress config, using the Babel AST utils.
+ *
+ * Injects the at the top of the config definition, based on the common patterns of:
```
ryanthemanuel, line 17:
```suggestion
 * Injects the export at the top of the config definition, based on the common patterns of:
```

packages/config/src/ast-utils/addToCypressConfigPlugin.ts
```diff
@@ -0,0 +1,235 @@
...
+import { print } from 'recast'
+
+const debug = debugLib('cypress:config:addToCypressConfigPlugin')
+
+interface AddToCypressConfigPluginOptions {
+  shouldThrow?: boolean
+}
+
+/**
+ * Standardizes our approach to writing values into the existing
```
ryanthemanuel, line 13: This comment seems incomplete.

tgriesser: It was indeed. Updated

ZachJW34 (commented):
Tested it and it works really well! Can we have an explicit e2e test that covers merging? Or do you think there is sufficient coverage based on the tweaks you made to that test?

ryanthemanuel (approved):
Tested this and it seems solid to me as well. Great unit test coverage as well. Nice work!

PR #21155, opened 2022-04-20
feat: UI updates to scaffolded files

rachelruderman:
- Closes UNIFY-1552

### User facing changelog
- In E2E and component testing configuration, shows the scaffolded files on the launchpad in a custom order
- Displays the scaffolded file names in bold to match Figma
- Updates scaffolded file descriptions to match content in Figma
- Small fixes to punctuation

### Additional details

These changes implement the design set out in Figma, resulting in an improved user experience during configuration.

**Before**
ui-order (2)

**After**
Screen Shot 2022-04-21 at 3 10 10 PM

### PR Tasks


packages/data-context/src/actions/WizardActions.ts
```diff
@@ -207,14 +209,18 @@ export class WizardActions {
     await this.ctx.fs.mkdir(supportDir, { recursive: true })
 
     let fileContent: string | undefined
+    let description: string = ''
 
     if (fileName === 'commands') {
       fileContent = commandsFileBody(language)
+      description = 'A support file that is useful for creating custom Cypress commands and overwriting existing ones.'
```
rachelruderman, line 216: Ideally these strings would be set by the frontend instead of the backend so we can store them in `en-US.json`, but doing so would require some backend changes that would have significantly bloated the scope for this ticket.
Would be great to do at another time

packages/launchpad/src/components/code/FileRow.cy.tsx
```diff
@@ -120,4 +120,8 @@ describe('FileRow', () => {
 
     cy.percySnapshot()
   })
+
+  it('has expected styles', () => {
+    cy.get('h2').should('have.class', 'font-semibold')
+  })
```
rachelruderman, line 126: As a general rule, do we write style tests like this or do we just consider it "done" thanks to the percy snapshot?

packages/data-context/src/actions/WizardActions.ts
```diff
@@ -190,13 +190,15 @@ export class WizardActions {
 
     assert(chosenFramework && chosenLanguage && chosenBundler)
 
-    return await Promise.all([
+    const scaffoldedFiles = await Promise.all([
```
ZachJW34, line 193: Rather than maintain a list of the expected order of these files, can we just order them here and write a comment? We loop through this array to render out the files changed so if it was written like:
```
return await Promise.all([
      this.scaffoldConfig('component'),
      this.scaffoldSupport('component', chosenLanguage),
      this.scaffoldSupport('commands', chosenLanguage),
      this.scaffoldComponentIndexHtml(chosenFramework),
      this.scaffoldFixtures(),
    ])
```

the order would be correct and it would remove the need to sort them.

packages/data-context/src/actions/WizardActions.ts
```diff
@@ -272,10 +278,12 @@ export class WizardActions {
     // only do this if config file doesn't exist
     this.ctx.lifecycleManager.setConfigFilePath(`cypress.config.${this.ctx.coreData.wizard.chosenLanguage}`)
 
+    const adjective = (testingType === 'e2e') ? 'E2E' : 'component testing'
+
     return this.scaffoldFile(
       this.ctx.lifecycleManager.configFilePath,
       configCode,
-      'Created a new config file',
+      `The Cypress config file where the ${adjective} dev server is configured.`,
```
ZachJW34, line 286: Not sure this wording makes sense for e2e. You _can_ configure a server for e2e inside of `setupNodeEvents` but it's not required. It should just say "The Cypress configuration file for {adjective} testing".

packages/launchpad/cypress/fixtures/scaffolded-files.json
```diff
@@ -0,0 +1,127 @@
+{
```
ZachJW34, line 1: We have a test in [project-setup.cy.ts](https://github.com/cypress-io/cypress/blob/691234cfdf780a3d9df3cdd43a8a04b6bf36c24e/packages/launchpad/cypress/e2e/project-setup.cy.ts#L192), we do a `verifyFiles` which we could tweak to verify the ordering as well. We could remove this fixture and have better coverage.

packages/data-context/src/actions/WizardActions.ts
rachelruderman: 💡 _work smarter, not harder_ lolllll this is so much simpler, thank you for pointing that out! Makin' those changes now!!

packages/launchpad/cypress/fixtures/scaffolded-files.json
```diff
@@ -0,0 +1,127 @@
...
+               "id": "1",
+               "relative": "cypress.config.ts",
+               "absolute": "/Users/rachel/Documents/GitHub/cypress/packages/launchpad/cypress.config.ts",
+               "contents": "Neque porro quisquam est qui dolorem ipsum quia dolor sit amet, consectetur, adipisci velit...",
+               "fileExtension": ".ts"
+           }
+       },
+       {
+           "status": "valid",
+           "description": "An example fixture for data imported into your Cypress tests, such as `cy.intercept()`.",
```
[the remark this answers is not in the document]
lmiller1990: > If there's a preference here to do a system test instead, lmk and I'll be happy to update!

Normally we don't add a system test if we can avoid it, since they take so long to run. I'd say E2E tests in cypress/e2e are fine for this kind of fix.

packages/data-context/src/actions/WizardActions.ts
lmiller1990: Not sure we need `async/await` here, either.

```ts
private scaffoldComponent () {
  return Promise.all([ /* ... */ ])
```

lmiller1990 (approved):
Just one minor comment. Looks great.

ZachJW34 (approved):

davidmunechika (approved):

packages/data-context/src/actions/WizardActions.ts
rachelruderman: @lmiller1990 true that, updating!

rachelruderman: Actually I'mma keep it because all things being equal, it was helpful to have an intermediate variable for `scaffoldedFiles` that could be easily console-logged after the `await`

PR #21157, opened 2022-04-21
feat: reflect the GQL baseError state in App, not just launchpad

marktnoonan:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1595 and https://cypress-io.atlassian.net/browse/UNIFY-1440

### User facing changelog

- `baseError` state will be surfaced in the app now, as well as the launchpad
- a loading spinner will appear when the config has been saved and is reloading, or the config has been reloaded manually with the "try again" button from the error state.

### Additional details

This PR moves BaseError.vue into `@packages/frontend-shared` and adds a subscription to this state in the App. It also adds some more calls for `toApp()` around the same places we call `toLaunchpad()` to keep state in sync.

One side not is that without a valid config, we were not initially able to serve the application at all, which broke reloading behavior, so there are some changes to avoid serving a 500 with an error payload in that situation. With the 500 avoided, we still hit some problems loading the app with no config resolved, so I added defaults for `namespace` and `socketIoRoute` so that the front end can loaded/

Previously the BaseError component was only in Launchpad and the browser App would not show these kinds of errors. Now they appear.

https://user-images.githubusercontent.com/8340719/166287918-502b9add-5a8a-4b88-a6d7-7f737986ee0c.mov

The 2nd half of the bug reported in [UNIFY-1440](https://cypress-io.atlassian.net/browse/UNIFY-1440) is also addressed by this PR, config values for project ID are now synced in both config and Dashboard Settings parts of the settings page:

https://user-images.githubusercontent.com/8340719/166306693-627285fa-13e5-4180-ada5-857dd1a8c60b.mov

### How to test manually

With the App and Launchpad open on a project, break the config file by setting an invalid value, or breaking the syntax of the the file. Errors should appear in both places. Refreshing should work in both places. "Try again" should work in both places. After fixing the config file, "try again" button should clear the error from both places.

When the App is open, it looks like config errors are the only thing that trigger a Base Error state, @tgriesser is this correct? I see some other things but they would only happen during migration. I've added tests related to the config state, if can think of other ways to get a baseError, I'll test them as well.

For UNIFY-1440 testing, change the `projectId` in the config whilst on the settings page and see that the new ID gets populated in both the config section and the Dashboard Settings section.

### PR Tasks


packages/data-context/src/DataContext.ts
```diff
@@ -338,6 +338,7 @@ export class DataContext {
     } else {
       this.update((coreData) => {
         coreData.baseError = { title, cypressError }
+        this.emitter.baseErrorChange()
       })
```
tgriesser, line 343:
```suggestion
      })
      this.emitter.baseErrorChange()

```

tgriesser (changes requested):
One minor change I noticed, we want to trigger the mutation outside the `update` block. Otherwise I think this is looking good

marktnoonan:
@tgriesser nice - the only things I want to do before getting it out of draft are

- add tests
- make sure clicking "try again" in launchpad clears the app as well, when I tested it didn't seem to

jennifer-shehane:
What happens if you refresh the browser when the error is shown?

tgriesser:
@jennifer-shehane you’ll see the error still

packages/data-context/src/sources/HtmlDataSource.ts
```diff
@@ -6,6 +6,7 @@
 import type { DataContext } from '../DataContext'
 import { getPathToDist, resolveFromPackages } from '@packages/resolve-dist'
 import _ from 'lodash'
+import configDefaults from '../../../app/config-defaults'
```
tgriesser, line 9: It seems like these are pulling the `config-defaults` we use for testing cypress-in-cypress, which is different than the normal config defaults.

Should we just use the cached `namespace` & `socketIoRoute` from the last known "good" config, since we presumably need to have that for the browser to have been launched in the first place? I think `fieldsFromLegacyCfg` should have that?

```
cfg.namespace = fieldsFromLegacyCfg.namespace
cfg.socketIoRoute = fieldsFromLegacyCfg.socketIoRoute
```

marktnoonan: @tgriesser I've made the changes we discussed and one or two small tweaks.

packages/app/src/layouts/default.vue
```diff
@@ -36,11 +48,64 @@
...
+
+gql`
+subscription CheckBaseError {
+  baseErrorChange {
+    ...MainAppQueryData
+  }
+}
+`
+
+useSubscription({ query: CheckBaseErrorDocument })
```
ZachJW34, line 98: Since this page will load for all routes, I think we should follow the same pattern we have for [Runner.vue](https://github.com/cypress-io/cypress/blob/b65371a6a12aa18d9cd187b5bbf9b006a49cabc7/packages/app/src/pages/Specs/Runner.vue#L69) and pause all queries/subscriptions when in run mode.

ZachJW34 (commented):
One small comment, will approve once addressed!

tgriesser (changes requested):
Yeah, let's add the `skip` conditional @ZachJW34 mentioned on the subscription and then I think it's good to merge!

packages/app/src/layouts/default.vue
marktnoonan: Yeah this is a good catch.

marktnoonan:
@tgriesser @ZachJW34 updated. I brought the whole comment about why it's pausing as well, it seems equally useful both places.

tgriesser (approved):

marktnoonan:
Will merge this after CI completes, had to update again to get a base build in Percy that wasn't cancelled.

PR #21160, opened 2022-04-21
fix: subscribe to changes in cypress.config.js in app

lmiller1990:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1163

### User facing changelog

Use subscriptions to update/respond correctly in app when `cypress.config.js` is updated. Now, if you update `cypress.config.js` on the runner page (eg change `viewportHeight`) it will re-execute the spec with the new config.

It can feel a bit unresponsive, since we re-run the child process to get the latest config. We will add an overlay with "Loading..." to give the user immediate feedback that we are reloading, that'll be part of another PR, see [this comment](https://github.com/cypress-io/cypress/pull/21160#issuecomment-1105249176).

### Additional details

I also found a bug where CT fails to reconnect to the dev server after you change `cypress.config.js`. https://cypress-io.atlassian.net/browse/UNIFY-1675. It's outside the scope of this ticket, though, but we should look into this as a fast follow up.

### PR Tasks


packages/app/src/pages/Specs/Runner.vue
```diff
@@ -48,18 +48,53 @@ subscription SpecPageContainer_specsChange {
...
+
+let initialLoad = true
+
+useSubscription({ query: Runner_ConfigChangeDocument }, () => {
+  // if the `config` changed, we want to reload the entire
+  // page and re-execute the current test with the latest config
+  // values
+  // subscriptions trigger on the initial page load,
+  // so we do not want to trigger `window.location.reload` on the
+  // first load, or we get stuck in an infinite loop.
```
tgriesser, line 92: What if we send the config values in the subscription as a new field with a `JSON` payload, and then set them on `window.__CYPRESS_CONFIG__`?

tgriesser (commented):
I think we are actually going to also want to show the loading spinner when the background process is refreshing, @marktnoonan was looking into that as a separate ticket

marktnoonan:
Yeah I was going to either write that up as a ticket or add it as a part of https://github.com/cypress-io/cypress/pull/21157

packages/app/src/pages/Specs/Runner.vue
lmiller1990: We could, but those are base64 encoded right now, we'd need to re-encode it as base64, or receive it as base64.

I'm not super happy we are sticking the config on `window`, but I guess it's probably the most reliable way to do this right now.

I'll give this a try.

lmiller1990:
@tgriesser @marktnoonan I agree we should show a spinner here too, is there a mock for this? Should it be included in this ticket? It could be implemented separately.

packages/app/src/pages/Specs/Runner.vue
lmiller1990: Your idea is quite a bit better in terms of reliability. I found something troubling; the config we source in `data-context` was missing some dynamic values like the `proxyUrl`. For now I patched it in, but we need to continue to aggressively triage the config problem we've got and find a single way to manage and source it. 

I'd say the next step would be a solid refactor of `project-base`, `open-project` and `server-{base,e2e,ct}` and the surrounding area, but this is quite high risk, so not sure we should do that right now. I might start a tech brief where we can gather our thoughts about this.

packages/app/src/runner/index.ts
```diff
@@ -323,21 +328,17 @@ async function initialize () {
...
-  // just stick config on window until we figure out how we are
-  // going to manage it
-  window.UnifiedRunner.config = config
-
   // window.UnifiedRunner exists now, since the Webpack bundle with
   // the UnifiedRunner namespace was injected by `injectBundle`.
   initializeEventManager(window.UnifiedRunner)
 
   window.UnifiedRunner.MobX.runInAction(() => {
-    const store = initializeMobxStore(window.UnifiedRunner.config.testingType)
```
lmiller1990, line 335: We can kill off most of `window.UnifiedRunner` soon

packages/app/src/pages/Specs/Runner.vue
```diff
@@ -48,20 +50,76 @@ subscription SpecPageContainer_specsChange {
...
+const configChangeHandler: SubscriptionHandlerArg<any, any> = (
+  _prev: Runner_ConfigChangeSubscription | undefined,
+  next: Runner_ConfigChangeSubscription,
+) => {
+  if (!next.configChange?.serveConfig) {
+    throw Error(`Did not get expected serveConfig from subscription`)
+  }
+
+  if (!initialLoad && specStore.activeSpec) {
+    try {
```
lmiller1990, line 102: Not super ideal, since we don't know that eventManager will absolute exist here (potential race condition, you changed `cypress.config.js` while the spec is still "Spec is loading...", so I just did a try/catch.

packages/data-context/src/sources/HtmlDataSource.ts
```diff
@@ -41,8 +42,32 @@ export class HtmlDataSource {
     throw err
   }
 
+  getUrlsFromLegacyProjectBase (cfg: any) {
```
lmiller1990, line 45: The config we get in the data-context doesn't have some of these important keys. They are dynamically added later, for example the `port` will only be added once the CT dev server has started.

We'll need to find a way to better manage the config in general, but this fixes the problem for now and links the "legacy" part of config in `project-base` and `server-{ct,e2e}` with the newer data-context stuff.

lmiller1990:
Thanks for the update @marktnoonan, looks like so far it's okay except for that weird bug you found in loom - thanks for capturing that. I'm not sure if I can reproduce, but I'll give it a try.

If we can reproduce, it could be a separate PR? I can see this blowing out if we try to fix any and all config errors here - my main goal was to hook up subscriptions.

packages/app/src/runner/index.ts
```diff
@@ -323,21 +328,17 @@ async function initialize () {
...
+    const store = initializeMobxStore(window.__CYPRESS_TESTING_TYPE__)
 
     store.updateDimensions(config.viewportWidth, config.viewportHeight)
   })
 
-  window.UnifiedRunner.MobX.runInAction(() => setupRunner(config.namespace))
```
marktnoonan, line 340: It looks like this function wasn't using `config.namespace` for anything, this is just cleanup?

marktnoonan (approved):
All looks good and works well in e2e and on settings page, happy to create a follow up issue for my last comment, though since it is testing-type related, I will hold off until we have the config reloading _not_ breaking CT, since that's a possible confounder.

packages/app/src/runner/index.ts
lmiller1990: Yep, unused.

packages/data-context/src/actions/DataEmitterActions.ts
```diff
@@ -35,6 +35,14 @@ abstract class DataEmitterEvents {
     this._emit('devChange')
   }
 
+  /**
+   * Emitted when cypress.config is re-executed and we'd like to
+   * either re-run a spec or update something in the App UI.
+   */
+  configChange () {
+    this._emit('configChange')
```
ryanthemanuel, line 43: Do we envision that this can eventually negate the need for some of the `toApp` and `toLaunchpad` calls elsewhere?

lmiller1990: Good question - for now I'm guessing we want to respond to any config changes in both. I can't think of any reason to selectively only update only `app` or only `launchpad` at this point.

tgriesser (approved):

ryanthemanuel (approved):
Looks good and works well. Nice work!

PR #21169, opened 2022-04-21
fix: update not supported plugins errors

estrada9166:
- Closes  [UNIFY-1524](https://cypress-io.atlassian.net/jira/software/projects/UNIFY/boards/20/backlog?assignee=5ea883deddd45f0b94f5d98d&selectedIssue=UNIFY-1524)

### User facing changelog

Some plugins are not supported in v-10, is for this reason that the error that is being thrown has been updated depending if we are able to identify that the reason is the invalid plugin

### Before
image

### After
image

### PR Tasks


lmiller1990 (approved):
Definitely a big improvement. We still show the original error "glob pattern not found" but I'm guessing this is somewhere deep in driver and not practical to change.

Did we see if any other (official) plugins are broken by 10.x?

packages/driver/src/cypress/runner.ts
```diff
@@ -987,6 +987,8 @@ const _runnerListeners = (_runner, Cypress, _emissions, getTestById, getTest, se
       hookName = getHookName(runnable)
       const test = getTest() || getTestFromHookOrFindTest(runnable)
 
+      const unsupportedPlugin = runnable.invocationDetails && runnable.invocationDetails.originalFile && runnable.err && runnable.err.message ? $errUtils.getUnsupportedPlugin(runnable.invocationDetails.originalFile, runnable.err.message) : null
```
chrisbreiding, line 990: Maybe pass the `runnable` to `$errUtils.getUnsupportedPlugin` and let it take care of this logic and return null if these properties aren't available?

chrisbreiding (changes requested):
I think it would be better if the error message were formatted like:

```
Cypress detected that the current version of @cypress/code-coverage is not supported. Update it to the latest version.

The following error was caught:

> glob pattern string required

Because this error occurred during a after all hook we are skipping all of the remaining tests.
```

lmiller1990:
Nice @estrada9166, for completion sake (and it it's not too difficult) can you add a screenshot of how the error looks? I think it'll be useful for historical context, if we ever look back, as well as for the final review and :heavy_check_mark:  :pray:

chrisbreiding (approved):

lmiller1990:
Alright got the double ✅ , let's ship it.

PR #21181, opened 2022-04-25
fix: update dev server error message

lmiller1990:
- Closes  https://cypress-io.atlassian.net/browse/UNIFY-1424
### User facing changelog

Improve error message for invalid dev-server config.

## Before

image

## After

image

### PR Tasks


packages/errors/src/errors.ts
```diff
@@ -1342,19 +1342,23 @@ export const AllCypressErrors = {
...
       }`
 
     return errTemplate`\
-      Your ${fmt.highlightSecondary(`configFile`)} is invalid: ${fmt.path(configFilePath)}
+      Your ${fmt.highlightSecondary(configFile)} is invalid: ${fmt.path(configFilePath)}
 
-      The ${fmt.highlight(`component.devServer()`)} must be a function with the following signature:
+      The ${fmt.highlight(`component.devServer`)} must be an object with a supported ${fmt.highlight('framework')} and ${fmt.highlight('bundler')}.
 
       ${fmt.code(code)}
```
lmiller1990, line 1364: Below 3 files were not used so I deleted them.

packages/errors/src/errors.ts
```diff
@@ -1342,19 +1342,23 @@ export const AllCypressErrors = {
   },
 
   CONFIG_FILE_DEV_SERVER_IS_NOT_A_FUNCTION: (configFilePath: string, setupNodeEvents: any) => {
```
ZachJW34, line 1344: We should change the name of this error. We throw it based on if the `devServer` is not an object _or_ a function.

ZachJW34 (changes requested):

packages/errors/src/errors.ts
lmiller1990: Good catch... I made this more generic - `CONFIG_FILE_DEV_SERVER_IS_NOT_VALID`.

https://github.com/cypress-io/cypress/pull/21181/commits/b8f1f353b35ec6c2123917177ef69d5ade6ad4da

PR #21187, opened 2022-04-25
chore: Updating system test node modules cache logic to support binary tests

### LINKED ISSUE #21123: node_module caching for binary system tests is flakey
### Current behavior

We generate node_module caches in CircleCI for our binary system tests and share them across multiple jobs. To do this, we hash the contents of all package.json/yarn.lock files within the `system-tests/projects` directory and use that as part of our cache key. However, depending on the executor architecture, the generated hash can differ for the same content, causing a mismatch.

We attempt to correct the mismatch by falling back to a recent cache that matches a subset of our key; however, we do not perform a subsequent install to ensure any missing dependencies are added. This has the potential to create volatility and flake in our build system.

### Desired behavior

We need to make sure that:
* Our cache keys are generated in a deterministic way across architectures
* If we don't have a valid cache, ensure that the node modules are properly installed for a project
* We don't fallback to state keys when we are actually doing partial cache fallbacks

### Test code to reproduce

Here's a recent example showing the mismatch: https://app.circleci.com/pipelines/github/cypress-io/cypress/36440/workflows/b32ce66c-e33a-478b-9888-e0046b815fa7/jobs/1459760?invite=true#step-112-1

But this occurs for any build at the moment. Whether or not it's a problem will depend on what else has built recently and how recently the package/lock contents have changed.

### Cypress Version

N/A

### Other

_No response_

tbiethman:
- Closes #21123 

During an investigation into some test flake, I noticed that the binary system tests were not consuming the generated node module cache for the system test projects. While this did not ending up being the root cause of the flake, I figured we'd get it fixed up to limit future confusion.

Changes made here include:

* Updated the cache keys for the state flags to not not match against our partial cache key fallbacks. We never want to fallback to the state cache when we are looking up the actual cache, and this update prevents it entirely.

* Removed the partial cache fallback for the `restore_cached_system_tests_deps` step. We should always have a cache, and if we don't, we'll just do a full install like normal to limit potential issues from the partial match deps.

* Moved saved cache location of system test node modules from `~/.cache/cy-system-tests-node-modules` to `/tmp/cy-system-tests-node-modules`. CircleCI had permissions issues applying the cache to this location on the plain Linux machine, due to the cache being generated on a Docker image.

* The `cache_key.sh` script was updated to generate a deterministic result when run on different build machines. Previously it was always mismatching on the Linux machine due to the expanded glob results being in a different order. This only appeared recently due to a number of similarly named projects being added to the repo.

### User facing changelog

N/A

### PR Tasks


emilyrohrbough (approved):

tbiethman:
I'm going to need to update the circle cache key env value before the `binary-system-tests` job will pass. I'll wait to do so until after working hours tonight to limit the thrash to builds today.

mjhenkes (approved):

circle.yml
```diff
@@ -201,7 +201,6 @@ commands:
           name: Restore system tests node_modules cache
           keys:
             - v{{ .Environment.CACHE_VERSION }}-{{ checksum "platform_key" }}-system-tests-projects-node-modules-cache-{{ checksum "system_tests_cache_key" }}
-            - v{{ .Environment.CACHE_VERSION }}-{{ checksum "platform_key" }}-system-tests-projects-node-modules-cache-
```
flotwig, line 204: this will also remove the cache fallback from the `install` step... but i guess that's the intent, i can see how yarn install from existing node_modules would differ from a fresh install (and be less deterministic)

flotwig (approved):
thank you for tracking this down :pray:

tbiethman:
> I'm going to need to update the circle cache key env value before the binary-system-tests job will pass. I'll wait to do so until after working hours tonight to limit the thrash to builds today.

Planning on doing this around 10PM CT.

PR #21193, opened 2022-04-26
feat: Update migration for projects with projectId

estrada9166:
- Closes  [UNIFY-1684](https://cypress-io.atlassian.net/browse/UNIFY-1684)

Do not migrate the user pre-extension if the project has projected on the config, with this change we are not going to display anything related to the pre-extension migration, and also, we are gonna add a custom specPattern in the config file.

The reason behind this change is to keep the spec history on the cloud, simple as possible.

### PR Tasks


packages/data-context/src/sources/MigrationDataSource.ts
```diff
@@ -62,6 +62,18 @@ export class MigrationDataSource {
     return this.ctx.coreData.migration.legacyConfigForMigration
   }
 
+  get shouldMigratePreExtension () {
+    return !this.legacyConfig.projectId
```
lmiller1990, line 66: I wonder if anyone does nested projectId, like

```json
{
  "e2e": {
    "projectId": "ABC"
  },
  "component": {
    "projectId": "DEF"
  }
}
```
Not sure if this is even valid, but if it is, we might need to consider this scenario, too.

packages/data-context/src/sources/MigrationDataSource.ts
```diff
@@ -62,6 +62,18 @@ export class MigrationDataSource {
...
+  }
+
+  get migrationOptions () {
```
lmiller1990, line 69: Seems like we have this twice, here and also here? https://github.com/cypress-io/cypress/pull/21193/files#diff-c2c1f114b54072e7f036c272e41454586ddba1378ada7bf61d38049e05e3d8f5R68

lmiller1990: Also what is the point of this getter? Why not just access `shouldMigratePreExtension` directly?

packages/data-context/test/unit/sources/migration/format.spec.ts
```diff
@@ -19,13 +19,27 @@ describe('formatMigrationFile', () => {
         { text: 'js', highlight: false },
       ])
     })
+
+    it('do not highlight the preExtension', () => {
```
lmiller1990, line 23:
```suggestion
    it('do not highlight the preExtension when migratePreExtension is false', () => {
```

system-tests/projects/migration-e2e-custom-integration-with-projectId/README.md
```diff
@@ -0,0 +1,35 @@
+## Migration E2E Custom Integration with projectId
+
+An e2e project with a custom `integrationFolder` named `src` and projectId. It uses the default `testFiles`. We will keep the custom `intergrationFolder`, but it'll be part of `e2e.specPattern`. We will rename their specs to use the `.cy.js` extension.
```
lmiller1990, line 3: Don't forget to update this to explain this fixture - it  says "We will rename their specs to use the `.cy.js` extension", which isn't right.

In this case, "this project has as `projectId` which means we do not migrate to the new `.cy.js` extension, because this will interfere with the historical data on Cypress Cloud" or something like that.

lmiller1990 (commented):
Some questions, code looks pretty thorough.

packages/data-context/src/sources/MigrationDataSource.ts
tgriesser: According to Brian, this should not be permitted, it needs to be a root level property only

lmiller1990: We might need to verify if this actually works in 9.x and see if anyone was using it like that, seems unlikely but you never know.

Eg: I'm saying if it *was* previously permitted, it's possible someone thought to do this.

estrada9166: I tried the projectId inside the e2e and it works fine in 9.x

The support for this should be something like ` return !this.legacyConfig.projectId && !this.legacyConfig.e2e?.projectId` - how common may this be? I guess it would be safer to support it

lmiller1990: That means `component.projectId` probably works, too. If someone either `e2e.projectId` OR `component.projectId`, I think we should not rename, just to be safe.

I have no idea how common this is, but since this is how we make $$$ I don't think we can just assume it's unlikely to happen - we need to get this 100% perfect. I think we should do as you suggested, and **also** add a check for `component.projectId` to be 100% safe. Thoughts?

estrada9166: For the moment we are not gonna take into account the CT for the history on the cloud - just the E2E tests; that's why I only update the `e2e.projectId`

packages/data-context/src/sources/migration/shouldShowSteps.ts
```diff
@@ -65,8 +65,14 @@ async function hasSpecFiles (projectRoot: string, dir: string, testFilesGlob: st
 export async function shouldShowAutoRenameStep (projectRoot: string, config: LegacyCypressConfigJson) {
   const specsToAutoMigrate = await getSpecs(projectRoot, config)
 
+  const e2eMigrationOptions = {
+    // If the configFile has projectId, we do not want to change the preExtension
+    // so, we can keep the cloud history
+    shouldMigratePreExtension: !config.projectId && !config.e2e?.projectId,
+  }
```
lmiller1990, line 72: Should this call this getter? https://github.com/cypress-io/cypress/pull/21193/files#diff-6dae8a6d4d0aac4cb2685a16ea20ceeb091f6113e7e882be5c14f89f579b7e52R69

lmiller1990 (approved):
One more comment, but not a blocker.

packages/data-context/src/sources/migration/shouldShowSteps.ts
estrada9166: This file lives on the `data-context` folder but is not using the` dataContext` itself - I'll call this a util/helper file

flotwig (approved):

Issue #20556, opened 2022-03-10
Why can I use cy.waitFor()?


Serhansolo:
### Current behavior

I noticed that I can use `cy.waitFor(string)` to wait for a cypress alias.  I can't find any documentation on the matter. While there is a good page for `cy.wait(string)` (https://docs.cypress.io/api/commands/wait). While this oddly seems to work whilst waiting for an API stub to resolve. I'm not sure if it is the expected behavior for `wait()`. 

My IDE tells me that `cy.waitFor()` is a valid function to use. I just don't think it is. It looks like it's wrongly exposed in the typing files.

![image](https://user-images.githubusercontent.com/213463/157644274-ea1b084d-e041-4e7e-ac76-71828c13d0ed.png)

### Desired behavior

I would not expect the waitFor() function to show in my IDE as a possible method. Since it is not described in the documentation. OR it should be described in the documentation with the intended behavior.

### Test code to reproduce

Just write `cy.wiat` in any IDE enabled with code hinting and typescript for cypress.

### Cypress Version

8.1.0.

### Other

Or this all has to do with my IDE setup, although I can't imagine it is since it happens with multiple (if not all) team members...

mike-mccormick:
@emilyrohrbough I can maybe help clarify this one; I'm seeing too but following the definition shows it's coming from `eventemitter2`:


![image](https://user-images.githubusercontent.com/2361501/161610579-a3f6dd0e-1b9d-4c5f-a683-35ae5e94085e.png)

other methods in it's d.ts also show with intellesense: 

![image](https://user-images.githubusercontent.com/2361501/161610856-5f19921c-16f5-4143-b623-cc550d45e58f.png)

Spitballing, but any chance it's coming through `EventEmitter`?

![image](https://user-images.githubusercontent.com/2361501/161612019-59529dcd-5606-4abf-ae1a-7bd3a61f7d21.png)

Issue #6683, opened 2020-03-09
Export TypeScript types from `driver` for NPM package


flotwig:
The NPM module's `types` are manually defined and kept up-to-date in `cli/types/index.d.ts`, but the code for these types is largely defined by the shapes of objects in `packages/driver`

Historically, this was done because the `driver` was all coffeescript, so types had to be manually defined. Now that it's starting to get converted to Typescript, we can potentially generate the type declaration file from the Typescript files.

This has some benefits:

- not having to manually update `index.d.ts` all the time
- better guarantees that what's being exported from the declaration file matches the implementation
- we can take advantage of better types internally

Challenges:

- more dtslint tests will be necessary, especially to ensure that there is not a backwards-incompatible change in the declarations when this change is made
- many files in the driver are still coffeescript and do not export types - may need to add external module declaration shims for now since some coffee files cannot be converted since there are open PRs against them

Related issues:

- https://github.com/cypress-io/cypress/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22type%3A+typings%22
- https://github.com/cypress-io/cypress/issues/4218

CypressJoseph:
This is definitely a good idea! Maybe we could revisit when the coffee-to-TS conversion is more fullsome?

Note that `driver/src` is still about 2/3 coffee: ```find packages/driver/src -type f | sed 's/.*\.//' | sort | uniq -c | sort
   1 json
   9 ts
  29 js
  60 coffee```

PR #21197, opened 2022-04-26
chore: improve pkg/driver types

sainthkh:
- 

### User facing changelog

N/A. It's adding types for internal use.

### Additional details
- Why was this change necessary? => Remove `any`s and make code easier to navigate.

N/A

### PR Tasks


packages/driver/src/cypress/command.ts
```diff
@@ -145,13 +149,4 @@ export class $Command {
...
 
-// mixin lodash methods
-_.each(['pick'], (method) => {
-  return $Command.prototype[method] = function (...args) {
-    args.unshift(this.attributes)
-
-    return _[method].apply(_, args)
-  }
-})
-
```
sainthkh, line 156: This code is added 5 years ago and no other lodash function is added to `$Command`. I guess it wouldn't later. So, I moved it inside the class for type.

packages/driver/src/cypress/state.ts
```diff
@@ -0,0 +1,35 @@
+/// <reference path="../../types/cy/commands/session.d.ts" />
```
sainthkh, line 1: New file is created because types defined in `.ts` cannot be imported to `d.ts` files without breaking `d.ts`'s auto import feature.

jennifer-shehane:
@sainthkh Thanks for the updates!

packages/driver/src/cy/assertions.ts
```diff
@@ -347,6 +349,8 @@ export const create = (Cypress, cy) => {
         if (_.isFunction(onRetry)) {
           return cy.retry(onRetry, options)
         }
+
+        return
```
BlueWinds, line 353: Do we need this trailing empty return? It looks very unusual to my eyes.

BlueWinds (approved):

packages/driver/src/cy/assertions.ts
sainthkh: It's because TypeScript wants all return paths. 

As the type of `cy.retry` became specific, TypeScript learned that this function returns a `Promise` above but nothing at the end.

sainthkh: Without this it's an error.

packages/driver/src/cypress/command.ts
```diff
@@ -136,6 +134,12 @@ export class $Command {
     return this
   }
 
+  pick (...args) {
+    args.unshift(this.attributes)
+
+    return _.pick.apply(_, args as [Record<string, any>, any[]])
+  }
```
emilyrohrbough, line 141: I don't necessarily think we need to be added this to every command instance. Why can't we use a static method?

emilyrohrbough (commented):
@sainthkh Can you share the long-term intended type changes? I'm having a hard time understanding the end-goal for the driver package.

It seems like our types patterns are getting inconsistent where some components are exporting their types from their file where other types are being managed in a standalone file, while other types are overriding the CLI types within `./driver/types`

sainthkh:
AFAIK, there is no rule or documentation about where and how we should define our types. But as a person who wrote a lot of type definitions for Cypress, here are my rules and goals:

* Use `cli/types` when the type should be provided to the users.
* Use `driver/types` when we're extending the provided types (e.g. `Window`) or it's not defined in our code. 
* Use each `ts` file and `import`/`import type` syntax when it's defined in our code. 

The reasons are below. It's a bit long because it's related with the history of Cypress code base. 


### Why are we using TypeScript?

Maybe, we should start with the benefit of TypeScript over plain JavaScript. I think there are 2 benefits:

* Static types make you think about the correct input and output of everything. This can help reduce bugs and write easier-to-read code. 
* Thanks to more info, tools can help us navigate and write code faster with intellisense. 

The goal of these type PRs is related with the second reason, **making Cypress driver code easier to navigate**. 

Without types, when a developer wants to know where the definition of some functions are, they have to search the entire project and read all the result. With types, they can save much time by just clicking the name. 


### `cli/types` - public types for users.

`driver` imports types from 3 sources:

* `cli/types` - `d.ts`
* `driver/types` - `d.ts`
* `each file`

`cli/types` were first created because **some users wanted TypeScript support of Cypress while the code base was mostly written in CoffeeScript**. It was inevitable because it's the only way to use TypeScript and CoffeeScript together. 

There was an issue like #6683 that the types should be generated from the `driver` directly. But it's not possible in the near future because `cy` commands are added dynamically to the class `$Cy` like below:

https://github.com/cypress-io/cypress/blob/331e5412f566490e75d45d52d8111836745c336c/packages/driver/src/cypress/commands.ts#L188-L193

When removing type todos in #20601, I thought about moving command options to `driver` and adding them to the `cli/type` on the build. But it would make the problem more complicated because public and internal interfaces are mixed in the project. So, I decided not to do that and imported types from `cli/types` when it's public. 

And removing it is a bit hard, because there are some APIs we need to hide from our definition. For example of #20556, `$Cy` uses `eventemitter2` internally, but it should not expose `waitFor` because it can confuse users. 


### `driver/types`

The next type file created is `driver/types`. It is created mainly to **avoid type problems** in `driver` when we're moving JavaScript to TypeScript. 

But there are some limitation of this file because:

* type check isn't stronger than plain `.ts` files. For example, there is currently `ActiveSessionsSessionData` in `internal-types.d.ts`. But this type doesn't exist. It should be `Cypress.Commands.Session.ActiveSessions`. But TypeScript doesn't alarm because non-existent types are `any` in `d.ts` files. 
* It doesn't help navigating files. When you want to learn more about something, it doesn't lead to the real definition, but the `d.ts`. It's not what we wanted. 

So, whenever possible, I try to define types inside the `.ts` files, and use `import` or `import type` directly from the defined `.ts` file to help navigate. 

But removing these `d.ts` files isn't the goal, because they are sometimes necessary when we're extending the given types like `Window` in `types/window.d.ts` in this PR.

packages/driver/src/cypress/command.ts
sainthkh: It's because of the line below:

https://github.com/cypress-io/cypress/blob/331e5412f566490e75d45d52d8111836745c336c/packages/driver/src/cypress/log.ts#L125

rachelruderman (approved):
LGTM, thank you for your contribution and the in-depth explanation in the comments 🙂

PR #21205, opened 2022-04-26
feat: implement the new design updates for the error pages

rachelruderman:
- Closes UNIFY-1429

### User facing changelog
1. Moves retry button from footer to below header
1. Adds docs button with varying text/link based on the error
1. Updates header text
1. Removes unused footer slot and unused `onReadDocs` prop

### Additional details
[Figma](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/Cypress-App?node-id=6%3A348) - this change improves the user experience by making the retry CTA more prominently featured. Previously, users may have not realized there was a retry button because they would need to scroll to see it.

**Before:**
Screen Shot 2022-04-27 at 6 10 56 PM

**After:**
Screen Shot 2022-04-28 at 11 44 09 AM

### PR Tasks


packages/launchpad/src/error/BaseError.vue
```diff
@@ -127,10 +141,29 @@ const { t } = useI18n()
 const props = defineProps<{
   gql: BaseErrorFragment
   retry?: () => void
```
rachelruderman, line 143: @marktnoonan I'm itching to make this required, let's discuss

packages/launchpad/src/error/BaseError.vue
```diff
@@ -4,15 +4,43 @@
...
+          </Button>
+
+          <Button
+            variant="outline"
+            data-testid="error-docs-button"
+            :prefix-icon="BookIcon"
+            prefix-icon-class="icon-dark-indigo-500 group-hocus:icon-dark-indigo-500 group-hocus:icon-light-indigo-50"
+            :href="t('launchpadErrors.generic.docsButton.DOCS_TYPE.link'.replace('DOCS_TYPE', docsType))"
+          >
+            {{ t('launchpadErrors.generic.docsButton.DOCS_TYPE.text'.replace('DOCS_TYPE', docsType)) }}
```
rachelruderman, line 39: note: better way of doing this

packages/launchpad/src/error/BaseError.vue
rachelruderman: Actually I'll leave as is for now so to not bloat the PR

packages/launchpad/src/error/BaseError.cy.tsx
```diff
@@ -1,20 +1,17 @@
...
-import Button from '@cy/components/Button.vue'
 import { BaseErrorFragmentDoc } from '../generated/graphql-test'
 import dedent from 'dedent'
 
 // Selectors
 const headerSelector = '[data-testid=error-header]'
 const messageSelector = '[data-testid=error-message]'
 const retryButtonSelector = '[data-testid=error-retry-button]'
-const customFooterSelector = '[data-testid=custom-error-footer]'
+const docsButtonSelector = '[data-testid=error-docs-button]'
```
marktnoonan, line 10: Not really related to your changes, but we could add some semantics to these selectors (I skipped messageSelector since it would be `[data-testid=error-message] p` which seems weirdly specific to me. But something like this would be good:

```js
const headerSelector = 'h1[data-testid=error-header]'
const messageSelector = '[data-testid=error-message]'
const retryButtonSelector = 'button[data-testid=error-retry-button]'
const docsButtonSelector = 'a[data-testid=error-docs-button]'
```

marktnoonan (approved):
Looks good! One non-blocking comment, feel free to implement or not.

https://github.com/cypress-io/cypress/pull/21157 will move this component into the frontend-shared package but doesn't change the structure, there might be a little work to reconcile these sets of changes, but probably not much.

packages/launchpad/src/error/BaseError.vue
```diff
@@ -127,10 +141,29 @@ const { t } = useI18n()
...
 }>()
 
 const markdownTarget = ref()
 const baseError = computed(() => props.gql)
 const { markdown } = useMarkdown(markdownTarget, computed(() => props.gql.errorMessage), { classes: { code: ['bg-error-200'] } })
+
+const getDocsType = (): string => {
+  const { errorType, errorStack } = baseError.value
+
+  // Full list of errors lives in "packages/errors/src/errors.ts", but cannot be imported to Vue
```
marktnoonan, line 153: Yeah if want this list we would have to move it into a separate file and import it on its own, there's a lot of non-frontend baggage that would run if we try to import from it on the front end. That said I think the code around this is fine.

PR #21207, opened 2022-04-26
fix: dedicated socket.io namespace for graphql-refetch messages

tgriesser:
- Closes UNIFY-1693

### User facing changelog

App properly receives graphql related notifications from the server

### Additional details

We need a dedicated socket.io [namespace](https://socket.io/docs/v3/namespaces/) for `data-context` pushes, rather than re-using the one provided via the runner via `window.ws`, because the event-manager calls `.off()` on its socket instance when Cypress execution is stopped. We need to make sure the events here are long-lived.

This also likely explains a lot of situations where we've seen `emitter.toApp()` "not working"

<a href="https://www.loom.com/share/2e2e0bc78722434c8212c83cc68150ee">
    <p>Before</p>
    
  </a>

<a href="https://www.loom.com/share/99513d7784b347f6aef8e9d8451463cb">
    <p>After</p>
    
  </a>

### PR Tasks


packages/frontend-shared/src/graphql/urqlClient.ts
```diff
@@ -161,15 +158,18 @@ interface AppPubSubConfig {
 
 type PubSubConfig = LaunchpadPubSubConfig | AppPubSubConfig
 
+// We need a dedicated socket.io namespace, rather than re-use the one provided via the runner
+// at window.ws, because the event-manager calls .off() on its socket instance when Cypress
+// execution is stopped. We need to make sure the events here are long-lived.
```
flotwig, line 163: What's the reason we need to call `ws.off()`? I see it's a relatively recent change to `event-manager`.

flotwig (approved):

packages/app/cypress/e2e/settings.cy.ts
```diff
@@ -77,6 +77,23 @@ describe('App: Settings', () => {
...
+
+      cy.startAppServer('e2e')
+      cy.loginUser()
+      cy.visitApp()
+      cy.get('.spec-list-container').scrollTo('bottom')
+      cy.contains('test1.js').click()
+      cy.get(`[href='#/settings']`).click()
+      cy.contains('Dashboard Settings').click()
+      cy.contains('Record Key')
+    })
```
flotwig, line 96: This PR touches 8 files and only adds one test, and the test doesn't specifically cover what changed in this PR, only a side-effect. @tgriesser what do you think about adding a test that the graphql ws survives even after the per-spec ws is closed?

packages/frontend-shared/src/graphql/urqlClient.ts
tgriesser: It's been around for [a lot longer](https://github.com/cypress-io/cypress/blame/676fc97f46f9798b5f17ee6d292924f8e1a0c82b/packages/runner-shared/src/event-manager.js#L626) - the recent changes were some code shuffling around:

image

flotwig: Ah gotcha, that's what I deserve for trusting the first blame :see_no_evil: Still, do you know why?

packages/app/cypress/e2e/settings.cy.ts
tgriesser: > and the test doesn't specifically cover what changed in this PR

Yeah I guess it depends on how you define that, it covers it in the sense that there's a failing test, and then changes which result in the test passing. To me that does cover the changes here, and in order to test the graphql ws surviving you need to restructure the code to make it "testable", which I'm not a huge fan of, particularly on the frontend where it's harder to `proxyquire` or something.

I'm curious what your approach to testing it might be?

estrada9166 (approved):

packages/app/cypress/e2e/settings.cy.ts
tgriesser: @flotwig ok, I renamed the test, made it a bit more robust against false positives, added better comments about what it's actually testing

packages/frontend-shared/src/graphql/urqlClient.ts
tgriesser: I guess we don't want to react to any WS events after the test is stopped 🤔, didn't dig into it

PR #21212, opened 2022-04-26
fix: restart dev-server on config change

ZachJW34:
- Closes [UNIFY-1236](https://cypress-io.atlassian.net/browse/UNIFY-1236)

### User facing changelog
Restart dev-server when config is refreshed

### Additional details
The component dev-server is killed whenever the Cypress config changes. A new child process is created, but the dev-server requires an explicit `start` for it to actually start since it is a function that is registered on the `dev-server:start` event. This was only being called once during project initialization so any subsequent changes to the config would never restart the server.

This PR moves the `devServer.start` event into the config loading process such that whenever the config is refreshed, the dev-server will be recreated.

https://user-images.githubusercontent.com/25158820/165393639-86e36241-c15c-4d02-b1ff-ab0caa6f033e.mov

### PR Tasks


packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -219,6 +219,26 @@ export class ProjectLifecycleManager {
...
+            path: this.projectRoot,
+            testingType: this._currentTestingType,
+            specPattern: this.ctx.modeOptions.spec || finalConfig.specPattern,
+            excludeSpecPattern: finalConfig.excludeSpecPattern,
+            additionalIgnorePattern: finalConfig.additionalIgnorePattern,
+          })
+        }
+
+        if (this._currentTestingType === 'component') {
+          const devServerOptions = await this.ctx._apis.projectApi.getDevServer().start({ specs: this.ctx.project.specs, config: finalConfig })
```
ZachJW34, line 233: There is an unhandled edge-case here. If the user modifies their webpack config such that the port of the dev-server is changed, serving tests will fail. We would need to shut down and restart the server in order to fix this, but this seems like a fairly large refactor and not just ct specific.

I haven't done an audit of all config options, but there are a few options that would require a server restart such as `watchForFileChanges` or `hosts` which we currently aren't handling.

tgriesser (changes requested):

packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
```diff
@@ -7,11 +7,11 @@ describe('Cypress In Cypress CT', { viewportWidth: 1500, defaultCommandTimeout:
   beforeEach(() => {
     cy.scaffoldProject('cypress-in-cypress')
     cy.findBrowsers()
-    cy.openProject('cypress-in-cypress')
-    cy.startAppServer('component')
```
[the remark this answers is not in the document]
ZachJW34: There was another test that were duplicating the `openProject` and `startAppServer` call with different arguments, I'm not a fan of calling these more than once so I decided to move them out and into each test.

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -296,6 +306,12 @@ export class ProjectLifecycleManager {
     await this.initializeConfig()
 
     if (this._currentTestingType && this.isTestingTypeConfigured(this._currentTestingType)) {
+      if (this._currentTestingType === 'component') {
+        // Since we refresh the dev-server on config changes, we need to close it and clean up it's listeners
+        // before we can start a new one. This needs to happen before we have registered the events of the child process
+        this.ctx._apis.projectApi.getDevServer().close()
```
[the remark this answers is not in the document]
ZachJW34: I'm honestly not too familiar with the IPC communication, I was just trying to lift and keep the same contract that existed before. There is a [close](https://github.com/cypress-io/cypress/blob/b326693879accd8d78c75f7b7bef8aa211404dc9/npm/webpack-dev-server/src/devServer.ts#L104) function returned with the dev-servers, maybe this is the missing link. Like you said, seems like we are just relying on the child process being auto-killed, I can confirm that the new server spawned doesn't conflict with the old as there would be a port in use error thrown but maybe there is a potential race condition here. I'll dig in a bit, could use some help in this area though

packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
```diff
@@ -156,4 +164,33 @@ describe('Cypress In Cypress CT', { viewportWidth: 1500, defaultCommandTimeout:
...
+  it('restarts dev server on config change', () => {
+    cy.openProject('cypress-in-cypress')
+    cy.startAppServer('component')
+    cy.visitApp()
+
+    cy.withCtx(async (ctx, { testState, sinon }) => {
+      sinon.stub(ctx._apis.projectApi.getDevServer(), 'close')
+      const devServerReady =
+        new Promise((res) => {
+          ctx._apis.projectApi.getDevServer().emitter.on('dev-server:compile:success', (err) => res(true))
```
ryanthemanuel, line 177: Are we doing anything with the value that's resolved? Should this be:

```suggestion
          ctx._apis.projectApi.getDevServer().emitter.on('dev-server:compile:success', (err) => res())
```

packages/app/cypress/e2e/specs_list_actual_git_repo.cy.ts
```diff
@@ -2,10 +2,10 @@ describe('Spec List - Git Status', () => {
   beforeEach(() => {
     cy.scaffoldProject('cypress-in-cypress')
     .then((projectPath) => {
-      cy.task('initGitRepoForTestProject', projectPath)
       cy.wait(500)
```
ryanthemanuel, line 5: If we're moving stuff around, is this wait still needed?

packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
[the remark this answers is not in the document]
ZachJW34: Addressed in [8cabb60](https://github.com/cypress-io/cypress/pull/21212/commits/8cabb60ac598975004fbdf9e276fd6da4d78287f)

packages/data-context/src/data/ProjectLifecycleManager.ts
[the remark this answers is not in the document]
ZachJW34: I dug into this a bit and landed on just removing the unused close functions and event. Killing the child process is safe in terms of cleanup, and we weren't calling the `close` functions from the dev-servers to begin with. I think we could put in a system to allow these child processes to have a cleanup step, but that's outside of the scope of this issue. I can followup with an issue if this is something we should pursue.

packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
ZachJW34: Addressed in [8cabb60](https://github.com/cypress-io/cypress/pull/21212/commits/8cabb60ac598975004fbdf9e276fd6da4d78287f)

packages/app/cypress/e2e/specs_list_actual_git_repo.cy.ts
ZachJW34: Nope! [4bd8b90](https://github.com/cypress-io/cypress/pull/21212/commits/4bd8b90c2bb90d6c2e3dd6b11a74e6f89edcc386)

packages/data-context/src/data/ProjectLifecycleManager.ts
lmiller1990: It would be CT specific since E2E does not use the dev-server.

Also isn't the port always randomly assigned each time webpack-dev-server starts?

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -230,16 +250,6 @@ export class ProjectLifecycleManager {
 
         await this.setInitialActiveBrowser()
 
-        if (this._currentTestingType && finalConfig.specPattern) {
```
lmiller1990, line 233: How come we don't need this anymore?

packages/server/lib/project-base.ts
```diff
@@ -308,46 +300,6 @@ export class ProjectBase<TServer extends Server> extends EE {
     options.onError(err)
   }
 
-  async initializeSpecsAndDevServer (updatedConfig: Cfg): Promise<{
```
lmiller1990, line 311: Nice, I love deleting code from project-base - I think we should be mostly getting rid of this entire thing, eventually.

lmiller1990 (commented):
Definitely fixes a lot of bugs, but looks like we need a bit of work around the hot reload - seems HMR dies after the restart.

I commented in the issue and added a video reproduction: https://cypress-io.atlassian.net/browse/UNIFY-1236?focusedCommentId=17992

packages/data-context/src/data/ProjectLifecycleManager.ts
ZachJW34: We still have it, I just moved it up to the top of the function such that we have the specs available for when we spawn the dev-server

packages/data-context/src/data/ProjectLifecycleManager.ts
ZachJW34: So far my testing has shown that the dev-servers will default to the same port used, I only saw a port switch when explicitly changing the port in the `webpackConfig`. What I have wired up isn't fullproof, but we have [UNIFY-1696](https://cypress-io.atlassian.net/browse/UNIFY-1696) to properly handle config changes that require a server restart.

ZachJW34:
@lmiller1990 I'll create a followup issue for vite not working. Since webpack is our primary audience, I'm ok with getting this merged. I believe [UNIFY-1696](https://cypress-io.atlassian.net/browse/UNIFY-1696) will fix the vite issue.

tgriesser (approved):

lmiller1990 (approved):
Known bug with the Vite restart, issue is logged.

PR #21234, opened 2022-04-27
fix(launchpad): avoid crash when authenticating before selecting project

flotwig:
- Closes UNIFY-1605

### User facing changelog

- Fixes a bug where Cypress would crash if a user tried to log in before selecting a project in interactive mode.

### Additional details

- Also moved some business logic out of `makeDataContext` and into `AuthActions`.
	- As part of this I moved the `Windows` calls to `ElectronApi`.
- Also added unit tests for `AuthActions#logIn`.

### PR Tasks


packages/data-context/src/actions/AuthActions.ts
```diff
@@ -3,12 +3,14 @@ import type { AuthenticatedUserShape, AuthStateShape } from '../data'
...
   getUser(): Promise<Partial<AuthenticatedUserShape>>
-  logIn(onMessage: (message: AuthStateShape) => void): Promise<AuthenticatedUserShape>
+  logIn(onMessage: (message: AuthStateShape) => void, utmCode: string): Promise<AuthenticatedUserShape>
   logOut(): Promise<void>
   resetAuthState(): void
 }
 
 export class AuthActions {
+  private cancelActiveLogin: (() => void) | null = null
+
```
flotwig, line 13: I noticed that this was being set on `coreData`, but never actually used outside of this class... so I made it a local private variable. Any reason I'm missing to not do this? There's nothing about `Actions` not holding local state in the `README`, but I noticed none of them seem to do it.

Or, well, they do have local state, but they put it in `coreData` :stuck_out_tongue:

packages/server/test/unit/gui/auth_spec.js
```diff
@@ -123,38 +122,24 @@ describe('lib/gui/auth', function () {
...
       sinon.stub(auth._internal, 'stopServer')
-      sinon.stub(windows, 'focusMainWindow').callsFake(() => {})
 
-      try {
-        await auth.start(() => {}, 'code', () => {
-          windows.focusMainWindow()
-        })
-      } catch (e) {
-        expect(e.message).to.eql('test error')
-      }
```
flotwig, line 154: this `auth.start` call has actually never rejected - `auth.start` errors are propagated via `onMessage`. so this `catch` block never fired.

packages/data-context/src/actions/ElectronActions.ts
```diff
@@ -32,15 +34,6 @@ export class ElectronActions {
...
 
-  hideBrowserWindow () {
-    // this.electron.browserWindow?.hide()
-    // if (this.isMac) {
-    //   this.ctx.electronApp?.dock.hide()
-    // } else {
-    //   this.electron.browserWindow?.setSkipTaskbar(true)
-    // }
-  }
-
```
flotwig, line 43: removed this and other related dead code... it was cropping up in searches and things like this should be recorded in tasks

packages/data-context/src/actions/AppActions.ts
```diff
@@ -20,7 +20,7 @@ export class AppActions {
   async setActiveBrowser (browser: FoundBrowser) {
     this.ctx.coreData.activeBrowser = browser
 
-    await this.ctx._apis.projectApi.insertProjectPreferencesToCache(this.ctx.lifecycleManager.projectTitle, {
+    this.ctx._apis.projectApi.insertProjectPreferencesToCache(this.ctx.lifecycleManager.projectTitle, {
```
flotwig, line 23:
```suggestion
    await this.ctx._apis.projectApi.insertProjectPreferencesToCache(this.ctx.lifecycleManager.projectTitle, {
```

packages/data-context/src/actions/AuthActions.ts
```diff
@@ -72,10 +86,10 @@ export class AuthActions {
       return
     }
 
-    this.setAuthenticatedUser(user as AuthenticatedUserShape)
+    this.setAuthenticatedUser(user)
 
     this.ctx.update((coreData) => {
-      coreData.cancelActiveLogin = null
+      this.cancelActiveLogin = null
     })
```
tbiethman, line 93: We can loose the wrapping `update` if cancelActiveLogin remains local.

packages/data-context/src/actions/AuthActions.ts
tgriesser: Yeah it was generally because we don't create a new `DataContext`, but we do reset/create a new `coreData` between tests, so it's cleaner to keep it there from a testing perspective. 

We originally wanted to keep state centralized in `coreData` so that we could eventually make it immutable to better track updates to it for debugging, but i'm not sure if we've been consistent in that regard.

I don't feel strongly about it either way, though we've started making `private` members [actually "private"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) elsewhere, so if you want to keep it here, want to change it to `#cancelActiveLogin`?

packages/server/lib/makeDataContext.ts
```diff
@@ -119,6 +107,8 @@ export function makeDataContext (options: MakeDataContextOptions): DataContext {
         return cache.removeAllProjectPreferences()
       },
       insertProjectPreferencesToCache (projectTitle: string, preferences: Preferences) {
+        // FIXME: this should be awaited (since it writes to disk asynchronously) but is not
+        // https://cypress-io.atlassian.net/browse/UNIFY-1705
```
tgriesser, line 111: Any reason we can't just fix that now? Looks like you should just need a `return cache.` here and a type change in `ProjectApiShape` to `Promise<void>`, and an `await` in `setProjectPreferences`

tgriesser: Edit: I see the comment you had in the jira, makes sense

packages/data-context/src/actions/AuthActions.ts
tgriesser: +1

tgriesser (changes requested):
Added some comments on the local state, don't have a strong opinion about it - looks good other than the removing the wrapping update @tbiethman pointed out

packages/data-context/src/actions/AuthActions.ts
flotwig: 7572f7811f

packages/data-context/src/actions/AuthActions.ts
flotwig: Kept it in as `#cancelActiveLogin`: 7572f7811f

tgriesser (approved):

tbiethman (approved):

PR #21235, opened 2022-04-27
feat: Presenting success banner after dependency installation

tbiethman:
- Closes UNIFY-1676, UNIFY-1606

### User facing changelog

* We now show a success banner on the "Install Dev Dependencies" wizard page after all dependences are determined to have been installed.

### Additional details

In addition to the banner being present, a few related components were updated:

* The Alert component wasn't styling its close button appropriately; that is now fixed and matches Figma.
* The verbiage on the "Install Dev Dependencies" page was updated in Figma; the implementation was updated to match.

Before:

After:

Screen Shot 2022-04-27 at 5 14 14 PM

Screen Shot 2022-04-27 at 5 14 56 PM

### PR Tasks


rachelruderman:
This is more of a product question than a question about your PR, but here goes 😸 if we can detect that they've updated their dependencies, why do we even show this banner? Why not automatically bring them to the next step in the migration?

packages/launchpad/cypress/e2e/scaffold-project.cy.ts
```diff
@@ -83,7 +83,11 @@ function scaffoldAndOpenCTProject (opts: {
   }
 
   cy.contains('Next Step').click()
+
+  cy.contains('Install Dev Dependencies').should('be.visible')
+  cy.contains('button', 'Waiting for you to install the dependencies...').should('be.disabled')
```
rachelruderman, line 88:
```suggestion
  cy.contains(cy.i18n.setupWizard.installDependencies.title).should('be.visible')
  cy.contains('button', cy.i18n.setupPage.install.waitForInstall).should('be.disabled')
```

Oooh having made this suggestion, it looks like we're storing the setup install strings in two separate objects: `setupWizard` and `setupPage`. Are they both referring to the same screen? If so could be a good opportunity for a small clean up

marktnoonan (changes requested):
Looks good! Minor change request. 

- There's a little double border situation at the top of the dependencies list in the state without the terminal prompt. 

- It would be good to also have this state, where `wizard.installDependenciesCommand` is falsy, covered in CT and Percy?

Screen Shot 2022-04-28 at 2 50 55 PM

packages/launchpad/cypress/e2e/scaffold-project.cy.ts
rachelruderman: fyi I just made this suggestion w/o testing it, gonna pull down locally now 😅

rachelruderman: Alrighty confirmed it works!

packages/launchpad/src/setup/InstallDependencies.vue
```diff
@@ -1,13 +1,26 @@
...
+      v-if="allDependenciesInstalled && showSuccessAlert"
+      #accessory
+    >
+      <Alert
+        v-model="showSuccessAlert"
+        class="w-full"
+        :icon="CircleCheck"
+        :title="t('setupPage.install.installationAlertSuccess')"
+        status="success"
+        dismissible
```
rachelruderman, line 21: What's the purpose of making this dismissable?

rachelruderman: If the user dismisses it, we're back in the UI state that the new banner intended to mitigate. Also, if it's not dismissable, I think we can get rid of the `showSuccessAlert` variable

tbiethman: The long and short is that it's shown as dismissible within Figma. I think the updates to this view's text mitigate the weirdness better than the banner does, so I'm personally not too concerned about it being dismissible. 

We can circle back around with design in a future ticket if there's a strong preference?

packages/launchpad/cypress/e2e/scaffold-project.cy.ts
tbiethman: I didn't know we exposed the strings that way, interesting! I committed your suggestion and will consolidate those strings.

tbiethman:
> This is more of a product question than a question about your PR, but here goes 😸 if we can detect that they've updated their dependencies, why do we even show this banner? Why not automatically bring them to the next step in the migration?

Good question, I'm not privy to the answer. I think that would have been a valid approach as well, if not a little more "auto-pilot" than the rest of our setup steps.

tbiethman:
@marktnoonan I got the double border removed and the component test updated to show in the appropriate state, you should see a difference in the Percy snapshots.

marktnoonan (approved):

rachelruderman (approved):

PR #21245, opened 2022-04-28
fix: Spec renaming opt out improvements

BlueWinds:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1628

### User facing changelog
Fixes a visual bug with the migration workflow, where we would display a table showing specs to be renamed even if the user had opted out.

Also combines two steps in the opt-out process, reducing clicks and making it clearer to the user what choice they are making.

Before, opting out of migrating spec extensions involved four steps:
Spec renaming 1
Spec renaming 2
Spec renaming 3
Spec renaming 4

Of note, it incorrectly displays that specs will be renamed even if the user has opted out and they won't actually.

After this PR, the flow is shorter:

After 1
After 2

And then once the option to keep their existing data is selected, a bit of extra text pops in,

After 3

The two modal pages have been combined into one (still displaying all text and information from the two modals), and once the user opts out, if there are no files renamings which will occur, then the table is hidden entirely.

### PR Tasks


jennifer-shehane (dismissed):
@BlueWinds The intention of the original designs was to have this as 2 modals to discourage renaming. So I know it was intentionally not meant to be a shorter workflow. We would need design and product approval to change the design of the modals. 

I think the original scope of the issue was to:

1.  fix this formatting and description after choosing to opt out. 
	Screen Shot 2022-04-28 at 2 00 03 PM
1. Show the correct preview of file changes if you are opting out.

BlueWinds:
I understand that the original purpose is to discourage it, and the original scope of the ticket. I saw a poor UX, and took the opportunity to improve it while retaining the original intentions of the design.

More to the point though, there's a difference between discouraging what we consider a bad idea, and actively hindering the user from doing what they've decided. We're still "discouraging renaming", by any reasonable measure of the user experience.

Have you tried out the change? Please try out the current experience and then the new one.

BlueWinds:
Also, pull request reviews are designed exactly for this sort of discussion - anyone with an interest can and should pull down this branch and try out the changes.

Product and design can and should chime in here! Pull request reviews are not only about the code itself. Please don't discourage developers - who you hired for our expertise - from proposing changes, or lock the process of making changes away behind needing to schedule meetings and get pre-approval.

jennifer-shehane:
@BlueWinds This is not a discouragement, just stating what next steps need to be for a change like this. It's not up to me to decide what the implementation should be. If we're making changes to the original Figma design, it needs to be rereviewed by design and product and approved. I know we don't have a very clear process today on how this should be done. Rachel is working on documenting this process and having some better solutions. For now I'll bring this up directly with them and ask how they'd like to move forward.

marktnoonan:
Just note since this is on my review list: I haven't reviewed this yet since it doesn't match the direction we've been given for how this should work. Even though I agree with @BlueWinds on PRs as a good place to discuss changes (and often the best place), and I am in favor of the actual change in this PR to streamline the UI, still seems at this point we need to implement it as intended, and move the general feedback about the behavior into other channels.

marktnoonan:
Oh I think I missed the last line of @jennifer-shehane's comment 

> For now I'll bring this up directly with them and ask how they'd like to move forward.

Should we put this is in draft or something to wait on this direction?

BlueWinds:
> Oh I think I missed the last line of @jennifer-shehane's comment
> 
> > For now I'll bring this up directly with them and ask how they'd like to move forward.
> 
> Should we put this is in draft or something to wait on this direction?

Had a review meeting today, Brian liked the changes and had a couple of suggestions, which I implemented in this latest commit. Updating the PR description with a new screenshot here in a sec.

packages/launchpad/src/migration/OptOutModal.vue
```diff
@@ -0,0 +1,149 @@
...
+          keypath="migration.renameAuto.modal.line6"
+        >
+          <CodeTag class="text-jade-500">
+            [filename].cy.[ext]
+          </CodeTag>
+        </i18n-t>
+      </MigrationListItem>
+    </MigrationList>
+    <Radio
+      v-model:value="value"
```
tbiethman, line 58: There's a lot of `value`s here: Radio's `value`, our `value` ref, the options' `value` attributes. Can we update the ref to be `selectedValue`/`radioValue`/something to help this be a bit more readable?

packages/launchpad/src/migration/OptOutModal.vue
```diff
@@ -0,0 +1,149 @@
...
+            </CodeTag>
+          </i18n-t>
+        </span>
+        <span
+          v-else
+          class="text-gray-800"
+        >
+          {{ props.hasCustomIntegrationFolder ? t('migration.renameAuto.modal.option2') : t('migration.renameAuto.modal.option3') }}
+          <i18n-t
+            v-if="value !== 'rename'"
```
tbiethman, line 83: Nit, but it'd be cool if we updated Radio to provide a `isChecked` value for the option slot. That's really what we're wanting to assert here, that this option is currently selected. What you have works fine, it's just a little indirect.

packages/launchpad/src/migration/OptOutModal.cy.tsx
```diff
@@ -0,0 +1,7 @@
+import OptOutModal from './OptOutModal.vue'
+
+describe('<OptOutModal/>', { viewportWidth: 1119 }, () => {
+  it('renders expected content', () => {
+    cy.mount(OptOutModal)
+  })
+})
```
tbiethman, line 7: Let's add a couple more component tests:
* with `hasCustomIntegrationFolder` enabled
* with `skip` or `renameFolder` selected so we can assert the additional text

We could take percy snapshots for all as well.

tbiethman:
One little thing I noticed is that the additional text we add to the skip/renameFolder option when selected seems to impact the modal layout and shift some elements up/down in the header/footer by a pixel. You can see both bouncing back and forth in this video:

https://user-images.githubusercontent.com/1711637/167212808-b3d84ec4-470e-42c2-864a-b47eb036f689.mov

You can also see it in the component test. Seems to be related to the code element and its increased line-height.

tbiethman (approved):
A couple nitpicky things aside, I like this change. I'm a big fan of removing modals where we can :P Approving for now, but I'd be happy to take another look if something drastically changes.

marktnoonan (changes requested):
Requesting changes as there's a couple of failures in CI, and one of them (the CT failure) is preventing the percy build from completing, so I can't fully review yet.

The code changes look good, though I agree with Tyler the `OptOutModal` test could use some assertions/snapshots, the E2E tests around migration do seem to be exhaustive in covering these paths. 

Happy to review again once CI is green and we can look at percy.

Looking at this does give me some general thoughts about the text here, but I'll take that to the feedback channel to address separately.

packages/launchpad/src/migration/OptOutModal.vue
BlueWinds: Good call, updated.

packages/launchpad/src/migration/OptOutModal.vue
BlueWinds: Added 'checked' to the Radio component.

BlueWinds:
> One little thing I noticed is that the additional text we add to the skip/renameFolder option when selected seems to impact the modal layout and shift some elements up/down in the header/footer by a pixel. You can see both bouncing back and forth in this video:
> You can also see it in the component test. Seems to be related to the code element and its increased line-height.

That was bothering me as I was working on this, but I hadn't spent the time to fix it. Fixed now.

packages/launchpad/src/migration/OptOutModal.vue
```diff
@@ -0,0 +1,149 @@
...
+      <MigrationListItem>
+        <i18n-t
+          scope="global"
+          keypath="migration.renameAuto.modal.line6"
+        >
+          <CodeTag class="text-jade-500">
+            [filename].cy.[ext]
+          </CodeTag>
+        </i18n-t>
+      </MigrationListItem>
```
rachelruderman, line 55: If you wanted to get mega-fancy making this dynamic, you can do:

```
      <MigrationListItem
        v-for="(listItem, index) in listItems.length"
        :key="index"
      >
        <i18n-t
          scope="global"
          :keypath="`migration.renameAuto.modal.listItems[${index}]`"
        >
          <CodeTag class="text-jade-500">
            [filename].cy.[ext]
          </CodeTag>
        </i18n-t>
      </MigrationListItem>
```

Where the script tag has:
```
const { t, messages, locale } = useI18n()
const { listItems } = messages.value[locale.value].migration.renameAuto.modal.listItems
```

and `en-US` has:

```
      "modal": {
        "title": "Change the existing spec file extension",
        "warning": "We recommend using the default extension to avoid inconsistencies, framework conflicts, and confusion with your team.",
        "listItems": [
          "Cypress now supports the ability to create new spec files from within the UI for both E2E and component specs.",
          "All new spec files created within Cypress will use the default pattern of: {0}",
          "We want to rename your existing specs so that they have a consistent filename pattern for both E2E and component testing.",
          "All documentation and example code will be using: {0}",
          "We've changed the placement of component specs to be next to their source files (e.g. src/Button.jsx and src/Button.cy.jsx)",
          "The new default pattern of {0} prevents targeting conflicts with other testing frameworks. (e.g. Jest)"
        ],
```

packages/launchpad/src/migration/RenameSpecsAuto.vue
```diff
@@ -139,8 +132,7 @@ const emits = defineEmits<{
   (eventName: 'selectOption', value: PossibleOption): void
 }>()
 
-const step1Modal = ref(false)
-const step2Modal = ref(false)
+const optOutModal = ref(false)
```
rachelruderman, line 135: Personal preference kinda thang (take it or leave it) I like to start booleans with a verb, i.e. `showOptOutModal`. It is a bit more wordy tho

BlueWinds: Agreed, prefer that as well. Changing.

packages/launchpad/src/migration/OptOutModal.vue
BlueWinds: That feels like we're beginning to make assumptions about the text contents in the code. Sure, right now they all only use `[filename].cy.[ext]` - but if we add another bullet point that needs a different code tag, we'd have to start refactoring the component.

The duplication here (6 `MigrationListItem`) isn't so bad that I feel the need to refactor it yet. If we had a dozen I'd agree entirely.

rachelruderman: All good in the hood! Just want to clarify one thang

> if we add another bullet point that needs a different code tag, we'd have to start refactoring the component.

If we add another bullet point, we wouldn't have to refactor -- the original one would be accessed as `{0}` and the additional one would be accessed as `{1}`

BlueWinds: Sure. We'd just then be cross-referencing via magic numbers, and using/ignoring different interpolated values in different strings, which is a bad sign all around.

BlueWinds:
Ok, took a bit, but should finally have all the tests passing here.

rachelruderman (approved):

marktnoonan (approved):
Tested it out and this works great 👍

PR #21250, opened 2022-04-28
feat: Display Cypress Dashboard metrics in the Specs Explorer

MuazOthman:
- Closes [CLOUD-577](https://cypress-io.atlassian.net/browse/CLOUD-577)

### User facing changelog
* Add new "Latest Runs" and "Average Duration" columns capable of displaying data from the Cypress Dashboard

### Additional details
A large portion of the changes in this PR are to build a GQL mechanism for retrieving & caching Cloud Project data in the app. Whenever possible, previously-cached data is queried and reused by components. As polling proceeds and detects items that may be stale, those individual items can be marked as eligible for re-fetch; once a fetch is initiated they report a status of actively fetching to support display of loading indicators, and finally record a "fetched" or "error" status that can be reflected in the UI.

Incorporate metrics from the Cypress Dashboard (cloud) into the app in the form of "latest runs" statistics and "average duration" measurements on a per-spec basis. The intent is to increase the value of logging into the dashboard from the app and expose potential benefits of recording test runs that users may not be aware of. Two new columns are being added to the Specs Explorer as shown in the screenshots below.

**Logged-out State**
For users who are logged out the columns will display disabled placeholders and be otherwise non-interactive.

Screen Shot 2022-06-26 at 4 42 31 PM

**Activation**
Mousing over the headers will display tooltips describing the action needed to activate these columns. Clicking the login button will start the dashboard login/signup flow and, when complete, will return the user.

Screen Shot 2022-06-26 at 4 30 38 PM

**Active State**
Once logged-in the Specs Explorer will display the latest statistics (if available) from the cloud. Each column will display a quickly-scannable overview of its measurement to indicate potential problem areas that may need improvement.

Screen Shot 2022-06-26 at 4 29 10 PM

Each column header displays an explanation of the data being displayed and provides a link to online docs for each.

Screen Shot 2022-06-26 at 4 29 31 PM

Screen Shot 2022-06-26 at 4 29 39 PM

**Near-realtime Updates**
The app will poll for updates to cloud data for visible specs (within the virtualized list or it's overflow rows) - this polling defaults to every 30 seconds but can be remotely accelerated/decelerated by the dashboard as needed to mitigate any load issues. A single request is made to check if the active project has a recording update within the last 30 seconds - if so, queries are launched for each spec to get the latest data for each. This data is reflected in the app by dynamically-updating row components.

In this screenshot, a recording is currently active - some specs at the top have already Passed, one spec is Running, and the remainder are Queued to be run.

Screen Shot 2022-06-26 at 4 35 11 PM 

**Latest Runs Details**
The row items in the "Latest Runs" column gain a mouseover tooltip that exposes additional details on the most recent runs for the given spec.

Screen Shot 2022-06-26 at 4 29 23 PM

**Not-yet-connected state**
If the user is logged in but the active project has not been connected to the dashboard (has no `projectId` in its config file) the user will be given an actionable header tooltip to connect the project.

Screen Shot 2022-06-26 at 4 48 37 PM

**Disconnected State**
If the user is logged in but the active project has an invalid `projectId` the user will be given an actionable header tooltip to re-connect the project.

Screen Shot 2022-06-26 at 4 31 47 PM

**Unauthorized State**
If the user is logged in and the project is connected to the dashboard but does not have access to the given `projectId` they will be given an action to request access. This will then update to show "request sent" once actioned.

Screen Shot 2022-06-26 at 4 32 40 PM

**Offline State**
If a loss of network connectivity is detected we now show a warning banner at the top of the Spec Explorer to notify the user.

Screen Shot 2022-06-26 at 4 30 17 PM

### Steps to test
- Confirm all "disconnected" states are correct when logged out
- Log in via the CTAs in the header tooltips
- Confirm all project data is shown correctly after login
- Perform actions such as recording runs, navigation, re-authentication, network throttling, etc.
- Confirm all states are shown and respond to actions correctly

### PR Tasks
- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4554

Pre-merge tasks:

- [x] Merge https://github.com/cypress-io/cypress/pull/22347


packages/frontend-shared/cypress/support/mock-graphql/fakeCloudSpecRun.ts
```diff
@@ -0,0 +1,45 @@
...
+    return {
+      id: `SpecRun_${idx}`,
+      status: s,
+      createdAt: new Date('2022-05-08T03:17:00').toISOString(),
+      runNumber: 432,
+      groupCount: 2,
+      specDuration: {
+        min: 143.3333, // 2:23
+        max: 159.767, // 3:40
+      },
```
tgriesser, line 27: These types are defined as `Int` in the `SpecDataAggregate` type, so this is an invalid representation

packages/app/src/specs/SpecHeaderCloudDataTooltip.vue
```diff
@@ -0,0 +1,124 @@
...
+  if (!props.gql.cloudViewer) return 'LOGGED_OUT'
+
+  if (!props.gql.currentProject?.cloudProject?.__typename) return 'NOT_CONNECTED'
+
+  if (props.gql.currentProject?.cloudProject?.__typename === 'CloudProjectNotFound') return 'NOT_FOUND'
+
+  if (props.gql.currentProject?.cloudProject?.__typename === 'CloudProjectUnauthorized') return 'UNAUTHORIZED'
+
+  return 'CONNECTED'
+})
```
tgriesser, line 112: Hmm, this seems correct but makes me wonder about a good way we could represent this in GraphQL, so you could just do:

```ts
currentProject {
  id
  cloudProjectStatus
}
```

Might take a look at that at some point, seems like a common thing we'll want to know across the app

packages/app/src/runs/RunResults.vue
```diff
@@ -19,56 +19,51 @@
...
 
-gql`
-fragment RunResults on CloudRun {
-  id
-  totalPassed
-  totalFailed
-  totalPending
-  totalSkipped
-}
-`
```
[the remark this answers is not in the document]
MuazOthman: Makes sense.

packages/app/src/runs/RunCard.vue
```diff
@@ -50,7 +50,7 @@
       </template>
       <template #right>
         <RunResults
-          :gql="run"
+          v-bind="run"
```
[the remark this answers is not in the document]
MuazOthman: I saw the other related comment and will address.

packages/app/src/specs/SpecRunSummary.vue
```diff
@@ -0,0 +1,87 @@
...
+  specFile: string|null
+}>(), {
+  run: null,
+  specFile: null,
+})
+
+const getAggregateTestCountString = (agg: SpecDataAggregate) => {
+  if (agg.min == null) return '--'
+
+  if (!agg.max || agg.min === agg.max) return agg.min!
```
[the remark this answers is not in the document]
tgriesser: I might be wrong on the specific change needed here, was mostly indicating that we should avoid `!` whenever we can (outside of tests maybe) b/c we can always achieve an equivalent result without lying to the typesystem 😄 . Should probably get a lint rule for this.

flotwig:
Some UI feedback...

1. The table's header overflow gets janky with smaller window sizes, I think this should either adaptively shorten the column title or ellipsize, but definitely not line break: 
image
2. Average duration looks weird, the letter spacing needs to be even, and there isn't enough margin between the numbers and the scrollbar: 
image
3. I'm not sure why these columns show up when we aren't in a project that records to the dashboard, is this intentional? Feels obtrusive when they're completely empty: 
image
4. Kinda minor, but shouldn't the column header be to the left of the scrollbar?
image

packages/app/src/specs/AverageDuration.vue
```diff
@@ -0,0 +1,37 @@
...
+fragment AverageDuration on RemoteFetchableCloudProjectSpecResult {
+  id
+  data {
+    ... on CloudProjectSpecNotFound {
+      retrievedAt
+    }
+    ... on CloudProjectSpec {
+      id
+      retrievedAt
+      averageDuration(fromBranch: $fromBranch)
```
tbiethman, line 27: Is the `$fromBranch` variable value the same defined here: https://github.com/cypress-io/cypress/pull/21250/files#diff-2bc060c37de5ac22696888b84cc57c497d163d9a6506969f06bd65bf7e560b02R49? Or somewhere else?

packages/app/src/specs/LastUpdatedHeader.cy.tsx
```diff
@@ -0,0 +1,29 @@
...
+    cy.mount(<LastUpdatedHeader isGitAvailable={false}/>)
+    cy.findByTestId('last-updated-header').trigger('mouseenter')
+    const tooltipText = defaultMessages.specPage.lastUpdated.tooltip.gitInfoUnavailable
+
+    const tooltipTextSplit = tooltipText.split('{0}').filter((s) => s)
+
+    for (let i = 0; i < tooltipTextSplit.length; i++) {
+      cy.get('.v-popper__popper--shown').contains(tooltipTextSplit[i])
+    }
+  })
```
tbiethman, line 28: Let's get some percy snapshots here as well.

packages/app/src/specs/AverageDuration.vue
MuazOthman: Correct.

packages/app/src/specs/RunStatusDots.vue
```diff
@@ -0,0 +1,204 @@
...
+
+const runs = computed(() => {
+  return props.gql?.data?.__typename === 'CloudProjectSpec' ? props.gql?.data?.specRuns?.nodes ?? [] : []
+})
+
+const dotClasses = computed(() => {
+  const statuses = ['placeholder', 'placeholder', 'placeholder']
+
+  if (runs.value && runs.value.length > 0) {
+    for (let i = 1; i < Math.min(runs.value.length, 4); i++) {
```
tbiethman, line 136: Might be worth a comment to indicate that index 0 represents the latest run and isn't considered here.

packages/app/src/specs/RunStatusDots.vue
```diff
@@ -0,0 +1,204 @@
...
+}>()
+
+const runs = computed(() => {
+  return props.gql?.data?.__typename === 'CloudProjectSpec' ? props.gql?.data?.specRuns?.nodes ?? [] : []
+})
+
+const dotClasses = computed(() => {
+  const statuses = ['placeholder', 'placeholder', 'placeholder']
+
+  if (runs.value && runs.value.length > 0) {
```
tbiethman, line 135: I think everything ends up working out, but we could check to see if this is greater than 1 instead of zero, since we start our indexing at 1.

packages/app/src/specs/SpecRunSummary.cy.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+            <SpecRunSummary run={runs[5]} specFileNoExtension="bankaccounts" specFileExtension=".spec.ts"/>
+
+            <SpecRunSummary run={runs[6]} specFileNoExtension="bankaccounts" specFileExtension=".spec.ts"/>
+
+            <SpecRunSummary run={runs[7]} specFileNoExtension="veryVeryVeryVeryVeryVeryVeryVeryVeryLongSpecFileName" specFileExtension=".spec.ts"/>
+          </div>
+        </div>
+      )
+    })
```
tbiethman, line 31: Let's get a percy snapshot here too.

packages/app/src/specs/SpecsList.vue
```diff
@@ -33,17 +79,44 @@
...
+          :gql="props.gql"
+          :header-text-key-path="'specPage.averageDuration.header'"
+          :connected-text-key-path="'specPage.averageDuration.tooltip.connected'"
+          :not-connected-text-key-path="'specPage.averageDuration.tooltip.notConnected'"
+          :no-access-text-key-path="'specPage.averageDuration.tooltip.noAccess'"
+          :docs-text-key-path="'specPage.averageDuration.tooltip.linkText'"
+          :docs-url="'https://on.cypress.io/specs-average-duration?utm_medium=Specs+Average+Duration+Tooltip&utm_campaign=Average+Duration'"
+          @showLogin="()=>showLogin('Specs Average Duration Tooltip')"
+          @showConnectToProject="showConnectToProject"
+        />
```
tbiethman, line 119: This template is quite large and complex at this point, which indicates to me we could make a few more components out of it. Not sure if changes like that are in scope at this point, but we should consider breaking this up before adding yet more features going forward.

packages/data-context/src/DataContext.ts
```diff
@@ -37,10 +37,12 @@ import { VersionsDataSource } from './sources/VersionsDataSource'
 import type { SocketIONamespace, SocketIOServer } from '@packages/socket'
 import { globalPubSub } from '.'
 import { InjectedConfigApi, ProjectLifecycleManager } from './data/ProjectLifecycleManager'
-import { CypressError, getError } from '@packages/errors'
+import type { CypressError /*, getError*/ } from '@packages/errors'
```
tbiethman, line 40: Can remove commented code here

packages/data-context/src/DataContext.ts
```diff
@@ -223,9 +235,9 @@ export class DataContext {
         // remote contract with the test runner. Showing the main overlay is too heavy-handed of an action
         // to take here, so we only show it in development, when we maybe did something wrong in our e2e
         // Cypress test mocking and want to know immediately in the UI that things are broken
-        if (process.env.CYPRESS_INTERNAL_ENV !== 'production') {
-          return this.onError(getError('DASHBOARD_GRAPHQL_ERROR', err), 'Cypress Dashboard Error')
-        }
+        // if (process.env.CYPRESS_INTERNAL_ENV !== 'production') {
+        //   return this.onError(getError('DASHBOARD_GRAPHQL_ERROR', err), 'Cypress Dashboard Error')
+        // }
```
tbiethman, line 240: Do we want to remove this?

tbiethman:
Looks like the column headers are still conflicting at smaller widths.

Screen Shot 2022-06-13 at 12 10 14 PM

tbiethman:
It appears navigating to a spec and then going back to the specs page consistently results in the runs data not loading. Not sure if this is the same race condition we talked about before?

https://user-images.githubusercontent.com/1711637/173408367-872c8282-eb2e-49e7-bd4b-2f994cb9be9e.mov

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -534,4 +535,21 @@ export class ProjectDataSource {
...
+    }
+
+    // Get the saved state & resolve the lastProjectId
+    const savedState = await source.savedState?.()
+
+    if (savedState?.lastProjectId) {
+      return savedState.lastProjectId
+    }
+
+    // Otherwise, we can try to derive the projectId by reading it from the config file (todo)
```
tbiethman, line 552: Do we need to do this todo?

MuazOthman:
@tbiethman the issue you posted video for seems to be similar to this case: https://cypressio.slack.com/archives/C035VB3HVAL/p1655050339126899

packages/frontend-shared/src/components/ResultCounts.vue
```diff
@@ -0,0 +1,68 @@
+<template>
```
tbiethman, line 1: Can we add some dedicated component tests for this new component?

flotwig:
Super minor UI feedback, but I noticed that if I double click, two tabs open up in my system browser. Users like to double click things when they shouldn't, maybe the click should be debounced.

packages/data-context/src/sources/GraphQLDataSource.ts
```diff
@@ -124,13 +124,19 @@ export class GraphQLDataSource {
...
   #pushFragment (params: PushNodeFragmentParams | PushQueryFragmentParams, isNode: boolean = false) {
     const docBuilder = new DocumentNodeBuilder({
+      isNode,
       parentType: params.info.parentType,
       fieldNodes: params.info.fieldNodes,
-    }, isNode)
+      // operation: params.info.operation,
+    })
+
+    // console.log(print(isNode ? docBuilder.queryNode : docBuilder.query))
```
mike-plummer, line 133: Can we remove the commented-out changes in this file?

packages/data-context/src/sources/CloudDataSource.ts
```diff
@@ -264,10 +299,27 @@ export class CloudDataSource {
...
+
+  async getCache () {
+    await this.#cloudUrqlClient.mutation(`
+      mutation Internal_showUrqlCache { 
+        _showUrqlCache
+      }
+    `, { }, {
+      fetchOptions: {
+        headers: {
+          // TODO: replace this with an exchange to filter out this request
```
mike-plummer, line 317: Are the TODO's in this file pending for this effort, or are these future cleanup items?

flotwig:
@MuazOthman this seems like it's doing a TON of traffic, it seems to be sending about 1,000 requests per minute on my system until I shut it off with `cypress-example-kitchensink`. It's about 1 MB/s of traffic from this process, and it's using a lot of CPU (40% when idling on specslist) presumably from parsing 1,000 responses per second.

Logs from `time CYPRESS_INTERNAL_ENV=staging DEBUG=cypress:network:agent yarn cypress:open |& tee logs.txt`
[logs.txt](https://github.com/cypress-io/cypress/files/8893459/logs.txt)

MuazOthman:
@flotwig can you please make sure to pull `muaz/CLOUD-577-spec-list-display-latest-runs` and rebuild?

MuazOthman:
@flotwig to set expectation, this PR should add one polling request (query to `cloudLatestRunUpdateSpecData`) every 30 seconds and when you first load the page, and one request (contains multiple queries to `cloudSpecByPath`) each time you scroll the spec list to bring in more specs into view with no cached cloud data (with 10 specs scanned ahead in each direction).

packages/app/src/specs/SpecsList.vue
```diff
@@ -129,15 +246,48 @@ import SpecItem from './SpecItem.vue'
...
+import { CloudData_RefetchDocument } from '../generated/graphql-test'
+import type { RemoteFetchableStatus } from '@packages/frontend-shared/src/generated/graphql'
 
 const route = useRoute()
 const { t } = useI18n()
 
+const isOnline = useOnline()
+const isOffline = ref(false)
+
+watch(isOnline, (newIsOnlineValue) => isOffline.value = !newIsOnlineValue)
```
mike-plummer, line 264: Is this really necessary? Seems like we're just using it to control display of the offline banner, so couldn't we just do a `!isOnline` check?

flotwig:
Nice, I was definitely not on the latest, it's no longer DDoSing the dashboard :smile: 

However, it does seem to spray the dashboard with 20 separate requests, one for each spec file, on project startup: 
image

This seems to be contrary to the one request for all files you mentioned.

packages/app/src/specs/SpecsList.vue
```diff
@@ -111,13 +209,32 @@
...
+    v-model="isLoginOpen"
+    :gql="props.gql"
+    :utm-medium="loginUtmMedium"
+    @loggedin="refreshPage"
+  />
+  <CloudConnectModals
+    v-if="isProjectConnectOpen"
+    :gql="props.gql"
+    @cancel="isProjectConnectOpen = false"
+    @success="isProjectConnectOpen = false; refreshPage()"
```
mike-plummer, line 222: Just to keep the implementation of these two modals in alignment, is the `isProjectConnectOpen` update necessary here, or will the `refreshPage` call take care of it the same way we're doing on `LoginModal` above?

MuazOthman:
@tgriesser I think passing a list of IDs to `loadRemoteFetchables` is still sending individual requests to remote per ID for now, or is it expected to batch right now? cc @flotwig

tbiethman:
The rendered values are very tight when scrollbars are visible (and can be overlaid by the scrollbar, depending on OS):

Screen Shot 2022-06-13 at 1 53 09 PM

This seems like it will be corrected one way or another with https://github.com/cypress-io/cypress/pull/22269, just wanted to call it and the current work out here.

packages/app/src/specs/SpecsList.vue
mike-plummer: Let me clarify my question - could we just use `v-if="!isOnline"` to control that modal and dispense with the logic around `isOffline`?

packages/app/src/specs/RunStatusDots.vue
MuazOthman: We do want this condition to pass if `runs.value.length===1`. We can change the condition to >=1 id that makes it more readable. My personal preference is `>0`, but I'm fine with changing it if that makes it more readable.

packages/app/src/specs/AverageDuration.vue
```diff
@@ -0,0 +1,37 @@
+<template>
+  <div
+    v-if="props.gql.data?.__typename === 'CloudProjectSpec' && props.gql.data.averageDuration"
```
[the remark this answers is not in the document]
MuazOthman: There's value in leaving this condition inside `v-if`: it helps TSC to validate:
1. `props.gql.data` is not undefined
2. `props.gql.data.__typename` is `CloudProjectSpec`, which then allows us to access `averageDuration` when rendering content inside the div.

Refactoring to use a computed property will require we change line 7 to something like:
`{{ (props.gql.data?.__typename === 'CloudProjectSpec' && props.gql.data.averageDuration) ? getDurationString(props.gql.data.averageDuration ??) : '' }}`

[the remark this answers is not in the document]
MuazOthman: I may be missing a better approach, so please feel free to correct me.

packages/app/src/specs/SpecsList.vue
```diff
@@ -231,9 +401,27 @@ const treeSpecList = computed(() => collapsible.value.tree.filter(((item) => !it
...
+
+// Watch the sizing of the specs list so we can detect when a scrollbar is added/removed
+// We then calculate the width of the scrollbar and add that as padding to the list header
+// so that the columns stay aligned
+useResizeObserver(containerProps.ref, (entries) => {
+  const specListContainer = entries?.[0]
+  const containerElement = specListContainer?.target as HTMLElement
+
+  if (containerElement) {
+    const displayedScrollbarWidth = containerElement.offsetWidth - containerElement.clientWidth
```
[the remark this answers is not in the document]
MuazOthman: The 'sticky' option makes perfect sense. In fact, this is also what @mike-plummer suggested. However, such solution will be challanging in cobination with how the virtialized list is currently implemented. It seems this whole page requires refactoring for both UX and performance reasons. This is yet another item to be addressed following the initial release.

packages/app/src/specs/SpecsList.vue
MuazOthman: it would work for `v-if` but it won't for `v-model` as that should be a changable value used to track dismissal.

packages/data-context/src/sources/CloudDataSource.ts
MuazOthman: I think this is meant for future work. Right, @tgriesser?

packages/data-context/src/sources/ProjectDataSource.ts
MuazOthman: Thoughts, @tgriesser?

packages/app/cypress/e2e/reporter_header.cy.ts
```diff
@@ -24,6 +24,8 @@ describe('Reporter Header', () => {
       cy.get('[data-cy="spec-file-item"]').should('have.length', 3)
       .should('contain', 'dom-content.spec')
 
+      cy.get('body').type('f')
```
MuazOthman, line 27: I'm not sure why the test won't pass as it was, at least not in my setup. I will need to monitor how CI handles this.

rachelruderman:
Hahaha just took my first peek at this and nearly jumped when I saw >100 changed files 🤣 in previous gigs I would have seen work of this magnitude to be broken down into smaller incremental PRs, what is the approach we take here at Cypress?

MuazOthman:
@rachelruderman yeah this wasn't handled ideally. It felt much smaller at the get go but kept ballooning and we had hard time breaking work items. I'll own this one, and I plan on using this as an example for what not to do as we dig deeper into ACI.

packages/app/src/specs/AverageDuration.vue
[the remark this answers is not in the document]
rachelruderman: I agree with @mike-plummer , using named variables would increase the readability for other developers

> Refactoring to use a computed property will require we change line 7
I don't think we would need to refactor line 7 at all? But could be a Vue thing I'm unfamiliar with

packages/data-context/src/sources/CloudDataSource.ts
rachelruderman: It would be ideal for the TODOs to be linked to a GitHub issue so they don't fall into the ever-expanding black hole of forgotten TODOs across the universe 😆 ⚫

packages/app/src/specs/RunStatusDots.vue
```diff
@@ -0,0 +1,206 @@
...
+    case 'OVERLIMIT':
+    case 'TIMEDOUT':
+      return { icon: ErroredIcon, spin: false }
+    case 'NOTESTS':
+    case 'CANCELLED':
+      return { icon: CancelledIcon, spin: false }
+    default:
+      return { icon: PlaceholderIcon, spin: false }
+  }
+})
```
rachelruderman, line 196: Can bypass the return on line 175 by using optional chaining. Up to you, just personal preference

```suggestion
  switch (latestRun.value?.status) {
    case 'PASSED':
      return { icon: PassedIcon, spin: false }
    case 'RUNNING':
      return { icon: RunningIcon, spin: true }
    case 'FAILED':
      return { icon: FailedIcon, spin: false }
    case 'ERRORED':
    case 'OVERLIMIT':
    case 'TIMEDOUT':
      return { icon: ErroredIcon, spin: false }
    case 'NOTESTS':
    case 'CANCELLED':
      return { icon: CancelledIcon, spin: false }
    default:
      return { icon: PlaceholderIcon, spin: false }
  }
})
```

packages/app/src/specs/AverageDuration.vue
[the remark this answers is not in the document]
MuazOthman: It is actually a TypeScript limitation. As I understand it, the TypeScript compiler and Vue tooling will evaluate `v-if` as if it was a regular `if` statement in code, meaning that the compiler will factor in the fact the condition must evaluate to `true` when evaluating the contents of the element with `v-if`.

This allows TypeScript to make the correct assumption that `props.gql.data` is a defined object with the shape correlated with `__typename === 'CloudProjectSpec'`, in contrast to other options in the union type where `__typename` is  `'CloudProjectSpecNotFound'` or `'CloudProjectUnauthorized'` which lack the definition of `averageDuration` field we're trying to access.

Maybe this screenshot can better show how the refactored code can throw off TSC
image

For reference, this is the GQL union type definition:
https://github.com/cypress-io/cypress/blob/d80e719b32c5621f5ac79e92d200c789d3e529dc/packages/graphql/schemas/cloud.graphql#L278-L281

The corresponding generated TypeScript code (not checked in since it is generated) can be found in `packages/app/src/generated/graphql.ts` and here's a formatted version of it:

```ts
type AverageDurationFragment = {
  readonly __typename?: "RemoteFetchableCloudProjectSpecResult"
  readonly id: string
  readonly data:
    | {
        readonly __typename?: "CloudProjectSpec"
        readonly id: string
        readonly retrievedAt: string | null
        readonly averageDuration: number | null
      }
    | {
        readonly __typename?: "CloudProjectSpecNotFound"
        readonly retrievedAt: string | null
      }
    | { readonly __typename?: "CloudProjectUnauthorized" }
    | null
}

```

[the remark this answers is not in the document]
rachelruderman: TIL 😸 thank you for the detailed explanation!

packages/app/cypress/e2e/specs_list_latest_runs.cy.ts
```diff
@@ -0,0 +1,570 @@
+import defaultMessages from '@packages/frontend-shared/src/locales/en-US.json'
```
MuazOthman, line 1: @mike-plummer this seems to be at a good stage for you to review when you get the chance

packages/app/src/specs/LastUpdatedHeader.vue
```diff
@@ -0,0 +1,60 @@
...
+      <div
+        class="flex flex-col text-sm text-center max-w-300px p-4 items-center"
+      >
+        <i18n-t
+          v-if="props.isGitAvailable"
+          scope="global"
+          keypath="specPage.lastUpdated.tooltip.gitInfoAvailable"
+        >
+          <ExternalLink
+            href="https://on.cypress.io/specs-last-updated?utm_medium=Specs+Last+Updated+Tooltip&utm_campaign=Last+Updated"
```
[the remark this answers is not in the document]
MuazOthman: Thanks for catching this, Mark.
I also realized that we're still sending `Test Runner` as `utm_source` for in the log in flow
https://github.com/cypress-io/cypress/blob/ad7668e5e3ace954921a43911c8b624f7b95794e/packages/server/lib/gui/auth.ts#L48

I'll address both cases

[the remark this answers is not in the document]
MuazOthman: Fixed. Please see: https://github.com/cypress-io/cypress/pull/21250/files/52d738fdbce6ccaef439f1831fbf42d656961f39..3003f560b4576254dfcd13ece3a98dfaf7bf63fd

packages/app/src/specs/RunStatusDots.vue
tbiethman: A length of 1 would result in a no-op for the for-loop, right? Which is fine functionally, but it seems like we could just skip it in that case.

packages/app/cypress/e2e/runs.cy.ts
```diff
@@ -107,6 +107,12 @@ describe('App: Runs', { viewportWidth: 1200 }, () => {
 
       cy.findByTestId('sidebar-link-runs-page').click()
 
+      // TODO: investigate the scenario for this test
```
MuazOthman, line 110: @estrada9166 can you please review this test? The new GQL implementation for remote queries (cloud) adds caching, which seems to interfere with this test. Busting the cache at his point as I added passes the test, but it may turn it into an unrealistic test. I'm not sure about the details of this scenario/feature. Can you please take a look?

packages/app/cypress/e2e/runs.cy.ts
```diff
@@ -147,6 +153,12 @@ describe('App: Runs', { viewportWidth: 1200 }, () => {
...
```
MuazOthman, line 156: @estrada9166 this is also related to caching. Can you review this one?

packages/app/cypress/e2e/runs.cy.ts
```diff
@@ -611,6 +623,12 @@ describe('App: Runs', { viewportWidth: 1200 }, () => {
       })
 
       cy.findByTestId('sidebar-link-runs-page').click()
+      // TODO: investigate the scenario for this test
```
MuazOthman, line 626: @estrada9166 this is also related to caching. Can you review this one?

packages/app/cypress/e2e/subscriptions/createCloudOrgModal-subscription.cy.ts
```diff
@@ -30,6 +30,12 @@ describe('App: Runs', { viewportWidth: 1200 }, () => {
 
       cy.findByTestId('sidebar-link-runs-page').click()
 
+      // TODO: investigate the scenario for this test
```
MuazOthman, line 33: @estrada9166 this is also related to caching. Can you review this one?

marktnoonan (commented):
I've added some comments, mostly on frontend stuff. I haven't reviewed the tests much yet, and didn't get into detail on the data-context and graphql changes (which I believe have been reviewed as PRs of their own going into this branch).

There are a few things that I'd consider blockers to merging this to develop, that we should discuss:

- The diff for this PR is 5,000 lines and the PR description is 6 words - it should document the changes and rationale for what has been done, explain how to test, the works. It's a big change and lots of head scratching is in our future if we merge without a real description. It will be easier for reviewers who have less context than me to review if there are the usual details and screenshots etc.
- There are a few existing open questions that don't seem to have been answered, I've tagged people in the threads to get responses, let's get them answered and either complete the TODOs or make the issues for them before we lose context.
- The main technical thing that might be blocker is the fallback to an empty string [on this line](https://github.com/cypress-io/cypress/blob/44641ba7090152bbbb4530f4fe97f23d36549722/packages/app/src/specs/SpecsList.vue#L548). An empty string doesn't seem to work downstream, so if there's a chance that the `id` on a `cloudSpec` might really be missing, we'd need to handle it differently. This seems really minor though.

To summarize my other comments:

- There are some places we can remove reactivity that's not needed - computed values and prop bindings that don't use dynamic data.
- We might need to add cleanup for some directory-expansion watchers in the specs list, and we can definitely create fewer watchers be being more selective about when they are needed.
- There are some opportunities to reduce the amount of code and write more idiomatic Vue
- A small amount of deleted code seems to have re-entered the chat after a merge conflict was resolved.

I hit two minor visual bugs. One is as you move your mouse to the login button, if you happen to pass over a table header, it can get stuck open and sit on top of the modal backdrop:

Screen Shot 2022-06-25 at 11 38 46 AM

The other is, at the smallest breakpoint here, the vertical alignment inside the rows gets wrong. The "New Spec" button doesn't wrap nicely either, though that's not a part of this PR:

Screen Shot 2022-06-25 at 4 17 24 PM

In terms of accessibility, we have some known issues to address and I found a couple of others that I point out in the comments. I'd like to create github issues for these problems and put TODOs pointing to them in the code so we can get things prioritized as quickly as possible after release. A few of the issues are solvable with small HTML changes and I'd suggest we either do those in this PR or, after it is merged, make a follow along PR into develop with the tweaks and the issue numbers for the future TODOs.

My next steps on Monday will be to review the test code itself, run the tests locally, and do more manual testing. A "how to test" section in the PR would be helpful for this so we can hit some kind of matrix of scenarios.

packages/frontend-shared/src/composables/useCollapsibleTree.ts
```diff
@@ -38,9 +45,13 @@ function collectRoots<T extends RawNode<T>> (node: UseCollapsibleTreeNode<T> | n
 }
 
 export const useCollapsibleTreeNode = <T extends RawNode<T>>(rawNode: T, options: UseCollapsibleTreeOptions, depth: number, parent: UseCollapsibleTreeNode<T> | null): UseCollapsibleTreeNode<T> => {
+  const { cache, expandInitially } = options
   const treeNode = rawNode as UseCollapsibleTreeNode<T>
   const roots = parent ? collectRoots<T>(parent) : []
-  const [expanded, toggle] = useToggle(!!options.expandInitially)
+  const [expanded, toggle] = useToggle(cache?.get(rawNode.id) ?? !!expandInitially)
+
+  watch(() => expanded.value, (newValue) => cache?.set(rawNode.id, newValue))
```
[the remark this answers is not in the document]
mike-plummer: Agreed, this was definitely doing more than it needed to and may have been a leak. I've reworked it a bit to instead wrap the `toggle` function to replace the logic from the `watch`, and only do so for non-hidden non-leaf nodes. This should eliminate any cleanup needed for this. Local testing looks good for both `SpecsList` and `InlineSpecList`

packages/app/cypress/e2e/reporter_header.cy.ts
```diff
@@ -24,11 +24,7 @@ describe('Reporter Header', () => {
       cy.get('[data-cy="spec-file-item"]').should('have.length', 3)
       .should('contain', 'dom-content.spec')
 
-      cy.get('input').clear()
-
-      cy.get('[data-cy="spec-file-item"]').should('have.length', '3')
```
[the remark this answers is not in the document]
mike-plummer: I can't explain why these lines were removed. The test passes with them added back in so I've reverted the change

packages/app/cypress/e2e/specs_list_e2e.cy.ts
```diff
@@ -112,19 +112,19 @@ describe('App: Spec List (E2E)', () => {
 
   describe('typing the filter', function () {
     it('displays only matching spec', function () {
-      cy.get('button').contains('14 Matches')
+      cy.get('button').contains('23 Matches')
```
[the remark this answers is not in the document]
lmiller1990: I think we'd add a task that fetches the correct number, if we do want to address this one in the future.

packages/data-context/src/sources/ProjectDataSource.ts
lmiller1990: This TODO doesn't make sense, we already tried this on one 544 - if it's not there, it's not available, unless they are in global mode.

Why do we even need to do this here? Eg - why grab from the savedState at all? 

The only use case I can think of is pre-fetching runs for the previous project, when we are launching in global mode and before we've chosen a project, but this seems like an optimization that's unlikely to be super useful. Why not just remotely fetch when they choose a project?

packages/data-context/test/unit/sources/RemoteRequestDataSource.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      expect(remoteRequestSource.unpackFetchableNodeId('UmVtb3RlRmV0Y2hhYmxlQ2xvdWRQcm9qZWN0U3BlY1Jlc3VsdDo1Y2MyNWQ4YTM5YTY1NGViMjNiNTI1NzM0NWFiYmY0MmJlNDBjOGQxOmV5SmhJam94TENKaUlqb2lNaUo5')).to.eql({
+        name: 'RemoteFetchableCloudProjectSpecResult',
+        operationHash: '5cc25d8a39a654eb23b5257345abbf42be40c8d1',
+        operationVariables: { a: 1, b: '2' },
+      })
+    })
+  })
+
+  describe('batchResolveRemoteFields', () => {
+    it('')
```
lmiller1990, line 60: Should we delete these, or make an issue? I'd recommend pick one - it's unlikely someone is going to look at this file and retroactively write these if we don't.

packages/frontend-shared/src/components/ResultCounts.vue
```diff
@@ -0,0 +1,68 @@
...
+        :class="result.class"
+      />
+      <span class="sr-only">{{ result.name }}</span>
+      {{ result.value }}
+    </div>
+  </div>
+</template>
+
+<script lang="ts" setup>
+export interface ResultCountsProps {
```
lmiller1990, line 22: I didn't think you can export from `script setup`. I am surprised this works 🤔

packages/frontend-shared/src/components/ResultCounts.vue
```diff
@@ -0,0 +1,68 @@
...
+  totalPassed: number|string|null
+  totalFailed: number|string|null
+  totalPending: number|string|null
+  totalSkipped: number|string|null
```
lmiller1990, line 26: Shouldn't this be either string or number, not both?

packages/frontend-shared/src/components/Tooltip.vue
```diff
@@ -1,28 +1,70 @@
 <template>
-  <Tooltip
+  <component
```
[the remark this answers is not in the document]
lmiller1990: I'd agree with Mark - looks like this is two components, masquerading as one.

Maybe an issue with what you'd like to see. So far, many of our TODOs have remained TODOs for now - I'd also like to get a handle on the frontend which is starting to get a bit messy.

packages/frontend-shared/src/components/Tooltip.vue
```diff
@@ -90,4 +132,42 @@ withDefaults(defineProps<{
   }
 }
 
+.v-popper__popper.v-popper--theme-menu {
```
lmiller1990, line 135: I did not see the mocks but do we need to do this? 
Overriding a third party library styles seems like it'll be very hard to maintain. We have no guarentee they won't change the classes in a minor version bump.

packages/frontend-shared/src/composables/useCollapsibleTree.ts
```diff
@@ -38,20 +45,37 @@ function collectRoots<T extends RawNode<T>> (node: UseCollapsibleTreeNode<T> | n
...
   const treeNode = rawNode as UseCollapsibleTreeNode<T>
   const roots = parent ? collectRoots<T>(parent) : []
-  const [expanded, toggle] = useToggle(!!options.expandInitially)
+  const [expanded, toggle] = useToggle(cache?.get(rawNode.id) ?? !!expandInitially)
+
   const hidden = computed(() => {
     return !!roots.find((r) => r.expanded.value === false)
   })
 
+  let wrappedToggle = toggle
```
lmiller1990, line 57: Rather than conditionally reassigning a function, what about just a single function that has a conditional? This seems more idiomatic in general, and a bit easier to understand.

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -150,6 +150,7 @@
                 <Auth
                   :gql="props.gql"
                   :show-logout="true"
+                  :utm-medium="'Nav'"
```
lmiller1990, line 153:
```suggestion
                  utm-medium="Nav"
```

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -170,6 +171,7 @@
       <LoginModal
         v-model="isLoginOpen"
         :gql="props.gql"
+        :utm-medium="'Nav'"
```
lmiller1990, line 174:
```suggestion
        utm-medium="Nav"
```

packages/frontend-shared/src/graphql/urqlClient.ts
```diff
@@ -114,30 +152,10 @@ export async function makeUrqlClient (config: UrqlClientConfig): Promise<Client>
...
-            })
-          }, 1000)
-        }
-
-        // eslint-disable-next-line
-        console.error(error)
+        showError(error)
       },
     }),
+    // detailedDebugExchange,
```
lmiller1990, line 158: Should we delete this comment?

packages/frontend-shared/src/graphql/urqlDetailedDebugExchange.ts
```diff
@@ -0,0 +1,65 @@
...
+
+window.__toPrintClear = function () {
+  window.__toPrint__ = {
+    queries: [],
+    logs: [],
+    data: {},
+  }
+}
+
+export const detailedDebugExchange: Exchange = ({ forward }) => {
```
lmiller1990, line 27: Seems this is not used?

packages/server/lib/makeDataContext.ts
```diff
@@ -73,8 +75,8 @@ export function makeDataContext (options: MakeDataContextOptions): DataContext {
       getUser () {
         return user.get()
       },
-      logIn (onMessage, utmCode) {
-        return auth.start(onMessage, utmCode)
+      logIn (onMessage, utmSource, utmMedium) {
```
lmiller1990, line 78: No types? Or are these somehow inferred?

lmiller1990 (commented):
I left a bunch of comments/questions. Happy to do a sync/pair code review for more clarity. I did NOT test this out locally yet.

packages/app/cypress/e2e/subscriptions/specChange-subscription.cy.ts
```diff
@@ -12,7 +12,8 @@ describe('specChange subscription', () => {
   describe('specs list', () => {
     it('responds to specChange event for an added file', () => {
       cy.get('[data-cy="spec-item-link"]')
-      .should('have.length', 14)
+      // cannot assert a length since this is a virtualized list
```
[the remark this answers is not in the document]
marktnoonan: It doesn't, but it makes it a bit more complex, since the number of nodes rendered is determined by the height of the list in the current conte plus the `overscan` number that is configured for the list in the Vue component that renders it. It will sometimes be less than then number of matching results for a given search term.

packages/frontend-shared/src/graphql/urqlDetailedDebugExchange.ts
tgriesser: This was useful when debugging issues we were seeing around the urql cache not hitting properly, but it's extremely verbose. I can add a comment here that it can be imported/added to the middleware if you need to do advanced debugging of things in URQL req/response payloads

packages/data-context/test/unit/sources/RemoteRequestDataSource.spec.ts
tgriesser: Yeah I think I meant to delete it - the batch resolution is implemented in #22457

packages/server/lib/makeDataContext.ts
```diff
@@ -93,8 +95,13 @@ export function makeDataContext (options: MakeDataContextOptions): DataContext {
       insertProjectToCache (projectRoot: string) {
         return cache.insertProject(projectRoot)
       },
-      getProjectRootsFromCache () {
-        return cache.getProjectRoots()
+      async getProjectRootsFromCache () {
+        return await Bluebird.map(cache.getProjectRoots(), async (projectRoot: string) => {
```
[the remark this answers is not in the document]
tgriesser: Yeah probably, the one thing that `Bluebird.map` has that `Promise.all` doesn't is the ability to control concurrency of promises in flight at a given time, which is extremely useful when you're dealing with remote requests. That isn't the case here though, so we can change to a `Promise.all`

packages/frontend-shared/src/graphql/urqlClient.ts
tgriesser: Yeah

packages/frontend-shared/src/components/ResultCounts.vue
tgriesser: I guess you can export types

packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -142,7 +143,9 @@ export class ProjectActions {
   async loadProjects () {
     const projectRoots = await this.api.getProjectRootsFromCache()
 
-    this.projects = projectRoots.map((projectRoot) => ({ projectRoot }))
+    this.ctx.update((d) => {
+      d.app.projects = [...projectRoots]
```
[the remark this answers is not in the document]
tgriesser: I think this is fine to keep since we don't have any strong guarantee that we don't accidentally mutate this elsewhere

packages/app/src/specs/SpecsList.vue
```diff
@@ -33,17 +80,48 @@
...
+          :docs-url="latestRunsDocsUrl"
+          data-cy="latest-runs-header"
+          @showLogin="showLogin('Specs Latest Runs Tooltip')"
+          @showConnectToProject="showConnectToProject"
+        />
+      </div>
+      <div class="hidden items-center justify-end truncate md:flex md:col-span-2">
+        <SpecHeaderCloudDataTooltip
+          :gql="props.gql"
+          header-text-key-path="specPage.averageDuration.header"
```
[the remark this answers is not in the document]
tgriesser: I don't know much about how we actually use the i18n stuff either

packages/app/src/runs/RunsErrorRenderer.vue
```diff
@@ -126,13 +126,27 @@ mutation RunsErrorRenderer_RequestAccess( $projectId: String! ) {
...
+
 const requestAccessMutation = useMutation(RunsErrorRenderer_RequestAccessDocument)
 
-function requestAccess () {
+async function requestAccess () {
   const projectId = props.gql.currentProject?.projectId
 
   if (projectId) {
-    requestAccessMutation.executeMutation({ projectId })
+    const result = await requestAccessMutation.executeMutation({ projectId })
```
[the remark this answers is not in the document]
tgriesser: @mike-plummer I think you might have worked on this recently - any thoughts here?

packages/frontend-shared/src/composables/useCollapsibleTree.ts
tgriesser: @mike-plummer thoughts on this one?

packages/frontend-shared/src/components/Tooltip.vue
tgriesser: @mike-plummer thoughts on this one?

packages/app/src/specs/SpecHeaderCloudDataTooltip.vue
```diff
@@ -0,0 +1,204 @@
...
+
+    return 'UNAUTHORIZED'
+  }
+
+  return 'CONNECTED'
+})
+
+const requestAccessMutation = useMutation(SpecHeaderCloudDataTooltip_RequestAccessDocument)
+
+async function requestAccess () {
```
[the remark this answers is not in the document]
tgriesser: Rather than a composable, I'd see it being encapsulated in a `RequestAccessButton.vue` or something. I'll defer to @mike-plummer on whether we want to change this though

packages/app/cypress/e2e/specs_list_latest_runs.cy.ts
```diff
@@ -0,0 +1,611 @@
...
+
+  // TODO: add link verification
+  // if (latestRunStatus !== 'PLACEHOLDER') {
+  //   cy.get(`${specRowSelector(specFileName)} [data-cy="run-status-dots"]`).validateExternalLink('https://google.com')
+  // }
+}
+
+function simulateRunData () {
+  cy.remoteGraphQLIntercept(async (obj) => {
+    const fakeRuns = (statuses: string[], idPrefix: string) => {
```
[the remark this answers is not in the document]
tgriesser: Yeah this should probably be defined elsewhere

tgriesser:
> This TODO doesn't make sense, we already tried this on one 544 - if it's not there, it's not available, unless they are in global mode.

> Why do we even need to do this here? Eg - why grab from the savedState at all?

It was something that @brian-mann had wanted in original designs, the ability to show recent runs for a project in global mode. This was mostly just illustrating it being possible, I clarified the `todo` to call out that we should add it if we ever want to do that.

packages/app/src/specs/RunStatusDots.vue
```diff
@@ -0,0 +1,210 @@
...
+  })
+})
+
+const latestRun = computed(() => {
+  if (runs.value == null || runs.value.length === 0 || runs.value[0] === null) {
+    return null
+  }
+
+  return runs.value[0]
+})
```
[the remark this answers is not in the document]
tgriesser: Fixed

packages/app/src/specs/SpecsList.vue
```diff
@@ -213,27 +400,56 @@ function handleClear () {
...
   if (!debouncedSearchString.value) {
-    return specs
+    return fuzzyFoundSpecs
   }
 
-  return fuzzySortSpecs(specs, debouncedSearchString.value)
+  return fuzzySortSpecs(fuzzyFoundSpecs, debouncedSearchString.value)
 })
 
+// Maintain a cache of what tree directories are expanded/collapsed so the tree state is visually preserved
```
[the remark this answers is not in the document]
mike-plummer: Absolutely agree that `SpecsList` has become unwieldy. I think we could at a minimum split this into a management component to maintain the overall data and interactions, and a second to handle the display aspects with the virtualized list and such. Not something I'm eager to do at this moment but I'd love to do this as a cleanup item.

packages/frontend-shared/src/composables/useCollapsibleTree.ts
mike-plummer: I think that makes sense - we aren't necessarily memoizing on the toggle function anywhere so there isn't a strong argument to avoid wrapping it. May as well always do the wrap and move the conditional inside

packages/frontend-shared/src/components/Tooltip.vue
mike-plummer: Unfortunately yes, if we want to change the default styles from `floating-vue` this is basically the approach we have to take. The library does have a [theme system ](https://floating-vue.starpad.dev/guide/themes.html#custom-theme) but it's really just a thin veneer to allow layering/sharing CSS defns - we're defining the `menu` theme here the same way we had already defined a `tooltip` theme just above this

packages/app/src/runs/RunsErrorRenderer.vue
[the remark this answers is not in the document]
mike-plummer: I restructured this a bit and used it as a pattern for `SpecHeaderCloudDataTooltip.vue` - my suspicion is that Muaz added the component-level state because the GQL state update wasn't circling back and causing the component to update. I don't know enough about the GQL stack yet to say whether that was expected behavior from the mutation, or whether this component needed some sort of subscription to be updated when that data updated. I agree that ideally this manual state management *shouldn't* be necessary - hopefully someone who knows more about our GQL setup can point me in the right direction.

packages/app/src/specs/SpecHeaderCloudDataTooltip.vue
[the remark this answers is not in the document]
mike-plummer: I agree that this is duplicated behavior and we should combine them. To do it right I would want a better understanding around the other comment regarding our manual state management here vs. the GQL mutation flowing the updated state back into the component

packages/app/cypress/e2e/specs_list_latest_runs.cy.ts
```diff
@@ -0,0 +1,611 @@
...
+
+      simulateRunData()
+
+      cy.visitApp()
+      cy.findByTestId('sidebar-link-specs-page').click()
+    })
+
+    it('shows correct tooltips', () => {
+      cy.findByTestId('latest-runs-header').trigger('mouseenter')
+      cy.contains('.v-popper__popper--shown', 'The status of your latest runs in the Cypress Dashboard')
```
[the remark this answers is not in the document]
mike-plummer: In this case it's first validating the popper shows specific text then on the next line validates that it does not contain an action button. Would probably be better to use a `.find` so it's more obvious it's proceeding within that scope

packages/app/src/runs/RunsErrorRenderer.vue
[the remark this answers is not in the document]
marktnoonan: > Based on mutation result, GrpahQL cache (urql client) updates

Isn't this the kind of behavior we'd need a subscription to get now? That (since the removal of `liveMutation`) we're no longer syncing with the client?

I might be conflating two different situations though.

packages/app/src/specs/SpecsList.vue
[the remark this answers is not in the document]
marktnoonan: Yeah we still need the paths somewhere, but moving that logic inside would reduce the duplication here.

packages/app/src/specs/SpecsList.vue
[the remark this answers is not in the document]
marktnoonan: We should definitely have a chat soon about the future of the specs list, I think we can simplify the specs list and the search functionality a whole lot and drop virtualization. I'll put something on the calendar for after the break and invite some the 4 of us, we can add more people if needed.

packages/app/src/specs/SpecsList.vue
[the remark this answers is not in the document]
mike-plummer: I restructured this component a bit to accept a `mode` prop and, based on that value, the component itself maintains the i18n keys and UTM URL for display. Moves ~30 lines of logic from `SpecsList` into `SpecHeaderCloudDataTooltip` and feels a bit better from an encapsulation perspective

packages/app/src/runs/RunsErrorRenderer.vue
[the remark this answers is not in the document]
tgriesser: > Isn't this the kind of behavior we'd need a subscription to get now?

No, in most cases this should happens automatically based on the requested fields in the mutation response. 

Looking at this again, I'm pretty sure the issue here is that because there isn't an identifier tying this to the locally resolved project, without what we're doing here, we need to add some updater logic to the `graphcacheExchange` to keep this in-sync correctly in the cache.

packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -197,6 +200,13 @@ export class ProjectActions {
   async addProject (args: AddProject) {
     const projectRoot = await this.getDirectoryPath(args.path)
 
+    const found = this.projects.find((x) => x.projectRoot === projectRoot)
```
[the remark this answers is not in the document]
tgriesser: Good catch, maybe this was a bad merge. Will update

packages/frontend-shared/cypress/support/mock-graphql/stubgql-Mutation.ts
```diff
@@ -119,4 +123,20 @@ export const stubMutation: MaybeResolver<Mutation> = {
 
     return {}
   },
+  loadRemoteFetchables (source, args, ctx) {
+    return args.ids.map((id) => {
+      const [__typename, hash] = atob(id).split(':')
```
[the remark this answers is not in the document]
tgriesser: Fixing on the frontend

packages/app/src/specs/SpecListGitInfo.vue
```diff
@@ -1,23 +1,33 @@
...
       placement="top"
-      class="h-full grid items-center"
+      class="h-full truncate"
+      data-cy="tooltip"
     >
-      <component
-        :is="classes.icon"
-        :class="classes.iconClasses"
-      />
+      <div class="flex h-full gap-9px justify-start items-center">
```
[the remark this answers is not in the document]
tgriesser: Changed from a `div` to a `button`

mike-plummer (approved):

packages/frontend-shared/src/gql-components/topnav/LoginModal.vue
```diff
@@ -107,13 +108,23 @@ import type { LoginModalFragment } from '../../generated/graphql'
 
 const online = useOnline()
 
+function continueAuth (isLoggedIn: boolean) {
+  if (isLoggedIn) {
+    emit('loggedin')
```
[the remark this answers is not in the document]
tgriesser: Seems like this had other side-effects actually

tbiethman (approved):

packages/app/src/specs/SpecsList.vue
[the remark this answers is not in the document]
marktnoonan: Agreed. Scrolling past more than a couple of screens to find a spec is not something I think most people will really do. I think they'll type the search term. Talked some with Ryan about the possible future implementations around this today, since we were talking anyway and it sounds like we are pretty open to different solutions around this list.

packages/app/src/runs/RunsErrorRenderer.vue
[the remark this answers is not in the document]
lmiller1990: Is this going to be a problem or can we dismiss this as "won't do"? Otherwise, do we need to file an issue (likely with something more detailed than "add a custom updater" since many people might not know what action is needed. What do you think?

pstakoun (approved):

lmiller1990 (approved):
All the major blockers are addressed. Approving, with the following in mind:

- Fast follows are reviewed and followed up, [epic link](https://github.com/cypress-io/cypress/issues/22547). There's definitely some technical debt to be audited and addressed.
- I did not battle test this personally, this was done and owned by ACI team. I focused on the code review primarily.

:heavy_check_mark:

PR #21103, opened 2022-04-15
fix(unify): Specs list within the test runner should be collapsed by default

davidmunechika:
- Closes [UNIFY-1421](https://cypress-io.atlassian.net/browse/UNIFY-1421?atlOrigin=eyJpIjoiODJkYzkxNDM5Mzc2NGE0OWI1NTcwOWQ1NGRkMmU3YzYiLCJwIjoiaiJ9)

### User facing changelog
The specs list within the test runner should be collapsed by default (currently the specs list is expanded by default)

### Additional details

Slack conversation:
https://cypressio.slack.com/archives/C02MYBT9Y5S/p1648592192269109

### PR Tasks

- [x] Have tests been added/updated?
- [ ] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [n/a] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [n/a] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [n/a] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

PR #21218, opened 2022-04-27
chore: cache Percy's Chromium browser at yarn install

flotwig:
- Closes 

### User facing changelog

n/a

### Additional details

- Percy can be sped up by not downloading their Chromium instance every single time we use Percy CLI. In particular, this will prevent flake like https://app.circleci.com/pipelines/github/cypress-io/cypress/36922/workflows/20dbee90-6950-4a81-b2f0-8337f65acc65/jobs/1482435 - see Slack thread: https://cypressio.slack.com/archives/C02BYH0P937/p1651017099108499

### PR Tasks

- [na] Have tests been added/updated?
- [na] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [na] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [na] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [na] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

PR #21254, opened 2022-04-29
fix: reporter resizing in small viewports and AUT scaling when specs list is open

marktnoonan:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1471

### User facing changelog

This PR fixes three issues:

1. A previously harmless issue with the reactivity of how the `isSpecsListOpen` value was revealed by #21103. It meant when the specs list was closed by default, expanding it didn't add the width of the specs list to the scaling math, so the AUT would get cut off.
2. In some situations, resizing the window could put the reporter and specs list into a "locked" state, which had several negative side effects.
3. The handle for resizing the inline specs list had become harder to "grab" as it was only active on the left hand side, due to some z-indexes changing that placed it below the reporter in the stacking order.

### Additional details

[Loom demo of the changes in this PR](https://www.loom.com/share/255baa2019164dc6b2811127cf8aaf37).

To solve the first issue, we now access the `isSpecsListOpen` via `runnerUIStore.isSpecsListOpen` in the `useRunnerStyle` composable, instead of destructuring it. This maintains its reactivity. 

To solve the second issue, without disrupting the idea that the panels in the runner should have some minimum widths, required some changes. We now check if the window size is too small for the content, and if so, allow some scrolling so that users can always find the AUT and adjust the reporter size if they don't have much room. 

There were some side effects of allowing overflow in the resizable panel system, so there are some CSS class changes to keep the visual appearance of everything correct, and also some general removal of unused code.

Reacting the the window size and updating the minimum allowed widths when the window is small meant adding some new values to the `runner-constants` file.

Some tests have been added that open and close the specs list and reduce the viewport size to test scrolling. In the absence of visual regression testing for this part of the app, I wanted to add a visibility-based check, and did create something that put elements in all corners of the iframe (example screenshot below). But visibility checks appeared to run _in the context of the iframe_ so would pass even if the iframe was outside the parent viewport.

Screen Shot 2022-04-27 at 6 53 32 AM

That said, doing that exercise and some other testing persuaded me that validating the displayed scale % was failing in all the places I would have wanted the visibility checks to fail, so the % values should be helpful going forward. Tests have been updated to include more opening and closing the specs lists, which would have caught the regression.

The user can now recover better in very small windows, or at least figure out what's going on with the help of scrolling. The AUT is scaled correctly in all cases I could find.

### How to test

Open any spec and try any combination of dragging, opening, closing, resizing, refreshing, etc that you can think of that could put the scaling or the UI in to a bad state.

Very open to feedback on the readability of the calculations to somebody with fresh eyes. Is is easy to follow? Would you be comfortable making changes to this?

### PR Tasks


packages/app/cypress/e2e/cypress-in-cypress.cy.ts
```diff
@@ -118,6 +118,12 @@ describe('Cypress in Cypress', { viewportWidth: 1500, defaultCommandTimeout: 100
       cy.get(`[data-cy="reporter-panel"]`).invoke('outerWidth').should('eq', 800)
       cy.percySnapshot('initial state')
 
+      cy.contains('[aria-controls=reporter-inline-specs-list]', 'Specs')
+      .click({ force: true })
```
marktnoonan, line 124: I'm not delighted about using`force: true`, but since the reporter itself is not the subject of the test and the element seems to like to detach from the DOM, this cuts down on flake. I see in some other tests we use the keyboard shortcut instead, presumably for the same reason. Would be good to address this better.

packages/app/src/runner/SpecRunnerHeaderRunMode.vue
```diff
@@ -57,16 +56,10 @@
 <script lang="ts" setup>
 import { computed } from 'vue'
 import { useAutStore } from '../store'
-import type { EventManager } from './event-manager'
 import SpecRunnerDropdown from './SpecRunnerDropdown.vue'
 import { allBrowsersIcons } from '@packages/frontend-shared/src/assets/browserLogos'
 import { useAutHeader } from './useAutHeader'
 
-const props = defineProps<{
```
marktnoonan, line 65: Event Manager was only needed for the Selector Playground, and since that doesn't appear in run mode it was fine to remove it from the run mode header.

Passing the width used to be important to manage the appearance of the header, but now that we have changed the offset and allow scrolling, we are also able to drop this explicit setting of width.

circle.yml
```diff
@@ -301,6 +301,9 @@ commands:
       - run:
           name: Install Node Modules
           command: |
+            # avoid installing Percy's Chromium every time we use @percy/cli
```
marktnoonan, line 304: This is already merged in develop (https://github.com/cypress-io/cypress/pull/21218), but adding it here since the develop -> 10.0-release merge is a bit delayed

packages/app/cypress/e2e/cypress-in-cypress.cy.ts
tbiethman: I wonder if waiting for the inner spec run to complete before we perform our assertions could fix some of that flake... something like:

```js
  cy.findByLabelText('Stats').within(() => {
    cy.get('.passed .num', { timeout: 10000 }).should('have.text', '2')
  })
```

packages/app/src/runner/SpecRunnerOpenMode.vue
```diff
@@ -13,11 +13,14 @@
     />
     <ResizablePanels
       v-else
-      :offset-left="64"
-      :max-total-width="windowWidth"
+      class="w-[calc(100vw-64px)]"
```
tbiethman, line 16: Can we interpolate the collapsedNavBarWidth value here?

packages/app/src/runner/SpecRunnerOpenMode.vue
```diff
@@ -54,19 +57,18 @@
...
             :gql="props.gql.currentProject"
             :event-manager="eventManager"
             :get-aut-iframe="getAutIframeModel"
-            :width="width"
           />
         </HideDuringScreenshotOrRunMode>
 
         <RemoveClassesDuringScreenshotting
-          class="h-full bg-gray-100 p-16px"
+          class="bg-gray-100 h-0 p-16px"
```
tbiethman, line 71: Doesn't seem like we need the background color here at this point. Do we need to set a height with the vertical overflow disabled now?

tbiethman (approved):
I have just a few minor comments. My testing has shown the AUT scaling correctly, specs list open or closed. And the responsive improvements to the runner at small widths are a big improvement!

packages/app/src/runner/SpecRunnerOpenMode.vue
marktnoonan: Setting an explicit height (even 0) prevents scrolling in a way that I don't think users are actually able to do, but that I saw in Cypress-in-Cypress tests, and was causing a failure. It seems possible that some users might experience this as well (say, with assistive tech?), so I'm inclined to leave it in. A more general fix for this probably involves touching the `body` element based on what route the user is on, and seems to go beyond the scope of this PR.

You're right the gray is redundant though, since it's now handled in the panel itself, which was otherwise leaving a weird white area at the bottom some times after the overflow changes. So I pulled out the grays.

packages/app/src/runner/SpecRunnerOpenMode.vue
marktnoonan: yes, and switching this to a plain old style attribute turned out to be the best way, we could sorta get Windi to do this by creating a CSS variable to pass to it, but the indirection gets very hard to keep up with at that point.

packages/app/src/runner/screenshot/RemoveClassesDuringScreenshotting.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <div :class="classes">
+  <div
+    :class="classes"
```
lmiller1990, line 3: Instead of passing `class` and `style` manually should we just remove `inheritAttrs: true` from here https://github.com/cypress-io/cypress/pull/21254/files#diff-f0a40aeadfe60360756bdb96e57193fb0aed8897794366a826c32c379cff4528R11-R13.

I believe the default is `true`, which passes all the attrs to the root element. docs: https://vuejs.org/api/options-misc.html#inheritattrs

packages/app/src/runner/useRunnerStyle.ts
```diff
@@ -3,39 +3,48 @@ import { computed, ref, watchEffect } from 'vue'
...
+  reporterWidth.value = initialReporterWidth
+  specListWidth.value = initialSpecsListWidth
 
   const containerWidth = computed(() => {
     const miscBorders = 4
-    let nonAutWidth = reporterWidth.value + (isSpecsListOpen ? specListWidth.value : 0) + (autMargin * 2) + miscBorders
+    const containerMinimum = 50
+    let nonAutWidth = reporterWidth.value + (runnerUIStore.isSpecsListOpen ? specListWidth.value : 0) + (autMargin * 2) + miscBorders
 
     if (window.__CYPRESS_MODE__ !== 'run') {
```
lmiller1990, line 38:
```suggestion
    if (window.__CYPRESS_MODE__ === 'open') {
```

lmiller1990 (commented):
Some minor comments. I gotta say there's a ton of complexity around the panel sizings, I hope we can extract it all out into a badass resizable library from this one day.

packages/app/src/runner/screenshot/RemoveClassesDuringScreenshotting.vue
marktnoonan: Turned out I could revert this change anyway, so I did, `style` here was added while exploring different ways of managing the sizes.

marktnoonan:
@tbiethman @lmiller1990 I've updated based on your comments.

I also did a round of manual testing on Windows and through Firefox. Everything is good except a small issue in FF in some situations, so I've avoided adding the scrolling part of this to Firefox and created https://cypress-io.atlassian.net/browse/UNIFY-1704 to follow up, so that we can keep moving.

lmiller1990:
Approved, ship it

PR #21255, opened 2022-04-29
feat: gracefully fail an error when mismatching Cypress version used in migration

lmiller1990:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1531

### User facing changelog

Gracefully fail with error when migrating a Cypress 9 project with Cypress 10 launched in global mode.

### Additional details

It's possible to launch Cypress 10 in global mode with the `--global` flag. If the user then attempts to migrate a Cypress 9 project using the globally launched Cypress 10, we will generate Cypress 10 APIs (`cypress.config.js`, etc). If they later try to launch their Cypress 9 project, using it's locally installed Cypress 9, it will fail, since it will be expecting things like `cypress.json`, not `cypress.config.js`.

We now catch this scenario and warn the user, asking them to install Cypress 10 in their migration project before continuing.

#### Testing

1. Create a new project 2. 
2. Install cypress 9 from npm3. 
3. `yarn cypress open` to create `cypress.json`4. 
4. Open global mode on this branch with `yarn gulp open --global`5. 
5. Attempt to migrate your project with 9.x6. 
6. Get this error. 

image

### PR Tasks


packages/data-context/src/actions/MigrationActions.ts
```diff
@@ -147,6 +168,13 @@ export class MigrationActions {
     })
   }
 
+  locallyInstalledCypressVersion (currentProject: string) {
+    const localCypressPkgJsonPath = resolveFrom(currentProject, path.join('cypress', 'package.json'))
+    const localCypressPkgJson = fs.readJsonSync(path.join(localCypressPkgJsonPath)) as { version: string }
```
ZachJW34, line 173: We should wrap this in a try catch as we won't display the proper error `MIGRATION_CYPRESS_NOT_FOUND` as it will throw when reading a non-existent file.

I tested this by clicking on a project that had no node_modules
Screen Shot 2022-04-29 at 10 01 10 AM

ZachJW34 (changes requested):
Code looks good but one small change for handling errors.

packages/data-context/src/actions/MigrationActions.ts
lmiller1990: https://github.com/cypress-io/cypress/pull/21255/commits/6c3aa9c787533dbf86853b58e91bb0120c5775e6

lmiller1990: Done!

ZachJW34 (approved):

packages/data-context/src/actions/MigrationActions.ts
```diff
@@ -28,6 +31,7 @@ import {
 } from '../sources/migration'
 import { makeCoreData } from '../data'
 import { LegacyPluginsIpc } from '../data/LegacyPluginsIpc'
+import resolveFrom from 'resolve-from'
```
flotwig, line 34: any reason we are using this and not `require.resolve`? https://github.com/sindresorhus/resolve-from/issues/8

we use `require.resolve` with `cwd` elsewhere to find the user's `typescript`: https://github.com/cypress-io/cypress/blob/6e03e020682e73e0e5480fe010fe818fce488205/npm/webpack-preprocessor/lib/typescript-overrides.ts#L20-L27

packages/errors/src/errors.ts
```diff
@@ -1491,6 +1491,25 @@ export const AllCypressErrors = {
...
+    `
+  },
+
+  MIGRATION_CYPRESS_NOT_FOUND: () => {
+    return errTemplate`
+      You are running Cypress version 10 in global mode, but we were unable to determine the version of Cypress the project you are attempting to migrate is using.
+
+      Ensure the project you are migrating has Cypress version 10 installed.
+
+      https://on.cypress.io/migration-guide
```
flotwig, line 1510: I felt like this wasn't super clear, I think because there are only 2 cases where we are "unable to determine the version":

1. if the `resolve-from` fails for some reason, which would be a bug
2. if `cypress/package.json` DOES exist but `version` is empty, which also seems like it could only be a bug

Since `!version` is always either a bug or a sign the project definitely doesn't have `cypress@10`, we should make the error message more direct:

```suggestion
      You are running Cypress 10 in global mode and attempting to migrate a project where an install of `cypress` cannot be found.

      Ensure that `cypress@10` is installed in the project you are attempting to migrate.

      https://on.cypress.io/migration-guide
```

flotwig (changes requested):

BlueWinds (approved):

packages/data-context/src/actions/MigrationActions.ts
lmiller1990: Hm, looks like `resolve-from` should be obsolete and we can remove it entirely from this code base. Good point.

lmiller1990:
Thanks for the review @flotwig, I'm OOO for 2 days but I will try and get this fixed up before then. 💯

packages/data-context/src/actions/MigrationActions.ts
lmiller1990: @flotwig `require.resolve` is a little different, it gets the actual path that will be require'd (eg `index.js`). In this case I want `package.json`, so `resolveFrom` is a little better, it'll find the folder, not a specific file.

marktnoonan:
I gave this a try and got the error as expected, nice! One thought about the wording of the error: it's not clear to me that the person is "trying to migrate a project" - all they've tried to do is open one. So the message might be more clear if it explains a user can't _open_ projects in Cy 10 that haven't yet been upgraded.

packages/data-context/src/actions/MigrationActions.ts
lmiller1990: Right 🤦 

I made the change here and also in another part of the code base and removed the `resolve-from` dependency from the monorepo. https://github.com/cypress-io/cypress/pull/21255/commits/56e615452957f81b0d879c772577fa680c367fb5

lmiller1990:
@marktnoonan I made the error a bit more clear by adding "open" as well as "migrate"": https://github.com/cypress-io/cypress/pull/21255/commits/3eb1199c9fa17f192d695eb5bea91e89f1bad9b3

marktnoonan:
@lmiller1990 just checking, did you intend to only update the wording in one of the two errors? MIGRATION_CYPRESS_NOT_FOUND got the new wording (including 'open') but MIGRATION_MISMATCHED_CYPRESS_VERSIONS didn't. Noticed this when somebody posted a screenshot of this error.

PR #21272, opened 2022-04-29
refactor: cleanup types & memoization for SpecsLists component

tgriesser:
### User facing changelog
N/A

### Additional details

Cleans up the memoization types & comparator in `SpecsList.vue` / `spec-utils`  so we can add more fields to the `fragment SpecsList`, as will be needed in the work @MuazOthman is doing

No user facing change, internal refactor only

### PR Tasks


packages/frontend-shared/src/utils/spec-utils.ts
```diff
@@ -118,35 +120,13 @@ export function makeFuzzyFoundSpec (spec: FoundSpec): FuzzyFoundSpec {
...
-  return specs.some((spec, idx) => spec.absolute !== oldSpecs[idx]?.absolute)
-}
-
 export function useCachedSpecs<S extends { absolute: string }> (
   specs: ComputedRef<Readonly<S[]>>,
-  compareFn?: SpecsComparator<S>,
 ): Ref<Readonly<S[]>> {
   const cachedSpecs: Ref<Readonly<S[]>> = ref([])
 
   watch(specs, (currentSpecs, prevSpecs = []) => {
```
elevatebart, line 128: It feels like `cachedSpecs` should become a computed property...
I might be mistaken.

packages/frontend-shared/src/utils/spec-utils.ts
```diff
@@ -118,35 +120,13 @@ export function makeFuzzyFoundSpec (spec: FoundSpec): FuzzyFoundSpec {
...
-    const comparer = compareFn?.(currentSpecs, prevSpecs) || defaultCompareFn(currentSpecs, prevSpecs)
-
-    const specsAreDifferent =
-        currentSpecs.length !== prevSpecs?.length || comparer
-
-    if (specsAreDifferent) {
+    if (!_.isEqual(currentSpecs, prevSpecs)) {
```
lmiller1990, line 129: Is this caching even necessary at all? Is it prevent some kind of performance degradation?

lmiller1990 (approved):
Code seems fine - it might have got lost in all the noise, but why do we even need to cache this? Is there some kind of costly re-render performed if we don't (and what is triggered said re-render?)

packages/frontend-shared/src/utils/spec-utils.ts
tgriesser: I'm not 100% sure, I think it has something to do with the search or the list toggling or something. IIRC tests do fail if it's removed. I can check again

marktnoonan: This was going to be my only comment: I think it's fine to merge as it, but it should also work fine if remove `useCachedSpecs` completely and just reference the specs out of GQL directly in the 2 spec list components.

If removing it causes any tests to fail I wouldn't worry about leaving it in for the time being, we could clean it up later. I did pull it down and try without this function and it seemed to work just fine.

Perf would seem to be impacted if the specs list is changing a lot in rapid succession, like maybe if you paste 1000 files in there or something, but otherwise the specs list itself doesn't change much, and it looks like this cache is kinda recreating Vue's built in caching for `computed`s anyway (only update the ref if the value changed). So I'm not sure we are protecting against any particular side effects here.

marktnoonan (approved):
Approved, seems fine to explore what Lachlan reported, to simplify further, but also no problem merging it as it is now.

PR #21279, opened 2022-04-30
fix: Allow submit button to be outside of the form for implicit submission

### LINKED ISSUE #21163: Implicit form submission should work when the submit button is not nested in the form, but is associated
### Current behavior

When there is a submit button outside of the form, but associated to the form by the `form` attribute

```ts
someInput.type('{enter}')
``` 
does not submit the form

I'm assuming this is because Cypress is searching for the buttons with 
```ts
form.find('input, button')
```
Instead of searching for every submit button/input and testing if it's nested or associated by the form attribute

My submit button is not a child of the form, but it looks like
```html
<form id="my-form-id" onSubmit={function(e) { console.log('submitted'); e.preventDefault() }}>
  <input type="text" />
</form>
<button form="my-form-id" type="submit">Save</button>
```

### Desired behavior

Cypress should implicitly submit the form on enter in a text field

### Test code to reproduce

```html
<form 
  id="my-form-id" 
  onSubmit={function(e) { console.log('submitted'); e.preventDefault() }}>
    <input type="text" />
</form>

<button 
  form="my-form-id" 
  type="submit">
    Save
</button>
```

### Cypress Version

9.5.3

### Other

_No response_

skvale:
- Closes #21163 

### User facing changelog
Submit buttons can be outside of the form for implicit submission

### Additional details
Implicit submit wasn't working when the submit button wasn't nested within the form

Implicit submit with the enter key works when the submit button is outside of the form, but associated with a form attribute

### PR Tasks

### DIFF
--- a/packages/driver/src/cy/commands/actions/type.ts
+++ b/packages/driver/src/cy/commands/actions/type.ts
@@ -173,7 +173,7 @@ export default function (Commands, Cypress, cy, state, config) {
     const win = state('window')
 
     const getDefaultButtons = (form) => {
-      const formId = form.attr('id')
+      const formId = CSS.escape(form.attr('id'))
       const nestedButtons = form.find('input, button')
 
       const possibleDefaultButtons: JQuery<any> = formId ? $dom.wrap(_.uniq([
--- a/packages/driver/src/cy/commands/actions/type.ts
+++ b/packages/driver/src/cy/commands/actions/type.ts
@@ -176,12 +176,12 @@ export default function (Commands, Cypress, cy, state, config) {
       const formId = form.attr('id')
       const nestedButtons = form.find('input, button')
 
-      const possibleDefaultButtons = formId ? $dom.wrap(_.uniq([
+      const possibleDefaultButtons: JQuery<any> = formId ? $dom.wrap(_.uniq([
         ...nestedButtons,
         ...cy.$$('body').find(`input[form='${formId}'], button[form='${formId}']`),
       ])) : nestedButtons
 
-      return $dom.wrap(possibleDefaultButtons).filter((__, el) => {
+      return possibleDefaultButtons.filter((__, el) => {
         const $el = $dom.wrap(el)
 
         return (
--- a/packages/driver/src/cy/commands/actions/type.ts
+++ b/packages/driver/src/cy/commands/actions/type.ts
@@ -173,7 +173,15 @@ export default function (Commands, Cypress, cy, state, config) {
     const win = state('window')
 
     const getDefaultButtons = (form) => {
-      return form.find('input, button').filter((__, el) => {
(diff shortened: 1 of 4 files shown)


jennifer-shehane:
@skvale Thanks for the contribution and for adding tests around the new behavior. Could you please sign our CLA? After that we’ll have someone from our team review as soon as possible.

packages/driver/src/cy/commands/actions/type.ts
```diff
@@ -173,7 +173,15 @@ export default function (Commands, Cypress, cy, state, config) {
     const win = state('window')
 
     const getDefaultButtons = (form) => {
-      return form.find('input, button').filter((__, el) => {
+      const formId = form.attr('id')
+      const nestedButtons = form.find('input, button')
+
+      const possibleDefaultButtons: JQuery<any> = formId ? $dom.wrap(_.uniq([
+        ...nestedButtons,
+        ...cy.$$('body').find(`input[form='${formId}'], button[form='${formId}']`),
```
skvale, line 181: I wasn't sure what `cy.$$` is here.  Maybe we're looking for `$dom('body')` instead?

flotwig: Here's the definition for `cy.$$`: https://github.com/cypress-io/cypress/blob/6e03e020682e73e0e5480fe010fe818fce488205/packages/driver/src/cypress/cy.ts#L349-L355

So it seems like it is in effect doing `$dom('body')`.

To decrease the statefulness of this code, the context can be set to `Node.ownerDocument`:

```suggestion
        ...$dom.query('body', form.attr('ownerDocument')).find(`input[form='${formId}'], button[form='${formId}']`),
```

packages/driver/src/cy/commands/actions/type.ts
```diff
@@ -173,7 +173,15 @@ export default function (Commands, Cypress, cy, state, config) {
     const win = state('window')
 
     const getDefaultButtons = (form) => {
-      return form.find('input, button').filter((__, el) => {
+      const formId = form.attr('id')
```
flotwig, line 176: I think we need to `CSS.escape` this to avoid issues with special characters breaking the selector:

```suggestion
      const formId = CSS.escape(form.attr('id'))
```

flotwig (changes requested):

packages/driver/src/cy/commands/actions/type.ts
skvale: Did you mean `form.prop('ownerDocument')` ?

flotwig: Yes, my bad.

flotwig (approved):

marktnoonan (approved):
Glad to see this :)

PR #21281, opened 2022-05-01
fix: moving away from runner and disconnecting websockets stops listening for changes to webpack build

ryanthemanuel:
- Closes UNIFY-1651

### User facing changelog

In component testing, when the user moves away from the spec runner (for example to the Runs page) and then goes back to the spec runner, reruns due to webpack dev server recompilations will now properly reset the state of the reporter.

### Additional details

The steps to recreate this issue were:

1. Launch the app and run a component test
2. Click the runs tab
3. Go back to the spec list tab
4. Run the component test from step 1 again
5. Put your computer to sleep
6. Wake the computer
7. Modify the test in step 1
8. The test re-runs but the spec totals don't reset

What ends up happening in this flow is we [set up the websocket](https://github.com/cypress-io/cypress/blob/10.0-release/packages/app/src/runner/index.ts#L38-L44) and [on connection](https://github.com/cypress-io/cypress/blob/10.0-release/packages/server/lib/socket-base.ts#L315-L325) we set up the server to join the `runner` room. When we [move away](https://github.com/cypress-io/cypress/blob/10.0-release/packages/app/src/runner/unifiedRunner.ts#L18) from the spec list tab, we [teardown](https://github.com/cypress-io/cypress/blob/10.0-release/packages/app/src/runner/index.ts#L182) the event manager and [turn off](https://github.com/cypress-io/cypress/blob/10.0-release/packages/app/src/runner/event-manager.ts#L590) all websocket events. When we return to the spec list tab, we [turn on](https://github.com/cypress-io/cypress/blob/10.0-release/packages/app/src/runner/event-manager.ts#L67) all websocket events again except for the initial connection event. When the computer goes to sleep, the websocket connection is killed, and when it is reestablished, the connection event doesn't fire again and thus we never join the `runner` room again. Thus, when we modify a test file and recompile the webpack bundle, the `runner` channel is [notified](https://github.com/cypress-io/cypress/blob/10.0-release/packages/server/lib/socket-ct.ts#L19-L21), but nothing is listening on the client. Thus, the [event manager](https://github.com/cypress-io/cypress/blob/10.0-release/packages/app/src/runner/event-manager.ts#L146) is not notified and the reporter is not [reset](https://github.com/cypress-io/cypress/blob/10.0-release/packages/app/src/runner/event-manager.ts#L615).

To fix this, we should ensure that we are setting up the `connect` listener along with the other listeners in `addGlobalListeners`. We only do this if that event is not already set up. This flow is broken only in component testing, but I added a corresponding test in e2e as well to ensure that flow stays stable as well.

In addition on this PR, we clean up some dead code that was discovered.

Before:

https://user-images.githubusercontent.com/4873279/166166747-61b349d7-637d-44e8-81ec-88849b36a01a.mov

After:

https://user-images.githubusercontent.com/4873279/166166828-6e0c7373-cec3-4ab5-828f-7691a18fbe37.mov

### PR Tasks

### DIFF
--- a/packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
+++ b/packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
@@ -156,4 +156,59 @@ describe('Cypress In Cypress CT', { viewportWidth: 1500, defaultCommandTimeout: 
     cy.get('#unified-runner').should('have.css', 'width', '333px')
     cy.get('#unified-runner').should('have.css', 'height', '333px')
   })
+
+  it('moves away from runner and back, disconnects websocket and reconnects it correctly', () => {
+    cy.openProject('cypress-in-cypress')
+    cy.startAppServer('component')
+
+    cy.visitApp()
+    cy.contains('TestComponent.spec').click()
+    cy.get('[data-model-state="passed"]').should('contain', 'renders the test component')
+    cy.get('.passed > .num').should('contain', 1)
+    cy.get('.failed > .num').should('contain', '--')
+
+    cy.get('[href="#/runs"]').click()
+    cy.get('[data-cy="app-header-bar"]').findByText('Runs').should('be.visible')
+
+    cy.get('[href="#/specs"]').click()
+    cy.get('[data-cy="app-header-bar"]').findByText('Specs').should('be.visible')
+
+    cy.contains('TestComponent.spec').click()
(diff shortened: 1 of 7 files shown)


brian-mann:
Tears of joy

lmiller1990 (approved):
Wow, great job tracking this down. I tested it out (including putting my computer to sleep) and it's working as expected.

tbiethman (approved):
Ran through repro steps, works as expected :+1: 👍

PR #21285, opened 2022-05-02
fix: cy.contains handles input[type=submit] correctly.

### LINKED ISSUE #21166: cy.contains does not find the input[type='submit'] by value when using RegExp
### Current behavior

```html
<form>
    <input type="submit" value="Save Changes" />
</form>
```
 
```js
cy.contains('Save Changes') // OK. yields input[type='submit'] element
cy.contains(/Save Changes/) // ERROR. Expected to find content: '/Save Changes/' but never did.
```

Fails with this error
image

### Desired behavior

Expect `cy.contains` with a **RegExp** works as with **String** when looking at `input[type="submit"]`

### Test code to reproduce

Repo with reproduction https://github.com/cyberxander90/cypress-test-tiny/tree/contains-problem

### Cypress Version

9.5.4

### Other

_No response_

sainthkh:
- Closes #21166

### User facing changelog

`cy.contains` handles `input[type=submit]` correctly.

### Additional details
- Why was this change necessary? => `cy.contains` cannot find the text of `input[type=submit]` elements. 

```html
<input type="submit" value="Text" />
```

```js
cy.contains(/Text/) // Before: fails. After: works.
```

### PR Tasks

### DIFF
--- a/packages/driver/src/dom/elements/find.ts
+++ b/packages/driver/src/dom/elements/find.ts
@@ -247,6 +247,10 @@ export const getContainsSelector = (text, filter = '', options: {
 
     // taken from jquery's normal contains method
     cyContainsSelector = function (elem) {
+      if (elem.type === 'submit' && elem.tagName === 'INPUT') {
+        return text.test(elem.value)
+      }
+
       const testText = normalizeWhitespaces(elem)
 
       return text.test(testText)
--- a/packages/driver/cypress/integration/commands/querying/querying_spec.js
+++ b/packages/driver/cypress/integration/commands/querying/querying_spec.js
@@ -1040,6 +1040,15 @@ describe('src/cy/commands/querying', () => {
       })
     })
 
+    // https://github.com/cypress-io/cypress/issues/21166
+    it('can find input type=submits by Regex', () => {
+      cy.contains(/input contains submit/).then(($el) => {
+        expect($el.length).to.eq(1)
+
+        expect($el).to.match('input[type=submit]')
+      })
+    })
+
     it('has an optional filter argument', () => {
       cy.contains('ul', 'li 0').then(($el) => {
         expect($el.length).to.eq(1)


packages/driver/src/dom/elements/find.ts
```diff
@@ -247,6 +247,10 @@ export const getContainsSelector = (text, filter = '', options: {
 
     // taken from jquery's normal contains method
     cyContainsSelector = function (elem) {
+      if (elem.type === 'submit' && elem.tagName === 'INPUT') {
+        return text.test(elem.value)
+      }
```
flotwig, line 252: How can we handle the default submit buttons like `<input type="submit">`?

The default label is up to the user agent. and can be locale-dependent: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit#omitting_the_value_attribute

And I have not been able to find a way to retrieve the default submit string... I've tried `.innerText`, I've tried `getComputedStyle`, I've tried `.value` and `.defaultValue`...

If we can't figure out a way to get the default `submit` value to work with `cy.contains`, then I think this change may be unwise, because it won't be immediately clear why `cy.contains` works with some submits and not others. In that case, we should probably recommend users to use `cy.get('input[type=submit][value=whatever]')` if they need to select a submit by value.

flotwig (changes requested):
Let's target this against 11.0 - while it's arguably a bug fix, I think the number of users who will have test behavior changes because of this will be significant enough for us to want to keep it in a breaking change.

packages/driver/src/dom/elements/find.ts
sainthkh: How about leaving it as a note in [`cy.contains` doc](https://docs.cypress.io/api/commands/contains#Notes)?

It's happening without changes in this PR. Example code below:

```html
<input id="aaa" type="submit" />
```

```js
it('t', () => {
  cy.visit('fixtures/a.html')

  cy.get('#aaa').should('have.value', 'Submit') // fail. value is ''.
})

it('t2', () => {
  cy.visit('fixtures/a.html')

  cy.contains('Submit') // fail
})
```

----

Maybe we can add the explanation like below:

### When `value` is omitted in `<input type=submit>`

The default label is used and can be locale-dependent when the `value` attribute is not specified in `<input type=submit>`. [Check this.](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit#omitting_the_value_attribute)

`value` attribute is an empty string and there is no programmatic way to retrieve the default label name set by the user agent. It can cause unexpected failure in your tests.

The solution is to:

* compare the empty string (e.g. `cy.get('#el').should('have.value', '')`
* set the `value` attribute (e.g. `<input type=submit value="Submit" />`)

flotwig: Ohhhhhhhhhhhhh, I see what you mean now. Thank you. This is just fixing the existing case for `RegExp`. In that case, it's of course just a `fix` and not breaking or a "new" concern.

In that case, I'll open a PR for the docs, this PR looks good to me

flotwig (approved):

packages/driver/src/dom/elements/find.ts
flotwig: https://github.com/cypress-io/cypress-documentation/pull/4608

rockhold: Sounds like we're good to remove the breaking change label then, @flotwig?

flotwig: @rockhold Oh yeah, thanks. Done.

ryanthemanuel (approved):

PR #21286, opened 2022-05-02
fix: remove unnecessary eventemitter2 type definitions from cy, Cypress

### LINKED ISSUE #20556: Why can I use cy.waitFor()?

sainthkh:
- Closes #20556

### User facing changelog

Remove unnecessary `eventemitter2` type definitions from `cy` and `Cypress`.

### Additional details
- Why was this change necessary? => Remove undocumented `eventemitter2` definitions from `cy` and `Cypress`. Especially, `cy.waitFor` is really confusing with `cy.wait` for beginners.
- Any implementation details to explain? => Created a new interface and cherry-picked the definitions. 

```ts
cy.waitFor() // Before: type passes. After: Type fails.
```

### PR Tasks


cli/types/tests/actions.ts
```diff
@@ -87,3 +87,17 @@ namespace CypressActionCommandOptionTests {
   cy.get('el').click({scrollBehavior: false})
   cy.get('el').click({scrollBehavior: true}) // $ExpectError
 }
+
+namespace CyEventEmitterTests {
```
tbiethman, line 91: Can we put a comment here (or alongside the types themselves) explaining why we'd expect errors for `waitFor`/`prependListener`? Without context, these assertions seem a little random.

cli/types/cypress-eventemitter.d.ts
```diff
@@ -1,7 +1,23 @@
...
+  (event: event | eventNS, listener: ListenerFn, options?: boolean|OnOptions): this|Listener
+}
+
+interface CyEventEmitter {
+  once: CyActions
+  on: CyActions
+  off: CyActions
+  emit: EventEmitter2['emit']
+  removeListener: EventEmitter2['removeListener']
+  removeAllListeners: EventEmitter2['removeAllListeners']
```
tbiethman, line 20: Our docs say that Cypress/cy are "standard Node event emitters": https://docs.cypress.io/api/events/catalog-of-events#Binding-to-Events

I'm wondering if we need to either:
* Change the wording there if we only intend to support a subset of the API
* Add more functions here to align with what [Node's event emitter](https://nodejs.org/dist/latest-v16.x/docs/api/events.html) provides, like `addListener`. The other missing APIs are probably less frequently used, but it's hard to say.

cli/types/tests/actions.ts
sainthkh: I wrote the reason like below:

```js
// `waitFor` doesn't exist in Node EventEmitter 
// and it confuses the users with `cy.wait`
```

cli/types/cypress-eventemitter.d.ts
sainthkh: I decided to change the type to `Omit<EventEmitter2, 'waitFor'>`.

Because other functions are harmless, they're really unlikely to be used in the real world cases, though.

ryanthemanuel (approved):

PR #21293, opened 2022-05-02
feat: hide matches from spec creation

estrada9166:
- Closes  [UNIFY-1629](https://cypress-io.atlassian.net/jira/software/projects/UNIFY/boards/20?assignee=5ea883deddd45f0b94f5d98d&selectedIssue=UNIFY-1629)

### User facing changelog

Hide the matches count on the spec creation modal

### Before
image

### After
image
image

### PR Tasks


ryanjwilke:
@ZachJW34 To be honest, this isn't something we had built into the designs to my knowledge so showing the `specPattern` component makes the UX harder to understand from my perspective. I would much rather see that we only bring this up in the event that it **doesn't** match the specPattern. Does that make sense?

ZachJW34:
@ryanjwilke checked out the designs, looks good to me. Screenshots look good @estrada9166, I'll pull down today and test!

packages/app/src/components/FileMatchIndicator.vue
```diff
@@ -3,9 +3,27 @@
...
 <script lang="ts" setup>
+import { computed } from 'vue'
+
+const props = withDefaults(defineProps<{
+  variant?: 'default' | 'empty'
+}>(), { variant: 'default' })
+
+const colorClasses = {
+  default: 'bg-jade-100 text-jade-600',
+  empty: 'bg-purple-100 text-purple-600',
```
ZachJW34, line 23: We should rename this prop, `empty` is confusing. What do you think about `info`?

ZachJW34 (approved):
The `FileMatchIndicator` is a weird name for this component now that we are reusing it, but we already have a `Badge` component so not sure what to rename it. Code looks good though so I'm going to ✅

packages/app/src/components/FileMatchIndicator.vue
estrada9166: Sounds great! gonna rename it!

marktnoonan (approved):
Works well! I think this state looks a little weird with the extra whitespace, but not something to hold up the PR over, we can come back to that at any point.

Screen Shot 2022-05-05 at 9 10 04 AM

ryanjwilke:
@marktnoonan That state does match the Figma designs; so I think it's good for now. There's a reason we added a little extra room here, but we can talk through all that via Slack if you're curious. Should be easy to change in the future if needed.

PR #21302, opened 2022-05-03
chore: upgrade urql dependencies

tgriesser:
### User facing changelog

N/A

### Additional details

Upgrading some dependencies & fixing types, to split out from another PR which requires these changes.

### PR Tasks


estrada9166 (approved):

packages/frontend-shared/cypress/support/mock-graphql/mountFragment.ts
```diff
@@ -9,22 +9,12 @@ import type { TypedDocumentNode } from '@urql/vue'
...
-/**
- * This variable is mimicing ipc provided by electron.
- * It has to be loaded run before initializing GraphQL
- * because graphql uses it.
- */
-
-(window as any).ipc = {
-  on: () => {},
-  send: () => {},
-}
```
flotwig, line 27: where'd this go?

packages/frontend-shared/cypress/support/mock-graphql/mountFragment.ts
```diff
@@ -217,14 +206,14 @@ declare global {
...
-      stubMutationResolver<Result, Variables, T extends TypedDocumentNode<Result, Variables>>(
+      stubMutationResolver<T extends TypedDocumentNode<any, any>>(
         document: T,
         resolver: MutationResolver<T>
       ): Cypress.Chainable<ClientTestContext>
       /**
        * Mount helper for a component with a GraphQL fragment, as a list
        */
-      mountFragmentList<Result, Variables, T extends TypedDocumentNode<Result, Variables>>(
+      mountFragmentList<T extends TypedDocumentNode<any, any>>(
```
flotwig, line 216: at first i thought this weakened types, but i guess this doesn't weaken types because `Result` and `Variables` are already `any` (and unused)? is that correct?

packages/frontend-shared/cypress/support/mock-graphql/mountFragment.ts
tgriesser: It's been gone in the 10.0 branch since like August, got replaced with the local express server

packages/frontend-shared/cypress/support/mock-graphql/mountFragment.ts
tgriesser: Right, yeah `T` extends that but the concrete values are still kept, there's new helpers that we use to extract the types from that (`ResultOf, VariablesOf`). The generics removed here previously was just another way to do the same but isn't quite compatible w/ the new helper approach.

PR #21305, opened 2022-05-03
chore: Distribute tests to desktop-gui containers. Make `desktop-gui` tests faster!

### LINKED ISSUE #21318: CI: Builds from external contributors duplicate tests run across mutliple machines
### Current behavior

When a build is created by an external contributor, the tests are not split up across the machines due to them not having access to the record keys. We still parallelize the job (up to 7 machines) which means we are using up CI time and increasing the likelihood that the job fails due to flake.

### Desired behavior

We should not be parallelizing jobs if the tests cannot be split up, or we should use our own logic for splitting the tests across machines

### Test code to reproduce

Check out a build from an external contributor e.g. https://app.circleci.com/pipelines/github/cypress-io/cypress/37271/workflows/1ad10bf4-86c3-4bff-bad8-044b3952ad4b/jobs/1499238/parallel-runs/5?filterBy=ALL

### Cypress Version

9.6.0

### Other

_No response_

sainthkh:
- Related with #21304
- Closes #21318

### User facing changelog

N/A. It's an internal CI change. 

#### Why was this change necessary? 

For some reason, a test in `settings_spec.js` of `desktop-gui` CI failed on all recent PRs of mine (#21286, #21285, #21197, #21283). So, I tested it with #21304 that it happens with an empty commit. 

I investigated this failure and learned that it's a flaky test and run 7 times ([5 passed, 2 failed](https://app.circleci.com/pipelines/github/cypress-io/cypress/37258/workflows/fa6c4589-3ccc-4dc6-9761-56419cd602a5/jobs/1498481/parallel-runs/0?filterBy=FAILED)). Because each container of desktop-gui is running all of the desktop-gui tests. 

It was the reason why `desktop-gui` is always super late to finish.

And it can cause unwanted flakiness.

#### What is affected by this change?

N/A

#### Any implementation details to explain?

* I distributed tests to the containers. Currently, each container runs 3 test suites. 
* I used the code at driver tests.
* It's much faster to run CI. `desktop-gui` finishes before others. 

N/A.

### PR Tasks

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -468,28 +468,13 @@ commands:
       - restore_cached_workspace
       - run:
           command: |
-            if [[ -v MAIN_RECORD_KEY ]]; then
-              # internal PR
-
-              cmd=$([[ <<parameters.percy>> == 'true' ]] && echo 'yarn percy exec --parallel -- --') || true
-              CYPRESS_KONFIG_ENV=production \
-              CYPRESS_RECORD_KEY=$MAIN_RECORD_KEY \
-              PERCY_PARALLEL_NONCE=$CIRCLE_SHA1 \
-              PERCY_ENABLE=${PERCY_TOKEN:-0} \
-              PERCY_PARALLEL_TOTAL=-1 \
-              $cmd yarn workspace @packages/runner cypress:run --record --parallel --group runner-integration-<<parameters.browser>> --browser <<parameters.browser>>
-            else
-              # external PR
-              TESTFILES=$(circleci tests glob "cypress/integration/**/*spec.*" | circleci tests split --total=$CIRCLE_NODE_TOTAL)
-              echo "Test files for this machine are $TESTFILES"
-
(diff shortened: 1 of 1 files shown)


flotwig (changes requested):
Hey @sainthkh, nice fix! We actually use Cypress parallelization, but that won't work for contributors since there is no MAIN_RECORD_KEY, hence the problem.

Can you update your PR to use the approach described in https://github.com/cypress-io/cypress/issues/21318 please? i.e. use `cypress run --record --parallel` if `MAIN_RECORD_KEY` is available, and `circleci tests split` if not.

sainthkh:
@flotwig I checked the CI and it seems that `desktop-gui-integration` and `runner-integration` are using multiple containers. 

`reporter-integration`, `ui-components-integration` also use `MAIN_RECORD_KEY`, but they don't use multiple containers yet. So, I didn't change anything for them.

flotwig (approved):
nice, thank you for fixing this slowness @sainthkh

PR #21324, opened 2022-05-03
fix(app): make projectId pull from resolved config

flotwig:
- Closes 

### PR Tasks


packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -103,7 +103,7 @@ export class ProjectLifecycleManager {
 
   async getProjectId (): Promise<string | null> {
     try {
-      const contents = await this.getConfigFileContents()
+      const contents = await this.ctx.project.getConfig()
```
flotwig, line 106: @tgriesser is this something that https://github.com/cypress-io/tech-briefs/pull/19/files will help with? Maybe there could be better guards around calling `this.getConfigFileContents` - outside of config file resolution, there shouldn't be any need for code to be reading the config file in isolation, which was the source of this bug.

tgriesser: Yeah I think it might help. I think there was an old `read` helper used by old integration tests that used `cypress.json` which read the config file in isolation, if we can rip that out we can make `getConfigFileContents ` private.

packages/app/cypress/e2e/settings.cy.ts
```diff
@@ -28,7 +28,7 @@ describe('App: Settings', () => {
   })
 
   describe('Cloud Settings', () => {
-    it('shows the projectId section when there is a projectId', () => {
+    it('shows the projectId section when there is a projectId and shows override from CLI', () => {
```
tgriesser, line 31: Any reason we didn't just add a new test vs change the existing one?

tgriesser (approved):

packages/app/cypress/e2e/settings.cy.ts
flotwig: I was hesitant to add a new test because they're pretty slow at startup, and this was half covered by the existing test.

davidmunechika (approved):

PR #21327, opened 2022-05-03
feat: restart browser/server on config change

estrada9166:
- Closes  [UNIFY-1696](https://cypress-io.atlassian.net/jira/software/projects/UNIFY/boards/20?assignee=5ea883deddd45f0b94f5d98d&selectedIssue=UNIFY-1696)

### User facing changelog

We want to restart the server or the browser if there's a change on the config file, with this change we are able to set on the config options if a change on it requires server restart or browser restart so we can apply those changes.

### PR Tasks


packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -240,12 +241,21 @@ export class ProjectLifecycleManager {
...
         if (!this.ctx.isRunMode) {
           if (!this._initializedProject) {
             this._initializedProject = await this.ctx.actions.project.initializeActiveProject({})
+          } else if (restartOnChange.server) {
+            this.ctx.project.setRelaunchBrowser(true)
+            this._initializedProject = await this.ctx.actions.project.initializeActiveProject({})
+          } else if (restartOnChange.browser) {
+            this.ctx.project.setRelaunchBrowser(true)
+            await this.ctx.actions.browser.closeBrowser()
+            await this.ctx.actions.browser.focusActiveBrowserWindow()
```
tgriesser, line 258: What happens if they haven't opened the browser yet, but they make a config change, does this end up launching the browser?

estrada9166: Good catch! I'm gonna update it to handle it

estrada9166: [2303fa5](https://github.com/cypress-io/cypress/pull/21327/commits/2303fa559703d5cbb54e8a2aacb0382f79f87abb)

ZachJW34 (commented):
Few small comments, looks good overall!

ZachJW34 (approved):
Paired and looks good!

flotwig (changes requested):
1. If I save a syntax error in my config with the browser open, I get a really confusing error and subsequent saves do not reload the browser. Is this intended, that auto-reloading breaks here and I have to click "Try Again"? image
2. After adding/removing `setupNodeEvents` and `userAgent` with the browser open, the browser reloads to a blank screen: 
image
3. It doesn't seem like `setupNodeEvents` causes the browser to restart, but it should, since it can affect browser launch args.

packages/config/src/options.ts
```diff
@@ -167,6 +171,7 @@ const resolvedOptions: Array<ResolvedConfigOption> = [
     validation: validate.isString,
     isFolder: true,
     canUpdateDuringTestTime: false,
+    requireRestartOnChange: 'server',
```
flotwig, line 174: this one should relaunch the browser since it is set at browser launch

packages/config/src/options.ts
```diff
@@ -191,12 +196,14 @@ const resolvedOptions: Array<ResolvedConfigOption> = [
     validation: validate.isBoolean,
     isExperimental: true,
     canUpdateDuringTestTime: false,
+    requireRestartOnChange: 'browser',
```
flotwig, line 199: this shouldn't require a full browser restart, since `window.fetch` gets set in the driver: https://github.com/cypress-io/cypress/blob/b0c8db34348be31f32795f3f67c3408403d59ace/packages/driver/src/cypress/cy.ts#L1024-L1030

estrada9166:
> If I save a syntax error in my config with the browser open, I get a really confusing error and subsequent saves do not reload the browser. Is this intended, that auto-reloading breaks here and I have to click "Try Again"?

I didn't add changes for this, because the current implementation for all the errors required to click try again, it happens as well from the first time we try to read the config

> It doesn't seem like setupNodeEvents causes the browser to restart, but it should, since it can affect browser launch args.

It was updated to restart the browser if the task `before:browser:launch` is registered and there's any change on the config, unless there's a change that require server restart

> After adding/removing setupNodeEvents and userAgent with the browser open, the browser reloads to a blank screen

Was there any error on the terminal or the console? I have not been able to reproduce it

tgriesser (approved):

flotwig (approved):

Issue #7020, opened 2020-04-15
Launch Hanging on Ubuntu when `snap` version of Chromium is installed


bhayward93:
### Current behavior:

`cypress open` does not open Cypress.
`cypress run` just seems to hang.

Logs:
https://pastebin.com/ULZwQEj8

Snapshot of behavior and logs:
![https://files.gitter.im/cypress-io/cypress/g4g2/image.png](https://files.gitter.im/cypress-io/cypress/g4g2/image.png)

System:
![image](https://user-images.githubusercontent.com/18596813/79396536-0e3c7b00-7f74-11ea-8620-5ef578229a6e.png)

### Desired behavior:

Should be able to run tests.

### Test code to reproduce

As this is happening prior to Cypress running I don't think its anything I can give reproduction steps for.

`DEBUG=* $(npm bin)/cypress open`
`DEBUG=* $(npm bin)/cypress run`

Are the commands I am running.

### Versions

v4.3.0 and v4.4.0

DieterE:
Same problem here. Happened after Ubuntu wanted to update so perhaps the list of last updated packages can be helpful:
```
cat /var/log/dpkg.log | grep " installed" | grep 2020-04-15
2020-04-15 18:32:56 status installed libssh-gcrypt-4:amd64 0.9.0-1ubuntu1.4
2020-04-15 18:32:57 status installed libssh-4:amd64 0.9.0-1ubuntu1.4
2020-04-15 18:32:57 status installed git-man:all 1:2.20.1-2ubuntu1.19.10.2
2020-04-15 18:32:57 status installed containerd:amd64 1.3.3-0ubuntu1~19.10.2
2020-04-15 18:32:58 status installed git:amd64 1:2.20.1-2ubuntu1.19.10.2
2020-04-15 18:33:03 status installed man-db:amd64 2.8.7-3
2020-04-15 18:33:03 status installed libc-bin:amd64 2.30-0ubuntu2.1
2020-04-15 18:55:28 status installed libxapian30:amd64 1.4.12-1
2020-04-15 18:55:28 status installed libept1.5.90:amd64 1.1+nmu3ubuntu1
2020-04-15 18:55:29 status installed synaptic:amd64 0.84.6ubuntu3
2020-04-15 18:55:29 status installed mime-support:all 3.63ubuntu1
2020-04-15 18:55:29 status installed hicolor-icon-theme:all 0.17-2
[...]

PR #21328, opened 2022-05-03
fix(launcher): support Firefox as a snap

### LINKED ISSUE #19793: Can’t use default Firefox on latest Ubuntu
### Current behavior

When trying to use Firefox from Cypress on a fresh install of Ubuntu 21.10, it just hangs with
“Firefox is already running, but not responding.”

Screenshot

### Desired behavior

Cypress should open Firefox.

### Test code to reproduce

`cypress open`

### Cypress Version

9.3.1

### Other

Ubuntu 21.10 has [moved Firefox](https://discourse.ubuntu.com/t/impish-indri-release-notes/21951) to a [snap](https://snapcraft.io/about).

The snap is configured to confine Firefox using the [home interface](https://snapcraft.io/docs/home-interface), which allows “access to non-hidden files.” The firefox profile Cypress creates at `~/.config/Cypress/cy/production/browsers/firefox-stable/interactive` is considered hidden because of the `.` in `.config`. Firefox can’t access the profile folder and gets confused and says it is already running.

A similar issue was already fixed for Chromium in #7039.

As a temporary workaround, if I run, `XDG_CONFIG_HOME=$HOME/cypress cypress open` so that Cypress creates and uses a firefox profile in a non-hidden directory, then Firefox opens and cypress tests run successfully.

flotwig:
- Closes #19793
- Unblocks UNIFY-1607

### User facing changelog

- Fixed an issue causing Linux users with Firefox installed via Snapcraft to encounter errors on launch.

### Additional details

- If we detect FF is running in a snap, we change the profilePath to the sandboxed one, like we did for Cr: https://github.com/cypress-io/cypress/issues/7020
- ~~Had to change `@types/node` to the correct version to use the `fs` features.~~
	- Gave up on this, too many failures. Created an issue: https://github.com/cypress-io/cypress/issues/21329

### PR Tasks

### DIFF
--- a/packages/launcher/lib/linux/index.ts
+++ b/packages/launcher/lib/linux/index.ts
@@ -3,8 +3,28 @@ import type { FoundBrowser, Browser, PathData } from '../types'
 import { notInstalledErr } from '../errors'
 import { utils } from '../utils'
 import os from 'os'
+import { promises as fs } from 'fs'
 import path from 'path'
 import Bluebird from 'bluebird'
+import which from 'which'
+
+async function isFirefoxSnap (binary: string): Promise<boolean> {
+  try {
+    const binaryPath = await which(binary)
+
+    // read the first 16kb, don't read the entire file into memory in case it is a binary
+    const fd = await fs.open(binaryPath, 'r')
+    const { buffer, bytesRead } = await fd.read({ length: 16384 })
+
+    fd.close()
+
+    return buffer.slice(0, bytesRead).toString('utf8').includes('exec /snap/bin/firefox')
+  } catch (err) {
+    log('failed to check if Firefox is a snap, assuming it isn\'t %o', { err, binary })
+
+    return false
+  }
+}
 
 function getLinuxBrowser (
   name: string,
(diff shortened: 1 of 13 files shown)


packages/launcher/lib/linux/index.ts
```diff
@@ -3,8 +3,33 @@ import type { FoundBrowser, Browser, PathData } from '../types'
...
+
+      // read the first 16kb, don't read the entire file into memory in case it is a binary
+      const fd = await fs.open(binaryPath, 'r')
+      // @ts-ignore - needs @types/node at least 16
+      // https://github.com/cypress-io/cypress/issues/21329
+      const { buffer, bytesRead } = await fd.read<Buffer>({ length: 16384 })
+
+      await fd.close()
+
+      return buffer.slice(0, bytesRead).toString('utf8').includes('exec /snap/bin/firefox')
```
BlueWinds, line 24: Today I learned about reading only the opening bytes of a file! This seemed odd (particularly the slice), so I poked around the docs and command line, and lo, you have done it correctly. :)

packages/launcher/lib/linux/index.ts
```diff
@@ -3,8 +3,33 @@ import type { FoundBrowser, Browser, PathData } from '../types'
...
 import { utils } from '../utils'
 import os from 'os'
+import { promises as fs } from 'fs'
 import path from 'path'
 import Bluebird from 'bluebird'
+import which from 'which'
+
+async function isFirefoxSnap (binary: string): Promise<boolean> {
+  try {
+    return await Bluebird.resolve((async () => {
```
BlueWinds, line 13: I'd like to avoid Bluebird promises in newly written code when possible. Especially since we're already inside an `async` function, do we need `Bluebird.resolve` here?

BlueWinds (approved):

packages/launcher/lib/linux/index.ts
flotwig: I'm using Bluebird.resolve to get access to `.timeout`. I could `Promise.race/new Promise/setTimeout` but I figure that unless we are planning on moving all the code in the code base to not use Bluebird, might as well use the utilities we have. I also generally agree that BB should be avoided in new code.

BlueWinds: That is something I would like to do at some point, remove bluebird from our existing code. But adding timeout utility functions is beyond the scope of this PR, fair enough.

tbiethman (approved):
Verified on my Ubuntu VM that both snap/non-snap firefox installations are launched by Cypress

nicktelindertvmw:
The snap version of firefox and chromium are not detected by cypress for 12.17.2, testing on  Ubuntu 22.04 and newer

jennifer-shehane:
@nicktelindertvmw Please open a new issue describing the behavior and a reproducible example

PR #21334, opened 2022-05-04
fix: align 10.x selector playground tooltips with 9.x

ZachJW34:
- Closes [UNIFY-1582](https://cypress-io.atlassian.net/browse/UNIFY-1582)

### User facing changelog
Align selector playground tooltips with 9.x and fix selector validity state after selecting an element

### Additional details
The tooltips for the selector playground were not in parity with 9.x. This PR brings them into parity and adds tests/screenshots around them.

I also fixed a bug (one liner) where the playground would be marked as invalid even after selecting an element with the playground. The validity state was never being reset upon choosing an element. I've added a test for this case.

Before:

https://user-images.githubusercontent.com/25158820/166712498-df96d4e2-9d5b-4846-84b2-4adc74715107.mov

After

https://user-images.githubusercontent.com/25158820/166711976-9a3ab1a2-504d-46a2-b577-590d8042d617.mov

### PR Tasks


packages/app/src/runner/selector-playground/SelectorPlaygroundTooltip.vue
```diff
@@ -0,0 +1,57 @@
...
+const textToShow = ref(props.hoverText)
+
+function mouseOver () {
+  shown.value = true
+}
+
+async function mouseLeave () {
+  shown.value = false
+  await promiseTimeout(200)
+  textToShow.value = props.hoverText
```
tbiethman, line 50: Rather than reseting the value on leave and dealing with the animation timeout, can we just set the value to hoverText before we show it in `mouseOver`?

ZachJW34: This is so simple! Had to change `mouseover` to `mouseenter` since mouseover triggers on any movement inside the element.

packages/app/src/runner/selector-playground/SelectorPlaygroundTooltip.vue
```diff
@@ -0,0 +1,55 @@
...
+
+const shown = ref(false)
+const textToShow = ref(props.hoverText)
+
+function mouseEnter () {
+  textToShow.value = props.hoverText
+  shown.value = true
+}
+
+async function mouseLeave () {
```
tbiethman, line 47: No need for async here with your changes 😃

packages/app/src/runner/selector-playground/SelectorPlaygroundTooltip.vue
```diff
@@ -0,0 +1,55 @@
...
+  clickText?: string
+}>()
+
+const shown = ref(false)
+const textToShow = ref(props.hoverText)
+
+function mouseEnter () {
+  textToShow.value = props.hoverText
+  shown.value = true
+}
```
tbiethman, line 45: This might be out of scope, but is there a desire to also show the tooltip when the button is focused? The previous implementation would show Copied!/Printed! when the button was selected with keyboard keys, but we only show the tooltips with mouse interactions with these changes.

Seems like we could additionally listen for focusin/focusout events and mirror the logic, but I'm sure there are some hellish interaction combinations to work through.

ZachJW34: Slightly janky but I've added keyboard support to the tooltips in [aa185df](https://github.com/cypress-io/cypress/pull/21334/commits/aa185df92647260a331d523fe57b31cd4850629e)

ZachJW34: I had an updated version of floating-vue locally, forgot to push this up.

packages/app/src/navigation/SidebarNavigation.cy.tsx
```diff
@@ -28,11 +28,11 @@ describe('SidebarNavigation', () => {
 
   it('shows tooltips on hover', () => {
     mountComponent(false)
-    cy.get('[data-cy="sidebar-header"').realHover()
+    cy.get('[data-cy="sidebar-header"').trigger('mouseenter')
```
ZachJW34, line 31: Very flaky locally with `realHover`. Was debugging a failing test here and couldn't get past this step in open mode.

packages/app/src/navigation/SidebarNavigationRow.vue
```diff
@@ -5,7 +5,6 @@
     :distance="8"
   >
     <div
-      v-bind="$attrs"
```
ZachJW34, line 8: With the update to `floating-vue`, the Tooltip binds the attributes of it's parent by default. This was causing two element have the `data-e2e-href` attribute. This was the only attribute being passed so we can safely remove the `v-bind`.

packages/app/src/navigation/SidebarNavigation.cy.tsx
marktnoonan: I generally prefer mouseenter as well unless is doesn't work, good change.

packages/app/src/runner/selector-playground/SelectorPlayground.cy.tsx
```diff
@@ -110,42 +110,75 @@ describe('SelectorPlayground', () => {
...
-        Yielded: undefined, // stubbed dom does not actually return anything
+        Yielded: 'Nothing',
       })
     })
   })
+
+  it('shows tooltips when buttons are focused', () => {
+    mountSelectorPlayground()
+
+    cy.get('[data-cy="playground-toggle"]').focus()
```
marktnoonan, line 165: Could maybe clean this up with aliases for the `cy.get`s?

packages/app/src/runner/selector-playground/SelectorPlaygroundTooltip.vue
ZachJW34: Addressed in [aa185df](https://github.com/cypress-io/cypress/pull/21334/commits/aa185df92647260a331d523fe57b31cd4850629e)

packages/app/src/runner/selector-playground/SelectorPlayground.cy.tsx
```diff
@@ -110,42 +110,75 @@ describe('SelectorPlayground', () => {
...
       })
     })
-
-    cy.get('[data-cy="playground-print-tooltip"]').should('be.visible').contains('Printed to console')
   })
 
-  it('prints elements when selected elements found', () => {
-    const { autIframe } = mountSelectorPlayground()
-
+  it('prints nothing to console when no selected elements found', () => {
```
marktnoonan, line 148: This test seems to be checking the behavior for the invalid state, not the "no selected elements found" state - I'm surprised that `cy.type('.foo-bar')`, wonder if this is related to resetting validity. Or does this test have the wrong name?

marktnoonan (changes requested):
Works great, I just have a question about one of the tests.

packages/app/src/runner/selector-playground/SelectorPlayground.cy.tsx
marktnoonan: Zach and I talked offline a bit, the issue I'm seeing was not introduced in this PR, so I've created https://cypress-io.atlassian.net/browse/UNIFY-1727 to follow up.

marktnoonan (approved):

tbiethman (approved):

PR #21337, opened 2022-05-04
fix: clear remote data by marking all query data as stale on logout

tgriesser:
- Closes UNIFY-1652

### User facing changelog

When logging out, the "Record Key" section, and any other stale cloud data will no longer be visible.

### Additional details

When logging out, we want to clear out any remote data that might have been cached on the client. We can do this by invalidating the `Query` field for the cache using [`cache.invalidate`](https://formidable.com/open-source/urql/docs/api/graphcache/#invalidate), on logout which will result in refetching all data with the latest info from the electron server.

### PR Tasks

### DIFF
--- a/packages/app/cypress/e2e/settings.cy.ts
+++ b/packages/app/cypress/e2e/settings.cy.ts
@@ -103,6 +103,30 @@ describe('App: Settings', () => {
       cy.contains('Record Key').should('not.exist')
       cy.contains('Record Key')
     })
+
+    it('clears nested cloud data (record key) upon logging out', () => {
+      cy.startAppServer('e2e')
+      cy.loginUser()
+      cy.visitApp()
+      cy.withCtx((ctx, o) => {
+        o.sinon.spy(ctx.actions.auth, 'logout')
+      })
+
+      cy.get(`[href='#/settings']`).click()
+      cy.contains('Dashboard Settings').click()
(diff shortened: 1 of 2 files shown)


estrada9166 (approved):
Nice!!

packages/app/cypress/e2e/settings.cy.ts
```diff
@@ -103,6 +103,30 @@ describe('App: Settings', () => {
...
+      cy.loginUser()
+      cy.visitApp()
+      cy.withCtx((ctx, o) => {
+        o.sinon.spy(ctx.actions.auth, 'logout')
+      })
+
+      cy.get(`[href='#/settings']`).click()
+      cy.contains('Dashboard Settings').click()
+      cy.contains('Record Key').should('exist')
+      cy.get(`[href='#/runs']`).click()
```
tbiethman, line 118: Do we need to navigate away from the Settings page and log out there so that data is queried again when we return after invalidation? Or could we just log out from the Settings page and see it update immediately after invalidation?

tgriesser: Currently before this change, if you're on the settings page it'll send a signal to refetch the mounted queries, the bug is when you're not on the settings page, the queries don't get refreshed.

BlueWinds (approved):

PR #21340, opened 2022-05-04
chore: cleanup dead code

tgriesser:
- Closes N/A

### User facing changelog

N/A

### Additional details
Removing some dead code that I came across while working on other tickets

### PR Tasks


packages/data-context/src/sources/HtmlDataSource.ts
```diff
@@ -35,7 +35,7 @@ export class HtmlDataSource {
         )
       } catch (e) {
         err = e
-        await this.ctx.util.delayMs(1000)
+        await new Promise((resolve) => setTimeout(resolve, 1000))
```
flotwig, line 38: might be worth just using `Bluebird.timeout` here and elsewhere in this PR... so when we strip `Bluebird` out these timeouts are more easily discovered and replaced with a shared helper. Blue and I recently discussed this here: https://github.com/cypress-io/cypress/pull/21328#discussion_r865463268

flotwig (approved):
lovely lovely lovely

packages/data-context/src/sources/HtmlDataSource.ts
tgriesser: I don't think we use Bluebird at all in the `data-context` package currently

flotwig: :man_shrugging:

PR #21341, opened 2022-05-04
fix: add min-width and min-height to _x16 icons

estrada9166:
- Closes  [UNIFY-1661](https://cypress-io.atlassian.net/browse/UNIFY-1661)

### Additional details

Add default `min-width` and `min-height` to icons that are `x16`, so, when resizing panels the icons do not get smaller, this approach will add the default value preventing future misses and being consistent across the app

### PR Tasks


packages/frontend-shared/vite.config.ts
```diff
@@ -53,6 +53,11 @@ const makePlugins = (plugins) => {
         cy: FileSystemIconLoader(path.resolve(__dirname, './src/assets/icons')),
         ...plugins.iconsOptions?.customCollections,
       },
+      iconCustomizer (collection, icon, props) {
+        if (icon.includes('_x16')) {
```
marktnoonan, line 57: At some point we could do this across the board for `_x24`, `_x32` and all the others - perhaps by checking the `_x${whatever}` and using that value for the inline height and width.

marktnoonan (approved):
Works great. Percy diff was huge due to many small adjustments but all of them look fine and measuring some of the changes against figma, they are more correct. Removed @mapsandapps as a reviewer as I think they are out of town.

Non-blocking idea suggested, but I'm happy for this to merge as-is.

PR #21343, opened 2022-05-04
chore: Following up on mochaEvent test performance on Windows

tbiethman:
- Closes UNIFY-1579

### User facing changelog

N/A

### Additional details

UNIFY-1579 was logged to track down or otherwise determine the need for increased timeouts for the mochaEvent tests on Windows. More raw data can be found on the JIRA, but to summarize:

* Cypress execution times for similar test suites in 9.x and 10.0-release are comparable on Windows, with 10.0-release generally performing (slightly) better. This indicates that there is not a generalized performance regression with 10.0-release in windows.
* Cypress execution times for in 10.0-release for the `app` integration tests (e.g. cypress-in-cypress tests) are much longer in Windows than they are on linux, even with greater resources available on the Windows build box. Increasing the Windows resource size from `medium` to `large` decreases the delta but does not erase it. The mochaEvents specs are a subset of the `app` tests and are similarly impacted.

It’s also important to understand how these mochaEvent tests are structured and execute. Each test in these specs performs the following tasks:

1. The outer Cypress loads the inner Cypress app 
2. The inner Cypress loads the runner and parses the spec under test
3. The inner Cypress executes the spec to completion
4. The inner Cypress emits the appropriate lifecycle events with mocha snapshot data
5. The outer Cypress performs the snapshot diffing and calls `done()` when complete

As they are written, steps 2 through 5 above all must occur within a single command timeout for the test to succeed. Given Window's performance with cypress-in-cypress tests, it's not surprising to me that these tests sometimes exceed the default timeout duration.

I attempted remove the default timeout increase from the suite and target individual tests that were more prone to failure; however, further testing showed that any of these specs can time out. So I left the increases where they were and with the same duration, as it has proven to be adequate.

Changes made in this PR include:
* Updating the comments around the increased timeouts to more definitively state why they are present.
* Updating the mochaEvent tests to directly visit their spec vs. navigating through the UI. This does not have an impact on their timing out, but it does make them complete quite a bit faster across all jobs.

### PR Tasks


lmiller1990 (approved):

lmiller1990:
Thanks for the PR description, it is appreciated! I'm glad we don't have an overall regression w.r.t windows perf.

lmiller1990:
Looks like we have some real windows fails, but they were not introduced here, so I'll merge this up.

PR #21345, opened 2022-05-04
chore: remove unused parts of design-system, move SCSS to runner-ct

flotwig:
- Closes 

### User facing changelog

n/a

### Additional details

- In 10.0, `design-system` is not used, and it has a quirky build process that gets in the way: https://cypressio.slack.com/archives/C011BAPC614/p1651619895578469
- Guessing it will be replaced by https://github.com/cypress-io/cypress-design?
- There is only 1 usage, in `packages/runner-ct/src/variables.scss`
- This PR removes all non-`scss` files from `design-system` and moves the scss to `packages/runner-ct/src/legacy-styles`
    - Some scss could probably be deleted but I don't know the best way to audit this and am not sure how useful a deep cleanup would be right now with the state of flux of the design system.
    - Most of the other changes in this PR are fallout from previously-implicitly-hoisted modules no longer being installed after removing `design-system`.
- Unexpected bonus: this brings the number of packages installed during local development from ~34392 to ~29726 (`yarn list | wc -l`), a 13.5% reduction 
    - Brings module size on disk from 4.2 GB to 3.5 GB (-16.6%) according to `du -ch **/node_modules/`
- Also removes some now-unused patch files

### PR Tasks


npm/create-cypress-tests/package.json
```diff
@@ -27,14 +27,16 @@
...
+    "recast": "0.20.4",
+    "semver": "7.3.7"
   },
   "devDependencies": {
     "@types/babel__core": "^7.1.2",
     "@types/inquirer": "7.3.1",
     "@types/mock-fs": "4.10.0",
     "@types/node": "14.14.31",
     "@types/ora": "^3.2.0",
+    "@types/semver": "7.3.9",
```
flotwig, line 39: these types are used in this package but apparently were only available due to the miracle of hoisting, and by virtue of design-system including them

packages/app/package.json
```diff
@@ -101,7 +101,6 @@
         "lodash",
         "mobx",
         "nanoid",
-        "path",
```
flotwig, line 104: Something, somewhere, was causing this internalization to fail with an inscrutable error: https://app.circleci.com/pipelines/github/cypress-io/cypress/37390/workflows/48a17960-8622-45a4-a3bf-ff73f4e2b33d/jobs/1504577?invite=true#step-111-31

My hypothesis is that something that was hoisted by `npm/design-system/package.json` is no longer available, so this stopped working across the repo. I was unable to determine what that "something" was though, so I just removed this line from the `app`, `launchpad`, and `frontend-shared` `package.json`s.

This only had one side-effect of breaking `path` in the FE tests, which we only use in one place and was easily replaced by a 1 line function: 4357e8198bfeaf1f2ea3ff0879a735da81b50213

packages/frontend-shared/package.json
```diff
@@ -102,7 +102,6 @@
         "gravatar",
         "human-interval",
         "lodash",
-        "path",
```
flotwig, line 105: https://github.com/cypress-io/cypress/pull/21345/files#r866233298

packages/launchpad/package.json
```diff
@@ -107,7 +107,6 @@
         "lodash/clone",
         "markdown-it",
         "@toycode/markdown-it-class",
-        "path",
```
flotwig, line 110: https://github.com/cypress-io/cypress/pull/21345/files#r866233298

packages/frontend-shared/cypress/support/mock-graphql/stubgql-Mutation.ts
```diff
@@ -1,17 +1,21 @@
...
-import path from 'path'
 
 import type { MaybeResolver } from './clientTestUtils'
 import { createTestCurrentProject, createTestGlobalProject } from './stubgql-Project'
 
+// `path` is externalized, so we use a simple implementation of `path.basename` here
+function basename (path: string) {
+  return path.split(/[/\\]/).slice(-1)[0]
+}
+
```
flotwig, line 10: see https://github.com/cypress-io/cypress/pull/21345/files#r866233298

packages/runner-ct/src/variables.scss
```diff
@@ -1,5 +1,5 @@
-@forward '~@cypress/design-system';
-@use '~@cypress/design-system' as *;
+@forward './legacy-styles';
+@use './legacy-styles' as *;
```
emilyrohrbough, line 2: Are these actually legacy styles? or just shared styles between the app & dashboard?

emilyrohrbough (approved):

packages/runner-ct/src/variables.scss
flotwig: https://cypressio.slack.com/archives/C011BAPC614/p1651786758401009?thread_ts=1651619895.578469&cid=C011BAPC614

ZachJW34 (approved):
NICE!

PR #21349, opened 2022-05-05
feat: migration landing page

marktnoonan:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1501

### User facing changelog

This PR adds a landing page with the Cypress 10 video, a continue button, and release noes link, that will appear before the migration for projects that need to migrate.

### Additional details

A [PR was merged](https://github.com/cypress-io/cypress-services/pull/4517) to the services repo to provide the vimeo embed code via an on link. This ensures that changes in Vimeo's implementation can be accommodated even with already-published versions of Cypress, to avoid repeating an incident where Vimeo abruptly changed.

This PR:

- adds `videoEmbedJson` to `migration` in GraphQL
- adds a function to fetch the json in `MigrationDataSource`
- tweaks link-button styles so the hover state is the same no matter what element the button uses
- updates many tests around migration that need to click through this new landing page now
- adds some e2e and component test around the new feature
- updates Main.vue with logic for when to show the landing page - this `v-if` logic is getting a bit heavy, ideally the rendering would be managed by a router but that's out of scope for this PR.
- cleaned up some unused stuff in the assets folder
- added MigrationLandingPage.vue itself

New landing page added: 

Screen Shot 2022-05-17 at 11 47 59 AM

### How to test

Open any project that needs migration (for example [these](https://github.com/cypress-io/cypress-component-testing-apps)). The new welcome screen should be there.

### PR Tasks


packages/graphql/src/schemaTypes/objectTypes/gql-Migration.ts
```diff
@@ -227,6 +227,13 @@ export const Migration = objectType({
       },
     })
 
+    t.string('videoEmbedJson', {
+      description: 'stringified JSON from an onlink, containing markup for the migration landing page video embed',
+      resolve: (source, args, ctx) => {
```
tgriesser, line 232:
```suggestion
    t.string('videoEmbedHtml', {
      description: 'Markup for the migration landing page video embed',
      resolve: (source, args, ctx) => {
```

tgriesser (changes requested):
Was thinking it might be better to just respond with the HTML string as the field rather than the raw json encoded payload which the client then needs to know about the shape of, also changes to a try/catch to wrap the `await response.json()` call which could throw

packages/data-context/src/sources/MigrationDataSource.ts
```diff
@@ -96,6 +96,25 @@ export class MigrationDataSource {
...
+      // fail silently, no user-facing error is needed
+    })
+
+    if (!response) {
+      return null
+    }
+
+    const json = await response.text()
+
+    return json
```
tgriesser, line 115:
```suggestion
    const embedOnLink = `https://on.cypress.io/v10-video-embed/${versionData.current.version}`
    try {
      const response = await this.ctx.util.fetch(embedOnLink, { method: 'GET' })
      const { videoHtml } = await response.json()
      return videoHtml
    } catch {
      // fail silently, no user-facing error is needed
      return null
    }
```

packages/graphql/src/schemaTypes/objectTypes/gql-Migration.ts
marktnoonan: Yeah, the case for keeping it as JSON is pretty weak, just the expectation that one day we might add more items to the payload, I'm fine with switching it here.

packages/launchpad/src/migration/MigrationLandingPage.vue
```diff
@@ -0,0 +1,91 @@
...
+
+.video {
+  aspect-ratio: 15 / 9;
+
+  // since we support pre-aspect-ratio browsers,
+  // the classic padding hack is used here
+  // https://css-tricks.com/aspect-ratio-boxes/#aa-the-pseudo-element-tactic
+  // https://codepen.io/una/pen/BazyaOM
+
+  @supports not (aspect-ratio: 15 / 9) {
```
ZachJW34, line 77: This is only rendered in electron which we control, do we need this?

packages/launchpad/src/Main.vue
```diff
@@ -82,14 +92,17 @@ import Spinner from '@cy/components/Spinner.vue'
...
+import MigrationLandingPage from './migration/MigrationLandingPage.vue'
 import { useI18n } from '@cy/i18n'
 import { computed, ref } from 'vue'
 import LaunchpadHeader from './setup/LaunchpadHeader.vue'
 import OpenBrowser from './setup/OpenBrowser.vue'
+import { useOnline } from '@vueuse/core'
 
 const { t } = useI18n()
 const isTestingTypeModalOpen = ref(false)
+const wasLandingPageShown = ref(false)
```
ZachJW34, line 104: Should this be stored in the backend project state? It is desired that we always show this for every project that needs to be migrated?

packages/launchpad/src/Main.vue
```diff
@@ -3,7 +3,17 @@
     <HeaderBar
       class="w-full z-10 fixed"
     />
-    <div class="px-24px pt-86px">
+
+    <MigrationLandingPage
+      v-if="currentProject?.needsLegacyConfigMigration && !wasLandingPageShown && online && videoHtml"
```
ZachJW34, line 8: If the user has an error (like not having Cypress installed int he project), it will be shown after they continue pass this page. Not sure if it's a better experience throwing the error sooner, but it felt like "Ah yeah, Cypress 10 looks dope" then boom error.

ZachJW34 (commented):
Looks awesome @marktnoonan! Responsiveness is real nice.

Just a few minor comments

packages/launchpad/src/Main.vue
marktnoonan: In so far as this has been described, it's "before migration" so I think this is the best place to start. It will be easy to add conditions in the future if needed, in a follow up PR.

ZachJW34 (approved):

packages/launchpad/src/Main.vue
marktnoonan: I guess let's see how it feels in the final run through, we could tweak

marktnoonan:
@tgriesser updated with your changes

tbiethman (approved):
This looks great!

tgriesser (approved):

PR #21352, opened 2022-05-05
Unify 1316 finalize viewport dropdown in spec runner header

rachelruderman:
- Closes 

### PR Tasks


packages/app/src/runner/SpecRunnerHeaderOpenMode.vue
```diff
@@ -77,29 +77,40 @@
...
+
+            <i18n-t
+              tag="p"
+              scope="global"
+              keypath="runner.viewportTooltip.configText"
+              class="mb-14px"
+            >
+              <!-- disable rule to prevent trailing space from being added to <InlineCodeFragment/> -->
+              <!-- eslint-disable-next-line vue/singleline-html-element-content-newline -->
+              <InlineCodeFragment class="text-xs">cypress.config.js</InlineCodeFragment>
```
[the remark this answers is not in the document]
jennifer-shehane: Yes, this needs to be dynamic.

packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
```diff
@@ -29,7 +29,7 @@ describe('Cypress In Cypress CT', { viewportWidth: 1500, defaultCommandTimeout:
 
       snapshotAUTPanel('browsers open')
       cy.contains('Canary').should('be.hidden')
-      cy.contains('The viewport determines the width and height of your application. By default the viewport will be 500px by 500px for Component Testing unless specified by a cy.viewport command.')
+      cy.contains(cy.i18n.runner.viewportTooltip.infoText.replace('{0}', '500px').replace('{1}', '500px').replace('{2}', 'component'))
```
jennifer-shehane, line 32: There was never consensus among the engineering team to move forward with using cy.i18n replacement values for text in the tests. There were quite a few people opposed to this convention since readability is hindered.

jennifer-shehane (changes requested):

packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
rachelruderman: Oooh does this mean I should use a hardcoded string?

BlueWinds: There never was a consensus - IIRC we ended up with "as feels appropriate to you."

packages/app/src/runner/SpecRunnerHeaderOpenMode.vue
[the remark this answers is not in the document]
marktnoonan: There's a `hasTypescript` field on `currentProject` in GQL: https://github.com/cypress-io/cypress/blob/b65371a6a12aa18d9cd187b5bbf9b006a49cabc7/packages/graphql/src/schemaTypes/objectTypes/gql-CurrentProject.ts#L116 that could be pulled to check this.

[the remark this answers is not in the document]
BlueWinds: `OpenConfigFileInIDE.vue` has an example of pulling the config path. 

```
        <span
          class="text-purple-500 cursor-pointer"
        >
          {{ props.gql.configFile ?? 'cypress.config.js' }}
        </span>

...stuff...

gql`
fragment OpenConfigFileInIDE on CurrentProject {
  id
  configFile
  configFileAbsolutePath
}
`
```

BlueWinds:
Looks pretty good in general, just that one dynamic value (and I haven't looked at the failing tests).

packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
marktnoonan: It's a tradeoff, should the test "know" the correct text, or should the content json be the source of truth? Tyler called it "asserting if the pipes work" vs "asserting what goes through them" or something like that, which I liked.

It's more readable with plain strings for sure - especially this example, I would agree with keeping it as a plain string here.

I've kinda come around to thinking component tests should always refer to the content from the json, to confirm the component is correctly hooked up to its source of content, and E2E tests might benefit from using plain strings to confirm the user-facing text is what is expected. There are other considerations but this would at least get all the advantages of both approaches, and a clear boundary between them. We haven't revisited the discussion in a while though.

marktnoonan (changes requested):
Great to get this in and the functionality and tests look good! I have a few small comments on the appearance.

The line heights and padding are a little off compared to Figma, so this looks sorta cramped to me:
Screen Shot 2022-05-05 at 6 55 14 PM

Compared to Figma:

Screen Shot 2022-05-05 at 6 55 07 PM

I think there's also a small error in Figma itself, we seem to typically refer to commands with parens after, so `cy.something()` (eg, to pluck something random from docs) how we refer to `cy.clock()` [here](https://docs.cypress.io/api/commands/clock#Yields), so we should probably refer to is as `cy.viewport()` (open to correction if this is not a standardized thing?).

One last style thing that is a bit of a fluke: there is some global CSS from the reporter package causing the `<strong>` tag to not actually end up bold for me when I open this in the runner (you can see this in my first screenshot), so we might need an explicit font weight set on the pixel values in this text.

Screen Shot 2022-05-05 at 6 56 46 PM

rachelruderman:
@marktnoonan Oh man that was a good catch! For some reason I didn't even think about the padding 🤦‍♀️ I just put in some changes, here's a few caveats:
ui-updates

marktnoonan:
Nice! The line height makes sense, I didn't check it directly, what looked off to me was the vertical space between the two inline code blocks. Looking deeper, it turns out that tailwind's `text-xs` utility class doesn't just set text size, it also reduces line height, so the spacing at the bottom is actually off a bit. We could fix this by setting an explicit text size - though I want us to move away from that, it's consistent with what we do elsewhere at the moment.

The spacing and the bold text are the main differences between how this looks in CT and how it looks in the real header:

Screen Shot 2022-05-06 at 8 42 57 AM

Since the one on the right is what users will actually see, that's the one we have to dial in. 

This is probably the influence of global styles, that belong to the runner page, not being present in CT - long term we should either clean those up with better scoping, or make sure they are present in component tests for runner components so we can see where/why we need some extra overrides. Those styles themselves are still needed because they support the reporter, which isn't part of the Vue app.

BlueWinds (approved):

marktnoonan:
@emilyrohrbough I agree with you on the icon issue, it was captured in https://cypress-io.atlassian.net/browse/UNIFY-1464

emilyrohrbough:
@marktnoonan oh awesome! glad this wasn't lost.

marktnoonan (approved):
Looks great!

PR #21354, opened 2022-05-05
chore: fix flaky windows tests

ryanthemanuel:
### User facing changelog

n/a

### Additional details

There's a variety of fixes in here including:

* Creating a standard mechanism for waiting for cypress in cypress tests to complete. The goal here is to provide several checkpoints along the way rather than waiting for one long timeout when they are finished. The checkpoints include:
  * Ensuring the pass/fail stats are reset
  * Ensuring that the test loader has been removed
  * Ensuring that the refresh button gets re-added
* Ensuring that the test runner has loaded before we try and type 'f' to open up the inline specs list
* Handle a race condition where we try and rerun the current spec even though we are about to go back to the specs list with an error since the spec no longer exists
* Properly handle platform differences when we are verifying scaffolded files

n/a

### PR Tasks


packages/app/cypress/e2e/support/execute-spec.ts
```diff
@@ -0,0 +1,32 @@
...
+  // First ensure the test is loaded
+  cy.get('.passed > .num').should('contain', '--')
+  cy.get('.failed > .num').should('contain', '--')
+
+  // Then ensure the tests are running
+  cy.contains('Your tests are loading...').should('not.exist')
+
+  // Then ensure the tests have finished
+  cy.get('[aria-label="Rerun all tests"]', { timeout: 30000 }).should('exist')
+}
```
tbiethman, line 30: Very cool! I appreciate the increased stability this will bring to our cypress-in-cypress tests.

tbiethman (approved):
Thanks for jumping on these!

mjhenkes (approved):

tgriesser (approved):
Changes look fine to me, nice work on standardizing some of the cypress-in-cypress behavior.

Do we even need the `isLoadingNodeEvents` property with the addition of `isFullConfigReady`?

tbiethman (approved):

packages/app/cypress/e2e/support/execute-spec.ts
```diff
@@ -0,0 +1,32 @@
...
+export const waitForSpecToFinish = () => {
+  // First ensure the test is loaded
+  cy.get('.passed > .num').should('contain', '--')
+  cy.get('.failed > .num').should('contain', '--')
+
+  // Then ensure the tests are running
+  cy.contains('Your tests are loading...').should('not.exist')
+
+  // Then ensure the tests have finished
+  cy.get('[aria-label="Rerun all tests"]', { timeout: 30000 }).should('exist')
```
brian-mann, line 29: You don’t need the should(‘exists’) it’s an implicit / default assertion

ryanthemanuel:
```suggestion
  cy.get('[aria-label="Rerun all tests"]', { timeout: 30000 })
```

packages/launchpad/cypress/e2e/support/getPathForPlatform.ts
```diff
@@ -0,0 +1,11 @@
+import { decodeBase64Unicode } from '@packages/frontend-shared/src/utils/base64'
+
+export function getPathForPlatform (posixPath: string) {
+  // @ts-ignore
+  const cy = window.Cypress
+  const platform = cy?.platform || JSON.parse(decodeBase64Unicode(window.__CYPRESS_CONFIG__.base64Config)).platform
```
brian-mann, line 6: This code is really ugly and seems like it shouldn’t be necessary. Also isn’t it Cypress.platform not cy.platform? 

Also what’s the situation where this hasn’t been decided yet?

ryanthemanuel: I moved this code to reuse it, I'm not sure what prompted the original structure of it.

ryanthemanuel: Fixed here: [`dcc4122` (#21354)](https://github.com/cypress-io/cypress/pull/21354/commits/dcc4122f12b1e3aef14bd72acfd79b47af73203b)

(This is used for validation in tests by the way, not in application code)

ryanthemanuel:
In the spirit of moving forward with this, I'm going to merge even with the remaining failures. They are not related to my changes and are newly existing on 10.0-release. Merging what I have will improve stability somewhat and we can then focus on fixing the remaining issues.

PR #21358, opened 2022-05-05
fix: Git status ages don't update in real time

MuazOthman:
- Closes [UNIFY-1668](https://cypress-io.atlassian.net/browse/UNIFY-1668)

### User facing changelog
Better real-time tracking in spec list for Git changes

### PR Tasks


packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -310,6 +310,10 @@ export class ProjectActions {
     this.ctx.emitter.specsChange()
   }
 
+  refreshSpecs (specs: FoundSpec[]) {
+    this.ctx.lifecycleManager.git?.setSpecs(specs.map((s) => s.absolute))
+  }
```
tgriesser, line 315: Nice, can we get an e2e test that covers it? Would prob go in `specs_list_actual_git_repo.cy.ts`

tgriesser (changes requested):

packages/data-context/src/actions/ProjectActions.ts
MuazOthman: I have one coming after the UI updates for https://cypress-io.atlassian.net/browse/CLOUD-604
image

packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -217,21 +207,15 @@ export class GitDataSource {
...
-
-      debug('stdout %s', stdout)
+      const statusResult = await this.#git.status()
 
       const changed: string[] = []
 
       // Go through each file, updating our gitInfo cache and detecting which
       // entries have changed, to notify the UI
-      for (const [i, file] of absolutePaths.entries()) {
+      for (const file of absolutePaths) {
```
tgriesser, line 216: So I forgot about this in the pairing session earlier, but part of the reason we had it the way we did previously was because simpleGit spawns a child process per command, which gets very expensive when the number of specs grows. The way we have it currently batches it all into a single command to save on child process overhead.

I haven't looked closely at their API to determine if there's a way to batch them, they do have some notes about it here but it's a little vague: https://github.com/steveukx/git-js#concurrent--parallel-requests

If we can figure out why it's broken with the current approach and how to fix, that might be better - if we determine it's worth it to stick to simple git, we prob want to change this into a `Promise.all(absolutePaths.map(...` rather than a for loop, because the `for / await` will block and run sequentially rather than concurrently.

packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -302,9 +302,13 @@ export class GitDataSource {
 
     debug('executing command `%s`:', cmd)
 
-    const result = await execa(cmd, { shell: process.env.SHELL || '/bin/bash' })
+    const result = await execa(cmd, { shell: process.env.SHELL || '/bin/bash', cwd: this.#gitBaseDir })
```
tgriesser, line 305: Can we add a test covering this change in the `GitDataSource.spec.ts`?

packages/data-context/src/actions/ProjectActions.ts
tgriesser: We could add one with this PR and then just adjust it in the next PR so we don't miss it. We typically like to have a test covering behavior changes in the PR that introduces them

tgriesser (changes requested):

packages/data-context/src/sources/GitDataSource.ts
MuazOthman: I'm not sure if there's value in writing a test for this. I was able to find the problem by opening the cypress app in from within the cypress folder via `yarn cypress:open` and then navigating to list specs for a project. The issue was due to the fact that the cypress project folder was the default working directory since the spawned process would inherit that from the main process, which may be an OS-level or NodeJS behavior.
If you think we still need a test, then would this be a negative test by opening cypress in global mode?

tgriesser: I meant in the unit tests file it should be pretty simple to simulate the behavior you're describing, just a `process.chdir` or something to make a failing test that this change would make green, just a general practice to guards against it being lost in a future refactor: https://github.com/cypress-io/cypress/blob/1ed93ffebd446532e3df1b50913a60874b4c6ef1/packages/data-context/test/unit/sources/GitDataSource.spec.ts#L36-L47

tgriesser (approved):

packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -83,8 +83,10 @@ export class GitDataSource {
     // Simple Git will error if the projectRoot does not exist.
     // This should never happen outside of testing code simulating
     // incorrect scenarios
+    debug('config: %o', this.config)
```
rachelruderman, line 86: At my old gig we always removed console logs & debugger statements; at Cypress do we leave debugger statements in?

rachelruderman (approved):
LGTM

packages/data-context/src/sources/GitDataSource.ts
flotwig: Yes, some of these `debug` statements are way too verbose to put in `cypress:*`. Developers use `DEBUG=cypress:*` frequently to get a high level view of what Cypress is doing, so we should try not to pollute it [more than it already is].

There is actually a `cypress-verbose:*` namespace available in this file as `debugVerbose`, where you can put logs like https://github.com/cypress-io/cypress/blob/383bdb1d3ddcae645dfabc76b65b2aeb319b6216/packages/data-context/src/sources/GitDataSource.ts#L315 and https://github.com/cypress-io/cypress/blob/383bdb1d3ddcae645dfabc76b65b2aeb319b6216/packages/data-context/src/sources/GitDataSource.ts#L234 if you do feel they will be valuable in production.

MuazOthman: @flotwig I’ll keep this recommendation in mind, but in this specific scenario these statements were there before my code but appear to be under my name in `git blame` because of a revert I needed to do. I think the other debug statements I added (lines 86, 308, and 314) may or may not need to be verbose, what do you think?

PR #21370, opened 2022-05-06
fix: UNIFY-1625 Runs tab not updating in real time

tgriesser:
- Closes UNIFY-1625

### User facing changelog

The runs tab will refresh with the latest runs & updates to any "running" rows in an interval, currently every 15 seconds.

### Additional details

This PR includes improvements to the overall GraphQL layer in the Cypress app, and combines changes made in two other PRs (#21408, #21412) related to changes needed as improvements for UNIFY-1625 (runs tab updates) & the work @MuazOthman is doing around app/cloud integration

- Adds tests for the expected behavior re: fetching updated runs
- Adds some formalization around the caching & invalidation strategy for using the urql client on the server
  - `CloudDataSource` now manages all of this behavior, and adds
  - Adds unit tests for the new API surface area
- Improves our "pushing data to the client" when remote data has resolved into the query
  - Previously, we sent a signal that something on the client has changed, which triggered re-querying
  - Now, we `execute` the part of the query which has been updated and send that down in a `pushFragment` subscription payload, so the client can update the cache directly
  - Adds unit tests for the `pushFragment` behavior
- Simplifies the remote query delegation, via `delegateCloudField`, removes batch resolving, which we will revisit in future PRs
- Adds a `refetchRemote` mutation field, which will act as a signal to the electron process that we want to skip the server-side cache, and update any values we are requesting
- Patches URQL w/ changes merged in https://github.com/FormidableLabs/urql/pull/2431, currently blocked on upgrading to the latest urql graphcache, see https://github.com/FormidableLabs/urql/discussions/2440

### PR Tasks


packages/data-context/src/DataContext.ts
```diff
@@ -184,7 +205,14 @@ export class DataContext {
...
   @cached
   get cloud () {
-    return new CloudDataSource(this)
+    return new CloudDataSource({
+      fetch: (...args) => this.util.fetch(...args),
+      getUser: () => this.user,
+      logout: () => this.actions.auth.logout().catch(this.logTraceError),
+      onError: (err) => {
+        return this.onError(getError('DASHBOARD_GRAPHQL_ERROR', err), 'Cypress Dashboard Error')
+      },
```
tgriesser, line 214: Need to revisit whether this is actually what we want to do, or if we should just be putting `<Alert />` components in various places

packages/graphql/src/schemaTypes/objectTypes/gql-CurrentProject.ts
```diff
@@ -68,7 +67,12 @@ export const CurrentProject = objectType({
           return null
         }
 
-        return cloudProjectBySlug(projectId, ctx, info)
+        return ctx.cloud.delegateCloudField({
```
tgriesser, line 70: Removes needing to create specific resolvers

packages/data-context/src/sources/CloudDataSource.ts
```diff
@@ -29,117 +36,239 @@ const REMOTE_SCHEMA_URLS = {
...
+                cache.invalidate(...args)
+              },
+            },
+          },
+        }),
+        errorExchange({
+          onError: (err, operation) => {
+            // Ignore the error when there's no internet connection or when the
+            // cloud session is not valid, we want to logout the user on the app
+            if (err.graphQLErrors.some((e) => e.message.includes('Unauthorized'))) {
```
estrada9166, line 113: I think this should be checking `error.networkError?.message === 'Unauthorized'` as well

estrada9166 (approved):
Looks good, my only comment is that we may want to check as well `err.networkError?.message === 'Unauthorized'` to logout the user

packages/app/src/runs/RunsContainer.vue
```diff
@@ -92,30 +94,95 @@ fragment RunsContainer on Query {
...
-    isOnlineRef.value = true
-    emit('reexecuteRunsQuery')
+  return undefined as any
+})
+
+const refetcher = useMutation(RunsContainer_FetchNewerRunsDocument)
+
+// 15 seconds polling
+const POLL_FOR_LATEST = 1000 * 15
+const timeout = ref<null | number>(null)
```
lmiller1990, line 151: No need for `ref`, can just use `let` - this doesn't seem to need to be reactive.

packages/data-context/src/DataContext.ts
```diff
@@ -141,6 +157,11 @@ export class DataContext {
     return new FileDataSource(this)
   }
 
+  @cached
+  get node () {
```
lmiller1990, line 161: This seems like a kind of bad name, considering we also use `node` as in `node.js` quite a bit in the code base. Maybe `GraphQLNodeDataSource`? 🤔 naming is hard.

tgriesser: Moving it to the `GraphQLDataSource` - `ctx.graphql.resolveNode`

lmiller1990 (approved):
I tested it using `CYPRESS_INTERNAL_ENV=production`, works as expected. My comments have been responded to.

PR #21386, opened 2022-05-09
fix: handle specs with white space in vite-dev-server

lmiller1990:
- Closes https://github.com/lmiller1990/vue-3-cypress-vite/issues/1 (external repo)

### User facing changelog

Correctly handle spec files with white space in `@cypress/vite-dev-server`.

### Additional details

We were not doing `encodeURI` which made it impossible to load specs with white space for projects using `@cypress/vite-dev-server`. Some users reported it in an example repo I created for Vue and Vite. https://github.com/lmiller1990/vue-3-cypress-vite/issues/1 

### PR Tasks

### DIFF
--- a/npm/vite-dev-server/cypress/components/spec with whitespace.spec.ts
+++ b/npm/vite-dev-server/cypress/components/spec with whitespace.spec.ts
@@ -1,0 +1,5 @@
+describe('spec with whitespace', () => {
+  it('should work', () => {
+    expect(1).to.eq(1)
+  })
+})
--- a/npm/vite-dev-server/src/makeCypressPlugin.ts
+++ b/npm/vite-dev-server/src/makeCypressPlugin.ts
@@ -83,7 +83,7 @@ export const makeCypressPlugin = (
     load (id) {
       if (id === 'cypress:spec-loaders') {
         return `export default {\n${specs.map((s) => {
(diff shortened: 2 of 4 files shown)


ZachJW34 (approved):

reintroducing:
@lmiller1990 I just went through an updated all of my dependencies across repos (there are 3, but realistically that shouldn't make a difference here). I got to upgrading @cypress/vite-dev-server from `2.2.2` to `2.2.3` and didn't think much of it, until a few of my Cypress component tests stopped working. Eventually, one by one, I undid my dependency updates until I found that this is the dependency that is causing issues. It only happens in my repo for components that use https://github.com/framer/motion for some reason (and when I remove that dependency it goes away), BUT, that isn't really an option as it is tied into my component library and used for animations. That being said, the error that comes up is the following:
Screen Shot 2022-05-23 at 12 43 04 PM

What's really odd is that, outside of just defining NODE_ENV=test in my npm script that runs the tests, NODE_ENV isn't being used at all in this codebase. I'm using the latest versions of everything (except this library as explained above).

This seems incredibly suspect, especially because the changes here don't really seem like something that would impact this, but I wanted to get this in here because maybe someone looking at this can identify why this would be happening?

reintroducing:
Quick update: the issue was fixed in framer-motion.

lmiller1990:
Great, glad this was fixed!

PR #21388, opened 2022-05-09
fix: show correct error message when a regex starting with `=` is passed to `cy.contains`.

### LINKED ISSUE #21108: Unrecognized expression for regex that ends with `/=\s+[0-6]/`
### Current behavior

I have a `cy.contains()` call with the following:

```javascript
cy.contains(/[56]\s+-\s+[0-6]\s+=\s+[0-6]/);
```
And it fails with:

```
Error: Syntax error, unrecognized expression: :cy-contains('/[56]\s+-\s+[0-6]\s+='\s+[0-6']/'), [type='submit'][value~='/[56]\s+-\s+[0-6]\s+='\s+[0-6']/']
```

It looks like there are extra single quotes? I'm not sure. The minimal code that throws the same error is:

```javascript
cy.contains(/=\s+[0-6]/);
```

### Desired behavior

Should not throw Error "Syntax error, unrecognized expression".

### Test code to reproduce

https://github.com/asartalo/cypress-test-cy-contains-fail

### Cypress Version

9.5.4

### Other

_No response_

sainthkh:
- Closes #21108

### User facing changelog

Correct error message is shown and Cypress is properly retried when a regex starting with `=` is passed to `cy.contains`.

### Additional details
- Why was this change necessary? => Cypress doesn't work correctly when a regex starting with `=` is passed to `cy.contains`.
- Any implementation details to explain? => It was happening because Sizzle, jQuery selector engine escapes `=` to handle attribute selectors. I changed our jquery config code to handle this exceptional case. 

```js
// Before: Fail with Syntax Error message + doesn't retry.
// After: Timed out error message + retry.
cy.contains(/=[0-6]/) 
```

### PR Tasks

### DIFF
--- a/packages/driver/cypress/integration/commands/querying/querying_spec.js
+++ b/packages/driver/cypress/integration/commands/querying/querying_spec.js
@@ -1230,6 +1230,16 @@ describe('src/cy/commands/querying', () => {
       cy.get('#backslashes').contains('"<OE_D]dQ\\')
     })
 
+    // https://github.com/cypress-io/cypress/issues/21108
+    it('shows correct error message when regex starts with =(equals sign)', () => {
+      cy.on('fail', (err) => {
+        expect(err.message).to.include('Expected to find content')
+      })
+
+      cy.visit('fixtures/dom.html')
+      cy.contains(/=[0-6]/, { timeout: 100 }).should('have.text', 'a=2')
+    })
+
     describe('should(\'not.exist\')', () => {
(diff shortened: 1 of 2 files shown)


BlueWinds:
For anyone looking at this later, note that the selector already finds the element if it exists on the page - this PR just ensures that Cy waits for it to exist if it doesn't already. See https://github.com/cypress-io/cypress/issues/21108#issuecomment-1102172717.

PR #21396, opened 2022-05-09
fix(unify): Have better default filename than `filename` for create spec file windows

davidmunechika:
- Closes [UNIFY-1630](https://cypress-io.atlassian.net/browse/UNIFY-1630?atlOrigin=eyJpIjoiODhiYjU2NTQzMjUwNDQwN2JiZGM4MzJkZDNkN2NhNDAiLCJwIjoiaiJ9)

### User facing changelog
In the create new spec modal, the default filename for e2e testing will be `spec` and the default filename for component testing will be `ComponentName`.

### Additional details
Before:
image

After:

Screenshot 2022-05-09 at 10 07 33 AM

Screenshot 2022-05-09 at 10 07 01 AM

### PR Tasks


packages/app/src/specs/CreateSpecModal.vue
```diff
@@ -109,7 +109,7 @@ const helpLink = computed(() => {
 const specFileName = computed(() => {
   const extension = props.gql.currentProject?.fileExtensionToUse ?? 'js'
 
-  const fileName = props.gql.currentProject?.defaultSpecFileName ?? `cypress/e2e/filename.cy.${extension}`
+  const fileName = props.gql.currentProject?.defaultSpecFileName ?? (props.gql.currentProject?.currentTestingType === 'e2e' ? `cypress/e2e/spec.cy.${extension}` : `src/ComponentName.spec.${extension}`)
```
ZachJW34, line 112: Looking through `defaultSpecFileName` in ProjectDataSource, it's a bit strange that we allow `defaultSpecFileName` to be null. Instead, we could just return the default and mark the field as `nonNull`, then we wouldn't have to have defaults on the frontend and backend.

packages/app/src/specs/CreateSpecModal.cy.tsx
```diff
@@ -113,3 +113,90 @@ describe('playground', () => {
...
+            fileExtensionToUse: 'js',
+            defaultSpecFileName: null,
+          },
+        }}
+        show={show.value}
+        onClose={() => show.value = false}
+      />
+    </div>))
+
+    cy.get('input').invoke('val').then((val) => {
```
ZachJW34, line 198: Can be reworked to `cy.get('input').invoke('val').should('contain', 'ComponentName.spec.js')`, same for the test above.

ZachJW34 (commented):
Looks good, though I think we can cleanup `defaultSpecFileName`. Also noticed that we are no longer using `codeGenGlobs` so I'll create an issue for removing it.

packages/app/src/specs/CreateSpecModal.vue
davidmunechika: @ZachJW34 How would I make `defaultSpecFileName` be a `nonNull` field if `currentProject` could still potentially be null?

tbiethman:
Heads up, looks like this PR and https://github.com/cypress-io/cypress/pull/21411 are making similar changes (for different reasons). Someone can anticipate some minor conflicts 😅

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -122,7 +122,7 @@ export function getDefaultSpecFileName (specPattern: string, fileExtensionToUse?
   }
 
   const splittedDirname = dirname.split('/').filter((s) => s !== '**').map((x) => replaceWildCard(x, 'e2e')).join('/')
-  const fileName = replaceWildCard(parsedGlob.path.filename, 'filename')
+  const fileName = testingType === 'e2e' ? replaceWildCard(parsedGlob.path.filename, 'spec') : replaceWildCard(parsedGlob.path.filename, 'ComponentName')
```
tbiethman, line 125:
```suggestion
  const fileName = replaceWildCard(parsedGlob.path.filename, testingType === 'e2e' ? 'spec' : 'ComponentName')
```

tbiethman (approved):

packages/app/src/specs/CreateSpecModal.vue
ZachJW34: At this stage, it doesn't make sense for `currentProject` to be null. Even if it was, the frontend code relies on there being a `currentProject` to determine the fileName anyway. The check for `currentProject` is more of a TS safety check, so we could replace the `return null` by throwing an error, or just return the default e2e spec name.

lmiller1990:
I tried and failed to rebase this - not enough context. I know you are finishing up soon @davidmunechika, will you have time rebase this?

packages/app/src/specs/CreateSpecModal.vue
ZachJW34: @tbiethman now that we are always returning a `defaultSpecFileName` can we remove
```
?? (props.gql.currentProject?.currentTestingType === 'e2e' ? `cypress/e2e/spec.cy.${extension}` : `src/ComponentName.spec.${extension}`)
```

ZachJW34 (approved):
One small comment, looks you you are working on fixing some failing windows test but going to ✅

tbiethman:
@ZachJW34 Yep, working on getting those Windows tests updated and your final comment addressed, I'll request another review if anything drastic changes.

packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
```diff
@@ -226,7 +226,7 @@ describe('getDefaultSpecFileName', () => {
     it('replace * for filename', () => {
       const defaultFileName = getDefaultSpecFileName('cypress/e2e/*.spec.ts', 'e2e')
 
-      expect(defaultFileName).to.eq('cypress/e2e/filename.spec.ts')
+      expect(defaultFileName).to.eq('cypress/e2e/spec.spec.ts')
```
rachelruderman, line 229: `spec.spec.ts` looks a little funky to me.... thoughts on something like `integration-test.spec.ts`?

packages/app/src/specs/CreateSpecModal.cy.tsx
```diff
@@ -103,3 +103,76 @@ describe('playground', () => {
...
+            }],
+            specs: [],
+            fileExtensionToUse: 'js',
+            defaultSpecFileName: 'spec.cy.js',
+          },
+        }}
+        show={show.value}
+        onClose={() => show.value = false}
+      />
+    </div>))
```
estrada9166, line 138: Should we use `cy.mountFragment` instead?

tbiethman: Yep that'll clean things up a bit. I was also able to remove one of these test cases now that we no longer flex within the component based on testing type.

estrada9166 (approved):

rachelruderman:
Based on the before/after screenshots in the description, I was wondering...
question

marktnoonan:
> Based on the before/after screenshots in the description, I was wondering... question

@rachelruderman This section was removed elsewhere, intentionally, I'm betting the "before" was taken before that change was merged into this branch.

PR #21409, opened 2022-05-09
fix: Unify 1232 the keyboard shortcuts button is not keyboard accessible

rachelruderman:
- Closes UNIFY-1232

### User facing changelog
N/A

### Additional details

1. Fixes keyboard accessibility for keyboard shortcuts icon
2. Resolves other accessibility errors in this view as detected by AXE devtools by adding alt tags, aria control labels, iframe title and semantic html
3. General changes for consistency, i.e. use of `findByTestId` and capitalization of `sidebar` variable
4. Adds `cypress-axe` and `cypress-plugin-tab` packages to `frontend-shared`

Accessibility gains in the form of keyboard tabbing in the sidebar and improved screenreader experience 💪 
#### Before
keyboard-after

#### After
keyboard-after

### PR Tasks


packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
```diff
@@ -62,17 +62,17 @@ describe('Cypress In Cypress CT', { viewportWidth: 1500, defaultCommandTimeout:
       cy.get('[data-model-state="passed"]').should('contain', 'renders the test component')
 
       // go to Settings page and back to spec runner
-      cy.contains('a', 'Settings').click()
+      cy.findByTestId('nav-row-settings').click()
```
tbiethman, line 65: I like how you updated all these to use consistent commands/selectors. 🥳 

Have you considered writing a custom cypress command to handle these 'navigate to route' flows that are so well defined? Something like:

```js
// packages/app/cypress/e2e/support/e2eSupport.ts

Cypress.Commands.add('navigateAppTo', (route: 'specs' | 'runs' | 'settings') => {
  return cy.findByTestId(`nav-row-${route}`).click()
})


// usage
cy.navigateAppTo('runs')
```

Might be a little more readable?

packages/app/cypress/e2e/sidebar_navigation.cy.ts
```diff
@@ -1,6 +1,47 @@
...
+      .tab().should('have.attr', 'data-cy', 'sidebar-header').should('have.attr', 'role', 'button')
+      .tab().should('have.attr', 'data-cy', 'nav-row-specs').should('have.prop', 'tagName', 'BUTTON')
+      .tab().should('have.attr', 'data-cy', 'nav-row-runs').should('have.prop', 'tagName', 'BUTTON')
+      .tab().should('have.attr', 'data-cy', 'nav-row-settings').should('have.prop', 'tagName', 'BUTTON')
+      .tab().should('have.attr', 'data-cy', 'keyboard-modal-trigger').should('have.prop', 'tagName', 'BUTTON')
+    })
+
+    it('has no axe violations', () => {
+      cy.injectAxe()
+      cy.checkA11y('[data-cy="sidebar"]')
```
tbiethman, line 25: 🔥 🔥 🔥

packages/app/src/navigation/SidebarNavigationHeader.vue
```diff
@@ -6,6 +6,8 @@
     :distance="8"
   >
     <div
+      role="button"
+      :aria-label="t('testingType.modalTitle')"
```
tbiethman, line 10: Setting an explicit label here will cause a screen reader to ignore the button's text content (project/branch). That'd probably be fine if the project were announced elsewhere (part of the title?), but I don't think it is. 

We could compose the label (`${project}, ${branch}, ${t('testingType.modalTitle')}`)? This button feels overloaded.

rachelruderman: Agh I agree, this button is definitely overloaded 

I think we could get away with `${project} - ${t('testingType.modalTitle')` with the rationale being: The user isn't depending on Cypress to tell them which git branch they're on, but they definitely are depending on Cypress to indicate the project name and how to switch testing types

We _might_ be able to get away with just `:aria-label="t('testingType.modalTitle')"` given that this PR adds a `title` attribute to the iframe with the projectId:

Screen Shot 2022-05-16 at 12 14 14 PM

@tbiethman what do you think? I can go either way

packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts
rachelruderman: Mark made a good point that I mistakenly turned this anchor tag into a button 😅 it's been reverted back to anchor tag so no diffs here 🎉

packages/app/cypress/e2e/sidebar_navigation.cy.ts
marktnoonan: 🎉

marktnoonan (approved):
Delighted to get this stuff tightened up, tested manually and all works great.

tbiethman (approved):

PR #21411, opened 2022-05-09
fix: use correct path for scaffolding spec on CT

estrada9166:
- Closes  [UNIFY-1614](https://cypress-io.atlassian.net/browse/UNIFY-1614)

### User facing changelog

We were scaffolding the CT specs on the root of `cypress` folder, in this case, if we identify that the specPatter used is the default one, we create the empty spec on `cypress/component`

### PR Tasks


packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -282,13 +282,15 @@ export class ProjectDataSource {
   }
 
   async defaultSpecFileName () {
-    const defaultFileName = 'cypress/e2e/filename.cy.js'
+    const getDefaultFileName = (testingType: TestingType) => `cypress/${testingType}/filename.cy.${this.ctx.lifecycleManager.fileExtensionToUse}`
```
lmiller1990, line 285: This doesn't seem correct. What if `cypress/component` does not exist? Considering the default is something like `src/**/*.cy.{js.jsx}`, it likely won't exist for most users.

Should we make sure the default placeholder is a valid path (or we don't care?)

We could check against the specPattern to make sure it matches (we've already got this logic [here](https://github.com/cypress-io/cypress/blob/10.0-release/packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts#L64-L76).

If it does not match, we could do something like:

```js
const root = specPattern.split('/')[0]
const getDefaultFileName = path.join(root, 'filename.cy', this.ctx.lifecycleManager.fileExtensionToUse)`
```

What do you think?

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -282,13 +282,15 @@ export class ProjectDataSource {
...
 
     try {
```
lmiller1990, line 287: Not directly related but do we need this `try/catch`?

lmiller1990 (commented):
One comment.

packages/data-context/src/sources/ProjectDataSource.ts
estrada9166: That's why it is checked [here](https://github.com/cypress-io/cypress/pull/21411/files/2edb1c396c31c65b00083548a8bc9a319ec1b02c#diff-6085376a4a2ce7de42456fe0abf38e4ddbd45dff4851894f456598e40e586cc7R305-R307) to be sure it's the default specPattern - so, this ^^ is just for the default paths

when there're custom specPatterns as you mention it is updated [here](https://github.com/cypress-io/cypress/pull/21411/files/2edb1c396c31c65b00083548a8bc9a319ec1b02c#diff-6085376a4a2ce7de42456fe0abf38e4ddbd45dff4851894f456598e40e586cc7R309) using the custom pattern and wild cards

lmiller1990: 🎉

lmiller1990 (approved):

lmiller1990 (approved):
Looks good, I think. The spec and pattern related logic is starting to get very confusing - I wonder if we should look to move it to either a) separate data source or b) separate module (or even just inside of `@packages/config`.

I think data-context should just be a thin layer on how we access data and write data, not the place where we handle the complexity. I don't think we need to do that now, but we should keep it in mind as we start to transition from release mode to stability and maintainability mode.

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -300,15 +302,19 @@ export class ProjectDataSource {
...
+      const specFileName = getDefaultSpecFileName(specPatternSet, this.ctx.coreData.currentTestingType, this.ctx.lifecycleManager.fileExtensionToUse)
 
       if (!specFileName) {
         return defaultFileName
       }
 
       return specFileName
     } catch {
-      return defaultFileName
+      return getDefaultFileName('e2e')
```
tbiethman, line 317: It looks like we could just try/catch the awaited specPatterns call and abort on line 302 in the case of a failure and not need separate error logic here.

If we do need to keep this, though, why not use currentTestingType instead of 'e2e'?

estrada9166: I left the try/catch but updated the testing type - the reason is bc `getDefaultSpecFileName` is using different packages and is safer to catch it if there's any possible error while trying to create the fileName depending of the specPattern

tbiethman (approved):

PR #21415, opened 2022-05-09
feat(cypress/schematic): add headed option and other fixes

### LINKED ISSUE #18745: Angular schematic ignores baseUrl if a devServerTarget is defined
### Current behavior

We recently made a switch to use the Cypress' official Angular schematic, @cypress/schematic. After doing so, some of our commands stopped working - particularly when a `devServerTarget` is specified.

For background, our app uses different routing when it's run locally. So instead of https://localhost:4200, our landing page lives at https://localhost:4200/foo. Well the `baseUrl` defined in our e2e configurations is meant to account for this. Oddly enough, though, when we run an e2e command that specifies the `devServerTarget` && the `baseUrl` the `baseUrl` gets completely ignored!

I've put together an example app. There is a workaround, which is to define the `deployUrl` for your app in that particular configuration in the `angular.json`. I'll add that this is probably best practice. But if I specify an option in my build config... I'd think it would get used. Right?

### Desired behavior

If I specify a `baseUrl` in my e2e configurations using the @cypress/schematic builder, then that `baseUrl` should be what Cypress uses when I do a `cy.visit()` or a `cy.log(${Cypress.config().baseUrl})`.

### Test code to reproduce

https://github.com/ashblox/base-url-problem

### Cypress Version

8.7.0

### Other

@cypress/schematic version 1.5.1

### LINKED ISSUE #21332: @cypress/schematic: can't `run` in headed mode
### Current behavior

Even when passing `headless: false` in `angular.json`, I can't find a way to set the `headed` flag on the actual runner.

### Desired behavior

Not setting `headless` or setting it explicitly to `false` through the builder should open cypress in headed mode

### Test code to reproduce

```sh
ng e2e --headless=false # Note that this is the default
```

### Cypress Version

9.6.0, but also affects ~9.5.x

### Other

https://github.com/cypress-io/cypress/blob/cb14ae62e7c4362d759ceb450f4af518ee486f90/npm/cypress-schematic/src/builders/cypress/index.ts#L70-L75

It seems that there is no way to actually set the `headed` flag explicitly. I kind-of fixed it by patching the schematic with this: 

```patch

--- a/node_modules/@cypress/schematic/src/builders/cypress/index.js
+++ b/node_modules/@cypress/schematic/src/builders/cypress/index.js
@@ -36,7 +36,8 @@ function initCypress(userOptions) {
     };
     const options = Object.assign(Object.assign(Object.assign({}, defaultOptions), userOptions), { 
         //@ts-ignore
-        dev: process.env.CYPRESS_ENV === 'test' });
+        dev: process.env.CYPRESS_ENV === 'test', headed: !userOptions.headless });
+        console.log(options)
     if (userOptions.configFile === undefined) {
         options.config = {};
     }
```

admah:
- Fixes #18745 - custom `baseUrl` can be specified via configuration in `angular.json`. More details are in README.md
- Fixes #21332 - `headed` is now a valid option in `run` mode

It appears like a large change because I renamed the sandbox app to prepare for multiple version support.


lmiller1990:
@admah you may need to sign in to CircleCI again, or it won't trigger CI to run.

npm/cypress-schematic/sandbox13/README.md
```diff
@@ -0,0 +1,27 @@
+# Sandbox
+
+This project was generated with [Angular CLI](https://github.com/angular/angular-cli) version 12.0.0.
```
lmiller1990, line 3: The title of the file (sandbox13) suggests this is angular 13, but the readme says angular 12?

lmiller1990 (commented):
I see the `headed` fix: https://github.com/cypress-io/cypress/pull/21415/files#diff-4ee798fa6c842712558f99121e5b14db00d1410c489de4bb52597724123be34eR11, but where is the code for "custom baseUrl can be specified via configuration in angular.json. More details are in README.md"? I don't see any changes to this meaning.

jordanpowell88 (changes requested):

lmiller1990 (approved):
Code seems fine from what I can see, just need CI to be ✅

jordanpowell88 (approved):

Issue #21236, opened 2022-04-27
Restore just-in-time `require`/`require` splitting in `cypress.js`


flotwig:
See https://github.com/cypress-io/cypress/pull/21174 for context.

We should remove these lines:

https://github.com/cypress-io/cypress/blob/244c1cc5c08112bda6bd60d0cef096b26a8595b6/packages/server/lib/cypress.js#L3-L10

Because `cypress.js` should not be slowed down by unused code:

https://github.com/cypress-io/cypress/blob/244c1cc5c08112bda6bd60d0cef096b26a8595b6/packages/server/lib/cypress.js#L12-L20

Or, remove the need for `cypress.js` to ever run in `node`.

Once `cypress.js`'s JIT `require`s are restored, we can clean up these areas:

* No longer will need to JIT require here, since we will no longer be requiring `makeDataContext` outside of `electron`: https://github.com/cypress-io/cypress/blob/244c1cc5c08112bda6bd60d0cef096b26a8595b6/packages/server/lib/makeDataContext.ts#L150-L168
* Any other areas where we currently have to JIT require to avoid triggering errors related to `electron` not being available.

Issue #21300, opened 2022-05-02
Visiting a cross origin page with a `cy.origin` command with no cy commands may cause subsequent tests to hang.


mjhenkes:
### Current behavior

Since we no longer wait for stability before ending a test test we may now end up in a race condition where an unload event initiated in a previous test will occur in a subsequent test, setting stability to false. If that load event happens for a cross origin page the test will hang, forever since stability won't be set back to true.

It was assumed that setting the iframes src to the `about:blank` page would cancel any current load events happening within the iframe, but the opposite appears to be true, if the iframe is currently loading

### Desired behavior

The second test should not hang based on commands run in the first test.

I believe we have to options to fix this, one more extreme than the other.

1. Wait for stability prior to loading the about blank page in [sessions/index](https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cy/commands/sessions/index.ts#L321). To achieve stability here, we will have know which spec bridge to continue to listen too and persist that state between tests. Currently this is handled within tests by the `latestActiveOriginPolicy` state, but that is cleared between tests. See relevant [event-manager lines](https://github.com/cypress-io/cypress/blob/develop/packages/runner-shared/src/event-manager.js#L538)

2. Revert the change that allows us to proceed in the event manager regardless of stability and instead allow `cy.visit` to successfully load cross-origin pages. This requires more investigation, but would have the dual effects of making it less confusing to visit pages (you could visit a cross origin page prior to entering a cy.origin block and allow us to remove code around syncing stability as you could not exit the command queue without being stable. Any any commands run after a visit would have to run against the correct origin. The downside of this approach would be that again, we'd be waiting on stability before exiting mocha blocks (but we've seen what not waiting can do) and we'd have to error in `cy.visit`s on load functions if someone attempts to run them against a cross origin url. On loads would also have to be setup prior to visiting.

### Test code to reproduce

The following test can be used to reproduce this issue in the 'driver' package of the cypress repo. This is most easily reproduced on firefox, but has also been observed on chrome.

```javascript
// The first test executes fast enough that the onunload event occurs at the beginning of the second test.
it.only('runs', () => {
  cy.visit('/fixtures/primary-origin.html')
  cy.get('a[data-cy="cross-origin-secondary-link"]').click()
  cy.origin('http://foobar.com:3500', () => {
    expect(true).to.equal(true)
  })
})

it.only('it hangs', () => {
  cy.log('you will not see me')
})
```

### Cypress Version

9.6.0

[...]

PR #17807, opened 2021-08-19
fix: Improve Big Sur browser detection performance

chrisbreiding:
- Closes #17773

### User facing changelog

- Projects will now open much faster on MacOS Big Sur with improved browser detection performance.

### Additional details

The root of the issue is [this bug in Electron](https://github.com/electron/electron/issues/26143), which causes spawning child processes to take 300-1000ms instead of the usual <30ms when using MacOS Big Sur and Electron 12+. We spawn a child process for each binary of each browser, plus potential fallback lookups if any of the standard ones fail. This causes the browser detection on Big Sur to take upwards of 10 seconds.

We considered using the Electron patch in that issue, but it's not a perfect solution (it was reverted in libuv due to failing tests), and would require us to build Electron ourselves, which is a large amount of work and responsibility.

As a workaround, we spawn a single child process (only incurring the performance penalty once instead of N+ times), and use it to call into our browser detection code from the user's Node.js. Since it's outside Electron, it doesn't suffer from the same bug and performs quickly. This reduces the time taken for browser detection from ~10s to ~1s.

### How has the user experience changed?

Users on Big Sur will see projects open much quicker.

### PR Tasks

- [x] Have tests been added/updated?
- [x] Has the original issue or this PR been tagged with a release in ZenHub? 
- N/A Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- N/A Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- N/A Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

Issue #21329, opened 2022-05-03
Update internal `@types/node`


flotwig:
`@types/node` should be upgraded across all Cypress packages. We lose out with incorrect/outdated typings.

Acceptance criteria:

* Update `@types/node` to the latest major version that matches our Node version (currently 16.x)
* Fix all type issues that arise
* See if we can remove any of these type overrides https://github.com/cypress-io/cypress/blob/cbda09aca5ef5ff31e29804f7374d4f145d36dfe/packages/ts/index.d.ts
* Try to find other type overrides/`ts-ignore`s that can now be removed
* Update `packages/electron/README.md#Upgrading-Electron` with information about ALSO bumping `@types/node` when Node is bumped.

PR #21418, opened 2022-05-10
chore(deps): Updating electron to v18 + node v16.13.2

tbiethman:
### User facing changelog

- Upgraded `electron` from `15.5.1` to `18.0.4`.
- Upgraded bundled Node.js version from `16.5.0` to `16.13.2`.
- Upgraded bundled Chromium version from 94.0.4606.81 to 100.0.4896.75.

### Additional details

The [Upgrading Electron](https://github.com/cypress-io/cypress/tree/develop/packages/electron#upgrading-electron) guide was followed to produce this PR.

I included inline comments where additional clarification was needed. A few changes of note:

* Issues were found with some of the `origin` related tests in newer versions of Firefox. An [issue](https://github.com/cypress-io/cypress/issues/21300) was logged, and the affected tests have been tweaked to pass in the meantime without being outright skipped.
* The workaround added with #17807 in regards to browser detection performance has been removed, as the Electron issue causing the slowdown has been corrected.

There shouldn't be any functional difference in the user experience beyond any changes arising from the bundled Chromium update. If other differences are seen, please raise them.

### PR Tasks


circle.yml
```diff
@@ -1247,7 +1257,7 @@ jobs:
 
   runner-integration-tests-firefox:
     <<: *defaults
-    resource_class: medium
+    resource_class: medium+
```
tbiethman, line 1260: The resources were bumped for the Firefox tests due to increased memory consumption. https://github.com/cypress-io/cypress/issues/21319 was logged to investigate further.

circle.yml
```diff
@@ -896,7 +904,9 @@ commands:
             yarn binary-build --platform $PLATFORM --version $(node ./scripts/get-next-version.js)
       - run:
           name: Zip the binary
-          command: yarn binary-zip --platform $PLATFORM
+          command: |
+            apt-get update && apt-get install -y zip
```
tbiethman, line 908: `zip` was removed from our base image generator in a recent slimming down effort. We can add it back/generate a new image if there's a concern with pulling it in here.

packages/driver/cypress/integration/commands/actions/trigger_spec.js
```diff
@@ -589,13 +589,10 @@ describe('src/cy/commands/actions/trigger', () => {
...
       it('passes options.animationDistanceThreshold to cy.ensureElementIsNotAnimating', () => {
-        const $btn = cy.$$('button:first')
-
-        const { fromElWindow } = Cypress.dom.getElementCoordinatesByPosition($btn)
-
         cy.spy(cy, 'ensureElementIsNotAnimating')
 
-        cy.get('button:first').trigger('tap', { animationDistanceThreshold: 1000 }).then(() => {
+        cy.get('button:first').trigger('tap', { animationDistanceThreshold: 1000 }).then(($btn) => {
+          const { fromElWindow } = Cypress.dom.getElementCoordinatesByPosition($btn)
```
tbiethman, line 595: These tests related to `animationDistanceThreshold` across trigger/type command tests were failing in a similar way in Firefox.

Firefox reports positions/scroll offsets with something like 13 digits of precision. We do a lot rounding in our [coordinate](https://github.com/cypress-io/cypress/blob/3e6d6bfe155234f782a890fd904b361c84d5c34e/packages/driver/src/dom/coordinates.ts#L174-L179) logic that could be impacted by this. These tests also seem to be influenced by how the AUT is being scaled; I can get them to pass/fail in develop by running the tests with the AUT scaled at different sizes.

I updated these tests to grab the reference element window coordinates after the element has been scrolled into position, which makes us less susceptible to variances within Firefox's reported scroll position. This allows them to pass consistently without need to add 1px buffers to our assertions.

packages/driver/cypress/integration/e2e/origin/commands/navigation.spec.ts
```diff
@@ -22,9 +22,18 @@ context('cy.origin navigation', () => {
...
+        // @ts-ignore
+        localWindow.cy_testCustomValue = true
+      })
+
+      cy.window().should('have.prop', 'cy_testCustomValue', true)
+
       cy.reload()
-      cy.get(':checkbox[name="colors"][value="blue"]').should('not.be.checked')
+
+      cy.window().should('not.have.prop', 'cy_testCustomValue', true)
```
tbiethman, line 36: Firefox now caches input values between reloads, causing this test to fail. Updated to use a window prop that we can be certain will be reset between reloads.

packages/web-config/tsconfig.json
```diff
@@ -56,7 +56,8 @@
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
     "noErrorTruncation": true,
-    "experimentalDecorators": true
+    "experimentalDecorators": true,
+    "skipLibCheck": true
```
tbiethman, line 60: `web-config` was the only package that didn't have skipLibCheck enabled. I attempted to update our node types to v16 but it seems to cause more problems than it fixed; https://github.com/cypress-io/cypress/issues/21329 was logged to dedicate more time this effort.

lmiller1990:
I wonder if this will solve https://github.com/cypress-io/cypress/issues/15932 🤔

tbiethman:
@lmiller1990 I was still able to reproduce #15932:

Screen Shot 2022-05-10 at 10 07 20 AM

packages/driver/cypress/integration/e2e/origin/origin.spec.ts
```diff
@@ -237,12 +247,13 @@ describe('cy.origin', () => {
...
           expect(err.codeFrame).to.exist
           expect(err.codeFrame!.frame).to.include('cy.origin')
+
           done()
         })
 
         const variable = () => {}
 
-        cy.origin('http://idp.com:3500', { args: variable }, (variable) => {
+        cy.origin('http://foobar.com:3500', { args: variable }, (variable) => {
```
tbiethman, line 256: This spec would hang across all browsers when visiting idp.com; switching to foobar.com like all other specs in this file caused the test to pass (well, fail in the way we expect it too). This is odd, but not sure if it's indicative of a larger problem or the same problem recorded in https://github.com/cypress-io/cypress/issues/21300.

tbiethman:
Note: we're currently pinned to electron 18.0.4 because 18.1.0 introduced a breaking change that has since been fixed (within the last day): https://github.com/electron/electron/pull/34150 

We should be able to consume the 18.2.2 release when it goes out.

AkechiShiro:
Any news when this dependency bump will happen, I'm just asking for some security related matters (the old dependencies have a few minors issues, nothing really critical)

Maybe soon, in a few weeks or it is still being at an early work of being adopted in the main branch ? Needs a few months of work still ?

tbiethman:
@AkechiShiro I'd anticipate these dependency bumps to be included in our next 9.x release in two weeks.

packages/driver/cypress/integration/commands/actions/trigger_spec.js
lmiller1990: I wonder if we can use a similar technique to remove this 1px hack for Firefox in 10.0-release? https://github.com/cypress-io/cypress/blob/10.0-release/packages/driver/cypress/e2e/commands/actions/type.cy.js#L360-L377

lmiller1990 (approved):
Great to see us finally updating electron. I left one comment about a potential update we can make in 10.0-release, that will be landing soon.

packages/driver/cypress/integration/e2e/origin/config_env.spec.ts
```diff
@@ -10,14 +10,17 @@
...
       // @ts-ignore
       window.top.__cySkipValidateConfig = true
+
+      // FIXME: Tests that end with a cy.origin command and enqueue no further cy
+      // commands may have origin's unload event bleed into subsequent tests
+      // and prevent stability from being reached, causing those tests to hang.
+      // We enqueue another cy command after each test to ensure stability
+      // is reached for the next test. This additional command can be removed with the
+      // completion of: https://github.com/cypress-io/cypress/issues/21300
+      cy.then(() => { /* ensuring stability */ })
```
mjhenkes, line 23: We still need this even after removing the before each? I wonder if there is another problem lurking here.

tbiethman: I'll confirm whether or not this is still necessary with the removal of the beforeEach. At one point I thought it was, but local runs seem to pass without it now, could have been an error on my part.

mjhenkes (approved):

packages/driver/src/cy/commands/actions/type.ts
```diff
@@ -360,9 +360,10 @@ export default function (Commands, Cypress, cy, state, config) {
               // We don't want to send it twice.
               !Cypress.isBrowser('firefox') ||
               // After Firefox 98,
-              // it sends a click event automatically if the element is a <button>
-              // it does not if the element is an <input>
-              (!isFirefoxBefore98 && $elements.isInput(event.target))
+              // it sends a click event automatically if the element is a <button>,
+              // but it does not if the element is an <input>.
+              // event.target is null when used with shadow DOM.
```
flotwig, line 365: How did this crop up during the Electron upgrade? Does this close an existing issue?

packages/launcher/lib/detect.ts
```diff
@@ -141,44 +140,13 @@ function checkOneBrowser (browser: Browser): Promise<boolean | HasVersion> {
...
 /** returns list of detected browsers */
-export const detect = (goalBrowsers?: Browser[], useDarwinWorkaround = true): Bluebird<FoundBrowser[]> => {
+export const detect = (goalBrowsers?: Browser[]): Bluebird<FoundBrowser[]> => {
   // we can detect same browser under different aliases
   // tell them apart by the name and the version property
   if (!goalBrowsers) {
     goalBrowsers = browsers
   }
 
-  // BigSur (darwin 20.x) and Electron 12+ cause huge performance issues when
```
flotwig, line 151: :tada: :tada: :tada: :tada: :tada: :tada:

packages/web-config/tsconfig.json
flotwig: `web-config` is used in several packages btw. What motivated this change?

flotwig (changes requested):
Really nice work @tbiethman, I know this was a bear of an upgrade.

Will this be a separate PR to close #21236?

packages/launcher/lib/darwin/util.ts
```diff
@@ -2,12 +2,8 @@ import { log } from '../log'
 import { notInstalledErr } from '../errors'
 import { utils } from '../utils'
 import * as fs from 'fs-extra'
-import * as os from 'os'
 import * as path from 'path'
 import * as plist from 'plist'
-import * as semver from 'semver'
-import type { FoundBrowser } from '../types'
-import * as findSystemNode from '@packages/server/lib/util/find_system_node'
```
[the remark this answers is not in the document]
flotwig: Oops, never mind, it's used in new places in `10.0-release`.

packages/driver/src/cy/commands/actions/type.ts
tbiethman: The Firefox version in the bundled image was bumped to v98 from v93. There have been PRs over the last few months to fix some piece-mail issues with FF98, starting with https://github.com/cypress-io/cypress/pull/21042. This seems to be the last hold-out. We have similar checks in place in other type logic: https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cy/commands/actions/type.ts#L373

packages/launcher/lib/darwin/util.ts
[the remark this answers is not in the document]
flotwig: Waaaait... It's used in a new place, but that function is never called by anything... I can clean it up after this makes its way into 10.0-release.

packages/web-config/tsconfig.json
tbiethman: Bumping to electron 18 caused `yarn type-check` to fail for this package due to conflicts in some deep transient dependencies. I assume this was due to electron's bumping to @types/node@16: https://github.com/cypress-io/cypress/pull/21418/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR8720-R8723. Looks like we might have been inadvertently(?) resolving against v15 before?

My type-config-fu is admittedly weak. It seems like we can resolve the failures if I add an explicit @types/node@14 dependency to this package. Would that be preferred? My understanding was that this flag forced a check on types we are actually referencing rather than on all dependencies.

flotwig: Hmm, this seems fine after reading the config doc: https://www.typescriptlang.org/tsconfig#skipLibCheck I don't think you need to change it. The doc says we can use Yarn `resolutions` maybe to pin it to 14 everywhere, but again, I don't think it's really an issue.

tbiethman:
@flotwig Regarding https://github.com/cypress-io/cypress/issues/21236, I was planning on following up with a quick PR after merging develop into 10.0-release. I've confirmed that Electron 18 does not show the same issues that 15.5.1 does with the increased verification time.

packages/driver/cypress/integration/e2e/origin/config_env.spec.ts
tbiethman: Can confirm it's not necessary, thanks for calling me on this one.

flotwig (approved):

AtofStryker (approved):

lmiller1990:
Recording breaking four ✅ - let's ship it 😎

tbiethman:
@lmiller1990 That comment has been addressed, seems like it's just not showing the full convo in Zach's review summary. We're good to go.

PR #21422, opened 2022-05-10
fix: test-recording instructions in Component Test mode

marktnoonan:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1637

### User facing changelog

In component testing mode, when no runs have been recorded to the dashboard, the command to copy will now include the `--component` flag, so that we run in component testing mode.

Additionally, the project root has been removed from the terminal prompt in all situations (per Brian's comment in this ticket). Terminal prompts now begin with `$` always, instead of `some-project: $`.

Before - a command is shown in component testing mode, but this command would run cypress in the default e2e mode, thus not record component tests:
Screen Shot 2022-05-10 at 6 01 56 PM

After - correct command is shown for component testing (it includes the `--component` flag:
Screen Shot 2022-05-10 at 5 59 46 PM

### How to test manually

Set up a project with no tests recorded in the dashboard and visit the runs page in the app, it should invite you to record your first run, and recommend the correct command for your testing type.

### PR Tasks


packages/app/cypress/e2e/runs.cy.ts
```diff
@@ -1,6 +1,35 @@
 import defaultMessages from '@packages/frontend-shared/src/locales/en-US.json'
 import type { SinonStub } from 'sinon'
 
+function scaffoldTestingTypeAndVisitRunsPage (testingType: 'e2e' | 'component') {
+  if (testingType === 'component') {
+    cy.scaffoldProject('component-tests')
```
ZachJW34, line 6: Can we use the `cypress-in-cypress` project? It has both testing types configured so we wouldn't have to create another system test project.

packages/app/src/runs/RunsEmpty.vue
```diff
@@ -58,7 +59,9 @@ const firstRecordKey = computed(() => {
     : '<record-key>'
 })
 const recordCommand = computed(() => {
-  return `cypress run --record --key ${firstRecordKey.value}`
+  const componentFlagOrSpace = props.gql.currentTestingType === 'component' ? ' --component ' : ' '
```
ZachJW34, line 62: You could do a `['arg1', 'arg2'].join(' ')` here if you wanted to to remove the space addition.

ZachJW34 (approved):
Few minor comments, not blocking

tbiethman (approved):
Code and percy snapshots look good 👍

packages/app/cypress/e2e/runs.cy.ts
marktnoonan: will merge when CI goes green

PR #21424, opened 2022-05-10
feat: distinguish app vs launchpad utm_source when using utm params

marktnoonan:
Closes https://cypress-io.atlassian.net/browse/UNIFY-1436

### User facing changelog

No user facing changes, but onlinks that previously had no `utm_source`, now correctly set a source of 'Binary: App' and 'Binary: Launchpad' depending on where they were clicked.

### Additional details

In 9.x the `utm_source` query param of `Test Runner` was added on the server as a part of `openExternal`. This PR removes that functionality completely from the server, and instead applies the same rule on the client side, in our `getUrlWithParams` helper function, to add the `utm_source` automatically to URLs that already include any other utm params.

Among other things, this makes the full length URL testable from the client side, and allows us to continue having well-formed links with accurate `href` attributes.

This PR covers all links in the docs menu and the growth prompt that's part of it, and any future links that use utm params will have the correct source appended automatically.

### How to test manually

Click any links in the docs menu in the launchpad and app and verify that when they take you to the docs pages, the utm params are present in the page URL and include the correct utm_source - `Binary: App` for the app and `Binary: Launchpad` for the browser.

### PR Tasks


tbiethman (approved):
Updated links look good, tested in app/launchpad. Couple things to note:

* I noticed that our changelog links don't have UTM params in 10.0, whereas they do in develop. For example: `https://docs.cypress.io/guides/references/changelog?source=dgui_footer&utm_medium=Footer&utm_campaign=Changelog&utm_source=Test%20Runner`. Not sure if that was intentionally scoped out, but we do launch those from both launchpad and app.
* This isn't new to your PR, but we have logic in our [openExternal](https://github.com/cypress-io/cypress/blob/04e5d70d48a1337b702f53ac9677350167eb7ae6/packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts#L96-L99) mutation that can also append query params. Seems like this could conflict with our [`getUrlWithParams`](https://github.com/cypress-io/cypress/blob/9ab0beea9ef19d370473f3cb16f25bc3679c2af6/packages/frontend-shared/src/utils/getUrlWithParams.ts#L19-L20) implementation if that arg is ever enabled, but it doesn't look like it is 🤷‍♂️ Would be worth removing if it's unused, especially if using it will result in bugs.

packages/frontend-shared/src/utils/getUrlWithParams.ts
```diff
@@ -5,6 +5,15 @@ export type LinkWithParams = {
 
 export const getUrlWithParams = (link: LinkWithParams) => {
   let result = link.url
+  const hasUtmParams = Object.keys(link.params).find((param) => param.startsWith('utm_'))
```
lmiller1990, line 8: Can use `some` to be more clear about the intention. When using `find` I usually expect the found variable is used, or an error is throw if it's not found. This is more "does it exist" in a true/false sense.

```suggestion
  const hasUtmParams = Object.keys(link.params).some((param) => param.startsWith('utm_'))
```

This is how Lodash's `findKey` works. https://github.com/lodash/lodash/blob/master/findKey.js#L23-L36

Or, if you want to be optimal and save a loop...

```js
function keyStartsWith (o, key) {
  for (const k in o) {
    if (k.startsWith(key)) return true
  }
  return false
}
```

I think `some` is the best candidate among `findKey`, `find` or the for loop.

packages/server/lib/gui/links.ts
```diff
@@ -1,29 +1,6 @@
...
-    if (_.find(opts.params, (_val, key) => _.includes(key, 'utm_'))) {
-      opts.params.utm_source = 'Test Runner'
-    }
-
-    url.search = new URLSearchParams(opts.params).toString()
-  }
-
-  return require('electron').shell.openExternal(url.href)
+export const openExternal = (url: string) => {
+  return require('electron').shell.openExternal(url)
```
lmiller1990, line 5: Any reason we inline the `require` instead of a top level `import`?

lmiller1990 (commented):
Two comments

packages/server/lib/gui/links.ts
tbiethman: That was changed as part of https://github.com/cypress-io/cypress/commit/cb5b547835ba5e58789b863589ced2212d2e8a13.

But good news, upgrading to electron 18 fixes this problem altogether and will allow that extra import to be removed. I'll get that taken out when https://github.com/cypress-io/cypress/pull/21418 lands and merges into 10.0-release.

packages/frontend-shared/src/utils/getUrlWithParams.ts
marktnoonan: oh that's awesome, I always forget about `some` but it's ideal here

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
```diff
@@ -93,8 +93,12 @@ export const mutation = mutationType({
       resolve: (_, args, ctx) => {
         let url = args.url
 
+        // the `port` param is included in external links to create a cloud organization
+        // so that the app can be notified when the org has been created
         if (args.includeGraphqlPort && process.env.CYPRESS_INTERNAL_GRAPHQL_PORT) {
-          url = `${args.url}?port=${process.env.CYPRESS_INTERNAL_GRAPHQL_PORT}`
+          const joinCharacter = args.url.includes('?') ? '&' : '?'
```
marktnoonan, line 99: @estrada9166 I made this small change, might as well future proof this for possible other params. @tbiethman I added a comment about the purpose of this, Alejandro showed me that it is actually used.

lmiller1990 (approved):
Seems fine to me - CI looks a bit red, but once that's fixed, we are good to go.

PR #21427, opened 2022-05-10
chore: fix windows tests in 10.0

ryanthemanuel:
### User facing changelog

n/a

### Additional details

Additional tests needed fixing in windows:

* The fuzzy spec search does not consistently use the windows separator. This PR ensures that we do.

### PR Tasks


packages/frontend-shared/src/utils/spec-utils.ts
```diff
@@ -4,8 +4,20 @@ import { ComputedRef, Ref, ref, watch } from 'vue'
 import _ from 'lodash'
 
 import type { UseCollapsibleTreeNode } from '../composables/useCollapsibleTree'
+import { decodeBase64Unicode } from './decodeBase64'
 
-const PATH_SEP = /[\/\\]/
+const getPlatform = () => {
+  // @ts-ignore
+  return window.Cypress?.platform || JSON.parse(decodeBase64Unicode(window.__CYPRESS_CONFIG__.base64Config)).platform
```
ryanthemanuel, line 11: This seems to be necessary since the specs list exists before Cypress necessarily exists

packages/frontend-shared/src/utils/spec-utils.ts
```diff
@@ -4,8 +4,11 @@ import { ComputedRef, Ref, ref, watch } from 'vue'
 import _ from 'lodash'
 
 import type { UseCollapsibleTreeNode } from '../composables/useCollapsibleTree'
+import { decodeBase64Unicode } from './decodeBase64'
 
-const PATH_SEP = /[\/\\]/
+const platform = JSON.parse(decodeBase64Unicode(window.__CYPRESS_CONFIG__.base64Config)).platform
```
lmiller1990, line 9: I don't think we should use `__CYPRESS_CONFIG__` in `frontend-shared`, it's not available in launchpad - it's `app` specific.

How about exporting this function and using it? https://github.com/cypress-io/cypress/blob/10.0-release/packages/app/src/runner/index.ts#L341-L343

packages/frontend-shared/src/utils/spec-utils.ts
```diff
@@ -4,8 +4,11 @@ import { ComputedRef, Ref, ref, watch } from 'vue'
...
+const regexSeparator = platform === 'win32' ? /\\/ : /\//
+const separator = platform === 'win32' ? '\\' : '/'
```
lmiller1990, line 11: Actually this entire file should not be in `frontend-shared`, it's only used in `app`. How about moving it?

lmiller1990 (changes requested):
Good fix, we move this entire file (and associated change) to `app`, and use the existing function to decode the __CYPRESS_CONFIG_ (see comment).

packages/app/src/specs/spec-utils.ts
```diff
@@ -2,10 +2,13 @@ import fuzzySort from 'fuzzysort'
...
 import { ComputedRef, Ref, ref, watch } from 'vue'
 import _ from 'lodash'
+import type { UseCollapsibleTreeNode } from '@packages/frontend-shared/src/composables/useCollapsibleTree'
+import { getRunnerConfigFromWindow } from '../runner'
 
-import type { UseCollapsibleTreeNode } from '../composables/useCollapsibleTree'
-
-const PATH_SEP = /[\/\\]/
+// @ts-ignore
+const platform = window.Cypress?.platform ?? getRunnerConfigFromWindow().platform
```
lmiller1990, line 9:
```suggestion
const platform = getRunnerConfigFromWindow().platform
```

I think - `window.Cypress` won't exist here, only on the runner page, but `__CYPRESS_CONFIG__` is defined from the very moment the app is opened (see here  https://github.com/cypress-io/cypress/blob/10.0-release/packages/data-context/src/sources/HtmlDataSource.ts#L119).

flotwig (approved):

ZachJW34 (approved):

lmiller1990:
Looks good... once CI is green, let's merge. I'll leave it to you 👍

lmiller1990 (approved):

AtofStryker (approved):

PR #21430, opened 2022-05-10
fix: ct testing support for node 17+

ZachJW34:
- Closes [UNIFY-1620](https://cypress-io.atlassian.net/browse/UNIFY-1620)

### User facing changelog
Fix ct apps using webpack-dev-server v4 not working for Node >= 17

### Additional details
Node 17+ changed how they resolve DNS names. Changing `localhost` to `127.0.0.1` is the fix for this issue

https://github.com/nodejs/node/issues/40537
https://github.com/Automattic/mongoose/issues/10917
https://github.com/chimurai/http-proxy-middleware/issues/705

To test this (using Node 17+), I had to make a few changes to the wherever we were building Webpack. Since we are using an older version of Webpack (< 5.6.1), we have to set `NODE_OPTIONS=--openssl-legacy-provider` for all Webpack build scripts. Without these changes, `yarn` and `yarn dev` will not work as it will fail with [digital envelope routines::unsupported](https://github.com/webpack/webpack/issues/14532)

CT works for Node 17+

https://user-images.githubusercontent.com/25158820/167738370-80bc8b89-a0af-41da-b623-67e82f012b1c.mov

### PR Tasks


scripts/gulp/tasks/gulpWebpack.ts
```diff
@@ -16,10 +19,27 @@ type RunWebpackCfg = {
...
+// Node 17+ ships with OpenSSL 3 by default, so we may need the option
+// --openssl-legacy-provider so that webpack@4 can use the legacy MD4 hash
+// function. This option doesn't exist on Node <17 or when it is built
+// against OpenSSL 1, so we have to detect Node's major version and check
+// which version of OpenSSL it was built against before spawning the process.
+//
+// Can be removed once the webpack version is upgraded to >= 5.61,
+// which no longer relies on Node's builtin crypto.hash function.
+function useLegacyOpenSSLProvider (env: Env) {
+  if (process.versions && semver.satisfies(process.versions.node, '>=17.0.0') && process.versions.openssl.startsWith('3.')) {
```
lmiller1990, line 36: Could use `semver` here, something like `semver.satisifies(prices.versions, '>=3')`.

scripts/use-legacy-openssl.js
```diff
@@ -0,0 +1,20 @@
...
+// which version of OpenSSL it was built against before spawning the process.
+//
+// Can be removed once the webpack version is upgraded to >= 5.61,
+// which no longer relies on Node's builtin crypto.hash function.
+
+const semver = require('semver')
+
+let opts = process.env.NODE_OPTIONS || ''
+
+if (process.versions && semver.satisfies(process.versions.node, '>=17.0.0') && process.versions.openssl.startsWith('3.')) {
```
lmiller1990, line 15: Same here, in fact already using it for `process.versions.node`, why not `process.versions.openssl`?

scripts/use-legacy-openssl.js
```diff
@@ -0,0 +1,20 @@
...
+  opts = `${opts} --openssl-legacy-provider`
+}
+
+// eslint-disable-next-line no-console
+console.log(opts)
```
lmiller1990, line 20: Is this `console.log` intentional/required? Should it be wrapped in a `if (process.env.NODE_VERSION !== 'production')`?

lmiller1990 (commented):
Some general comments 👍

scripts/gulp/tasks/gulpWebpack.ts
ZachJW34: Works with a tweak since Node 17.0.0 uses openssl=`3.0.0-quic` which is considered a pre-release so I've included `{ includePrerelease: true }`.

npm/webpack-dev-server/src/createWebpackDevServer.ts
```diff
@@ -68,7 +68,7 @@ function webpackDevServer4 (
   const { devServerConfig: { cypressConfig: { devServerPublicPathRoute } } } = config
   const WebpackDevServer = config.sourceWebpackModulesResult.webpackDevServer.module
   const webpackDevServerConfig = {
-    host: 'localhost',
+    host: '127.0.0.1',
```
flotwig, line 71: In general, we should never use `'localhost'` in the App due to issues like this. We've even had users with an incorrect `localhost` entry in `/etc/hosts`, which causes mind boggling issues.

packages/extension/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "main": "index.js",
   "scripts": {
-    "build": "gulp build",
+    "build": "cross-env NODE_OPTIONS=\"$(node ../../scripts/use-legacy-openssl.js)\" gulp build",
```
flotwig, line 7: Does this work in Windows? IDK how the command substitution is handled cross-platform

flotwig (changes requested):
@ZachJW34 We already have some tests that validate running `--e2e` in Node versions (including 17.3.0), I think it would be a good idea to add a test for CT as well: https://github.com/cypress-io/cypress/blob/b0c8db34348be31f32795f3f67c3408403d59ace/system-tests/test-binary/node_versions_spec.ts

Otherwise I can see this issue cropping back up. Let me know if I can assist with this, the Docker test experience on Mac is not great, you may want to create a Linux VM.

packages/extension/package.json
ZachJW34: Good point I'll give it a test!

npm/webpack-dev-server/src/createWebpackDevServer.ts
ZachJW34: I was absolutely baffled as to what was going on until I stumbled upon the issues I listed above.

ZachJW34:
@flotwig I'll see how far I can get adding the tests and if I get stuck I'll reach out!

ZachJW34:
@flotwig Added binary system tests, tried it out on WSL which was painful but I eventually got it working

packages/extension/package.json
ZachJW34: Didn't work on Windows, changed to running webpack in a child process so the env variables could be manipulated more easily

lmiller1990 (approved):

system-tests/test-binary/node_versions_spec.ts
```diff
@@ -9,6 +9,15 @@ function smokeTestDockerImage (dockerImage: string) {
     specDir: 'tests',
     project: 'todos',
   })
+
+  systemTests.it(`can run in ${dockerImage}`, {
```
flotwig, line 13: clean up the hierarchy of tests here? right now it's: 

```
e2e binary node versions > can run in ${dockerImage}
e2e binary node versions > can run in ${dockerImage}
```

but something like this would be better:

```
binary node versions > ct > can run in ${dockerImage}
binary node versions > e2e > can run in ${dockerImage}

ZachJW34: Good catch, addressed in [0654882](https://github.com/cypress-io/cypress/pull/21430/commits/06548825f43c6878abe43d6ff06997d6c8f2b5a0)

flotwig (approved):

PR #21437, opened 2022-05-11
fix: remove breaking options from testing type on migration

estrada9166:
- Closes  [UNIFY-1526](https://cypress-io.atlassian.net/jira/software/projects/UNIFY/boards/20?assignee=5ea883deddd45f0b94f5d98d&selectedIssue=UNIFY-1526)

### User facing changelog

We were removing some breaking options from the root level on the migrated config but, we were not removing those from the testing type config, with this change, we are removing every possible breaking option from the nested testing type config.

### PR Tasks


lmiller1990 (approved):

BlueWinds (changes requested):
Changes look good. `packages/data-context/test/unit/sources/migration/codegen.spec.ts` has a bunch of unit tests covering this function, can add a new one to cover this case.

BlueWinds (approved):

PR #21442, opened 2022-05-11
fix: handle errors on plugins migration child process

estrada9166:
- Closes  [UNIFY-1744](https://cypress-io.atlassian.net/browse/UNIFY-1744)

### User facing changelog

When the plugins file for migration, had an incorrect path or an error, we were not handling correctly the errors and it was affecting the main process, with this change we handle the child process error without affecting the Cypress process

image

image

### PR Tasks


lmiller1990 (approved):
Looks good, we pair-reviewed this.

tgriesser (approved):

PR #21462, opened 2022-05-13
fix: do not include non specs when globbing using --spec in run mode

lmiller1990:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1755

### User facing changelog

Fix a bug where non-spec files are included in run mode when using `--spec` with a glob.

### Additional details

Previously, we allowed to glob by anything - even things outside of the defined `specPattern`. This is incorrect - we only want to let things run that are within the `specPattern`. This is how 9.x functions. It also prevents situations where Cypress hangs, or gives a strange error, when trying to run something invalid, like an `.mp4` or other non-spec file.

After making this change, many system tests failed, since they relied on the previous, incorrect behaviour of running things outside of the `specPattern`, which is now  uses the `cy.js` (some were still using `spec.js` extension). I updated all the spec files and the associated snapshots to use the new `cy.js` pattern.

I also unskipped a test in `system_node_spec.js`, which was skipped for no obvious reason.

## Before 

Finds three files (two of which are non-specs when using --spec with glob). It finds `timers.cy.js.mov, timers.ts, timers.cy.js`.

```
$ yarn cypress:run --spec "**/timers*"                            lmiller1990/UNIFY-1755

=========================================================================================

  (Run Starting)

  │ Cypress:        9.6.1                                                                          │
  │ Browser:        Electron 94 (headless)                                                         │
  │ Node Version:   v16.14.2 (/Users/lachlan/.nvm/versions/node/v16.14.2/bin/node)                 │
  │ Specs:          3 found (timers.cy.js.mov, timers.ts, timers.cy.js)                            │
  │ Searched:       **/timers*                                                                     │


```

## After (finds one spec - same as 9.x)

```
$ yarn cypress:run --spec "**/timers*"                            lmiller1990/UNIFY-1755

=========================================================================================

  (Run Starting)

  │ Cypress:        9.6.1                                                                          │
  │ Browser:        Electron 94 (headless)                                                         │
  │ Node Version:   v16.14.2 (/Users/lachlan/.nvm/versions/node/v16.14.2/bin/node)                 │
  │ Specs:          1 found (timers.cy.js)                                                         │
  │ Searched:       **/timers*                                                                     │


```

### PR Tasks


packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -210,21 +211,45 @@ export class ProjectDataSource {
...
 
-  async findSpecs (
-    projectRoot: string,
-    testingType: Cypress.TestingType,
-    specPattern: string[],
-    excludeSpecPattern: string[],
-    globToRemove: string[],
-  ): Promise<FoundSpec[]> {
-    const specAbsolutePaths = await this.ctx.file.getFilesByGlob(
+  async findSpecs ({
```
ZachJW34, line 214: Is all of this info derivable? Seems like the info either comes from the `config` or `ctx` but every caller has to repeat the same steps to grab all this info when we could be doing it here. Not that many places we call this though

packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
```diff
@@ -143,46 +141,95 @@ describe('findSpecs', () => {
...
+    await ctx.fs.ensureDir(projectRoot)
+    await Promise.all(fixture.map((element) => ctx.fs.outputFile(path.join(projectRoot, element), '')))
   })
 
   afterEach(async () => {
-    await ctx.fs.remove(temporary)
+    await ctx.fs.remove(projectRoot)
+  })
+
+  it('exludes specs outside `specPattern`, even if passing a generic glob', async () => {
```
ZachJW34, line 159: Would adding an e2e test covering this behavior be worth it? I feel like this behavior is pretty important and having more coverage around it would help since we don't have a test that covers the whole flow (starting at command line `--spec <specPattern>`).

ZachJW34 (approved):
Tested and works great, a few comments but nothing non-blocking

packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
lmiller1990: Sure, we can probably get it "for free" from a system test using `spec: '**/*` or something like this. I'll add it.

flotwig (approved):
Tested and this fixes the issue I was having. Thanks for jumping on this @lmiller1990, this would have been a bad one to miss.

packages/data-context/src/sources/ProjectDataSource.ts
lmiller1990: @ZachJW34 looks into this but it doesn't seem like a good idea. If we try to derive it during the unit test, we need to spin up a full `configManager`, and all the overhead that comes with it. The alternative is a "fake" config manager... But then, what are we accomplishing? It's not really any more coverage than what we have here, but a lot more overhead.

I agree this forwarding of options is not really ideal, though. The main problem is we do not normalise the specPatterns from `config` internally, so we need to check if it's either `string | string[] | undefined` everywhere. Ultimately, if we can rework the config layer to normalise everything, it'll be a lot cleaner. Eg - anything that is defined as `string | string[] | undefined` needs to become an array (empty representing `undefined`).

We also should have a full config we can use for unit tests - but at that point, I'd argue an e2e or integration test is better, anyway, like you suggested below.

packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
lmiller1990: Ok, I added it as an integration test in `cypress_spec.js`. https://github.com/cypress-io/cypress/pull/21462/commits/0b55b3146d2bde62c78ca0813ad0deb9d52893fb#diff-0416ec0277f3a43b360ddb7f3b56280f1ab1851592bc36e90fcffe74ee571ee8R396-R399

Notice I added a `qux.mp4` but we only call open browser once - that's because we detect `qux.cy.js`, but not the invalid `qux.mp4`.

PR #21490, opened 2022-05-13
fix: use resolved port for vite

ZachJW34:
- Closes [UNIFY-1621](https://cypress-io.atlassian.net/browse/UNIFY-1621)

### User facing changelog
Fix vite ct tests hanging when default port 3000 is in use

### Additional details
This PR let's Vite find a server that's not in use. Vite will choose another port if the default port (3000) is in use, so we should derive the port from the created server rather than doing it ourselves.

Before:

https://user-images.githubusercontent.com/25158820/168335860-6af739bc-9c14-4e3b-93dc-30ed4b635f9a.mov

After:

https://user-images.githubusercontent.com/25158820/168335946-02ff3c78-43e0-42ae-85b7-84eaf1642453.mov

### PR Tasks


lmiller1990 (approved):

npm/vite-dev-server/src/devServer.ts
```diff
@@ -24,10 +23,9 @@ export async function devServer (config: ViteDevServerConfig): Promise<Cypress.R
   const server = await devServer.create(config, vite)
 
   debug('Vite server created')
-  const port = await getPort({ port: 3000 })
 
-  debug('Starting Vite Server on port %i', port)
-  await server.listen(port)
+  await server.listen()
+  const port = server.config.server.port!
```
flotwig, line 28: Kinda minor, but the Vite docs seem ambiguous about if port exists when you haven't specified a port to `server.listen()`:

```js
    /**
     * Specify server port. Note if the port is already being used, Vite will
     * automatically try the next available port so this may not be the actual
     * port the server ends up listening on.
     */
    port?: number;
```

So maybe it would be a good idea to guard against `undefined`, just in case the implementation changes and this is no longer available immediately after `await server.listen()`.

```suggestion
  const port = server.config.server.port!

  if (!port) throw new Error('Missing vite dev server port.')
```

flotwig (approved):
Nice, getPort is something we should avoid when we can!

npm/vite-dev-server/src/devServer.ts
lmiller1990: I like this, I think we generally error loud and clear rather than assuming something that's absolutely required will always exist. Better debugging and bug reports!

ZachJW34: Done in [f4d92cb](https://github.com/cypress-io/cypress/pull/21490/commits/f4d92cbc1b723842a71176717fbbff1bc253aa2a)!

PR #21500, opened 2022-05-13
fix: display cy.mount command log

estrada9166:
- Closes  [UNIFY-1671](https://cypress-io.atlassian.net/browse/UNIFY-1671)

### User facing changelog

Update vue2 to log the mount command, it was previously not being logged as we do for the different frameworks

### PR Tasks


npm/vue2/src/index.ts
```diff
@@ -290,6 +292,30 @@ const injectStyles = (options: StyleOptions) => {
...
+  if (componentOptions.name) {
+    return componentOptions.name
+  }
+
+  if (componentOptions.__file) {
+    const filepathSplit = componentOptions.__file.split('/')
+    const fileName = filepathSplit[filepathSplit.length - 1]
+
+    // remove the extension .js, .ts or .vue from the filename to get the name of the component
+    const baseFileName = fileName.replace(/\.(js|ts|vue)?$/, '')
```
lmiller1990, line 310: If `filepathSplit` does not have a `/`, `fileName` will be `undefined`, then we will get "Cannot call replace on undefined" here.

I think we need

```js
const fileName = filepathSplit[filepathSplit.length - 1] ?? DEFAULT_COMP_NAME
```

lmiller1990 (changes requested):

ZachJW34 (approved):
Looks good!

tbiethman (approved):

lmiller1990 (approved):

PR #21503, opened 2022-05-14
elevatebart: use design system windicss config

elevatebart:
- Closes 

### PR Tasks


packages/app/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "dependencies": {},
   "devDependencies": {
+    "@cypress-design/css": "0.1.0",
     "@graphql-typed-document-node/core": "^3.1.0",
```
lmiller1990, line 23: Not directly related but do you think we should consider changing to TailwindCSS at this point for future sustainability? It has way more contributors, and just looking at the repo we have 18 issues (Tailwind) vs 172(Windi). 

I feel like Tailwind will likely outlast and be a more sustainability option. I have no idea of the cost of migration - should be small, since Windi is "Tailwind compatible". 

I don't see a strong benefit to Windi since Tailwind now has a JIT compiler, which was the main issue Windi solved (speed).

Thoughts?

elevatebart: Unless there is a compelling reason (bug, needed enhancement), I would prefer to avoid unnecessary changes.
Tailwind is great and really active, but this does not justify yet the potential fixing of the problems.

Among the differences, TailwindCSS is a PostCSS-based plugin instead of being independent in WindiCSS. It means that in WindiCSS the plugin controls the configuration. With TailwindCSS, we would have to export a TailwindCSS config.

There are a few configurations as well that would need to migrate. Colors, safelist, plugins.

packages/app/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "dependencies": {},
   "devDependencies": {
```
lmiller1990, line 21: Just learned that https://github.com/cypress-io/cypress-design exists!

lmiller1990 (approved):
Seems fine - just a bunch of config.

It looks like some Percy screenshots are different, but they don't appear to be related to the style/color changes you've made in this PR.

ZachJW34 (approved):

jennifer-shehane (changes requested):
@elevatebart Update the base branch to be against develop before merging.

elevatebart:
@rachelruderman nope just on a green CI

PR #16247, opened 2021-04-28
feat(webpack-dev-server): support custom index-template for webpack dev server

lmiller1990:
- Closes #16243

### User facing changelog

Support passing a custom `index-template.html` to the Webpack Dev Server `startDevServer` function.

### Additional details

Some frameworks like Next.js expect certain HTML tags to exist in the document, [as explained here](https://github.com/cypress-io/cypress/issues/16243#issuecomment-828121505). Specifically an element with the id `__next_css__DO_NOT_USE__` should exist when the component is imported. This is used for CSS modules by `style-loader`. This PR uses the ability to pass a custom `index-template.html` to provide a modified `index.html` when using the  Next.js plugin.

This PR has a secondary use of letting users include things like CDN links in their default template, like Storybook's `preview-head.html`. They'll also need to include `<div id="__cy_root">`, though. Asking them to include this tag is not really ideal, so I'm not really advertising this feature at the moment, other than a small note in the Webpack Dev Server README.

Long term, I think we should take a custom `index.html` and inject `__cy_root` programatically, so the user does not need to think about this sort of thing when providing a custom template. This is a good solution for now, and gets us closer to bridge the gap between Storybook as far as "Cypress as a Design System".

### How has the user experience changed?

CSS modules (default for Next.js) work out of the box.

### PR Tasks

- [x] Have tests been added/updated?
- [ ] Has the original issue or this PR been tagged with a release in ZenHub? 
- [ ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [ ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [ ] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

PR #21507, opened 2022-05-16
chore: document next.js specific component template

lmiller1990:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1608

### User facing changelog

Document Next.js specific selector injected into `component-index.html`.

### Additional details

Prior context: https://github.com/cypress-io/cypress/pull/16247 

Basically, Next.js hard-codes a specific DOM node and expects it to exist for injecting CSS via style-loader. You can see the line of code here: https://github.com/vercel/next.js/blob/5f3351dbb8de71bcdbc91d869c04bc862a25da5f/packages/next/build/webpack/config/blocks/css/loaders/client.ts#L24

We need this too, so Next.js components render the same as the user expects when using Cypress Component Testing.

### PR Tasks


packages/scaffold-config/package.json
```diff
@@ -9,6 +9,7 @@
     "check-ts": "tsc --noEmit",
     "clean": "rimraf ./src/*.js ./src/**/*.js ./src/**/**/*.js ./test/**/*.js || echo 'cleaned'",
     "clean-deps": "rimraf node_modules",
+    "test": "yarn test-unit",
```
lmiller1990, line 12: These were not running on CI because this was missing so I added it.

system-tests/project-fixtures/next/cypress/support/component-index.html
```diff
@@ -5,6 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
lmiller1990, line 5: Above file packages/scaffold-config/test/unit/frameworks.spec.ts was obsolete (code was moved to `data-context` and tests added there) so I deleted it. The reason we did not notice was because `scaffold-config` was not running on CI (now it is as of this PR).

marktnoonan (changes requested):
This looks good and will help users avoid breaking the index html. It's especially good to get those unit tests running in CI!

One change request - one of the new changes is causing E2E failures in scaffolding component tests, I think due to including the `projectRoot` in the glob? But not too familiar with this.

Unrelated: I noticed when I got in that spec is that there's a TODO [here](https://github.com/cypress-io/cypress/blob/1289b01947a28a77048114aa63cbcd62d9ff52ee/packages/launchpad/cypress/e2e/scaffold-component-testing.cy.ts#L11-L12) where the PR that was referred to has been merged, should we go on and make a ticket for that TODO?

packages/scaffold-config/src/frameworks.ts
```diff
@@ -131,7 +131,14 @@ export const WIZARD_FRAMEWORKS = [
...
     supportStatus: 'alpha',
-    componentIndexHtml: componentIndexHtmlGenerator('<div id="__next_css__DO_NOT_USE__"></div>'),
+    /**
+     * Next.js uses style-loader to inject CSS and requires this element to exist in the HTML.
+     * @see: https://github.com/vercel/next.js/blob/5f3351dbb8de71bcdbc91d869c04bc862a25da5f/packages/next/build/webpack/config/blocks/css/loaders/client.ts#L24
+     */
+    componentIndexHtml: componentIndexHtmlGenerator([
+      `<!-- Used by Next.js to inject CSS. -->\n`,
+      `<div id="__next_css__DO_NOT_USE__"></div>`,
+    ].join(' '.repeat(8))),
```
ZachJW34, line 141: It would make more sense to have this logic in `componentIndexHtmlGenerator` since it is aware of the proper spacing required to produce a formatted html, but this is the only use case so not a big deal.

ZachJW34 (approved):
Code changes make sense, going to approve so once they are fixed up you can merge

packages/scaffold-config/src/frameworks.ts
lmiller1990: Not sure I agree here - I think `componentIndexHtmlGenerator` should not be too smart, and we should be injecting the additional markup from the outside. I don't think it needs to know about the different frameworks we support.

I tried to do something like:

```js
componentIndexHtml: dedent`
  <!-- stuff -->
`
```

But it won't work, since you end up with a `dedent` string inside a `dedent` string, and it doesn't format as you'd expect. This was the only way I could use the injection API that was implemented by the original developer and maintain the spacing. Not ideal, but it is what it is 🤷

packages/scaffold-config/src/detect.ts
```diff
@@ -135,7 +135,7 @@ export function detectLanguage (projectRoot: string, pkgJson: PkgJson): 'js' | '
     joinPosix(projectRoot, 'cypress', '**/*.{ts,tsx}'),
   ]
 
-  const tsFiles = globby.sync(globs, { onlyFiles: true, gitignore: true })
+  const tsFiles = globby.sync(globs, { onlyFiles: true, gitignore: true, cwd: projectRoot })
```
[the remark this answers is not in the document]
lmiller1990: Hm good catch - I needed `cwd` for the tests (that were not running) to pass. I'll try your suggestion of removing `projectRoot` from the globs - I think that makes more sense anyway, we should be able to assume that cypress is running in the projectRoot.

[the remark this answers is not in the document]
lmiller1990: Fixed.

lmiller1990:
> This looks good and will help users avoid breaking the index html. It's especially good to get those unit tests running in CI!
> 
> One change request - one of the new changes is causing E2E failures in scaffolding component tests, I think due to including the `projectRoot` in the glob? But not too familiar with this.
> 
> Unrelated: I noticed when I got in that spec is that there's a TODO [here](https://github.com/cypress-io/cypress/blob/1289b01947a28a77048114aa63cbcd62d9ff52ee/packages/launchpad/cypress/e2e/scaffold-component-testing.cy.ts#L11-L12) where the PR that was referred to has been merged, should we go on and make a ticket for that TODO?

I fixed the tests 👍 

Not sure about the TODO, we could, or it might be already captured in some other tests. Likely a low priority.

packages/scaffold-config/src/detect.ts
```diff
@@ -131,11 +131,11 @@ export function detectLanguage (projectRoot: string, pkgJson: PkgJson): 'js' | '
...
 
   const globs = [
-    joinPosix(projectRoot, '**/*tsconfig.json'),
-    joinPosix(projectRoot, 'cypress', '**/*.{ts,tsx}'),
+    joinPosix('**/*tsconfig.json'),
+    joinPosix('cypress', '**/*.{ts,tsx}'),
   ]
 
-  const tsFiles = globby.sync(globs, { onlyFiles: true, gitignore: true })
+  const tsFiles = globby.sync(globs, { onlyFiles: true, gitignore: true, cwd: projectRoot, ignore: ['node_modules'] })
```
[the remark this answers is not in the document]
lmiller1990: I think we can just use the `async` call of `globby` to avoid EMFILE errors. Unfortunately, making this function as `async` has a huge cascading effect requiring 10s of files to be changed. We end up calling a function (that now needs to be `awaited` in a few constructors, too, which makes it more painful, since a constructor cannot be `async`, the design and flow of some parts of the system would need to be reconsidered.

We should still make this change, it's just not as trivial as "use await, mark async".

PR #21518, opened 2022-05-16
fix: check the error constructor before using

### LINKED ISSUE #21428: Tests finish without running all commands if Error is not a constructor
### Current behavior

I have noticed this rare but unfortunate behavior, same was noticed by other people on my team.
Sometimes a test passes, but when you look at it, it is missing lots of commands, as if it exited early. 
For example the test below should have a lot more commands, yet it has finished already somehow

err1

When poking around, I noticed the Devtools is showing an error that is NOT reflected in the Command Log "ErrorConstructor is not a constructor" Seems somehow the `specWidow.Error` can be undefined, which causes a problem when we grab the user stack

err2

The error can be thrown only inside

```js
const captureUserInvocationStack = (ErrorConstructor, userInvocationStack) => {
  if (!userInvocationStack) {
    const newErr = new ErrorConstructor('userInvocationStack');
    userInvocationStack = newErr.stack; // if browser natively supports Error.captureStackTrace, use it (chrome) (must be bound)
    // otherwise use our polyfill on top.Error

    const captureStackTrace = ErrorConstructor.captureStackTrace ? ErrorConstructor.captureStackTrace.bind(ErrorConstructor) : Error.captureStackTrace;
    captureStackTrace(newErr, captureUserInvocationStack); // On Chrome 99+, captureStackTrace strips away the whole stack,
    // leaving nothing beyond the error message. If we get back a single line
    // (just the error message with no stack trace), then use the original value
    // instead of the trimmed one.

    if (newErr.stack.match('\n')) {
      userInvocationStack = newErr.stack;
    }
  }

  userInvocationStack = normalizedUserInvocationStack(userInvocationStack);
  return userInvocationStack;
};
```

Which receives the `ErrorConstructor` from its callers, but does not check if it is defined or a function before calling `new ...`

A few more places from the stack that call this function
err3
err4

A conditional breakpoint worked to stop there on undefined `Error` argument
err5

err7

### Desired behavior

This is very rare but happens and is very concerning. Especially since it lowers the confidence in the passing tests. 
What I would like to do is a simple guard: if there is no `ErrorConstructor`, do not try to use it. Having no stack is preferred to skipping the rest of the test with PASS status.

### Test code to reproduce

Really don't know, since this function is called a lot at the start of the test. 
I will try to trace it better, but a little defensive programming would not hurt still

### Cypress Version

I have seen it in v8 and v9

### Other

_No response_

bahmutov:
- Closes #21428

### User facing changelog

Very rarely the test runner has an internal crash trying to use `new specWindow.Error(...)` while that error constructor is unavailable. With that crash, the current test finishes early without running all test commands. This change guards against it and ensures that the test completes

### Additional details

I used the guard in my local version of Cypress and it seems to be harmless, yet fixes the early exit

### PR Tasks

- [x] Have tests been added/updated? I added a simple sanity test, but nothing full, since it is hard to recreate the conditions for the error (see the original issue)


jennifer-shehane:
Thanks @bahmutov, we'll take a look at reviewing as soon as we can.

flotwig (changes requested):
Hey @bahmutov, nice fix. It seems to be causing some test failures, once those are resolved this LGTM.

Also HIIII! :smile:

bahmutov:
ughh, the tests ... are not really testing this spec window error, are they

rachelruderman:
Unfortunately we have to close this issue due to inactivity. Please comment if there are passing tests and we can reopen.

PR #21519, opened 2022-05-16
fix: update dropdown shadow

estrada9166:
- Closes  [UNIFY-1490](https://cypress-io.atlassian.net/browse/UNIFY-1490)

### Additional details

The Figma designs are `0px 1px 3px rgba(0, 0, 0, 0.15);` but the current change is `0px 1px 3px rgba(0, 0, 0, 0.2)` - It needs to be a little bit different because the colors on Figma seems to be different based on the contrast of colors of the app (Figma)

## Before
image
image

## After
image
image

### PR Tasks


tbiethman (approved):
I think this is contrasting much better than before. Can we log a follow-up issue with design to synchronize the value used?

marktnoonan (approved):

PR #21527, opened 2022-05-17
feat(app): use longest common prefix for default spec filename

flotwig:
- Closes UNIFY-1636

### User facing changelog

- The longest common prefix of spec filenames has been added as a fallback for default spec filename generation.

### Additional details

- Adds unit tests for new functionality and integration tests around `defaultSpecFileName`.
- Fixes a small bug with the `getFilenameFromSpecPattern`: https://github.com/cypress-io/cypress/pull/21527#discussion_r874333679
- See Jira and Slack thread for rationale and details.
- Test failures match those in `10.0-release`.

Default filename generation is more likely to be what the user is looking for. Updated five-step algorithm:

https://github.com/cypress-io/cypress/blob/43edac62d2e53a5785b54b7ca7c35e454fd06931/packages/data-context/src/sources/ProjectDataSource.ts#L327-L348

### PR Tasks


packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -113,36 +142,34 @@ export function getDefaultSpecFileName (specPattern: string, testingType: Testin
...
-    dirname = dirname.replace('**', 'cypress')
-  }
+  // If a spec can be in any root dir, go ahead and use "cypress/"
+  if (dirname.startsWith('**')) dirname = dirname.replace('**', 'cypress')
 
   const splittedDirname = dirname.split('/').filter((s) => s !== '**').map((x) => replaceWildCard(x, testingType)).join('/')
   const fileName = replaceWildCard(parsedGlob.path.filename, 'filename')
 
   const extnameWithoutExt = parsedGlob.path.extname.replace(parsedGlob.path.ext, '')
+    || `.cy.${fileExtensionToUse}`
```
flotwig, line 155: I noticed a bug with a spec pattern like `cypress/component-tests/**/*` (no extension), this seemed to be the fix.

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -312,40 +339,48 @@ export class ProjectDataSource {
...
-  async defaultSpecFileName () {
-    const getDefaultFileName = (testingType: TestingType) => `cypress/${testingType}/${testingType === 'e2e' ? 'spec' : 'ComponentName'}.cy.${this.ctx.lifecycleManager.fileExtensionToUse}`
+  async defaultSpecFileName (): Promise<string | null> {
+    const defaultFilename = `${this.ctx.coreData.currentTestingType === 'e2e' ? 'spec' : 'ComponentName'}.cy.${this.ctx.lifecycleManager.fileExtensionToUse}`
+    const defaultPathname = path.join('cypress', this.ctx.coreData.currentTestingType ?? 'e2e', defaultFilename)
 
     try {
       if (!this.ctx.currentProject || !this.ctx.coreData.currentTestingType) {
-        return getDefaultFileName('e2e')
+        return null
```
tbiethman, line 348: defaultSpecFileName is nonNull in the schema now. I think throwing an error here would be appropriate, if we don't want to just return the e2e default like below.

tgriesser: Or we can change `defaultSpecFileName` to be nullable in the schema, either way

packages/data-context/src/sources/ProjectDataSource.ts
tgriesser: > I noticed a bug with a spec pattern like cypress/component-tests/**/* (no extension), this seemed to be the fix.

What was the bug exactly?

tgriesser (changes requested):
Agree w/ @tbiethman - we shouldn't return `null` for a `nonNull` field or that'll break the GraphQL query, we either need to give it a string return or change the GraphQL field to remove the `nonNull` constraint

packages/data-context/src/sources/ProjectDataSource.ts
tbiethman: This was recently changed to be nonNull to remove some of the guards/duplicated default logic that was present in app. Agreed it's not a big deal either way tho, we could default to an empty string in the app and have a similar result.

flotwig: Whoops, fixed

packages/data-context/src/sources/ProjectDataSource.ts
flotwig: Check out this test:

https://github.com/cypress-io/cypress/blob/34fa8cbf5f9308bd7438df24b5c4bb0da423e60c/packages/data-context/test/unit/sources/ProjectDataSource.spec.ts#L437-L441

Before this change and the other change to remove double-slashes, the result here would be `cypress/component-tests//filename`. I assumed we don't want to give people a default filename with no extension or double slashes.

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -312,40 +339,48 @@ export class ProjectDataSource {
...
-  async defaultSpecFileName () {
-    const getDefaultFileName = (testingType: TestingType) => `cypress/${testingType}/${testingType === 'e2e' ? 'spec' : 'ComponentName'}.cy.${this.ctx.lifecycleManager.fileExtensionToUse}`
+  async defaultSpecFileName (): Promise<string> {
+    const defaultFilename = `${this.ctx.coreData.currentTestingType === 'e2e' ? 'spec' : 'ComponentName'}.cy.${this.ctx.lifecycleManager.fileExtensionToUse}`
+    const defaultPathname = path.join('cypress', this.ctx.coreData.currentTestingType ?? 'e2e', defaultFilename)
 
     try {
       if (!this.ctx.currentProject || !this.ctx.coreData.currentTestingType) {
-        return getDefaultFileName('e2e')
+        throw new Error('Failed to get default spec filename, missing currentProject/currentTestingType')
```
tbiethman, line 348: Do you want to move this outside of the try/catch? We could cut out the middle man and just return defaultPathname otherwise.

tgriesser (changes requested):

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -339,13 +339,13 @@ export class ProjectDataSource {
...
 
-  async defaultSpecFileName (): Promise<string | null> {
+  async defaultSpecFileName (): Promise<string> {
     const defaultFilename = `${this.ctx.coreData.currentTestingType === 'e2e' ? 'spec' : 'ComponentName'}.cy.${this.ctx.lifecycleManager.fileExtensionToUse}`
     const defaultPathname = path.join('cypress', this.ctx.coreData.currentTestingType ?? 'e2e', defaultFilename)
 
     try {
       if (!this.ctx.currentProject || !this.ctx.coreData.currentTestingType) {
-        return null
+        throw new Error('Failed to get default spec filename, missing currentProject/currentTestingType')
```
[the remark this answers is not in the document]
flotwig: > it's legal to reference defaultSpecFileName in a GraphQL query before you've chosen a testing type

That's why I thought it made more sense to throw - why would we want to return a value here when it has no valid meaning? With throwing, tests fail in an obvious way if we start doing something silly like requesting `defaultSpecFileName` before selecting a project/testingType. With a `null` or `''`, tests will possibly silently pass and we will build our codebase upon a silly thing like requesting `defaultSpecFileName` before selecting a project/testingType.

packages/data-context/src/sources/ProjectDataSource.ts
flotwig: oh woof I didn't even notice that this is the same thing

packages/graphql/src/schemaTypes/objectTypes/gql-CurrentProject.ts
```diff
@@ -135,7 +135,8 @@ export const CurrentProject = objectType({
       },
     })
 
-    t.nonNull.string('defaultSpecFileName', {
+    // nullable so we can throw if it can't be decided
+    t.string('defaultSpecFileName', {
       description: 'Default spec file name for spec creation',
```
tgriesser, line 140:
```suggestion
    t.string('defaultSpecFileName', {
      description: 'Default spec file name for spec creation, nullable / throws if it can't be decided',
```

tgriesser (changes requested):

packages/graphql/src/schemaTypes/objectTypes/gql-CurrentProject.ts
flotwig: hahahaha

flotwig: unterminated string literal

tbiethman (approved):

PR #21536, opened 2022-05-17
fix: Spec List: Added tooltips to git icons created/modified

MuazOthman:
- Closes [CLOUD-576](https://cypress-io.atlassian.net/browse/CLOUD-576)

### User facing changelog

- On Spec list page, show tooltips for git status icons when created/modified for clarity
- On Spec list page, address the case when no git information is available to show no icons and file update time from file system

image
image

### PR Tasks


marktnoonan:
The description says "na" for "Have tests been added/updated?", but this seems like some new behavior that should be covered in a test. Are there existing tests that already cover these tooltips?

ZachJW34:
I'm seeing some weird behavior where the tooltip won't show up after the icon changes. Not sure if it is this PR that introduces this or not

https://user-images.githubusercontent.com/25158820/168901984-54b14bb4-56d5-4eb5-9af1-e2e72692e551.mov

packages/app/src/specs/SpecListGitInfo.vue
```diff
@@ -6,22 +6,23 @@
...
       <component
         :is="classes.icon"
         :class="classes.iconClasses"
       />
       <template
-        v-if="classes.showTooltip"
         #popper
       >
         <div>
           <p class="max-w-sm text-sm truncate overflow-hidden">
```
emilyrohrbough, line 18: Can we conditionally render this if there is a tooltip message to show?

packages/app/src/specs/SpecListGitInfo.vue
```diff
@@ -6,22 +6,23 @@
...
       placement="top"
       class="h-full grid items-center"
-      :disabled="!classes.showTooltip"
     >
       <component
         :is="classes.icon"
         :class="classes.iconClasses"
       />
       <template
-        v-if="classes.showTooltip"
```
emilyrohrbough, line 16: Can we conditionally render this if there is tooltip text to show?

packages/app/src/specs/SpecListGitInfo.vue
```diff
@@ -66,30 +67,36 @@ const classes = computed(() => {
...
       iconClasses: 'icon-light-gray-500',
-      showTooltip: true,
     },
   }[props.gql?.statusType || 'unmodified']
 })
 
-const gitTooltipSubtext = computed(() => {
+const tooltipMainText = computed(() => {
+  switch (props.gql?.statusType) {
+    case 'unmodified': return props.gql?.subject
```
emilyrohrbough, line 84: what does this props.gql?.subject show?

packages/app/src/specs/SpecListGitInfo.vue
```diff
@@ -6,22 +6,23 @@
...
-            {{ props.gql.subject }}
+            {{ tooltipMainText }}
           </p>
-          <p class="text-xs">
-            {{ gitTooltipSubtext }}
+          <p
+            v-if="tooltipSubtext"
+            class="text-xs"
+          >
+            {{ tooltipSubtext }}
```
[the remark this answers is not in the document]
MuazOthman: The parent `p` is already rendered conditionally on line 22, can you elaborate on the additional condition I'm missing?

packages/app/src/specs/SpecListGitInfo.vue
emilyrohrbough: Since the default was `null` I assumed there was cases when there wasn't content.

Is there always a tooltip even when the SCM is not available and/or git is not used in the project?

packages/app/src/specs/SpecListGitInfo.vue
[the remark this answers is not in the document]
emilyrohrbough: nope! I missed it. Should have looked closer at the Vue syntax. Was thinking React syntax here. Sorry 😅

packages/app/src/specs/SpecListGitInfo.vue
MuazOthman: This is the commit subject. I think #21139 can further show previous work updated in this PR.

packages/app/src/specs/SpecListGitInfo.vue
MuazOthman: If there's no SCM data available, this whole component won't be rendered
https://github.com/cypress-io/cypress/blob/06a02831e71432bd3837b9f4824c109c77ccbe52/packages/app/src/specs/SpecsList.vue#L105

packages/graphql/schemas/schema.graphql
```diff
@@ -760,6 +760,7 @@ type GitInfo {
 enum GitInfoStatusType {
   created
   modified
+  no_git_info
```
tgriesser, line 763: Not seeing the related `gql-` enum change, was this edited in the SDL file directly?

Also, we generally use camelCase everywhere for variable naming, unless it's `CONSTANT_CASE`, so `noGitInfo`

tgriesser (changes requested):

packages/graphql/schemas/schema.graphql
MuazOthman: I corrected to follow convention.
The main update is in `packages/types/src/git.ts`
https://github.com/cypress-io/cypress/blob/b8884b58f1d4b4fa5db90cb7867dcd593c370727/packages/types/src/git.ts#L1

MuazOthman:
@ZachJW34 this issue related to tooltips refusing to show up seems odd, would you have time to help look into it tomorrow maybe?
Meanwhile, I guess we can get this PR merged if since it was an existing issue and the updates in this PR need to go into 10.0.

packages/frontend-shared/src/components/Tooltip.cy.tsx
```diff
@@ -1,24 +1,58 @@
...
     })
 
-    cy.get('.default-slot')
+    cy.get('.default-slot0')
+    .realHover()
+
+    cy.contains('Tooltip 0')
+  })
+
+  it('playground - interactive', { viewportHeight: 400 }, () => {
```
marktnoonan, line 25: Since interactive is intended to have a different appearance than regular tooltips, could we add a percy snapshot to this and the other test, and update the test descriptions?

marktnoonan (changes requested):
@MuazOthman could we please have a bit more description on this PR, it seems to do more than just add tooltips for the icons. Can we give context to the other changes made? It's hard to review without this context. A good example is the new tooltip on the header:

Screen Shot 2022-05-24 at 12 45 13 PM

This is not mentioned in the PR or the ticket as far as I can tell, and it looks different to other existing tooltips. What's my reference point to approve this change? I'm sure there is a figma link but don't know where to find it.

There are some other refactors as well. These may be improvements but it's not clear how they relate to what's described in the PR.

The main change request I have is that this PR appears to introduce something that breaks keyboard navigation through the specs list. After tabbing onto an icon that tiggers a tooltip, it gains focus and the next tab stop is the browser address bar:

https://user-images.githubusercontent.com/8340719/170088843-00b1944a-5fac-414d-a823-774c45119a61.mov

I confirmed this doesn't happen on 10.0-release - the icons still show tooltips on focus but I can immediately tab down through the next rows as expected.

On the other hand the column header does not receive keyboard focus and so doesn't reveal its tooltip to keyboard users. Ideally the header and the icons in the rows would behave the same way when focused.

I made a suggestion in the e2e spec for where to add coverage related to the tooltip content changing based on the git status - I think without this coverage we would not catch a regression if these tooltips were were removed/changed.

packages/frontend-shared/src/components/Tooltip.vue
```diff
@@ -2,7 +2,11 @@
   <Tooltip
     v-if="!disabled"
     :popper-class="{'no-arrow': hideArrow}"
-    :popper-hide-triggers="['hover']"
+    :popper-triggers="['hover']"
+    :hide-triggers="['hover']"
+    :auto-hide="!isInteractive"
+    :theme="theme"
+    :placement="placement ?? 'auto'"
```
marktnoonan, line 9: I did a little more testing and found what I think are some other side effects related to these changes that surface when using the keyboard, leading to unexpected states:

Screen Shot 2022-05-24 at 6 39 19 PM

After comparing tooltips in general with 10.0-release, it seems something in here is also affecting the tooltips in the Selector Playground - both their positioning and focus behavior. Moving focus out to the address bar immediately when a button that has a tooltip is focused. Let me know if you can reproduce, here's what I did:

https://www.loom.com/share/598305649aa741499ca1f909b80e31db - 10.0-release
https://www.loom.com/share/469f94e459554d729c5fe13b63489f2a - this branch

This seem like pretty extreme effects for such a small set of changes, so maybe something else is up with my local env. Though the changed positioning in the Selector Playground is also captured in percy [here](https://percy.io/cypress-io/cypress/builds/18347831/changed/1028975721?browser=chrome&browser_ids=23&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=1000&widths=536%2C800%2C1000%2C1500).

marktnoonan:
Awesome, thanks @MuazOthman, I will check it out. Does this mean we can remove `packages/app/src/specs/LastUpdatedHeader.vue` itself since it's not used? And what do you think of adding the assertions I mentioned in this comment to verify that the new tooltips appear and are mapped correctly to the statuses: https://github.com/cypress-io/cypress/pull/21536#discussion_r880715651

MuazOthman:
@ZachJW34 the issue you found is now fixed thanks to @tbiethman. Can you please verify the fix, @ZachJW34?

MuazOthman:
@marktnoonan file removed.

packages/app/src/specs/SpecsList.vue
```diff
@@ -43,7 +43,9 @@
           t('specPage.componentSpecsHeader') : t('specPage.e2eSpecsHeader') }}
       </div>
       <div class="flex items-center justify-between">
-        <div>{{ t('specPage.lastUpdatedHeader') }}</div>
+        <div :is-git-available="isGitAvailable">
```
marktnoonan, line 46: Looks like we can remove this attribute from the div and the `isGitAvailable` computed value, then this file won't be touched by the PR.

MuazOthman: I removed that, but the localized string path was updated so this will touch the file.

packages/app/src/specs/SpecListGitInfo.vue
```diff
@@ -66,30 +70,40 @@ const classes = computed(() => {
...
+const tooltipSubtext = computed(() => {
   if (props.gql?.statusType === 'unmodified') {
     return t('specPage.rows.gitTooltipSubtext', {
       author: props.gql.author,
       shortHash: props.gql.shortHash,
     })
   }
 
-  return ''
+  return null
```
rachelruderman, line 106: How come we changed the return type from string to string or null?

packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -89,31 +88,46 @@ export class GitDataSource {
     debug('config: %o', this.config)
     try {
       this.#git = simpleGit({ baseDir: this.config.projectRoot })
-      this.#gitBaseDir = this.config.projectRoot
     } catch {
-      this.#git = simpleGit()
+      // suppress exception if git cannot be found
```
rachelruderman, line 92: I hit another spot in the code where an exception was suppressed and it took me foreverrrrrsssss to track it down. Is there any way we can surface an error, i.e. the error still shows in the console, but it does not impact the UI?

packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -89,31 +88,46 @@ export class GitDataSource {
...
+    } catch {
+      this.#gitErrored = true
+    }
+  }
+
+  #refreshAllGitData () {
+    this.#verifyGitRepo().then(() => {
+      const toAwait = [
+        this.#loadAndWatchCurrentBranch(),
+      ]
```
rachelruderman, line 119: I like!

packages/app/src/specs/SpecListGitInfo.vue
```diff
@@ -1,27 +1,31 @@
 <template>
   <div
-    class="h-full grid gap-9px grid-cols-[16px,auto] git-info-row items-center"
+    class="h-full grid gap-9px git-info-row items-center"
+    :class="{'grid-cols-[16px,auto]': classes.icon}"
     data-cy="git-info-row"
   >
     <Tooltip
+      v-if="classes.icon"
+      :key="props.gql?.statusType ?? undefined"
```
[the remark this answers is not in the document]
MuazOthman: It seems that the `Tooltip` component from `floating-vue` binds events to its contents when it is mounted, which we're losing when the contents change, causing the tooltip not to show up.
Adding the `key` attribute and causing it to change value when the target changes forces Vue to re-render the DOM, which allows for the events to be bound correctly. This may not be an ideal solution, but given the use case here it doesn't create much overhead as these changes are triggered by file status changes, which are not expected to be frequent within 60 seconds, which is the frequency used to update the file status icon.
[This article](https://michaelnthiessen.com/understanding-the-key-attribute) explains some more details about `key` in Vue. In summary:
> The key attribute tells Vue how your data relates to the HTML elements it's rendering to the screen. When your data changes, Vue uses these keys to know which HTML elements to remove or update, and if it needs to create any new ones.

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -121,7 +120,15 @@
     "searchPlaceholder": "Search Specs",
     "componentSpecsHeader": "Component specs",
     "e2eSpecsHeader": "E2E specs",
-    "lastUpdatedHeader": "Last updated",
+    "lastUpdated": {
+      "header": "Last updated",
+      "tooltip": {
```
marktnoonan, line 125: We don't need this tooltip text in the i18n just yet, might as well remove it until the PR with the header updates comes along? Not a blocking request though.

packages/app/src/specs/SpecListGitInfo.vue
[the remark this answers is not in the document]
marktnoonan: Yeah a `key` seems like the right move here 👍

packages/app/src/specs/SpecListGitInfo.vue
MuazOthman: I guess this was me missing on how JS evaluates falsy strings 😆 . I'm reverting it.

packages/app/src/specs/SpecListGitInfo.vue
```diff
@@ -66,30 +70,40 @@ const classes = computed(() => {
...
-      showTooltip: false,
     },
     unmodified: {
       icon: CommitIcon,
       iconClasses: 'icon-light-gray-500',
-      showTooltip: true,
+    },
+    noGitInfo: {
+      icon: null,
+      iconClasses: 'icon-light-gray-500',
```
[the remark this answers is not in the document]
MuazOthman: Good catch, removed `iconClasses` for this case.

packages/data-context/src/sources/GitDataSource.ts
MuazOthman: `debug` added

packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -1,5 +1,5 @@
 import execa from 'execa'
```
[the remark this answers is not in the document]
MuazOthman: I added a new test to cover this case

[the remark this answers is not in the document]
MuazOthman: Just saw this test fail for Windows, investigating.

marktnoonan (commented):
I see a couple of tasks left:

- Update this test mentioned here: https://github.com/cypress-io/cypress/pull/21536/files#r880715651. At the moment, if I completely remove the tooltip from `SpecListGitInfo.vue`, the `specs_list_actual_git_repo.cy.ts` test still passes, which seems wrong to me, it should be yelling about the tooltips being missing. Ideally the main tooltip with the git commit info would also be covered here, but I know that's not part of this PR, so it's your call whether to add it or not.

- Update PR description to explain what's expected in the "no git info" state to provide context for the `GitDataSource` changes.

And just a side note, is the right box ticked here? Not familiar with our ZenHub tagging process but wondering if you meant to check the top box instead.

Screen Shot 2022-05-26 at 8 22 20 AM

ZachJW34:
@MuazOthman verified that the tooltip is acting how it should! I'm curious, what was the fix?

packages/app/src/specs/SpecListGitInfo.vue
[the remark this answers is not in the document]
MuazOthman: @ZachJW34 this was the fix for the issue you found

packages/app/cypress/tasks/git.ts
```diff
@@ -30,3 +30,11 @@ export async function initGitRepoForTestProject (projectPath: string) {
 
   return null
 }
+
+export async function resetGitRepoForTestProject (projectPath: string) {
+  const git = simpleGit({ baseDir: projectPath })
+
+  await git.reset(['--hard'])
+
+  return null
```
rachelruderman, line 39: How come you're returning `null`? Is TS requiring you to have a return value?

MuazOthman: I guess I stole the pattern from the function above it. Upon digging up, it seems it is required to return something per the comment on the code shown below
https://github.com/cypress-io/cypress/blob/d6df1069dc934e4a017e20917231616f472cd0b3/cli/types/cypress.d.ts#L5472-L5477

marktnoonan (approved):
All looks great 👍

rachelruderman (approved):

PR #21538, opened 2022-05-17
feat: launch project if `cypress open --e2e/--ct --browser` is used

flotwig:
- Closes UNIFY-1480

### User facing changelog

- If a testing type and browser are passed to `cypress open` when launching a specific project, the browser will automatically launch.

### Additional details

- This makes 10.0 have parity with 9.x, where users can run 1 `cypress open` command and not have to do further interaction to access specs.

I'm not sure if we need to update the wording of the UI for when we are launching the browser automatically. This PR doesn't change any of the messaging.

### PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4480


BlueWinds (approved):
Nice. Tested locally and I love it.

lmiller1990 (commented):
Tested and it works 👌 
Potential improvement! https://github.com/cypress-io/cypress/pull/21544/files thoughts @flotwig?

packages/data-context/src/actions/MigrationActions.ts
```diff
@@ -78,12 +78,9 @@ export async function processConfigViaLegacyPlugins (projectRoot: string, legacy
 
     const cwd = path.join(projectRoot, pluginFile)
 
-    const { CYPRESS_INTERNAL_E2E_TESTING_SELF, ...env } = process.env
```
tgriesser, line 81: Only thing was wondering about - was this just unnecessary code or did this need to be removed for some reason? I was looking at where it was added in https://github.com/cypress-io/cypress/pull/20786 and couldn't tell if it was necessary

tgriesser (approved):
LGTM, had 1 question but otherwise 👍

packages/data-context/src/actions/MigrationActions.ts
flotwig: It seemed to be a no-op. I noticed because vscode was highlighting that `CYPRESS_INTERNAL_E2E_TESTING_SELF` is unused. I really can't wait til we can turn unused code detection back on :stuck_out_tongue:

tgriesser: Not unused, it was destructuring it out so that it doesn't get passed down as an env var to the child processes when we're opening sub-projects during e2e testing, but I think that is maybe no longer necessary.

PR #21540, opened 2022-05-17
fix: disable ct testing when run via binary

ZachJW34:
- Closes [UNIFY-1695](https://cypress-io.atlassian.net/browse/UNIFY-1695)

### User facing changelog
Disable CT Testing when invoked from the binary

### Additional details
CT Testing requires the binary to be invoked from the cli due to the need to source the user's `node_module` dependencies. If the project is not invoked from the cli, we will now disable the option for the user to choose CT Testing from the page.

I did a small refactor of the testing type status. Since we are now adding another one (disabled), the attributes were getting hard to manage. I also had to make some changes to the `Card.vue` component as it had `pointer-events-none` when disabled which was preventing the link from being clicked.

https://user-images.githubusercontent.com/25158820/168918630-ea76b0f3-8578-4247-8cc2-b36dc210c938.mov

### PR Tasks


tgriesser (changes requested):
Change seems fine to me - based on the test failures I'm assuming there's something else in  the e2e setup layer that needs to be adjusted to make this work?

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -520,6 +521,10 @@
     "component": {
       "name": "Component Testing",
       "description": "Build and test your components from your design system in isolation in order to ensure each state matches your expectations."
+    },
+    "componentDisabled": {
+      "description": "Testing your components requires {0} as an NPM dependency for this project.",
+      "link": "installing Cypress"
```
BlueWinds, line 527: I'd prefer active over passive voice here.

```suggestion
      "description": "To test your components you must {0} as an NPM dependency for this project.",
      "link": "install Cypress"
```

packages/graphql/src/schemaTypes/objectTypes/gql-Query.ts
```diff
@@ -107,6 +107,11 @@ export const Query = objectType({
       resolve: (_, args, ctx) => ctx.coreData.scaffoldedFiles,
     })
 
+    t.nonNull.boolean('invokedFromCli', {
+      description: 'Whether the app was invoked from the CLI',
```
BlueWinds, line 111:
```suggestion
      description: 'Whether the app was invoked from the CLI', False if user is using the binary without npm.
```

Just imagining coming in on this later and not being sure how else the app can be invoked.

BlueWinds (approved):
Some small suggestions, looks good overall. Thanks for the component refactors.

tgriesser (approved):

packages/frontend-shared/src/locales/en-US.json
ZachJW34: @BlueWinds going with your suggestion

PR #21541, opened 2022-05-17
chore: Removing workaround for long Cypress verification times

### LINKED ISSUE #21236: Restore just-in-time `require`/`require` splitting in `cypress.js`

tbiethman:
- Closes #21236
- Related #21174

### User facing changelog

N/A

### Additional details

Our previous usage of Election v15.4.0 resulted in much longer verification times 10.0-release than with previous versions, up to 2x/3x as long. But now that we're using Electron v18, we no longer see this increased verification duration, even without the workaround that was added with https://github.com/cypress-io/cypress/pull/21174 🪄. So I'm yanking it back out and reverting the JIT requires that were added. 

We're sub-5s again:
Screen Shot 2022-05-17 at 5 39 06 PM

### PR Tasks


packages/server/lib/makeDataContext.ts
```diff
@@ -1,5 +1,5 @@
 import { DataContext, getCtx, clearCtx, setCtx } from '@packages/data-context'
-import type { OpenDialogOptions, SaveDialogOptions, BrowserWindow } from 'electron'
+import electron, { OpenDialogOptions, SaveDialogOptions, BrowserWindow } from 'electron'
```
tgriesser, line 2: I am wondering how this is working, because we use `makeDataContext` in the tests (e2ePluginSetup & some of the mocha tests). 

Maybe v18 it handles it better?

tgriesser (changes requested):
Change seems fine to me - could we change the circle config and get a build for this so we can try it out and confirm things look good?

packages/server/lib/makeDataContext.ts
tbiethman: Hmm, this is what we were doing before: https://github.com/cypress-io/cypress/pull/21174/files#diff-f4f06c00a1bc66aafb7f77e5c5ea7b04152df534668fe536d8deb2cdbe950e67L2

Seems to be fine, but if there are concerns we can keep this as-is?

tgriesser: Yeah no concerns, if the tests are passing I'm good with it

packages/server/lib/makeDataContext.ts
```diff
@@ -157,16 +157,16 @@ export function makeDataContext (options: MakeDataContextOptions): DataContext {
...
       showSaveDialog (window: BrowserWindow, props: SaveDialogOptions) {
-        return require('electron').dialog.showSaveDialog(window, props)
+        return electron.dialog.showSaveDialog(window, props)
       },
       copyTextToClipboard (text: string) {
-        require('electron').clipboard.writeText(text)
+        electron.clipboard.writeText(text)
       },
       // These instances of JIT requiring gui/windows can be removed
       // once https://github.com/cypress-io/cypress/issues/21236 is fixed
```
flotwig, line 172: I think it should be possible to remove these guys as well

tbiethman: Good call, I should have seen that that comment. I'll get these updated too.

tgriesser (approved):

lmiller1990 (approved):

PR #21542, opened 2022-05-18
fix: update gql error handler

estrada9166:
- Closes 

### Additional details

Update error handling on gql cloud requests, so the errors can be propagated to the UI and we can handle them with some alerts

[Here](https://app.circleci.com/pipelines/github/cypress-io/cypress/38252/workflows/de5bc3d0-0345-425c-9538-14b7409cf991/jobs/1549246) is a failing test with the previous implementation - that is passing with this new change

### PR Tasks


packages/data-context/src/sources/CloudDataSource.ts
```diff
@@ -174,6 +153,22 @@ export class CloudDataSource {
...
 
+  #formatWithErrors (data: OperationResult<any, any>) {
+    // If we receive a 401 from the dashboard, we need to logout the user
+    if (data.error?.response?.status === 401) {
+      this.params.logout()
+    }
+
+    if (data.error?.graphQLErrors[0]) {
+      this.params.onError(data.error?.graphQLErrors[0])
+    }
```
tgriesser, line 164: I think we can get rid of this and the `onError` we pass to `CloudDataSource` entirely

tgriesser (changes requested):

packages/data-context/src/sources/CloudDataSource.ts
estrada9166: Updated in [cd05ee9](https://github.com/cypress-io/cypress/pull/21542/commits/cd05ee95576355255aa228cbac2f2ff5b58a434b)

packages/data-context/src/sources/CloudDataSource.ts
```diff
@@ -174,6 +152,18 @@ export class CloudDataSource {
     return `${print(config.document)}-${stringifyVariables(config.variables)}`
   }
 
+  #formatWithErrors (data: OperationResult<any, any>) {
+    // If we receive a 401 from the dashboard, we need to logout the user
```
tgriesser, line 156:
```suggestion
  #formatWithErrors: (data: OperationResult<any, any>) => {
    // If we receive a 401 from the dashboard, we need to logout the user
```

Needs to be an arrow fn to reference `this` correctly the way we're passing it to `then`

tgriesser (changes requested):

packages/data-context/src/sources/CloudDataSource.ts
```diff
@@ -183,6 +152,23 @@ export class CloudDataSource {
...
+      // Do nothing on failure here
+      this.invalidate({ __typename: 'Query' }).catch()
+    }
+
+    return {
+      ...data,
+      errors: data.error?.graphQLErrors,
+    }
+  }
+
```
tgriesser, line 171:
```suggestion
  #formatWithErrors = async (data: OperationResult<any, any>) => {
    // If we receive a 401 from the dashboard, we need to logout the user
    if (data.error?.response?.status === 401) {
      await this.params.logout()
    }

    if (data.error && data.operation.kind === 'mutation') {
      await this.invalidate({ __typename: 'Query' })
    }

    return {
      ...data,
      errors: data.error?.graphQLErrors,
    }
  }

```

tgriesser (changes requested):

tgriesser (approved):

PR #21543, opened 2022-05-18
fix: consider `cypress.config.ts`as ES module by default

lmiller1990:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1769

### User facing changelog

- correctly generate ES modules code for projects using `cypress.config.ts` (even without `type="module"`, it's still considered ESM if they are using TS)
- fix weird indentation for cjs based `cypress.config.js`

### PR Tasks


packages/config/test/__fixtures__/has-e2e.config.mjs
```diff
@@ -0,0 +1,3 @@
+export default {
+  e2e: {}
+}
```
lmiller1990, line 3: Below file, `packages/config/test/ast-utils/addToCypressConfig.spec.ts` is marked as "too big to show diff" for some silly reason. It's the main part of this PR - the tests adding this missing coverage. Make sure you expand it and look.

The problem was I thought the `isusingEsModules` flag was the only thing I should check - this actually means "is using type: module". We need to include a check for `ts` which defaults to ESM.

We don't currently cover the case of using TS with CommonJS. I am not sure if this is a consideration we need to make, I've not seen any projects doing this.

lmiller1990: Here's the regression test described in the original issue: 

https://github.com/cypress-io/cypress/pull/21543/files#diff-8a7cf6dc3057090b692b41192285cc863933594137e517ae272da9b736a0196bR125-R155

I also added coverage for `mjs`.

packages/config/test/ast-utils/addToCypressConfig.spec.ts
```diff
@@ -19,223 +19,374 @@ const { addTestingTypeToCypressConfig } = proxyquire('../../src/ast-utils/addToC
...
+              bundlerConfigPath: './webpack.config.js',
+            },
+            isProjectUsingESModules: false,
+          })
+
+          // we generate this using Babel which intentionally does **NOT**
+          // handle formatting, so the output format is kind of weird.
+          // we rely on the user's eslint or prettier to format this properly.
+          const expected = dedent`
+          const webpackConfig = require('./webpack.config.js')module.exports = {
```
tgriesser, line 374: This seems like unexpected behavior and I feel like we should be able to handle this better - if we're adding the import, could we just consistently add it to the top of the config file with a newline or something?

tgriesser (changes requested):
Maybe these are in the snapshots, but I didn't see anything here confirming that we're not including `.ts` extension in the imported config, like this:

image

Which should be

image

It looks like we were previously keeping the extension because the `require` has a `.ts` here, which is why i'm calling it out

image

packages/config/test/ast-utils/addToCypressConfig.spec.ts
lmiller1990: Babel does not support adding formatting what-so-ever, unfortunately. We should be able to run it through something like `@babel/generator`, I'll look into this.

lmiller1990: Oh, easy... recast has a `prettyPrint`, let's just use that. It also solves the "what if they don't have prettier" problem - the basic `prettyPrint` is quite good.

tgriesser (changes requested):
So @ZachJW34 can maybe chime in on this, but I think what we had originally wanted to do with the `webpack.config.js`, was make it work like vite, where we would detect it internally and source automatically if it were the standard naming (`webpack.config.{js|ts}` at the root), which is different than what was added in #21432

The change in currently breaks for vite, because we already source the config here:

https://github.com/cypress-io/cypress/blob/c2b270f1280d71476469f5ab1f4cd1794a869bcb/npm/vite-dev-server/src/resolveConfig.ts#L20-L31
 
https://github.com/cypress-io/cypress/blob/c2b270f1280d71476469f5ab1f4cd1794a869bcb/npm/vite-dev-server/src/constants.ts#L1-L6

So with this change we're now importing the same vite config twice and merging them, which results in errors. We either need to remove the internal sourcing in vite, or make the behavior consistent with webpack, which would eliminate needing all of the changes to the AST, and runtime error if the config can't be sourced

My vote would be to revert #21432 and make it consistent where we use the webpack we detect automatically, and then tell them to provide the config if we don't have one, but I'm fine with either - we just need to pick one b/c right now vite breaks out of the box

tgriesser:
Closing, re: UNIFY-1774

PR #21544, opened 2022-05-18
chore: improve test context method to avoid ts overrides

lmiller1990:
Target https://github.com/cypress-io/cypress/pull/21538

- update `createTestDataContext` to receive `modeOptions`
- avoid using `ts-expect-error`

### DIFF
--- a/packages/data-context/test/unit/data/ProjectLifecycleManager.spec.ts
+++ b/packages/data-context/test/unit/data/ProjectLifecycleManager.spec.ts
@@ -3,31 +3,36 @@ import type { DataContext } from '../../../src'
 import { createTestDataContext } from '../helper'
 import sinon from 'sinon'
 
-describe('ProjectLifecycleManager', () => {
-  let ctx: DataContext
+const browsers = [
+  { name: 'electron', family: 'chromium', channel: 'stable', displayName: 'Electron' },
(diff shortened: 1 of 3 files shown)


packages/data-context/test/unit/data/ProjectLifecycleManager.spec.ts
```diff
@@ -48,15 +53,15 @@ describe('ProjectLifecycleManager', () => {
     })
 
     it('uses cli --browser option and launches project if `--project --testingType` were used', async () => {
+      ctx = createDataContext({
```
lmiller1990, line 56: This is the improvement - we can pass mode options instead of `ts-expect-error`

packages/types/src/modeOptions.ts
```diff
@@ -29,7 +29,6 @@ export interface RunModeOptions extends CommonModeOptions {
 export type TestingType = 'e2e' | 'component'
 
 export interface OpenModeOptions extends CommonModeOptions {
-  _?: (string | null)[] | null
```
flotwig, line 32: this is a real property, it contains argv that did not match a known option. but since it's unused doesn't matter here

flotwig (approved):
nice, thanks @lmiller1990

Issue #21535, opened 2022-05-17
Cypress doesn't run on an environment with precedence of ipv6 over ipv4


insane-developer:
### Current behavior

I have an environment with precedence of ipv6 over ipv4, i.e. this code
`node -e "c=require('net').createConnection({port: 30000 /* any unused port will do*/},()=>{});c.on('error', e=>console.error('address:', e.address))"`
returns `address: ::1`
On this environment cypress creates SNI server with [hardcoded ipv4 address](https://github.com/cypress-io/cypress/blob/develop/packages/https-proxy/lib/server.js#L271) and therefore cannot connect to it [later on](https://github.com/cypress-io/cypress/blob/develop/packages/https-proxy/lib/server.js#L157), because nodejs effectively use address `::1` for localhost in this case.

Here are relevant parts of logs:
```
[13:23:06] OUT 2022-05-17T10:23:06.582Z cypress:server:server-base server open
[13:23:06] OUT 2022-05-17T10:23:06.590Z cypress:network:client-certificates loaded client certificates for 0 URL(s)
[13:23:06] OUT 2022-05-17T10:23:06.593Z cypress:server:server-e2e createServer connecting to server
[13:23:06] OUT 2022-05-17T10:23:06.596Z cypress:server:server-base Server listening on  { address: '127.0.0.1', family: 'IPv4', port: 35179 }
[13:23:06] OUT 2022-05-17T10:23:06.596Z cypress:server:appdata path: /home/sandbox/.config/Cypress/cy/production/proxy
[13:23:06] OUT 2022-05-17T10:23:06.599Z cypress:https-proxy:ca checking CA version { actualVersion: 1, CA_VERSION: 1 }
[13:23:06] OUT 2022-05-17T10:23:06.619Z cypress:https-proxy Created SNI HTTPS Proxy Server { port: 44261 }
...
[13:23:06] OUT 2022-05-17T10:23:06.943Z cypress:server:browsers:electron setting user agent to: Mozilla/5.0 (Linux; Android 8.0.0; Pixel 2 XL Build/OPD1.170816.004) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4183.102 Mobile Safari/537.36
[13:23:06] OUT 2022-05-17T10:23:06.944Z cypress:server:browsers:electron clearing cache
[...]

PR #21545, opened 2022-05-18
fix: pass family parameter to connect method

insane-developer:
- Closes https://github.com/cypress-io/cypress/issues/21535

### User facing changelog
n/a

### Additional details

`family: 4` explicitly forces Socket.connect to use ipv4 address.
Without this change Socket.connect may use ipv6 address which whould lead to error due to the explicit use of `127.0.0.1` [here](https://github.com/cypress-io/cypress/blob/develop/packages/https-proxy/lib/server.js#L271).

### PR Tasks


jennifer-shehane (changes requested):
@insane-developer Could you please sign our CLA? We can’t accept any PRs without this signed.

flotwig (approved):
Looks good to me @insane-developer. We try to use `127.0.0.1` as much as possible to avoid bugs like this, but this is a much better fix.

ryanthemanuel (approved):

PR #21550, opened 2022-05-18
fix: Ensuring spec file presence prior to webpack-dev-server compilation

tbiethman:
- Closes UNIFY-1697

### Additional details

There is a race condition between our webpack compilation and our emission of the `dev-server:specs:changed` event. When a spec file is removed from the file system, we attempt to compile prior to updating our internal `files` property within the plugin, causing the loader to reference files that no longer exist.

I updated the `CypressCTWebpackPlugin` to filter out files in state that no longer exist on the file system during the beforeCompile phase. This prevents both the initial error from being presented at file removal time and allows the file to be immediately executed when re-created.

https://user-images.githubusercontent.com/1711637/169138230-2e4a3493-0726-4a1c-9eb1-2074fadc9aef.mov

### PR Tasks

### DIFF
--- a/npm/webpack-dev-server/cypress/e2e/webpack-dev-server.cy.ts
+++ b/npm/webpack-dev-server/cypress/e2e/webpack-dev-server.cy.ts
@@ -1,3 +1,4 @@
+/// <reference path="../support/e2e.ts" />
 describe('Config options', () => {
   it('supports supportFile = false', () => {
     cy.scaffoldProject('webpack5_wds4-react')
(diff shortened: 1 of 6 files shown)


npm/webpack-dev-server/src/CypressCTWebpackPlugin.ts
```diff
@@ -68,6 +68,16 @@ export class CypressCTWebpackPlugin {
...
   };
 
+  private beforeCompile = () => {
+    if (!this.compilation) {
+      return
+    }
+
+    // Ensure we don't try to load files that have been removed from the file system
+    // but have not yet been detected by the onSpecsChange handler
+    this.files = (this.files || []).filter((file) => fs.existsSync(file.absolute))
```
lmiller1990, line 78: I am not sure we want to use `existsSync` here, should we use the async `pathExists` from `fs-extra` and `tapAsync`?

Eg: https://github.com/cypress-io/cypress/compare/tbiethman/fix/unify-1697-ct-file-removed...lmiller1990/fix/unify-1697-ct-file-removed?expand=1

lmiller1990: https://webpack.js.org/api/compiler-hooks/#beforecompile can work with `tapAsync`

lmiller1990: Or - is the point we **must** do this async, so we avoid the race condition?

lmiller1990 (commented):
Just a quick question around sync vs async!

npm/webpack-dev-server/cypress/e2e/webpack-dev-server.cy.ts
```diff
@@ -8,4 +8,39 @@ describe('Config options', () => {
     cy.contains('App.cy.jsx').click()
     cy.get('.passed > .num').should('contain', 1)
   })
+
+  // https://cypress-io.atlassian.net/browse/UNIFY-1697
+  it('filters missing spec files from loader during pre-compilation', () => {
```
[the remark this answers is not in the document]
lmiller1990: They are quite cool, but I wonder if a regression test would be best added in the actual `npm/webpack-dev-server`?

npm/webpack-dev-server/src/CypressCTWebpackPlugin.ts
tbiethman: We just need this to execute prior to the compile phase, which should happen either sync or async. Given that we're not actively manipulating the compile params here, we might as well use async and whatever benefit that might get us.

Your example PR was helpful, thank you. I thought about going with a Promise.all rather than running them in series. Something like:

```js
    const foundFiles = await Promise.all(this.files.map(async (file) => {
      try {
        const exists = await fs.pathExists(file.absolute)

        return exists ? file : null
      } catch (e) {
        return null
      }
    }))

    this.files = foundFiles.filter((file) => file !== null)
```

Thoughts?

tbiethman:
I updated this to use an async beforeCompile implementation, seems to work great. I also added a mocha integration test to get slightly more granular coverage of this behavior. To help with this, I added a new event (`dev-server:compile:done`) to our existing `done` hook so I could more easily test that a failure is not occurring during the compile without relying on a timeout.

npm/webpack-dev-server/src/CypressCTWebpackPlugin.ts
lmiller1990: Seems fine 💯

lmiller1990 (approved):
Tests were improved, looks good. @ZachJW34 had previously QA'd so I'll merge this up.

PR #21551, opened 2022-05-18
fix: clean up some unexpected behavior and designs

elevatebart:
While checking if we could safely remove the safelist, I found a few odd designs.

## Two color class bugs

In a few places around the app, we have multiple utility classes triggering the color of the text or the background.

Since CSS classes don't take order into account, one cannot have the following code expect a deterministic rendering:

```html
<button class="text-gray-600 text-indigo-600">
  Am I blue or gray?
</button>
```

It will depend on the position of both classes in the CSS file. 

This use case is obvious. But when dynamic stuff is involved, it can be tricky. Look at the following vue use case.

```vue
<template>
  <button class="text-indigo-600" :class="{ 'text-gray-600': disabled }">
    Am I blue or gray?
  </button>
</template>
```

If the CSS class `text-gray-600` is defined before `text-indigo-600` in the CSS file, this button will never be gray when disabled. More deceptively, when building, the order might not be exactly the same and something that works in dev might not work in production.

## Retry is not in the designs

Everywhere there is an error in Figma I only see `Try again` not `retry`.


packages/frontend-shared/src/components/Card.vue
```diff
@@ -40,9 +40,10 @@
     <!-- this button can be focused via tab key and allows card hocus styles to appear
     as well as allows a keyboard user to "activate" the card with spacebar or enter keys -->
     <button
-      class="font-medium mx-8px mb-8px text-indigo-500 text-18px leading-24px focus:outline-transparent"
+      class="font-medium mx-8px mb-8px text-18px leading-24px focus:outline-transparent"
       :class="{
-        'text-gray-700 cursor-default': disabled
+        'text-gray-700': disabled,
```
elevatebart, line 45:
```suggestion
        'text-gray-700 cursor-default': disabled,
```

tbiethman (approved):
Changes look good, the Percy snapshots we have show diffs I'd expect from these changes (and the existing flake). I appreciate the thorough PR description too, I'll have to keep an eye out for these scenarios going forward.

We can point this PR to develop now 🚀

marktnoonan:
These look like good changes, gonna rerun CI for Percy.

elevatebart:
@marktnoonan when you have a minute I would love your eyes on this one.

ZachJW34 (approved):

PR #21560, opened 2022-05-19
fix: revert #21432 automatically find/merge config for component testing

tgriesser:
This reverts commit a355ad61593df38c75e9df109b2cce740c466a61 (#21432)

- Related to UNIFY-1774

### User facing changelog

Reverting this commit because it is breaking our current vite behavior, and we would instead prefer to change our webpack behavior to work similar to vite, with an error if we can't find a user config when it's required

### PR Tasks


lmiller1990:
Yep, we talked about this and are reverting it for the bug-hunt, and will follow up w/ better errors if the config is not found.

ryanthemanuel (approved):

PR #21563, opened 2022-05-19
fix: UNIFY-1774 error if component config is not sourced for webpack/vite

tgriesser:
- Closes UNIFY-1774

### User facing changelog

Provides a better error if the user does not have webpack / vite config, and we cannot detect it.

### Additional details

If a user does not supply a `webpackConfig` / `viteConfig`, and we cannot infer one based on the framework, default location, etc. we should be erroring with a descriptive indication that they need to provide one, otherwise component testing won't work.

We should also be changing the webpack config to source from the `webpack.config.{js|ts|cjs|mjs}` from the root of their project, as we do with vite

Example error if you provide `bundler: 'vite'` but no `vite.config` is found.

image

### PR Tasks


npm/vite-dev-server/src/devServer.ts
```diff
@@ -10,6 +10,7 @@ export type ViteDevServerConfig = {
   specs: Cypress.Spec[]
   cypressConfig: Cypress.PluginConfigOptions
   devServerEvents: NodeJS.EventEmitter
+  onConfigNotFound?: (devServer: 'vite', cwd: string, lookedIn: string[]) => void
```
lmiller1990, line 13: We probably need a `@cypress/dev-server-utils` package that defines a kind of interface that all our dev-servers must respect, which would include passing a `onConfigNotFound` error handler. We have a good idea of how future dev servers will look now, so post-10.0 we can start to make more concrete docs on how to make one.

Some things that might go there:

- `findConfig` (use `findUp`)
- common interface
- ??

lmiller1990:
Picked this up and pushed some commits, added screenshot to PR description.

marktnoonan:
To test this out, I opened the new system tests projects from global mode. The `missing-webpack-config` project showed the error as expected:

Screen Shot 2022-05-31 at 10 28 48 AM

The `missing-vite-config` project opened right up without any error shown, not sure what to make of that yet. Locally, all the tests are passing, both errors appear, so I would have expected both errors when opening the projects as well. Will keep testing, probably there really is a vite config lurking somewhere.

tgriesser:
> Will keep testing, probably there really is a vite config lurking somewhere.

Yeah, the current behavior does a `find-up` without bounds, ideally this would stop traversal at the project root or the git root or something, but this has been the behavior going back for so long that I'm hesitant to change it at the last minute

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -94,6 +92,42 @@ export function makeWebpackConfig (
...
+  if (!userWebpackConfig && !frameworkWebpackConfig) {
+    debug('Not user or framework webpack config received. Trying to automatically source it')
+
+    const { default: findUp } = await importModule('find-up')
+
+    configFile = await findUp(configFiles, { cwd: projectRoot } as { cwd: string })
+
+    if (configFile) {
+      debug('found webpack config %s', configFile)
+      const sourcedConfig = configFile.endsWith('mjs') ? await import(configFile) : require(configFile)
```
ZachJW34, line 106: If we are already using `importModule`, which works for both ES and cjs, should we use it here?
```suggestion
      const sourcedConfig = await importModule(configFile)
```

npm/webpack-dev-server/src/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const configFiles = [
+  'webpack.config.ts',
+  'webpack.config.js',
+  'webpack.config.mjs',
```
ZachJW34, line 4: Have we tried testing a project that uses `.mjs`? I tried and got an error, but all I did was change an existing project `webpack.config.js` to `webpack.config.mjs` so not sure if it on my side.

Screen Shot 2022-05-31 at 10 04 11 AM

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -94,6 +92,42 @@ export function makeWebpackConfig (
...
+
+      debug('config contains %o', userWebpackConfig)
```
ZachJW34, line 108: This will always be empty unless the debug is moved down a few lines
```suggestion
      debug('config contains %o', sourcedConfig)
```

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -94,6 +92,42 @@ export function makeWebpackConfig (
...
+      if (sourcedConfig && typeof sourcedConfig === 'object') {
+        userWebpackConfig = sourcedConfig.default ?? sourcedConfig
+      }
```
ZachJW34, line 111: It's common for configs to be functions, can we refactor to something like this:

```suggestion
      if (sourcedConfig) {
        const source = sourcedConfig.default ?? sourcedConfig

        if (typeof source === 'object') {
          userWebpackConfig = source
        }

        if (typeof source === 'function') {
          userWebpackConfig = source()
        }
      }
```

ZachJW34 (commented):
Tested and it works really well. Have a few questions that I don't think are blockers

npm/webpack-dev-server/src/makeWebpackConfig.ts
tgriesser: Is there a standard for what should be passed if the config is a function? I figured that could be a separate follow-up PR to this since it might need more definition around the behavior, and in the meantime you could just import the function it into your config file and invoke it as you'd expect.

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -94,6 +92,42 @@ export function makeWebpackConfig (
...
+    }
+
+    if (!userWebpackConfig) {
+      debug('could not find webpack.config!')
+      if (config.devServerConfig?.onConfigNotFound) {
+        config.devServerConfig.onConfigNotFound('webpack', projectRoot, configFiles)
+        // The config process will be killed from the parent, but we want to early exit so we don't get
+        // any additional errors related to not having a config
+        process.exit(0)
```
brian-mann, line 120: this does not look/feel right... will `ipc.send(...)` even resolve considering its async if you are immediately killing the process?

packages/errors/src/errors.ts
```diff
@@ -1530,6 +1530,23 @@ export const AllCypressErrors = {
...
+
+  DEV_SERVER_CONFIG_FILE_NOT_FOUND: (devServer: 'vite' | 'webpack', root: string, searchedFor: string[]) => {
+    const devServerConfigFile = `${devServer}Config`
+
+    return errTemplate`\
+      You are using ${fmt.highlight(devServer)} for your dev server, but a configuration file was not found. We traversed upwards from:
+      
+      ${fmt.highlightSecondary(root)}
+      
+      looking for a file named:
```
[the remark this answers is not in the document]
tgriesser: This is because it's a continuation from the previous sentence, `we traversed upward from ... looking for a file named:`

npm/webpack-dev-server/src/makeWebpackConfig.ts
ZachJW34: I'm cool with it being a followup, here is the docs for the function signature:
https://webpack.js.org/configuration/configuration-types/#exporting-a-function

npm/webpack-dev-server/src/makeWebpackConfig.ts
tgriesser: It it does otherwise the system tests wouldn't be passing. Could log an issue to follow up on this, because changing this will require a bit more investigation/work.

BlueWinds (approved):

npm/webpack-dev-server/src/constants.ts
BlueWinds: Wanted to add a +1 here for getting this fixed ASAP. I have run into so much pain over the years with `type: 'module'`, and we should avoid adding to it if possible. Let's open a new issue for this and any other follow-ups, make sure we don't lose track.

tgriesser:
Logged https://github.com/cypress-io/cypress/issues/21983 to track follow up items

PR #21574, opened 2022-05-19
fix: support loading config files for projects with type module, without esbuild

JessicaSachs:
- Closes [UNIFY-1779](https://cypress-io.atlassian.net/browse/UNIFY-1779)

@flotwig or @lmiller1990 I could use a hand in figuring out why the existing test suite [here](https://github.com/cypress-io/cypress/blob/10.0-release/system-tests/test/config_modules_spec.ts#L40-L51) does not fail.

Do we always have `esbuild` in the require callstack? I thought we sandboxed the node_modules well enough in the temp dir.

### User facing changelog
We now support Cypress Config files for project that have `"type": "module"` in their package.json, but do not have a dependency on `esbuild`.


packages/server/lib/plugins/child/run_require_async_child.js
```diff
@@ -118,7 +118,7 @@ function run (ipc, file, projectRoot) {
 
       return (await bundleRequire({ filepath: file })).mod
     } catch (err) {
-      if (err.stack.includes(`Cannot find package 'esbuild'`)) {
+      if (err.stack.includes(`Cannot find package 'esbuild'`) || err.stack.includes(`Cannot find the module 'esbuild'`)) {
```
JessicaSachs, line 121:
```suggestion
      if (err.stack.includes(`Cannot find package 'esbuild'`) || err.stack.includes(`Cannot find module 'esbuild'`)) {
```

packages/server/lib/plugins/child/run_require_async_child.js
```diff
@@ -118,7 +118,7 @@ function run (ipc, file, projectRoot) {
 
       return (await bundleRequire({ filepath: file })).mod
     } catch (err) {
-      if (err.stack.includes(`Cannot find package 'esbuild'`)) {
+      if (err.stack.includes(`Cannot find package 'esbuild'`) || err.stack.includes(`Cannot find module 'esbuild'`)) {
```
JessicaSachs, line 121: I can't actually find an instance where "Cannot find package" is the error message -- at first I was worried that specific node versions might have changed the language of the stack, but I think I just made a typo in the first place.


```suggestion
      if (err.stack.includes(`Cannot find module 'esbuild'`)) {
```

packages/server/lib/plugins/child/run_require_async_child.js
```diff
@@ -118,7 +118,7 @@ function run (ipc, file, projectRoot) {
 
       return (await bundleRequire({ filepath: file })).mod
     } catch (err) {
-      if (err.stack.includes(`Cannot find package 'esbuild'`)) {
+      if (err.stack.includes(`Cannot find module 'esbuild'`)) {
```
tgriesser, line 121:
```suggestion
      if (err.code === 'MODULE_NOT_FOUND') {
```

tgriesser (dismissed):

packages/server/lib/plugins/child/run_require_async_child.js
tgriesser: Actually, I think we should just be try/catch'ing the `require.resolve` and then if that fails do the early return

tgriesser: Like this: https://github.com/cypress-io/cypress/pull/21576/files#diff-d25a897afab55c6be61f5c5064524f60036f4ef0e6b399ebb7f0c851ea0a22b7

JessicaSachs: No, that doesn't make sense. MODULE_NOT_FOUND is a legit error that can come from any module that bundleRequire fails to find when it tries to load the config file

JessicaSachs: While using the error code looks nice, that’s a legit error we conditionally want to throw. In your proposed change we also SWAP preferring node’s import over bundleRequire, which is not something we want to do. This OSS a change in functionality. bundleRequire shims esm-only libraries like `findUp` Which is why we prefer it. 

This PR is a typo fix. We’re code frozen. The code flow is well-tested and I don’t want to change it.

tgriesser: @JessicaSachs Did you look at the code thing I linked, it does not use the `err.code` and doesn't change the control flow, it's:

```ts
try {
  require.resolve('esbuild')
} catch {
  debug(`User doesn't have esbuild. Going to use native node imports.`)

  // We cannot replace the initial `require` with `await import` because
  // Certain modules cannot be dynamically imported
  return await import(file)
}

debug(`They have esbuild, so we'll load the configFile via bundleRequire`)
const { bundleRequire } = require('bundle-require')

return (await bundleRequire({ filepath: file })).mod
```

Which is just being more explicit about what we're guarding for failure, doesn't rely on string matching, which might be different based on node version or change at some point in the future, and doesn't change the current control flow. 

Regardless, we should make sure we can get a failing test in here with whatever the change ends up being.

tgriesser (changes requested):

packages/server/lib/plugins/child/run_require_async_child.js
JessicaSachs: The only functional issue I have with the above code is that you no longer throw the errors from `bundleRequire`, which may include TS compilation errors, etc.

**On the stability of the current code:**
The `err.stack` string hasn't changed in a long time -- it's safe to assert on across node versions and I think you yourself suggested I make this change ~5 weeks ago. The reason it's different is because I made a typo initially.

**On why this wasn't caught in our infrastructure:**
Because we install a lot of packages automatically to make our system-tests run faster, `esbuild` was always in the `require.resolve` chain as a transitive dependency and we never hit this code path in the tests.

So yes, agreed, we need to fix our system test infra. I've been talking with @flotwig about WHY the tests I wrote didn't fail.

tgriesser: Ok, @JessicaSachs and I met on this - it's fine, let's just land this change for now to get it working

tgriesser (approved):

ryanthemanuel (approved):

flotwig (changes requested):
Would like to see some tests for this, either fixing the existing tests or a new integration test. This error message comparison seems prone to regressions.

JessicaSachs:
There was a test already written, but it wasn't failing correctly due to differences between prod and test.

When running system-tests, `require.resolve` uses the node_modules closest to `cypress/packages/server/lib/plugins/child/run_require_async_child.js` to figure out which `esbuild` to use -- so, when it came time to test this code path, esbuild was always present and we were never hitting the codepath that the test:

https://github.com/cypress-io/cypress/blob/c2cc967227c6359203e3c1bdb99fe54e8fef996f/system-tests/projects/config-cjs-and-esm/config-with-js-module/package.json

was designed to cover.

By adding `paths: [process.cwd()]` into the `require.resolve` statement, the test actually tests something and will now correctly fail.

flotwig (approved):

packages/server/lib/plugins/child/run_require_async_child.js
```diff
@@ -111,14 +111,14 @@ function run (ipc, file, projectRoot) {
     try {
       debug('Trying to use esbuild to run their config file.')
       // We prefer doing this because it supports TypeScript files
-      require.resolve('esbuild')
+      require.resolve('esbuild', { paths: [process.cwd()] })
```
JessicaSachs, line 114: The absence of this line was actually a bug and it was only working on my machine because of some GLOBAL transitive deps with `esbuild`.

To debug this, you can add

```js
require.resolve.paths('esbuild')
```

to a production-built binary installed in a real project. We should really consider moving some of these low-dependency or no-dependency tests OUTSIDE of the system-tests and into the `binary-system-tests` phase.

tgriesser (changes requested):
Looks like there's a test failing now just FYI

jennifer-shehane (dismissed):
@JessicaSachs Update the base branch to be against develop before merging.

PR #21578, opened 2022-05-20
chore: add regression test for ts detection

lmiller1990:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1765 (already closed, adding regression test)

### User facing changelog

Fix a bug where `tsconfig.json` in `node_modules` incorrectly flags that the user's project is using TypeScript.

### Additional details

As per [this tech brief](https://github.com/cypress-io/tech-briefs/pull/22) we changed how we detect TS for the user project. We are using `globby` to detect `tsconfig.json` in the project root, and using the `gitignore: true` flag to ignore files in common places like `node_modules`.

This does not work for one use case - where the user is not using `git`, which means `node_modules` is incorrectly included when globbing for `tsconfig.json`. This was fixed as side effect of another, unrelated PR: https://github.com/cypress-io/cypress/commit/aa4c0b5e09ac4a7049a332556b8c0182dbb0d669#diff-ce8622dde6ed83c3ceb55b5766f02a3dd06e26b84a51e0de44083e9246a768bfR138 but a regression test was not added at that point, so I'm adding one now.

The language is correctly inferred for projects that do not use `git`.

### PR Tasks

### DIFF
--- a/packages/scaffold-config/test/unit/detect.spec.ts
+++ b/packages/scaffold-config/test/unit/detect.spec.ts
@@ -304,4 +304,16 @@ describe('detectLanguage', () => {
 
     expect(actual).to.eq('ts')
   })
+
+  it('ignores node_modules when checking for tsconfig.json', async () => {
+    const projectRoot = await scaffoldMigrationProject('pristine-cjs-project')
+
+    await fs.mkdirp(path.join(projectRoot, 'node_modules', 'some-node-module'))
+    await fs.writeFile(path.join(projectRoot, 'node_modules', 'some-node-module', 'tsconfig.json'), '')
+    const pkgJson = fs.readJsonSync(path.join(projectRoot, 'package.json'))
+
+    const actual = detectLanguage(projectRoot, pkgJson)
+
+    expect(actual).to.eq('js')
+  })
 })


ZachJW34 (approved):

packages/scaffold-config/test/unit/detect.spec.ts
```diff
@@ -304,4 +304,16 @@ describe('detectLanguage', () => {
 
     expect(actual).to.eq('ts')
   })
+
+  it('ignores node_modules when checking for tsconfig.json', async () => {
```
emilyrohrbough, line 308: Can we add a check for this glob patter as well to have full coverage?
 https://github.com/cypress-io/cypress/blob/10.0-release/packages/scaffold-config/src/detect.ts#L135

packages/scaffold-config/test/unit/detect.spec.ts
```diff
@@ -304,4 +304,16 @@ describe('detectLanguage', () => {
...
+    const projectRoot = await scaffoldMigrationProject('pristine-cjs-project')
+
+    await fs.mkdirp(path.join(projectRoot, 'node_modules', 'some-node-module'))
+    await fs.writeFile(path.join(projectRoot, 'node_modules', 'some-node-module', 'tsconfig.json'), '')
```
emilyrohrbough, line 312: Could we have stubbed these to achieve the same results instead of needing to write to the file system?'

Also are these being cleaned up? I'm prob missing it

packages/scaffold-config/test/unit/detect.spec.ts
lmiller1990: We've got a few tests for `ts` files inside of `cypress` already, for example: https://github.com/cypress-io/cypress/blob/10.0-release/packages/scaffold-config/test/unit/detect.spec.ts#L255-L263

packages/scaffold-config/test/unit/detect.spec.ts
lmiller1990: It's using the `scaffoldMigrationProject` function which uses the system testes `Fixtures` code - it's not cleaned up, but the files are all created inside of `/tmp`, so I don't think it's necessary to clean up - it's just doing whatever the `system-tests/fixtures` code already does.

flotwig: For a unit test I'd prefer to use something like https://github.com/tschaub/mock-fs or `sinon`, writing to the filesystem is unnecessarily slow and potentially flaky for this kind of test. Really the only time you /need/ to write to the real FS is when launching multiple processes, like in the system tests.

flotwig (approved):

PR #21579, opened 2022-05-20
feat: (origin) handle waiting for aliased intercepts

### LINKED ISSUE #21317: Handle Aliases in cy.origin callback

mschile:
- Closes #21317 

### User facing changelog

* Updated `cy.origin` to support waiting for aliased intercepted requests.

### Additional details

* Aliased intercepted requests are stored in `Cypress.state` within the primary domain. As such, when waiting on an aliased intercepted request within `cy.origin`, a message is sent to the primary domain to handle the waiting. The response or error is then sent back to the the secondary domain.

**Before:**
Screen Shot 2022-05-19 at 10 45 34 PM

**After:**

**Waiting on intercepted request:**
Screen Shot 2022-05-19 at 10 30 42 PM

**Error when request not called:**
Screen Shot 2022-05-19 at 10 36 48 PM

### PR Tasks


packages/driver/src/cy/aliases.ts
```diff
@@ -31,14 +31,14 @@ export const create = (cy: $Cy) => ({
   getAlias (name, cmd, log) {
     const aliases = cy.state('aliases') || {}
 
-    // bail if the name doesnt reference an alias
+    // bail if the name doesn't reference an alias
     if (!aliasRe.test(name)) {
       return
     }
 
+    // slice off the '@'
```
emilyrohrbough, line 39: 👍🏻

packages/runner/webpack.config.ts
```diff
@@ -85,8 +85,7 @@ mainConfig.resolve = {
 
 // @ts-ignore
 const crossOriginConfig: webpack.Configuration = {
-  mode: 'production',
```
emilyrohrbough, line 88: Was this intentional?

packages/driver/src/cy/commands/waiting.ts
```diff
@@ -62,6 +63,16 @@ export default (Commands, Cypress, cy, state) => {
         // avoid circular reference
         options: _.omit(options, '_log'),
       })
+
+      // if this came from the spec bridge, we need to set a few additional
+      // properties to ensure the log displays correctly
+      if (options.isCrossOriginSpecBridge) {
+        log.set({
+          displayName: 'wait',
+          name: 'wait',
```
emilyrohrbough, line 72:
```suggestion
          name: 'wait',
```

packages/driver/src/cy/commands/waiting.ts
```diff
@@ -62,6 +63,16 @@ export default (Commands, Cypress, cy, state) => {
...
+          message: '',
```
emilyrohrbough, line 73: Doesn't this hide the wait arguments?

packages/driver/cypress/integration/commands/waiting_spec.js
```diff
@@ -68,7 +68,7 @@ describe('src/cy/commands/waiting', () => {
 
           return null
         })
-        .wait('@fetch').then((xhr) => {
+        .wait('@fetch.response').then((xhr) => {
```
AtofStryker, line 71: why is `.response` now needed over just the alias? Is this something that will break users?

packages/driver/cypress/integration/e2e/origin/commands/waiting.spec.ts
```diff
@@ -1,33 +1,341 @@
...
+        expect(actualLog).to.deep.equal(expectedLog)
+      })
+    })
+
+    it('doesn\'t log when log: false', () => {
+      const response = { foo: 'foo' }
+
+      cy.intercept('/foo', response).as('foo')
+
+      cy.origin('http://www.foobar.com:3500', { args: { response } }, ({ response }) => {
```
AtofStryker, line 173: does the response need to be passed in to `cy.origin` here?

packages/driver/src/cy/commands/waiting.ts
```diff
@@ -270,8 +281,51 @@ export default (Commands, Cypress, cy, state) => {
     })
   }
 
+  Cypress.primaryOriginCommunicator.on('wait:for:xhr', ({ args: [str, options] }, originPolicy) => {
+    options.isCrossOriginSpecBridge = true
+    waitString(null, str, options).then((responses) => {
+      // remove props that aren't serializable (these aren't used by consumers so are okay to remove)
+      const omitFn = (response) => _.omit(response, ['subscriptions', 'setLogFlag'])
```
AtofStryker, line 288: Shouldn't `preprocessForSerialization` omit non serializable props already?

packages/driver/src/cy/commands/waiting.ts
```diff
@@ -270,8 +281,51 @@ export default (Commands, Cypress, cy, state) => {
...
+      const preprocessedResponses = _.isArray(responses) ? responses.map(omitFn) : omitFn(responses)
+
+      Cypress.primaryOriginCommunicator.toSpecBridge(originPolicy, 'wait:for:xhr:end', { responses: preprocessedResponses })
```
AtofStryker, line 291: I wonder if we should bake this handling into the `communicator` itself as it already handles specific serialization for some namespaced events. That way the pre/post processing is just contained to the communicator.

packages/driver/src/cy/commands/waiting.ts
```diff
@@ -270,8 +281,51 @@ export default (Commands, Cypress, cy, state) => {
...
+    }).catch((err) => {
+      options._log?.error(err)
+      Cypress.primaryOriginCommunicator.toSpecBridge(originPolicy, 'wait:for:xhr:end', { err: preprocessForSerialization(err) })
+    })
+  })
+
+  const delegateToPrimaryOrigin = ([_subject, str, options]) => {
+    return new Promise((resolve, reject) => {
```
AtofStryker, line 299: now that we have a use case for promise based communicator events, I wonder if we can also add a generic method on the communicator to wrap an one time event in a promise and define some type of generic CB function in an `async`/`await` context, like:

```js
const promisifyToPrimary(toPrimaryEvent, toPrimaryEventArgs, fromPrimaryEvent, onFromPrimaryEventResponse){
  return new Promise((resolve, reject) => {
     Cypress.specBridgeCommunicator.once(fromPrimaryEvent, async (args) => {
         try {
             const genericResults = await onFromPrimaryEventResponse(args)
             resolve(genericResults)
         } catch(e){
             reject(e)
         }
     })
  })
  
  Cypress.specBridgeCommunicator.toPrimary(ToPrimaryEvent, args)
}
```

then you can just make the `onFromPrimaryEventResponse` generic, like

```js
    if (err) {
          if (options.log) {
            Cypress.state('onBeforeLog', (log) => {
              // skip this 'wait' log since it was already added through the primary
              if (log.get('name') === 'wait') {
                // unbind this function so we don't impact any other logs
                cy.state('onBeforeLog', null)

                return false
              }

              return
            })
          }

          throw err
        }

       return responses
```

packages/runner/webpack.config.ts
AtofStryker: I think we still want to keep the simple config if I remember correctly

mschile: Yes, as discussed in standup today, we want to use the same config as `cypress_runner` which uses `commonConfig` to ensure we are getting sourcemaps for development.

packages/driver/src/cy/commands/waiting.ts
mschile: No, the aliases are still displayed:
Screen Shot 2022-05-20 at 11 55 56 AM

packages/driver/src/cy/commands/waiting.ts
mschile: Updated [957b80e](https://github.com/cypress-io/cypress/pull/21579/commits/957b80e549ec301eb382a3cef29a232288013c58)

packages/driver/cypress/integration/e2e/origin/commands/waiting.spec.ts
mschile: No, it doesn't, updated [957b80e](https://github.com/cypress-io/cypress/pull/21579/commits/957b80e549ec301eb382a3cef29a232288013c58)

packages/driver/src/cy/commands/waiting.ts
mschile: Yeah, it does, updated [957b80e](https://github.com/cypress-io/cypress/pull/21579/commits/957b80e549ec301eb382a3cef29a232288013c58)

packages/driver/src/cy/commands/waiting.ts
mschile: Typically I would wait until there is a second use-case before creating a generic pattern. Do we know of another use-case? Thoughts on implementing now or waiting?

emilyrohrbough (approved):

packages/driver/src/cy/commands/waiting.ts
mschile: Clearing the message is also needed otherwise it will display details from the command 'origin' command which is running on the primary.

packages/driver/src/cy/commands/waiting.ts
AtofStryker: I'm OK with waiting. But I wonder if we want to document someplace that if we see this pattern creep up, that we implement it in the communicator

packages/driver/src/cy/commands/waiting.ts
```diff
@@ -5,6 +5,7 @@ import { isDynamicAliasingPossible } from '../net-stubbing/aliasing'
 import ordinal from 'ordinal'
 
 import $errUtils from '../../cypress/error_utils'
+import { preprocessForSerialization } from '../../util/serialization'
```
AtofStryker, line 8:
```suggestion
```

packages/driver/src/cy/commands/waiting.ts
```diff
@@ -270,8 +283,47 @@ export default (Commands, Cypress, cy, state) => {
     })
   }
 
+  Cypress.primaryOriginCommunicator.on('wait:for:xhr', ({ args: [str, options] }, originPolicy) => {
+    options.isCrossOriginSpecBridge = true
+    waitString(null, str, options).then((responses) => {
+      Cypress.primaryOriginCommunicator.toSpecBridge(originPolicy, 'wait:for:xhr:end', { responses: preprocessForSerialization(responses) })
+    }).catch((err) => {
+      options._log?.error(err)
+      Cypress.primaryOriginCommunicator.toSpecBridge(originPolicy, 'wait:for:xhr:end', { err: preprocessForSerialization(err) })
```
AtofStryker, line 292:
```suggestion
      Cypress.primaryOriginCommunicator.toSpecBridge(originPolicy, 'wait:for:xhr:end', { err })
```

packages/driver/src/cy/commands/waiting.ts
mschile: Discussed in standup today and we are going to wait before creating a generic pattern. I don't think there is really a good spot to document this but something that we will keep in mind for the future.

packages/driver/src/cy/commands/waiting.ts
mschile: updated [1b15780](https://github.com/cypress-io/cypress/pull/21579/commits/1b15780a25ec1035bef5f4a915b3b1d060d1abfc)

packages/driver/src/cy/commands/waiting.ts
```diff
@@ -270,8 +283,47 @@ export default (Commands, Cypress, cy, state) => {
     })
   }
 
+  Cypress.primaryOriginCommunicator.on('wait:for:xhr', ({ args: [str, options] }, originPolicy) => {
+    options.isCrossOriginSpecBridge = true
+    waitString(null, str, options).then((responses) => {
+      Cypress.primaryOriginCommunicator.toSpecBridge(originPolicy, 'wait:for:xhr:end', { responses: preprocessForSerialization(responses) })
```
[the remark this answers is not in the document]
mschile: updated [1b15780](https://github.com/cypress-io/cypress/pull/21579/commits/1b15780a25ec1035bef5f4a915b3b1d060d1abfc)

packages/driver/src/cy/commands/waiting.ts
mschile: updated [1b15780](https://github.com/cypress-io/cypress/pull/21579/commits/1b15780a25ec1035bef5f4a915b3b1d060d1abfc)

packages/driver/src/cy/commands/waiting.ts
mschile: The custom serialization was removed. [1b15780](https://github.com/cypress-io/cypress/pull/21579/commits/1b15780a25ec1035bef5f4a915b3b1d060d1abfc)

AtofStryker (approved):

Issue #21585, opened 2022-05-20
ffmpeg doesn't always run at low priority


rrauenza:
### Current behavior

During capture, it looks like ffmpeg runs at low priority:

`0 S gitlab-+  91530  90804 66  99  19 - 517602 unix_s 09:41 ?       00:11:41 /opt/gitlab-runner/NCiG8AGt/0/.../cypress/.cache/Cypress/9.2.0/Cypress/resources/app/node_modules/@ffmpeg-installer/linux-x64/ffmpeg -f image2pipe -use_wallclock_as_timestamps 1 -i pipe:0 -y -vcodec libx264 -filter:v crop='floor(in_w/2)*2:floor(in_h/2)*2' -preset ultrafast /opt/gitlab-runner/NCiG8AGt/0/.../foo.spec.ts.mp4`

I think it is this code:

```
const startCapturing = () => {
    return new Bluebird((resolve) => {
      const cmd = ffmpeg({
        source: pt,
        priority: 20,
      })
      .videoCodec('libx264')
      .outputOptions('-preset ultrafast')
      .on('start', (command) => {
        debug('capture started %o', { command })

        return resolve({
          cmd,
          startedVideoCapture: new Date,
        })

```

But there seems to be another phase -- where I can see up to 600% cpu utilization (affecting other cicd jobs) that is not niced.  

It must be a second pass compressing, because that's when the -crf option is passed.

I think priority: 20 needs to be added here:

```
 return new Bluebird((resolve, reject) => {
    debug('processing video from %s to %s video compression %o',
      name, cname, videoCompression)

    const command = ffmpeg()
```

### Desired behavior

I'd like ffmpeg to always be niced since it is so cpu intensive.

Even better, an option like videoCompression to make this configurable?   And/or a thread limit passed to the command.

### Test code to reproduce

any code will work as long as videos are made.

### Cypress Version

cypress@^9.2.0:

### Other

I can make a PR for priority passed to ffmpeg.

If we want a user configurable option, then that would require more discussion.

rrauenza:
Note:  There is one risk that the dev team should consider, which is that this could make cypress sessions take longer on cicd systems with multiple jobs, because the final ffmpeg will be deprioritized.  

I will argue that this is better than causing other jobs to timeout due to lack of CPU.

rrauenza:
![image](https://user-images.githubusercontent.com/1376251/169610167-80eab71a-9f64-45af-81f7-b837d0aac619.png)

[...]

PR #21587, opened 2022-05-20
fix: Change ffmpeg to post process at priority 20

rrauenza:
- Closes https://github.com/cypress-io/cypress/issues/21585

### User facing changelog
 
Run the post processing ffmpeg at lower priority, similar to when capturing

### Additional details

When running concurrent cicd, the ffmpeg consumes a LOT of CPU.  My experience showed 600%.

No

### PR Tasks

- [n/a ] Have tests been added/updated?
- [n/a ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [n/a ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [ n/a] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?


rrauenza:
PR: https://github.com/cypress-io/cypress/issues/21585

BlueWinds:
Yeah, that's why we push it off until after the tests complete - and then when there's nothing else competing for CPU, allow it to process as quickly as it can. "One thread per CPU, process as quickly as possible" is the intended behavior here.

Generally you don't want to be running multiple instances of Cypress on the same machine at the same time. The preferred pattern is to use docker containers for parallelization - and then use Docker's facilities to manage resources.

BlueWinds:
On further consideration, I think I'm fine with the change - if ffmpeg is the only thing running, then it won't matter if priority is set lower, it won't slow down video processing.

I updated the PR with latest develop in order to kick off the automated tests - once they're green I'll verify that they haven't been slowed down and we can merge in.

packages/server/lib/video_capture.ts
```diff
@@ -283,7 +283,9 @@ export async function process (name, cname, videoCompression, ffmpegchaptersConf
     debug('processing video from %s to %s video compression %o',
       name, cname, videoCompression)
 
-    const command = ffmpeg()
+    const command = ffmpeg(
+      priority: 20,
```
BlueWinds, line 287:
```suggestion
      priority: 20
```

packages/server/lib/video_capture.ts
```diff
@@ -283,7 +283,9 @@ export async function process (name, cname, videoCompression, ffmpegchaptersConf
     debug('processing video from %s to %s video compression %o',
       name, cname, videoCompression)
 
-    const command = ffmpeg()
+    const command = ffmpeg(
+      priority: 20
+    )
```
BlueWinds, line 288:
```suggestion
    const command = ffmpeg({
      priority: 20
    })
```

BlueWinds:
@rrauenza - Looks mostly good, just need to clean up the dangling comma so it'll pass linting.

rrauenza:
@BlueWinds Actually... that lint error was really good because I didn't put the priority: 20 into a {}.  I don't understand the lint error... there *was* a trailing comma, but the code was wrong.   Fix pushed just now.

BlueWinds (approved):

BlueWinds:
@rrauenza - Looks good to me. We're holding off merging anything else in before the 10.0.0 release tomorrow, but will try and get this in for 10.0.1.

PR #21589, opened 2022-05-20
fix: order projects by most recently opened

ZachJW34:
- Closes [UNIFY-1745](https://cypress-io.atlassian.net/browse/UNIFY-1745)

### User facing changelog
Sort projects in global mode by most recently opened

### Additional details
We had all the logic for managing the project list, we just had two sources of truth. This PR rewires the loading of projects so that we always pull from the cache where the ordered list we want is stored.

https://user-images.githubusercontent.com/25158820/169606759-113d8e00-fb08-4d3f-8bc2-d4b4fe4175b2.mov

### PR Tasks


packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -195,12 +194,7 @@ export class ProjectActions {
   async addProject (args: AddProject) {
     const projectRoot = await this.getDirectoryPath(args.path)
 
-    const found = this.projects.find((x) => x.projectRoot === projectRoot)
-
-    if (!found) {
-      this.projects.push({ projectRoot })
-      await this.api.insertProjectToCache(projectRoot)
-    }
+    await this.updateProjectList(() => this.api.insertProjectToCache(projectRoot))
```
tbiethman, line 197: We'll call update twice if args.open is true (here and within setCurrentProject). Might be worth moving this into an else below and save us a few cycles?

ZachJW34: Good catch, added in [f2984b8](https://github.com/cypress-io/cypress/pull/21589/commits/f2984b8c598476a444c7212bac28d84f1ad663a9)

tgriesser (approved):

tbiethman (approved):

PR #21593, opened 2022-05-20
fix: Successfully create new specs files that do not have a known extension

tbiethman:
- Closes UNIFY-1787

### Additional details

When we process the proposed file name for a new spec created in the app, we try to find a known spec file extension (`.cy`, `.spec`, `_spec`, etc.) within in. We do this so that if the file name as provided is found to be a duplicate, we can update the name without mangling this extension, which would in all likelihood result in the file not matching the spec pattern. 

As reported in the JIRA issue, the file name would be generated incorrectly if it didn't end with one of our expected spec extensions. This was due to our falling back to the _file_ extension in the absence of an expected spec extension, inadvertently doubling the file extension in the resulting file name. I corrected this by instead falling back to nothing and letting the file parsing continue like normal.

### PR Tasks

### DIFF
--- a/packages/app/cypress/e2e/specs.cy.ts
+++ b/packages/app/cypress/e2e/specs.cy.ts
@@ -448,7 +448,7 @@ describe('App: Specs', () => {
           cy.contains('button', defaultMessages.createSpec.createSpec).should('not.be.disabled').click()
           cy.contains('h2', defaultMessages.createSpec.successPage.header)
 
-          cy.get('[data-cy="file-row"]').contains(getPathForPlatform('src/e2e/spec.js')).should('be.exist')
+          cy.get('[data-cy="file-row"]').contains(getPathForPlatform('src/e2e/spec.js')).should('be.visible')
         })
       })
 
--- a/packages/app/cypress/e2e/specs.cy.ts
+++ b/packages/app/cypress/e2e/specs.cy.ts
@@ -433,7 +433,9 @@ describe('App: Specs', () => {
             await ctx.actions.file.writeFileInProject('cypress.config.js', config)
           })
 
-          cy.contains('No Specs Found').should('be.visible')
+          // Timeout is increased here to allow ample time for the config change to be processed
+          cy.contains('No Specs Found', { timeout: 10000 }).should('be.visible')
+
           cy.findByRole('button', { name: 'New Spec' }).click()
           cy.contains('Create new empty spec').click()
 
@@ -708,7 +710,8 @@ describe('App: Specs', () => {
           await ctx.actions.file.writeFileInProject('cypress.config.js', config)
         })
 
-        cy.contains('No Specs Found').should('be.visible')
+        // Timeout is increased here to allow ample time for the config change to be processed
+        cy.contains('No Specs Found', { timeout: 10000 }).should('be.visible')
         cy.findByRole('button', { name: 'New Spec' }).click()
 
         cy.findByRole('dialog', {
--- a/packages/app/cypress/e2e/specs.cy.ts
+++ b/packages/app/cypress/e2e/specs.cy.ts
@@ -420,6 +420,34 @@ describe('App: Specs', () => {
 
           cy.visitApp().get('[data-cy-row]').contains('MyTest.cy.js')
         })
+
+        it('generates spec with file name that does not contain a known spec extension', () => {
+          cy.withCtx(async (ctx) => {
+            let config = ctx.actions.file.readFileInProject('cypress.config.js')
+
+            config = config.replace(
+                `specPattern: 'src/**/*.{cy,spec}.{js,jsx}'`,
+                `specPattern: 'src/e2e/**/*.{js,jsx}'`,
+            )
+
+            await ctx.actions.file.writeFileInProject('cypress.config.js', config)
+          })
+
+          cy.contains('No Specs Found').should('be.visible')
+          cy.findByRole('button', { name: 'New Spec' }).click()
+          cy.contains('Create new empty spec').click()
+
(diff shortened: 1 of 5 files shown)


packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -371,13 +355,11 @@ export class ProjectActions {
         : codeGenCandidate
     }
 
-    const specFileExtension = getFileExtension()
     const codeGenPath = getCodeGenPath()
 
     const newSpecCodeGenOptions = new SpecOptions(this.ctx, {
       codeGenPath,
       codeGenType,
-      specFileExtension,
```
tbiethman, line 380: I moved the specFileExtension logic into SpecOptions to:
1. co-locate it with the relevant logic
2. make it a little more testable

packages/data-context/src/codegen/spec-options.ts
```diff
@@ -6,10 +6,18 @@ import type { CodeGenType } from '@packages/graphql/src/gen/nxs.gen'
...
 }
 
+// Spec file extensions that we will preserve when updating the file name
+// due the existence of duplicate files.
+//
+// Example:
+//   Button.cy.ts   -> Button-copy-1.cy.ts
+//   Button_spec.js -> Button-copy-1_spec.js
+//   Button.foo.js  -> Button.foo-copy-1.js
+export const expectedSpecExtensions = ['.cy', '.spec', '.test', '-spec', '-test', '_spec']
```
tbiethman, line 19: While the immediate problem is fixed, the existence of this set is problematic. It's still possible for users with a custom spec pattern to create a new file and have it not match their spec pattern after our '-copy-x' is injected. We're getting into edge-case, explicitly recommended against patterns here, but ideally it'd still work.

If we want this to be fool-proof, we'll need to detect duplicated filenames prior to submission so that the user can intelligently navigate this. Once we reach this point in the lifecycle, we're shooting in the dark.

packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -346,22 +346,6 @@ export class ProjectActions {
...
-
-    const getFileExtension = () => {
-      if (erroredCodegenCandidate) {
-        return ''
-      }
-
-      return (
-        possibleExtensions.find((ext) => {
-          return codeGenCandidate.endsWith(ext + parsed.ext)
-        }) || parsed.ext
```
tbiethman, line 361: This `|| parsed.ext` fallback was the root of the issue and was replaced with `|| ''` in the moved logic here: https://github.com/cypress-io/cypress/pull/21593/files#diff-1dbda8880bf7cf51476b46c4e8344284db3c540b89672fe2dc63b3d61ae8228fR45

packages/data-context/src/codegen/spec-options.ts
flotwig: Do you want to open a task to track this improvement? I feel like we'll see a few reports about this...

packages/data-context/test/unit/codegen/spec-options.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+    ctx = createTestDataContext()
+
+    ctx.update((s) => {
+      s.currentProject = tmpPath
+    })
+  })
+
+  describe('getCodeGenOptions', () => {
+    context('unique file names', () => {
+      for (const specExtension of expectedSpecExtensions) {
```
flotwig, line 25: nit: I like to add an additional assertion that the expected tests run when generating the suite like this, since otherwise a regression in `expectedSpecExtensions` could possibly not cause a failure. Even something like `expect(expectedSpecExtensions).to.deep.eq(['.cy', '.spec', '.test', '-spec', '-test', '_spec'])`. But I guess this could also be construed as testing the implementation, not the feature, so it's just a thought.

flotwig (approved):

marktnoonan (approved):
Tested, works well. I noticed some alignment stuff in our warning that could use a tweak. But it's unrelated to the PR, I'll make a ticket.

Left one other non-blocking comment.

Screen Shot 2022-05-23 at 12 32 40 PM

packages/app/cypress/e2e/specs.cy.ts
```diff
@@ -420,6 +420,36 @@ describe('App: Specs', () => {
...
+          cy.findAllByLabelText(defaultMessages.createSpec.e2e.importEmptySpec.inputPlaceholder)
+          .as('enterSpecInput')
+
+          cy.get('@enterSpecInput').invoke('val').should('eq', getPathForPlatform('src/e2e/spec.js'))
+          cy.contains(defaultMessages.createSpec.e2e.importEmptySpec.invalidSpecWarning).should('not.exist')
+
+          cy.contains('button', defaultMessages.createSpec.createSpec).should('not.be.disabled').click()
+          cy.contains('h2', defaultMessages.createSpec.successPage.header)
+
+          cy.get('[data-cy="file-row"]').contains(getPathForPlatform('src/e2e/spec.js')).should('be.exist')
```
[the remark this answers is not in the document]
tbiethman: Updated to `be.visible`

packages/data-context/test/unit/codegen/spec-options.spec.ts
tbiethman: Agreed, added a test to validate the array is as we expect.

packages/data-context/src/codegen/spec-options.ts
tbiethman: Logged https://cypress-io.atlassian.net/browse/UNIFY-1810

PR #21598, opened 2022-05-23
chore: compare `cy.screenshot` images in percy

marktnoonan:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1566

The main change here is a circle.yml update, to use Percy to diff the screenshot images generated using the `cy.screenshot()` command from [screenshot.cy.tsx](https://github.com/cypress-io/cypress/blob/e1403d4bc5d5b273ad41922f256a8a03294eefbf/packages/app/src/runner/screenshot/screenshot.cy.tsx#L141). We do this currently on develop, so this ticket was just catch up. 

When I did get these uploads working, there were some differences to investigate. Most were are result of global styles and only required changes to test code, not to the app itself. 

After those changes to the test code, images only contained some differences in the presence/absences of scrollbars, so this PR includes changes to the app around overflow behavior, to fix those differences. 

In addition to these Percy snapshots, I used ImageMagick to compare images generated by the screenshot system tests on both branches, confirming that none of these changes have altered those images.

### User facing changelog

Scrollbars in screenshots will now be present/absent in the same situations in 9.x and 10.0. There were some situations where this previously was not the case in 10.0.

### Additional details

The rest of the PR is:

- a change to the screenshot.cy.tsx test itself to remove global styles. It's a component test, so it mounts in the context of the global app styles. Since global styles are different between 9.x and 10.0, there was some noise in the initial diffs caused by font and line height differences at the global level. Removing the 10.0 global styles in this test makes it much easier to diff with 10.0-release.
- Some changes to overflow styles to bring the presence/absence of scrollbars into line between these screenshots when taken from 10.0 and 9.x. These differences hadn't shown up before due to the images being the same, apart from scrollbars. These changes involve taking a similar approach to the current `is-screenshotting` body class to add an `screenshot-scrolling` class. This allows us to conditionally add/remove CSS rules related to overflow without adding/removing classes with JavaScript in the nested Vue components. When managing the conditional classes in the Vue components, the styles were modified too late for the screenshot.
- There's a tweak to the overflow of the `specs-list-container` which fixes another edge case that only appeared when scrollbars are present.

Since the latest Percy diff will compare this branch against 10.0-release, I've captured some diffs here between these screenshots on this branch and develop. What can be seen in these screenshots is a 1px shift to the left of all the content, since on 10.0-release (in [this PR](https://github.com/cypress-io/cypress/pull/21019)) we have previously fixed a 1px border bug in component test screenshots.

That means the 1px shifts, shown here in the diff with develop, are expected:

percy diff 4

percy diff 3

percy diff 2

percy diff 1

No UX changes.

### PR Tasks


packages/app/src/pages/Specs/Runner.vue
```diff
@@ -147,6 +147,24 @@ iframe.aut-iframe {
...
+
+#resizable-panels-root {
+  overflow-x: auto;
+  overflow-y: hidden;
+}
+
+.is-screenshotting #resizable-panels-root {
+  overflow-x: visible;
+  overflow-y: visible;
+}
```
marktnoonan, line 166: I'd actually like to consolidate more of our screenshot css rules into this structure, instead of conditionally managing tailwind classes with JS. But to keep this PR focused, I'm not doing that here, just moving what's needed.

packages/app/src/runner/ResizablePanels.vue
```diff
@@ -1,9 +1,10 @@
 <template>
   <div
-    class="flex overflow-y-hidden"
+    id="resizable-panels-root"
+    class="flex"
     :class="{
       'select-none': panel1IsDragging || panel2IsDragging,
-      'overflow-x-auto': !isFirefox
+      'overflow-x-hidden': isFirefox
```
marktnoonan, line 7: this lets the `!important` that comes with the utility class override the otherwise-preferred overflow.

packages/app/src/runner/SpecRunnerOpenMode.vue
```diff
@@ -13,7 +13,7 @@
     />
     <ResizablePanels
       v-else
-      :style="{width: `calc(100vw - ${collapsedNavBarWidth}px)`}"
+      :style="{width: `calc(100vw - ${screenshotStore.isScreenshotting ? 0 : collapsedNavBarWidth}px)`}"
```
marktnoonan, line 16: This one was just a loose end with open-mode screenshots.

packages/app/src/runner/screenshot/screenshot.cy.tsx
```diff
@@ -37,6 +37,16 @@ const Layout: FunctionalComponent = () => {
...
 const captureTypes = ['fullPage', 'viewport', 'runner'] as const
 
+function removeGlobalStyles () {
+  cy.get('style').each((item) => {
+    if (item[0].dataset.cy !== 'injected-style-tag') {
+      item.remove()
+    }
+  })
+
+  return cy.get('style').should('have.length', 1)
```
marktnoonan, line 47: This now means that only the existing CSS injected by the component will be applied. Other CSS targeting body, HTML, global reset CSS etc will not apply here. 

We may want to change this later, but for now this gives us the clearest diff with 10.0-release.

packages/app/src/specs/InlineSpecListTree.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div
     v-bind="containerProps"
-    class="pt-8px specs-list-container overflow-y-auto overflow-x-hidden"
+    class="pt-8px specs-list-container overflow-hidden"
```
marktnoonan, line 4: Since the scrolling in the specs list takes place within a virtual list, the overflow-x value of `auto` was causing a weird empty space at the bottom of the parent, to do with how the `vh` works in the `specs-list-container` height calculation below.

packages/app/src/runner/screenshot/screenshot.cy.tsx
rachelruderman: >This now means that only the existing CSS injected by the component will be applied. 
This comment has great context!! What do you think about including it in the code?

packages/app/src/runner/screenshot/screenshot.cy.tsx
```diff
@@ -45,7 +55,9 @@ describe('screenshot', () => {
         styles,
       })
 
-      cy.screenshot(`percy/component_testing_takes_a_screenshot_viewport_${capture}`, { capture })
+      removeGlobalStyles().then(() => {
+        cy.screenshot(`percy/component_testing_takes_a_screenshot_viewport_${capture}`, { capture })
+      })
```
rachelruderman, line 60:
```suggestion
      removeGlobalStyles()
      cy.screenshot(`percy/component_testing_takes_a_screenshot_viewport_${capture}`, { capture })
```

AFAIK you don't need `.then` because Cypress runs promises sequentially and you're not using the return value from `removeGlobalStyles` within `cy.screenshot`

tbiethman (approved):
I compared the generated screenshots to what's in develop, everything lines up how we'd expect. Also validated that they're showing up in Percy.

Great work on this @marktnoonan. It's nice to have regression coverage here, puts us in a good spot to refactor/condense some of these styles going forward.

packages/driver/src/cy/commands/screenshot.ts
```diff
@@ -98,6 +98,10 @@ const scrollOverrides = (win, doc) => {
   // hide scrollbars
   doc.documentElement.style.overflow = 'hidden'
 
+  // this body class is used to set some other overflow-related CSS
+  // around the resizable panels in the Runner
+  document.querySelector('body')?.classList.add('screenshot-scrolling')
```
ZachJW34, line 103: Not sure why TS is happy with this, if `document.querySelector('body')?` is ever null (which it shouldn't be) it would be safer to use `document.querySelector('body')?.classList?.add('screenshot-scrolling')`

I would just do `document.body.classList.add('screenshot-scrolling')`

packages/app/src/runner/screenshot/screenshot.cy.tsx
```diff
@@ -134,8 +150,11 @@ describe('screenshot', () => {
       }
 
       mount(() => <Comp />, { style }).then(() => {
-        cy.screenshot(`percy/large_component_hardcoded_size_viewport_${viewport[0]}_${viewport[1]}`, { capture: 'viewport' })
-        cy.screenshot(`percy/large_component_hardcoded_size_fullPage_${viewport[0]}_${viewport[1]}`, { capture: 'fullPage' })
+        removeGlobalStyles().then(() => {
+          cy.pause()
```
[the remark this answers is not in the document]
ZachJW34: bump

ZachJW34 (commented):
Other than the pause looks good to me!

ZachJW34 (approved):

packages/driver/src/cy/commands/screenshot.ts
marktnoonan: I got warnings in vscode when I did plain `document.body` so not sure what's up.

PR #21599, opened 2022-05-23
fix: issue with compilation failures in component testing

ryanthemanuel:
- Closes UNIFY-1437

### User facing changelog

Component specs with webpack compilation errors no longer exit the server process in run mode, their results get sent to the dashboard, and the exit code is specified properly. Component specs with vite compilation errors no longer hang in run or open mode.

### Additional details

Each framework handles compilation and linting errors slightly differnently. This PR focuses on compilation errors. An explanation of the current, proposed interim (on this PR), and proposed ideal states can be found in [this google doc](https://docs.google.com/spreadsheets/d/1sCHVKhAntzEVdkvoVGLz5Kieg-ImuoTb9Ghj6N8GirA/)

## Compilation Errors when using the Webpack Framework  
  
The main problem with this scenario is that we currently emit a webpack failure on compilation errors and in one of the handlers, we immediately exit the server process. This doesn't allow us to return the appropriate status codes, report to the dashboard appropriately, or continue running specs that don't have compilation errors.
  
The interim solution to this problem is to no longer exit the process. Webpack automatically injects the compilation error into the problematic module and Cypress's generic error mechanism handles this correctly.
  
Longer term, we could investigate displaying a custom error for this scenario rather than the generic uncaught exception message.

## Compilation Errors when using the Vite Framework
  
The main problem with this scenario is that esbuild errors out during the optimization process and assets are not generated.
The interim solution to this problem is to add an esbuild plugin that changes the errors to warnings. Vite will then not end up generating an asset for the bad asset and we will get a 500 error whenever the asset is loaded. This will end up getting this scenario to an acceptable state for both open and run mode, however it won't allow for the compilation error message to be displayed to the user. Instead it will be a more generic error.
  
In the long run, we will try and make the error more user friendly and specific. Some investigation will be needed to figure out:
* How to inject the compilation error in the problematic asset in such a way that Cypress can catch it and display it to the user in a non-generic way
* Can this be done with the vite-dev-server? Or do we need to try and create a static compilation in run mode? Initial investigation into this seemed to cause a stall in starting testing, so we will need to watch performance. If we do this, how do we solve open mode?

### PR Tasks


npm/webpack-dev-server/src/makeDefaultWebpackConfig.ts
```diff
@@ -41,5 +41,24 @@ export function makeDefaultWebpackConfig (
...
+  } as any
+
+  if (config.sourceWebpackModulesResult.webpackDevServer.majorVersion === 4) {
+    return {
+      ...finalConfig,
+      devServer: {
+        client: {
+          overlay: false,
+        },
+      },
```
ZachJW34, line 53:
```suggestion
      devServer: {
        ...finalConfig.devServer,
        client: {
          ...finalConfig?.devServer?.client
          overlay: false,
        },
      },
```

The way we are merging would overwrite any options specified by the user, so we should do this in a way that preserves these values.

Would also be good to check if we can swap the spread of the client to see if we can preserve the overlay if the user explicitly sets this (most configs should have a null value here and it gets defaulted to true later but not sure on this)

Same for v3 below

system-tests/project-fixtures/react/src/App.cy.jsx
```diff
@@ -4,5 +4,5 @@ import { App } from './App'
 
 it('renders hello world', () => {
   mount(<App />)
-  cy.get('h1').contains('Hello World')
+  cy.get('h1').contains('Hello World').click()
```
ZachJW34, line 7: Do we need this `click()`?

system-tests/lib/system-tests.ts
```diff
@@ -296,12 +296,13 @@ Bluebird.config({
   longStackTraces: true,
 })
 
+let cacheNodeModulesDir
```
ZachJW34, line 299: Why were these changes needed?

ZachJW34 (changes requested):
One change for the merging of the webpackConfig, the rest are minor. Tested and it works great, HUGE improvement

system-tests/project-fixtures/react/src/App.cy.jsx
ryanthemanuel: Yes. Without it, if we are displaying an overlay the tests will pass (even though they shouldn't). I can maybe throw a comment on there to make that more clear why I'm doing it.

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -14,14 +14,14 @@ const removeList = [
   // https://github.com/cypress-io/cypress/issues/15865
   'HtmlWebpackPlugin',
 
-  // This plugin is an opitimization for HtmlWebpackPlugin for use in
-  // production environments, not relevent for testing.
+  // This plugin is an optimization for HtmlWebpackPlugin for use in
```
BlueWinds, line 17: I know how to spell, I swear. >.>

npm/webpack-dev-server/test/makeWebpackConfig.spec.ts
```diff
@@ -19,6 +19,9 @@ describe('makeWebpackConfig', () => {
         output: {
           publicPath: '/this-will-be-ignored',
         },
+        devServer: {
+          overlay: true, // This will be ignored as well
```
BlueWinds, line 23:
```suggestion
          overlay: true, // This will be overridden by makeWebpackConfig.ts
```

system-tests/project-fixtures/react/src/App.cy.jsx
```diff
@@ -4,5 +4,5 @@ import { App } from './App'
 
 it('renders hello world', () => {
   mount(<App />)
-  cy.get('h1').contains('Hello World')
```
BlueWinds, line 7: Why are we clicking on hello world?

npm/webpack-dev-server/test/devServer-e2e.spec.ts
```diff
@@ -179,74 +178,6 @@ describe('#devServer', () => {
     })
   })
 
-  it('emits dev-server:compile:error event on error compilation', async () => {
```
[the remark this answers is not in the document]
ryanthemanuel: So there still is a `dev-server:compile:success` event. It's just in a different spot, but as I was thinking about this, I like having all of the compilation hooks in one place (they were definitely spread out and confusing before). With how everything works together we don't need the error events in vite or in webpack (however, that could change as this evolves in the future).

Unified the location of the compilation hooks here:

[`fd175c8` (#21599)](https://github.com/cypress-io/cypress/pull/21599/commits/fd175c863f92b811319ae4792a1ee940cc8a3e2f)

npm/webpack-dev-server/test/makeWebpackConfig.spec.ts
ryanthemanuel: Updated: [`fd175c8` (#21599)](https://github.com/cypress-io/cypress/pull/21599/commits/fd175c863f92b811319ae4792a1ee940cc8a3e2f)

system-tests/project-fixtures/react/src/App.cy.jsx
ryanthemanuel: I want to ensure that the component is interactable (i.e. there's no overlay) I added a comment for this: [`fd175c8` (#21599)](https://github.com/cypress-io/cypress/pull/21599/commits/fd175c863f92b811319ae4792a1ee940cc8a3e2f)

npm/webpack-dev-server/src/CypressCTWebpackPlugin.ts
```diff
@@ -93,42 +93,6 @@ export class CypressCTWebpackPlugin {
     callback()
   }
 
-  /*
```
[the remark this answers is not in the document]
ryanthemanuel: See my comment below: https://github.com/cypress-io/cypress/pull/21599#discussion_r881999136

npm/vite-dev-server/src/resolveConfig.ts
```diff
@@ -46,6 +46,23 @@ export const createViteDevServerConfig = async (config: ViteDevServerConfig, vit
...
     configFile,
     optimizeDeps: {
+      esbuildOptions: {
+        incremental: true,
+        plugins: [
+          {
+            name: 'esbuild-plugin',
+            setup (build) {
+              build.onEnd(function (result) {
+                // We don't want to completely fail the build here on errors so we treat the errors as warnings
```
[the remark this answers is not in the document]
ryanthemanuel: Added here: [`fd175c8` (#21599)](https://github.com/cypress-io/cypress/pull/21599/commits/fd175c863f92b811319ae4792a1ee940cc8a3e2f)

npm/vite-dev-server/cypress/e2e/react.cy.ts
```diff
@@ -56,5 +64,53 @@ for (const project of VITE_REACT) {
...
+      })
+
+      cy.waitForSpecToFinish()
+      cy.get('.passed > .num').should('contain', 1)
+
+      // Cause the problem again
+      cy.withCtx(async (ctx) => {
+        await ctx.actions.file.writeFileInProject(
+          `src/AppCompilationError.cy.jsx`,
+          `
```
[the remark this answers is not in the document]
ryanthemanuel: Updated to use dedent on the latest commit

system-tests/lib/system-tests.ts
ryanthemanuel: Good catch. These were needed at one point to standardize output, but I refactored a few things and missed these. They are now removed:

[`fd175c8` (#21599)](https://github.com/cypress-io/cypress/pull/21599/commits/fd175c863f92b811319ae4792a1ee940cc8a3e2f)

system-tests/project-fixtures/react/src/App.cy.jsx
ryanthemanuel: Comment added here: [`fd175c8` (#21599)](https://github.com/cypress-io/cypress/pull/21599/commits/fd175c863f92b811319ae4792a1ee940cc8a3e2f)

npm/webpack-dev-server/src/makeDefaultWebpackConfig.ts
ryanthemanuel: This is just modifying the default config (defined on line 27) before we've even added anything from the user-specified options, so we should be fine doing this. I did add a couple of tests to make sure that user specified dev server options are honored (except the overlay) here: [`fd175c8` (#21599)](https://github.com/cypress-io/cypress/pull/21599/commits/fd175c863f92b811319ae4792a1ee940cc8a3e2f)

packages/server/lib/modes/run-ct.ts
```diff
@@ -1,25 +1,9 @@
 import type { LaunchArgs } from '@packages/types'
 
-import { openProject } from '../open_project'
-
 export const run = (options: LaunchArgs, loadingPromise: Promise<void>) => {
   // TODO: make sure if we need to run this in electron by default to match e2e behavior?
   options.browser = options.browser || 'electron'
   options.runAllSpecsInSameBrowserSession = true
 
-  require('../plugins/dev-server').emitter.on('dev-server:compile:error', (error: Error) => {
```
tgriesser, line 10: 🔥 good riddance

BlueWinds (approved):

ZachJW34 (approved):

tgriesser (approved):
Changes & rationale makes sense - good test coverage and it looks like the other review comments have been addressed, fine with me as long as tests look good 👍

system-tests/project-fixtures/react/src/App.cy.jsx
JessicaSachs: `.should('be.visible')` btw. Though `.click()`'s actionability checks are more aggressive (and will cover more)

npm/vite-dev-server/cypress/support/e2e.ts
```diff
@@ -17,4 +17,4 @@
 import '@packages/frontend-shared/cypress/e2e/support/e2eSupport'
 
 // Alternatively you can use CommonJS syntax:
-// require('./commands')
+require('./commands')
```
JessicaSachs, line 20: I don't like mixing these. Why would you require instead of import? Not gonna block but stick to one. require is nonstandard in the browser.

ryanthemanuel: I can update this. I had just uncommented what was commented.

ryanthemanuel: Updated here: [`dd566a0` (#21599)](https://github.com/cypress-io/cypress/pull/21599/commits/dd566a0562093ca231c4270c3cc76d873e013833)

npm/webpack-dev-server/test/makeWebpackConfig.spec.ts
```diff
@@ -38,4 +43,47 @@ describe('makeWebpackConfig', () => {
...
+        projectRoot: '.',
+        supportFile: '/support.js',
+        devServerPublicPathRoute: '/test-public-path', // This will be overridden by makeWebpackConfig.ts
+      } as Cypress.PluginConfigOptions,
+      webpackConfig: {
+        output: {
+          publicPath: '/this-will-be-ignored',
+        },
+        devServer: {
+          magicHtml: true,
```
JessicaSachs, line 61: This was added in Webpack 4.1 only. Will we break user config for Webpack 4? Is there a sys test for Webpack 4.0?

packages/server/lib/modes/run-ct.ts
JessicaSachs: god bless

JessicaSachs (commented):
Concerned about `magicHtml` usage. What's the deal with that?

npm/webpack-dev-server/test/makeWebpackConfig.spec.ts
ryanthemanuel: @JessicaSachs this is a unit test. I'm just verifying that user defined config will be kept.

JessicaSachs: Yeah I saw that was just like "why **this** flag?". Cool.

PR #21617, opened 2022-05-24
fix: handle custom (default) integration folder for migration

estrada9166:
- Closes  [UNIFY-1784](https://cypress-io.atlassian.net/browse/UNIFY-1784)

### Additional details

If the integrationFolder was set on the legacy config as `cypress/integration` which is the default value, we do not want to add it to the migrated config which should handle it as the new default one

### PR Tasks


tbiethman (approved):
Works great and adds good test coverage  🚀

marktnoonan (approved):
Tested locally, works great.

PR #21620, opened 2022-05-24
test(launchpad): skip failure due to recaptcha on windows, enable more windows jobs

flotwig:
- Closes UNIFY-1813

### User facing changelog

N/A

### Additional details

- Was only able to reproduce in CircleCI, not on a regular Windows Server VM.
- Looking at the requests, it seems like ReCAPTCHA requests are the last things to go out, and then a `vimeocdn` request is never sent.
- My assumption is that ReCAPTCHA is preventing us from loading Vimeo, so we can skip this on Windows.
	- I'm not sure what else could be done. Maybe we can hire someone to solve CAPTCHAs for our CI pipeline as an alternative? :smile: 
- Manually tested Windows launchpad and the video does work.
- Enables Windows app and launchpad tests on all PRs now that they pass.

### PR Tasks


packages/launchpad/cypress/e2e/migration.cy.ts
```diff
@@ -161,9 +161,13 @@ describe('Opening unmigrated project', () => {
 
     // Vimeo's implementation may change, this is just a high level check that
     // the expected iframe code is being returned and that there is vimeo-related network traffic
-    cy.get('[data-cy="video-container"] iframe[src*=vimeo]').should('be.visible')
+    cy.get('[data-cy="video-container"] iframe[src*=vimeo]').as('videoContainer').should('be.visible')
```
tbiethman, line 164: Looks like we don't need the alias?

tbiethman (approved):

packages/launchpad/cypress/e2e/migration.cy.ts
flotwig: whoops, thanks
```suggestion
    cy.get('[data-cy="video-container"] iframe[src*=vimeo]').should('be.visible')
```

mschile (approved):

marktnoonan (approved):

PR #21621, opened 2022-05-24
chore: clean up `debug` statements in preparation for 10.0 release, add `debug` docs

flotwig:
### User facing changelog

If we want to consider this user-facing:

* Removed some verbose `DEBUG=cypress:*` lines and moved others to the `DEBUG=cypress-verbose:*` namespace.

### Additional details

- I noticed that we have added a lot of debug lines to our already-bloated `DEBUG` namespaces.
- With 10.0, there are going to be a lot of changes and potentially a lot of bug reports, so it's important that we take the opportunity to clean up `DEBUG` and make it easier to read and more useful before we start getting reports with `DEBUG` logs in them.
- This PR:
  - Documents the way that `debug` namespaces are handled in `cypress`. Previously, the way we named debug namespaces was not documented.
  - Removes or moves to `cypress-verbose` a lot of `debug` statements, some added in 10.0, some existing in 9.x
    - Of note, `packages/launcher` is now using namespaces instead of just `cypress:launcher`
  - Updates some namespaces to follow the written standard

`debug` output is easier to parse and make sense of.

### PR Tasks

### DIFF
--- a/guides/debug-logs.md
+++ b/guides/debug-logs.md
@@ -19,7 +19,7 @@ cypress:{packageName}:{relative path to file from src root, using : to separate 
 Exceptions to these rules:
 * The `cli` uses `cypress:cli:*`.
 * NPM packages should use `{moduleName}` as a prefix instead of `cypress`, like `cypress-webpack-preprocessor` for `npm/webpack-preprocessor`.
-* In some places, like the proxy package, it's more useful to attach `debug` messages to something besides the module. In that case, it's okay to create namespaces as you see fit. But at least begin with `cypress:{packageName}` or `cypress-verbose:{packageName}`
+* In some places, like per-request in the `proxy` package, it's more useful to attach `debug` messages to something besides the module (like individual HTTP requests). In that case, it's okay to create namespaces as you see fit. But at least begin with `cypress:{packageName}` or `cypress-verbose:{packageName}`
 
 ## Using debug logs
 
--- a/guides/debug-logs.md
+++ b/guides/debug-logs.md
@@ -1,0 +1,42 @@
+# Debug Logs
+
+Many Cypress packages use the [`debug` module][debug] to log runtime info to the console.
+
+## Choosing a namespace
+
+The naming scheme for debug namespaces should generally follow this pattern:
+
+```
+cypress:{packageName}:{relative path to file from src root, using : to separate directories, minus index if applicable}
+
+# examples:
+# packages/server/lib/util/file.js -> cypress:server:util:file
+# packages/launcher/windows/index.ts -> cypress:launcher:windows
+```
+
+`cypress-verbose` can be used instead of `cypress` if the logs are overly verbose and would make the output of `DEBUG=cypress:*` unreadable.
+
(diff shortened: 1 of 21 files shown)


guides/debug-logs.md
```diff
@@ -0,0 +1,42 @@
+# Debug Logs
+
+Many Cypress packages use the [`debug` module][debug] to log runtime info to the console.
+
+## Choosing a namespace
+
+The naming scheme for debug namespaces should generally follow this pattern:
+
+```
+cypress:{packageName}:{relative path to file from src root, using : to separate directories, minus index if applicable}
```
tgriesser, line 10: I wonder if in the future we can do this automatically, similar to what we do [in the dashboard](https://github.com/cypress-io/cypress-services/blob/9d8899e5c68f2221f81b0a5a79bf1e659d313d71/packages/debug/src/debug.ts#L21-L37)

```
import { makeDebug } from '@packages/debug'
const debug = makeDebug(__filename)
```

packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -326,16 +326,14 @@ export class GitDataSource {
     const stdout = result.stdout.split('\n')
 
     if (result.exitCode !== 0) {
-      debug(`error... stderr`, result.stderr)
+      debug(`error... %o`, result)
```
MuazOthman, line 329:
```suggestion
      debug(`command execution error: %o`, result)
```

packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -326,16 +326,14 @@ export class GitDataSource {
...
     }
 
     if (stdout.length !== absolutePaths.length) {
-      debug('error... stdout:', stdout)
+      debug('error... %o', result)
```
MuazOthman, line 333:
```suggestion
      debug('unexpected command execution result: %o', result)
```

MuazOthman (commented):
It looks much cleaner now. Thanks!

guides/debug-logs.md
flotwig: Yeah, I would love to do this in this repo too so the burden isn't on individual contributors and reviewers to know these guidelines.

tgriesser:
> Do you think this utility will remain long-term?

@flotwig I think we should remove it, it gives such a silly amount of logs when you do `DEBUG=cypress*` that it should probably be removed. @ryanthemanuel I saw you had added the comment for it, have you found any use in it with anything you've needed to work on?

ryanthemanuel:
> > Do you think this utility will remain long-term?
> 
> @flotwig I think we should remove it, it gives such a silly amount of logs when you do `DEBUG=cypress*` that it should probably be removed. @ryanthemanuel I saw you had added the comment for it, have you found any use in it with anything you've needed to work on?

I actually used it a fair amount when I was trying to debug a few race condition type issues with the lifecycle manager. Maybe we can namespace it even further so that it doesn't show up with `cypress*`, something like `private-cypress-trace` or something?

PR #21629, opened 2022-05-25
fix(cypress/schematic): run configuration does not work anymore without baseUrl 

### LINKED ISSUE #21555: @cypress/schematic: Run configuration does not work anymore without baseUrl
### Current behavior

After upgrading from `1.6.0` to `1.7.0` all tests stopped working because `cy.visit('/')` is trying to access a file instead of a URL. Now the `baseUrl` has to be defined additionally in order to run the tests.

### Desired behavior

Cypress takes the host and the port of the `devServerTarget`. The host and the port should be defined once and not twice. There should be a reference instead.

## Current workaround:
Add `baseUrl: http://localhost:{myPort}` to the `cypress-run` and `cypress-open` options in `angular.json` or to `cypress.json`

### Test code to reproduce

angular.json:
```json
...
      "architect": {
        "build": {
          "builder": "@angular-devkit/build-angular:browser",
          "options": {
            ...
          },
          "configurations": {
            "test": {
              ...
            }
          },
          "defaultConfiguration": ""
        },
        "serve-test": {
          "builder": "@angular-devkit/build-angular:dev-server",
          "options": {
            "browserTarget": "myProject:build:test",
            "port": 5000
          },
        },
        "cypress-open": {
          "builder": "@cypress/schematic:cypress",
          "options": {
            "devServerTarget": "myProject:serve-test",
            "watch": true,
            "headless": false,
            "browser": "chrome"
          },
          "configurations": {
            "production": {
              "devServerTarget": "myProject:serve-test"
            }
          }
        }
...
```

### Cypress Version

9.6.1

### Other

Angular core: `12.2.16`
cli, @angular-devkit/build-angular: `12.2.17`

admah:
Fixes #21555.

Users can now add custom options to `devServerTarget` in `angular.json` to specify `baseUrl`:

```json
"options": {
  "host": "localhost",
  "port": 4200
},
```

### DIFF
--- a/npm/cypress-schematic/src/builders/cypress/cypressBuilderOptions.ts
+++ b/npm/cypress-schematic/src/builders/cypress/cypressBuilderOptions.ts
@@ -4,8 +4,6 @@ export interface CypressBuilderOptions extends JsonObject {
   baseUrl: string
   configFile: string
   browser: 'electron' | 'chrome' | 'chromium' | 'canary' | 'firefox' | 'edge' | string
-  devServerTarget: string
-  devServerBaseUrl: string
   env: Record<string, string>
   quiet: boolean
   exit: boolean
--- a/npm/cypress-schematic/src/builders/cypress/cypressBuilderOptions.ts
+++ b/npm/cypress-schematic/src/builders/cypress/cypressBuilderOptions.ts
@@ -4,6 +4,7 @@ export interface CypressBuilderOptions extends JsonObject {
   baseUrl: string
   configFile: string
   browser: 'electron' | 'chrome' | 'chromium' | 'canary' | 'firefox' | 'edge' | string
+  devServerTarget: string
   env: Record<string, string>
   quiet: boolean
   exit: boolean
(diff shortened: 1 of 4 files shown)


npm/cypress-schematic/src/builders/cypress/cypressBuilderOptions.ts
```diff
@@ -5,6 +5,7 @@ export interface CypressBuilderOptions extends JsonObject {
   configFile: string | false
   browser: 'electron' | 'chrome' | 'chromium' | 'canary' | 'firefox' | 'edge' | string
   devServerTarget: string
+  devServerBaseUrl: string
```
ZachJW34, line 8: This option would never be used since it is derived from the dev-server, so I think we can remove this.

admah: Good point!

ZachJW34: A general rule is that if the option is available in the `schema.json` it should be listed here as well, so I'd add back in the `devServerTarget` since you shouldn't have to cast the type in the builder.

ZachJW34 (approved):

admah:
@ZachJW34 what do we typically do with the Percy tests when they fail? Nothing I touched should have altered those in any way.

ZachJW34:
@admah Your changes have no impact on UI so we can ignore the Percy diffs. I reran the build that flaked, so let's get this merged once everything (besides Percy) is green

PR #21630, opened 2022-05-25
fix: UNIFY-1408, warnings should be nested appropriately & clearable

tgriesser:
- Closes UNIFY-1408, UNIFY-1368 - warnings persisting after dismissal, across testing types
- Closes UNIFY-1604, auto opening global mode project
- Closes UNIFY-1732, clicking into projects does not clear global error

### User facing changelog

Several minor bugfixes, mostly all related to how we're storing error & warning state

### Additional details
- Ensures that warnings are cleared via mutation when the "X" is clicked, rather than maintaining that state on the frontend
- Adds the concept of "layering" the warnings & errors with the part of the lifecycle they're related to, so when switching testing types / projects / browsers, warnings & errors are cleaned up appropriately
- Adds `errorWarningChange` subscription, emitted when error/warning state changes
- Adds concept of "shared" global subscriptions, for state changes that affect both apps

Bonus:
- While adding a test for the warnings & errors, fixed "opening" the project from global mode drop-zone (UNIFY-1604) and clearing the error when going back to the global projects list (UNIFY-1732)
- Fixes for a few other flaky tests

### Steps to test
- Trigger a warning, by having `experimentalStudio` set as a config prop, or a `baseUrl` on a port that is not running,  observe the warning can be cleared

### PR Tasks


packages/data-context/src/DataContext.ts
```diff
@@ -352,36 +367,60 @@ export class DataContext {
...
+      })
 
-      this.emitter.toLaunchpad()
-      this.emitter.toApp()
+      this.emitter.errorWarningChange()
     }
   }
 
-  onWarning = (err: CypressError) => {
+  onWarning = (err: CypressError, target?: 'testingType' | 'project') => {
```
ZachJW34, line 400: I see one instance of `target` being specified, but I removed it and didn't see any differences. Since we send all the warnings in an array, is this needed?

packages/launchpad/cypress/e2e/global-mode.cy.ts
```diff
@@ -314,4 +317,19 @@ describe('Launchpad: Global Mode', () => {
...
   })
+
+  describe('error state', () => {
+    it('should not persist the error when going back to the main screen projects', () => {
+      cy.openGlobalMode()
+      cy.addProject('config-with-import-error')
+      cy.addProject('todos')
+      cy.visitLaunchpad()
+      cy.contains('[data-cy="project-card"]', 'todos').should('be.visible')
+      cy.contains('[data-cy="project-card"]', 'config-with-import-error').should('be.visible')
```
ZachJW34, line 328: Can add `.click()` to the end of this line and remove the line below

ZachJW34 (commented):
Not sure if this is related to this PR, but when refreshing the config I'd expect that warnings are dismissed:


https://user-images.githubusercontent.com/25158820/170517771-e51d88c9-86c9-4de6-96c8-825634ca3488.mov

packages/launchpad/cypress/e2e/global-mode.cy.ts
tgriesser: Ah yeah I was hitting a detached element but it was due to another version

packages/data-context/src/DataContext.ts
tgriesser: Yeah it actually looking to guard that it doesn't get cleared if it wasn't supposed to, but agree it seems unneeded. Will remove

tgriesser:
@ZachJW34 fixed, and added more tests around the expected behavior

packages/data-context/src/actions/DevActions.ts
```diff
@@ -50,7 +49,7 @@ export class DevActions {
   }
 
   dispose () {
-    this._chokidar?.close()
+    this._chokidar?.close().catch(() => {})
```
tbiethman, line 52: 👍

ZachJW34 (approved):
Looks like there are some failing tests but I'm going to throw a ✅  so you can get this merged once they're addressed

tbiethman:
In global mode, seems like warnings are not presented when a project is re-opened. Seems like the issuedWarnings set may not be getting reset there? I guess that only happens when we destroy the data context.

https://user-images.githubusercontent.com/1711637/170582479-71e78a1c-9d25-4e47-b5b3-ee0b499292be.mov

tgriesser:
> In global mode, seems like warnings are not re-presented when a project is re-opened. Seems like the issuedWarnings set may not be getting reset there? I guess that only happens when we destroy the data context.

Hmm, good point, i guess we might want to reset that when we clear the project?

tbiethman (approved):
Nice, it's not every day you get to close four tickets with one PR.

PR #21632, opened 2022-05-25
fix: percy snapshots screenshotName should always be populated

rachelruderman:
- Closes UNIFY-1739

### User facing changelog
n/a

### Additional details
Before, when Percy snapshots were taken in the `afterEach` block, they all get the same name, resulting in the error below in CI.

### Steps to test
1. Look at Percy CI in this PR and confirm no errors in console for `SnapshotControls` component
2. Look at Percy CI in a different PR and confirm there are errors in the console for `SnapshotControls` component
3. Search through e2e and component tests in CircleCI to confirm there are no other Percy errors including `snapshot must be unique`
 
**Before:**
Screen Shot 2022-05-27 at 8 28 39 AM

**After:**
Screen Shot 2022-05-27 at 8 28 30 AM

### PR Tasks

- [na] Have tests been added/updated? -- unfortunately there are no existing tests for this file, would be good to add tests for the whole file in a separate PR


marktnoonan (changes requested):
Percy shows 256 snapshots as "new" for this PR. It looks like that's mostly due to the extra `>` characters added in this the snapshot names. 

For me the best way to confirm that this PR meets the goal of capturing snapshots that are currently missing due to these errors, is if we can see that all current snapshots are 1:1 and the previously absent snapshots have been added.

I do think the new names are more readable with the divider character, but since the names are what Percy uses to identify the snapshots over time, we will lose some history by doing this and should probably make that change separately.

If making the `afterEach` hooks work can be done conditionally where the `screenshotName` is generates, that's cool, or if we have to move the calls to `cy.percySnapshot` out of the hooks and into the `it` blocks, that would be fine.

packages/ui-components/cypress/support/customPercyCommand.ts
```diff
@@ -212,7 +212,9 @@ export const installCustomPercyCommand = ({ before, elementOverrides }: {before?
 
     const titlePath = test.titlePath()
 
-    const screenshotName = name ? titlePath.concat(name).filter(Boolean).join(' > ') : ''
```
MuazOthman, line 215: Do we know why we needed to use `filter(Boolean)`? it suggests that we suspected the array to have falsy values. If that still holds then maybe we need to keep it??

rachelruderman: I have a gut feeling we don't need it, I took it out in the most recent change to test my hypothesis 🙂 we'll find out when the CI results come through!

rachelruderman: Confirmed, no impact to the screenshotName!

**How to test:**
1. In the code look for `cy.percySnapshot` where name is being passed in, i.e. `cy.percySnapshot('tall viewport')`
2. In this PR, search for the Percy snapshot with "tall viewport"
3. In another PR, [i.e. this one](https://percy.io/cypress-io/cypress/builds/18459177/search/preview/1035275312?browser=chrome&searchParam=short%20viewport&viewLayout=side-by-side&viewMode=new&width=1500), search for the Percy snapshot with "tall viewport"
4. Confirm that the snapshot names are the same

MuazOthman (approved):

packages/ui-components/cypress/support/customPercyCommand.ts
MuazOthman: Documenting a conversation @rachelruderman and I had about this:

I’m basically wondering if there’s a scenario where `test.titlePath()` may return an array with some items being empty/null. The answer turned out to be yes, if we elected to pass in an empty string in `describe` or `it`. Rachel suggested there is value in ending up with `something > > something` in this case as it will help draw attention to the undesirable use of an empty string, encouraging developers to use a more proper value.

marktnoonan (approved):
Diff looks good, happy to see approx 40 snapshots from 3 separate specs now present that were getting swallowed by this error previously. Great that this adds the ability to take snapshots in `afterEach` hooks, since that is something people are moved to do anyway and this error is hard to catch if you don't know about it.

PR #21635, opened 2022-05-25
fix: Add hover states for test titles in reporter

BlueWinds:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1737

### User facing changelog
Hovering over test names now highlights the row to indicate clickability. Unifies styles for hovering over test name / hook name / command name.

### Additional details
Adds first component tests (primarily percy snapshots) to `packages/reporter`.

### Steps to test
1. After checking out branch, use `yarn dev` to recompile reporter css, otherwise you won't see the updates.
2. `cypress open` any project.
3. Select any configured testing type, launch any browser, run any spec.
4. Mouse over the command list - notice rows highlight under the mouse.

Old behavior:

Old Behavior

New behavior:

New Behavior

### PR Tasks

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -494,7 +494,7 @@ commands:
       package:
         description: package to target
         type: enum
-        enum: ['frontend-shared', 'launchpad', 'app']
+        enum: ['frontend-shared', 'launchpad', 'app', 'reporter']
       browser:
         description: browser shortname to target
         type: string
@@ -1132,6 +1132,7 @@ jobs:
             run-frontend-shared-component-tests-chrome,
             run-launchpad-component-tests-chrome,
             run-launchpad-integration-tests-chrome,
+            run-reporter-component-tests-chrome,
             run-webpack-dev-server-integration-tests,
             run-vite-dev-server-integration-tests
       - run:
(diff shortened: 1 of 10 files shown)


tbiethman:
Changes look great, but we need create a new `reporter-component-tests` job in circle.yml to run the new tests in CI.

packages/reporter/cypress/support/component.ts
```diff
@@ -0,0 +1,8 @@
+import { mount } from 'cypress/react'
+import 'cypress-real-events/support'
+import { installCustomPercyCommand } from '@packages/ui-components/cypress/support/customPercyCommand'
+
+import '../../src/main.scss'
+
+Cypress.Commands.add('mount', mount)
```
ZachJW34, line 7: You mentioned in slack there was an issue with 
```
declare global {
  namespace Cypress {
    interface Chainable {
      mount: typeof mount
    }
  }
}
```

I added this and included `cypress/support/component.ts` in the tsconfig and everything compiled fine, plus it removed some of the ts warnings.

ZachJW34 (approved):
The gifs are synchronized! Looks great, one small comment

packages/reporter/cypress/support/component.ts
BlueWinds: The issue with that was because I wasn't using our webpack configs (and therefor everything was broken, that just being one example). I'll add the type addition back in.

JessicaSachs (approved):

circle.yml
```diff
@@ -2324,6 +2340,11 @@ linux-workflow: &linux-workflow
         percy: true
         requires:
           - build
+    - run-reporter-component-tests-chrome:
+        context: [test-runner:launchpad-tests, test-runner:percy]
```
tbiethman, line 2344: I'm not super familiar with this setup, but I think you could:
1. Use the `test-runner:cypress-record-key` context here rather than `test-runner:launchpad-tests`
2. Update `run-new-ui-tests` to parameterize the CYPRESS_RECORD_KEY variable it uses (currently hardcoded to `$TEST_LAUNCHPAD_RECORD_KEY`, you want `$MAIN_RECORD_KEY` with the cypress-record-key context)

That might get you passed the current errors and keep the existing `reporter-integration-tests` build recording as-is.

PR #21653, opened 2022-05-27
fix: make ScriptError display at full height in Runner

marktnoonan:
- Closes https://cypress-io.atlassian.net/browse/UNIFY-1763

### User facing changelog

Corrects the height of the `ScriptError` component in Cypress 10. Not really a user facing change since this hasn't been released.

### Additional details

This component ended up too short as a result of changes to the surrounding elements in a previous PR. The change wasn't noticed at the time since there was not an e2e test for this state, and the component test wasn't affected by the DOM changes so it looked fine in Percy.

This PR corrects the height by changing some CSS classes, and uses the AUT header store to correctly react to wrapping in the header. It also adds a test that confirms the state appears, by adding a bad spec to the `cypress-in-cypress` project in `system-tests`. Adding a project there breaks a lot of assertions about the number of items in the specs list, so 3 spec files are updated to account for the new spec in the project folder. Even though it means updating several tests, I still think this is the most logical place to add the new spec file that triggers this error state.

### Steps to test

Run `yarn dev` to open the test runner for this branch, and visit an end-to-end spec that works. Make a syntax error in the spec that's open so that the ScriptError will appear. See `system-tests/projects/cypress-in-cypress/cypress/e2e/malformed-spec-file.spec.js` in this branch for an example of a bad test that causes this state. 

Once you see the error, please resize the reporter and the window, open and close the specs list etc. The error display area should always be fully visible, never get cut off and never get too small.

Before this PR:
script error display too short

After this PR:
script error

### PR Tasks


tbiethman (approved):
Tested as described, error renders with expected height 👍

ZachJW34 (approved):
Works great!

PR #21659, opened 2022-05-27
refactor: remove GitInfo from run mode

tgriesser:
- Closes N/A

### User facing changelog

N/A

### Additional details

While debugging some test flake, noticed there's a ton of noise generated by the `GitDataSource` in our own tests, which means there's enough overhead with using it to be noticeable. Since we don't have any product reasons (yet) to need to reference this info in CI, it seems like it'd be fine to skip using it if we're in run mode

### PR Tasks


packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -405,17 +405,19 @@ export class ProjectLifecycleManager {
...
-        onBranchChange: () => {
-          this.ctx.emitter.branchChange()
-        },
-        onGitInfoChange: (specPaths) => {
-          this.ctx.emitter.gitInfoChange(specPaths)
-        },
-      })
+      if (!this.ctx.isRunMode) {
+        s.currentProjectGitInfo = new GitDataSource({
+          isRunMode: this.ctx.isRunMode,
```
tgriesser, line 410: Will come back & update this if we decide on this change

flotwig (approved):

estrada9166 (approved):

PR #21665, opened 2022-05-27
fix(bug): indeterminate checkbox state doesn't get changed

### LINKED ISSUE #19098: Indeterminate checkbox state doesn't get changed via uncheck({ force: true })
### Current behavior

Cypress does not uncheck checkboxes that have indeterminate state.

### Desired behavior

Cypress should detect that it's an indeterminate checkbox, then click on it twice to uncheck.

### Test code to reproduce

```js
cy.get("input[aria-label='Toggle Select All Columns']")
  .uncheck({ force: true })
```

### Cypress Version

9.1.0.1027187

### Other

https://css-tricks.com/indeterminate-checkboxes/

kshastri:
- Closes #19098

### User facing changelog
Now the checkbox will show an appropriate checked or unchecked state and will not remain in indeterminate state on the browser UI. 

### Additional details
using .prop(indeterminate, false) here because as per the removeProp documentation it can lead to unexpected results and it was indeed not working as expected while I tried to use it. 

### Steps to test
2 tests have been added in check_spec.js 

1.  removes indeterminate prop when checkbox is checked
2. removes indeterminate prop when checkbox is unchecked

**Before**
image

**After**
image

### PR Tasks

### DIFF
--- a/packages/driver/src/cy/commands/actions/check.ts
+++ b/packages/driver/src/cy/commands/actions/check.ts
@@ -113,6 +113,13 @@ const checkOrUncheck = (Cypress, cy, type, subject, values: any[] = [], userOpti
           cy.ensureVisibility($el, options._log)
         }
 
+        // if the checkbox is in an indeterminate state, checking or unchecking should set the
+        // prop to false to move it into a "determinate" state
+        // https://github.com/cypress-io/cypress/issues/19098
+        if ($el.prop('indeterminate')) {
+          $el.prop('indeterminate', false)
+        }
+
         if (options._log) {
           const inputType = $el.is(':radio') ? 'radio' : 'checkbox'
 
--- a/packages/driver/cypress/integration/commands/actions/check_spec.js
+++ b/packages/driver/cypress/integration/commands/actions/check_spec.js
@@ -71,6 +71,21 @@ describe('src/cy/commands/actions/check', () => {
       })
     })
 
+    // https://github.com/cypress-io/cypress/issues/19098
+    it('removes indeterminate prop when checkbox is checked', () => {
+      const intermediateCheckbox = $(`<input type='checkbox' name="indeterminate">`)
+
+      intermediateCheckbox.prop('indeterminate', true)
+
+      $('body').append(intermediateCheckbox)
+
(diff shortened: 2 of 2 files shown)


kshastri:
@chrisbreiding, @BlueWinds - I have resolved the issue. Please have a look when you have time.

kshastri:
@[jennifer-shehane](https://github.com/jennifer-shehane) : Could you please get the 6 failing checks in CircleCI? The e2e test executed fine in my local for the test cases, I had added. Not sure why these 6 checks are failing.

packages/driver/src/cy/commands/actions/check.ts
```diff
@@ -113,6 +113,13 @@ const checkOrUncheck = (Cypress, cy, type, subject, values: any[] = [], userOpti
           cy.ensureVisibility($el, options._log)
         }
 
+        // if the checkbox is in an indeterminate state, checking or unchecking should set the
+        // prop to false to move it into a "determinate" state
+        // https://github.com/cypress-io/cypress/issues/19098
+        if ($el.prop('indeterminate')) {
+          $el.prop('indeterminate', false)
+        }
+
```
BlueWinds, line 122: So interesting note: `cy.uncheck()` on an indeterminate checkbox doesn't emit a click event.

At first I thought this was a bug, but the more I think about it, the more it seems like the correct behavior. Users can't uncheck indeterminate checkboxes - clicking on them checks them, which was a case that already worked as expected.

**Indeterminate checkboxes are already checked or unchecked**, and removing indeterminate on what would otherwise be a no-op is purely a developer QOL feature. There's no browser behavior to emulate, so not emitting any events seems correct.

BlueWinds (approved):

mjhenkes (approved):

BlueWinds:
@kshastri - Looks good to me. We're going to hold off merging this until after the 10.0.1 release tomorrow, but should get it in by end of week for 10.0.1.

kshastri:
> @kshastri - Looks good to me. We're going to hold off merging this until after the 10.0.1 release tomorrow, but should get it in by end of week for 10.0.1.

This is good to know. Thank you @BlueWinds

PR #21982, opened 2022-05-31
fix: Cypress version loading constantly

estrada9166:
- Closes  [UNIFY-1738](https://cypress-io.atlassian.net/browse/UNIFY-1738)

### Additional details

The push fragment was being called correctly, and the subscription as well, what was going on was that the test may be fast enough that the resolved subscription was not being updated before the test completed making the Percy snapshots be flaky.

Making this change as the comment on the code says, may help the test to be consistent 

### PR Tasks


marktnoonan:
@estrada9166 to confirm - this was also an issue outside of tests, sometimes the launchpad would start without this content in real usage. Though it looks like this would also fix real usage.

PR #21994, opened 2022-06-01
fix: scroll of inline specs list

marktnoonan:
### User facing changelog

In the Runner page, the inline specs list will now scroll as expected.

### Additional details

The overflow styles were changed incorrectly in [this PR](https://github.com/cypress-io/cypress/pull/21598) last week. Removing the `overflow-hidden` class restores the correct behavior of the specs list. The `!important` of the tailwind class clobbered the overflow that the `useVirtualList` sets on the wrapper of the virtual list.

### Steps to test

Visit a spec and confirm that the inline specs list is scrollable vertically.

### PR Tasks

### DIFF
--- a/packages/app/src/specs/InlineSpecList.cy.tsx
+++ b/packages/app/src/specs/InlineSpecList.cy.tsx
@@ -34,6 +34,12 @@ describe('InlineSpecList', () => {
     .should('exist')
     .and('have.length', 7)
 
+    // overflow is required for the virtual list to work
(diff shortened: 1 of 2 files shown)


jennifer-shehane (commented):
@marktnoonan Any way to add a Cypress test to actually scroll and check this? That would be a better test than relying on this specific style. Use `scroll` commands within Cypress?

mjhenkes (approved):

ZachJW34 (approved):

Issue #22012, opened 2022-06-01
Cypress 10 migration tool does not work with custom support file path


ddzz:
### Current behavior

Our project uses a custom directory structure for our Cypress tests. Our support file is in `repo/src/platform/testing/e2e/cypress/support/`. In the Cypress 10 migration tool, clicking on `Rename the support file for me` does nothing, and I can't proceed to the next step.

<img width="1200" alt="Screen Shot 2022-06-01 at 3 02 29 PM" src="https://user-images.githubusercontent.com/3535749/171491953-74d22559-e8a5-494f-a043-a613a5c05246.png">

### Desired behavior

I should be able to use the Cypress 10 migration tool with a custom directory structure.

### Test code to reproduce

Our project is open-source: https://github.com/department-of-veterans-affairs/vets-website

The Cypress configuration is in the `/config` directory.

### Cypress Version

10.0.0

jennifer-shehane:
@ddzz Thanks for reporting and providing a reproducible example. We're getting a fix in for this now.

PR #22015, opened 2022-06-01
fix: do not show incorrectly support file migration step

estrada9166:
- Closes  #22012 

### User facing changelog

We were showing the supportFile migration step, based on a default `supportFile` which was a hardcoded string, making the migration flow, show this step even if the custom supportFile set on the config file includes it partially, like for example "repo/src/platform/testing/e2e/cypress/support/index.js"

### Steps to test

1. Create a supportFile in a path that is not the default but includes `cypress/support/index.`
2. Run the migration

### PR Tasks


packages/data-context/src/sources/migration/utils.ts
```diff
@@ -0,0 +1,9 @@
+import _ from 'lodash'
+
+export const isDefaultSupportFile = (supportFile: string) => {
+  if (_.isNil(supportFile) || !_.isBoolean(supportFile) && supportFile.match(/^cypress\/support($|\/index($|\.(ts|js|coffee)$))/)) {
```
estrada9166, line 4: Added the `coffee` ext here

tgriesser (approved):

packages/launchpad/cypress/e2e/migration.cy.ts
```diff
@@ -503,6 +503,40 @@ describe('Full migration flow for each project', { retries: { openMode: 0, runMo
     checkOutcome()
   })
 
+  it('completes journey for migration-e2e-custom-supportFile', () => {
```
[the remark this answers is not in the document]
flotwig: This is testing a custom support file which doesn't exist. #22012 is a custom support file which does exist. Unless I'm missing something, you need a test case for "a custom support file which does exist"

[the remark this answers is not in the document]
flotwig: In his case, I'd expect these to all fail:

```
    // supportFile is false - cannot migrate
    cy.get(renameSupportStep).should('not.exist')
    cy.get(setupComponentStep).should('not.exist')
    cy.get(configFileStep).should('exist')
```

packages/data-context/src/sources/migration/utils.ts
ZachJW34, line 4: I know this code was lifted, but what happens if the user has `./cypress/support/index.js` (prefixed with "./"). This is the default file but the regex wouldn't match

estrada9166: Maybe we should update the regex for something like `(^|.+/)cypress\/support($|\/index($|\.(ts|js|coffee)$))` what do you think?

estrada9166: [2e5b5d6](https://github.com/cypress-io/cypress/pull/22015/commits/2e5b5d6c9bda0efd3245056f9746e920e00a5b93)

flotwig (approved):

Issue #22054, opened 2022-06-02
Cypress 10 watch all files when executing cypress run


anlambert:
### Current behavior

Today, I took some time to upgrade the project I am working on to Cypress 10 (see [details](https://forge.softwareheritage.org/D7947)).

The upgrade went fine (nice work on the migration tool, btw) and I could still run my e2e tests locally without an issue.

However when executing the tests on our Jenkins instance, the following error is raised when executing `cypress run` (see [build log](https://jenkins.softwareheritage.org/blue/organizations/jenkins/DWAPPS%2Ftests-on-diff/detail/tests-on-diff/1874/pipeline/)):
```
[2022-06-02T17:33:36.831Z] $ /var/lib/jenkins/workspace/DWAPPS/tests-on-diff/node_modules/.bin/cypress run --config numTestsKeptInMemory=0 --config watchForFileChanges=false

[2022-06-02T17:33:38.697Z] [627:0602/173338.326243:ERROR:sandbox_linux.cc(377)] InitializeSandbox() called with multiple threads in process gpu-process.

[2022-06-02T17:33:38.697Z] [627:0602/173338.333408:ERROR:gpu_memory_buffer_support_x11.cc(44)] dri3 extension not supported.

[2022-06-02T17:33:41.180Z] 

[2022-06-02T17:33:41.180Z] tput: No value for $TERM and no -T specified

[2022-06-02T17:33:41.180Z] ====================================================================================================

[2022-06-02T17:33:41.180Z] 

[2022-06-02T17:33:41.180Z]   (Run Starting)

[2022-06-02T17:33:41.180Z] 

[2022-06-02T17:33:41.180Z]   ┌────────────────────────────────────────────────────────────────────────────────────────────────┐

[2022-06-02T17:33:41.180Z]   │ Cypress:        10.0.1                                                                         │

[2022-06-02T17:33:41.180Z]   │ Browser:        Electron 100 (headless)                                                        │

[2022-06-02T17:33:41.180Z]   │ Node Version:   v14.19.3 (/usr/bin/node)                                                       │

[2022-06-02T17:33:41.180Z]   │ Specs:          25 found (add-forge-now-request-create.cy.js, add-forge-now-request-dashboard. │

[2022-06-02T17:33:41.180Z]   │                 cy.js, add-forge-now-requests-moderation.cy.js, admin.cy.js, api-tokens.cy.js, │

[...]

PR #22067, opened 2022-06-02
fix: watch only specs matching specPattern

estrada9166:
- Closes  Second part of https://github.com/cypress-io/cypress/issues/22054#issuecomment-1145214532

### Additional details

- We were watching all the files instead of watching the ones related with the specPattern

### PR Tasks

### DIFF
--- a/packages/data-context/src/sources/VersionsDataSource.ts
+++ b/packages/data-context/src/sources/VersionsDataSource.ts
@@ -100,8 +100,9 @@ export class VersionsDataSource {
   }
 
   private async getVersionMetadata (): Promise<Record<string, string>> {
+    const now = new Date().toISOString()
     const DEFAULT_RESPONSE = {
-      [pkg.version]: new Date().toISOString(),
+      [pkg.version]: now,
     }
 
     if (this.ctx.isRunMode) {
@@ -112,21 +113,18 @@ export class VersionsDataSource {
 
     try {
       response = await this.ctx.util.fetch(NPM_CYPRESS_REGISTRY)
+      const responseJson = await response.json() as { time: Record<string, string>}
+
+      debug('NPM release dates received %o', { modified: responseJson.time.modified })
+
+      return responseJson.time ?? now
     } catch (e) {
       // ignore any error from this fetch, they are gracefully handled
       // by showing the current version only
       debug('Error fetching %o', NPM_CYPRESS_REGISTRY, e)
-    }
 
-    if (!response) {
       return DEFAULT_RESPONSE
     }
-
-    const responseJson = await response.json() as { time: Record<string, string>}
-
-    debug('NPM release dates received %o', { modified: responseJson.time.modified })
-
-    return responseJson.time
   }
 
   private async getLatestVersion (): Promise<string> {
@@ -153,36 +151,32 @@ export class VersionsDataSource {
       manifestHeaders['x-machine-id'] = id
     }
 
-    let manifestResponse
-
     try {
-      manifestResponse = await this.ctx.util.fetch(url, {
+      const manifestResponse = await this.ctx.util.fetch(url, {
         headers: manifestHeaders,
       })
+
+      debug('retrieving latest version information with headers: %o', manifestHeaders)
+
+      const manifest = await manifestResponse.json() as { version: string }
+
+      debug('latest version information: %o', manifest)
+
+      return manifest.version ?? pkg.version
     } catch (e) {
       // ignore any error from this fetch, they are gracefully handled
       // by showing the current version only
       debug('Error fetching %o', url, e)
-    }
 
-    if (!manifestResponse) {
       return pkg.version
+    } finally {
+      this._initialLaunch = false
     }
-
-    debug('retrieving latest version information with headers: %o', manifestHeaders)
-
-    const manifest = await manifestResponse.json() as { version: string }
-
-    debug('latest version information: %o', manifest)
-
-    this._initialLaunch = false
-
-    return manifest.version
   }
 
   private static async machineId (): Promise<string | undefined> {
     try {
-      return nmi.machineId()
+      return await nmi.machineId()
     } catch (error) {
       return undefined
     }
--- a/packages/data-context/test/unit/sources/VersionsDataSource.spec.ts
+++ b/packages/data-context/test/unit/sources/VersionsDataSource.spec.ts
@@ -1,6 +1,8 @@
 import { expect } from 'chai'
 import os from 'os'
 import sinon from 'sinon'
+import { Response } from 'cross-fetch'
+
 import { DataContext } from '../../../src'
 import { VersionsDataSource } from '../../../src/sources'
 import { createTestDataContext } from '../helper'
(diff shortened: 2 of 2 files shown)


tgriesser (changes requested):
I think the changes seems like it'll work, but can we some tests confirming the behavior of the functionality here - in the unit test layer should be fine, but current unit test added doesn't tell us anything more than that there's an extra function added to the `ignored`. 

Maybe something like creating a directory with a few files that we wouldn't expect to be watched and make sure that it's not called?

We could create a `_makeSpecWatcher` method that just returns the `specWatcher` instance in so it's easier to test in isolation:

```
this._specWatcher = this._makeSpecWatcher()
```

tgriesser (approved):

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -335,17 +336,40 @@ export class ProjectDataSource {
...
+    return chokidar.watch('.', {
+      ignoreInitial: true,
+      ignorePermissionErrors: true,
+      cwd: projectRoot,
+      ignored: ['**/node_modules/**', ...excludeSpecPattern, ...additionalIgnorePattern, (file: string) => {
+        // don't ignore things that look like directories,
+        // using statSync, falling back to whether it has an extension name
+        // if that were to throw
+        try {
+          if (fs.statSync(file).isDirectory()) {
```
flotwig, line 360: `ignored`'s second argument has `Fs.Stats`, so you don't need to do a synchronous `stat` here: https://github.com/paulmillr/chokidar#path-filtering

Also, avoid doing synchronous FS operations, it blocks the entire server when you do this.

packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
```diff
@@ -442,6 +444,74 @@ describe('getPathFromSpecPattern', () => {
...
+
+    const SUPPORT_FILE = path.join(this.specWatcherPath, 'cypress', 'support', 'e2e.js')
+    const SPEC_FILE1 = path.join(this.specWatcherPath, 'cypress', 'e2e', 'foo.cy.js')
+    const SPEC_FILE2 = path.join(this.specWatcherPath, 'cypress', 'e2e', 'some', 'new', 'folder', 'foo.cy.js')
+
+    ctx.actions.file.writeFileInProject(SUPPORT_FILE, '// foo')
+    ctx.actions.file.writeFileInProject(SPEC_FILE1, '// foo')
+    ctx.actions.file.writeFileInProject(SPEC_FILE2, '// foo')
+
+    await delay(500)
```
flotwig, line 506: This is bad practice for unit tests, it makes the tests really slow if we have delays like this. Is there anything you could `await` or use `done` instead?

packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
```diff
@@ -442,6 +444,74 @@ describe('getPathFromSpecPattern', () => {
...
+    expect(Array.from(allFiles)).to.eql([
+      SPEC_FILE1,
+      SPEC_FILE2,
+    // Paths don't start with '/' when emitted
+    ].map((p) => p.slice(1)))
+  })
```
flotwig, line 512: So is the assertion here that we "aren't watching `SUPPORT_FILE`"? I think a comment or an explicit assertion would make it clearer:

```suggestion
    // by the end, we should only be watching specs, not support
    expect(Array.from(allFiles)).to.eql([
      SPEC_FILE1,
      SPEC_FILE2,
    // Paths don't start with '/' when emitted
    ].map((p) => p.slice(1)))
  })
```

packages/data-context/src/sources/ProjectDataSource.ts
tgriesser: Updating

flotwig (changes requested):
Looks good, couple small things

packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
```diff
@@ -467,48 +470,83 @@ describe('_makeSpecWatcher', () => {
...
+    specWatcher.on('add', (filePath) => allFiles.add(filePath))
+    specWatcher.on('change', (filePath) => allFiles.add(filePath))
+
+    writeFiles()
+
+    let attempt = 0
+
+    while (allFiles.size < 3 && attempt++ <= 100) {
+      await delay(10)
+    }
```
flotwig, line 540: eheheheh ok :smile:

flotwig (approved):

packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
flotwig: Tests were made more descriptive in another way, resolved

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -11,6 +11,7 @@ import { defaultSpecPattern } from '@packages/config'
 import parseGlob from 'parse-glob'
 import micromatch from 'micromatch'
 import RandExp from 'randexp'
+import fs from 'fs'
```
lmiller1990, line 14: Should this be `fs-extra`? I think we prefer this over `fs` everywhere. We should lint against this, probably.

lmiller1990 (approved):
Seems fine, I wonder if we eventually want to watch `chokidar` to work around all the weirdness, and possibly include a `gitignore: true` arg like `globby`...

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -335,17 +336,46 @@ export class ProjectDataSource {
...
     // the 'all' event includes: add, addDir, change, unlink, unlinkDir
     this._specWatcher.on('all', onProjectFileSystemChange)
   }
 
+  _makeSpecWatcher ({ projectRoot, specPattern, excludeSpecPattern, additionalIgnorePattern }: { projectRoot: string, excludeSpecPattern: string[], additionalIgnorePattern: string[], specPattern: string[] }) {
+    return chokidar.watch('.', {
+      ignoreInitial: true,
+      ignorePermissionErrors: true,
+      cwd: projectRoot,
+      ignored: ['**/node_modules/**', ...excludeSpecPattern, ...additionalIgnorePattern, (file: string, stats?: fs.Stats) => {
```
lmiller1990, line 355:
```suggestion
    return chokidar.watch(projectRoot, {
      ignoreInitial: true,
      ignorePermissionErrors: true,
      ignored: ['**/node_modules/**', ...excludeSpecPattern, ...additionalIgnorePattern, (file: string, stats?: fs.Stats) => {
```

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -335,17 +336,46 @@ export class ProjectDataSource {
...
+        // We need stats arg to make the determination of whether to watch it, because we need to watch directories
+        // chokidar is extremely inconsistent in whether or not it has the stats arg internally
```
lmiller1990, line 357: I also saw some comments around `**/node_modules/**` not always working as expected: https://github.com/paulmillr/chokidar/issues/773

Could do [like this](https://github.com/paulmillr/chokidar/issues/773#issuecomment-536793469)

```js
if (file.includes('node_modules')) {
  return true
}
```

To be extra safe?

lmiller1990 (approved):
Seems generally find, added some comments around code style + potential improvement based on some historical comments about chokidar's weirdness around ignoring globs and the fact it doesn't always seem to work.

lmiller1990 (commented):
Added some non blocking comments.

PR #22069, opened 2022-06-02
fix: only show active router link styles on parent page

### LINKED ISSUE #21873: Indicate how to get to Specs Page from Runner (minor style tweaks to sidebar)
## **Summary**
The navigation item in the left hand menu that goes to Specs is styled in an “inert” state, as the active item in the menu.

During the bug hunt, several Cypress employees on a video call with me could not see how to get back to the specs list from the runner, which is fair since the link that navigates there is not styled to be clickable (other links have `cursor: pointer` and hover states, this does not, since we are already inside that part of the app). This video shows the problem in more detail:

[https://www.loom.com/share/f95449b235474cf2b2d0ef096122c764](https://www.loom.com/share/f95449b235474cf2b2d0ef096122c764|smart-link)

The fix here is small: since every item in the side bar is a valid navigation target from the Spec Runner, they should all look and behave the same way so the user knows it is valid to click any of the three items, so we should turn off the “active” state of the spec item. 

This was part of the feedback noted in [https://cypress-io.atlassian.net/browse/UNIFY-1667](https://cypress-io.atlassian.net/browse/UNIFY-1667|smart-link) - but the title of that ticket would make it harder to track.

## **Acceptance Criteria**
* Specs List item in the navigation should only be active when on the Specs List page, not the runner

### **Resources**
Any Notion documents, Google documents, Figma Boards

┆Issue is synchronized with this [Jira Task](https://cypress-io.atlassian.net/browse/UNIFY-1688) by [Unito](https://www.unito.io)
┆author: Mark Noonan
┆epic: UI/UX Improvements
┆friendlyId: UNIFY-1688
┆priority: Medium
┆sprint: Fast Follows 1
┆taskType: Task

rachelruderman:
- Closes #21873

### User facing changelog

Minor UI tweak to better indicate how to return to the Specs list when the test runner is active

### Additional details

During the bug hunt, several Cypress employees on a video call could not see how to get back to the specs list from the runner, which is fair since the link that navigates there is active, so it is not styled to be clickable (other links have cursor: pointer and hover states, this does not, since we are already inside that part of the app). 

This change makes it so the navigation links are only styled as active when the URL is an exact match. Now, if the user goes to a child page (i.e. test runner is a child page of the Specs page) the link to the parent page will appear clickable as it no longer has active styles.

### Steps to test

1. `yarn dev`
2. Select "E2E testing"
3. Confirm that the "Specs" nav link has active styles (purple color, default cursor on hover)
4. Click on an E2E spec, i.e. `logger.cy.ts`
5. Confirm that the "Specs" nav link no longer has active styles (gray color, pointer cursor on hover)
6. Confirm that clicking the "Specs" nav link returns the user to the "Specs" parent page

**Before:**
Screen Shot 2022-06-02 at 4 45 10 PM

**After:**
Screen Shot 2022-06-02 at 4 45 28 PM

### PR Tasks


packages/app/src/navigation/SidebarNavigation.vue
```diff
@@ -38,12 +38,12 @@
...
         <RouterLink
           v-for="item in navigation"
-          v-slot="{ isActive }"
+          v-slot="{ isExactActive }"
           :key="item.name"
           :to="item.href"
         >
           <SidebarNavigationRow
-            :active="isActive"
+            :active="isExactActive"
```
rachelruderman, line 46: I opted for a global change instead of a Specs-only exception to ensure all navigation links continue to have the same behavior. 

This means that if any of the other pages have links to a child view with a different URL (whether it be query params, pathname addition, etc) the navigation link will lose its active styles. This is intentional so that the user can more intuitively understand how to return to the parent page.

packages/app/cypress/e2e/sidebar_navigation.cy.ts
```diff
@@ -235,6 +235,20 @@ describe('Sidebar Navigation', () => {
...
+    it('Specs sidebar nav link is not active when a test is running', () => {
+      cy.location('hash').should('equal', '#/specs')
+      cy.contains('.router-link-exact-active', 'Specs')
+
+      cy.findAllByTestId('spec-item').first().click()
+      cy.location('hash').should('contain', '#/specs/runner')
+      cy.contains('.router-link-exact-active', 'Specs').should('not.exist')
+      cy.percySnapshot()
+
+      cy.get('a[href="#/specs"]').click()
```
MuazOthman, line 247: does it make sense to use a more specific selector like `cy.findAllByTestId('nav-link-spec')` just in case another link was added later?

MuazOthman (approved):

MuazOthman (approved):

packages/app/cypress/e2e/sidebar_navigation.cy.ts
MuazOthman: The update looks elegant!

packages/app/src/navigation/SidebarNavigation.vue
marktnoonan: Yeah I agree with doing this as the default. It's likely in the future, if we have nested router views, there would be cases where the exact route changes but the user is still "on" the main page. But I'm fine with not worrying about that till we get to that point.

marktnoonan (approved):
Works great and feels much more clear 👍

PR #22070, opened 2022-06-03
fix: update sample spec file so it passes

rachelruderman:
- Closes 

### PR Tasks

### DIFF
--- a/packages/data-context/test/unit/codegen/code-generator.spec.ts
+++ b/packages/data-context/test/unit/codegen/code-generator.spec.ts
@@ -125,9 +125,10 @@ describe('code-generator', () => {
           status: 'add',
           file: fileAbsolute,
           content: `${dedent`
-            describe('my-integration-file.js', () => {
-              it('should visit', () => {
-                cy.visit('/')
+            context('Navigation', () => {
+              it('cy.visit() - visit a remote url', () => {
+                cy.visit('https://example.cypress.io/commands/navigation')
+                cy.contains('h1', 'Navigation')
               })
             })` }\n`,
         },


packages/data-context/test/unit/codegen/code-generator.spec.ts
```diff
@@ -125,11 +125,11 @@ describe('code-generator', () => {
           status: 'add',
           file: fileAbsolute,
           content: `${dedent`
-            describe('my-integration-file.js', () => {
-              it('should visit', () => {
-                cy.visit('/')
```
flotwig, line 130: The alternative would've been `cy.visit('https://example.cypress.io')`, is there a reason not to do that? Shows off a lil bit more of Cypress.

rachelruderman: @flotwig I started off with that in [this commit](https://github.com/cypress-io/cypress/pull/22070/commits/85dac33c868fe78c4281752ef6026d8636808544) using
```
context('Navigation', () => {
  it('cy.visit() - visit a remote url', () => {
    cy.visit('https://example.cypress.io/')
  })
})
``` 

but, on further reflection, realized it looked a little bit too much like a scaffolded example spec when the user had opted for "Create empty spec" instead of "Scaffold example specs"

I can go either way though! Whatever y'all prefer 🙂

flotwig: Above my pay grade, but my vote is the `cy.visit`!

rachelruderman: Conclusion of Slack discussion: `cy.visit` it is!
https://cypressio.slack.com/archives/C01TQQCQLJY/p1654279272535549?thread_ts=1654278777.307619&cid=C01TQQCQLJY

flotwig (approved):

tgriesser (approved):

PR #20470, opened 2022-03-03
fix: config migration when --config-file flag is set

estrada9166:
- Closes  [UNIFY-1176](https://cypress-io.atlassian.net/browse/UNIFY-1176) [UNIFY-571](https://cypress-io.atlassian.net/jira/software/projects/UNIFY/boards/20?assignee=5ea883deddd45f0b94f5d98d&selectedIssue=UNIFY-571)

### PR Tasks

- [ ] Have tests been added/updated?
- [ ] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [ ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [ ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [ ] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

PR #22073, opened 2022-06-03
fix: remove projects from breadcrumbs entirely for non-global mode projects

### LINKED ISSUE #21771: Remove 'Projects' from breadcrumbs entirely for non global mode projects
The ‘projects’ breadcrumb was updated to not be clickable when opened outside of global mode, but the ‘Projects’ still displays. This is confusing to users that have never used global mode as it looks like it should be clickable or there is some areas to view ‘Projects’. 

The ‘Projects’ breadcrumb should be removed if the project wasn’t launched via global mode (updated this direction as that was the conclusion in the linked Slack thread).

Screen Shot 2022-03-24 at 10 59 00 AM

Followup from [https://cypress-io.atlassian.net/browse/UNIFY-1257](https://cypress-io.atlassian.net/browse/UNIFY-1257|smart-link)  based on feedback given in Slack: 

[https://cypressio.slack.com/archives/C02MYBT9Y5S/p1648064742752759](https://cypressio.slack.com/archives/C02MYBT9Y5S/p1648064742752759|smart-card)

┆Issue is synchronized with this [Jira Task](https://cypress-io.atlassian.net/browse/UNIFY-1390) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/11573">Screen Shot 2022-03-24 at 10.59.00 AM.png</a>
┆author: Jennifer Shehane
┆epic: UI/UX Improvements
┆friendlyId: UNIFY-1390
┆priority: Medium
┆sprint: Fast Follows 1
┆taskType: Task

rachelruderman:
- Closes #21771

### User facing changelog
The `Projects` breadcrumb is now only shown in global mode

### Additional details
Showing the disabled `Projects` breadcrumb outside of global mode was confusing to users as it looked like it should be clickable or there was some areas to view ‘Projects’. 

See PR comments for additional implementation details

### Steps to test
**Non-global mode:**
1. From root, `cd packages/frontend-shared`
2. `yarn dev`
3. In launchpad, confirm header does not have `Projects > ` breadcrumb

**Global mode:**
1. From root, `yarn dev`
2. In launchpad, confirm header has disabled `Projects > ` breadcrumb
3. Click a project
4. Confirm header has enabled `Projects > ` breadcrumb
5. Click Projects
6. Confirm `Projects` breadcrumb is disabled again

**Before:**
Screen Shot 2022-06-27 at 4 01 52 PM

**After:**
Screen Shot 2022-06-27 at 4 00 19 PM

### PR Tasks

- na Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- na Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?


rachelruderman:
> @rachelruderman whoops just noticed this, @mike-plummer and I looked at this same issue Friday - must have missed the assignment. We made pretty much [the same changes](https://github.com/cypress-io/cypress/compare/develop...21771-remove-projects-when-not-a-link#diff-e78ec7277b30ed685300b453fa55dc7e912265aeab19eb41668d6f3e26e44de8R283) that you have. I see this is still in draft but might as well share where we were:
> 
> The main difference is we needed to update the `hasLinkToProjects` logic to account for still needing to show this on the global mode Projects page itself, where the gray-text version is in the design. I was actually about to put the disabled link pattern back in too, since on the projects page it might be a good signal to users that this _will_ become an active link some time in the future when you want to get back.
> 
> Lastly I think we hit something around `isInGlobalMode` becoming stale and that maybe the logic was gonna need a further tweak in order to correctly still show the Projects title if you go back and forth between the projects page and a project. But I haven't done any more work since pairing with Mike, and the purpose of pairing was his onboarding, not really the issue itself.

Ahahahaha I love that our assertions were practically identical! This is awesome, I had the same thought about showing disabled Projects page but wanted to confirm with design first, and I didn't know about `isInGlobalMode` so that's super helpful. Thank you for the heads up, it's much appreciated!

packages/frontend-shared/src/gql-components/HeaderBarContent.cy.tsx
```diff
@@ -189,62 +243,8 @@ describe('<HeaderBarContent />', { viewportWidth: 1000, viewportHeight: 750 }, (
     cy.contains(`${defaultMessages.topNav.updateCypress.title} 8.7.0`).should('not.exist')
   })
 
-  it('displays the active project name and testing type', () => {
```
rachelruderman, line 192: I moved these tests inside the `breadcrumbs` describe block

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -265,7 +265,7 @@ fragment HeaderBar_HeaderBarContent on Query {
     branch
     isLoadingNodeEvents
   }
-  projectRootFromCI
+  isInGlobalMode,
```
rachelruderman, line 268: Hi @estrada9166 ! It looks like you added `projectRootFromCI` in [this PR from March](https://github.com/cypress-io/cypress/pull/20470). Was there a reason you created `projectRootFromCI` instead of using `isInGlobalMode`?

packages/data-context/src/DataContext.ts
```diff
@@ -132,10 +132,6 @@ export class DataContext {
     return this._config.localSettingsApi
   }
 
-  get isGlobalMode () {
-    return this.appData.isInGlobalMode
-  }
```
rachelruderman, line 137: In the existing code, the variable `isGlobalMode` was used to expose `isInGlobalMode`. This led to some confusion and was an unnecessary duplication, so I removed it in favor of using `this.ctx.modeOptions.global` in the one location where `this.ctx.isGlobalMode` was being used

packages/data-context/test/unit/actions/MigrationActions.spec.ts
```diff
@@ -91,7 +90,6 @@ describe('MigrationActions', () => {
       ctx.update((coreData) => {
         coreData.currentProject = currentProject
         coreData.currentTestingType = 'e2e'
-        coreData.app.isInGlobalMode = true
       })
```
rachelruderman, line 93: This was some very sticky logic! Before this PR, `isInGlobalMode` really meant `currentProject is falsy`. In `packages/graphql/src/schemaTypes/objectTypes/gql-Query.ts` we had the following logic:
```
t.nonNull.boolean('isGlobalMode', {
      description: 'Whether the app is in global mode or not',
      resolve: (source, args, ctx) => !ctx.currentProject,
    })
```
As a result, some data massaging was necessary to keep `isInGlobalMode` true in `MigrationActions.spec.ts` when the currentProject was created.

Now, this logic has been removed and `isGlobalMode` refers ONLY to whether the app was opened with the flag `--global` per the `modeOptions` object

packages/data-context/test/unit/actions/MigrationActions.spec.ts
```diff
@@ -91,7 +90,6 @@ describe('MigrationActions', () => {
       ctx.update((coreData) => {
         coreData.currentProject = currentProject
         coreData.currentTestingType = 'e2e'
-        coreData.app.isInGlobalMode = true
```
rachelruderman, line 94: This was some very sticky logic! In `packages/graphql/src/schemaTypes/objectTypes/gql-Query.ts` we had the following logic:
```
t.nonNull.boolean('isGlobalMode', {
      description: 'Whether the app is in global mode or not',
      resolve: (source, args, ctx) => !ctx.currentProject,
    })
```
In other words,`isInGlobalMode` really meant `there is no current project`. As a result, some data massaging was necessary to keep `isInGlobalMode` true in `MigrationActions.spec.ts` when the currentProject was created.

With this PR, `isGlobalMode` means one thing and one thing only: whether the app was opened with the flag `--global` per the `modeOptions` object

packages/frontend-shared/src/gql-components/HeaderBarContent.cy.tsx
```diff
@@ -282,7 +279,6 @@ describe('<HeaderBarContent />', { viewportWidth: 1000, viewportHeight: 750 }, (
     cy.contains(cloudViewer.fullName).should('be.visible')
     cy.contains(cloudViewer.email).should('be.visible')
     cy.findByRole('button', { name: text.login.actionLogout }).should('be.visible')
-    cy.percySnapshot()
```
rachelruderman, line 285: I added an `afterEach` hook with `cy.percySnapshot` in this file, so anywhere there was a snapshot at the end of a test I removed it

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -18,37 +18,35 @@
           class="h-32px mr-18px w-32px"
           src="../assets/logos/cypress-dark.png"
         >
-        <nav
-          role="navigation"
-          aria-label="Breadcrumbs"
-        >
```
rachelruderman, line 24: Updated per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -18,37 +18,35 @@
...
+        <nav>
```
rachelruderman, line 21: Updated per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role

rachelruderman:
Screen Shot 2022-06-28 at 2 09 28 PM

^ In reviewing this PR, we would expect that `Projects` is only removed when it was previously inactive. However, in the Percy snapshots for `HeaderBarContent`, `Projects` was active AND it was removed. This is expected because the previous logic was:
```
```

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -271,7 +268,7 @@ fragment HeaderBar_HeaderBarContent on Query {
     branch
     isLoadingNodeEvents
   }
-  projectRootFromCI
+  isGlobalMode,
```
tgriesser, line 271: Doesn't make a difference, since the character is ignored:

```suggestion
  isGlobalMode
```

packages/frontend-shared/cypress/support/component.ts
```diff
@@ -20,6 +21,7 @@ cy.gqlStub = GQLStubRegistry
 
 Cypress.on('uncaught:exception', (err) => !err.message.includes('ResizeObserver loop limit exceeded'))
 
+configure({ testIdAttribute: 'data-cy' })
```
lmiller1990, line 24: What about putting this in `packages/frontend-shared/cypress/support/common.ts`? We also do the same `configure` call [here](https://github.com/cypress-io/cypress/blob/develop/packages/app/cypress/component/support/index.ts#L30). That same file also [imports common](https://github.com/cypress-io/cypress/blob/develop/packages/app/cypress/component/support/index.ts#L1), so if we do `config` in `common`, then we could also remove it from `app` and just have it in one place.

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -18,56 +18,53 @@
...
+                :role="canClearTestingType ? undefined : 'link'"
+                :href="canClearTestingType ? 'choose-testing-type' : undefined"
+                :class="canClearTestingType ? 'text-indigo-500 hocus-link-default' :
                   'text-gray-700'"
-                :ariaDisabled="!hasLinkToCurrentProject"
+                :ariaDisabled="!canClearTestingType"
                 @click.prevent="clearTestingType"
               >
-                {{ props.gql?.currentProject?.title }}
+                {{ currentProject?.title }}
```
lmiller1990, line 64: Should we just do a null check on the outer `<li>` like on line 44? Seems we should never be rendering this `<a>` without a `currentProject`.

Line 44: https://github.com/cypress-io/cypress/pull/22073/files#diff-e78ec7277b30ed685300b453fa55dc7e912265aeab19eb41668d6f3e26e44de8R44

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -18,56 +18,53 @@
...
               </a>
               <template
-                v-if="props.gql?.currentProject?.branch"
+                v-if="currentProject?.branch"
```
lmiller1990, line 67: Outer check on the `<li>` would mean we don't need this `v-if`, too.

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -97,10 +94,10 @@
               />
             </li>
             <li
-              v-if="props.gql?.currentProject?.currentTestingType"
+              v-if="currentProject?.currentTestingType"
               class="inline-block"
             >
-              {{ t(`testingType.${props.gql?.currentProject?.currentTestingType}.name`) }}
+              {{ t(`testingType.${currentProject?.currentTestingType}.name`) }}
```
lmiller1990, line 100: Same here - does the first `v-if` not mean we don't need the following `?`?

lmiller1990 (commented):
Nice job, couple of small comments/questions.

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -18,56 +18,53 @@
           class="h-32px mr-18px w-32px"
           src="../assets/logos/cypress-dark.png"
```
marktnoonan, line 19: Completely unrelated but let's put an `alt` attr on this image while we're in this file, oops!

marktnoonan (commented):
Looks good, just a couple non-blocking thoughts. Updated to trigger CI since percy build didn't complete. @lmiller1990's comment about the rendering logic makes sense, if we can simplify, that's great.

astone123 (approved):
Works well for me 👍🏻

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -18,56 +18,53 @@
...
                 class="font-medium"
-                :role="hasLinkToCurrentProject ? undefined : 'link'"
-                :href="hasLinkToCurrentProject ? 'choose-testing-type' : undefined"
-                :class="hasLinkToCurrentProject ? 'text-indigo-500 hocus-link-default' :
+                :role="canClearTestingType ? undefined : 'link'"
+                :href="canClearTestingType ? 'choose-testing-type' : undefined"
+                :class="canClearTestingType ? 'text-indigo-500 hocus-link-default' :
                   'text-gray-700'"
-                :ariaDisabled="!hasLinkToCurrentProject"
+                :ariaDisabled="!canClearTestingType"
```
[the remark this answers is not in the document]
marktnoonan: I have a PR coming hot on the heels of this one that will remove the link functionality itself, so let's not sweat this: https://github.com/cypress-io/cypress/pull/22648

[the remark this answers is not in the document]
tbiethman: I'll save my breadcrumb related comments for your PR @marktnoonan, thanks for the heads up

packages/frontend-shared/cypress/support/component.ts
rachelruderman: I like!

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
rachelruderman: Nice catch!

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
rachelruderman: Oh wow! This comment made me realize my VScode accessibility extension isn't working as expected. Thanks for the catch 🎣

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
rachelruderman: Ahhh it turns out this `v-if` is needed indeed. Although we have a currentProject, it may not necessarily have a branch. I.e. the user is working locally and hasn't created a git branch yet

lmiller1990 (approved):

Issue #877, opened 2017-11-06
.should('not.be.visible') fails when elem out of viewport


dwelle:
-  Operating System: win7x64
- Cypress Version: 1.0.3

### Is this a Feature or Bug?

bug

### Current behavior:

```
cy.get().should('not.be.visible')
```

fails even when DOM element is not in viewport.

It seems [I'm not the only one](https://gitter.im/cypress-io/cypress?at=5a004fd332e080696e6e6f8a) reporting this behavior.

### Desired behavior:

Acc to [doc](https://docs.cypress.io/guides/core-concepts/interacting-with-elements.html#Scrolling), only actionable commands should autoscroll DOM to viewport.

### How to reproduce:

```js
// ensure small viewport
cy.viewport( 999, 200 );
// ensure scrollbar is disabled, for good measure (though it doesn't seem Cypress cares)
cy.window().then( window => {
    window.$("body").css("overflow-y", "hidden");
});
// manually test for whether elem is out of viewport -- PASSES
cy.get(".elem").first().then( $el => {

    const bottom = Cypress.$( cy.state("window") ).height();
    const rect = $el[0].getBoundingClientRect();

    expect( rect.top ).to.be.greaterThan( bottom );
    expect( rect.bottom ).to.be.greaterThan( bottom );
    expect( rect.top ).to.be.greaterThan( bottom );
    expect( rect.bottom ).to.be.greaterThan( bottom );
});
// FAILS
cy.get(".elem").first().should("not.be.visible");
```

brian-mann:
I don't believe this is a bug. We wrote the visibility calculations to take into account elements outside of the viewport.

What I mean is - an element is considered visible if the user in could in any way interact with it - even if they needed to scroll to it.

The reason this rule *has* to be in place is because scrolling is a mutation. If Cypress first attempted to scroll elements on every single `be.visible` assertion it could have dramatic side effects that can cause all kinds of problems.

jennifer-shehane:
This is the visibility logic in our code, in case you want to investigate before we are able to: https://github.com/cypress-io/cypress/blob/code-of-conduct/packages/driver/src/dom/visibility.coffee#L17

jennifer-shehane:
Yes, what @brian-mann explains above is actually true. The example in the kitchen sink works because we take into account elements being clipped by a parent container, but we don't take into account the viewport size when calculating visibility. It's complicated logic. 

[...]

PR #22094, opened 2022-06-03
fix: top nav browsers list scroll

### LINKED ISSUE #22032: Unable to scroll down the browsers list in Cypress 10
### Current behavior

Unable to scroll down the browsers list in the app-header-bar in the cypress test runner.

#### Test runner overview page

image

#### Test run view

image broken.

image

#### Cypress choose browser window

The ui/app is frozen when testing in Opera. unable to focus and close the browser true the ui. It needs a restart. 

image

### Desired behavior

See a full list of browsers that are configured. 

### Test code to reproduce

```
  config.browsers.push({
      name: 'opera',
      channel: 'stable',
      family: 'chromium',
      displayName: 'Opera',
      version: '84.0.4316.42',
      path: 'C:\\Users\\USERNAME\\AppData\\Local\\Programs\\Opera\\opera.exe',
      majorVersion: 84
  },
  {
      name: 'brave',
      channel: 'stable',
      family: 'chromium',
      displayName: 'Brave',
      version: '99.1.36.116',
      path: 'C:\\Program Files\\BraveSoftware\\Brave-Browser\\Application\\brave.exe',
      majorVersion: 99
  },
  {
      name: 'vivaldi',
      channel: 'stable',
      family: 'chromium',
      displayName: 'Vivaldi',
      version: '5.1.2567.66',
      path: 'C:\\Users\\USERNAME\\AppData\\Local\\Vivaldi\\Application\\vivaldi.exe',
      majorVersion: 5
  },
  {
      name: 'srware iron',
      channel: 'stable',
      family: 'chromium',
      displayName: 'SRWare Iron',
      version: '98.0.5000.0', 
      path: 'C:\\Program Files\\SRWare Iron (64-Bit)\\chrome.exe',
      majorVersion: 98
  });
  
  return config;
}
```

### Cypress Version

10

### Other

_No response_

ZachJW34:
- Closes #22032

### User facing changelog
Allow browsers list in top nav to be scrollable

### Additional details
We didn't have any scroll css on the browsers list in the top nav. This PR fixes this for all use cases of the `VerticalBrowserListItems.vue`.

I've added a test to verify it can scroll, and also added a missing percy snapshot for the unsupported browser tooltip

### Steps to test
You can run the `renders with functional browser menu when show-browsers prop is true` from `HeaderBarContent.cy.tsx`. It has a long list of browsers and there is no scrolling on the element other than the full page scroll.

Before:

Screen Shot 2022-06-03 at 10 41 12 AM

After:

Screen Shot 2022-06-03 at 10 40 54 AM

### PR Tasks

### DIFF
--- a/packages/frontend-shared/src/gql-components/HeaderBarContent.cy.tsx
+++ b/packages/frontend-shared/src/gql-components/HeaderBarContent.cy.tsx
@@ -26,9 +26,8 @@ describe('<HeaderBarContent />', { viewportWidth: 1000, viewportHeight: 750 }, (
       .should('not.exist')
     })
 
-    cy.get('[data-cy="top-nav-browser-list-item"]')
-    .last().should('not.be.visible')
-    .scrollIntoView().should('be.visible')
+    cy.get('[data-cy="top-nav-browser-list-item"]').parent()
+    .should('have.class', 'overflow-auto')
 
     cy.contains('Version unsupported')
     .scrollIntoView()
@@ -40,7 +39,7 @@ describe('<HeaderBarContent />', { viewportWidth: 1000, viewportHeight: 750 }, (
       .trigger('mouseenter')
     })
 
-    cy.contains('Unsupported browser')
+    cy.contains('Unsupported browser').should('be.visible')
 
     cy.percySnapshot('unsupported browser tooltip')
   }),
--- a/packages/frontend-shared/src/gql-components/HeaderBarContent.cy.tsx
+++ b/packages/frontend-shared/src/gql-components/HeaderBarContent.cy.tsx
@@ -16,6 +16,8 @@ describe('<HeaderBarContent />', { viewportWidth: 1000, viewportHeight: 750 }, (
     .should('be.visible')
     .click()
 
+    cy.percySnapshot('after browsers open')
+
     cy.contains('Edge Canary')
     .should('be.visible')
     .closest('[data-cy="top-nav-browser-list-item"]')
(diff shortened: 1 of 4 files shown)


rachelruderman (approved):
Tried and tested, looking good!

flotwig (approved):
Nice! Looks great on Linux.

packages/frontend-shared/src/gql-components/HeaderBarContent.cy.tsx
```diff
@@ -24,15 +26,23 @@ describe('<HeaderBarContent />', { viewportWidth: 1000, viewportHeight: 750 }, (
       .should('not.exist')
     })
 
+    cy.get('[data-cy="top-nav-browser-list-item"]')
+    .last().should('not.be.visible')
+    .scrollIntoView().should('be.visible')
```
marktnoonan, line 31: hmm, in light of https://github.com/cypress-io/cypress/issues/877 I'm actually a bit surprised this works. I ran into this exact case in a different context, and asserting visibility did not work.

marktnoonan (approved):
Works well, I am a bit puzzled about why `'not.be.visible'` works in that scrolling test, but it does!

packages/frontend-shared/src/gql-components/HeaderBarContent.cy.tsx
```diff
@@ -24,15 +26,23 @@ describe('<HeaderBarContent />', { viewportWidth: 1000, viewportHeight: 750 }, (
...
     .within(() => {
       cy.get('[data-cy="unsupported-browser-tooltip-trigger"]')
       .should('exist')
+      .trigger('mouseenter')
     })
 
-    cy.percySnapshot('after browsers open')
+    cy.contains('Unsupported browser')
+
+    cy.percySnapshot('unsupported browser tooltip')
```
marktnoonan, line 45: We probably need a visibility check to get the tooltip displayed before Percy takes the snapshot, it seems to take it too early - I think the content is is the DOM but the tooltip has not faded in yet. Some of our other tooltip Percy tests are a bit flaky and I'd guess it's for similar reasons.

packages/frontend-shared/src/gql-components/HeaderBarContent.cy.tsx
ZachJW34: Good catch! I adjusted the test to just check for overflow.

The reason it was passing was because the browsers list (without overflow) is long enough that the last element renders off the screen. If the list was shortened, the assertion would fail.

packages/frontend-shared/src/gql-components/HeaderBarContent.cy.tsx
ZachJW34: Added visibility check!

PR #22099, opened 2022-06-03
fix: #22081 Cypress crashing behind proxy

### LINKED ISSUE #22081: cypress >=10.0.0 crashes if behind company proxy when using the first time
### Current behavior

after installing cypress 10.0.1 or 10.0.2 i run cypress with `npx cpress open` which starts cypress ui. cypress loads correctly but when i click on `E2E Testing` it crashes after a few seconds.
cypress tries to access the url `https://download.cypress.io/desktop.json` but fails with:

`invalid json response body at https://download.cypress.io/desktop.json reason: Unexpected token < in JSON at position 0
FetchError: invalid json response body at https://download.cypress.io/desktop.json reason: Unexpected token < in JSON at position 0
    at C:\Users\<username>\AppData\Local\Cypress\Cache\10.0.2\Cypress\resources\app\node_modules\node-fetch\lib\index.js:273:32
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at VersionsDataSource.getLatestVersion (C:\Users\<username>\AppData\Local\Cypress\Cache\10.0.2\Cypress\resources\app\packages\data-context\src\sources\VersionsDataSource.js:125:26)`

i use cypress behind a company firewall which i set as environment variable `HTTP_PROXY=http://my-company-proxy.com:proxy-port`. actually i was able to install cypress itself with this configuration, but using it, fails.

### Desired behavior

cypres should run as previous versions behind a company firewall.

### Test code to reproduce

use a company firewall.

### Cypress Version

10.0.2

### Other

_No response_

tgriesser:
- Closes #22081

### User facing changelog

Fixes situations where Cypress launch would crash behind a network proxy due to unhandled promise rejection on non-json payload.

### Additional details

The response was valid, but the response payload was non-json, causing an error. Moved both the response handling & `res.json` promise inside the `try` block

### PR Tasks

### DIFF
--- a/packages/data-context/src/sources/VersionsDataSource.ts
+++ b/packages/data-context/src/sources/VersionsDataSource.ts
@@ -100,8 +100,9 @@ export class VersionsDataSource {
   }
 
   private async getVersionMetadata (): Promise<Record<string, string>> {
+    const now = new Date().toISOString()
     const DEFAULT_RESPONSE = {
-      [pkg.version]: new Date().toISOString(),
+      [pkg.version]: now,
     }
 
     if (this.ctx.isRunMode) {
@@ -112,21 +113,18 @@ export class VersionsDataSource {
 
     try {
       response = await this.ctx.util.fetch(NPM_CYPRESS_REGISTRY)
+      const responseJson = await response.json() as { time: Record<string, string>}
+
+      debug('NPM release dates received %o', { modified: responseJson.time.modified })
+
+      return responseJson.time ?? now
     } catch (e) {
       // ignore any error from this fetch, they are gracefully handled
       // by showing the current version only
       debug('Error fetching %o', NPM_CYPRESS_REGISTRY, e)
-    }
 
-    if (!response) {
       return DEFAULT_RESPONSE
     }
-
-    const responseJson = await response.json() as { time: Record<string, string>}
-
-    debug('NPM release dates received %o', { modified: responseJson.time.modified })
-
-    return responseJson.time
   }
 
   private async getLatestVersion (): Promise<string> {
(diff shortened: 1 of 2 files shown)


ryanthemanuel (approved):

packages/data-context/src/sources/VersionsDataSource.ts
```diff
@@ -100,8 +100,9 @@ export class VersionsDataSource {
   }
 
   private async getVersionMetadata (): Promise<Record<string, string>> {
```
flotwig, line 102: kinda a general question - shouldn't an error here basically never crash the process? at worst, we should be blanking the GUI and displaying the error in open mode. why can this even propagate out? is there work formalized to fix this?

we'll certainly have other exceptions in this layer, none of which should probably fully crash Cy in open mode

flotwig (approved):

packages/data-context/src/sources/VersionsDataSource.ts
tgriesser: It's b/c there wasn't a catch handler here: https://github.com/cypress-io/cypress/blob/ce647c29df7afc230e1189e3ec728787b1dd16b1/packages/data-context/src/sources/VersionsDataSource.ts#L94-L98

It's not a common pattern for us to do that but we were doing it here b/c we weren't wanting to block the caller on awaiting it

flotwig: But what I'm saying is, if the `this.ctx.update` callback throws, that still shouldn't crash open mode. Cypress is a GUI app, we should recover as gracefully as we can in open mode. Of course things like actual unhandled exceptions/rejections should crash the process, but for things like this, there *is* an appropriate point in the call stack where we could give a better error/recovery steps to the error?

My point is that it shouldn't be possible for GUI to crash unless something is irrecoverably wrong. That's kinda the point desktop-gui was at.

tgriesser: Yeah, agreed we should figure out better ways to do a better job to surface the issue and allow to continue when in open mode.

PR #22105, opened 2022-06-03
fix: Make spec row clickable across entire width

### LINKED ISSUE #21783: Rows in Specs List should be clickable all the way across
**Current Behavior**

Only parts of the the rows in the specs list are clickable: 

https://user-images.githubusercontent.com/1271364/171227982-7a2b1c8f-9a94-4294-83f7-2bb45fc77407.mp4

**Expected Behavior**

Clicking on the whole row should produce the same result as clicking on a folder name or a spec name.

This row has changed, it now contains interactive content and needs a distinct hover state: see this [slack thread](https://cypressio.slack.com/archives/C02MYBT9Y5S/p1652298497800099) for context.

For the hover state, this [figma prototype](https://www.figma.com/proto/PAOozuCxtfetqP8fUVvg0y/branch/xqSbhaPZdYYMax8LMOi8UP/Cypress-App?page-id=0%3A1&node-id=3799%3A60922&viewport=-7970%2C-3439%2C0.5&scaling=scale-down&starting-point-node-id=427%3A3959) shows the icon on the left being highlighted with color, and the spec title text darkening by a few shades when any part of the row is hovered. 

Since some parts of the row are now clickable, only clicking on non-interactive parts should trigger the navigation.

┆Issue is synchronized with this [Jira Bug](https://cypress-io.atlassian.net/browse/UNIFY-1434) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/12039">Screen Shot 2022-05-20 at 9.46.17 AM.png</a>
┆author: Mark Noonan
┆epic: UI/UX Improvements
┆friendlyId: UNIFY-1434
┆priority: Medium
┆sprint: Fast Follows 1
┆taskType: Bug

mike-plummer:
* Move click-sensitive row action wrapper to surround entire row
* Add styles to highlight spec icon on hocus to match Figma
* Small text highlight style change to match Figma

- Closes #21783 

### User facing changelog
* Entire row in Specs List will respond to clicks
* Styling updates to make current row in Specs List more obvious

### Additional details

[Figma design](https://www.figma.com/proto/PAOozuCxtfetqP8fUVvg0y/branch/xqSbhaPZdYYMax8LMOi8UP/Cypress-App?page-id=0%3A1&node-id=3799%3A60922&viewport=-7970%2C-3439%2C0.5&scaling=scale-down&starting-point-node-id=427%3A3959)

According to #21783 this was brought up a lot during v10 feedback testing. Current behavior only performs the row's action (toggle expansion for directories or navigate to spec for Spec rows) if you click directly on the row title item. This makes the entire row actionable by default.

Theoretically we may introduce row items that themselves want to respond to click events (opening a tooltip permanently, for example) - in that case the child can ensure the click event is not propagated upwards thus preventing the row from actioning.

### Steps to test
1. Open Cypress app into a project with existing tests within directories
2. Navigate to Specs List
3. Validate that clicking directly on a Spec Name navigates, validate that clicking the row *away from* the spec name also navigates
4. Validate that keyboard navigation to Spec row is operational and that pressing `enter` triggers navigation
5. Validate that clicking directly on a directory name/icon toggles expansion, validate that clicking the row *away from* the directory name/icon also toggles expansion
6. Validate that keyboard navigation to Directory row is operational and that pressing `enter` toggles expansion
7. Validate that hovering over a Spec row adds indigo coloration to spec icon

**Current**

https://user-images.githubusercontent.com/11482842/171961246-5ce90429-cf41-4f3b-95ac-df5456e0bc79.mov

**Updated**

https://user-images.githubusercontent.com/11482842/171961261-40ace242-5555-4292-ac2b-e7e502882344.mov

### PR Tasks

- [ na] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)?
- [ na] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?

### DIFF
--- a/packages/app/src/specs/SpecsList.vue
+++ b/packages/app/src/specs/SpecsList.vue
@@ -69,24 +69,18 @@
           :key="row.index"
           :data-cy="row.data.isLeaf ? 'spec-list-file' : 'spec-list-directory'"
           :data-cy-row="row.data.data?.baseName"
+          :is-leaf="row.data.isLeaf"
+          :route="{ path: '/specs/runner', query: { file: row.data.data?.relative?.replace(/\\/g, '/') } }"
+          @toggleRow="row.data.toggle"
         >
           <template #file>
-            <RouterLink
-              v-if="row.data.isLeaf && row.data"
-              :key="row.data.data?.absolute"
-              class="focus:outline-transparent"
-              :to="{ path: '/specs/runner', query: { file: row.data.data?.relative?.replace(/\\/g, '/') } }"
-              data-cy="spec-item-link"
-              @click.meta.prevent="handleCtrlClick"
-              @click.ctrl.prevent="handleCtrlClick"
-            >
-              <SpecItem
-                :file-name="row.data.data?.fileName || row.data.name"
-                :extension="row.data.data?.specFileExtension || ''"
-                :indexes="row.data.data?.fileIndexes"
-                :style="{ paddingLeft: `${((row.data.depth - 2) * 10) + 22}px` }"
-              />
-            </RouterLink>
+            <SpecItem
+              v-if="row.data.isLeaf"
+              :file-name="row.data.data?.fileName || row.data.name"
+              :extension="row.data.data?.specFileExtension || ''"
+              :indexes="row.data.data?.fileIndexes"
+              :style="{ paddingLeft: `${((row.data.depth - 2) * 10) + 22}px` }"
+            />
 
             <RowDirectory
               v-else
@@ -96,7 +90,7 @@
               :style="{ paddingLeft: `${(row.data.depth - 2) * 10}px` }"
               :indexes="getDirIndexes(row.data)"
               :aria-controls="getIdIfDirectory(row)"
-              @click="row.data.toggle"
+              @click.stop="row.data.toggle"
             />
           </template>
 
@@ -240,11 +234,6 @@ const { containerProps, list, wrapperProps, scrollTo } = useVirtualList(treeSpec
 // If you are scrolled down the virtual list and list changes,
 // reset scroll position to top of list
 watch(() => treeSpecList.value, () => scrollTo(0))
-
-function handleCtrlClick () {
-  // noop intended to reduce the chances of opening tests multiple tabs
-  // which is not a supported state in Cypress
-}
 
 function getIdIfDirectory (row) {
   if (row.data.isLeaf && row.data) {
--- a/packages/app/src/specs/SpecsListRowItem.vue
+++ b/packages/app/src/specs/SpecsListRowItem.vue
@@ -10,11 +10,11 @@
       @click.meta.prevent="handleCtrlClick"
       @click.ctrl.prevent="handleCtrlClick"
     >
-      <div>
+      <div data-cy="specs-list-row-file">
         <slot name="file" />
       </div>
 
-      <div>
+      <div data-cy="specs-list-row-git-info">
         <slot name="git-info" />
       </div>
     </component>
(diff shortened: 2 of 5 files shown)


packages/app/src/specs/SpecsListRowItem.vue
```diff
@@ -1,14 +1,40 @@
...
-    </div>
-
-    <div>
-      <slot name="git-info" />
-    </div>
+  <div data-cy="specs-list-row">
+    <component
+      :is="isLeaf ? 'RouterLink' : 'div'"
+      class="h-full outline-none border-gray-50 ring-inset grid group focus:outline-transparent focus-within:ring-indigo-300 focus-within:ring-1 children:cursor-pointer"
+      :class="[isLeaf ? 'grid-cols-2' : 'grid-cols-1']"
```
tbiethman, line 6: I noticed that the directory rows render a little off-center:

Screen Shot 2022-06-06 at 4 24 39 PM

I think that's because we flex this class to `grid-cols-1` when it's not a leaf, but we continue to render the git info div on line 17 regardless.

That kinda has me thinking why we need the git-info slot here at all? Could we instead make this a concern of the `SpecItem` component and consolidate all the 'leaf'-related logic in one place?

packages/app/src/specs/SpecsListRowItem.vue
```diff
@@ -1,14 +1,40 @@
...
-      <slot name="file" />
-    </div>
-
-    <div>
-      <slot name="git-info" />
-    </div>
+  <div data-cy="specs-list-row">
+    <component
+      :is="isLeaf ? 'RouterLink' : 'div'"
+      class="h-full outline-none border-gray-50 ring-inset grid group focus:outline-transparent focus-within:ring-indigo-300 focus-within:ring-1 children:cursor-pointer"
```
tbiethman, line 5: For the directory rows, it looks like there's some conflict between the transparent outline focus styles and the top border that acts as the divider:

Screen Shot 2022-06-06 at 4 52 35 PM

You might just need to use `focus:outline-none` within the RowDirectory component and let these focus-within styles take precedent.

packages/app/src/specs/SpecsListRowItem.vue
lmiller1990: I'm not sure if this is still the case but we used to have multiple views for this table, and I think some code was shared with the spec list on the runner page. That might explain some slots that don't seem to do much.

It's possible some old code is still hanging around from when we were still figuring the UI for this page out. Don't be afraid to delete code that seems weird/incorrect/not useful.

mike-plummer: I don't think this is a dead/legacy code issue, rather just that we currently don't display anything in the "Last Updated" slot for directories from the newly-added Git Info data in this table. Since there's a non-zero chance we may decide to put something in that column for directories in the future I reverted the style change so the directory renders as expected and just has an empty second column

flotwig (approved):

packages/app/src/specs/SpecsList.vue
```diff
@@ -96,7 +90,7 @@
               :style="{ paddingLeft: `${(row.data.depth - 2) * 10}px` }"
               :indexes="getDirIndexes(row.data)"
               :aria-controls="getIdIfDirectory(row)"
-              @click="row.data.toggle"
+              @click.stop="row.data.toggle"
```
tbiethman, line 93: Do we need both the click handler here and the one that emits the `@toggleRow` event from SpecsListRowItem? Seems like the SpecsListRowItem handler should catch clicks on the RowDirectory at this point.

tbiethman (approved):
Updates look good 👍 👍

lmiller1990 (approved):

PR #22119, opened 2022-06-06
fix: prevent Cypress from crashing when getting machine-id fails

### LINKED ISSUE #22110: Cypress >= 10 GUI crashes on Windows as non-admin user when registry editing is disabled
### Current behavior

When opening the GUI for Cypress and then clicking on type of test (either e2e or component, doesn't matter which), it displays "Initializing Config... Please what while we load your project and find browsers installed on your system" and shortly after crashes with the console error below:

Error: Error while obtaining machine id: Error: Command failed: %windir%\System32\REG.exe QUERY HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography /v MachineGuid
ERROR: Registry editing has been disabled by your administrator.

    at ChildProcess.exithandler (node:child_process:406:12)
    at ChildProcess.emit (node:events:390:28)
    at ChildProcess.emit (node:domain:475:12)
    at maybeClose (node:internal/child_process:1064:16)
    at Socket.<anonymous> (node:internal/child_process:450:11)
    at Socket.emit (node:events:390:28)
    at Socket.emit (node:domain:475:12)
    at Pipe.<anonymous> (node:net:687:12)

    at C:\Users\swaana\AppData\Local\Cypress\Cache\10.0.1\Cypress\resources\app\node_modules\node-machine-id\dist\index.js:1:7964
    at ChildProcess.exithandler (node:child_process:413:5)
    at ChildProcess.emit (node:events:390:28)
    at ChildProcess.emit (node:domain:475:12)
    at maybeClose (node:internal/child_process:1064:16)
    at Socket.<anonymous> (node:internal/child_process:450:11)
    at Socket.emit (node:events:390:28)
    at Socket.emit (node:domain:475:12)
    at Pipe.<anonymous> (node:net:687:12)

### Desired behavior

Getting the unique machine id should not crash for non admin users. The issue is caused by the underlying library being used (node-machine-id) which is calling %windir%\System32\REG.exe to fetch a unique id from the registry. Even though it is only reading the key the command itself is sensitive on Windows machines and therefore closed for non-admin users. This library seems to be badly maintained as the last publish was 3 years ago and for a year there has already been an [open PR](https://github.com/automation-stack/node-machine-id/pull/57) to fix this but without any response.

For my situation I am able to get this ID as a non-admin by using Powershell ("Powershell.exe -Command (Get-ItemProperty -Path HKLM:\SOFTWARE\Microsoft\Cryptography).MachineGuid") and there are probably other ways as well. So the issue is with the library itself and not with there being limitations due to user privileges.

### Test code to reproduce

Due to the system setup required for this I cannot reproduce this with a repo or a test. It requires a Windows setup with a user that is not an Administrator and perhaps also extra security preventing a user from running REG.exe as is the case for the company where I work.

### Cypress Version

10.0.1

### Other

_No response_

Swaana:
- Closes #22110 

### User facing changelog
The GUI will continue to run if getting machine-id fails.

### Additional details
The underlying npm package `node-machine-id` requires admin rights on Windows to get the machine id. This is causing the Cypress GUI to crash for Windows users that do not have administrator rights. The fix will allow for the failure as the ID is optional.

### Steps to test
To test this it requires a Windows machine and a non-admin user. Perhaps also an extra setting to prevent the user from running `%windir%\System32\REG.exe`. Running Cypress without this change will cause the application to crash right after selecting the type of test to perform, and after the change the application will run as normal.

The GUI will continue to run if getting machine-id fails.

### PR Tasks

### DIFF
--- a/packages/data-context/src/sources/VersionsDataSource.ts
+++ b/packages/data-context/src/sources/VersionsDataSource.ts
@@ -182,7 +182,7 @@ export class VersionsDataSource {
 
   private static async machineId (): Promise<string | undefined> {
     try {
-      return nmi.machineId()
+      return await nmi.machineId()
     } catch (error) {
       return undefined
     }
--- a/packages/data-context/test/unit/sources/VersionsDataSource.spec.ts
+++ b/packages/data-context/test/unit/sources/VersionsDataSource.spec.ts
@@ -90,7 +90,7 @@ describe('VersionsDataSource', () => {
     it('resets telemetry data triggering a new call to get the latest version', async () => {
       const currentCypressVersion = pkg.version
 
-      nmiStub.throws()
+      nmiStub.rejects('Error while obtaining machine id')
       ctx.coreData.currentTestingType = 'component'
 
       fetchStub


tgriesser:
> @tgriesser is this another example of something that we could catch higher up in the data-context in GUI mode?

@flotwig  This is another example of an `unhandledRejection` which we are currently treating as a fatal error in all cases. I think the adjustment would be to send error reports on these / log to the console but not crash open mode on them.

https://github.com/cypress-io/cypress/blob/e2243ee733c56f2c3c8c58877439fd2f250d15fd/packages/server/lib/unhandled_exceptions.js#L1-L7

tgriesser:
It's because we're not `await`'ing the promise:

```ts
try {
  return nmi.machineId()
} catch (error) {
  return undefined
}
```

vs 

```ts
try {
  return await nmi.machineId()
} catch (error) {
  return undefined
}
```

the former will lead to an `unhandledRejection` if it throws and isn't caught higher up in the chain - which it isn't because it's not in the `try` block in `getLatestVersion`, same problem as https://github.com/cypress-io/cypress/pull/22099. 

Has nothing to do with GraphQL, it's more to do with the fact that we're now using async/await rather than promise chains everywhere and not always ensuring there's proper awaiting. I think it's something we can introduce linting for but IIRC there were problems with false positives & negatives.

tgriesser (approved):

PR #22168, opened 2022-06-07
fix: properly run multiple specs in run and headed mode on linux and windows in chrome

### LINKED ISSUE #22155: `cypress run`ing multiple specs in headed mode, in chrome, on linux will fail after the first spec
### Current behavior

Running `cypress run` with multiple specs in headed mode, in chrome, on linux will fail after the first spec with the following error:

image

### Desired behavior

All specs should be able to run with no error

### Test code to reproduce

1. Run `cypress run --headed --browser chrome` on a project with multiple specs
2. Notice that the second spec fails with the error `There was an error reconnecting to the Chrome DevTools protocol. Please restart the browser.`

### Cypress Version

10.0.0

### Other

_No response_

ryanthemanuel:
- Closes #22155 

### User facing changelog

Users can now properly `cypress run` multiple specs in headed mode on linux and windows in chrome

### Additional details

In 10.0.0 we updated run mode to close a browser tab and open a new one in between specs rather than killing the browser instance and launching a new one. Headed mode on linux and windows works slightly differently than in other circumstances in that when the last browser tab closes, the browser process exits. This prevents us from being able to launch a new tab since we no longer have a connection to the browser. We will fix this by launching a new tab to `about:blank` prior to closing the last tab. This will allow us to maintain that connection.

### Steps to test

1. Pick a project with multiple e2e specs
2. On linux and/or windows, run `cypress run --headed --browser --chrome` for the project
3. Ensure that all specs succeed

Before:

https://user-images.githubusercontent.com/4873279/172425241-3871575d-fc6e-49d6-8f94-418b222bcd11.mov

After:

https://user-images.githubusercontent.com/4873279/172425210-16467a9a-da2b-470f-95e7-819dda980b00.mov

### PR Tasks


packages/server/lib/modes/run.js
```diff
@@ -1251,7 +1251,7 @@ module.exports = {
       //   await openProject.closeBrowser()
       // } else {
       debug('attempting to close the browser tab')
-      await openProject.closeBrowserTabs()
+      await openProject.resetBrowserTabsForNextTest(shouldKeepTabOpen)
```
ryanthemanuel, line 1254: Passing this `shouldKeepTabOpen` variable around allows us to not open a new tab on the last spec. It's a lot of changes but there's probably a small performance improvement on the last spec at least.

packages/server/lib/browsers/browser-cri-client.ts
```diff
@@ -125,35 +129,28 @@ export class BrowserCriClient {
...
-   */
-  closeCurrentTarget = async (): Promise<void> => {
+  resetBrowserTargets = async (shouldKeepTabOpen: boolean): Promise<void> => {
     if (!this.currentlyAttachedTarget) {
       throw new Error('Cannot close target because no target is currently attached')
     }
 
+    let target
+
+    if (shouldKeepTabOpen) {
```
ryanthemanuel, line 150:
```suggestion
    // If we are keeping a tab open, we need to first launch a new default tab prior to closing the existing one
    if (shouldKeepTabOpen) {
```

packages/server/lib/browsers/browser-cri-client.ts
```diff
@@ -67,7 +67,7 @@ const retryWithIncreasingDelay = async <T>(retryable: () => Promise<T>, browserN
 }
 
 export class BrowserCriClient {
-  private currentlyAttachedTarget: CRIWrapper.Client | undefined
+  private _currentlyAttachedTarget: CRIWrapper.Client | undefined
```
BlueWinds, line 70: I consider private values with public getters to be a JS anti-pattern - it adds code and verbosity for no good reason that I can tell. Very much like to avoid C# or Java getters and setters in our code.

(enforcing access control is not, IMO, a good reason)

BlueWinds (approved):

ZachJW34 (approved):

PR #22175, opened 2022-06-07
fix: retry on EMFILE always, lint sync FS calls

### LINKED ISSUE #22023: EMFILE eventually crashes `cypress open` on Linux
### Current behavior

If you're near the open file limit, cypress will crash with an EMFILE error when you try to use it. 
image

Since it's EMFILE (too many open files) and not an error about "too many file watchers", I think we are either accidentally opening way too many files, or are not using fs-extra's EMFILE queuing like we should be, or probably both.

### Desired behavior

Don't crash

### Test code to reproduce

--	

### Cypress Version

10.0.0

### Other

_No response_

flotwig:
- Closes #22023
- Addresses parts of #22028

### User facing changelog

- Fixed an issue where Cypress would crash with an EMFILE error if the process ran into the file descriptor limit. Cypress will now retry file reads using `graceful-fs`.

### Additional details

- Applies `graceful-fs` to `fs` on startup+nextTick to block EMFILE errors - previously, only `fs-extra` had `graceful-fs` protections
- ~~Adds warnings for sync FS method usage in the server when not in production~~
	- I moved this to another PR: #22176 Since it will require more work to update sync node_modules/remaining existing sync call sites so we aren't flooded by warnings.
- Adds lint rules for sync FS method usage with ignores for scripts, tests, and non-server code
	- The `no-restricted-syntax` rule is crazy versatile, you can write arbitrary AST queries to lint using `esquery` syntax.
- Updates trivial-to-fix places where we use sync FS methods to use async FS methods
- Adds eslint-ignores in areas where it doesn't matter (debug code, etc...)
- Adds TODOs and eslint-ignores where we need to do further work to remove sync FS calls

### PR Tasks


packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
```diff
@@ -563,16 +565,11 @@ describe('_makeSpecWatcher', () => {
 
     specWatcher.on('add', (filePath) => allFiles.add(filePath))
     specWatcher.on('change', (filePath) => allFiles.add(filePath))
+    specWatcher.on('unlink', (filePath) => allFiles.delete(filePath))
```
flotwig, line 568: @tgriesser I didn't think we'd be seeing this test so soon again :P I had to modify this test to pass after making so much stuff `async`, and this line was missing. I think the test is still testing what it's supposed to, can you verify?

.eslintrc.js
```diff
@@ -55,6 +74,16 @@ module.exports = {
         message: 'os.userInfo() will throw when there is not an `/etc/passwd` entry for the current user (like when running with --user 12345 in Docker). Do not use it unless you catch any potential errors.',
       },
     ],
+    'no-restricted-syntax': [
```
lmiller1990, line 77: Wow

npm/vite-dev-server/src/plugins/cypress.ts
```diff
@@ -38,6 +38,8 @@ export const Cypress = (
   const indexHtmlFile = options.cypressConfig.indexHtmlFile
 
   let specsPathsSet = getSpecsPathsSet(specs)
+  // TODO: use async fs methods here
+  // eslint-disable-next-line no-restricted-syntax
   let loader = fs.readFileSync(INIT_FILEPATH, 'utf8')
```
lmiller1990, line 43: Does this "do not use sync FS calls" also need to be applied in npm, or just things we know are shipping in the binary (and never published separately on npm). It's possible a user is consuming these outside of Cypress (however unlikely) and they may not have access to wrapped `fs` module with async extensions. 

Seems safest to either us things we know exist in the Node.js standard lib, or have these depend on `fs-extra`, or use `fs.promises` (didn't know about this!) to making async versions of the functions we need in each `npm` module.

lmiller1990 (commented):
One comment/question around extending this to `npm`, the rest looks like pretty simple migrations from sync->async. :clap: 

Also windows is passing on this PR - did you re-run it, or did these changes reduce some flake?

npm/vite-dev-server/src/plugins/cypress.ts
flotwig: I'd use `fs.promises` here if I was gonna change this. I started to, but then stopped because it actually requires a user-facing (I think?) type change to make this function async.

Since this is most likely to run in the plugins process, it should be fine to just use `fs.promises`, since we already run `gracefulFs.gracefulify(fs)` in the child process:

flotwig:
> Also windows is passing on this PR - did you re-run it, or did these changes reduce some flake?

I think I just got lucky.

npm/vite-dev-server/src/plugins/cypress.ts
lmiller1990: Not too sure this would require a user facing change (this function is internal to the Vite dev server) but sure, happy to leave it like this for now. If we do change it `fs.promises` seems good.

lmiller1990 (approved):
Seems fine, got an answer on my one question.

tbiethman (approved):
Changes are thorough :+1: I ran a few smoke tests locally and everything is working good. Also added a few sync calls to test the new eslint rule, that is very cool.

Should we log issues for the few remaining TODOs or will we just keep https://github.com/cypress-io/cypress/issues/22028 open for that work?

flotwig:
@tbiethman I'll update #22028, I think it can encompass the remaining work on this (warn on all sync usage on a monkey-path level, remove all sync usage)

PR #22189, opened 2022-06-08
chore: Update Volar identifier for VSCode Marketplace

astone123:
### Additional details

When installing recommended VSCode extensions in my IDE, I noticed this warning

Screen Shot 2022-06-08 at 9 37 13 AM

It looks like the identifier for the Volar extension changed from `johnsoncodehk.volar` to `vue.volar` at some point. This PR updates the recommended extensions to use the new identifier so that VSCode can find the Volar extension when installing recommended extensions. The links to the VSCode marketplace on the web seemed to still work, but I updated those as well just so that we're using the correct identifier.

### PR Tasks


flotwig (approved):
Thanks for opening a PR @astone123

mike-plummer (approved):

astone123:
@mike-plummer @flotwig any idea why Percy is showing visual changes for this? Seems like there shouldn't be any based on my changes

marktnoonan:
The changes in Percy are all flake, some have to do with loaders/spinners, a lot are an element in the header that actually _is_ flaky in the launchpad and has a ticket. Percy is great but with the new app some of the snapshots haven't been given enough attention around exactly at what moment to take the snapshot, or what elements to block out. So we get quite a bit of noise in the diffs based on the position of a spinner or the precise status in the command log, which is based on performance at the time, and not deterministic. There's some WIP out there that will address a lot of this, though I bet we can make small improvements before that lands.

lmiller1990:
Nice, going to merge - windows CI is red here, but I'm confident the errors are resolved in https://github.com/cypress-io/cypress/pull/22234#pullrequestreview-1002252834, and you did not introduce any just by changing some markdown files.

PR #22226, opened 2022-06-09
fix: Truncate spec directories and file names instead of wrapping

### LINKED ISSUE #21895: Truncate spec name/folder when it's too long
## **Current Behavior**
While working on [https://cypress-io.atlassian.net/browse/UNIFY-1661](https://cypress-io.atlassian.net/browse/UNIFY-1661|smart-link)  found that long directories are being shown completely and not truncated

image-20220504-182935

**Desired Behavior**

Truncate spec name/folder if it’s too long

## **Reproducible Steps**
1. Create this spec `super-long-directory-name-with-children-long-long-long-long/super-long-directory-name-with-children-long-long-long-long copy/super-long-directory-name-with-children-long-long-long-long02/super-long-directory-name-with-children-long-long-long-long-2/spec-super-long-directory-name-with-children-long-long-long-long-2.cy.js``
1. visit the app

┆Issue is synchronized with this [Jira Bug](https://cypress-io.atlassian.net/browse/UNIFY-1726) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/11927">image-20220504-182935.png</a>
┆author: Alejandro Estrada
┆epic: UI/UX Improvements
┆friendlyId: UNIFY-1726
┆priority: Low
┆sprint: Backlog
┆taskType: Bug

astone123:
- Closes #21895

### User facing changelog

Spec directories and file names are now truncated if they exceed a certain width

### Additional details

In the spec list, directory paths and file names were wrapping to show the entire value rather than truncating when they overflowed max width. This lead to long file/dir names clogging up the UI and overflowing the set height for the row.

This change affects the spec lists for e2e and component testing. See the related issue for more details

### Steps to test

- Open the spec list for an application (e2e or ct) and create a directory and/or spec file with a long name (shrink window width if necessary)
- Verify that when the file/dir name is longer than the width of the spec column in the table, the name is truncated
- Verify that you can view the full name of the file/dir by hovering over the truncated text

**Before**
Screen Shot 2022-06-09 at 12 08 29 PM

**After**
Screen Shot 2022-06-09 at 12 14 37 PM

### PR Tasks

### DIFF
--- a/packages/app/src/specs/HighlightedText.vue
+++ b/packages/app/src/specs/HighlightedText.vue
@@ -1,7 +1,6 @@
 <template>
   <span
     :title="props.titleText"
-    :class="{'truncate': truncate}"
   >
     <span
       v-for="({char, highlighted}, idx) in characters"
@@ -24,8 +23,8 @@
 import { computed } from 'vue'
 
 const props = withDefaults(
-  defineProps<{text: string, indexes: number[], highlightClasses?: string, titleText?: string, truncate?: boolean}>(),
-  { text: '', indexes: () => [], highlightClasses: 'text-white', titleText: '', truncate: false },
+  defineProps<{text: string, indexes: number[], highlightClasses?: string, titleText?: string}>(),
+  { text: '', indexes: () => [], highlightClasses: 'text-white', titleText: '' },
 )
 
 const characters = computed(() => {
--- a/packages/app/src/specs/HighlightedText.vue
+++ b/packages/app/src/specs/HighlightedText.vue
@@ -1,7 +1,5 @@
 <template>
-  <span
-    :title="props.titleText"
-  >
+  <span>
     <span
       v-for="({char, highlighted}, idx) in characters"
       :key="idx"
@@ -23,8 +21,8 @@
 import { computed } from 'vue'
 
 const props = withDefaults(
-  defineProps<{text: string, indexes: number[], highlightClasses?: string, titleText?: string}>(),
-  { text: '', indexes: () => [], highlightClasses: 'text-white', titleText: '' },
+  defineProps<{text: string, indexes: number[], highlightClasses?: string}>(),
+  { text: '', indexes: () => [], highlightClasses: 'text-white' },
 )
 
 const characters = computed(() => {
--- a/packages/app/src/specs/HighlightedText.vue
+++ b/packages/app/src/specs/HighlightedText.vue
@@ -1,5 +1,8 @@
 <template>
-  <span>
+  <span
+    :title="props.titleText"
+    :class="{'truncate': truncate}"
+  >
     <span
       v-for="({char, highlighted}, idx) in characters"
       :key="idx"
@@ -21,8 +24,8 @@
 import { computed } from 'vue'
 
 const props = withDefaults(
-  defineProps<{text: string, indexes: number[], highlightClasses?: string}>(),
-  { text: '', indexes: () => [], highlightClasses: 'text-white' },
+  defineProps<{text: string, indexes: number[], highlightClasses?: string, titleText?: string, truncate?: boolean}>(),
+  { text: '', indexes: () => [], highlightClasses: 'text-white', titleText: '', truncate: false },
 )
 
 const characters = computed(() => {
--- a/packages/app/src/specs/SpecFileItem.cy.tsx
+++ b/packages/app/src/specs/SpecFileItem.cy.tsx
@@ -24,7 +24,7 @@ describe('SpecFileItem', () => {
   })
 
   it('truncates spec name if it exceeds container width and provides title for full spec name', () => {
-    const specFileName = 'VeryLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongSpecName'
+    const specFileName = `${'Long'.repeat(20)}Name`
 
     // Shrink viewport width so spec name is truncated
     cy.viewport(400, 850)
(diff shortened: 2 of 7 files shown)


packages/app/src/specs/HighlightedText.vue
```diff
@@ -21,8 +24,8 @@
 import { computed } from 'vue'
 
 const props = withDefaults(
-  defineProps<{text: string, indexes: number[], highlightClasses?: string}>(),
-  { text: '', indexes: () => [], highlightClasses: 'text-white' },
+  defineProps<{text: string, indexes: number[], highlightClasses?: string, titleText?: string, truncate?: boolean}>(),
```
lmiller1990, line 27: Seems truncate will never be false from what I can see? Do we need this prop?

lmiller1990 (commented):
Might be missing something, but seems like truncate is never false? Any reason we need it as a prop?

Also might be good to add a test - have a huge spec, then make sure only the first N char are rendered. What do you think?

packages/app/src/specs/HighlightedText.vue
astone123: There are some cases where `truncate` will be false:

- Use in `SpecItem.vue`. In this case, the outer div has the `truncate` class on it since there are two instances of `HighlightedText` being used side by side in this component (one for the file name and one for the extension), so using the `truncate` prop wouldn't do what we want in this case.
- Use in `SpecFileItem.vue`. Similar situation
- Use in `DirectoryItem.vue`. In this case the truncation is handled by the parent component as well.

I am re-thinking adding this as a prop though... seems like we should handle the truncation outside of the `HighlightedText` component. I'll mess with that

packages/app/src/specs/SpecFileItem.vue
```diff
@@ -9,17 +9,21 @@
...
-      :indexes="indexes.filter((idx) => idx < fileName.length)"
-      class="font-medium pl-8px whitespace-nowrap"
-      :class="selected ? 'text-white' : 'group-focus:text-indigo-300 text-gray-400 group-hover:text-indigo-300'"
-    />
-    <HighlightedText
-      :text="extension"
-      :indexes="indexes.filter((idx) => idx >= fileName.length).map(idx => idx - fileName.length)"
-      class="text-gray-700"
-    />
+    <div class="text-gray-400 truncate">
```
astone123, line 12: Needed to apply the text color class to the outer container otherwise the ellipsis wouldn't match the text color for the spec

packages/app/src/specs/SpecFileItem.cy.tsx
```diff
@@ -22,4 +22,23 @@ describe('SpecFileItem', () => {
 
     cy.percySnapshot()
   })
+
+  it('truncates spec name if it exceeds container width and provides title for full spec name', () => {
+    const specFileName = 'VeryLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongSpecName'
```
lmiller1990, line 27: Nit: can be more concise:

```js
`${'Long'.repeat(20)}Name`
```

lmiller1990 (approved):

ryanjwilke (approved):
Screenshots look good

packages/app/src/specs/HighlightedText.vue
```diff
@@ -1,5 +1,7 @@
 <template>
-  <span>
+  <span
+    :title="props.titleText"
```
marktnoonan, line 3: The `title` attribute isn't ideal for this, it renders differently depending on the operating system and is not exposed to keyboard users since it only appears when you hover with the mouse. I think we should be using our existing Tooltip component here. See in this screenshot, the title on the left vs the standard tooltip on the right:

Screen Shot 2022-06-13 at 7 32 37 AM

One other issue here is that since this `HighlightedText` component is used twice to show what matches the search term, we end up having two span elements right next to each other with the same title. We only need it once. Whether we switch to tooltip or not, we can move this up a level in the DOM - to the div that wraps the two instances of `HighlightedText`? That can either have the title element or be the activator for the tooltip.

We already do the "truncate + tooltip" pattern is `HeaderBarContent.vue` which could be another good reference.

marktnoonan (changes requested):
This looks good, much better than the wrapping! 

My recommendation is that we should use our regular Tooltip component instead of the `title` attribute to maintain consistency with other tooltips on the page and be more accessible for keyboard navigation. More details in the comment.

I did discover some other issues with the current markup while testing this. Those are unrelated to this PR though. An example is that wrapping the whole row in a link somewhat clobbers the internal contents of the row for screen readers. On the longer names, chrome appears to cut off after 95 characters when putting the link's name into the accessibility tree. So the git info isn't announced, we just trail off in the middle of the file name.

We are probably going to need to revisit the underlying markup here sooner rather than later as we add more columns. Ultimately this list should probably become a HTML table. So if using the tooltip introduces problems, let's keep it a title. This PR is still an improvement, and as we deal with the increasing complexity of this list, we can revisit the tooltips holistically.

packages/app/src/specs/HighlightedText.vue
astone123: @marktnoonan I agree that we should use tooltips here rather than the title attribute. I looked into doing that and I think that the scope is going to be more than this PR since if we go forward with tooltips I'd like to conditionally render them based on whether or not the text is actually being truncated or not, which is going to involve some extra logic.

What do you think about moving forward with this as-is and changing from title attribute to truncation in a separate task? I will move the title attribute for the spec items to a higher level in the DOM here though

astone123: @marktnoonan I created an issue to handle this #22323

marktnoonan: I thought we had settled on just a delay opening (with same timing as the `title` attr), but sure let's keep the title attr for now and get this improvement in.

marktnoonan (approved):
Tried it out locally and all works well 👍

tbiethman (approved):

PR #22234, opened 2022-06-09
fix: windows

ryanthemanuel:
### User facing changelog

N/A

### Additional details

The main goal here is to fix flaky tests in windows. This was accomplished via:

* Bump the resource class of the windows executor
* Add some waits to a few places that seem to rely on `postFragment` subscriptions

Neither of these are ideal solutions; however, they don't appear to be new problems (I can recreate them in at least v10.0.3). Consequently we want to make the builds more stable and follow up on these issues later.

### Steps to test

Check that CI succeeds

N/A

### PR Tasks


circle.yml
```diff
@@ -2619,15 +2619,15 @@ windows-workflow: &windows-workflow
...
+        resource_class: windows.xlarge
         requires:
           - windows-build
 
     - create-build-artifacts:
         <<: *full-windows-workflow-filters
         name: windows-create-build-artifacts
         executor: windows
-        resource_class: windows.medium
+        resource_class: windows.xlarge
```
lmiller1990, line 2630: A little unsettling that we don't know why this is required, but glad to have windows CI back to green.

packages/data-context/src/actions/MigrationActions.ts
```diff
@@ -32,9 +32,6 @@ import {
 } from '../sources/migration'
```
lmiller1990, line 32: Just to clarify for reviewers: every file change below this line is reverted from https://github.com/cypress-io/cypress/pull/22118 and you don't need to review it to review this PR.

lmiller1990 (approved):
Looks good, I will find out why https://github.com/cypress-io/cypress/pull/22118 has (frustrating) windows on CI fails. I'm fairly sure they are just flake, they've passed before, but I'll find out.

lmiller1990:
Hey @ryanthemanuel I do not think we need to revert #22118 anymore - can you just merge [this commit](https://github.com/cypress-io/cypress/commit/c5f1025e7a10b53372aaa6ef88ace74f2eed65c0) and it'll be fine. I checked out #22118, added that one commit, and CI is green. [Link to CI run with 100% windows tests passing first run](https://app.circleci.com/pipelines/github/cypress-io/cypress?branch=lmiller1990%2F21939-with-windows-fix&filter=all).

circle.yml
mjhenkes: Agreed. 😕

mjhenkes (approved):

PR #22250, opened 2022-06-10
feat: Add support for optional env key to browser config object

ryanjduffy:
### User facing changelog

n/a

### Additional details

In building out support to [integrate replay with cypress](https://github.com/replayio/replay-cli/tree/main/packages/cypress),  I discovered there isn't a way to set environment variables for the browser without putting the onus on the end user. Further, since the browser launch seems to happen in a different process than the plugin, I couldn't inject those values into `process.env` and have them picked up by the existing logic.

This adds support for an optional `env` key on the `Browser` which is then merged with the `defaultBrowserEnv` and `process.env`. I placed it between those two so the default browser family config (`defaultBrowserEnv`) would be overwritten by browser instance config (`browser.env`) which would be overwritten by the current `process.env`.

Some [prior art from Playwright](https://playwright.dev/docs/api/class-browsertype#browser-type-launch)

### Steps to test

Still figuring out the testing process for this. I monkeypatched it into a project to test the implementation.

### PR Tasks


emilyrohrbough:
@ryanjduffy Thank you for opening the pull request for this change. I would like to learn a bit more about what you are trying to accomplish here and why this change is needed. I think I have a good assumption of how you intended to utilize this change, but I'd like to fully understand before providing feedback.

I do want to know, I see you have added an additional key to the `browser` object to allow passing environment variables. This `browser` object is not a configuration option for Cypress, rather it is the result of browser that has been detected by Cypress when a run starts. 

You said you tried to config this from the plugins - this process is kicked off once the browser has successfully launched. I do however have a [`before:browser:launch` event](https://docs.cypress.io/api/plugins/browser-launch-api) that you can hook into to configure your browser launch options. It does not support ENV at this time, however it does allow for browser args to be passed to configure how the browser is launched. Have you looked into this API at all?

ryanjduffy:
:wave: @emilyrohrbough 

Replay uses a couple environment variables to configure our browser forks. Some are required (e.g. `RECORD_ALL_CONTENT` for firefox) and some may need to vary by browser instance when running in parallel (like `RECORD_REPLAY_METADATA_FILE`). We've tried to build these into our plugins/adapters/whatever you want to call them so users aren't encumbered by the implementation details.

* Our [playwright adapter](https://github.com/replayio/replay-cli/blob/d60a0f22d2b16e23be6feea6bc1e36828040f7c6/packages/playwright/src/index.ts#L11-L32) which sets the relevant env variables
* Our [cypress plugin](https://github.com/replayio/replay-cli/blob/d60a0f22d2b16e23be6feea6bc1e36828040f7c6/packages/cypress/src/index.ts#L14-L46) which uses `before:browser:launch` (but of course not to configure the ENV :) )

emilyrohrbough:
@ryanjduffy Thanks for the response. I see what you are attempting to do here. Looks like you plugin is injecting a custom detected browser in the Cypress configuration and you'd like to expand the configuration to include envs to handle setting the envs for users when they leverage your plugins. That's great you're looking to set them up for success!

Updating the browser types like this is not how'd I'd expect to go about this change, as this is a not how Cypress-detected browsers will be able to pick up environment variables. The solution I'd expect to work before for all use cases is to add `env` to the [`launchOptions` API](https://docs.cypress.io/api/plugins/browser-launch-api#Syntax) to allow individuals to configure these from the `before:launch:browser` hook.

```js
setupNodeEvents(on, config) {
  on('before:browser:launch', (browser, launchOptions) => {
    launchOptions.env ={
      RECORD_ALL_CONTENT: 1
    }

    return launchOptions
  })
  return config
},
```
Please let me know if you are still interested in taking this work on. If so, I can act as a resource to get you on the right track if you need any help!

ryanjduffy:
@emilyrohrbough That sounds reasonable to me. I'll try that out soon. Thanks for the feedback.

emilyrohrbough:
@ryanjduffy Awesome! I am going to move this PR into a draft for the time being while you work on this.

emilyrohrbough:
@ryanjduffy Hey! Have you had a chance to take a look at this?

ryanjduffy:
Sadly, no. It's still on my radar but I haven't gotten into it.

emilyrohrbough:
@ryanjduffy no problem! Just wanted to check in.

emilyrohrbough:
@ryanjduffy Are you fine if I close this till you are ready to take another look?

PR #22269, opened 2022-06-12
feat: Column width updates for latest runs and average duration work

pstakoun:
- Closes 

### User facing changelog

Changes the Spec Explorer column sizes and adds padding to the rightmost column for Dashboard Integration data so the data is not flush with the scrollbar.

### PR Tasks

### DIFF
--- a/packages/app/src/specs/SpecsList.vue
+++ b/packages/app/src/specs/SpecsList.vue
@@ -69,17 +69,17 @@
     />
     <div
       v-if="specs.length"
-      class="mb-4 grid grid-cols-10 children:font-medium children:text-gray-800"
+      class="mb-4 grid grid-cols-12 children:font-medium children:text-gray-800"
       :style="`padding-right: ${scrollbarOffset}px`"
     >
       <div
-        class="flex col-span-4 items-center justify-between"
+        class="flex col-span-5 items-center justify-between"
         data-cy="specs-testing-type-header"
       >
         {{ props.gql.currentProject?.currentTestingType === 'component' ?
           t('specPage.componentSpecsHeader') : t('specPage.e2eSpecsHeader') }}
       </div>
-      <div class="flex items-center justify-between whitespace-nowrap">
+      <div class="flex col-span-2 items-center justify-between truncate">
         <LastUpdatedHeader :is-git-available="isGitAvailable" />
       </div>
       <div class="flex col-span-2 items-center justify-end truncate">
@@ -171,7 +171,7 @@
               />
               <div
                 v-else-if="row.data.isLeaf && row.data.data?.cloudSpec?.fetchingStatus === 'FETCHING'"
-                class="bg-gray-50 rounded-[20px] w-80px animate-pulse"
+                class="bg-gray-50 rounded-[20px] w-full animate-pulse"
               >
                 &nbsp;
               </div>
--- a/packages/app/src/specs/SpecsList.vue
+++ b/packages/app/src/specs/SpecsList.vue
@@ -80,7 +80,7 @@
     <div
       v-if="specs.length"
       class="mb-4 grid grid-cols-5 children:font-medium children:text-gray-800"
-      :style="`padding-right: ${scrollbarOffset && scrollbarOffset + 20}px`"
+      :style="`padding-right: ${scrollbarOffset + 20}px`"
     >
       <div
         class="flex col-span-2 items-center justify-between"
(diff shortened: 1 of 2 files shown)


packages/app/src/specs/SpecsList.vue
```diff
@@ -69,20 +69,20 @@
...
       <div
-        class="flex col-span-3 items-center justify-between"
+        class="flex col-span-4 items-center justify-between"
         data-cy="specs-testing-type-header"
       >
         {{ props.gql.currentProject?.currentTestingType === 'component' ?
           t('specPage.componentSpecsHeader') : t('specPage.e2eSpecsHeader') }}
       </div>
-      <div class="flex col-span-2 items-center justify-between">
+      <div class="flex items-center justify-between whitespace-nowrap">
```
pstakoun, line 82: not sure if there is a better way to do this - this was the only header that was wrapping on smaller screens

packages/app/src/specs/SpecsList.vue
```diff
@@ -170,7 +171,7 @@
               />
               <div
                 v-else-if="row.data.isLeaf && row.data.data?.cloudSpec?.fetchingStatus === 'FETCHING'"
-                class="bg-gray-50 rounded-[20px] w-full animate-pulse"
+                class="bg-gray-50 rounded-[20px] w-80px animate-pulse"
```
pstakoun, line 174: what I would love to do here is do full width unless it would overlap with the previous column, and then do the max that doesn't overlap, but I did this for simplicity - still debating if it makes it better or worse

pstakoun: Here's an example of what I mean:

Screen Shot 2022-06-12 at 9 37 44 PM

MuazOthman: It seems to be correctly reflecting the 1:2 ratio for Last updated vs. Latest runs, keeping in mind Last updated is left aligned and Latest runs is right aligned. I believe the overlapping is the text in Last updated overflowing. I added background colors to these columns in the screenshot below to visualize it. I think we need to handle the text overflow for the first 2 columns. 
image

MuazOthman: I was PR-surfing when I found this PR seemingly addressing the text wrapping issue #22226

packages/app/src/specs/SpecsList.vue
```diff
@@ -102,6 +102,7 @@
           @showConnectToProject="showConnectToProject"
         />
       </div>
+      <div />
```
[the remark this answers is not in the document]
pstakoun: We totally can. My thinking was that on larger widths an additional column gives a bit more padding to center the data a bit more, but this is not entirely necessary.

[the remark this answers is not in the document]
pstakoun: It would probably have to be inner padding or something if we wanted to maintain the column widths right? Not entirely sure on where would be best to put it without misaligning them.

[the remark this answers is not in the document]
tbiethman: Setting right padding here should do it: https://github.com/cypress-io/cypress/blob/peter/feat/aci-columns/packages/app/src/specs/SpecsList.vue#L116

But we'd need to add whatever value we use there to the scrollbarOffset value to keep the column headers aligned: https://github.com/cypress-io/cypress/blob/peter/feat/aci-columns/packages/app/src/specs/SpecsList.vue#L73

MuazOthman (approved):

packages/app/src/specs/SpecsList.vue
```diff
@@ -79,17 +79,17 @@
     />
     <div
       v-if="specs.length"
-      class="mb-4 grid grid-cols-7 children:font-medium children:text-gray-800"
-      :style="`padding-right: ${scrollbarOffset}px`"
+      class="mb-4 grid grid-cols-5 children:font-medium children:text-gray-800"
+      :style="`padding-right: ${scrollbarOffset && scrollbarOffset + 20}px`"
```
tbiethman, line 83: Is there a reason we don't want to include a little padding, even when scrollbars may not be present? macOS temporary scrollbars can overlay the content as-is:

image

pstakoun: @tbiethman what browser does this?

tbiethman: That was from Chrome on macOS, with "show scroll bars" set to automatic:

Screen Shot 2022-06-14 at 1 07 59 PM

pstakoun: Addressed in last commit

PR #22270, opened 2022-06-13
chore: add component testing metadata to manifest fetch request

### LINKED ISSUE #21880: Add the configured dev server and framework to the metadata that is sent on the version data source call
## **Summary**
The version data source call to `https://download.cypress.io/desktop.json` is used to track usage. Currently, we send:

* Cypress Version
* OS Name
* Architecture
* Testing Type
* Machine ID

We would like to add the following to this list:

* Configured Dev Server for Component Testing
* Configured Framework for Component Testing

┆Issue is synchronized with this [Jira Task](https://cypress-io.atlassian.net/browse/UNIFY-1703) by [Unito](https://www.unito.io)
┆author: Ryan Manuel
┆epic: Chores
┆friendlyId: UNIFY-1703
┆priority: High
┆sprint: Fast Follows 2
┆taskType: Task

lmiller1990:
- Closes #21880

### User facing changelog

N/A

### Additional details

Includes`x-framework` and `x-dev-server` headers when requesting the latest Cypress version.

### Steps to test

Note: we only make this request in open mode.

There's no easy way to test this without modifying the code or monitoring your network traffic. You could do:

1. Pick a project (any, say launchpad or app) with component testing configured.
2. Add `console.log(manifestHeaders)` here: https://github.com/cypress-io/cypress/compare/lmiller1990/21880?expand=1#diff-f4439168518f09f406948b6f67428518c25a6ac04a9b766f86b2dfbe1964ac2eR1633. 
3. See if the `x-framework` and `x-dev-server` are correct.4. 

I also added a test that verifies this.

### PR Tasks

### DIFF
--- a/packages/data-context/src/sources/VersionsDataSource.ts
+++ b/packages/data-context/src/sources/VersionsDataSource.ts
@@ -29,16 +29,15 @@ export class VersionsDataSource {
   constructor (private ctx: DataContext) {
     this._initialLaunch = true
     this._currentTestingType = this.ctx.coreData.currentTestingType
-    this.#ensureData()
   }
 
-  #ensureData () {
+  #populateVersionMetadata () {
     let versionData = this.ctx.coreData.versionData
 
     if (!versionData) {
       versionData = {
-        latestVersion: this.getLatestVersion().catch((e) => pkg.version),
-        npmMetadata: this.getVersionMetadata().catch((e) => ({})),
+        latestVersion: this.#getLatestVersion().catch((e) => pkg.version),
+        npmMetadata: this.#getVersionMetadata().catch((e) => ({})),
       }
 
       this.ctx.update((d) => {
@@ -65,7 +64,7 @@ export class VersionsDataSource {
    * }
    */
   async versionData (): Promise<VersionData> {
-    const versionData = this.#ensureData()
+    const versionData = this.#populateVersionMetadata()
     const [latestVersion, npmMetadata] = await Promise.all([
       versionData.latestVersion,
       versionData.npmMetadata,
@@ -93,13 +92,13 @@ export class VersionsDataSource {
       this._currentTestingType = this.ctx.coreData.currentTestingType
       this.ctx.update((d) => {
         if (d.versionData) {
-          d.versionData.latestVersion = this.getLatestVersion()
+          d.versionData.latestVersion = this.#getLatestVersion()
         }
       })
     }
   }
 
-  private async getVersionMetadata (): Promise<Record<string, string>> {
+  async #getVersionMetadata (): Promise<Record<string, string>> {
     const now = new Date().toISOString()
     const DEFAULT_RESPONSE = {
       [pkg.version]: now,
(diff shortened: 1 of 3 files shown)


packages/data-context/src/sources/VersionsDataSource.ts
```diff
@@ -93,13 +92,13 @@ export class VersionsDataSource {
...
         if (d.versionData) {
-          d.versionData.latestVersion = this.getLatestVersion()
+          d.versionData.latestVersion = this.#getLatestVersion()
         }
       })
     }
   }
 
-  private async getVersionMetadata (): Promise<Record<string, string>> {
+  async #getVersionMetadata (): Promise<Record<string, string>> {
```
lmiller1990, line 101: Trying to get some consistency by using `#` over `private`. I really think this I better for various reasons:

- `#` is a native JS; better than a proprietary feature
- `#` is actually private; `private` is just compile time private (I think)
- With `#`, you can see a method is private where it's called. With `private`, you need to look at the definition.

This file has both types, so I just the one I think is better and made them all the same.

packages/data-context/src/sources/VersionsDataSource.ts
```diff
@@ -29,16 +29,15 @@ export class VersionsDataSource {
   constructor (private ctx: DataContext) {
     this._initialLaunch = true
     this._currentTestingType = this.ctx.coreData.currentTestingType
-    this.#ensureData()
```
lmiller1990, line 32: If we call this when the data source is initialised (eg when the entire lifecycle starts) the config has not been read yet, and we cannot get the `config.component` key. So, we don't call this immediately anymore, it's called just-in-time by `versionData`.

packages/data-context/src/sources/VersionsDataSource.ts
mike-plummer: Today I learned I completely missed the fact that JS is gaining true private fields/methods. Cool!

Personally I've liked using Typescript's `private` in the past because I'm still able to use those items directly in unit tests which sometimes makes testing a bit cleaner/easier. But that's the lazy developer within me - if a class is hard to test using its public interface it's probably a hint that I haven't structured it very well

mike-plummer (commented):
LGTM!

Outside the code, do we need to update any docs regarding analytics/usage data we collect within privacy notices/usage agreements/etc?

mike-plummer (approved):

ZachJW34 (approved):

lmiller1990:
@mike-plummer 

> if a class is hard to test using its public interface it's probably a hint that I haven't structured it very well

This is my thoughts, too - I like it when languages force me to write good tests, even if writing good tests is not always fun😓.

I'm not sure grabbing framework version is that valuable - would we also grab the dev server version? I think npm downloads is a more reliable metric for this kind of thing. I'm thinking we only want to grab data the minimum data we need.

Definitely happy to continue this discussion internally around what we should grab, if we want to make changes we can create a new issue - trying to keep scope creep in mind, which is something we've traditionally not been great at.

packages/data-context/src/sources/VersionsDataSource.ts
brian-mann: TIL this also exists, very cool

PR #22284, opened 2022-06-13
feat: Auto-focus and select file name for create new spec modal

### LINKED ISSUE #21865: Auto-select filename text in Create Spec modal
## **Summary**
I wonder if it's possible to make it easier for the user to change the spec name in this dialog by a) focusing the input, b) selecting the filename, and c) using an explicit background color for the highlighted area, like [https://www.loom.com/share/778f2b74129849dc8df6be8e06252963](https://www.loom.com/share/778f2b74129849dc8df6be8e06252963|smart-link)

### Current

Screen Shot 2022-05-31 at 10 58 46 AM

### Desired

Highlighted this text and probably a nicer color than this orange. 

Screen Shot 2022-05-31 at 10 58 58 AM

[https://cypressio.slack.com/archives/C02MYBT9Y5S/p1650563604384479](https://cypressio.slack.com/archives/C02MYBT9Y5S/p1650563604384479|smart-card)

┆Issue is synchronized with this [Jira Task](https://cypress-io.atlassian.net/browse/UNIFY-1654) by [Unito](https://www.unito.io)
┆author: Ben Alman
┆epic: UI/UX Improvements
┆friendlyId: UNIFY-1654
┆priority: High
┆sprint: Backlog
┆taskType: Task

astone123:
- Closes #21865 

### User facing changelog

Opening the create spec modal will automatically focus the input and select the file name of the new spec

### Additional details

This change is an improvement to the UX for creating a new spec. Most of the time, the user will probably want to edit the file name to override the default. By auto-focusing the input and selecting the part of the file path that the user will most likely want to edit, we can speed up the spec creation process resulting in a more convenient UX. See related issue for more info.

### Steps to test

- Open the spec list for an application (e2e or ct)
- Click the "+ New Spec" button in the top-right corner
- Verify that when the modal opens, the text input is focused and the file name in the path is selected for you to edit

**After**
image

### PR Tasks


packages/app/src/specs/CreateSpecModal.cy.tsx
```diff
@@ -44,6 +44,17 @@ describe('<CreateSpecModal />', () => {
     cy.get(modalSelector).should('be.visible')
 
     cy.percySnapshot()
+  }),
```
ZachJW34, line 47: Remove comma and add space

packages/app/src/specs/generators/EmptyGenerator.vue
```diff
@@ -190,6 +195,25 @@ const hasError = computed(() => !isValidSpecFile.value && !!specFile.value)
...
 
+function getFileNameIndexes (inputValue: string, match: RegExpMatchArray): number[] {
+  return [match.index || 0, inputValue.indexOf(match[0]) + match[0].length]
+}
+
+onMounted(() => {
+  // Focus text input and try to pre-select file name
+  inputRef.value?.focus()
+
+  const fileNameRegex = /[ \w-]+(?=\.)/
```
ZachJW34, line 206: Can we get a comment here? My regex skills are sub-par so a comment would be helpful.

ZachJW34 (approved):
Nice test! Few small things that I'd like to see addressed bu going to ✅

packages/app/src/specs/generators/EmptyGenerator.vue
ZachJW34: I dig it!

mike-plummer (commented):
I know this isn't part of your issue, but I personally couldn't even tell there *was* any highlighted text until I really squinted my eyes. We may want to revisit the highlight color of these text boxes with UX to make sure we have sufficient contrast

packages/app/src/specs/generators/EmptyGenerator.vue
```diff
@@ -190,6 +195,27 @@ const hasError = computed(() => !isValidSpecFile.value && !!specFile.value)
...
+  inputRef.value?.focus()
+
+  // This regex matches a group of word characters, spaces, and '-' characters before a '.' character.
+  // For example 'path/to/spec/spec-name.cy.tsx' ---> 'spec-name'
+  const fileNameRegex = /[ \w-]+(?=\.)/
+  const inputValue = inputRef.value?.value
+  const match = inputValue?.match(fileNameRegex)
+
+  if (inputValue && match) {
+    const [startSelectionIndex, endSelectionIndex] = getFileNameIndexes(inputValue, match)
```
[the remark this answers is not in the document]
astone123: Yeah I think I used a function there in an effort to make things more readable, but looking at it now it doesn't make much sense

packages/app/src/specs/generators/EmptyGenerator.vue
```diff
@@ -190,6 +195,24 @@ const hasError = computed(() => !isValidSpecFile.value && !!specFile.value)
...
+
+  // This regex matches a group of word characters, spaces, and '-' characters before a '.' character.
+  // For example 'path/to/spec/spec-name.cy.tsx' ---> 'spec-name'
+  const fileNameRegex = /[ \w-]+(?=\.)/
+  const inputValue = inputRef.value?.value
+  const match = inputValue?.match(fileNameRegex)
+
+  if (inputValue && match) {
+    const startSelectionIndex = match.index || 0
+    const endSelectionIndex = inputValue.indexOf(match[0]) + match[0].length
```
mike-plummer, line 210: I think you're still doing a bit of extra work here - couldn't this `inputValue.indexOf(match[0])` be replaced by `startSelectionIndex`?

astone123: yeah you're right, it could be 👍🏻

mapsandapps:
> @mapsandapps Any thoughts on this? [#22284 (review)](https://github.com/cypress-io/cypress/pull/22284#pullrequestreview-1004829818)
> 
> We had talked earlier about maybe changing the highlight color. I'm guessing we'd want this to be consistent across the entire app though, so it might be worth creating a separate issue for this

I checked with Ryan, and it should be indigo-100

packages/app/src/specs/generators/EmptyGenerator.vue
```diff
@@ -190,6 +195,24 @@ const hasError = computed(() => !isValidSpecFile.value && !!specFile.value)
 
 const result = ref<GeneratorSuccessFileFragment | null>(null)
 
+onMounted(() => {
+  // Focus text input and try to pre-select file name
+  inputRef.value?.focus()
+
+  // This regex matches a group of word characters, spaces, and '-' characters before a '.' character.
```
lmiller1990, line 202: As a nit/discussion item but not a hard blocker, I want to get your opinion on how we should be testing this kind of code, specifically the "business logic" part, which is parsing the spec.

Eg, I wonder if a pure function like this would be better:

```js
const fileNameRegex = /[ \w-]+(?=\.)/

export function getHighlightRange (filename: string) {
  const match = inputValue?.match(fileNameRegex)

  if (!match) {
    return null
  }

  const startSelectionIndex = match.index || 0
  const endSelectionIndex = startSelectionIndex + match[0].length
   
  return [startSelectionIndex, endSelectionIndex]
}
```

It would be easy to test in isolation (no need for the UI framework to come into play), and against lots of different strings - right now, we only test it against one string.

Finally, you'd the import this into the component, and your `onMount` would be

```js
onMounted(() => {
  if (!inputRef?.value) {
    return
  }

  const match = getSelectionRange(specFile)
  
  if (!match) {
    return
  }

  inputRef.value.setSelectionRange(match[0], match[1])
})
```

More a point of discussion around best practices and testing philosophy in general, thoughts? It's a bit more overhead, but I like the separation of concerns.

lmiller1990 (commented):
Left some comments/questions

packages/app/src/specs/generators/EmptyGenerator.vue
astone123: Yeah this is a good discussion to have. I like the idea of having at least one component test that tests the "happy path" just so that we are sure that in most cases the component is behaving as expected. Having at least one test that covers everything together is crucial imo. But I also have this same thought

> I'm not sure the ideal of re-mounting a component many times for lots of regexp test is ideal.

One thing I'd question about that though: is the regex the _only_ thing we'd be testing? I'd argue that it's not. What's important is the experience for the user for different situations (i.e. when the default file path that we provide doesn't match the regex) which I think we can get better from a CT.

So at this point my thought is to add more component test cases that provide different default strings and see how the component behaves. That's probably what I'll do to finish out this PR, but I'm not sure if I love that approach because it doesn't seem very efficient.

I also like the idea of separating the logic out into its own function like `getSelectionRange` for the sake of cleaning things up a bit if nothing else. As far as using CT vs a normal unit test for something like this, I think I'm torn. Would like to hear what the rest of the team thinks

astone123:
Planning on addressing the selection highlight color in this issue
https://github.com/cypress-io/cypress/issues/22322

tbiethman (approved):

packages/app/src/specs/generators/EmptyGenerator.vue
lmiller1990: Good points! No right or wrong here.

For now, happy with the unhappy case you've added: https://github.com/cypress-io/cypress/pull/22284/commits/50851bd064854d9f42ff87c2c26e782a2b678c5c

✅

PR #22292, opened 2022-06-13
chore: prep npm packages for use with Cypress v10

ZachJW34:
### User facing changelog
na

### Additional details
[npm-release](https://app.circleci.com/pipelines/github/cypress-io/cypress/39434/workflows/5ea799d1-79dc-4326-8f0b-8fc80caff111/jobs/1611883) is timing out on `@cypress/webpack-dev-server`. It is failing with 
```
The authenticity of host 'github.com (140.82.112.3)' can't be established.`
```
The package.json for `@cypress/webpack-dev-server` is missing some required fields. It used to have these but we dropped them when doing the overhaul. I tested this change by ssh-ing into the npm-release job and verifying that it was able to move past this timeout.

I also removed the configured registry for `@cypress/vue2` as we have had issues in the past releasing a pakcage configured with the `http://` since it's deprecated in favor of `https://` which is the default value.

### Steps to test
Job shows what's going on, I was only able to reproduce it on the CI machine

na

### PR Tasks

BREAKING CHANGE: new version of packages for Cypress v10

### DIFF
--- a/npm/webpack-dev-server/package.json
+++ b/npm/webpack-dev-server/package.json
@@ -44,5 +44,15 @@
   },
   "files": [
     "dist"
-  ]
+  ],
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/cypress-io/cypress.git"
+  },
+  "homepage": "https://github.com/cypress-io/cypress/tree/master/npm/webpack-dev-server#readme",
+  "bugs": "https://github.com/cypress-io/cypress/issues/new?template=1-bug-report.md",
+  "publishConfig": {
+    "access": "public"
+  }
 }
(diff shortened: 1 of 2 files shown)


npm/webpack-dev-server/package.json
```diff
@@ -44,5 +44,15 @@
   },
   "files": [
     "dist"
-  ]
+  ],
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/cypress-io/cypress.git"
```
flotwig, line 51: oh, @ryanthemanuel i bet this is why we were getting "authenticity cannot be established" - we were using git+ssh, and having `repository.url` set can force npm to use `https` instead, avoiding `known_hosts` verification. kudos @ZachJW34 for actually tracking this down

flotwig: although i don't think you were touching these files when you first ran into it :confused:

ZachJW34: Glad I could (unknowingly) help 😅

flotwig (approved):

mjhenkes (approved):

PR #22325, opened 2022-06-15
chore: make spec migration screen text more generic

### LINKED ISSUE #21926: Update wording of 'rename specs' area when the specs are not actually being renamed
## **Summary**
I know this step is for renaming and moving but it feels odd to have everything say ‘rename’ when in this case there isn’t a rename and just a move:

Screen Shot 2022-05-19 at 2 07 18 PM

[https://cypressio.slack.com/archives/C02MYBT9Y5S/p1652990922196199](https://cypressio.slack.com/archives/C02MYBT9Y5S/p1652990922196199|smart-card)

**Acceptance Criteria**

If we are not renaming the spec extension (like for an existing Dashboard project)

1. Have heading say ‘Move existing specs’
1. Have sub description say ‘In this step, we’ll automatically move your existing spec files.
1. Remove the sub-heading in the content that says ‘We recommend automatically…..etc'

┆Issue is synchronized with this [Jira Task](https://cypress-io.atlassian.net/browse/UNIFY-1801) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/12035">Screen Shot 2022-05-19 at 2.07.18 PM.png</a>
┆author: Matthew Schile
┆epic: UI/UX Improvements
┆friendlyId: UNIFY-1801
┆priority: Medium
┆sprint: Fast Follows 1
┆taskType: Task

lmiller1990:
- Closes #21926

### User facing changelog

N/A

### Additional details

Make the text more generic, as to cover all the cases where we automatically change specs:

- rename
- move
- rename and move

### Steps to test

Just open launchpad to any Cypress 9 project and have a look.

Improved language in migration UI to be more clear about intended and recommended behavior.

## Before

image

## After

Screen Shot 2022-06-16 at 12 14 58 am

### PR Tasks

- [na] Have tests been added/updated? **It's just a json file, so no, there isn't any to update**
- [x] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only) **YES** 10.1.1

### DIFF
--- a/packages/frontend-shared/src/locales/en-US.json
+++ b/packages/frontend-shared/src/locales/en-US.json
@@ -543,7 +543,7 @@
     "after": "After",
     "heresWhy": "here's why:",
     "renameAuto": {
-      "title": "We recommend automatically updating your specs in this step",
+      "title": "We recommend automatically renaming your specs in this step",
       "changeButton": "change",
       "changedSpecFolder": "We've changed the default spec folder from:",
       "changedSpecExt": "We've changed the default spec file extension from:",
@@ -597,8 +597,8 @@
       "title": "Migration Helper",
       "description": "Complete the steps below to migrate your project to Cypress 10",
       "step1": {
-        "title": "Rename existing specs",
-        "description": "In this step, we'll automatically change existing spec files to match the new convention.",
+        "title": "Migrate existing specs",
+        "description": "In this step, we'll automatically rename and/or move your existing spec files as needed.",
         "button": "Rename these specs for me",
         "buttonSkip": "Skip renaming specs",
         "buttonRenameFolder": "Rename the folder for me"
--- a/packages/frontend-shared/src/locales/en-US.json
+++ b/packages/frontend-shared/src/locales/en-US.json
@@ -543,7 +543,7 @@
     "after": "After",
     "heresWhy": "here's why:",
     "renameAuto": {
-      "title": "We recommend automatically renaming your specs in this step",
+      "title": "We recommend automatically updating your specs in this step",
       "changeButton": "change",
       "changedSpecFolder": "We've changed the default spec folder from:",
       "changedSpecExt": "We've changed the default spec file extension from:",
@@ -598,7 +598,7 @@
       "description": "Complete the steps below to migrate your project to Cypress 10",
       "step1": {
         "title": "Rename existing specs",
-        "description": "In this step, we'll automatically rename and move your existing spec files.",
+        "description": "In this step, we'll automatically change existing spec files to match the new convention.",
         "button": "Rename these specs for me",
         "buttonSkip": "Skip renaming specs",
         "buttonRenameFolder": "Rename the folder for me"


packages/frontend-shared/src/locales/en-US.json
```diff
@@ -543,7 +543,7 @@
     "after": "After",
     "heresWhy": "here's why:",
     "renameAuto": {
-      "title": "We recommend automatically renaming your specs in this step",
+      "title": "We recommend automatically updating your specs in this step",
```
marktnoonan, line 546: "updating" sounds to broad to me, what we _recommend_ is definitely renaming. This title probably doesn't need a change.

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -598,7 +598,7 @@
       "description": "Complete the steps below to migrate your project to Cypress 10",
       "step1": {
         "title": "Rename existing specs",
-        "description": "In this step, we'll automatically rename and move your existing spec files.",
+        "description": "In this step, we'll automatically change existing spec files to match the new convention.",
```
marktnoonan, line 601: This also seems too broad compared to what we have now. How about:

> In this step, we'll automatically rename and/or move your existing spec files as needed.

Which keeps the scope narrow to what we will actually do, and allows it to be correct even if we are only doing one thing or the other. We could leave all other text the same I think.

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -598,7 +598,7 @@
       "description": "Complete the steps below to migrate your project to Cypress 10",
       "step1": {
         "title": "Rename existing specs",
```
marktnoonan, line 600: In the screenshot, this title is changed, but it's not here?

marktnoonan (changes requested):
"Updating specs" describes something potentially much broader than renaming and moving. I've suggested a smaller change that might still be generic enough.

packages/frontend-shared/src/locales/en-US.json
lmiller1990: I actually had some variation of this but ended up flip flopping back to "update", but happy to go with this.

packages/frontend-shared/src/locales/en-US.json
lmiller1990: Oh, let's use "Migrate existing specs" for the title. It's the perfect description for what we are doing, it's the migration UI, after all.

packages/frontend-shared/src/locales/en-US.json
lmiller1990: Good point, reverted

marktnoonan (approved):
I'm good with this, migrate is indeed a good title imo.

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -597,8 +597,8 @@
       "title": "Migration Helper",
       "description": "Complete the steps below to migrate your project to Cypress 10",
       "step1": {
-        "title": "Rename existing specs",
-        "description": "In this step, we'll automatically rename and move your existing spec files.",
+        "title": "Migrate existing specs",
+        "description": "In this step, we'll automatically rename and/or move your existing spec files as needed.",
```
rachelruderman, line 601: Simpler version, up to you! 
```suggestion
        "description": "In this step, we'll automatically update your existing spec files.",
```

rachelruderman: Hahaha just caught up on the whole thread on wording between you and @marktnoonan . My 2 cents is that `rename and/or move` is a bit wordy and makes me think, whereas "update" covers both and I don't have to cerebrally digest what's happening. It's a simple thing we're doing, imo the markup should be short & sweet

marktnoonan: When I read "update" I feel like like it could do absolutely anything to my files. It's not clear that the changes are limited to moving/renaming, that in this step we aren't going to actually touch the internals of files (like we do other places).

rachelruderman: What if it's "update your existing spec file paths" instead of "update your existing spec files"?

fwiw I'm not married to either approach and will support whatever you choose 😃

rachelruderman (approved):

packages/frontend-shared/src/locales/en-US.json
lmiller1990: I'm happy with either, just gonna ship this as is. I think they are both fine and combined with the UI, it's pretty clear what's going on.

PR #22334, opened 2022-06-15
fix: add tests for React 18, bump peerDeps

### LINKED ISSUE #21381: Cypress Testing fails to build on React version 18
### Current behavior

Hey guys, we've encountered an error that is disrupting our build pipeline :( (oh no!)

Currently Cypress only allows react 16 in the package.json. These days some other packages require at least React 18, making this Cypress tester incompatible with those packages. When installing this with pnpm, or another package manager, warnings about incompatible react versions cause builds to fail, disrupting the build pipeline.

PNPM:
```
 WARN  Issues with peer dependencies found
.
└─┬ @cypress/react
  ├── ✕ unmet peer @types/react@"^16.9.16 || ^17.0.0": found 18.0.8
  ├── ✕ unmet peer react@"^=16.x || ^=17.x": found 18.1.0
  └── ✕ unmet peer react-dom@"^=16.x || ^=17.x": found 18.1.0
```

We were able to get past this by ignoring all warnings, and errors, but obviously this is not
something we want to do.

Turns out, from our testing it a bit, Cypress seems to work just fine with React 18 (after ignoring the errors and just running everything manually). But we didn't have time to test it extensively yet.

From doing some research it seems like the biggest bottleneck comes from the Enzyme package being limited at React 16. And this package is not getting many updates these days.

### Desired behavior

We should be able to run Cypress on react 18.

We should look into migrating away from Enzyme, if possible. Potentially to rely on React Testing Library.

### Test code to reproduce

Following this official getting-started guide produces the failure state.

https://www.cypress.io/blog/2021/04/06/cypress-component-testing-react/

Using pnpm, and installing the missing peer dependencies, you will reach a request saying

```
 WARN  Issues with peer dependencies found
.
└─┬ @cypress/react
  ├── ✕ unmet peer @types/react@"^16.9.16 || ^17.0.0": found 18.0.8
  ├── ✕ unmet peer react@"^=16.x || ^=17.x": found 18.1.0
  └── ✕ unmet peer react-dom@"^=16.x || ^=17.x": found 18.1.0
```

If you downgrade react to react 16, then other packages start failing. It becomes difficult to find a golden package configuration that meets everything.

### Cypress Version

 Cypress package version: 9.6.0

### Other

Cypress package version: 9.6.0
Cypress binary version: 9.6.0
Electron version: 15.5.1
Bundled Node version: 16.5.0

### LINKED ISSUE #21946: React 18 support - today logs a bunch of warnings for ReactDom.render being deprecated
## **Current Behavior**
Thought we had an issue for React-18 support but couldn't find one. `ReactDOM.render` is deprecated and logs a bunch of warnings

Screen Shot 2022-04-21 at 12 43 59 PM

┆Issue is synchronized with this [Jira Bug](https://cypress-io.atlassian.net/browse/UNIFY-1639) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/11846">Screen Shot 2022-04-21 at 12.43.59 PM.png</a>
┆author: Zach Williams
┆epic: CT Architecture
┆friendlyId: UNIFY-1639
┆priority: Medium
┆sprint: Backlog
┆taskType: Bug

lmiller1990:

### User facing changelog

Add support for React 18 to `npm/react`.

### Additional details

- updated `peerDependencies` to allow installing `@cypres/react` with `react@18`
-  some specific code in vite and webpack dev servers to account for the potentially not-available `react-dom/client` code

Since React projects on <= 17 do not have `react-dom/client`, I did a dynamic import:

```js
if (react!8) {
  import('react-dom/client').then(/* mount component */).catch(...)
}
```

This is valid in regular ESM, but both webpack and vite are NOT regular ESM... Vite does pre-optimizating, and webpack has it's own `require` function.

My solution:
1. vite plugin to ignore the missing `react-dom/client` request if the user is not on react 18
2. add `react-dom/client` as `externals` for React <= 17 projects. Then we can catch the error, instead of webpack failing to compile due to a missing module.

I cannot get the system-tests to work with React 18, but I did build the binary and test it out - seems like it's working as expected. 

Many other libs have this problem, eg storybook who solves this with `IngorePlugin` for webpack: https://github.com/storybookjs/storybook/pull/17215/files#diff-bc92d51dd98fe78ac7c9251c1b8d10d45bdfc8b872be68c12ace18f783205fd7R16-R18.

Another solution would be **remove** the react-dom/client dependency, maybe with https://webpack.js.org/plugins/normal-module-replacement-plugin/, but I cannot get this to work, either. I think the problem is that `cypress/react` is NOT a module, but just some inline code we bundle in `node_modules/cypress`, but I'm not really sure.

Either way, I think we need React 18 support soon, and this seems to be the best solution, other than creating `cypress/react18`, which is not really ideal.

Another solution might be having `cypress/react18` in the binary, and dynamically replacing `cypress/react` with a webpack or Vite plugin under the hood. cc @JessicaSachs, thoughts?

### PR Tasks


npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -155,6 +157,21 @@ export async function makeWebpackConfig (
     ],
   }
 
+  try {
```
lmiller1990, line 160: This is a bit of a hack. Since we cannot assume `react-dom/client` exists, we import it using a dynamic import: https://github.com/cypress-io/cypress/pull/22334/files#diff-deaf32d610fd1088b9d1a39d5b49198b98fcaba2d2b1b8942df73ed9961551e0R142

```js
import('does-not-exist').then(...).catch(...)
```

This works fine for "real" ESM, but webpack is not real ESM, but it's own implementation - it does analysis of all modules, including dynamic imports, and will error.

To tell webpack NOT to do this, we can just add it as an external - then webpack will assume it exists at runtime. We can now catch the error!

Unfortunately, this does not seem to work how you'd expect during tests - possibly due to how we use symlinks, or the fact that we ship our module is a kind of weird way (inline inside of `node_modules/cypress`.

Storybook solves this differently, using `IgnorePlugin`: https://github.com/storybookjs/storybook/pull/17215/files#diff-bc92d51dd98fe78ac7c9251c1b8d10d45bdfc8b872be68c12ace18f783205fd7R16-R18

But I could not get this to work, I am not sure why - possibly due to the weird way we bundle `cypress/react`.

lmiller1990:
In favor of https://github.com/cypress-io/cypress/pull/22437 for now.

PR #22347, opened 2022-06-15
fix: Preserve tree expansion state 

mike-plummer:
Video: https://cypressio.slack.com/archives/C035VB3HVAL/p1655219943591069

Addressing a reactivity issue we currently have in our `useCollapsibleTree` usage. In ACI, row data is loaded as the user scrolls and periodically in the background - these fetches flow back through GQL and present as a "new" list of specs (even though it's the same specs, just with new deeply-nested content) which triggers the tree data to get rebuilt. This manifests as the tree state re-setting on scroll and is likely contributing to some of the scroll lagginess we're seeing.

Without reworking the implementation which would impact multiple UI areas at present this is a next-best: incorporate an external cache that can be used to persist the expansion state of the directories so that when the tree does get rebuilt the state is maintained. This cache is then cleared when the user searches or changes the list of specs because we want the directories to expand in those situations to match 10.1.0 behavior.

### Additional details

I wasn't able to find a good way to test this in CT; since SpecList is using `mountFragment` and GQL I couldn't find a way to mutate the GQL data and trigger a reactive data update within a single mount cycle. If someone knows how to do this I'd love to learn how

### Steps to test

1. Open E2E mode with a project that has at least one recorded run in the cloud (preferably one with 
2. Log in if not already
3. Observe latest run data begin to load
4. Collapse a directory and scroll down
5. Observe directory remains collapsed as lower rows scroll into viewport and load cloud data
	- Current behavior would result in directory resetting to expanded

SpecsList directory expansion state should persist as ACI data is loaded during scroll

### PR Tasks


MuazOthman (approved):
Looks good and functional. I defer to @marktnoonan to advise in terms of performance and memory use, but it seems to have minimal memory footprint using a sparse map.

lmiller1990 (approved):
Nice fix with minimal changes. The virtual list feels kind of bleh, I think we need to rethink this approach. If we do away with highlighting characters during fuzzy search I think we can vastly cut down on the number of DOM nodes and avoid the need virtualize it, but another battle for another day.

marktnoonan (approved):
Forgot to drop this approval before I logged off yesterday, but this looks good to me. The old behavior was really noticeable, especially when even collapsing a row would cause new items to render, re-expanding the row you just collapsed. Good solution.

PR #22356, opened 2022-06-16
chore: Address skipped specs in server package

### LINKED ISSUE #21966: Audit and address spec skips in server package related to unification dev
## **Summary**
There are a large number of specs skipped in the server package related to 10.0 development. These tests should be removed if they’re no longer necessary, or corrected if they’re providing necessary coverage. 

## **Acceptance Criteria**
* Skipped tests that are no longer needed should be removed
* Skipped tests for workflows that do not otherwise have coverage should be updated
* No skipped tests should be left without a documented reason for their current state and an associated issue

### **Resources**
Any Notion documents, Google documents, Figma Boards

### **Open Pull Requests**
Any PRs needed for review

┆Issue is synchronized with this [Jira Task](https://cypress-io.atlassian.net/browse/UNIFY-1393) by [Unito](https://www.unito.io)
┆author: Tyler Biethman
┆epic: Testing / Quality
┆friendlyId: UNIFY-1393
┆priority: Medium
┆sprint: Fast Follows 1
┆taskType: Task

astone123:
- Closes #21966

### User facing changelog

n/a

### Additional details

Some test suites were skipped during the unification effort. We need to re-visit those skipped tests and either revise them to cover new workflows, get coverage elsewhere and/or remove them.

I marked the diff with explanations for the changes that I made. Look through the changes and make sure that you agree with what changes were made to the tests.

I revised and/or deleted most single tests that were skipped, but there were some entire test files that were skipped and I think that we should handle those in separate tickets. You can see all of the related tickets organized in [this epic](https://github.com/cypress-io/cypress/issues/22494)

### PR Tasks


packages/server/test/integration/cypress_spec.js
```diff
@@ -1731,26 +1702,6 @@ describe('lib/cypress', () => {
       })
     })
 
-    // NOTE: skipped because we want to ensure this is captured in v10
```
astone123, line 1734: This workflow is covered by a system test `system-tests/test/network_error_handling_spec.js`

packages/server/test/unit/config_spec.js
```diff
@@ -578,27 +570,27 @@ describe('lib/config', () => {
       })
 
       // TODO:(lachlan): after mega PR
-      context.skip('specPattern', () => {
+      context('specPattern', () => {
```
astone123, line 573: Just want to point out that fixing this test uncovered that we aren't validating the `specPattern` property in our latest version. This would be an issue if a user had a `specPattern` in their config that wasn't either a string or an array of strings. For example, if I had 
```
specPattern: 5
```
in my configuration, I would see this error
Screen Shot 2022-06-17 at 12 27 02 PM

now, I would see this error

Screen Shot 2022-06-17 at 12 25 47 PM

packages/config/src/options.ts
```diff
@@ -419,6 +419,11 @@ const resolvedOptions: Array<ResolvedConfigOption> = [
     canUpdateDuringTestTime: false,
     requireRestartOnChange: 'server',
   },
+  {
+    name: 'specPattern',
```
marktnoonan, line 423: This is a good catch! We seem to also declare the default values in other cases.  Looks like we can use the same `options.testingType` ternary that's being used for `viewportWidth` etc to populate the defaults that are declared in this file already in the `defaultSpecPattern` object. What do you think?

marktnoonan (changes requested):
Some good changes here! Requesting one change in the validation, to include the default `specPattern` values. I might have one or two other questions tomorrow.

For the tests you mentioned in the description that aren't addressed in this PR, let's make those issues and link them up before merging the PR. It might be good to reach out to @tgriesser and get some notes on what is needed to fix them, or confirm if, say, [these unit tests](https://github.com/cypress-io/cypress/blob/110f8ae02dd2b6fd695856aaa77e3b9d96c939dc/packages/server/test/unit/util/settings_spec.js#L13%7Csmart-link) are still needed, given that the comment says this is tested through open mode now.

packages/server/test/unit/config_spec.js
lmiller1990: Lol, as soon as we skipped a test, the functionality regressed and no-one noticed. Obvious lesson is don't skip tests!!!!! 🤦

packages/config/src/options.ts
```diff
@@ -419,6 +419,12 @@ const resolvedOptions: Array<ResolvedConfigOption> = [
...
+    defaultValue: (options: Record<string, any> = {}) => options.testingType === 'component' ? defaultSpecPattern.e2e : undefined,
+    validation: validate.isStringOrArrayOfStrings,
+    canUpdateDuringTestTime: false,
+  },
```
lmiller1990, line 427: Am I missing something, this seems the opposite of what we want:

```js
options.testingType === 'component' ? defaultSpecPattern.e2e
```

If we are using component testing, default to the e2e spec pattern?

I think you want these: https://github.com/cypress-io/cypress/blob/50cc582543f714fd363eb785194bd1de638dd48d/packages/config/src/options.ts#L109-L112

Maybe it should be

```js
defaultValue: (options: Record<string, any> = {}) =>defaultSpecPattern[options.testingType],
```

packages/server/test/integration/cypress_spec.js
lmiller1990: We also have it here for launchpad:

https://github.com/cypress-io/cypress/blob/develop/packages/launchpad/cypress/e2e/config-files-error-handling.cy.ts#L242-L271

packages/server/test/unit/config_spec.js
```diff
@@ -1525,7 +1518,6 @@ describe('lib/config', () => {
             supportFile: { value: false, from: 'config' },
             supportFolder: { value: false, from: 'default' },
             taskTimeout: { value: 60000, from: 'default' },
-            specPattern: { value: '**/*.*', from: 'default' },
```
lmiller1990, line 1528: Why did this get removed?

packages/server/test/unit/config_spec.js
```diff
@@ -1569,23 +1561,25 @@ describe('lib/config', () => {
 
         return config.mergeDefaults(obj, options)
         .then((cfg) => {
+          ['platform', 'arch'].forEach((x) => {
+            expect(cfg.resolved[x]).to.exist
```
lmiller1990, line 1565: I just realised we could potentially do

```js
expect(cfg.resolved.arch).to.eq({ value: os.arch(), from: 'default' })
```

Or something - thoughts? We'd still delete after, so the snapshot is system agnostic.

lmiller1990 (commented):
Some comments

lmiller1990:
As for adding missing coverage, I'd agree we make new tickets for each one (or maybe can group some). Would you like to to that? You could then reference this one in their description.

I think
- plugins https://github.com/cypress-io/cypress/blob/efede20d9ed1222139a68c4b6002d59abd872e8f/packages/server/test/unit/plugins/child/run_plugins_spec.js
- run spec https://github.com/cypress-io/cypress/blob/f669c101e13a40722b53e9556792b54dae9abbf2/packages/server/test/unit/modes/run_spec.js

Are probably the more important ones (and likely can be replaced by a handful of integration tests, instead of 50+ unit tests).

packages/server/test/unit/config_spec.js
```diff
@@ -1634,7 +1628,6 @@ describe('lib/config', () => {
             supportFile: { value: false, from: 'config' },
             supportFolder: { value: false, from: 'default' },
             taskTimeout: { value: 60000, from: 'default' },
-            specPattern: { value: '**/*.*', from: 'default' },
```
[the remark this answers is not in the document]
astone123: Yeah I believe so

packages/config/src/options.ts
astone123: Yeah that was wrong... I ended up removing the defaultValue because it was causing issues. It seems like `options.testingType` was undefined sometimes when inside of the `defaultValue` function. I don't think that we need that property here based on the tests

packages/config/src/options.ts
astone123: I added that but then I just removed it again... I'm confused about what `defaultValue` does here. It looks like it was messing up some of the validation in the tests when I had it in there 😕

packages/server/test/unit/gui/events_spec.js
```diff
@@ -1,191 +0,0 @@
...
-const EE = require('events')
-const extension = require('@packages/extension')
-const electron = require('electron')
-const Promise = require('bluebird')
-const chromePolicyCheck = require(`../../../lib/util/chrome_policy_check`)
-const ProjectBase = require(`../../../lib/project-base`).ProjectBase
-const errors = require(`../../../lib/errors`)
-const browsers = require(`../../../lib/browsers`)
-const { openProject } = require('../../../lib/open_project')
-const events = require(`../../../lib/gui/events`)
```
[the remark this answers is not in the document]
marktnoonan: Let's make a ticket to clean those out, iirc this is stuff @tgriesser kept around for reference purposes during a transition. Doesn't quite make sense to remove them in this PR for sure. But maybe Tim can tell us if it's safe

packages/config/src/options.ts
marktnoonan: Let's make an issue and add a comment for this then, totally worth investigate.

lmiller1990 (approved):
Looks good, my changes have been incorporated, or spawned new issues.

packages/config/src/options.ts
astone123: Added an issue and a comment here

packages/server/test/unit/config_spec.js
```diff
@@ -578,27 +572,27 @@ describe('lib/config', () => {
       })
 
       // TODO:(lachlan): after mega PR
```
tbiethman, line 574: The TODO has been addressed, should be safe to remove at this point.

marktnoonan (approved):

lmiller1990:
I do no know what's going on with the failing windows test, but it's happening in multiple branches, and I do not think it was introduced here. I am going to force merge this, then look into the windows CI fail.

PR #22363, opened 2022-06-16
fix: process_profiler follow up work for v10

### LINKED ISSUE #22022: process_profiler is broken in 10.0.0
### Current behavior

process_profiler is broken: https://docs.cypress.io/guides/references/troubleshooting#Log-memory-and-CPU-usage

1. The namespace was changed to verbose in the 10.0 merge, which broke the documented behavior.
2. "plugins" group detection is broken - it was probably not updated for 10.0. There actually seems to be a lot of the old plugin code still in here that should be removed.
3. `procs.params` can be undefined which causes it to fail
4. No full e2e tests

### Desired behavior

1. Change the _printGroupedProcesses namespace back to `debug` from `debugVerbose`
2. Fix plugin PID detection so the grouping works.
3. Check proc.params for truthiness before using `.includes`
4. Add better e2e tests

### Test code to reproduce

--

### Cypress Version

10.0.0

### Other

_No response_

tbiethman:
- Closes #22022

### User facing changelog

* Updated the process profiler to output Cypress's summarized CPU and memory metrics when the 'cypress:server:util:process_profiler' debug stream is enabled.

### Additional details

The plugin group had stopped working due to some of the called APIs in the ProjectLifecycleManager being removed. I added a few APIs to enable the profiler to once again access this process info and properly group it. 

### Steps to test

1. Launch cypress with `DEBUG=cypress:server:util:process_profiler`
2. Open a project (`launchpad`), select testing type
3. View process profiler output, ensure 'plugin' group is visible and populated

These changes align with our currently documented behavior.

### PR Tasks


packages/server/lib/plugins/index.ts
```diff
@@ -0,0 +1,28 @@
+import {
+  getCtx, registerServerPluginHandler,
+} from '@packages/data-context'
+
+export const registerEvent = (event, callback) => {
+  getCtx().lifecycleManager.registerEvent(event, callback)
+}
+
+export const getPluginPid = () => {
+  return getCtx().lifecycleManager.eventProcessPid
```
tbiethman, line 10: With type checking here, we shouldn't see similar regressions going forward.

packages/server/lib/util/process_profiler.ts
```diff
@@ -208,7 +208,7 @@ export const _aggregateGroups = (processes: Process[]) => {
 export const _printGroupedProcesses = (groupTotals) => {
   const consoleBuffer = concatStream((buf) => {
     // get rid of trailing newline
-    debugVerbose(String(buf).trim())
+    debug(String(buf).trim())
```
ryanthemanuel, line 211: Are we intentionally updating this to debug rather than debugVerbose?

tbiethman: Yes, we document this as being available with the non-verbose namespace: https://docs.cypress.io/guides/references/troubleshooting#Log-memory-and-CPU-usage

ryanthemanuel (approved):

flotwig (approved):
Awesome!!!

PR #22365, opened 2022-06-16
fix(cli): don't include workspaces.nohoist in public NPM package

### LINKED ISSUE #22147: nohoist config is ignored
### Current behavior

when installing the cypress package via yarn, i get a warning saying the nohoist config is ignored

### Debug logs

warning nohoist config is ignored in "cypress" because it is not a private package. If you think nohoist should be allowed in public packages, please submit an issue for your use case.

### Cypress Version

10.0.3

### Package Manager

yarn

### Operating system

Mac

### Other

_No response_

flotwig:
- Closes #22147

### User facing changelog

Fixed an issue causing a "nohoist config is ignored" warning when installing `cypress` with yarn.

### Steps to test

Install latest prerelease comment on this branch on Linux, note there is no warning.

### PR Tasks

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -27,8 +27,7 @@ mainBuildFilters: &mainBuildFilters
     branches:
       only:
         - develop
-        - 10.0-release
-        - linux-arm64
+        - issue-22147-nohoist
 
 # usually we don't build Mac app - it takes a long time
 # but sometimes we want to really confirm we are doing the right thing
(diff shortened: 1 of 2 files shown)


ryanthemanuel (commented):
Looks good. Is there a way to write a test for this change?

flotwig:
@ryanthemanuel I don't see an obvious way to test it since it's in the build step. I could add an assertion in buiild.js or post-build.js, but it would be overly unit testy imo.

I think this script should be updated in general, we shouldn't need to do all this wacky merging of 2 package.jsons, I think all of the should be all in one place as much as possible. So no `_.extend(json`. We'd be better off since changes to `cli/package.json` couldn't have this unintended impact.

ryanthemanuel:
> @ryanthemanuel I don't see an obvious way to test it since it's in the build step. I could add an assertion in buiild.js or post-build.js, but it would be overly unit testy imo.

I figured as such. Approved!

ryanthemanuel (approved):
Looks good!

ZachJW34 (approved):

PR #22384, opened 2022-06-17
fix: make virtual-list loading more efficient

ZachJW34:
### User facing changelog
Increase rendering speed of items in spec-list

### Additional details
When quickly scrolling the spec list, there is a flash of un-rendered content due to the the virtual list having to re-render so many elements. This only became an issue with the recent ACI work that added more components to each row.

This fix defers the rendering of many of the columns such that the row + filename is given priority, leading to a more responsive feel to the table. It doesn't fix the issue entirely as fast scrolls can still cause a white flash, but it does improve the experience.

This PR __DOES NOT__ fix the scroll jank on Chrome. That is an issue in [Chrome v102](https://github.com/petyosi/react-virtuoso/issues/675) that has been fixed in v103. The scroll behavior works find in Firefox + Electron

### Steps to test
Check out muaz/CLOUD-577-spec-list-display-latest-runs and choose a project with a decent amount of tests. Scroll through the list quickly (or use the scrollbar) to see flashes on run-rendered content

Before:

https://user-images.githubusercontent.com/25158820/174340939-a5043086-1ddf-40c0-aa60-88a6be58b145.mov

After:

https://user-images.githubusercontent.com/25158820/174340957-81ef1e3a-2108-4b98-9bc8-2a3e56dfb7a9.mov

### PR Tasks

### DIFF
--- a/packages/app/src/specs/SpecsListRowItem.vue
+++ b/packages/app/src/specs/SpecsListRowItem.vue
@@ -2,7 +2,7 @@
   <div data-cy="specs-list-row">
     <component
       :is="isLeaf ? 'RouterLink' : 'div'"
-      class="h-full outline-none border-gray-50 pr-20px ring-inset grid grid-cols-7 md:grid-cols-9 group focus:outline-transparent focus-within:ring-indigo-300 focus-within:ring-1 children:cursor-pointer"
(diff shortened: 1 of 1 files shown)


astone123 (approved):
Looks like a good improvement 👍🏻

tbiethman (approved):

ZachJW34:
Going to merge, there are some flaky windows builds but I noticed these on the destination branch as well.

PR #22387, opened 2022-06-17
chore: clean up git ignore

emilyrohrbough:
Clean up git ignore. 
Noticed cypress/videos weren't ignored for packages/reporter and assumed the root-level git ignore covered this.

### User facing changelog

n/a 

dev expereince - all ignored files are in one place

### PR Tasks


.gitignore
```diff
@@ -21,9 +25,11 @@ packages/data-context/src/**/*.js
...
 packages/errors/test/**/*.js
 
-# from driver
-packages/driver/cypress/videos
-packages/driver/cypress/screenshots
+# from graphql, compiled .js files
+/packages/graphql/src/**/*.js
+
+# from https-proxy
+packages/https-proxy/ca/
```
flotwig, line 32: what is this again? for local tests?

flotwig (commented):
III actually like having `.gitignore`s for the specific packages, what's the reason to move everything to the top level?

Having stuff like `cypress/videos` and `node_modules` in root gitignore makes sense, but I feel like things such as `packages/https-proxy/ca` and `packages/types/src/**/*.js` belong alongside the packages and the associated code that is generating those files.

marktnoonan:
Since we already have a lot of package-specific stuff in the global `.gitignore`, it feels like the ones that were moved are actually exceptions currently, and moving them brings things into conformance with the current package.

I'm not too concerned with exactly we do here. I do have a slight preference for "one git repo, one `.gitignore`". Happy to let some more discussion happen though.

emilyrohrbough:
@flotwig I made this change for consistency in the repository and to declutter / delete some of the files we need to manage.  Personally, I like one level because it's easy to find all of the information for the repo in one place and declutters the repo. That being said, don't feel super strongly one way or the other -- I just want consistency.

lmiller1990:
I don't mind one way or another, but if we are moving to "one gitingore to rule them all" I think we tear of the bandaid and do it **everywhere** including `npm` and `system-tests/projects` (some of those have gitignores, for whatever reason). Eg - if we just have a single `.gitignore` in the entire repo, that makes it much easier to search.

If we want to go with the "one gitignore per package", the top level gitignore should be pretty much empty. 

Let's make a decision and just go for it, I'm happy with one repo === one gitignore, maybe we have a block comment for each package-specific ignore in the main one, in this case.

emilyrohrbough:
closing as stale. will revisit at another time

Issue #22263, opened 2022-06-11
Cypress 10.1.0 "Spec" UI not remembering settings


DobQA:
### Current behavior

Firstly after upgrading to Cypress 10.0 (macOS) Im little surprised at how the UI feels very unpolished. The UI and navigation should have been tested little more. There is a seperate thread on number of clicks required, for myself the biggest annoyance which I feel is a bug;

When opening my "Specs" list, it has multiple folders and .js files. 
I often use the "Search" or "Collapse" features within the "Specs" list.

1- I run the test I'm intrested in.
2- After the test finishes, on the left hand side I click "Specs" button
3- My search results are all cleared, my collapsed tests are uncollapsed

This then confuses one as beats the point of the features

### Desired behavior

User search and collapsed states are saved, "Specs" button should act as a back button to the "Specs" list.

### Test code to reproduce

1) Make sure you collapse some tests in spec list or do a search.
2) Run a test from long list of specs.
3) After the run finishes, on the left side click "Specs" button
4) All previous settings are cleared

### Cypress Version

10.1.0

lmiller1990:
Hi! Thanks for trying out Cypress 10. This is very good suggestion, I think it should work like a text editor/IDE (like VS Code). I see two separate features/proposals here:

1. remember collapsed state in spec list (both on the landing page, and the run page)
2. remember search text (both pages)

I think these are both good ideas. I will forward them to the design team 👍 

We are collecting all the feedback around Cypress 10 (UI/UX, features, etc) and prioritizing accordingly. Please stay tuned for lots of updates in the near future.

Also, if you have any other ideas or feedback, please make them heard - issues are good (like this one) or you can also create a discussion (here: https://github.com/cypress-io/cypress/discussions). 👊

marktnoonan:
Reading this carefully, at first I thought you were describing a bug with the Inline Specs List that appears on the left hand side of the runner. But that is working as expected, and I think instead you might not be aware of its existence at all. There are two things in the UI that could be called "the specs button on the left" and there are two kinds of "collapsing" that can happen, so it took me a second to put together.

But first, there is an [issue](https://github.com/cypress-io/cypress/issues/21837) open that should be done in some release fairly soon, that will save the search term on the main specs list page. Saving the collapsed folder state is a good one as well, and we will probably create a new issue for that, or roll it into the saved search feature.

But for now, I want to confirm you are aware of the specs list that opens with this button on the top left of the runner:

[...]

PR #22388, opened 2022-06-17
chore: dont render unnecessary dom elements

emilyrohrbough:
Updates to only render dom elements that are needed. Hiding elements causes the page to get bloated with content that isn't needed (I see you legacy studio code). 

### User facing changelog

n/a

### PR Tasks


packages/app/src/runner/ResizablePanels.vue
```diff
@@ -10,7 +10,7 @@
     @mousemove="handleMousemove"
   >
     <div
-      v-show="showPanel1"
+      v-if="showPanel1"
```
tbiethman, line 13: Does this has a noticeable impact on presentation speed with really large spec lists? I'm curious if hiding this rather than unmounting it was intentional @marktnoonan

emilyrohrbough: Hmmm interesting perspective. I know it's rendering in Run mode as hidden when we really don't need to render it at all.

marktnoonan: The main reason to keep the `v-show` is that it avoids the state of the list getting blown away when the list is collapsed, since collapsing won't destroy the component. We don't seem to have a cypress-in-cypress test for this behavior, but we probably should add one. Especially for the moment (until the main specs list performs the same way) I am directing people to the inline specs list who like having the search term saved, or their collapsed folders staying collapsed - [for example here](https://github.com/cypress-io/cypress/issues/22263#issuecomment-1155438058). So we should keep that behavior.

There probably is some value for long lists as well, to just render them once and be done with it. 

In run mode, we don't pass anything to this slot, so the extra div itself doesn't seem so painful. But if we do want to remove it, we could wrap the elements related to it (this div and the sibling resize handle) in a `template` and add a new prop to ResizablePanels to control a `v-if` on the `<template>` - mainly to avoid this component having to know about "run" vs "open".

emilyrohrbough: fixed: [`e884de5` (#22388)](https://github.com/cypress-io/cypress/pull/22388/commits/e884de51998273e7b7bf1838f677f5002ad6260c)

tbiethman (approved):

mjhenkes (approved):

PR #22393, opened 2022-06-19
feat: auto-collapse sidebar for window width < lg

MuazOthman:
- Closes 

### User facing changelog

Automatically collapse sidebar, and keep it collapsed, when window width is narrow to better utilize screen real estate.

### Steps to test

1. Start with a window width >=1024px, make sure the sidebar is expanded.
2. Decrease the window width to become below 1024px, the sidebar should automatically collapse.
3. As long as window width is below 1024px, the sidebar should be collapsed with no option to be expanded.
4. Going back to a window width >=1024px should automatically expand the sidebar if it was expanded before.
5. Going back to a window width >=1024px should keep the sidebar collapsed if it was collapsed before.

### PR Tasks

### DIFF
--- a/packages/app/cypress/e2e/settings.cy.ts
+++ b/packages/app/cypress/e2e/settings.cy.ts
@@ -1,5 +1,7 @@
 import type { SinonStub } from 'sinon'
 
+const SidebarSettingsLinkSelector = '[data-cy="sidebar-link-settings-page"]'
+
 describe('App: Settings', () => {
   before(() => {
     cy.scaffoldProject('todos', { timeout: 50 * 1000 })
@@ -12,7 +14,7 @@ describe('App: Settings', () => {
   it('visits settings page', () => {
     cy.startAppServer('e2e')
     cy.visitApp()
-    cy.findByText('Settings').click()
+    cy.get(SidebarSettingsLinkSelector).click()
 
     cy.get('div[data-cy="app-header-bar"]').should('contain', 'Settings')
     cy.findByText('Device Settings').should('be.visible')
@@ -22,7 +24,7 @@ describe('App: Settings', () => {
   it('shows a button to log in if user is not connected', () => {
     cy.startAppServer('e2e')
     cy.visitApp()
-    cy.findByText('Settings').click()
+    cy.get(SidebarSettingsLinkSelector).click()
     cy.findByText('Project Settings').click()
     cy.get('button').contains('Log In')
   })
@@ -35,7 +37,7 @@ describe('App: Settings', () => {
 
       cy.startAppServer('e2e')
       cy.visitApp()
-      cy.findByText('Settings').click()
+      cy.get(SidebarSettingsLinkSelector).click()
       cy.findByText('Dashboard Settings').click()
       cy.findByText('Project ID').should('be.visible')
       cy.get('[data-cy="code-box"]').should('contain', 'fromCli')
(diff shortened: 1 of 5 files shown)


pstakoun (approved):
Tested this out and feels great

FabyoSK (approved):

marktnoonan (changes requested):
This does work and seems like a pretty typical pattern. I'm not sure I like taking away the user's ability to re-expand the list, even though I see why we might want to auto-close it for them. Ideally if we are auto closing something, a user can hover on the menu itself, and expand it over the content main page content so they can still pick an option. [This pattern here](https://vuetifyjs.com/en/components/navigation-drawers/#expand-on-hover) in Vuetify is a good example. But that's a bigger UX thing and doesn't block this PR imo, since the one other place we collapse the menu it cannot be expanded by the user.

The changes needed in this PR are related to the tests, lots of tests seem to expect the nav to be expanded on these pages and will need to be adjusted before this can merge.

There's also some vertical alignment problems on the rows on this branch at smaller breakpoints, not sure what that's about, possibly it is already fixed on the main feature branch:

Screen Shot 2022-06-23 at 4 51 51 PM

marktnoonan:
@pstakoun @mike-plummer what's next for this PR?

1. We want this and we update it?
2. We don't want this right now and we close it?

marktnoonan:
Since our new process will need design 👀 on this PR, maybe @elevatebart or @mapsandapps can get us some confirmation ahead of time that this behavior is good to go, before we prioritize somebody cleaning up the conflicts and fixing the tests.

Thinking about it more, here's my concern:

1. The button to expand the navigation is already invisible, you have to hover over a particular place to know you can toggle the nav
2. Now there is a secret rule that will mean that at certain sizes, this button doesn't appear at all even when you hover, so things get even less discoverable

If we are going to decide for users when to expand/collapse the nav as a responsive thing, or on certain pages like the runner, maybe we should fully remove the button to toggle it, since there are now fewer situations where that's possible to do, and we override your preference anyway when we think it's better a different way. I think it's weird to have interaction patterns that are only valid some of the time, with no indication of what those times are. Better to just not have it.

But the root issue of space here could also be solved by having the means of toggling the nav be more discoverable in the first place, so if a user wants more space, they can easily see how to get it. This feedback has already been given be a few people in a more general sense. With a visible control, even if we do sometimes disable the resizing, we can also visually remove the button that triggers it, so it's always clear whether resizing is available or not.

mapsandapps:
@marktnoonan (or whomever): where did the request & requirements for this feature come from? there's no associated ticket. (just wanting context.)  

i can take this to design, yeah.

mapsandapps (approved):
approved by design ✅

lmiller1990:
Looks like we got three ✅ but since the original branch is merged, someone will need to rebase this. The OP is off for the next little while, anyone want to rebase?

packages/app/cypress/e2e/settings.cy.ts
```diff
@@ -139,7 +141,7 @@ describe('App: Settings', () => {
       cy.loginUser()
 
       cy.visitApp()
-      cy.findByText('Settings').click()
+      cy.findByTestId('sidebar-link-settings-page').click()
```
marktnoonan, line 144: Hmm, should this be `cy.get(SidebarSettingsLinkSelector)`? Or it's probably more typical for all the others to be `findByTestId('sidebar-link-settings-page')`, since the selector is a test id.

marktnoonan (approved):
Approved! One tiny optional suggestion on the tests.

PR #22411, opened 2022-06-21
fix: show docs menu in column style when viewport is small.

### LINKED ISSUE #21842: Docs dropdown is cut off when on smaller window screens.
## **Summary**
The docs dropdown isn’t responsive when the window is a smaller size. 

## **Acceptance Criteria**
* Make docs dropdown responsive so all content is shown in smaller window sizes and not cut off. 

Screen Shot 2022-04-21 at 10 02 36 AM

[https://cypressio.slack.com/archives/C02MW5QBD4K/p1650557009496909](https://cypressio.slack.com/archives/C02MW5QBD4K/p1650557009496909|smart-card)

┆Issue is synchronized with this [Jira Task](https://cypress-io.atlassian.net/browse/UNIFY-1602) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/11807">Screen Shot 2022-04-21 at 10.02.36 AM.png</a>
┆author: Matthew Schile
┆epic: UI/UX Improvements
┆friendlyId: UNIFY-1602
┆priority: Lowest
┆sprint: Backlog
┆taskType: Task

sainthkh:
- Closes #21842

### User facing changelog

Show docs menu in column style when the viewport is small.

### Additional details
- Why was this change necessary? => When viewport is small, the left side of the docs menu is cut off.

I need help. I implemented this by creating a new class, `docs-menu-flex-direction`. 

I tried it with `flex-col md:flex-row` but it doesn't work. The generated windicss code isn't injected at `<head>`. Did I miss something? 

I wrote the test could be reusable. 

### Steps to test

Click the `Docs` button. 

**Before:**

image

**After:**

image

### PR Tasks


sainthkh:
Most of them failed because of the 503 failure from npm and netlify.

packages/frontend-shared/src/gql-components/topnav/TopNav.vue
```diff
@@ -143,7 +143,8 @@
     </template>
     <div
       v-if="docsMenuVariant === 'main'"
-      class="flex p-16px gap-24px"
+      data-cy="docs-menu-container"
+      class="flex p-16px gap-24px docs-menu-flex-direction"
```
rachelruderman, line 147: @sainthkh we've been using Tailwind CSS in the Vue components. According to their docs on [flex direction](https://tailwindcss.com/docs/flex-direction) and [responsive design](https://tailwindcss.com/docs/responsive-design), this should do the trick. Can you try it out? Thanks!
```suggestion
      class="flex p-16px gap-24px flex-col md:flex-row"
```

packages/frontend-shared/src/gql-components/topnav/TopNav.vue
```diff
@@ -357,3 +358,15 @@ const changelogLink = (version) => {
...
+.docs-menu-flex-direction {
+  flex-direction: column;
+}
+
+@media(min-width: 768px) {
+  .docs-menu-flex-direction {
+    flex-direction: row;
+  }
+}
+</style>
```
rachelruderman, line 372:
```suggestion
```
Remove in favor of Tailwind CSS

rachelruderman (changes requested):

packages/frontend-shared/src/gql-components/topnav/TopNav.vue
rachelruderman: AHHHH nevermind, I just re-read your GitHub description with the explanation on it not working

packages/frontend-shared/src/gql-components/topnav/TopNav.vue
rachelruderman: AHHHH nevermind, I just re-read your GitHub description with the explanation on it not working

packages/frontend-shared/src/gql-components/topnav/TopNav.vue
rachelruderman: @sainthkh I just tried it out with the Tailwind CSS did work for me, can you confirm? Make sure to refresh the launchpad to see changes since it doesn't hot reload

rachelruderman: Alrighty I put in the Tailwind CSS change and reorganized the tests to be scoped in a `responsive design` context block. I also broke them out into two tests to take advantage of the `percySnapshot` command in the `afterEach`. Sorry it took so long to get in and big thanks for your contribution 🙏

packages/frontend-shared/src/gql-components/HeaderBarContent.cy.tsx
```diff
@@ -162,6 +162,44 @@ describe('<HeaderBarContent />', { viewportWidth: 1000, viewportHeight: 750 }, (
...
+    // https://github.com/cypress-io/cypress/issues/21842
+    it('shows docs menu correctly on small viewports', () => {
+      // Simulate the small viewport.
+      cy.viewport(767, 800)
+
+      // show docs menu
+      cy.contains('button', text.docsMenu.docsHeading).click()
+
+      // docs menu flex direction is column when viewport width is small
+      cy.findByTestId('docs-menu-container').should('have.css', 'flex-direction', 'column')
```
marktnoonan, line 188: Let's add a `cy.percySnapshot()`for each viewport. I know they won't run for this outside contributor PR but I think it's a better test than asserting the flex-direction. It would be easy to rearrange the internals of this component in a way that breaks one breakpoint (say by adding an intermediate layout container) but still passes this test.

marktnoonan (changes requested):
This is an improvement! I'd like to add some Percy coverage for visual regression, so that's the blocking change request here. 

I'd also suggest, optionally, instead of testing the flex direction property, we should assert something like the width/height of the menu. Or nothing, and let visual regression be the only check. I think the `flex-direction` might provide false confidence and could continue passing even if internal changes bork flex behavior. I'm generally against checking CSS properties directly if there's some expected user-facing property we can check instead.

Example - in narrow viewports, the menu width should be less than the height, and in wide viewports, the menu height should be less than the width. This would confirm we've meaningfully changed the aspect ratio, without worrying about implementation details (flex/grid/CSS columns could all do this kind of thing).

But not blocking, since we can always improve that separately later.

packages/frontend-shared/src/gql-components/HeaderBarContent.cy.tsx
rachelruderman: @marktnoonan `cy.percySnapshot` comes along free for the ride -- it's in the top `afterEach` :)

PR #22427, opened 2022-06-21
feat: Update ACI tooltip copy

pstakoun:
- Closes 

### PR Tasks

### DIFF
--- a/packages/frontend-shared/src/locales/en-US.json
+++ b/packages/frontend-shared/src/locales/en-US.json
@@ -135,8 +135,8 @@
       "headerShort": "Runs",
       "tooltip": {
         "connected": "The {0} in the Cypress Dashboard",
-        "notConnected": "View the {0} in the Cypress Dashboard",
-        "noAccess": "Request access to view the {0} in the Cypress Dashboard",
+        "notConnected": "Connect to the Cypress Dashboard to see the {0}",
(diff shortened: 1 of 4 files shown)


mike-plummer (commented):
Changes look okay, but you've got some e2e test failures. See `settings.cy.ts`

mike-plummer (approved):
👍 Only comment is the discrepancy between sometimes saying "the Dashboard" and other times "the Cypress Dashboard". I saw y'all were discussing that earlier so I'll assume everyone is on board with it as-is

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -426,7 +426,7 @@
       "errorNotFound": "Project not found",
       "errorNotFoundButton": "Connect it again",
       "errorNotLoggedIn": "You are not connected to the Cypress Dashboard",
-      "errorNotLoggedInButton": "Log in to Cypress Dashboard"
+      "errorNotLoggedInButton": "Log in to the Cypress Dashboard"
```
pstakoun, line 429: I'm trying to figure out where this is used - can't find usages of `errorNotLoggedInButton`.

mike-plummer: Looks like several of these strings aren't referenced (most of the `error*` entries under `recordKey` anyways)

PR #22435, opened 2022-06-21
fix: Address GQL runs data being lost after navigation

mike-plummer:
* Average Runs data persistence
Update GQL query to work around issue that caused data to not hydrate properly when moving back from the Runs page

* Tooltip behavior
Minor tooltip tweaks to improve behavior when moving rapidly between SpecList headers

### Steps to test
1. Load project with recorded run(s)
2. Open Specs List, observe latest runs data load
3. Navigate to the Cloud Runs page then back to Specs
4. Observe previously-loaded latest runs data is still present
	* Without this fix, latest runs appear unloaded and do not trigger re-load
5. Move mouse rapidly between Specs List headers, it should no longer be possible to have multiple header tooltips remain open

### PR Tasks

### DIFF
--- a/packages/app/cypress/e2e/specs_list_latest_runs.cy.ts
+++ b/packages/app/cypress/e2e/specs_list_latest_runs.cy.ts
@@ -352,6 +352,26 @@ describe('ACI - Latest runs and Average duration', { viewportWidth: 1200, viewpo
         .should('have.attr', 'aria-expanded', 'true')
       })
     })
+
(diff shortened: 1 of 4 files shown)


tgriesser (approved):
Cool, still would like to investigate further to understand where exactly this is getting lost in translation, but glad it's fixed for now - do you mind following up with a test that covers this so we don't regress.

pstakoun (approved):
Works for me, and tooltips feel good too!

PR #22437, opened 2022-06-22
feat: support React 18

### LINKED ISSUE #21381: Cypress Testing fails to build on React version 18

### LINKED ISSUE #21946: React 18 support - today logs a bunch of warnings for ReactDom.render being deprecated

lmiller1990:

### User facing changelog

Add support for React 18 to `npm/react`.

### Additional details

- updated `peerDependencies` to allow installing `@cypress/react` with `react@18`
- some specific code in vite and webpack dev servers to account for the potentially not available `react-dom/client` code

The problem and solution is described in detail in the README [in this repo](https://github.com/lmiller1990/es-modules-example) which is an example project used for an internal presentation at Cypress.

### Steps to test

- Build binary (local should be fine too, but binary is more "production like")
- Try using Cypress CT with a React 18 project (using new `react-dom/client` API) 
  -  For example 
    - webpack: `yarn create react-app my-app` 
    - vite: `yarn create vite --template react my-vite-react-app`
  - Point Cypress at your app
  - Configure React using Launchpad
  - Should work, no errors in console, etc.
- No warning/error in console about deprecated React 18 API etc.

### PR Tasks


npm/vite-dev-server/client/reactDomClientPlaceholder.js
```diff
@@ -0,0 +1,2 @@
+// Placeholder ...
```
mike-plummer, line 1: Might be nice to describe the intent/purpose of this placeholder. I guess there is a comment in your `react18.ts` Vite plugin so I guess it's documented, but maybe we can just hint to it here `Placeholder for React18 plugin`

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -72,6 +77,67 @@ function modifyWebpackConfigForCypress (webpackConfig: Partial<Configuration>) {
   return webpackConfig
 }
 
+const versionRe = /(\d.+?)\./
```
mike-plummer, line 80: Same comment regarding `semver` here - I'd like it if we could use the lib just to avoid the logic duplication, but I get that adding a Node dependency isn't always desirable

npm/react/src/mount.ts
```diff
@@ -41,6 +41,36 @@ const injectStyles = (options: MountOptions) => {
...
+    throw Error(`Could not extract major version from ${semver}`)
+  }
+
+  return toNum
+}
+
+async function importReactModules () {
+  const react = await import('react')
+  const majorVersion = getMajorVersion(react.version)
+  const reactDomImport = majorVersion <= 17
```
mike-plummer, line 60: Total nitpick: Below we do a `< 18` check, might be good to do that same comparison here just so that it pops up if somebody does a global search for stuff that might be related to React 18

mike-plummer (commented):
Very cool stuff, Lachlan! I know you're still working on this - just had a couple nitpicks/questions as I was perusing

npm/react/src/mount.ts
lmiller1990: Good idea, thanks

npm/vite-dev-server/client/reactDomClientPlaceholder.js
lmiller1990: Will do, thanks!

npm/react/src/mount.ts
```diff
@@ -41,6 +41,36 @@ const injectStyles = (options: MountOptions) => {
  **/
 export const mount = (jsx: React.ReactNode, options: MountOptions = {}) => _mount('mount', jsx, options)
 
+const versionRe = /(\d.+?)\./
```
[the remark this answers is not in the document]
lmiller1990: Had the same thought on the semver library - I am more conscious about adding dependencies that bundle each time the user runs a test, since they could impact spec time. We grab their internal regexp: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

I think our regexp do the same thing - I can definitely add some unit tests around this regexp, now I've figured out the hard parts w.r.t the dependencies and system tests.

npm/webpack-dev-server/src/makeWebpackConfig.ts
lmiller1990: I like to keep the `cypress/{vue,react}` bundles really lean, but we definitely need a lot more tests around this regexp. I might grab the relevant code from the semver library, starting w/ the regexp  and test cases perhaps: https://gist.github.com/jhorsman/62eeea161a13b80e39f5249281e17c39

lmiller1990:
All the tests are passing :tada: the outstanding ones are something to do with Percy, not actual broken tests.

I'm off next week, but we will review and get this merged the week after (11th July). Thanks for waiting everyone!

system-tests/projects/react18/vite.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from 'vite'
+import react from '@vitejs/plugin-react'
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  plugins: [react()],
```
ZachJW34, line 6: Should we move this up into the `react` system test fixture?

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -72,6 +77,67 @@ function modifyWebpackConfigForCypress (webpackConfig: Partial<Configuration>) {
...
+ * We ignore `react-dom/client` if the React is present and
+ * we the version is <= 17.
+ * We support webpack v4 and v5, and the `IgnorePlugin` is slightly
+ * different between the two, so we check which version of webpack
+ * we are using and match the API.
+ */
+function makeIgnorePlugin (devServerConfig: WebpackDevServerConfig) {
+  debug('making react-version specific ignore plugin')
+
+  let webpackModule
```
ZachJW34, line 113: We don't need to do another detection of the webpack module since we [already have it](https://github.com/cypress-io/cypress/blob/871e54fa6b38bcb714d2ea73bdf90759dbdf0466/npm/webpack-dev-server/src/makeWebpackConfig.ts#L82). Is there something wrong with using `webpack.IgnorePlugin` directly?

npm/react/package.json
```diff
@@ -37,10 +37,10 @@
     "vite-plugin-require-transform": "1.0.3"
   },
   "peerDependencies": {
-    "@types/react": "^16.9.16 || ^17.0.0",
+    "@types/react": "^16.9.16 || ^17.0.0 || ^18.0.0",
     "cypress": "*",
-    "react": "^=16.x || ^=17.x",
-    "react-dom": "^=16.x || ^=17.x"
+    "react": "^=16.x || ^=17.x || ^=18",
```
ZachJW34, line 42: Any reason to not use "|| ^=18`.x`"?

lmiller1990: 8b07baf

npm/webpack-dev-server/src/makeWebpackConfig.ts
lmiller1990: Actually yes - if we just do `require('webpack')` it grabs the one we bundle with Cypress. If that is different to the one they are using (eg, dist'd one from Next.js, which might be a different version of webpack) it might not work. We need to make sure the `IgnorePlugin` comes from the same version of webpack that they are using.

It's possible this is only a bug during our system tests because of symlinks, but I had to add this or the Next.js tests would fail.

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -140,21 +206,44 @@ export async function makeWebpackConfig (
...
-        projectRoot,
-        devServerEvents,
-        supportFile,
-        webpack,
-      }),
-    ],
+    plugins: [],
+  }
+
+  try {
```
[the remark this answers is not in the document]
lmiller1990: Sure - b21c6bb

npm/webpack-dev-server/src/makeWebpackConfig.ts
lmiller1990: Let's just use semver! c8fee62

npm/react/src/mount.ts
lmiller1990: Made function and using semver: c8fee62

system-tests/test/component_testing_spec.ts
```diff
@@ -68,3 +68,47 @@ describe('component testing projects', function () {
...
+        testingType: 'component',
+        browser: 'chrome',
+        snapshot: true,
+        expectedExitCode: 0,
+      })
+    })
+  }
+})
+
+// TODO: Get these to pass.
```
lmiller1990, line 92: This works now

npm/react/src/mount.ts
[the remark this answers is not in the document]
lmiller1990: I ended up adding semver since we can just import function-by-function which is nice and small

ZachJW34 (approved):
Only holdup I have is the handling of the `IgnorePlugin`. If we can't use the webpack module from `sourceWebpackModulesResult.webpack.module` because it doesn't point to the correct Webpack version for Next projects then we have a bigger issue (we wouldn't be compiling projects with the expected webpack version and that would be bad).

I tested this and it works great, we chatted a bit about the above so I'm going to approve, hopefully we can remove the extra webpack sourcing.

astone123 (approved):
Works well for me 👍🏻

lmiller1990:
Sodatea from Vite team suggests looking here for 504 related:

- https://github.com/vitejs/vite/blob/cda5442fc5eae9a571536fb2fcea9d9f16c28365/packages/vite/src/node/server/middlewares/transform.ts#L203
- https://github.com/vitejs/vite/blob/cda5442fc5eae9a571536fb2fcea9d9f16c28365/packages/vite/src/node/plugins/importAnalysis.ts#L224-L230

Reproduction: https://github.com/lmiller1990/vite-react-17-error-with-cypress (uses this pre-release). This is only a problem for React 17 and Vite. And not happening in system-tests, only in prod. I don't understand why.

ZachJW34:
In response to https://github.com/cypress-io/cypress/pull/22633#issuecomment-1183976204:

The only reason we are concerned about the behavior of what bundlers/preprocessors will do is because of the bundler hacks we are doing in order to support React17 and React18 in a single package. Ideally, a single package is best as it provides the user with the best experience ("it just works" idealogy), but as you pointed out we are violating this since e2e tests will fail due to the webpack-preprocessor not having the included bundler hacks if it happens to import the mount.

In Slack you mentioned publishing another package (`@cypress/react18`) and I am leaning in that direction. Generally, I don't think we should introduce bundler hacks to support multi-version support. If we can get away with it without the hacks then sure, but this PR is leading us down a rabbit hole and I think that's a sign to go forward with separating the support into multiple packages.

As we expand our framework support, I fear these hacks will continue to grow. @lmiller1990 what are your thoughts on proceeding with what we have now vs publishing multiple packages? If we decide on the multiple packages, we will have some things to think through with regards to versioning.

BlueWinds:
> In Slack you mentioned publishing another package (@cypress/react18) and I am leaning in that direction. Generally, I don't think we should introduce bundler hacks to support multi-version support. If we can get away with it without the hacks then sure, but this PR is leading us down a rabbit hole and I think that's a sign to go forward with separating the support into multiple packages.

This is the path we've gone down with the vue packages already (vue (for vue3) / vue2) - when we need to support another version, we release another module. I personally would have preferred we go with vueN (supporting version N) rather than vue (supporting vue3) and vueN (supporting older version N).

lmiller1990:
I too am leaning towards two packages. I (finally) solved all the problems here, but what a mess - just because we can, doesn't mean we should, comes to mind.

Main question to answer: when does `cypress/react18` become the "main" package (ala `cypress/react`) and the current `cypress/react` become `cypress/react-17`?

I wish we just did `cypress/vue2` and `cypress/vue3` instead of `cypress/vue` - maybe we should just do this, pairing w/ each major. Historically these libraries have been good at not breaking things in minor/patches.

Also we can loop in Jordan for Angular - Angular drops a new major once every 6 months? We should fine a pattern that fits angular.

BlueWinds:
> Maybe we alias `cypress/vue` to `cypress/vue3` and promote importing the major you want?

That seems like a good way forward - don't break anyone, create universal support for a new pattern.

> Also we can loop in Jordan for Angular - Angular drops a new major once every 6 months? We should fine a pattern that fits angular.

Hm, I know very little about angular - getting someone who does to chime in sounds like a good step forward. Shouldn't assume that every package operates the same.

lmiller1990:
@BlueWinds ok, I'll prepare a tech brief and include you and Jordan and Zach.

astone123:
With this approach, what happens if a breaking change comes out of a minor or patch release from react/vue/angular? 

Will we internalize those changes in the major package, or will we create a new package for the patch/minor?

Issue #17853, opened 2021-08-23
Tests causing CI agents to run out of memory on cypress 8.2 & above


vaibhavganatra:
### Current behavior

The Azure CI Agent machine running out of memory when running cypress tests on 8.2 & above on Azure Pipeline CI agents (Linux OS). The cypress dashboard eventually times out as well. 

<--- Last few GCs --->

```
[31:0x1c7300000000]  8599171 ms: Mark-sweep 1835.4 (2072.0) -> 1832.6 (2071.6) MB, 205.8 / 0.0 ms  (average mu = 0.263, current mu = 0.223) allocation failure scavenge might not succeed
[31:0x1c7300000000]  8599492 ms: Mark-sweep 1835.0 (2072.0) -> 1832.6 (2071.6) MB, 294.6 / 0.0 ms  (average mu = 0.166, current mu = 0.082) allocation failure scavenge might not succeed
```

<--- JS stacktrace --->

```
The Test Runner unexpectedly exited via a exit event with signal SIGSEGV

Please search Cypress documentation for possible solutions:

https://on.cypress.io

Check if there is a GitHub issue describing this crash:

https://github.com/cypress-io/cypress/issues

Consider opening a new issue.

----------

Platform: linux (Debian - 10.10)
Cypress Version: 8.3.0
```

### Test code to reproduce

This has been consistently reproducible on 8.2 and 8.3. Running the same spec files on the same agents on version 8.1 and below works as expected.

### Cypress Version

8.3

cellog:
a note: in a much earlier version of cypress, we fixed this by adding
```json
  "numTestsKeptInMemory": 1,
```

to our cypress-ci.json, this may help if you're not already doing this

stychu:
I can observe similar issues since 8.2 and higher. My AZURE pipelines just run until the pipeline timeout occurs once I upgrade above 8.1.

On Cypress 8.1 everything works flawlessly

jennifer-shehane:
Is everyone in this thread experiencing this issue in Azure only?

jennifer-shehane:
Could anyone run Cypress in [debug mode](https://on.cypress.io/troubleshooting#Print-DEBUG-logs) mode and print the entire set of logs here?

cupcakepanda89:
I experience this when running cypress in aws codebuild too

VadimSkuratovskyi:
I've got the same issue with 8.3.1 on Azure Linux agents: 
[...]

PR #22460, opened 2022-06-22
fix: memory leak caused by storing base64 encoded files recieved by CDP `Network.requestWillBeSent`

DJSdev:
- Closes https://github.com/cypress-io/cypress/issues/17853

### User facing changelog

Fixed a memory leak caused by storing `data:` url containing base64 encoded files unnecessarily. Fixed: #17853 

#### Summary
There is a memory leak in the `CdpAutomation` class that causes Cypress to store data urls as requests in the `pendingBrowserPreRequests` array.

#### Background
After updating to Cypress any version higher than 8.2, my e2e suite crashes with the following error after about an hour:
```
8672:0000371000344000]  3709798 ms: Mark-sweep 1952.5 (2070.7) -> 1946.9 (2071.4) MB, 450.0 / 0.0 ms  (average mu = 0.789, current mu = 0.668) task; scavenge might not succeed
[8672:0000371000344000]  3711538 ms: Mark-sweep 1953.7 (2071.4) -> 1947.8 (2072.4) MB, 350.9 / 0.0 ms  (average mu = 0.793, current mu = 0.798) task; scavenge might not succeed

<--- JS stacktrace --->

[8672:0622/112740.931:ERROR:node_bindings.cc(143)] Fatal error in V8: Reached heap limit Allocation failed - JavaScript heap out of memory
```

While the tests run, I inspected task manager and I noticed the Cypress server process growing in size until the machine's memory is exhausted causes this issue.

#### Investigation
I started off by adding code to Cypress to have it dump it's heap memory every 90 seconds. After inspecting the heap memory in the DevTools console I discovered that Cypress is storing hundreds/thousands of copies of a font file.

image

This is a font-file used by my application and it's referenced like so:

image

I discovered in Chrome, when a CSS style sheet uses the `url()` function in the `src` field with a base64 encoded file ([a `data:` url](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs)), it still trigger Chrome to "send" a network request even though it's not actually going to send one.

image

Cypress subscribes to the `Network.requestWillBeSent` event in CDP. When Chrome goes to 'fetch' the base64 encoded file, it sends a `requestWillBeSent` event to Cypress which stores the url (literally a whole file) in the `pendingBrowserPreRequests` array. Every time the page is loaded a request would be sent adding to this buffer until it eventually crashes.

#### Solution
The solution I went with is very basic. Simply filtering out `data:` urls from from the `Network.requestWillBeSent` event handling logic should clear it up.

Another thought was to have logic clear out stale entries from the array, but that's a lot of work for URLs that shouldn't be stored to begin with.

### Steps to test

I can generate a test repo that reproduces this issue if anyone requests it.

I tested this with on version 8.2 and 10.2 and verified the issue existed on both versions. (Currently, my team is on 8.0 where this issue doesn't happen. This issue has stopped us from upgrading for a very long time).

I verified after implementing this fix on 10.2, the memory leak no longer happens and data urls are no longer stored in the `pendingBrowserPreRequests` array.

Nothing has changed!

### PR Tasks

- [ ] Have tests been added/updated? There weren't any tests for the function I modified. It seems this private function is difficult to hit due CDP being used to trigger the private function. If anyone has suggestions or examples, I'd be more than happy to add tests.

### DIFF
--- a/packages/server/lib/browsers/cdp_automation.ts
+++ b/packages/server/lib/browsers/cdp_automation.ts
@@ -202,6 +202,15 @@ export class CdpAutomation {
     // in Firefox, the hash is incorrectly included in the URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1715366
     if (url.includes('#')) url = url.slice(0, url.indexOf('#'))
 
+    // Filter out "data:" urls from being cached - should fix: https://github.com/cypress-io/cypress/issues/17853
+    // Chrome sends `Network.requestWillBeSent` events with data: urls which it won't actually fetch
+    // Example data url: "data:font/woff;base64,<base64 encoded string>"
+    if (url.startsWith('data:')) {
+      debugVerbose('skipping `data:` url %s', url)
+
+      return
+    }
+
     // Firefox: https://searchfox.org/mozilla-central/rev/98a9257ca2847fad9a19631ac76199474516b31e/remote/cdp/domains/parent/Network.jsm#397
     // Firefox lacks support for urlFragment and initiator, two nice-to-haves
     const browserPreRequest: BrowserPreRequest = {


mjhenkes:
@DJSdev, Thanks so much for this PR. Would it be possible to get a test written for it?

DJSdev:
I'll do my best to get a test written for it. It's in a weird place.

packages/server/lib/browsers/cdp_automation.ts
```diff
@@ -202,6 +202,15 @@ export class CdpAutomation {
     // in Firefox, the hash is incorrectly included in the URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1715366
     if (url.includes('#')) url = url.slice(0, url.indexOf('#'))
 
+    // Filter out "data:" urls from being cached - should fix: https://github.com/cypress-io/cypress/issues/17853
+    // Chrome sends `Network.requestWillBeSent` events with data urls which that won't actually be fetched
```
samtsai, line 206:
```suggestion
    // Chrome sends `Network.requestWillBeSent` events with data urls which won't actually be fetched
```

BlueWinds:
> I'll do my best to get a test written for it. It's in a weird place.

`packages/server/test/unit/browsers/cdp_automation_spec.ts` has the existing tests for our CDP integration, and has the helper `sendDebuggerCommand` which ought to make it easy to send a mocked up `Network.requestWillBeSent`.

I'd be fine with a test that just emits a  `Network.requestWillBeSent`event and verifies that `onBrowserPreRequest` isn't called when it's with a data url.

DJSdev:
Unit tests have been added. I added some bonus tests around `onNetworkRequestWillBeSent` and `onResponseReceived` since there wasn't any around these two functions.

> existing tests for our CDP integration, and has the helper sendDebuggerCommand which ought to make it easy to send a mocked up Network.requestWillBeSent.

I wasn't exactly sure how the `sendDebuggerCommand` worked in regards to these tests. Hopefully triggering the events with the stubbed `onFn()` will suffice. Looking forward to feedback :)

emilyrohrbough (approved):

BlueWinds (approved):

BlueWinds:
We have been having some flake problems with our automated test suite recently, which we're working on addressing in several other PRs. Given that the currently failing tests were known to be flaky before this branch, I'm going to exercise my judgement and merge this without waiting for a completely green build.

Chili82:
Is this fix available on 10.2.0 version?

BlueWinds:
We're hoping to get 10.2.1 out tomorrow, but it's not ready quite yet.

SadiqRahim:
Does this also mean, the fix would resolve the issue where it closes the test before it kick start with the error "The Test Runner unexpectedly exited via a close event with signal SIGILL"?

BlueWinds:
Doesn't sound related to this fix or the linked issue.

PR #22473, opened 2022-06-23
fix: change "resizable panels handlers" width to make it more convenient to use

Oustinger:
Closes #22473
Closes https://github.com/cypress-io/cypress/issues/22608

### User facing changelog
Changed resizable panels' handlers width to make it more convenient to use

### Additional details
Resizable panels' handlers have too big width that overlap scrollbar.
As user, you have troubles with scrolling tests' log window.
That's why you need to do it less.
I resolved it with changing resizable panels' handlers' className: `w-16` -> `w-10`

### Steps to test
1. Open and run any test
2. Expend inforamtion in tests' log window to see a scrollbar (if it's not  visible)
3. Try to scroll using cursor

Before: [video](https://recordit.co/jRUcJrPG0V)

After: [video](https://recordit.co/z384pudkL3)

### PR Tasks


marktnoonan:
Nice, this was a good thing to raise. Initially this looks good and seems justified, without making the click target too small. That's the real reason it's so wide, so users don't have to be super precise with the resizing. I'll pull it down and test but seems like this does not lose too much of that.

Seems like we should do this for the `panel1ResizeHandle` as well so the behavior is the same for both. Panel 1 is the inline specs list and can also have a scrollbar in the same places.

For reference, Panel 1 can be opened by clicking the "specs" button here.

Screen Shot 2022-06-20 at 9 34 03 PM

lmiller1990:
cc @marktnoonan

resolves: https://github.com/cypress-io/cypress/issues/22608

marktnoonan (approved):
Tested locally and both work great

Oustinger:
@marktnoonan Am I need to do any thing else? Or it's ok that PR is still be opened?

lmiller1990:
@Oustinger nothing left, just need CI green - I'll take a look.

PR #22483, opened 2022-06-23
fix: remove ctx.cloud.reset in tests, handle infinite loop in stale results

tgriesser:
Related to #21250

### Additional details
Correctly fixes tests failing due to the introduction of the new queries on the specs page, properly handles an infinite loop that was occurring when an incomplete response was sent back for a `stale` fetch result. This would cause an infinite re-fetch loop while the server was looking to fulfill the stale value. Instead, now if we detect this situation we clear the urql cache in both the electron process and the client.

### Steps to test

Existing tests should be passing with only minor, expected modifications

### PR Tasks

### DIFF
--- a/packages/data-context/src/sources/CloudDataSource.ts
+++ b/packages/data-context/src/sources/CloudDataSource.ts
@@ -52,12 +52,6 @@ export interface CloudExecuteRemote extends CloudExecuteQuery {
   operationType?: OperationTypeNode
   requestPolicy?: RequestPolicy
   onUpdatedResult?: (data: any) => any
-  /**
-   * Triggered when we have an initial stale response that is not fulfilled
-   * by an additional fetch to the server. This means we've gotten into a bad state
-   * and we need to clear both the server & client side cache
-   */
-  invalidateCache: () => any
 }
 
 export interface CloudExecuteDelegateFieldParams<F extends CloudQueryField> {
@@ -71,6 +65,12 @@ export interface CloudDataSourceParams {
   fetch: typeof fetch
   getUser(): AuthenticatedUserShape | null
   logout(): void
+  /**
+   * Triggered when we have an initial stale response that is not fulfilled
+   * by an additional fetch to the server. This means we've gotten into a bad state
+   * and we need to clear both the server & client side cache
+   */
+  invalidateClientUrqlCache(): void
 }
 
 /**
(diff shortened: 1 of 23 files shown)


packages/graphql/src/stitching/remoteSchemaWrapped.ts
```diff
@@ -65,13 +68,16 @@ export const remoteSchemaWrapped = wrapSchema<DataContext>({
...
-        obj.context?.graphql.pushResult({
+        context.graphql.pushResult({
           result: result?.[info.fieldName] ?? null,
           source: obj.rootValue,
           info,
-          ctx: obj.context,
+          ctx: context,
         })
       },
+      invalidateCache () {
```
tgriesser, line 78: Not super happy with the naming of this, might change to `onInvalidateCache`, and centralize the client & server invalidation calls as like `ctx.invalidateGraphQLCache()`.

packages/graphql/src/stitching/remoteSchemaWrapped.ts
```diff
@@ -65,13 +68,16 @@ export const remoteSchemaWrapped = wrapSchema<DataContext>({
       // When we respond eagerly with a result, but receive an updated value
       // for the query, we can "push" the data down using the pushFragment subscription
       onUpdatedResult (result) {
-        obj.context?.graphql.pushResult({
```
mike-plummer, line 68: Looks like the TS typings are good with it, but the original code treated `context` as nullable. Any risk of that here?

tgriesser: No, because it's assigned above so it can't change

estrada9166 (approved):

packages/data-context/src/sources/CloudDataSource.ts
```diff
@@ -48,9 +48,16 @@ export interface CloudExecuteQuery {
...
+  fieldName: string
   operationType?: OperationTypeNode
   requestPolicy?: RequestPolicy
   onUpdatedResult?: (data: any) => any
+  /**
+   * Triggered when we have an initial stale response that is not fulfilled
+   * by an additional fetch to the server. This means we've gotten into a bad state
+   * and we need to clear both the server & client side cache
+   */
+  invalidateCache: () => any
```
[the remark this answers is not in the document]
tgriesser: Nah it should be void, will updated

PR #22489, opened 2022-06-23
fix: Increase timeout for npm-webpack-dev-server tests

BlueWinds:
### User facing changelog
No user facing changes.

### Additional details
The system tests in `sourceRelativeWebpackModules.spec.ts` have been extremely flaky the past few days, to the point where it's more "consistent failure" than it is "flake".

It seems to be simply a timeout issue - the 10s we've allocated aren't enough to successfully yarn install all the dependencies for `webpack5_wds4-react` and similar system test projects.

While I'd very much like to rewrite these tests so as not to take so long, there's no quick solution here and it would probably involve changing how we install system test dependencies entirely (or mocking them out not to actually install), something beyond the scope of my efforts at the moment.

Therefore, timeout increase. Green builds are super important.

### PR Tasks

### DIFF
--- a/npm/webpack-dev-server/test/.mocharc.js
+++ b/npm/webpack-dev-server/test/.mocharc.js
@@ -1,4 +1,4 @@
 module.exports = {
   spec: 'test/**/*.spec.ts',
-  timeout: 10000,
+  timeout: 15000,
 }


AtofStryker (approved):

tbiethman (approved):

BlueWinds:
Because this change affects only one set of CI jobs, and they've passed cleanly on the first try (while completely failing today in many other branches), I'm going to merge this in without waiting for a completely green build so that I can move on and address other failing tests separately.

Issue #22461, opened 2022-06-22
Migration Helper loads endlessly


Maleckidd:
### Current behavior

Hi! After update to latest version 10.2.0 (from 9.1.0) and run 'cypress open', I have clicked button 'Continue to Cypress 10', Migration Helper have been shown, but it's loading endlessly (screenshot).  There is no errors in terminal or cypress window
Here is my repo: https://github.com/Maleckidd/CypressExample-OWASPJuiceShop
![image](https://user-images.githubusercontent.com/38877513/175133370-c350ef71-7b1a-48c0-915e-2669a17e7f7f.png)

### Cypress Version

10.2.0

### Package Manager

npm

### Operating system

Mac

lmiller1990:
Thanks for the reproduction - I confirmed it's happening on linux, too. I will find out the problem and fix it.

lmiller1990:
I found the problem. The migration tool is not respecting when `pluginsFile` and `supportFile` point to directories containing `index.js`. I think `pluginsFile` and `supportFile` are supposed to point to files (thus the name) but it looks like, by Node.js default module resolution, directories with `index.js` also worked in Cypress 9.

I'm fixing the migration helper now. You can wait until the patch is merged and try again, or you can just change your `cypress.json` to point to the files directly, like this:

```json
{
   "baseUrl": "http://localhost:3000/",
   "integrationFolder": "JuiceShopTests",
   "fixturesFolder": "fixtures",
   "pluginsFile": "plugins/index.js",
   "screenshotsFolder": "screenshots",
   "supportFile": "support/index.js"
}
```

I made these changes and it works.

Maleckidd:
thx, workaround works!

PR #22501, opened 2022-06-24
fix: handle case of implicit plugins/index.js files during migration

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/22461

### User facing changelog

Fix a bug in migration where the old `pluginsFile` API is not respected during migration, causing the migration tool to hang without a response.

### Additional details

The migration tool is not respecting when `pluginsFile` points to a directory containing `index.js`. For example:

```
- root
--- cypress
------ plugins
-------- index.js
```

If you set your `cypress.json` to

```json
{
   "pluginsFile": "cypress/plugins"
}
```

Cypress 9 would figure out you wanted `cypress/plugins/index.js`, since we just did `require(config.pluginsFile)`. It's not really documented, but it worked, so we should respect this during migration.

Although this was not explicitly documented, users coming from this usage encounter a migration screen that doesn't show any error, which is a terrible first experience when using Cypress 10, so I think it's better to just support this than error confusingly.

### Steps to test

1. Grab reproduction from https://github.com/cypress-io/cypress/issues/22461
2. Try to migrate it

Note: `supportFile` pointing to a directory with `index.js` also worked previously, although not explicitly documented. This throws a reasonable error telling the user the problem, so it's not as problematic as the case of `pluginsFile`, where it just sat on a screen with a spinner.

Can migrate without getting stuck.

### PR Tasks


packages/data-context/src/sources/migration/codegen.ts
```diff
@@ -188,6 +190,67 @@ function formatObjectForConfig (obj: Record<string, unknown>) {
...
+
+      debug('found pluginsFile %s', pathToIndex)
+
+      return pathToIndex
+    }
+  }
+
+  debug('error, could not find path to pluginsFile!')
+
+  throw Error(`Could not find pluginsFile. Received projectRoot ${projectRoot} and pluginPath: ${pluginPath}`)
```
emilyrohrbough, line 251:
```suggestion
  throw Error(`Could not find pluginsFile in %s`, pluginLoc)
```

emilyrohrbough (approved):

tbiethman (approved):
Works good. I appreciate the extra logs in here :+1:

mjhenkes (approved):

PR #22510, opened 2022-06-24
fix: Keep copy button in view, remove button when file already written

### LINKED ISSUE #21860: Remove 'copy' button for code blocks where the code is already written to file
## **Summary**
It's unclear why there's a Copy button in each code block here when it has already written the files to disk for me

image (43)

## **Acceptance Criteria**
* Should…
* Should also…

### **Resources**

Slack thread pointing out that we should also do this for the end of the spec creation flow, where we show the created spec. Also clarifies that this is just a selective removal of the `copyButton` prop, which defaults to false, in some usages of the ShikiHighlight.vue component. 

[https://cypressio.slack.com/archives/C02MYBT9Y5S/p1652456320690329?thread_ts=1652453834.024469&cid=C02MYBT9Y5S](https://cypressio.slack.com/archives/C02MYBT9Y5S/p1652456320690329?thread_ts=1652453834.024469&cid=C02MYBT9Y5S|smart-card)

[https://cypressio.slack.com/archives/C02MYBT9Y5S/p1650562301003789](https://cypressio.slack.com/archives/C02MYBT9Y5S/p1650562301003789|smart-card)

┆Issue is synchronized with this [Jira Task](https://cypress-io.atlassian.net/browse/UNIFY-1647) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/11856">image (43).png</a>
┆author: Jennifer Shehane
┆epic: UI/UX Improvements
┆friendlyId: UNIFY-1647
┆priority: Low
┆sprint: Backlog
┆taskType: Task

mike-plummer:
- Closes #21860 , #21862 

### User facing changelog
* Copy button is removed in situations where the content has already been written to disk and doesn't need to be acted upon
* Ensure copy button remains within bounds when content becomes scrollable

### Additional details
* Copy button has been removed in the following locations:
  - Scaffolding cy config files
  - Final step of spec generator

### Steps to test
1. Open a new project in Cypress (no config file present)
2. Configure e2e testing type, scaffolding files list should open
3. Observe that file entries no longer have a "Copy" button present
4. Proceed to runner, use "+ New Space" button and create a new spec
5. Observe that the success modal no longer has a "Copy" button for spec file content

There is no longer a place in the Cypress app that exhibits the scrollable issue with the Copy button - this can be demoed via component testing under `frontend-shared > ShikiHighlight.cy.tsx`

* Copy buttons have been removed in the locations specified above

* Copy button should now remain within scroll viewport (see videos)

**Before**

https://user-images.githubusercontent.com/11482842/175563108-f5820d33-2ca8-4bf2-9c3f-302a1c2d14a8.mov

**After**
_Note:_ The copy button has been removed from this page in this PR - this video is just using it since it's a convenient place to show the scrollable-sticky behavior

https://user-images.githubusercontent.com/11482842/175563146-f8a3f39f-5d30-469d-be69-6e09f91c4605.mov

### PR Tasks


mike-plummer:
@marktnoonan After removing the Copy button from the two places you called out the scroll issue no longer manifests in the app - only other locations with a copy button don't display enough content to scroll. Adding this styling change would future-proof to an extent, but the only way I could find to get the desired behavior was to use `position: sticky` which, while pretty well supported, isn't 100%

packages/frontend-shared/src/components/ShikiHighlight.cy.tsx
```diff
@@ -85,4 +85,23 @@ describe('<ShikiHighlight/>', { viewportWidth: 800, viewportHeight: 500 }, () =>
...
+    it('should render when specified "true"', () => {
+      cy.mount(() => <div class="p-12"><ShikiHighlight code={code} lang="ts" copyButton={true} /></div>)
+      cy.get('button').should('be.visible')
+      cy.percySnapshot()
+    })
+
+    it('should remain visible when content becomes scrollable', () => {
+      cy.mount(() => <div class="p-12"><ShikiHighlight code={code.repeat(5)} lang="ts" copyButton={true} /></div>)
+      cy.get('button').should('be.visible')
+      cy.scrollTo('bottom', { duration: 500 })
```
ZachJW34, line 104: Should we have another assertion of the button visibility after scroll?

packages/frontend-shared/src/components/ShikiHighlight.vue
```diff
@@ -64,7 +64,7 @@ shikiWrapperClasses computed property.
       v-if="copyButton"
       variant="outline"
       tabindex="-1"
-      class="bg-white absolute"
+      class="bg-white ml-auto -mt-[32px] sticky margin-top:"
```
ZachJW34, line 67: No need for brackets, can just do `-mt-32px`. Also, not sure what `margin-top:` is used for.

packages/launchpad/src/setup/ScaffoldedFiles.vue
```diff
@@ -12,6 +12,7 @@
       :file-path="file.file.relative"
       :file-extension="file.file.fileExtension"
       :description="file.description || undefined"
+      :copy-button="false"
```
ZachJW34, line 15: Defaults to false so this can be removed

packages/frontend-shared/src/components/ShikiHighlight.cy.tsx
```diff
@@ -85,4 +85,23 @@ describe('<ShikiHighlight/>', { viewportWidth: 800, viewportHeight: 500 }, () =>
...
+    })
+
+    it('should render when specified "true"', () => {
+      cy.mount(() => <div class="p-12"><ShikiHighlight code={code} lang="ts" copyButton={true} /></div>)
+      cy.get('button').should('be.visible')
+      cy.percySnapshot()
+    })
+
+    it('should remain visible when content becomes scrollable', () => {
+      cy.mount(() => <div class="p-12"><ShikiHighlight code={code.repeat(5)} lang="ts" copyButton={true} /></div>)
```
ZachJW34, line 102: Would be good to throw a `max-height` on the wrapper to properly showcase this change. Since there is no height set, it is a full page scroll and the button location changes as a result.

ZachJW34 (approved):
Some minor comments, going to approve but would like to see them addressed.

packages/frontend-shared/src/components/ShikiHighlight.vue
mike-plummer: Good catch, apparently I got ahead of myself with the copy-pasta

packages/frontend-shared/src/components/ShikiHighlight.cy.tsx
marktnoonan: This test doesn't do what you'd hope. The visibility check doesn't actually account for scroll position in most cases. It passes even if you remove the `sticky` class since the element is "visible" in the sense that user could interact with it.

This comment here explains: https://github.com/cypress-io/cypress/issues/877#issuecomment-342187174 and there are some suggestions around how to test this.

It's a really unintuitive part of Cypress. Here's where I tested something similar if you want a reference (we didn't merge this branch but it's still out there), maybe this could be abstracted into a command: https://github.com/cypress-io/cypress/blob/f38e5741cd0bbac484bd408835c85a445bb721a9/packages/app/src/specs/InlineSpecList.cy.tsx#L51-L61

Ideally we'd have a test that fails on the old code and passed on the new code.

packages/launchpad/src/components/code/FileRow.vue
```diff
@@ -54,7 +54,7 @@
       :code="content"
       line-numbers
       copy-on-click
-      copy-button
+      :copy-button="copyButton"
```
marktnoonan, line 57: How about we don't set this at all? There's no instance of FileRow that's receiving the prop, is it something FileRow needs to know how to handle? I might just not be seeing it.

marktnoonan (changes requested):
Looks good! A couple of small changes requested.

packages/frontend-shared/src/components/ShikiHighlight.cy.tsx
ZachJW34: You called me out previously for this and I forgot lol

mike-plummer: Good callout, Mark. This is definitely a hairy problem once you start thinking about situations with nested scrollpanes it gets really tricky to determine what's truly "visible" since an element may be outside its immediate scroll container but within the bounds of the window. It would be nice to come up with a command that handles a more general case like that so I may take a stab at that, but your logic would handle this particular scenario since we don't have a nested scrollpane here so that will be my fallback

marktnoonan: There's also the option I forgot to mention which is a Percy snapshot at both scroll positions, that would catch this and would be nice to have in here anyway.

It is good to have a Cypress test fail as well, so I don't see why we can't do both 👍

packages/frontend-shared/cypress/support/component.ts
```diff
@@ -11,13 +11,67 @@ declare global {
...
+    const isBoundedBottom = elementBounds.top >= scrollBounds.top
+
+    return isBoundedLeft && isBoundedRight && isBoundedTop && isBoundedBottom
+  }
+
+  function _validateWithinViewport () {
+    /* Naive implementation - does not consider nested scroll containers, only checks root container bounds */
+    cy.document({ log: false }).then((doc) => {
+      // Build rectangle reflecting bounds of root scroll container (or body if not scrollable)
+      const rootViewport = doc.scrollingElement ?? doc.body
```
marktnoonan, line 53: I suppose the next iteration of this command and receive a selector as a parameter for the parent scroll pane and then we could default to this behavior.

tbiethman (approved):
Good thinking with the custom command, seems very useful!

PR #22513, opened 2022-06-24
chore: improve scroll performance of specs lists

marktnoonan:
- Closes 

### User facing changelog

Performance improvement in scroll behavior of virtualized specs lists.

### Additional details

This PR improves the scroll performance of the main specs list as well as the inline specs list by inlining components where possible. Mostly this meant inlining SVGs instead of using the icon system that componentizes them. 

One other change is that, unless there is actually a search term, we no longer render the HightlightText component to show spec and folder names, we just render the names themselves, since no text needs to be highlighted. On my 2019 MacBook Pro, this seems to give Vue enough time to render the rows as the scroll is coming to stop, especially when the list is unfiltered, which means it will be at its longest.

This PR also lifts the AverageDuration component up to the specs list level, since it is small, and doesn't need its own lifecycle.

Since this is working around a bug in chrome that seems worse when scrolling with the mousewheel or trackpad, and the mouse cursor hovering over the rows, we now ignore pointer events on the specs list while it is being scrolled. This seems to provide some extra benefit, and at worst doesn't cause any harm.

**Alternatives considered**

- Delaying rendering the more expensive components until scrolling has fully stopped
  - This was OK for scrolling but the UX seemed worse. Still an option.
 - Lifting up more components to the Specs List level. 
   - This would provide more benefits for perf but add lots more complexity to an already long component, and since I think perf is OK as it is, the extra complexity could make it harder to debug things in this area.
 

New behavior - after this PR:

https://user-images.githubusercontent.com/8340719/175573689-b0bdf392-346e-4e30-aba9-8b20d3b15e3c.mov

Old behavior - before this PR:

https://user-images.githubusercontent.com/8340719/175585457-848d2e51-be8e-4b4c-a217-88c01f950061.mov

### PR Tasks


packages/app/src/specs/SpecsList.vue
```diff
@@ -174,7 +178,7 @@
 
           <template #git-info>
             <SpecListGitInfo
-              v-if="row.data.isLeaf && row.data.data?.gitInfo"
+              v-if=" row.data.isLeaf && row.data.data?.gitInfo"
```
pstakoun, line 181: Intentional space here?

pstakoun:
@baus @rockhold how would this affect our code freeze if it goes in? Do we want to point it to develop after we get the current feature branch reviewed?

baus:
@pstakoun We could probably merge the branch in first and then follow-up with this PR into develop. I'll let you make the call on how important you think this is for the release

packages/app/src/specs/DirectoryItem.vue
```diff
@@ -3,11 +3,46 @@
     :title="props.name"
     class="flex text-sm py-4px items-center"
   >
-    <i-cy-chevron-down-small_x16
+    <svg
```
lmiller1990, line 6: Why does inlining these SVGs improve performance? This doesn't seem correct to me. The Vue compiler can tell when a component is static and should be able to make this optimisation for us under the hood.

Inlining these seems bad for maintainability - it's not clear where this icon came from, or how it was generated. 

As an aside, I have no idea how our icons work - where is `i-cy-chevron-down-small_x16` and how *is* it generated?

Also, we should be able to add a `v-once` to hint to Vue that this component is static. https://vuejs.org/api/built-in-directives.html#v-once. This might have the same impact as inlining the SVG.

packages/app/src/specs/RowDirectory.vue
```diff
@@ -4,37 +4,75 @@
...
       class="text-gray-600 truncate"
     >
       <HighlightedText
+        v-if="props.highlightText"
         :text="props.name"
         :indexes="props.indexes"
         class="font-medium"
         highlight-classes="text-gray-1000"
       />
+      <span v-else>
```
lmiller1990, line 59: This seems like the real problem, not the SVG/icons. Can you estimate the impact of these two optimizations?

Also, if we are making two optimizations, would 2 PRs be better? I'm almost sure that one of these is doing the majority of the performance improvements, and I think the SVG change is going to be a massive headache in the future (inline SVG is basically write-only, no way to update/change/associate with a design system).

packages/app/src/specs/DirectoryItem.vue
lmiller1990: I know the SFC compiler quite well and I'd like to dig into this more. We hoist static components, so the icons we generate should be as optimised as an inline SVG under the hood. 

I'll need to learn a little more about our icon system. I think the real bug is in there, not in Vue - I have a feeling it's generating components that are not very optimised. 

Do you know much about our icons? Can we sync on this?

lmiller1990 (commented):
Some comments/questions

I think we need to be cautious about some optimizations, mainly the SVG ones which introduce a lot of technical debt. We should also make sure we are actually measuring the impact of 2 separate optimisations, instead of bundling them together (2 PRs, imo)..

packages/app/src/specs/RowDirectory.vue
marktnoonan: Hey, thanks for reviewing! Yes let's sync (maybe at the start of your day tomorrow).

The `<HighlightText>` changes weren't that impactful on their own. It removes 2 component instances per row. My understanding is that it's not really about the number of elements the component renders, as much as the overhead incurred by having the component instances exist at all: https://vuejs.org/guide/best-practices/performance.html#avoid-unnecessary-component-abstractions - even totally renderless components would incur this cost (per the docs: I haven't measured).

Inlining the SVGs dramatically cuts down on the number of component instances per row, so cumulatively the impact seemed larger for the Highlight Text removals. But my measurement of that is imperfect - it's "how does the scrolling feel to me as I make these changes".

This branch was pretty much a spike to see if anything could be done about Chrome 102, since we have lots of users who will be seeing the UI presented that way. Here's a few thoughts on the debt side:

1. Yes this is debt and less maintainable than using the icon system
2. This in theory could be scheduled to reverted after Chrome 103 is stable and users are using it. Though even Electron 100 was a bit janky for me here, so we might want to keep some of the optimizations. Would need more data/testing to decide.
3. If we really want the virtualized list and will extend the features even further, long term we probably need to be mindful of the number of component instances that are getting mounted in general. We may want to use an SVG sprite setup, not the main icon system, in virtualized lists to keep them as lean as possible - that's probably what a mature version of this spike would look like.
4. We'll hopefully revisit all of this soon anyway for a11y, so there's another opportunity to unwind this.
5. Ultimately, I think we should be filtering and sorting on the server, and possibly doing traditional pagination on the front end, not virtualizing a list of all specs. This would solve, and future-proof, a lot of what's going on here. And then the code would have less need to consider the instances, since rendering time would be less critical no matter what the arch is.

marktnoonan:
Closing as Chrome 103 is stable now and solves the worst of this, we will regroup soon to consider the approach to the specs list more generally.

Issue #22389, opened 2022-06-17
Cannot re-load invalid project configuration after re-trying once


astone123:
### Current behavior

When a project is loaded into the launcher with an invalid configuration, and the "Try Again" button has been clicked once, the launcher will not clear the error once the project configuration has been corrected.

The video below shows the behavior

https://www.loom.com/share/65bd4e88a93b47c6bd96bcd4f72fdcb4

### Desired behavior

When my project is loaded with an invalid configuration and then I correct the configuration, when I click the "Try Again" button, I should see the "Welcome to Cypress" screen and be able to select a testing type.

### Test code to reproduce

Steps to reproduce:
1. Provide an invalid project configuration
2. Open the project in Cypress
3. Click "Try Again" button on error message screen (while configuration is still invalid)
4. Correct project configuration
5. Click "Try Again"
6. Notice that even though your configuration is now valid, you can't run tests unless you leave the project or close the Cypress app

### Cypress Version

10.1.0

PR #22514, opened 2022-06-24
fix: Correctly catch and clear errors during resetErrorAndLoadConfig mutation

### LINKED ISSUE #21785: Pressing 'Try again' button with config error appears to do nothing
It’s clear that “Try again” does work. But there’s no indication that it’s “trying” and that if it fails again, it has done so after trying. It just appears to do nothing.

Slack thread: 

[https://cypressio.slack.com/archives/C02MW5QBD4K/p1649088983846219](https://cypressio.slack.com/archives/C02MW5QBD4K/p1649088983846219|smart-card)

┆Issue is synchronized with this [Jira Bug](https://cypress-io.atlassian.net/browse/UNIFY-1445) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/11901">Screen Shot 2022-04-26 at 4.31.08 PM.png</a>
┆author: Chris Breiding
┆epic: UI/UX Improvements
┆friendlyId: UNIFY-1445
┆priority: Medium
┆sprint: Fast Follows 1
┆taskType: Bug

astone123:
- Closes #21785 
- https://github.com/cypress-io/cypress/issues/22389

### User facing changelog

Fixes config re-load issue where the config error message wasn't clearing correctly

### Additional details

The user could not re-load a project configuration after they had clicked the "try again" button while their configuration was still invalid. This was because of two different issues:

- The `resetErrorAndLoadConfig` mutation wasn't catching errors that were thrown in `refreshLifecycle`
- The base error that resulted from the above issue above never got cleared on reset because of a syntax error

See https://github.com/cypress-io/cypress/issues/22389 for more info

### Steps to test

I tested this by introducing a configuration error in my project config and then opening the project in Cypress. You'll be presented with an error page that has the "Try Again" button. 

- Click the try again button while the config is still invalid
- Verify that the error title does not disappear from the page
- Fix the configuration
- Click the try again button again
- Verify that the error clears and you are taken to the "Welcome to Cypress" screen

### PR Tasks


astone123:
Looking for some guidance on how to add at test for this... I have some changes out but I haven't been able to get one to work for a component test. The issue is that the disabled state comes and goes so fast that by the time I make an assertion, the state is already gone.

I tried stubbing the retry function to return a promise that waits a few seconds, but that didn't seem to work either. Maybe we don't need a test for something this small? Let me know what you think

packages/frontend-shared/src/gql-components/error/BaseError.vue
```diff
@@ -148,6 +149,13 @@ const markdownTarget = ref()
 const baseError = computed(() => props.gql)
 const { markdown } = useMarkdown(markdownTarget, computed(() => props.gql.errorMessage), { classes: { code: ['bg-error-200'] } })
 
+let retrying = false
```
mike-plummer, line 152: This should probably be a `ref`, otherwise you won't get any reactivity out of it and the state will get wiped out on rerender

```
const retrying = ref(false)
...
retrying.value = true
```

packages/frontend-shared/src/gql-components/error/BaseError.vue
```diff
@@ -23,8 +23,9 @@
             variant="outline"
             data-testid="error-retry-button"
             :prefix-icon="RestartIcon"
-            prefix-icon-class="icon-dark-indigo-500"
-            @click="retry?.(baseError.id)"
+            :prefix-icon-class="`${retrying && 'animate-spin'} icon-dark-indigo-500`"
+            :disabled="retrying"
+            @click="retryFn()"
```
mike-plummer, line 28: Totally a personal preference, but I prefer to name my event handlers with an `on*` convention so they're more easily recognizable/extensible. In this case, something like `onRetry` or `onRetryClick`

packages/frontend-shared/src/gql-components/error/BaseError.vue
astone123: Good call 👍🏻 still learning about Vue and reactivity stuff

packages/frontend-shared/src/gql-components/error/BaseError.vue
```diff
@@ -148,6 +149,14 @@ const markdownTarget = ref()
 const baseError = computed(() => props.gql)
 const { markdown } = useMarkdown(markdownTarget, computed(() => props.gql.errorMessage), { classes: { code: ['bg-error-200'] } })
 
+const retrying = ref(false)
+
+const onRetry = async () => {
+  retrying.value = true
+  await props.retry?.(baseError.value.id)
```
ZachJW34, line 156: Change the type of props.retry to `(id: string) => Promise<void>`

ZachJW34, line 156: Change the type of props.retry to `(id: string) => Promise<void>`

ZachJW34 (changes requested):
You can test this out by doing:

```ts
it('shows loading state when clicking "Try again"', () => {
    let _res: any

    cy.mountFragment(BaseErrorFragmentDoc, {
      render: (gqlVal) => (
        <BaseError gql={gqlVal} retry={() => new Promise((res) => _res = res)} />),
    })

    cy.contains('button', 'Try again').click().should('be.disabled').within(() => {
      cy.get('svg').should('have.class', 'animate-spin')
    })

    cy.then(() => _res())
    cy.contains('button', 'Try again').should('not.be.disabled')
  })
```

I wouldn't take a percy snapshot of this since it would be flaky with the animation.

packages/frontend-shared/src/gql-components/error/BaseError.vue
```diff
@@ -23,8 +23,9 @@
             variant="outline"
             data-testid="error-retry-button"
             :prefix-icon="RestartIcon"
-            prefix-icon-class="icon-dark-indigo-500"
-            @click="retry?.(baseError.id)"
+            :prefix-icon-class="`${retrying && 'animate-spin'} icon-dark-indigo-500`"
```
lmiller1990, line 26: I think it's more common to use the syntax made popular by the `classes` library, often seen in React projects, using the object syntax. If the value is `true`, the class is rendered. Docs: https://vuejs.org/guide/essentials/class-and-style.html#binding-html-classes

```js
:prefix-icon-class="{ 'animate-spin': retrying, icon-dark-indigo-500` }"
```
It looks like `BaseError` would need to be updated:

```js
const props = defineProps<{
  prefixIconClass?: string | Record<string, string | boolean>
}>()
```

Seems like a quick win.

packages/frontend-shared/src/gql-components/error/BaseError.vue
lmiller1990: I just noticed we pass a function as a prop and handle it. This is unusual - normally you'd do `emit("retry")` then receive the result as a prop, instead of awaiting a no-op (assuming `Promise<void>` is actually the correct return type). 

Edit: this will not always be a promise: https://github.com/cypress-io/cypress/pull/22514/files#diff-b580358883f946c09563d80b0301fc0e521b5ff8435513cc5abb9c671f655dccR156

I know this wasn't introduced here, but we should probably change it to emit and event... but not a blocker.

lmiller1990: How about this:

1. define new emit, something like `(event: 'retry', id: string): void`
2. `emit("retry", baseError.value.id)` here
3. in the parent, listen for `@retry="resetErrorAndLoadConfig"
4. In there, do something like:

```js
const retrying = ref(false)
const resetErrorAndLoadConfig = async (id: string) => {
  retrying.value = true
  if (!mutation.fetching.value) {
    await mutation.executeMutation({ id }) // do we care about result of this? I don't think so
  }
  retrying.value = false
}
```

5. Pass `retrying` prop to `BaseError`.

We'd be lifting the state from BaseError into the parent (as they call it in the [React docs](https://reactjs.org/docs/lifting-state-up.html). I wonder if BaseError should be the one to decide if it's errored, or it should be a "dumb" or "presentation" component only?

Thoughts @astone123 @ZachJW34 ? Bit of work, but I think it's good to start attacking some of the debt we introduced while learning Vue and building the unification project.

lmiller1990 (commented):
Cool - functionality seems good, I think this might be a chance to reduce some of the debt incurred (mainly around non idiomatic Vue code) introduced over the last little while. Thoughts?

Also, wonder if we can get a test for this. The refactor I suggested would let us easily unit test the BaseError as a dumb/presentational component, and e2e test the whole workflow.

packages/launchpad/src/Main.vue
```diff
@@ -150,8 +150,10 @@ const mutation = useMutation(Main_ResetErrorsAndLoadConfigDocument)
 
 const resetErrorAndLoadConfig = (id: string) => {
   if (!mutation.fetching.value) {
-    mutation.executeMutation({ id })
+    return mutation.executeMutation({ id })
   }
+
+  return
 }
 const query = useQuery({ query: MainLaunchpadQueryDocument })
```
lmiller1990, line 158: Cool - functionality seems good, I think this might be a chance to reduce some of the debt incurred (mainly around non idiomatic Vue code) introduced over the last little while. Thoughts?

Also, wonder if we can get a test for this. The refactor I suggested would let us easily unit test the BaseError as a dumb/presentational component, and e2e test the whole workflow.

We should be able to test this, maybe we can pair on it? I will think about some testing strategies in the meantime.

packages/frontend-shared/src/gql-components/error/BaseError.vue
astone123: @lmiller1990 I'm definitely up for making this more idiomatic. With React, I'm used to passing functions down from parent to child to handle state changes like this, but I agree that it would be better to "lift state" into the parent (talking about the `retrying` state).

Will work on this

ZachJW34: The emit makes more sense here so I think it's a good idea and worth implementing. Can we remove the extra `retrying` ref and just use the value of `mutation.fetching.value` as our loading state?

packages/app/src/layouts/default.vue
```diff
@@ -24,7 +24,8 @@
       <BaseError
         v-if="query.data.value?.baseError"
         :gql="query.data.value.baseError"
-        :retry="resetErrorAndLoadConfig"
+        :retrying="retrying"
```
ZachJW34, line 27: It would be better to use `:retrying="mutation.loading.value"` here. If the button wasn't disabled, you could run into a problem with double clicking and resetting the loading state. Not a big deal since it is disabled, but we could get rid of the extra ref tracking the loading state.

ZachJW34 (approved):
One small comment, not blocking. Good refactor!

packages/app/src/layouts/default.vue
lmiller1990: Good point - if we can drop `retrying` which seems to be 1:1 to `loading.value` that would be nice.

packages/frontend-shared/src/gql-components/error/BaseError.cy.tsx
```diff
@@ -69,18 +69,28 @@ describe('<BaseError />', () => {
       .should('have.attr', 'href', docsButton.configGuide.link)
     })
 
-    it('calls the retry function passed in', () => {
+    it('emits a \'retry\' event when clicked', () => {
```
lmiller1990, line 72: Nit, I like to just use backticks instead of escaping with `\` but either or is fine.

lmiller1990 (commented):
Nice, just need to address the last commit around the `retrying` ref (do we need it) and we can sync on testing (haven't really thought about how to do this yet, but shouldn't be too hard).

astone123:
I want to also fix #22389 in this PR because I fixed part of it, and it should make this issue unnecessary. Currently trying to figure out why the error title disappears after trying again the first time

lmiller1990:
Should we move this to draft until it's ready for re-review @astone123 ?

astone123:
@lmiller1990 yeah, I wasn't sure how to move it back to draft

lmiller1990:
@astone123 You can click "move to draft" under the reviewers:

image

This button is hard to find for some reason, I had the same experience when they introduced this feature.

packages/data-context/src/actions/ErrorActions.ts
```diff
@@ -22,7 +22,7 @@ export class ErrorActions {
       }
 
       if (d.baseError?.id === id) {
-        d.baseError === null
+        d.baseError = null
```
astone123, line 25: This is why the base error wasn't getting cleared even after calling `resetErrorAndLoadConfig` 😅

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
```diff
@@ -33,7 +33,7 @@ export const mutation = mutationType({
       },
       resolve: async (_, args, ctx) => {
         ctx.actions.error.clearError(args.id)
-        await ctx.lifecycleManager.refreshLifecycle()
+        await ctx.lifecycleManager.refreshLifecycle().catch(ctx.lifecycleManager.onLoadError)
```
astone123, line 36: This is why the config error was getting set to base error rather than project error (and why the title was disappearing). The first time around, `lifecycleManager.onLoadError` was called when the configuration error was thrown.

However, the second time around (after clicking "try again"), it was called here, but wasn't caught so it went up to base error with no title. Now that we're catching it here, it stays at the current project level and has the correct title.

packages/launchpad/cypress/e2e/config-files-error-handling.cy.ts
```diff
@@ -147,8 +147,31 @@ describe('Launchpad: Error System Tests', () => {
...
+
+    cy.get('h1').should('contain', 'Welcome to Cypress')
+  })
+
+  it(`clears the error correctly after first 'try again' attempt`, () => {
+    cy.scaffoldProject('config-with-ts-syntax-error')
+    cy.openProject('config-with-ts-syntax-error')
+    cy.visitLaunchpad()
+    cy.contains('h1', cy.i18n.launchpadErrors.generic.configErrorTitle)
+    cy.percySnapshot()
```
ZachJW34, line 165:
```suggestion
    cy.contains('h1', cy.i18n.launchpadErrors.generic.configErrorTitle)
```
We can remove this percySnapshot, it's not covering any more than the test above.

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
lmiller1990: Nice job tracking this down.

packages/data-context/src/actions/ErrorActions.ts
lmiller1990: :facepalm: 

I wonder if there's an eslint rule for this. If not, writing one would be an interesting side project.

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -797,7 +797,7 @@ export class ProjectLifecycleManager {
    * centrally in the e2e tests, as well as notify the "pending initialization"
    * for run mode
    */
-  private onLoadError = (err: any) => {
+  onLoadError = (err: any) => {
```
lmiller1990, line 800: Can we not type this as `err: Error`? (or maybe there' a `CypressError` type, can't quite remember).

lmiller1990 (approved):
Two minor nits, going to tentatively approve

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
astone123: I have @ZachJW34 to thank for helping me track it

PR #22530, opened 2022-06-27
chore: refactor server/lib/config.ts to pkg/config

### LINKED ISSUE #21781: Refactor config logic in server to the config package
## **Summary**
There are several functions that still remain in https://github.com/cypress-io/cypress/blob/10.0-release/packages/server/lib/config.ts that could potentially moved to the config package. This will allow us to use those functions directly rather than having a `configApi` in DataContext and keep all of the config logic in one place

┆Issue is synchronized with this [Jira Task](https://cypress-io.atlassian.net/browse/UNIFY-1431) by [Unito](https://www.unito.io)
┆author: Ryan Manuel
┆epic: Chores
┆friendlyId: UNIFY-1431
┆priority: Medium
┆taskType: Task

sainthkh:
- Part 1 of #21781

### User facing changelog

N/A. Internal refactoring.

### Additional details
- Why was this change necessary? => Having them in `pkg/server` makes us create `configApi` inside `DataContext` and not be able to call them directly. 
- What is affected by this change? => None. It's a just code cleanup.

At first, I thought this issue would be a simple copy and paste of a few lines of code. But I learned that it's a lot bigger than I thought. 

To refactor `server/lib/config.ts`, I had to refactor these files, too:

* server/lib/util/config.ts
* server/lib/util/coerce.ts
* server/lib/util/keys.ts
* server/lib/util/origin.ts
* server/lib/util/path_helpers.ts

Because they're used mostly inside `config.ts` and related with manipulating config values. 

----

As we're migrating our code from `commonjs` to `esm`, the code structure also had to be fixed. 

For this PR, I created 3 files to place the code.

* config/src/project/index.ts -> The public api or the project config functions used outside the `config` package. I named them project because the type definitions were in `ProjectConfigManager`, `ProjectLifecycleManager`.
* config/src/project/util.ts -> The internal functions to be used to implement the public api at `index.ts`. They are `export`ed to be unit-tested. 
* config/src/util.ts -> Some utility functions like `hideKeys` and `coerce` are used outside `pkg/config`. They live here. 

As for `origin.ts`, I moved it to `pkg/network/lib/uri.ts` because it more suits the place.

----

I renamed some functions to make it clear.

* `keys` -> `hideKeys`
* `isDefault` -> `isResolvedConfigPropDefault`

----

I added types for the functions. I added some `any` because I'm not sure about them now. 

### Steps to test

No new features; ensure existing tests are passing.

N/A. It's a code cleanup.

### PR Tasks


packages/config/src/project/utils.ts
```diff
@@ -0,0 +1,342 @@
+import Bluebird from 'bluebird'
+import Debug from 'debug'
+import fs from 'fs-extra'
+import _ from 'lodash'
+import path from 'path'
+
+import { getProcessEnvVars, CYPRESS_SPECIAL_ENV_VARS } from '@packages/server/lib/util/config'
```
emilyrohrbough, line 7: Any way we can drop pulling in `@packages/server` or `@packages/data-context`?

sainthkh: I'm trying. I guess we can remove `@package/server` from here.

emilyrohrbough: So excited to see these changes!

packages/config/src/project/index.ts
```diff
@@ -0,0 +1,134 @@
+import _ from 'lodash'
+import Debug from 'debug'
+import deepDiff from 'return-deep-diff'
+
+import errors, { ConfigValidationFailureInfo, CypressError } from '@packages/errors'
+import { getCtx } from '@packages/data-context/src/globalContext'
```
sainthkh, line 6: This dependency cannot be removed until we move [`getFilesByGlob `](https://github.com/cypress-io/cypress/blob/1b688694fe7ad7de7ce4ac215024cd128d68bf39/packages/data-context/src/sources/FileDataSource.ts#L31-L57) to somewhere else.

Or call it inside `data-context` and pass the value to this function. 

Maybe, this can be solved in the future PRs.

sainthkh: I researched it after writing this comment. It seems that it will work if we add `supportFiles` member to `setupFullConfigWithDefaults`.

I'll do this after this PR is merged. This PR is big and I tried not to change code much.

packages/config/src/project/index.ts
```diff
@@ -0,0 +1,134 @@
...
+  validateNoBreakingConfig,
+} from '../browser'
+import {
+  setPluginResolvedOn,
+  mergeDefaults,
+} from './utils'
+
+const debug = Debug('cypress:config:project')
+
+// TODO: any -> SetupFullConfigOptions in data-context/src/data/ProjectConfigManager.ts
```
sainthkh, line 22: Using `data-context/src/data/ProjectConfigManager.ts` type definitions breaks the type. It seems that some types are missing. I postponed it for now because I need to read a lot of code to be sure that the new types are correct.

packages/data-context/src/globalContext.ts
```diff
@@ -0,0 +1,49 @@
+import type { DataContext } from './DataContext'
```
sainthkh, line 1: I created this file to bypass circular dependency.

packages/network/lib/uri.ts
```diff
@@ -10,24 +10,24 @@ import url, { URL } from 'url'
...
 
-const portIsDefault = (port) => {
+const portIsDefault = (port: string | null) => {
   return port && DEFAULT_PORTS.includes(port)
 }
 
-const parseClone = (urlObject) => {
+const parseClone = (urlObject: any) => {
   return url.parse(_.clone(urlObject))
 }
```
sainthkh, line 26: This is really weird code. [`url.parse` expects `string` as its first arg](https://nodejs.org/api/url.html#urlparseurlstring-parsequerystring-slashesdenotehost), but our code is written for the case that it's not a `string` but an object.

I'll check this behavior and fix this in the part 2 of this PR.

packages/server/lib/config.ts
```diff
@@ -1,597 +1,12 @@
...
-import { fs } from './util/fs'
-import keys from './util/keys'
-import origin from './util/origin'
-import pathHelpers from './util/path_helpers'
 
-import type { ConfigValidationFailureInfo, CypressError } from '@packages/errors'
+export const setupFullConfigWithDefaults = configUtils.setupFullConfigWithDefaults
 
-import { getCtx } from './makeDataContext'
+export const updateWithPluginValues = configUtils.updateWithPluginValues
```
sainthkh, line 7: These are here because they're used in some tests in `pkg/server`

packages/server/test/unit/config_spec.js
```diff
@@ -26,40 +19,6 @@ describe('lib/config', () => {
...
-      }
-      const options = {}
-
-      await config.mergeDefaults(cfg, options)
-
-      expect(errors.throwErr).not.to.be.called
-    })
-  })
-
   context('.get', () => {
```
sainthkh, line 22: The tests under `.get` more suit `pkg/data-context/test/unit/data/ProjectLifecycleManager`. 

If you agree, I'll move them in the next PR.

packages/ts/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     /* Basic Options */
     "target": "ES2018",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', or 'ESNEXT'. */
     "module": "commonjs",                     /* Specify module code generation: 'commonjs', 'amd', 'system', 'umd' or 'es2015'. */
-    "lib": ["es2018", "ES2020.Promise"],                             /* Specify library files to be included in the compilation:  */
+    "lib": ["es2018", "ES2020.Promise", "ES2021.String"],                             /* Specify library files to be included in the compilation:  */
```
sainthkh, line 6: `String.ReplaceAll` is used in several pacakges. So, I added it here to avoid surprise.

packages/config/src/project/index.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  debug('config is %o', config)
+
+  // flatten the object's properties into the master config object
+  config.envFile = envFile
+  config.projectRoot = projectRoot
+  config.projectName = projectName
+
+  // @ts-ignore
+  return mergeDefaults(config, options, cliConfig)
```
lmiller1990, line 41: How come we need to ts-ignore here?

packages/network/lib/uri.ts
```diff
@@ -10,24 +10,24 @@ import url, { URL } from 'url'
 
 // yup, protocol contains a: ':' colon
 // at the end of it (-______________-)
-const DEFAULT_PROTOCOL_PORTS = {
+const DEFAULT_PROTOCOL_PORTS: Record<string, string> = {
```
lmiller1990, line 13: Seems this is never going to change - instead of  `Record<string, string>` we could do `as const` and get better type safety - that way, only `http` and `https` would be valid. `DEFAULT_PROTOCOL_PORTS.foo` would raise a typing error.

packages/network/lib/uri.ts
```diff
@@ -10,24 +10,24 @@ import url, { URL } from 'url'
...
   'https:': '443',
   'http:': '80',
 }
 
 const DEFAULT_PORTS = _.values(DEFAULT_PROTOCOL_PORTS)
 
-const portIsDefault = (port) => {
+const portIsDefault = (port: string | null) => {
```
lmiller1990, line 20: Could also improve typing here - `port: '80' | '443'` are the only valid ports, right?

lmiller1990: Not sure how easy/hard these types are to add - it might make other places in the code base complain.

lmiller1990 (approved):
Glad to see the slow march from CJS+JS to TS, and gradual improvements to type safety. This generally looks like a pretty straight forward refactor - I don't see any obvious problems, just made a few recommendations around (potential) type safety improvements.

packages/config/src/project/index.ts
sainthkh: It's because of the many errors inside and outside `mergeDefaults` like these:

image

image

I'll try to fix this in the subsequent PR.

packages/network/lib/uri.ts
sainthkh: Done.

packages/network/lib/uri.ts
sainthkh: It seems that current behavior is correct because the purpose of this function is to check if the url uses the default port or **not**. We can use other ports like `3000`, `8080` for internal sites

packages/config/src/project/index.ts
lmiller1990: Alright, I figured it was something like this.

packages/server/lib/util/args.js
```diff
@@ -139,6 +138,7 @@ const tryJSONParse = (str) => {
 }
 
 const JSONOrCoerce = (str) => {
+  coerce
```
emilyrohrbough, line 141: typo?

packages/data-context/src/globalContext.ts
emilyrohrbough: Where were you seeing this at?

emilyrohrbough (changes requested):

packages/server/package.json
```diff
@@ -102,7 +102,6 @@
     "randomstring": "1.1.5",
     "recast": "0.20.4",
     "resolve": "1.17.0",
-    "return-deep-diff": "0.4.0",
     "sanitize-filename": "1.6.3",
```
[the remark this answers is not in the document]
sainthkh: No, we can't. Many things are imported from `@packages/config` at `lib/cypress.js`, `lib/scaffold.js`, etc.

packages/server/lib/util/args.js
sainthkh: Thanks. Removed it.

packages/config/src/project/index.ts
```diff
@@ -0,0 +1,134 @@
+import _ from 'lodash'
+import Debug from 'debug'
+import deepDiff from 'return-deep-diff'
+
+import errors, { ConfigValidationFailureInfo, CypressError } from '@packages/errors'
```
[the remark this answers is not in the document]
sainthkh: Done

packages/config/src/project/index.ts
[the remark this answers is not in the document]
sainthkh: `getCtx` is called in 2 functions:

1. `updateWithPluginValues` of `project/index.ts`
2. `setSupportFileAndFolder` of `project/utils.ts`

The solution for the first case is simple. We can simply add `configFile` member inside `buildBaseFullConfig` of `ProjectConfigManager`.

But the problem is the second one. 

To call the `ctx.file.getFilesByGlob`, we need `obj.supportFile`. But the problem is that global `supportFile` option is legacy. It is splitted and moved under `e2e` or `component` from 10. 

It means that current `setSupportFileAndFolder` can be buggy or completely removable in the current version. I'm not 100% sure about this. I need to think about a test case that shows this problem. Maybe #22776 is the related case.

The goal of this PR is copying and moving `server/lib/config.ts` to `pkg/config` and adding types as much as possible. I focused on this because this PR can be too big and it makes reviewers hard to review this PR. 

I think checking/fixing bugs and adding more specific types should be done in the future PRs. What do you think?

packages/data-context/src/globalContext.ts
sainthkh: `pkg/config` uses `getCtx` and `pkg/data-context` uses `pkg/config` functions. It causes circular dependency and breaks Cypress. 

This file could be removed and merged back to `index.ts` after fixing [the problems I mentioned in this comment](https://github.com/cypress-io/cypress/pull/22530#discussion_r927213805).

packages/config/src/project/index.ts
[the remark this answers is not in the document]
emilyrohrbough: > To call the ctx.file.getFilesByGlob, we need obj.supportFile. But the problem is that global supportFile option is legacy. It is splitted and moved under e2e or component from 10.

This log is correct, at this point we are testing the flattened configuration for the given testing type, so `obj.supportFile` is valid.

I think making a few small changes seems needed to move this code over and would be worth making a small change to remove this circular dependency. Simply passing int he attribute needed is probably the simplest approach until / if we ever want to revisit how this is composed.

```js
// data-context/ProjectConfigManger.ts
setupFullConfigWithDefaults({
  projectRoot,
  projectName,
  configFile,
  config,
  envFile,
  options,
  cliConfig
}, 
ctx.file.getFilesByGlob)
```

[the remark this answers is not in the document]
sainthkh: I never thought about passing the function as an argument to remove the dependency. Thanks for the idea. 

Unfortunately, this idea even cannot completely remove `pkg/data-context` from `config`. It can be removed from `src`. But it had to be added to `test` to run tests. 

However, I think this change is much better.

rachelruderman:
Hi @sainthkh ! Can you update `Steps to test` in the description? I realize this is a refactor PR so it may be as simple as "No new features; ensure existing tests are passing"

packages/data-context/src/globalContext.ts
emilyrohrbough: @sainthkh I think this can be removed now with your latest changes!

packages/config/test/project/utils.spec.ts
```diff
@@ -0,0 +1,1182 @@
+import '@packages/server/test/spec_helper'
+import { getCtx } from '@packages/data-context/src/globalContext'
```
sainthkh, line 2: We need `globalContext` here. Otherwise, we need to import the entire `pkg/data-context`.

packages/config/src/project/utils.ts
```diff
@@ -0,0 +1,525 @@
...
+  }
+
+  if (supportFilesByGlob.length === 0) {
+    if (obj.resolved.supportFile.from === 'default') {
+      return errors.throwErr('DEFAULT_SUPPORT_FILE_NOT_FOUND', relativeToProjectRoot(obj.projectRoot, obj.supportFile))
+    }
+
+    return errors.throwErr('SUPPORT_FILE_NOT_FOUND', relativeToProjectRoot(obj.projectRoot, obj.supportFile))
+  }
+
```
sainthkh, line 297: Return value is used to throw errors.

packages/data-context/src/globalContext.ts
sainthkh: Unfortunately, it can't. [As I wrote in the comment below,](https://github.com/cypress-io/cypress/pull/22530#discussion_r932805254) we need `getCtx` from `pkg/data-context` dependency in tests. [Check this code](https://github.com/cypress-io/cypress/pull/22530/files#r935032311)

I thought about injection dummy functions to the test. But [its return value is used to throw errors](https://github.com/cypress-io/cypress/pull/22530/files#r935034146). So, it can cause false positive when `getFilesByGlob` is changed.

I think the only way to roll back this code is to:

1. solve that [`obj.supportFile` issue](https://github.com/cypress-io/cypress/pull/22530#discussion_r927213805)
2. pass `supportFile` from `pkg/data-context`
3. remove `pkg/data-context` from tests in `pkg/config`
4. remove this `globalContext` file.

packages/config/test/project/utils.spec.ts
```diff
@@ -0,0 +1,1182 @@
...
+        supportFile: false,
+      })
+
+      return setSupportFileAndFolder(obj, getCtx().file.getFilesByGlob)
+      .then((result) => {
+        expect(result).to.eql({
+          projectRoot,
+          supportFile: false,
+        })
+      })
```
emilyrohrbough, line 550: You should be able to stub this out and we should not be pulling fixtures into this package.

```suggestion
       const getFilesByGlob = sinon.stub()
      return setSupportFileAndFolder(obj, getCtx().file.getFilesByGlob)
      .then((result) => {
        expect(result).to.eql({
          projectRoot,
          supportFile: false,
        })
        expect(getFilesByGlob).not.have.been.called
      })
```

sainthkh: Thanks, I should have thought more about the solutions. I finally removed `data-context` from `config`.

packages/data-context/src/globalContext.ts
emilyrohrbough: able to delete now with the latest commits?

emilyrohrbough (approved):

rachelruderman (changes requested):
@sainthkh looks like we have a [failing server performance test](https://app.circleci.com/pipelines/github/cypress-io/cypress/41600/workflows/6a533794-3022-4466-ba2d-1d8d5d9e42bb/jobs/1724019):
Screen Shot 2022-08-10 at 9 50 18 AM

sainthkh:
@rachelruderman For some reason, context was cleared before `before` in `proxy_performance_spec`. I solved the problem by adding new context before calling `getCtx`. I didn't create stub here because we're testing performance. Creating stub can make test good-for-nothing test result.

rachelruderman (approved):
Looks great, thank you for your contribution @sainthkh !

PR #22536, opened 2022-06-27
test: Addressing launchpad test flake in Windows

tbiethman:
This PR addresses the consistently failing spec for the launchpad's global mode tests in Windows CI. See comments for details.

### Steps to test

* Open up the launchpad, navigate to a project, and select a testing type
* Navigate back to the testing type selection
* Select a testing type again (the same one or different, doesn't matter)
* Inspect the project name link in the nav breadcrumbs and make sure there is no `role` attribute set on it (we want the implicit role on the `a` tag to take priority in this case)

### PR Tasks

### DIFF
--- a/packages/frontend-shared/src/gql-components/HeaderBarContent.vue
+++ b/packages/frontend-shared/src/gql-components/HeaderBarContent.vue
@@ -27,6 +27,7 @@
               <!-- context for use of aria role and disabled here: https://www.scottohara.me/blog/2021/05/28/disabled-links.html -->
               <!-- the `href` given here is a fake one provided for the sake of assistive technology. no actual routing is happening. -->
               <a
+                :key="hasLinkToProjects"
                 class="font-medium"
                 :class="hasLinkToProjects ? 'text-indigo-500 hocus-link-default' :
                   'text-gray-700'"
@@ -52,6 +53,7 @@
               <!-- context for use of aria role and disabled here: https://www.scottohara.me/blog/2021/05/28/disabled-links.html -->
               <!-- the `href` given here is a fake one provided for the sake of assistive technology. no actual routing is happening. -->
               <a
+                :key="hasLinkToCurrentProject"
                 class="font-medium"
                 :role="hasLinkToCurrentProject ? undefined : 'link'"
                 :href="hasLinkToCurrentProject ? 'choose-testing-type' : undefined"
--- a/packages/frontend-shared/src/gql-components/HeaderBarContent.vue
+++ b/packages/frontend-shared/src/gql-components/HeaderBarContent.vue
@@ -26,8 +26,9 @@
             <li class="inline-block">
               <!-- context for use of aria role and disabled here: https://www.scottohara.me/blog/2021/05/28/disabled-links.html -->
               <!-- the `href` given here is a fake one provided for the sake of assistive technology. no actual routing is happening. -->
+              <!-- the `key` is used to ensure the role/href attrs are added and removed appropriately from the element. -->
               <a
-                :key="hasLinkToProjects"
+                :key="Boolean(hasLinkToProjects).toString()"
                 class="font-medium"
                 :class="hasLinkToProjects ? 'text-indigo-500 hocus-link-default' :
                   'text-gray-700'"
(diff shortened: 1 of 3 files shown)


system-tests/projects/todos/cypress/support/component-index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
```
tbiethman, line 1: I noticed a bunch of compilation errors around the failing spec due to this file missing from the todos project. `todos` is configured for component testing but did not include this file. I don't think this was causing a problem in the tests outside of the chatty log, but now it's set up correctly and the log is chatty no more.

packages/launchpad/cypress/e2e/global-mode.cy.ts
```diff
@@ -152,7 +152,8 @@ describe('Launchpad: Global Mode', () => {
 
     it('updates breadcrumb when selecting a project and navigating back', () => {
       const getBreadcrumbLink = (name: string, options: { disabled: boolean } = { disabled: false }) => {
-        return cy.findByRole('link', { name }).should('have.attr', 'aria-disabled', options.disabled ? 'true' : 'false')
+        // The timeout is increased to account for variability in configuration load times in CI.
+        return cy.findByRole('link', { name, timeout: 10000 }).should('have.attr', 'aria-disabled', options.disabled ? 'true' : 'false')
```
tbiethman, line 156: In addition to the role lookup failing, the test would occasionally have to retry due to the config loading taking a bit longer than expected. So I just bumped the timeout here to give us plenty of room.

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -51,7 +53,9 @@
             <li class="inline-block">
               <!-- context for use of aria role and disabled here: https://www.scottohara.me/blog/2021/05/28/disabled-links.html -->
               <!-- the `href` given here is a fake one provided for the sake of assistive technology. no actual routing is happening. -->
+              <!-- the `key` is used to ensure the role/href attrs are added and removed appropriately from the element. -->
               <a
+                :key="Boolean(hasLinkToCurrentProject).toString()"
```
tbiethman, line 58: I noticed that the failure here was that `findByRole` couldn't find the breadcrumb links appropriately: https://dashboard.cypress.io/projects/sehy69/runs/12235/overview/83c7515d-deb1-414a-953c-b74c08865063

I could see in the recording that the links are enabled. And `findByRole` did find the 'Projects' and 'todos' elements, but it reported their role as...nothing?
```
  listitem:

  Name "":

  Name "":

  Name "":

  Name "":

  Name "":

---

  :                                 // <-- Wat

  Name "Projects":

  Name "todos":

---

  link:

  Name "v10.2.0":
```

I found this odd, so I took a look at those elements as rendered in the test and noticed that as we transitioned through the lifecycle, we would end up in a state where href/role attrs have gone from unset, to set, to unset again. However, once set, they would not be removed from the rendered element. So we would end up in a state where we have an `a` tag, with an `href` attribute, and a blank `role` set on it explicitly.

So long story short, I added a key to make sure these attributes we don't want to set actually do get removed. I tried using `null` rather than `undefined` in these cases to try and get these attributes to be completely removed from the same element instance, with no success.

And what do you know, this change fixed the test. As it turns out, our windows and linux builds are running these tests on different versions of Chrome. Windows is using Chrome 103 (the most recent build that it downloads onto machine at test time) vs. the linux build's Chrome 100 it uses from the docker image. This spec is actually failing in develop with Chrome 103, regardless of platform.

BlueWinds (approved):

rachelruderman:
What a rockstar, thank you!! Do you mind filling out `Steps to test`? Given the comments I imagine in this case it's as simple as "Run `packages/launchpad/cypress/e2e/global-mode.cy.ts` on Chrome 103 and confirm it passes`, but would be great to have in the PR description for posterity

tbiethman:
@rachelruderman Steps to test have been added!

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
marktnoonan: Excellent! And ugh. What a strange combination of events. I'm also real surprised that explicitly setting `undefined` which is also the _natural_ value for the role makes a difference, as it didn't seem to make any difference to Chrome's accessibility tree. But hurray for getting these passing.

As a side note, I'm suggesting we just [not have this link at all any more](https://github.com/cypress-io/cypress/issues/21911#issuecomment-1147674155 ). If anybody has thoughts one way or the other.

ZachJW34 (approved):

tgriesser:
Amazing

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
tbiethman: It looks like previous versions of chrome would leave the "link" value for the attribute, even when we unset it. This would allow the test to succeed, but it's not what we were wanting/expecting DOM-wise.

Chrome 103 gets a little closer and removes the role value from the element, but leaves the attribute just hanging there, like a boolean attribute: `<div role>Wat</div>`. This....is worse functionally but more indicative of what we're wanting, which is to unset the role.

I too would have figured that setting undefined would have cleared it, but I imagine this is some virtual dom weirdness where it's hard to interpret "no value" and "no attribute at all" through a single API.

Issue #22544, opened 2022-06-27
Cypress migration tool doesn't appear + stack trace in the console


kaiyoma:
### Current behavior

I'm currently following the instructions in the migration video and even the simplest steps aren't working. I upgraded from version 9 to version 10.2.0 and ran `npx cypress open`, as instructed. The migration video says that the migration helper will appear in the window, but it does not. Instead, I see the "Configuration Files" screen.

Also, there is a stack trace in the console:

```
$ npx cypress open
It looks like this is your first time using Cypress: 10.2.0

[STARTED] Task without title.
[TITLE]  Verified Cypress! C:\Users\Kyle Getz\AppData\Local\Cypress\Cache\10.2.0\Cypress
[SUCCESS]  Verified Cypress! C:\Users\Kyle Getz\AppData\Local\Cypress\Cache\10.2.0\Cypress

Opening Cypress...

node:internal/modules/cjs/loader:943
  throw err;
  ^

Error: Cannot find module 'C:\Users\Kyle'
Require stack:
- internal/preload
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:940:15)
    at Function.Module._load (node:internal/modules/cjs/loader:785:27)
    at Module.require (node:internal/modules/cjs/loader:1012:19)
    at Module._preloadModules (node:internal/modules/cjs/loader:1292:12)
    at loadPreloadModules (node:internal/bootstrap/pre_execution:486:5)
    at prepareMainThreadExecution (node:internal/bootstrap/pre_execution:77:3)
    at node:internal/main/run_main_module:7:1 {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ 'internal/preload' ]
}
```

It looks like Cypress is trying to access something in my home directory, but is failing when encountering a directory with spaces in it, which is pretty common in Windows. Was the new version of Cypress not tested on Windows at all?

### Desired behavior

No stack trace and I should see the migration tool in the window.

### Test code to reproduce

Upgrade from version 9 to version 10 on Windows.

### Cypress Version

10.2.0

lmiller1990:
I thought we had resolved the "spaces in username" issue but apparently not.

I will fix this right now and hopefully this can get into the release in the next 24 hours. I'm sorry you encountered this problem.

Dup: https://github.com/cypress-io/cypress/issues/22297#issuecomment-1167898416

lmiller1990:
Should be fixed in https://github.com/cypress-io/cypress/pull/22550

Issue #22297, opened 2022-06-14
Cypress doesn't load or open its window


singque09:
### Current behavior

Cypress won't load and it is indicating a vague cannot find module when I've upgraded to v10.1.0 it was working previously with 9.7.0 where it loads all spec lists. I did try the following:
- `cypress open -b chrome --e2e`
- `cypress open`

all of the above did was open the window but it is stuck in a loading state.

### Debug logs
```
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module 'C:\Users\Jolambert'
Require stack:
- internal/preload
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at Module._preloadModules (node:internal/modules/cjs/loader:1276:12)
    at loadPreloadModules (node:internal/bootstrap/pre_execution:483:5)
    at prepareMainThreadExecution (node:internal/bootstrap/pre_execution:77:3)
    at node:internal/main/run_main_module:7:1 {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ 'internal/preload' ]
}
```
### Cypress Version

10.1.0

### Package Manager

yarn

### Operating system

Windows

jeffreymb:
I resolved the issue by renaming my user profile directory as described in these instructions: https://www.tenforums.com/tutorials/89060-change-name-user-profile-folder-windows-10-a.html

samp-reston:
Tested on 10.0.3 and 10.0.0. Error isn't present. Only on 10.1.0. Node version v18.3.0.

singque09:
> I resolved the issue by renaming my user profile directory as described in these instructions: https://www.tenforums.com/tutorials/89060-change-name-user-profile-folder-windows-10-a.html

I think this is just a temporary fix but absolute folder pathing is working before 10.1.0.

marktnoonan:
Thanks for these details everyone. I've attempted to reproduce by opening Cypress on Windows with a project stored in `C:\Sites\empty project` where I would have expected the space in the path to cause the same problem people are reporting here. But Cypress opened fine for me in this situation.

Could I get some more information to help narrow this down - Windows 11, node 18.3.0, what else? There's nothing about 10.1.0 that stands out as a possible explanation.

[...]

PR #22550, opened 2022-06-28
fix: use posix path for ts-node loader

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/22544
- Closes https://github.com/cypress-io/cypress/issues/22297

### User facing changelog

Correctly escape `ts-node` require when running on windows.

### Additional details

 Fixes a bug where TypeScript users with spaces in their path could not launch Cypress. Introduced in https://github.com/cypress-io/cypress/commit/abd986aa411af3ec358056a8458b74cb52952b01#diff-b0bd90502ad464c0d541a0e47822d15fec9ba5434dc4225d2993305e1ec7d30aR19. `require.resolve` returns a path which we pass to `NODE_OPTIONS` in the form of `--require <path>` which errors if there is white space in the path. The fix is to change the path to use posix separators.

### Steps to test

- go on windows
- put cypress in a directory with white space
- `yarn dev` wasn't working - now it is
- you should also be able to load a project that's got white space in the path
- you also should be able to migrate a project w/ white space - this might have explained some migration issues, too, for anyone migrating after this regression.
### PR Tasks


packages/data-context/src/actions/MigrationActions.ts
```diff
@@ -34,11 +34,11 @@ import {
 } from '../sources/migration'
 import { makeCoreData } from '../data'
 import { LegacyPluginsIpc } from '../data/LegacyPluginsIpc'
-import { hasTypeScriptInstalled } from '../util'
+import { hasTypeScriptInstalled, toPosix } from '../util'
 
 const debug = debugLib('cypress:data-context:MigrationActions')
 
-const tsNode = require.resolve('@packages/server/lib/plugins/child/register_ts_node')
+const tsNode = toPosix(require.resolve('@packages/server/lib/plugins/child/register_ts_node'))
```
emilyrohrbough, line 41: Is this helper necessary? Can we use path.joint out of the box?
```suggestion
const tsNode = require.resolve(path.join('@packages', 'server', 'lib','plugins','child', 'register_ts_node'))
```

packages/data-context/src/data/ProjectConfigIpc.ts
```diff
@@ -17,7 +17,7 @@ const debug = debugLib(`cypress:lifecycle:ProjectConfigIpc`)
 const CHILD_PROCESS_FILE_PATH = require.resolve('@packages/server/lib/plugins/child/require_async_child')
```
emilyrohrbough, line 17: Lets fix this one as well.

packages/data-context/src/actions/MigrationActions.ts
tgriesser: @emilyrohrbough I believe the intent here is to properly escape the actual resolved path that's returned from `require.resolve`, which `path.join` wouldn't help with

lmiller1990: Yes, the helper is the fix. We need to posix-ify the path, even on windows.

mjhenkes (approved):

packages/data-context/src/data/ProjectConfigIpc.ts
lmiller1990: I wonder if this is actually a problem? Eg - I added regression tests, none of them fail. Kind of hesitant to add this helper unless it's actually necessary.

We use it like this:

```js
    const proc = fork(CHILD_PROCESS_FILE_PATH, configProcessArgs, childOptions)
```

I think `fork` must be smart enough to escape the path on its own.

lmiller1990: Edit: I see https://docs.cypress.io/guides/references/advanced-installation#Binary-cache

- Windows: /AppData/Local/Cypress/Cache

Since Cypress is always in a directory w/o a space, this has never been a problem. If it's never been a problem, I am kind of hesitant to add this change right now before shipping (without another round of manual testing and a way to reproduce the error it's meant to be fixing.

tgriesser (approved):

PR #22559, opened 2022-06-28
chore: Adding record key context for internal ui jobs

tbiethman:
### User facing changelog

N/A

### Additional details

With https://github.com/cypress-io/cypress/commit/9101a9066c44fa2bb5bcbca64351c30cb5457cae, we added a dependency on a new context to all the `run-new-ui-tests` jobs. Without this context set, we'd fall into the external contributor build logic, which seems like it might be the source of a few new failures we're seeing.

### Steps to test

Gotta watch the build

### PR Tasks

- [ ] Have tests been added/updated?
- [ ] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [ ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [ ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?

PR #22569, opened 2022-06-28
test: Addressing yet more flaky windows tests

tbiethman:
### Additional details

Changes include:

* Tweaked a test in global-mode.cy.ts to avoid clicking a link before it was enabled
* Updated the circle config to provide appropriate context vars to the windows ui (app/launchpad) builds. Shoutout to @ZachJW34 for noticing that I missed these earlier!

### Steps to test

The app/launchpad tests in windows should be green now, I think this covers the recent flake we've seen. So PRs should be happier. The kitchensink build that executes in develop is still failing though.

### PR Tasks

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2712,7 +2712,7 @@ windows-workflow: &windows-workflow
         name: windows-run-app-integration-tests-chrome
         executor: windows
         resource_class: windows.large
-        context: test-runner:launchpad-tests
+        context: [test-runner:cypress-record-key, test-runner:launchpad-tests]
         requires:
           - windows-build
 
@@ -2720,7 +2720,7 @@ windows-workflow: &windows-workflow
         name: windows-run-launchpad-integration-tests-chrome
         executor: windows
         resource_class: windows.large
-        context: test-runner:launchpad-tests
+        context: [test-runner:cypress-record-key, test-runner:launchpad-tests]
         requires:
           - windows-build
 
--- a/packages/launchpad/cypress/e2e/global-mode.cy.ts
+++ b/packages/launchpad/cypress/e2e/global-mode.cy.ts
@@ -30,8 +30,11 @@ describe('Launchpad: Global Mode', () => {
         .dropFileWithPath(projectPath)
       })
 
-      cy.contains('Welcome to Cypress!')
-      cy.get('a').contains('Projects').click()
+      cy.contains('Welcome to Cypress!').should('be.visible')
+      cy.findByRole('link', { name: 'Projects' })
+      .should('have.attr', 'aria-disabled', 'false')
+      .click()
+
       cy.get('[data-cy="project-card"]')
       .should('have.length', 1)
       .should('contain', 'todos')


packages/launchpad/cypress/e2e/global-mode.cy.ts
```diff
@@ -30,8 +30,11 @@ describe('Launchpad: Global Mode', () => {
         .dropFileWithPath(projectPath)
       })
 
-      cy.contains('Welcome to Cypress!')
-      cy.get('a').contains('Projects').click()
```
tbiethman, line 34: This link goes from disabled to enabled as the page loads. Updated the selector here to validate that it's enabled prior to clicking it.

circle.yml
```diff
@@ -2712,15 +2712,15 @@ windows-workflow: &windows-workflow
...
+        context: [test-runner:cypress-record-key, test-runner:launchpad-tests]
         requires:
           - windows-build
 
     - run-launchpad-integration-tests-chrome:
         name: windows-run-launchpad-integration-tests-chrome
         executor: windows
         resource_class: windows.large
-        context: test-runner:launchpad-tests
+        context: [test-runner:cypress-record-key, test-runner:launchpad-tests]
```
tbiethman, line 2723: I missed these earlier in https://github.com/cypress-io/cypress/pull/22559

packages/launchpad/cypress/e2e/global-mode.cy.ts
BlueWinds: Hm. cy.click() waits for the `disabled` property to be false - https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Disability. So this could possibly be a bug with the button, where we have it `aria-disabled=true` but `disabled=false`?

BlueWinds: I'm not against merging this as-is, it's just a confusing state to be in because our actionability checks around clicking are pretty extensive.

tbiethman: This is just a bog standard link with some attribute juggling to keep it accessible while still "disabled": https://github.com/cypress-io/cypress/blob/develop/packages/frontend-shared/src/gql-components/HeaderBarContent.vue#L58-L65. The actual `disabled` attribute isn't used here, so it might be sailing through our actionability checks?

BlueWinds: Hm, yeah. It looks like it's actually our actionability checks that are incomplete! We don't respect aria-disabled at all. 

I'm a little surprised there's no existing cypress issues that mention `aria-disabled`. Let's merge this as-is, and do you mind opening a follow-up to add aria-disabled to our actionability checks?

tbiethman: Good call, logged: https://github.com/cypress-io/cypress/issues/22578

lmiller1990 (approved):
Pair reviewed earlier, looks good

PR #22572, opened 2022-06-28
fix: Require react-dom as a dependency for React projects

### LINKED ISSUE #22371: Cypress implicitly expects react-dom to be installed during component testing setup
### Current behavior

Cypress 10.1.0 - Components setup using web interface does not work when select React and vite.

```
[15940:0617/071413.082:ERROR:gpu_init.cc(446)] Passthrough is not supported, GL is disabled, ANGLE is                                                                               d, ANGLE is 
X [ERROR] Could not resolve "react-dom"

    node_modules/cypress/react/dist/cypress-react.esm-bundler.js:9:26:
      9 Γöé import * as ReactDOM from 'react-dom';
        Γò╡                           ~~~~~~~~~~~
  You can mark the path "react-dom" as external to exclude it from the bundle, which will 
remove this error.

7:26:45 AM [vite] error while updating dependencies:
TypeError: Cannot read properties of undefined (reading 'outputs')
    at runOptimizeDeps (C:\Users\nlrdn\GravitySource\rluisn-devtest\Cypress-TestFramework\node_modules\vite\dist\node\chunks\dep-8f5c9290.js:39993:49)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async runOptimizer (C:\Users\nlrdn\GravitySource\rluisn-devtest\Cypress-TestFramework\node_modules\vite\dist\node\chunks\dep-8f5c9290.js:50498:38)
```

### Desired behavior

Cypress does not crashes when using React and vite Into DevServer

### Test code to reproduce

yarn add cypress --dev
npx cypress open 

3- Then in the first page select components *beta 
4- When the combo box appear, select React and Vite as bundle combination.
5- Then execute the command in the terminal yarn add -D vite react 
6- And create the vite.config.ts

```ts
//vite.config.ts
import { defineConfig, loadEnv } from 'vite'

// https://vitejs.dev/config/

module.exports = defineConfig(({ command, mode }) => {

    // Load env file based on `mode` in the current working directory.
    // Set the third parameter to '' to load all env regardless of the `VITE_` prefix.
    const env = loadEnv(mode, process.cwd(), '')

    if (command === 'serve') {
        return {
            // dev specific config
            define: {
                __APP_ENV__: env.APP_ENV
            }
        }
    } else {
        // command === 'build'
        return {
            // build specific config
            define: {
                __APP_ENV__: env.APP_ENV
            },
            build: {
                lib: {
                    entry: 'src/index.ts',
                    formats: ['es']
                },
                rollupOptions: {
                    // make sure to externalize deps that shouldn't be bundled
                    // into your library
                    external: /^lit/,
                    output: {
                        // Provide global variables to use in the UMD build
                        // for externalized deps
                        globals: {
                            lit: 'lit'
                        }
                    }
                }
            }
        }
    }
})
```

### Cypress Version

10.1.0

### Other

_No response_

astone123:
- Closes #22371 

### User facing changelog

React DOM has been explicitly named as a required dependency for component testing in React applications

### Additional details

For component testing in applications that use React, we require both `react` and `react-dom` libraries, but only `react` is listed as a required dependency, so the wizard will allow users to continue without notifying them if their project doesn't have `react-dom` as a dependency, and then when they try to run a test, it will fail because `react-dom` is required.

This PR adds `react-dom` as a required dependency for React applications using Vite and ones created with Create React App so that users know to install `react-dom` in their project if it wasn't already installed.

### Steps to test

Follow these steps with a Create React App app and a React app using Vite:

- Create a new application
- Either load the project into Cypress via global mode or run `yarn cypress open` in the project directory
- Select component testing
- Verify that when the wizard checks the project's dependencies, it includes `react-dom`
- Verify that if `react-dom` doesn't exist in the project, the wizard reflects this and prompts you to install it

Screen Shot 2022-06-28 at 3 33 50 PM

### PR Tasks

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2390,22 +2390,22 @@ linux-x64-workflow: &linux-x64-workflow
         requires:
           - build
     - run-frontend-shared-component-tests-chrome:
-        context: [test-runner:launchpad-tests, test-runner:percy]
+        context: [test-runner:cypress-record-key, test-runner:launchpad-tests, test-runner:percy]
         percy: true
         requires:
           - build
     - run-launchpad-integration-tests-chrome:
-        context: [test-runner:launchpad-tests, test-runner:percy]
+        context: [test-runner:cypress-record-key, test-runner:launchpad-tests, test-runner:percy]
         percy: true
         requires:
           - build
     - run-launchpad-component-tests-chrome:
-        context: [test-runner:launchpad-tests, test-runner:percy]
+        context: [test-runner:cypress-record-key, test-runner:launchpad-tests, test-runner:percy]
         percy: true
         requires:
           - build
     - run-app-integration-tests-chrome:
-        context: [test-runner:launchpad-tests, test-runner:percy]
+        context: [test-runner:cypress-record-key, test-runner:launchpad-tests, test-runner:percy]
         percy: true
         requires:
           - build
@@ -2418,7 +2418,7 @@ linux-x64-workflow: &linux-x64-workflow
         requires:
           - system-tests-node-modules-install          
     - run-app-component-tests-chrome:
-        context: [test-runner:launchpad-tests, test-runner:percy]
+        context: [test-runner:cypress-record-key, test-runner:launchpad-tests, test-runner:percy]
         percy: true
         requires:
           - build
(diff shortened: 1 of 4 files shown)


ZachJW34 (changes requested):
`react-dom` should be added to `type: 'nextjs'` config as well since it relies on the `@cypress/react` mounting lib

lmiller1990 (approved):
Looks good once we add React DOM to Next.js as a dependency, too.

ZachJW34 (approved):

PR #22588, opened 2022-06-29
fix: restore truncating behavior in Last Updated column

marktnoonan:
### User facing changelog

The "Git info" in the specs list will now truncate correctly when the "Last Updated" column is narrower than the content width.

### Additional details

This regressed as part of implementing feedback on a recent PR. I've added Percy snapshots for the specs list component test at screen sizes that should cause this overflow, to prevent future regressions.

The narrowest Percy snapshot will look incorrect due to vertical alignment problems in the rows at that breakpoint. This is a separate, known issue that will be solved in a different PR. While I don't like putting a known wrong snapshot in the Percy build, I'd rather add the coverage now with everything else so that when the problem is corrected, the diff will reflect the correction.

### Steps to test

Open Cypress in any project. Visit the specs list and resize the window to be narrow enough to cause overflow of the text in the columns.

**Before (last updated column cut off):**
Screen Shot 2022-06-28 at 10 02 53 PM

**After (last updated column with ellipsis):**
Screen Shot 2022-06-29 at 8 47 38 AM

### PR Tasks

  - added percy snapshots


packages/app/src/specs/SpecsList.cy.tsx
```diff
@@ -121,6 +121,23 @@ describe('<SpecsList />', { keystrokeDelay: 0 }, () => {
       cy.get('@showCreateSpecModalSpy').should('have.been.calledOnce')
     })
 
+    it('displays as expected at various widths', () => {
```
marktnoonan, line 124: Two things about this, which I can correct myself when back in the office tomorrow or somebody is free to take over:

1. This test should be combined with the `when screen is narrow` type tests lower down in this file
1. The Percy snapshots probably need to weight for the git info to be visible so we can confirm the overflow behavior, [this snapshot here ](https://percy.io/cypress-io/cypress/builds/19277314/changed/1082226949?browser=chrome&browser_ids=23&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=2000&widths=500%2C650%2C800%2C1000%2C1200%2C1280%2C1500%2C2000)is the right size but since the content isn't rendered it doesn't capture the overflow.

marktnoonan: Moved things around a little and added a wait to allow for our 50ms delay built into rendering some of the row contents.

packages/app/src/specs/SpecsList.cy.tsx
```diff
@@ -164,6 +164,25 @@ describe('<SpecsList />', { keystrokeDelay: 0 }, () => {
...
+        cy.viewport(500, 850)
+        cy.percySnapshot('narrowest')
+        cy.viewport(650, 850)
+        cy.percySnapshot('narrow')
+        cy.viewport(800, 850)
+        cy.percySnapshot('medium')
+        cy.viewport(1200, 850)
+        cy.percySnapshot('wide')
+        cy.viewport(2000, 850)
+        cy.percySnapshot('widest')
```
rachelruderman, line 184: Are we using the [Tailwind CSS viewport breakpoints](https://tailwindcss.com/docs/responsive-design)? If so, it would be nice to mirror the same values

rachelruderman (approved):
Works as expected! One nice-to-have comment on the test breakpoint values

packages/app/src/specs/SpecsList.cy.tsx
marktnoonan: I believe columns are using those breakpoints for changing titles and showing/hiding, but the truncating behavior kinda falls in between, depending on how wide the column is, so I think I'd rather leave this as is it.

Though a more general page level percy helper command could be really useful so we cover more breakpoints without having to be too explicit about it 🤔

marktnoonan: A `cy.percyResponsive` that did the viewports and snapshots for you.

PR #22606, opened 2022-06-29
fix: Ensure that file watchers are closed before Electron exits

astone123:
Closes https://github.com/cypress-io/cypress/issues/22026

This PR uses a `will-quit` hook for Electron to ensure that we close all of the DataContext file watchers before the application exits.

See the related issue for more info.

### User facing changelog

* Fixed an issue where file watchers were not completely closed prior to the Cypress app quitting, causing crashes on exit in some macOS environments.

**Testing**:
- Check out develop
- Follow the repro steps in the related issue to reproduce a SIGABORT error upon exiting the application
- Check out this branch
- Follow the steps again and verify that the SIGABORT error will not happen


packages/data-context/src/DataContext.ts
```diff
@@ -429,7 +429,11 @@ export class DataContext {
   }
 
   async destroy () {
-    const destroy = util.promisify(this.coreData.servers.gqlServer?.destroy || (() => {}))
```
astone123, line 432:
```
util.promisify(() => {})
```
This noop didn't work because it never resolved. We replaced it with `() => Promise.resolve()`

packages/server/lib/modes/interactive.ts
```diff
@@ -170,6 +173,36 @@ export = {
       makeGraphQLServer(),
     ])
 
+    const clearCtxAndQuit = async () => {
```
astone123, line 176: This file contains the bulk of the changes. Most of the other changes in this diff are related to `await`ing function calls that are now async

packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -311,7 +311,8 @@ export class ProjectConfigManager {
   }
 
   onLoadError = (error: any) => {
-    this.closeWatchers()
+    // Tyler TODO: Do we need to close here? Making this async has downstream effects
```
astone123, line 314: @tbiethman any idea about this?

tbiethman: My comment is related to our calling this in the ProjectLifecycleManager (https://github.com/cypress-io/cypress/blob/develop/packages/data-context/src/data/ProjectLifecycleManager.ts#L802); making this async would have ramifications to a number of APIs there without some refactoring.

It might be necessary though. Keeping the watcher open while in the error state would result in `refreshLifecycle` being [called on change](https://github.com/cypress-io/cypress/pull/22606/files#diff-d81c2004d6fb4e9796dcb278f97ccf435abb59ddbb4b41d8f9706bf5f26f8af7R338), which wouldn't be expected in the previous code. We could try forcing an error and seeing what impact this has.

packages/server/test/integration/cypress_spec.js
```diff
@@ -449,10 +452,13 @@ describe('lib/cypress', () => {
       })
     })
 
-    it('scaffolds out support + files if they do not exist', function () {
+    // NOTE: This test is skipped because it causes the subsequent test to fail intermittently
+    it.skip('scaffolds out support + files if they do not exist', async function () {
```
astone123, line 456: @lmiller1990 @estrada9166 I marked this as skipped because it was causing the test after it to intermittently timeout in its current state. Maybe we could meet to discuss what this test actually does? In its current state I don't think we even make it to the block that calls `cypress.start`, so I'm not sure what this is actually testing.

Reading through the code here doesn't make much sense to me. It looks like we're actually asserting that support + files **are not** scaffolded despite what the test title says.

packages/server/lib/modes/interactive.ts
```diff
@@ -170,6 +173,36 @@ export = {
...
+      try {
+        await clearCtx()
```
lmiller1990, line 178: There is a comment saying we should not call this in runtime: https://github.com/cypress-io/cypress/pull/22606/files#diff-1827debd8088884005aba0e14a94e1e6815ceb1ad567db970f9f327d9feb7c9dR27

If we are doing this, I think we need to change the comment?

Also, maybe a comment here explaining why we clearCtx before closing? Could link to the relevant GH issue

packages/server/lib/modes/run.js
```diff
@@ -1650,6 +1650,7 @@ module.exports = {
   async run (options, loading = Promise.resolve()) {
     if (require('../util/electron-app').isRunningAsElectronProcess({ debug })) {
       const app = require('electron').app
+      const { clearCtx } = require('@packages/data-context')
```
lmiller1990, line 1653: Was there a reason to inline this require instead of putting it at the top?

Will make an eventual ESM convert easier.

packages/server/test/integration/cypress_spec.js
lmiller1990: I think we can nuke this, we don't auto scaffold support etc anymore, that is part of the launchpad onboarding. The fact this passes at all is probably because we stub a bunch of things? Weird.

lmiller1990 (commented):
Sorry it took so long to review, I missed this one. I left a few tiny comments, once those are good let's ship it.

packages/server/lib/modes/run.js
tbiethman: The existing data-context require used to be at root but was changed with https://github.com/cypress-io/cypress/pull/20932/files#diff-58fd4a8e28c13b52b19c42346eade66ed1de5a8b139541dff4e94fef42c3c701R666-R667. I assume due to how the run script is getting loaded?

Regardless, I am going to revert the changes to this file for run mode. I don't think the process.exit handler is deterministic here, just like it wasn't before in the ProjectLifecycleManager. And because we don't initialize watchers in run mode, the teardown isn't necessary prior to exit.

packages/data-context/src/data/ProjectConfigManager.ts
tbiethman: I'm attempting some refactoring here to allow us to keep closing the watchers here. The good news is we have lots of tests around this, the bad news is those tests are now failing 😞

tbiethman:
@astone123 @lmiller1990 I have addressed the current PR feedback and pushed a few new commits.

The biggest change was addressing the TODO regarding closing the file watchers when the ProjectConfigManager errored. I was able to keep closing these by untangling some of the circular logic between the ProjectLifecycleManager and the ProjectConfigManager. You can that see in [the diff](https://github.com/cypress-io/cypress/pull/22606/files#diff-63757aa782fe9ed7c93e5f91cd7afdcf8f617d2b4eea12d1f51d09747fb28247L801-L802). Doing it this way allows us to keep the lifecycle manager error handling functionally equivalent while allowing for async handling within the config manager itself. I'm happy to answer more questions around that, otherwise I think we're good to go for a final review.

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -717,10 +710,8 @@ export class ProjectLifecycleManager {
     }
   }
 
-  destroy () {
-    this.resetInternalState()
-    // @ts-ignore
-    process.removeListener('exit', this.onProcessExit)
```
ryanthemanuel, line 723: Do we not need to remove this anymore?

tbiethman: Nope, we no longer add the listener in the first place.

lmiller1990 (approved):
My comments are addressed. Just need CI to pass and we can merge?

PR #22612, opened 2022-06-30
chore: debugging CT performance slowness

### LINKED ISSUE #22353: CT Performance Improvements: v10 slower than v9
**Edit from @lmiller1990: this is getting worked on: https://github.com/cypress-io/cypress/pull/23104. Component Testing performance will be improved, initially released under an experimental flag to let use continue making incrementally performance updates. See latest comments for latest updates.**

### Current behavior

After upgrading to the latest version, we have an issue with slowness when we run `cypress run --component`. The tests themselves are not slow, but it seems there is some delay between specs runs as you can see in this video 👇 .

https://user-images.githubusercontent.com/6830426/174029644-aa31da35-ee35-4a04-8474-b0eab9a72869.mp4

We were not able to detect the exact issue, but we would appreciate any help 🙏

I was trying to run the tests with `DEBUG` and this might be potentially the issue 👇 
image

### Desired behavior

The tests should run smoothly without any delay.

### Test code to reproduce

We use cypress on public repo https://github.com/toptal/picasso.

We have cypress 10 on the current master.

```shell
yarn
yarn test:integration
```

for debugging it is faster to run
```shell
# these need to be run only once
yarn
yarn build:package

yarn test:setup cypress run --component
```

### Cypress Version

10.1.0

### Other

We use cypress with [Happo](https://happo.io/) for visual testing. The first idea was that it might be the problem since Happo didn't adapt to v10 yet, but it is not the case, I have tried to remove it from the repo and it didn't help.

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/22353 hopefully

WIP, this has a long way to go - need to settle on the final implementation. More to come!

### User facing changelog

Go faster

### PR Tasks

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -38,7 +38,7 @@ macWorkflowFilters: &darwin-workflow-filters
     - equal: [ develop, << pipeline.git.branch >> ]
     - equal: [ linux-arm64, << pipeline.git.branch >> ]
     - matches:
-          pattern: "-release$"
+          pattern: "lmiller/22353-ct-single-session"
           value: << pipeline.git.branch >>
 
 linuxArm64WorkflowFilters: &linux-arm64-workflow-filters
--- a/circle.yml
+++ b/circle.yml
@@ -27,7 +27,7 @@ mainBuildFilters: &mainBuildFilters
     branches:
       only:
         - develop
-        - issue-22147-nohoist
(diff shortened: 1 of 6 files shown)


lmiller1990:
In favor: https://github.com/cypress-io/cypress/pull/23104

Issue #22610, opened 2022-06-30
Firefox display bug


melsacramento:
### Current behavior

When running cypress 10 and using firefox(v102) as the browser. There is a problem with the display (see attached image)

This did not occur in cypress 9

One workaround to this is to minimize the browser display, but it can be very inconvenient as then its hard to see the test that are running

![image](https://user-images.githubusercontent.com/88633931/176597412-13041cd4-87f0-46d0-abd5-dad251ab6ba8.png)

### Desired behavior

Expect the display to look normal

### Test code to reproduce

Unable to give source code, though as long as its a lengthy one it seems to happen

### Cypress Version

10.3.0

### Other

I am also using kubuntu 22.04 if that helps

marktnoonan:
Thanks for reporting @melsacramento, this looks like another manifestation of https://github.com/cypress-io/cypress/issues/21881, which I am actually testing a fix for right now, and expect it to be included in the next release of Cypress.

melsacramento:
Thank you very much, looking forward to the fix

PR #22620, opened 2022-06-30
fix: avoid overflow issue in Firefox

marktnoonan:
- Closes https://github.com/cypress-io/cypress/issues/21881
- Closes https://github.com/cypress-io/cypress/issues/22610

### User facing changelog

The Spec Runner will no longer overflow unexpectedly in Firefox in certain situations.

### Additional details

It appears that what's really going on here is that Firefox is basing the `scrollHeight` of the container for the panels off of the initial height of the AUT before it gets scaled. So even though everything looks correct, a hidden scrollable area is created at the bottom of the window. Other browsers do not seem to calculate the scroll height this way, and this can't be triggered in any other browser. Firefox itself returns to the correct visual state after some window resizing.

Most of the time this is fine, users can't scroll to that area accidentally, due to `overflow: hidden`. What's surprising is when you hover over certain parts of the command log, it causes FF to scroll the panels area to the bottom for some reason. This can happen without devtools open if users make their window small enough.

This PR removes the AUT container `div` from its place in the document flow using `position: absolute`, which fixes this issue in Firefox as there's never any possibility this element would take up its original space. Due to the AUT scaling logic, the AUT viewport never exceeds the size of the window and always takes up the right amount of space anyway.

It's hard to pin down the exact reason this regressed from 9.x - the DOM surrounding the AUT is very different in 10.x, with a grid and many other layers.

### Steps to test

- Open any Component or E2E spec in Firefox and reduce the height of the window so that the AUT scales down significantly. 
- Run the spec, and hover on the commands
- Observe that the UI does not shift, whereas before this change, it might have
- Use `cy.screenshot()` and confirm the screenshots look as you expect. I've tested and found no problems, all the tests and screenshot diffs themselves still pass, but every time we touch this area, we should double check that nothing has been disturbed.

The state below is no longer possible to reach in FF:

Screen Shot 2022-05-02 at 8 51 54 AM

### PR Tasks

- [na] Have tests been added/updated? - Cypress-in-Cypress tests cannot run in Firefox


packages/app/src/runner/ResizablePanels.vue
```diff
@@ -4,7 +4,6 @@
     class="flex"
     :class="{
       'select-none': panel1IsDragging || panel2IsDragging,
-      'overflow-x-hidden': isFirefox
```
BlueWinds, line 7: Love that we can remove browser-specific styling and just apply one thing that works everywhere. +1

BlueWinds:
You would not believe how excited I am for this fix. I use FF as my primary dev browser, and often have the dev tools quite large, leading to me seeing the bug basically constantly. Thank you!

BlueWinds (approved):

packages/app/src/runner/useRunnerStyle.ts
```diff
@@ -62,10 +62,15 @@ export const useRunnerStyle = () => {
   })
 
   const viewportStyle = computed(() => {
+    // in the below styles, `position: absolute` is required to avoid certain edge cases
+    // that can happen with overflow (mainly, in Firefox, but not always) since, see
```
ZachJW34, line 66:
```suggestion
    // that can happen with overflow (mainly, in Firefox, but not always)
```

ZachJW34 (approved):
Works great! Small nit with the comment

PR #22624, opened 2022-06-30
fix: UTM parameters for Latest Runs column

### LINKED ISSUE #22623: Append UTM params to "Latest Runs" URL in Specs Explorer
### Current behavior

URLs in `RunStatusDots.vue` do not contain UTM params

### Desired behavior

Update `RunStatusDots.vue` to append UTM params to the dashboard URL - needs to be done both for the row-level link and the tooltip-level link. Ideally this would use the `getUrlWithParams` utility, but this may need to be updated to work with prebuilt URLs that already contain params.

utm_source: the regular Binary: App/Launchpad source that we get from the helper function
utm_medium: Specs Latest Runs Dots
utm_campaign: for this one, let's use the status itself as is displayed in the tooltip: so Failed, Passed, Timed Out, etc.

### Test code to reproduce

1. Open cloud-connected project
2. Mouseover Latest Runs row data, observe link does not contain UTM params

### Cypress Version

10.2.0

### Other

_No response_

mike-plummer:
- Closes #22623 

### User facing changelog
* Add UTM parameters to dashboard URL for Latest Runs content

### Additional details
* Update util function to work with URLs that have existing params - previous implementation could end up with duplicate `?` causing a malformed URL if input URL already contained query params like the URLs we get back for direct-linking to the dashboard

### Steps to test
1. Open cloud-connected project with recorded run(s)
2. Login if not already logged in
3. Observe Latest Runs data in Specs Explorer rows, mouse over
4. Inspect - observe that link for Latest Runs "dots" in row and in tooltip that is opened on mouseover have links that contain UTM parameters

Latest Runs links to dashboard now contain UTM parameters

### PR Tasks


flotwig (approved):

packages/frontend-shared/src/utils/getUrlWithParams.ts
```diff
@@ -7,14 +7,29 @@ export type LinkWithParams = {
...
+    if (link.url.includes('?')) {
+      // If input URL already includes params we should preserve them
+      url = link.url.substring(0, link.url.indexOf('?'))
+      searchParams = new URLSearchParams(link.url.substring(link.url.indexOf('?')))
+    } else {
+      url = link.url
+      searchParams = new URLSearchParams()
+    }
+
+    Object.entries(link.params).forEach(([key, value]) => searchParams.append(key, value))
```
ZachJW34, line 30: Looks like this can be simplified by using `new URL(...)`, which will preserve the searchParams if already provided.

```suggestion
    const url = new URL(link.url)

    Object.entries(link.params).forEach(([key, value]) => url.searchParams.append(key, value))

    result = url.href
```

packages/frontend-shared/src/utils/getUrlWithParams.ts
```diff
@@ -7,14 +7,29 @@ export type LinkWithParams = {
 
 export const getUrlWithParams = (link: LinkWithParams) => {
   let result = link.url
-  const hasUtmParams = Object.keys(link.params).some((param) => param.startsWith('utm_'))
 
-  if (hasUtmParams) {
-    link.params.utm_source = getUtmSource()
-  }
+  if (Object.keys(link.params).length > 0) {
```
ZachJW34, line 11: Nit: You can store `Object.keys(link.params)` so you don't have to call it twice.

ZachJW34 (changes requested):
Suggested tweak that makes the helper function a bit cleaner

packages/frontend-shared/src/utils/getUrlWithParams.ts
mike-plummer: I started out with that approach, but hit a hurdle since it doesn't handle "partial" URLs like `./my/path` unless we manually provide a base URL as a separate argument. I wasn't sure whether it was safe to assume a "localhost"-relative path in all circumstances so I took the manual parsing approach to leave the relative-pathing decisions up to the browser when the link was resolved

ZachJW34: Ahh makes sense, your approach seems good then.

ZachJW34 (approved):

PR #22633, opened 2022-06-30
fix: remove CT side effects from mount when e2e testing

### LINKED ISSUE #22589: Adding a command for component mount to e2e support file causes an error
### Current behavior

I'm having issues launching e2e tests if I include the custom `cy.mount` command used for component testing in the support file for e2e tests. I'm not sure if this is intentional or not.

Launching any e2e test with the custom `cy.mount` command included results in the following error:

image

**Moving the custom `cy.mount` command to `component.tsx` does seem to fix the issue**

My `support` folder is structured as follows:
```
support
 - commands.tsx
 - component.ts
 - component-index.html
 - e2e.ts
 - TestsWrapper.tsx
```

Both `e2e.ts` and `component.ts` include identical code (import the `commands.tsx` file):
```
import './commands'
import '@cypress/code-coverage/support'
```

The custom `cy.mount` command looks as follows:
```
Cypress.Commands.add('mount', (component, options = {}) => {
  const wrapped = <TestsWrapper>{component}</TestsWrapper>
  return mount(wrapped, options)
})
```

The `TestsWrapper` is a React wrapper component that includes several Providers, but that doesn't seem to be causing the issue here. Writing the command as its simplest version, `Cypress.Commands.add('mount', mount)`, still throws the same error.

My logic here was that I wanted to keep the commands separate from other setup code for improved composition, as this approach is also recommended in the documentation [here](https://docs.cypress.io/api/cypress-api/custom-commands).
The error however wasn't particularly verbose as to what exactly is going wrong. Maybe the command is somehow executed on initialization?

### Desired behavior

I believe that there are a few ways to improve on this behaviour.

First of all, possibly specifying in documentation to **not include `cy.mount` in the e2e support file** would help.

Secondly, improving the error reported through Cypress in this specific case to include something like "Are you possibly including the `mount` command for component testing in end-to-end testing?" might also be helpful.

Finally, as I'm not 100% sure if this is a bug or not, in the case it is, the custom `mount` command inclusion should not throw an error at all.

### Test code to reproduce

I've forked your `cypress-test-tiny` repo, added `create-react-app`, initialized component testing and moved around the code in `cypress/support` file to adhere to the structure I've mentioned above.

The error occurs on e2e testing.

Repo link [here](https://github.com/bozskejkaja/cypress-mount-command-e2e-crash).

### Cypress Version

10.3.0

### Other

Thanks for the awesome project! You've been saving me hours of headaches for the past several years. ✌️

ZachJW34:
- Closes #22589

### User facing changelog
CT side effects from `import { mount } from 'cypress/{react,vue}'` are disabled when running e2e tests.

### Additional details
Though we scaffold CT with the mount registration in `cypress/support/component.js`, a user might lift this registration to the scaffolded `commands.js` file, which is not specific to testing type. There are side effects associated with the mount import that are required for CT testing but they will cause errors when e2e testing.

Though we have e2e and component specific support files, users might be inclined to keep all of their custom commands in a single file as documented [here](https://docs.cypress.io/api/cypress-api/custom-commands). Doing so should not cause these errors to surface, so I've stubbed out the CT side effects when running e2e (and improved a few errors thrown due to not finding the root selector).

Ideally, I don't want the `mount` import to have side effects. I think there should be an explicit import/registration that would be used inside of the `component.js`support file, but that would be a breaking change so I've left some comments around a future refactor of this.

### Steps to test
The original issue has a good reproduction and I've included a system test to verify the new behavior. You can scaffold a basic project with e2e and component testing and lift the `Cypress.Commands.add('mount', ...)` registration into `commands.js` then try to run an e2e test and the error will be thrown.

Before: 
Screen Shot 2022-06-30 at 2 14 00 PM

After:
Screen Shot 2022-06-30 at 2 09 08 PM

### PR Tasks


system-tests/projects/e2e-with-mount-import/cypress.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  e2e: {
```
lmiller1990, line 2: Instead of adding a new project, couldn't we use an existing one that has CT and E2E and just update the `support/e2e.js` to import `cypress/react`? Just to minimize the number of example projects.

lmiller1990 (commented):
Two small comments.

ZachJW34:
@astone123 Due to how the packages are bundled with the binary, they don't get synced with a normal `yarn watch`. Can you try checking out the branch and running `yarn`? This will rebuild all the packages, then you can try running an e2e test. I used the `reporter` package as it had both ct and e2e setup

astone123:
@ZachJW34 I tried running `yarn` on your branch, and I see that it works for the reporter package, but I still see 

```
No element found that matches selector [data-cy-root]. Please use the mount utils to mount it properly
```

In my project with that `commands.js` file when I try to run an e2e test

npm/mount-utils/src/index.ts
```diff
@@ -200,6 +200,17 @@ export const injectStylesBeforeElement = (
 }
 
 export function setupHooks (optionalCallback?: Function) {
+  // Consumed by the framework "mount" libs. A user might register their own mount in the scaffolded 'commands.js'
+  // file that is imported by e2e and component support files by default. We don't want CT side effects to run when e2e
+  // testing so we early return.
+  // System test to verify CT side effects do not pollute e2e: system-tests/test/e2e_with_mount_import_spec.ts
+  if (Cypress.testingType !== 'component') {
+    // eslint-disable-next-line no-console
+    console.log('Skipping component setup as Cypress is not running in component mode. You are probably seeing this message due to an inclusion of a "mount" import in your support file.')
```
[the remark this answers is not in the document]
ZachJW34: Gonna remove it. We should probably outright block the use of `mount` in e2e tests but that was outside of the scope of this PR so I didn't tweak it.

system-tests/projects/e2e-with-mount-import/cypress.config.js
ZachJW34: Moved to `component-tests` project, keeping the system-test project bloat down is good with me

ZachJW34:
@astone123 We chatted, testing this change with a project outside of the monorepo will resolve `cypress/react` from the external project and thus not have the updated changes.

lmiller1990 (approved):

lmiller1990:
Looks good, let's get this green and ship it.

astone123 (approved):

ryanthemanuel (approved):

lmiller1990:
I thought about this a bit more @ZachJW34 - the current approach doesn't really disable side effects, it just returns early in some places in the mount adapters.

In https://github.com/cypress-io/cypress/pull/22437 we do a dynamic import for `react-dom` depending on the react version. Unfortunately, if the supportFile registers `mount` from `cypress/react`, the default webpack preprocessor will attempt to optimize `react-dom/client` and throw an error if it doesn't exist. This will happen in projects using React <=17 and E2E.

In the `cypress/webpack-dev-server`, we've got special logic to handle this - but not in the default preprocessor, and I don't know we'd want to leak CT specific things into webpack-preprocessor.

I wonder if there's a smart way to actually **ignore** the component testing mount import and associated custom commands completely, rather than just returning early (thus truly having no side effects).

I'll sync up w/ @ZachJW34 on this, but I am not sure the approach in here is ideal.

PR #22635, opened 2022-06-30
fix: Styling of specs list at small breakpoints

### LINKED ISSUE #22626: In the specs list the vertical alignment of Specs is incorrect
### Current behavior

At smaller breakpoints row content is offset vertically

175789330-562f6ef8-5a69-4791-b523-ed8539c916ab

### Desired behavior

Spec row content should remain vertically centered regardless of screen width

### Test code to reproduce

None

### Cypress Version

10.3.0

### Other

_No response_

mike-plummer:
- Closes #22626 

### User facing changelog
* Fix styling issues of Specs Explorer at narrow viewport widths

### Additional details
* Fix issue where spec rows aren't vertically centered at narrow viewport widths
  - This was caused by using CSS `visibility` rather than `display` which caused layout space to be preserved while hiding the item visually when we actually wanted the item to be fully removed from the layout flow at the lower breakpoint. It was being correctly handled in the header but not in the row items
* Per product request, increase min width of "New Spec" button to prevent wrapping. We may want to look into responsively scaling the `Button` component's padding & icon size but that feels like a larger issue to be addressed with UX

### Steps to test
1. Open a project with specs in the Cypress app, navigate to Specs Explorer
2. Narrow viewport to < 768px, observe spec rows remain vertically centered and button text does not wrap

**Before**
Screen Shot 2022-06-30 at 3 46 18 PM

**After**
Screen Shot 2022-06-30 at 3 45 47 PM

### PR Tasks


emilyrohrbough (approved):

emilyrohrbough:
@mike-plummer sorry for the suppppper slow review!

Issue #21271, opened 2022-04-29
CYPRESS_NO_COMMAND_LOG=1 leaves faulty empty space


Agalias:
### Current behavior

Hello, currently when running `CYPRESS_NO_COMMAND_LOG=1 npm run cy:open` and opening any test - there is some space left where Test Runner Command Log should be. Probably because of reporter

![Screenshot 2022-04-28 at 17 33 43](https://user-images.githubusercontent.com/77224571/165777677-0ea3577c-a132-4f65-82c1-480339efb369.png)

Probably empty space is left there because `<div class="reporter-wrap" style="left: 0px;"></div>` is not deleted. If to delete it and resize the Test Runner window - empty space will disappear.

https://user-images.githubusercontent.com/77224571/165968558-71fa981b-70ca-44a5-aec1-bf55b118b6fe.mov

### Desired behavior

There is no space left where Test Runner Command Log should be when using `CYPRESS_NO_COMMAND_LOG=1` environment variable:
![Screenshot 2022-04-28 at 17 40 19](https://user-images.githubusercontent.com/77224571/165778113-d7e6b449-10cc-48b3-9c4c-e98289f38837.png)
 

### Test code to reproduce

Should be reproducible anywhere, I tested it on the fresh 9.6.0 Cypress project with getting started examples.

### Cypress Version

8.7.0, 9.6.0

davidmunechika:
Thanks for opening an issue. I was able to reproduce on 9.6.0 and it looks like a regression occurred sometime after 5.4.0 (when it was implemented).

rockhold:
I tested this out in v10.0.2 and realized we don't yet have support for this environment variable in v10+. Here is a related issue for implementing that support: #21721

We should ensure that the spacing issue called out here is taken care of as part of that work.

PR #22641, opened 2022-06-30
fix: respect CYPRESS_NO_COMMAND_LOG in env

marktnoonan:
- Closes https://github.com/cypress-io/cypress/issues/21721
- Closes https://github.com/cypress-io/cypress/issues/21271

### User facing changelog

The `CYPRESS_NO_COMMAND_LOG` environment variable will hide the command log when set to `1`. This is documented [here](https://docs.cypress.io/guides/references/troubleshooting#Disable-the-Command-Log) but hasn't been working correctly for a few major versions (left an empty space) and was not yet implemented in Cy 10.x.

### Additional details

This PR makes a `hideCommandLog` boolean available under `window.__CYPRESS_CONFIG`, since we want this variable accessible on the client side at startup time in run mode, which does not use GraphQL. 

In order to prevent rendering the reporter, we avoid rendering the element it would mount into. This let me find a few places where in 10.x, we assume it will always be there. So I added some checks where needed, and uncoupled finding the reporter element, which is allowed to fail, from finding the runner element, which should still throw if it cannot be found.

The tests include some (near!) duplication to cover run mode as well as open mode. I thought about making a custom command but for just 2 instances of a very short test, it didn't seem worth it.

### Steps to test

- Set `CYPRESS_NO_COMMAND_LOG=1` before any Cypress `run` or `open` command and open a spec.
- Verify that you don't see the Command Log or Inline Specs List with this setting enabled.
- Verify that in E2E and CT mode that the Application/Component Under Test is correctly centered and scale behavior is good. The AUT area should never overflow the window no matter how small.
- In open mode, the navigation bar on the left should be present and collapsed, in run mode, is should not be there at all.

Cypress can run tests without the reporter again:

Screen Shot 2022-06-30 at 7 11 27 PM

### PR Tasks


packages/app/src/store/runner-ui-store.ts
```diff
@@ -40,6 +41,7 @@ export const useRunnerUiStore = defineStore({
       reporterWidth: runnerConstants.defaultReporterWidth,
       automationStatus: automation.CONNECTING,
       randomString: `${Math.random()}`,
+      noCommandLog: window.__CYPRESS_CONFIG__.noCommandLog,
```
emilyrohrbough, line 44: I find it's best to name vars with words that have read with true/value meanings. For example `showCommandLog` if true, it shoes, if false, it hides it or hasCommandLog . Can we rename noCommandLog to something that reads like the truthy / falshiness of the value?

packages/app/cypress/e2e/cypress-in-cypress-run-mode.cy.ts
```diff
@@ -62,6 +62,35 @@ describe('Cypress In Cypress - run mode', { viewportWidth: 1200 }, () => {
...
+    })
+
+    cy.visitApp()
+    simulateRunModeInUI()
+    cy.contains('dom-content.spec').click()
+
+    cy.contains('http://localhost:4455/cypress/e2e/dom-content.html').should('be.visible')
+    cy.findByLabelText('Stats').should('not.exist')
+    cy.findByTestId('specs-list-panel').should('not.be.visible')
+    cy.findByTestId('reporter-panel').should('not.be.visible')
```
emilyrohrbough, line 90: Does this render the hidden div no content in it?

packages/app/src/store/runner-ui-store.ts
marktnoonan: I agree in general - I kept it like this to not lose the connection to the actual name of the environment variable, but the runner UI short is probably the best place to break that connection, and convert it from "information in the config that came from the CLI" to a more readable boolean name, will update.

packages/app/cypress/e2e/cypress-in-cypress-run-mode.cy.ts
marktnoonan: Right, the panels are managed with a v-show, which means they are in the DOM with `display: none` on them until they appear. This is useful for the first panel since it has to appear instantly when opened, and preserve state when closed, but actually come to think of it, it's not important for the 2nd panel, since the command log cannot currently be toggled while the application is open, so we can change to a `v-if` here and lose div.

packages/app/src/store/runner-ui-store.ts
marktnoonan: I'm going with `hideCommandLog` every where to at least keep the truthy/falsy patten matched with `CYPRESS_NO_COMMAND_LOG=1`. I think it does read better this way.

packages/app/src/runner/utils.ts
```diff
@@ -8,7 +8,7 @@ export const REPORTER_ID = 'unified-reporter'
 function getElementById (id: string) {
   const el = document.querySelector<HTMLElement>(`#${id}`)
 
-  if (!el) {
+  if (!el && id !== REPORTER_ID) {
```
lmiller1990, line 11: This isn't a good idea - `getElementById` is generic and used by multiple things, having a check for some specific constant in here definitely is a code smell. Most likely this check should move to whatever function calls `getElementById` that led to this change, probably `getReporterElement`?

packages/app/src/runner/utils.ts
```diff
@@ -23,8 +23,8 @@ export function getReporterElement () {
   return getElementById(REPORTER_ID)
 }
 
-export function empty (el: HTMLElement) {
-  while (el.lastChild) {
+export function empty (el: HTMLElement | null) {
```
lmiller1990, line 26: Same here - `empty` looks like it's designed to empty a `HTMLElement` - this change reflects we don't understand the different workflows and states our system can be in. Rather than proceeding with "try to empty", we should ensure we only call `empty` when we know we need to.

So, I think the check should be placed where this was prompted. Something like:

```js
if (!someElement) {
  // someElement doesn't exist for X reason, return
  return
}

empty(someElement)
```

lmiller1990 (changes requested):
I left some comments. These are not nitpicks, but actual problems that I think we need to address. Adding a null check here or there is what led to `packages/server` having thousands of paths that no-one really understands, since the correct and valid states of the system are muddied by the amount of things than can be null.

We should ensure what is already a pretty complex front-end does not divulge into the same state, and I think it starts here - add the null checks where and if they are needed, rather than pushing them down into the core functions that we composable the rest of the system from. I think the "business logic" should be on the outer parts of the system, where we compose things (using things like `getElementById` and `empty`, and not leak into the core functions.

packages/app/src/runner/utils.ts
marktnoonan: I think I should remove this helper itself, and use `document.getElementById` directly in `getRunnerElement` and `getReporterElement`, since only `getRunnerElement` actually needs to throw, and this helper function doesn't get used elsewhere.

packages/app/src/store/runner-ui-store.ts
marktnoonan: I have a few reasons I like it in the store. This global variable stuff is a weird little exception for values that get used in run mode and can't come out of gql - I'd rather the components no have to know/care about that. It feels weird to grab it from the window in the Open Mode runner component, but also weird to have 2 slightly different ways to get the same data depending on which mode we are in. Putting it in the store, both components who need to use this value can get it in the same way and not care about the source, and both components will update correctly over time if something about the source changes.

I don't feel super strongly about this, but I think it's better than the alternatives.

packages/app/src/runner/index.ts
```diff
@@ -227,7 +227,7 @@ function runSpecCT (spec: SpecFile) {
   // watches spec file for changes.
   getEventManager().setup(config)
 
-  const $runnerRoot = getRunnerElement()
+  const $runnerRoot = getRunnerElement() as HTMLElement
```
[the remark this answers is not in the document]
marktnoonan: This was because `getElementById` had an internal condition where it could sometimes not return an element (since the reporter is not guaranteed). Will address this by just removing this helper and putting the check in `getRunnerElement`.

packages/app/src/runner/utils.ts
lmiller1990: Sure, that works!

lmiller1990 (approved):
My comments have been fixed :heavy_check_mark:

flotwig (approved):
Lookin' good.

packages/app/src/runner/SpecRunnerOpenMode.vue
```diff
@@ -22,7 +22,7 @@
       :min-panel2-width="minWidths.reporter"
       :min-panel3-width="minWidths.aut"
       :show-panel1="runnerUiStore.isSpecsListOpen && !screenshotStore.isScreenshotting"
-      :show-panel2="!screenshotStore.isScreenshotting"
+      :show-panel2="!screenshotStore.isScreenshotting && !hideCommandLog"
```
rachelruderman, line 25: Just wondering, how come we're not mirroring the env var name? I.e. why `NO_COMMAND_LOG - hideCommandLog` instead of `HIDE_COMMAND_LOG - hideCommandLog` or `NO_COMMAND_LOG - noCommandLog` ?

marktnoonan: It started out that way, changed it based on @emilyrohrbough's [feedback here](https://github.com/cypress-io/cypress/pull/22641#discussion_r911969672) since in general boolean names are formed with verbs like `isThing` or whatever.

emilyrohrbough (approved):

packages/app/src/runner/SpecRunnerOpenMode.vue
marktnoonan: Yeah I think this is marginal either way. Since tracing this takes you to the store, and the store shows how the value is initialized, I'm still just slightly in favor of "readability when somebody has to read/modify the Vue files" since that's probably the much more common situation for this code to be read.

PR #19572, opened 2022-01-05
fix: open browser at correct time during lifecycle

tgriesser:
We should be "initializing" the server via `initializeActiveProject` right around when we choose the testing type, after the config is loaded, prior to opening the Browser. A project should not be considered "loaded" until the server is started, as the creation of the server affects the configuration used by the open app, such as the server port & routing.

We're not doing this at the moment, and instead have added in a codepath to create it just in time, when the browser is launched, causing race conditions and subtle bugs when the app is loaded.

This is not currently caught by E2E tests, because we are calling `initializeActiveProject` in there, so things appear to work properly, even though this is not the actual behavior of the app.

- [x] Remove `initializeActiveProject` from e2e tests
- [x] Remove JIT creation from `open_project.launch`
- [x] Create the open project as part of the loading process of the testing type
- [x] Handle cleanup so that the test setup/teardown works properly

----

Fixes: [UNIFY-883]
Fixes: [UNIFY-881]

[UNIFY-883]: https://cypress-io.atlassian.net/browse/UNIFY-883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[UNIFY-881]: https://cypress-io.atlassian.net/browse/UNIFY-881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #22648, opened 2022-07-01
fix: resolve "project" breadcrumb problems

marktnoonan:
- Closes https://github.com/cypress-io/cypress/issues/21911

Keeping in draft till some related changes here land: https://github.com/cypress-io/cypress/pull/22073

### User facing changelog

In the Launchpad, the project's name will no longer link to the "Choose Testing Type" step of the the launchpad. This behavior was removed because in many situations it is not valid to go back to the testing type picker, and in the situation where it is valid, there's already a "Switch testing types" link on the same page.

### Additional details

Being able to "go back" midway through a migration, set up, or during an error state can lead to problems and confusion, but having a piece of text that only becomes a link on one screen doesn't seem that helpful to users. It should be a link, or not.

I considered making the rules more complex, or finding a way to safely clean up and navigate to this screen, which would address the the problem in the linked issue as well, but both options seem to me like we'd be maintaining these rules long term for quite low value. 

After discussion with Ryan, we have design approval to not make this a link.

### PR Tasks


marktnoonan:
@emilyrohrbough I know you like this link, let me know if you like it enough we should talk more about it before this PR happens.

emilyrohrbough:
@marktnoonan If Ryan has signed off, I trust he has considered the tradeoffs and user impact of removing this.

Thanks for tagging me! I appreciate it 😄  You're right that I do like it and am sad to see this go. It feels very natural to navigate from here given its a fairly upper left corner is an established pattern for browsers and filesystem finders. This area is where I tend to go for navigation (after keypad swipes) for navigating forward/back.

marktnoonan:
@tbiethman this is good feedback for design, I'm happy for now to be able to prevent the user reaching an invalid/confusing state, but it's a good point about the overall sense of how navigation works in the launchpad.

ZachJW34 (approved):
Sorry for the late review! Works great

One small nit: You can remove `isLoadingNodeEvents` from the query in `HeaderBarContent.vue`.

lmiller1990:
Looks like once https://github.com/cypress-io/cypress/pull/22648#pullrequestreview-1043521781 is addressed we can merge?

lmiller1990:
LGTM, are we going to do https://github.com/cypress-io/cypress/pull/22648#pullrequestreview-1043521781?

Otherwise let's get CI green and merge this up.

marktnoonan:
@lmiller1990 👍 , updated. 

The reason I didn't do this right away was I got slightly distracted with wondering how many other places we could potentially remove `isLoadingNodeEvents` - it looks like there were several files where it is present in the GQL data requested but not actually being referenced elsewhere in those components. So I started to wonder if it was included for some other reason than directly affecting the UI, like it has some side effect that we want, and I wanted to look a little further. After checking the history, it seems safe to remove the `isLoadingNodeEvents` from HeaderBarContent.vue, since it was added at the same time it was used, so I've done that.

@tgriesser there's [this PR](https://github.com/cypress-io/cypress/pull/19572/files#diff-99636480d4ea376f639524746fcddf8126cd1ad0abf0f5a2b570d6bb56085722) where the some of the change is just to add some fields including `isLoadingNodeEvents` to the some queries, and those fields are not actually referenced by the components. What's the purpose of that? These are the usages that made me ever so slightly cautious before removing.

PR #22659, opened 2022-07-01
chore: ensure percy nonce updates with new workflow trigger

emilyrohrbough:
Currently we are using the SHA means only one successfully Percy run per commit which means we can't rerun a CircleCI workflow "from start" -- we an only re-run "from failed". This changes update to use CIRCLE_WORKFLOW_ID for Percy Nonce to allow you to kick off a full workflow restart with the possibility of finalizing Percy more than once per commit.

I have hit this multiple times, especially w hen trying to release & a new empty commit must be pushed to successfully re-run CI.


tbiethman (commented):
My current (admittedly shaky) understanding of this is that the $CIRCLE_SHA1 is ideal for PRs, and that $CIRCLE_WORKFLOW_ID is would be idea for 'develop', as that is the only branch that that auto-approves without waiting for all jobs to successfully complete: https://github.com/cypress-io/cypress/blob/de4cb660505e39a0ae3947638e7e640b069c795f/scripts/wait-on-circle-jobs.js#L18-L21

From back in March: [Internal Slack](https://cypressio.slack.com/archives/C033N3S6ZU4/p1648567644184489)

emilyrohrbough:
@tbiethman I'd be curious what @marktnoonan says on the review.

My understanding was that we were using the commit because it was guaranteed to be unique since percy can only finalize once per nonce. 

So many be the root issue is really that since we always auto-finalized in develop regardless of the rob status, we will always fail a retried develop/master run.

marktnoonan:
The timing! I am writing up other Percy things this afternoon.

The tradeoff here is that we open ourselves up to partial builds in Percy when we do re-run from failed, which means in order to review a full percy diff for a given PR, all of the tests have to pass the first time (if we use CIRCLE_WORKFLOW_ID for the nonce). When it worked that way, the percy diff associated with a PR would only contain the snapshots uploaded during the _last_ run, even if the last run was just repeating a failing job. This means often we had 100s of missing images in the diff and it wasn't possible to review in a meaningful way. Those snapshots had been uploaded to Percy but were in a black hole of unfinalized builds with the previous workflow id.

This is why we switched to the commit sha for the nonce - so that repeated runs will accumulate into the same open percy build, and finalize it as a full diff at the end. Since a full re-run from start can be done with an empty commit, it's not super painful, though it would be better not to need that. 

I was going to suggest using `CIRCLE_WORKFLOW_ID` when the branch is `develop`, and keep commit sha when it's a PR? That would match the needs of both situations: 

- Develop currently always finalizes the first time, which means both "re-run from failed" and "re-run from start" don't work, which can needlessly block things from going green when trying to release - 100% needs to be fixed to make releasing better.
- When PRs "rerun from failed" they won't create new partial builds, they'll add to the build for the given commit.

We ought to have done this before merging 10.0, but since parts of Percy are still WIP it looks like it went on the backburner.

tbiethman:
@marktnoonan that makes sense to me. Did we investigate _not_ auto-approving builds in develop and waiting there as well? In doing so we would trade using an approved partial build as a baseline (when the develop merge flakes) for a completely missing build for that commit, unless the build gets retried until completion. I guess that doesn't seem like a great option either 🤷‍♂️

emilyrohrbough:
@marktnoonan 
I think I am following you... and when looking deeper into these details to expand on what I was noticing for you, It looks like I updated the wrong env I noticed that was persisting per pipeline retries. It should have been `CIRCLE_WORKFLOW_WORKSPACE_ID`.

When comparing runs within a pipeline original & those rerunning from failed, it seemed to me that the `CIRCLE_WORKFLOW_WORKSPACE_ID` matched - it seems to be associated to a set of runs. Then, when a "run from start" is triggered, a new `CIRCLE_WORKFLOW_WORKSPACE_ID` for each workflow in the pipeline.

Take a look at  the linux-64 workflow runs for the `cypress39940` pipeline on this branch. 
- When clicking the first Linux-64 run (7 days ago) in the pipeline, it's assigned `CIRCLE_WORKFLOW_WORKSPACE_ID=755a53c8-0f1d-43fa-98dc-6b477e3befcd`. 
- When you click into the last linux-64 run (1 hr ago) it also has `CIRCLE_WORKFLOW_WORKSPACE_ID=755a53c8-0f1d-43fa-98dc-6b477e3befcd`.

marktnoonan:
@emilyrohrbough `CIRCLE_WORKFLOW_WORKSPACE_ID` sounds really good, that seems like it would work exactly how we want and never need the extra step pushing commits! Let's try it.

@tbiethman we did talk about that, there's a long recorded meeting I can dig up with me and Brian going over all the options and accepting the tradeoffs.

emilyrohrbough:
@marktnoonan 👍🏻 cool. it's an easy fix to revert if this doesn't work like we'd expect. I updated my branch to use `CIRCLE_WORKFLOW_WORKSPACE_ID`.

tbiethman (approved):
Good find @emilyrohrbough , let's give it a shot.

marktnoonan (approved):
Initially looks good! I've just done a rerun from failed, let's confirm it all works before we merge. What we want to see on the rerun from failed is the same build ID logged in the output. And then after a rerun from start, a whole new build with a different nonce and a different ID. 

If you want to check the nonce being used, it's not in the UI but they do pass the data to the front end via their API: https://percy.io/api/v1/builds/19347755?include-metadata=true

Screen Shot 2022-07-01 at 4 49 42 PM

I'm approving now so you can merge when you're happy with it after I"m offline.

emilyrohrbough:
@marktnoonan Did this match like you expected?

marktnoonan:
@emilyrohrbough confirmed through looking at the percy history of builds for this, we have expected combinations:

Same commit can rerun from failed and will use have the same nonce each time, as it should. But also the same commit can run more than once and have a different nonce each time, as it should in regular re-run. 

This is GREAT 🥇

emilyrohrbough:
@marktnoonan 🎉 whoo!!

PR #22661, opened 2022-07-01
fix: add tooltip to New Spec button

jljohnson001:
- Closes [Issue 21858](https://github.com/cypress-io/cypress/issues/21858)

### User facing changelog
n/a

### Additional details
**Implementation Details:**
We have a few different tooltip styles in the project. Specifically the affected page has 2 (Created, SpecHeaderCloudDataTooltip). In lieu of introducing a third tooltip style to the page (exactly matching "Run All Tests") I opted to match the Created tooltip:
image

### Steps to test
1. Mouse Hover over New Spec button

**Before:**
Note: Screenshot is on New Spec hover
image

**After:**
image

### PR Tasks

### DIFF
--- a/packages/app/src/specs/SpecsListHeader.cy.tsx
+++ b/packages/app/src/specs/SpecsListHeader.cy.tsx
@@ -62,6 +62,14 @@ describe('<SpecsListHeader />', { keystrokeDelay: 0 }, () => {
     .should('have.been.called')
   })
 
+  it('provides expected tooltip content', () => {
+    cy.get('.v-popper').trigger('mouseenter')
+    cy.findByTestId('new-spec-tooltip').should('be.visible')
+        .and('have.text', 'Create New Spec')
+
+    cy.percySnapshot()
+  })
+
   it('emits a spec pattern event', () => {
     const onShowSpecPatternModal = cy.stub().as('show-spec-pattern-modal')
     const search = ref('')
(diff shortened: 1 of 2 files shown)


marktnoonan (changes requested):
Hi! https://github.com/cypress-io/cypress/issues/21858 is not referring to the "New Spec" button on the main specs list, instead it's talking about the "New Spec" button in the inline specs list in the Spec runner:

171218512-b2a7aaa3-e191-45d4-ab74-6d11c4be23a6

This can be opened and closed with this button when you are on a spec:

Inline specs list

I think the style you picked is still correct though. We would just need to move this tooltip to the button in the inline specs list. Since the main specs list button already has a text label, I don't think we need the extra tooltip there.

The other note I'd make is that we should capitalize this way to be consistent with our other capitalization: "Create new spec". 

I coincidentally found an asset from a not-yet finalized design for something else that happens to how how this tooltip should look:

Screen Shot 2022-07-13 at 10 05 07 AM

Thanks for picking up the issue! Sorry about the vague issue description, it looks like it was migrated from our Jira board as more of a feedback note and could have been more clear about the intended outcome.

nagash77:
@jljohnson001 please contact us once you have had a chance to respond to comments from the Cypress team and we will be happy to open this PR back up.

jljohnson001:
@nagash77 and @marktnoonan thank you for the feedback. I have updated this code to include your suggestions.

PR #22736, opened 2022-07-11
refactor: Replace use of html-webpack-plugin with custom lightweight version

ZachJW34:
### User facing changelog
n/a

### Additional details
Using HtmlWebpackPlugin is overkill for our needs. Using a custom version allows us to drop two dependencies (webpack 4 and 5 versions)

### Steps to test
System tests

Same behavior as before

https://user-images.githubusercontent.com/25158820/178296180-cdd68479-6be3-4d96-bdf8-fb0ad6a17955.mov

### PR Tasks


npm/webpack-dev-server/src/IndexHtmlWebpackPlugin.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    const indexHtmlContent = await this.getIndexContent(compilation, compiler)
+    const entryPointScriptTags = entryPointPublicPaths.map((entryPath) => `<script defer src="${entryPath}"></script>`).join('')
+    const indexContentWithScripts = indexHtmlContent.replace(/<\/head>/, `${entryPointScriptTags}</head>`)
+
+    return indexContentWithScripts
+  }
+
+  private async getIndexContent (compilation: Compilation, compiler: Compiler): Promise<string> {
```
lmiller1990, line 59: Cool, nice to see you working on this. html webpack plugin does a lot of things - do we know what things it does that we don't need (eg, what makes it overkill?)

ZachJW34: HtmlWebpackPlugin contains a whole bunch of [options](https://github.com/jantimon/html-webpack-plugin#options) that we don't need and don't surface to the user. The only thing we rely on is the live-reloading and the injection of script tags. The one thing that we would lose is the handlebar templating syntax that HtmlWebpackPlugin handles.

I could change the flow a bit and have this plugin be a fallback in case we can't `require('html-webpack-plugin')`. If the user doesn't have this, they would just need a barebones plugin which we would provide.

emilyrohrbough:
Closing as stale. Can re-visit when needed.

PR #22741, opened 2022-07-11
fix: Resolve Git Info when using alternative shells or Windows

### LINKED ISSUE #22454: Support getting spec Git info when active shell differs in syntax from bash
### What would you like?

I'd like to be able to see Git info for my project even if my active shell uses a different syntax than bash. Currently I use [Fish Shell](https://fishshell.com/) as my default shell on Mac. When Cypress tries to load Git info for my project, it uses syntax that isn't supported by my active shell (Fish in this case).

The result is that my project's Git info doesn't show in the spec list

```
[cy:open:dev:95630]:   cypress:data-context:sources:GitDataSource command execution error: { stdout: '', stderr: "fish: Unsupported use of '='. In fish, please use 'set IFS $'\n''.", code: 0, failed: false, killed: false, signal: null, cmd: "IFS=$'\n" + `'; for file in {"/Users/adams/projects/cypress-example-kitchensink/cypress/e2e/2-advanced-examples/actions.cy.js", ... 
```

Screen Shot 2022-06-22 at 10 31 44 AM

### Why is this needed?

This improvement would be useful for people who use different default shells and run Cypress on their machine. 

### Other

When we execute the command to get the Git info, we use the active shell and fall back to bash.

```typescript
const result = await execa(cmd, { shell: process.env.SHELL || '/bin/bash', cwd: this.#gitBaseDir })
```

Forcing the use of bash fixes the issue in my case but I imagine that there are some cases where we need to use the active shell.

mike-plummer:
* Update *nix flow to use /bin/sh rather than user's active shell
* Add unit tests for filename edge cases

- Closes #22454 

### User facing changelog
* Allow Git Information to be retrieved for specs when using less common UNIX shells like Fish
* Fix issue that prevented displayed "Last Updated" information on Windows for Git repositories without a commit history

### Additional details
Existing implementation attempted to use the user's active shell when executing Git info commands, falling back to Bash if no shell was specified on the env variable. This worked fine for *most* users since most Linux flavors default to Bash, and Mac uses Zsh which happens to understand the syntax we were using. However, some users with less common shells like Fish were seeing failures since it uses a different variable syntax.

This PR updates our Posix logic to use `/bin/sh` (the `execa` default) which should be available on all *nix flavors (Bash isn't 100% available on all flavors). It also addresses an edge case where files with double quotes in their path would break the command, and adds unit tests to validate special character edge cases.

### Steps to test

**UNIX Systems**
Validate that files with special characters are properly handled and that alternative user shells don't break the experience
1. Create a new empty Cypress project
2. Initialize your project as a Git repo (`git init`)
3. Open project in the app
4. Create an empty spec named `file"withOneDoubleQuote.cy.js`
5. Observe that "Last Updated" info displays in the Specs Explorer along with a "Created" icon
    * App successfully handled file with double quote character when processing Last Updated info
6. Close the app
7. Install the FISH shell (`brew install fish`) or [via another install mechanism](https://fishshell.com/)
8. In your command prompt switch your active shell to Fish (`fish`) and open Cypress
9. Observe that "Last Updated" info displays in the Specs Explorer along with a "Created" icon
    * App successfully acquired Git information despite running in Fish shell

**Windows Systems**
Validate that "Last Updated" info is displayed for empty Git repositories
1. Create a new empty Cypress project, open in the app
2. Create an empty spec
3. Observe that "Last Updated" info displays in Specs Explorer, but no "created/modified" icon is displayed
4. Close the app
5. Initialize your project as a Git repo (`git init`)
6. Reopen project in Cypress, observe that last updated info still displays and now shows a "Created" icon in Specs Explorer
    * App successfully acquired "Last Updated" info on Windows for an empty Git repository (no commit history)

* Last Updated information now populates when using an alternative shell on Unix systems
 
|  State | Before | After |
| --- | --- | --- |
| Non-Git Project in Fish | FISH-Before-NoGit | FISH-After-NoGit |
| Git project in Fish | FISH-Before-EmptyGit | FISH-After-EmptyGit |
 
* Last Updated information now populates on Windows for projects that are newly-initialized Git repositories
 
|  State | Before | After |
| --- | --- | --- |
| Non-Git Project | WIN-Before-NoGit | WIN-After-NoGit |
| Empty Git project | WIN-Before-EmptyGit | WIN-After-EmptyGit |
| Git Project w/ history |  WIN-Before-NonEmptyGit | WIN-After-NonEmptyGit |

### PR Tasks

- [na] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


packages/data-context/test/unit/sources/GitDataSource.spec.ts
```diff
@@ -99,6 +99,64 @@ describe('GitDataSource', () => {
     expect(modified.lastModifiedTimestamp).not.to.be.undefined
   })
 
+  it(`handles files with special characters on ${os.platform()}`, async () => {
```
BlueWinds, line 102: Suggest adding a comment linking to to https://github.com/cypress-io/cypress/issues/22454 in this test. I love when tests link issues so I can tell exactly what bug they're protecting against. :)

BlueWinds (approved):

astone123 (approved):
This works for me using Fish 👍🏻 🐟

packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -339,19 +339,21 @@ export class GitDataSource {
 
   async #getInfoPosix (absolutePaths: readonly string[]) {
     debug('getting git info for %o:', absolutePaths)
-    const paths = absolutePaths.map((x) => `"${path.resolve(x)}"`).join(',')
+    const paths = absolutePaths
+    .map((x) => x.replace(/\"/g, '\\"'))
```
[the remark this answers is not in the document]
mike-plummer: Done

packages/data-context/src/sources/GitDataSource.ts
```diff
@@ -339,19 +339,22 @@ export class GitDataSource {
 
   async #getInfoPosix (absolutePaths: readonly string[]) {
     debug('getting git info for %o:', absolutePaths)
-    const paths = absolutePaths.map((x) => `"${path.resolve(x)}"`).join(',')
+    const paths = absolutePaths
+    .map((p) => path.resolve(p))
```
lmiller1990, line 343: Any reason we do three `map` here? That's a lot of extra loops.  Can't we do the modification in a single `map`?

packages/data-context/test/unit/sources/GitDataSource.spec.ts
```diff
@@ -49,9 +49,9 @@ describe('GitDataSource', () => {
 
     // create a file and modify a file to express all
     // git states we are interested in (created, unmodified, modified)
-    const fooSpec = path.join(e2eFolder, 'foo.cy.js')
-    const aRecordSpec = path.join(e2eFolder, 'a_record.cy.js')
-    const xhrSpec = path.join(e2eFolder, 'xhr.cy.js')
+    const fooSpec = path.join(e2eFolder, 'foo.cy.js').replace(/\\/g, '/')
```
lmiller1990, line 52: Could potentially use https://github.com/cypress-io/cypress/blob/develop/packages/data-context/src/util/file.ts ?

lmiller1990 (commented):
Just some small comments, sorry I took so long to review this.

PR #22571, opened 2022-06-28
chore: Refactor chainer / Commands.add for readability

BlueWinds:
### User facing changelog
No user-facing changes.

### Additional details
This PR is the first of hopefully several with some cleanup and refactoring of a very hard to read and central part of our code base: How commands are added, queued, and executed.

There are no functional changes. The code is logically identical - but hopefully easier to read, with less internal state tracking and passing around of variables "to be used at some later point."

### Steps to test
All cypress tests execute identically. The CI build passing without test changes should show this well enough, but feel free to manually play with it as well.

### How has the user experience changed?
UX is unchanged. DX, reading this code, should be far better.

### PR Tasks
- [n/a] Have tests been added/updated? No test changes, because new code is functionally equivalent.
- [n/a] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [n/a] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [n/a] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?

PR #22742, opened 2022-07-11
chore: Refactor cy.state('subject') and `cy.then()`

BlueWinds:
### User facing changelog
`cy.state('subject')` is deprecated, and reading from it will log a warning to the console. Prefer `cy.currentSubject()` instead. 

### Additional details
This PR is a follow up to https://github.com/cypress-io/cypress/pull/22571, and refactors how Cypress keeps track of the current subject. No behavior has changed, but this is in preparation for addressing Detached DOM errors (#7306).

While we would normally consider internal state to be internal, `cypress-testing-library` relies on `cy.state('subject')`. Therefore, this PR includes a shim to keep that call working as before. Once this PR is released, we will follow up with them to move over to the new call, ensuring that no end users (with or without cypress-testing-library) ever experience disruption (beyond the console warning).

### Steps to test
The automated tests cover this code extensively, especially the `packages/driver` e2e tests.

No change. There is less indirection around how commands are added and invoked, which results in less memory churn and CPU usage (but only a ms here or there, not really noticeable).

### PR Tasks
- [x] Have tests been added/updated? Yes


packages/driver/src/cy/logGroup.ts
```diff
@@ -27,6 +27,7 @@ export default (Cypress, userOptions: Cypress.LogGroup.Config, fn: Cypress.LogGr
   const endLogGroupCmd = $Command.create({
     name: 'end-logGroup',
     injected: true,
+    args: [],
```
flotwig, line 30: what's this change for?

packages/driver/src/cypress/command_queue.ts
```diff
@@ -247,7 +249,7 @@ export class CommandQueue extends Queue<$Command> {
       // we're finished with the current command so set it back to null
       this.state('current', null)
 
-      this.state('subject', subject)
```
flotwig, line 250: This PR removes `cy.state('subject')`, right? I think we should be careful here - it's not a publicly-documented API, but it seems like it's used in a couple of Cypress plugins, testing-library/cypress-testing-library and Lakitna/cypress-commands (tests only):

https://github.com/search?q=cy.state%28%27subject%27%29&type=code

What do you think about still setting `cy.state('subject')` just for backwards-compatibility, and opening a second PR for 11.0.0 that will remove it entirely, so the plugins have some time to update? `cypress-testing-library` gets 3MM npm downloads a month, so I'm worried about breaking those users.

BlueWinds: This PR changes the formatting of it, to be an object rather than a single entry, in preparation for changing it even further down the line (into an even more complex data structure). You're right though, if testing-library is relying on it, then this is a breaking change and I absolutely cannot make it as part of a `chore`.

Hm. This is really rough though. The call `cy.state('subject')` is exactly what I'm trying to get rid of - it's the source of detached DOM errors, distilled to its essence. Let me think on this a bit and figure out what to do here.

packages/driver/src/util/serialization/index.ts
```diff
@@ -72,7 +72,7 @@ const convertObjectToSerializableLiteral = (obj): typeof obj => {
     })
 
     currentObjectRef = Object.getPrototypeOf(currentObjectRef)
-  } while (currentObjectRef)
+  } while (currentObjectRef && currentObjectRef !== Object.prototype && currentObjectRef !== Date.prototype)
```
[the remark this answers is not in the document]
BlueWinds: This is a performance optimization I noticed while debugging. It's otherwise unrelated to this PR.

If you turn on "break on exceptions" and add "break on caught exceptions" in the chromium debugger, you can see that we hit this code literally thousands of times for every test, because for every object we try to serialize, we recurse down to Object.prototype and iterate over its properties, throwing and catching dozens of "can't serialize native code" exceptions.

Same with dates - we see them as objects, recurse down into the prototype, and then throw and catch dozens of exceptions.

packages/driver/src/cy/logGroup.ts
BlueWinds: Consistency. This homogenizes `$Commands` created here with commands created everywhere else in Cypress. Adding the empty array here means that elsewhere in the PR I can assume that all commands have an args array.

I've honestly forgotten exactly where it was throwing an error, but it was somewhere pretty arcane, and this one-line change meant I didn't have to add existence checks deep inside the command-queue.

packages/driver/src/cypress/command_queue.ts
BlueWinds: Ok, so thinking about this further, I think we can in fact leave `cy.state('subject')` 'mostly working', and add a deprecation warning to it. The basic problem is that the *it does not map cleanly to the current subject*, and never has. I guarantee you every third-party project using `cy.state('subject')` has subtle bugs around the use of `cy.within` and `cy.as`.

That's what I mean when I say "mostly working" - I can add a shim to keep it working as-is (only a little bit broken).

BlueWinds: So current plan:
1. Add a shim on `cy.state('subject')` in this PR so it keeps working as a `chore`. Merge this PR.
2. In next PR, introduce selectors / function chains as subject, as a `feat`. Keep the shim working.
3. Open a PR with `cypress-testing-library` to move them to the new functionality.
4. Remove shim in breaking 11.x release eventually.

packages/driver/src/cy/commands/querying/within.ts
```diff
@@ -76,6 +78,10 @@ export default (Commands, Cypress, cy, state) => {
       })
     }
 
+    // TODO: Rework cy.within to use chainer-based subject chaining, rather than its custom withinSubject state.
+    // For now, we leave this logic in place and just ensure that the new rules don't interfere with it.
+    cy.breakSubjectLinksToCurrentChainer()
```
emilyrohrbough, line 83: Could you explain this? If I am understanding correctly, it seem like this would break the following? 


```
cy.get('.form').within(() => { do something}).should('contain', 'submit')
```

packages/driver/src/cypress/cy.ts
```diff
@@ -1257,11 +1259,40 @@ export class $Cy extends EventEmitter2 implements ITimeouts, IStability, IAssert
...
+  }
+
+  linkSubject (fromChainerId, toChainerId) {
+    const links = this.state('subjectLinks') || {}
+
+    links[fromChainerId] = toChainerId
+    this.state('subjectLinks', links)
+  }
+
+  breakSubjectLinksToCurrentChainer () {
```
emilyrohrbough, line 1276: Thoughts on adding in some docs on what these are doing? Seems like a lot of code to replace `.state('subject')`

Curious on what rules/scenarios are  `breakSubjectLinksToCurrentChainer` for should be used?

BlueWinds: Ok, so that novel of a comment aside, I'm unsure if I want to document this *right now* - I believe the next step (introducing the concept of 'selectors' and subject chains) will change how this works again, and commenting the intermediate stage extensively didn't feel quite right.

With that said, I'm not exactly sure how much will change - it may be worth documenting the current state in-code, rather than just in my head or in PR comments like this. I'll add a comment here (but probably not a complete description of the system).

packages/driver/src/cy/commands/querying/within.ts
BlueWinds: So the key is that cy.within works differently from cy.then - it *never changes the subject.* According to our docs (https://docs.cypress.io/api/commands/within#Yields):

>  .within() yields the same subject it was given from the previous command.
> Trying to return a different element the .within callback have no effect:
```
cy.get('#within-yields')
  .within(() => {
    // we are trying to return something
    // from the .within callback,
    // but it won't have any effect
    return cy.contains('Child element').should('have.class', 'some-child')
  })
  .should('have.id', 'within-yields')
```

So cy.within always matches the two conditions I mentioned in my other comment - the callback contains cypress commands, and we do not want to use their resolved subject as the new subject of the parent chainer. The existing automated tests back up this behavior.

I didn't know this about cy.within - I learned it by looking at the tests, being confused, then looking at the docs to see if the tests were correct. :rofl: I can't speak to why it works this way, just that it does and I'm trying to retain all the existing rules.

packages/driver/src/cypress/cy.ts
BlueWinds: Nevermind, I started writing and then words came out. Many lines of comments added around the new cy functions.

BlueWinds:
This PR is slightly larger than I'd like, apologies for lumping a couple of changes into the same batch. I'll try and start a separate branch for any further (non-review-related) changes.

packages/driver/src/cypress/command_queue.ts
lmiller1990: Nice catch here. We should probably make a PR against testing-library once this lands, to save them the time/effort, and while this is fresh in our minds.

lmiller1990 (commented):
I never realized how complex the driver and the "subject" concept was until I started following these PRs. I also don't see many references to `currentSubject()` in our tests - should there be more of those, if this is becoming part of the public API (or is it remaining private, like `cy.state('subject')`?

I'm still internalizing how this all works, and don't feel like I've gone into enough depth for an approval to have much weight, but good to see incremental, careful improvements to this core bit of our API.

BlueWinds:
> I never realized how complex the driver and the "subject" concept was until I started following these PRs. I also don't see many references to `currentSubject()` in our tests - should there be more of those, if this is becoming part of the public API (or is it remaining private, like `cy.state('subject')`?

It's remaining "mostly private". You only need to read it in exceptional cases. I'd say it should be entirely private, but since the community *is* reading it sometimes, we need to continue to support them.

`cypress-testing-library` is reading the subject directly because they want to integrate with our `.within()` command - they mostly care about `cy.state('withinSubject')`, rather than `cy.state('subject')`. I want to look at our implementation of .within to see if I can make their use easier, but that's something to mess with at a later date.

packages/driver/src/cy/commands/querying/within.ts
emilyrohrbough: Yeah, `.within()` is a bit interesting in that regard -- but curious what these new rules are exactly? It seems this logic is maintaining parity with the current logic?

Does your TODO comment mean you'd like to change this behavior later on or that you need to remove `cy.breakSubjectLinksToCurrentChainer()` ?

packages/driver/src/cy/logGroup.ts
emilyrohrbough, line 30: Given this is an injected command, why do we need this?

packages/driver/src/cypress.ts
```diff
@@ -218,6 +218,24 @@ class $Cypress {
...
+    /*
+     * As part of the Detached DOM effort, we're changing the way subjects are determined in Cypress.
+     * While we usually consider cy.state() to be internal, in the case of cy.state('subject'),
+     * cypress-testing-library, one of our most popular plugins, relies on it.
+     * https://github.com/testing-library/cypress-testing-library/blob/1af9f2f28b2ca62936da8a8acca81fc87e2192f7/src/utils.js#L9
+     *
+     * Therefore, we've added this shim to continue to support them. The library is actively maintained, so this
+     * shouldn't need to stick around too long (written 07/22).
+     */
+    Object.defineProperty(this.state(), 'subject', {
```
emilyrohrbough, line 231: Does this warning show in the reporter as a log or in the devTools console? 

Is this observed in run mode?

packages/driver/src/cy/commands/querying/within.ts
BlueWinds: Yes, this logic is maintaining parity with existing functionality. The 'two rules' I mentioned are just guidelines to help understand when `breakSubjectLinksToCurrentChainer` is appropriate to use.

The TODO comment is that I'll need to refactor `cy.within` as part of the effort of implementing selectors / fixing detached DOM. The goal will still be to maintain the existing functionality as part of any further work in here.

packages/driver/src/cypress/command_queue.ts
emilyrohrbough: Does PR likely needs to be flagged for release so a note can be made in the changelog that this usage has been deprecated if we are porting this functionality?

packages/driver/src/cypress.ts
BlueWinds: In the devtools console.

It does not show up at all in run mode.

ryanthemanuel (commented):
Looking good. Nice work!

packages/driver/src/cypress/cy.ts
```diff
@@ -1257,11 +1259,100 @@ export class $Cy extends EventEmitter2 implements ITimeouts, IStability, IAssert
...
+   * But you don't always want to use the inner chainer's subject for the outer chainer. Consider:
+   * `cy.then(() => { cy.get('foo').click(); return 'success' }).log()`
+   *
+   * In this case, we want to break the connection between the inner chainer and the outer one, so that we can
+   * instead use the return value as the new subject. Is this case, you'll want cy.breakSubjectLinksToCurrentChainer().
+   */
+  linkSubject (fromChainerId, toChainerId) {
+    const links = this.state('subjectLinks') || {}
+
+    links[fromChainerId] = toChainerId
```
emilyrohrbough, line 1315: is `fromChainerId` the child / chained command? ex:
cy.get().should().should() or c1 -> c2 -> c3 would map to:
```
 {
   'c2': 'c1',
   'c3': 'c1,
  }
```

packages/driver/src/cypress/command_queue.ts
BlueWinds: Good call. I've updated the PR description to note the change and the rational behind it.

ryanthemanuel (approved):

packages/driver/src/cypress/cy.ts
emilyrohrbough: These comments are amazing! 🤩

packages/driver/src/cypress/cy.ts
BlueWinds: Yep, exactly. I will add a note to change this to child / parent in my next PR, that will be clearer variable names.

packages/driver/src/cy/logGroup.ts
BlueWinds: Without this, we get errors in `commands/querying/root.cy.js` and `commands/querying/within.cy.js`, saying "args is not iterable." from this line in command_queue.ts:

  `args = [command.get('chainerId'), ...args]`

I could add a guard against commands that don't have args in there, or I could ensure that every command has an args array and not need to guard against it. I prefer consistent object shapes over null checks when possible.

The follow-up PR I'm currently working on removes the concept of 'injected' entirely (while maintaining all existing functionality and without editing any of the related tests).

packages/driver/src/cypress/cy.ts
emilyrohrbough: ahhh thats right. 👍🏻 thanks!

PR #22755, opened 2022-07-12
fix: save spec filter term

marktnoonan:
- Closes https://github.com/cypress-io/cypress/issues/21837

### User facing changelog

Cypress will remember the spec filter in the "Search Specs" field during your session, and also when a project is reopened. The same filter will be used in the main specs list as well as the inline specs list present in the Test Runner.

### Additional details

Previously, the search term used to find specs was not preserved when you navigated away from the specs list, and was not carried over between the inline specs list and the main specs list. This was a source of frustration for users, especially who are used to this behavior from before Cypress 10.

In earlier versions of Cypress, this term was saved in local storage in the launchpad Electron app. That approach doesn't work in Cypress 10 because the filtering now happens in the browser app, so users would be saving different search terms depending on which browser they are using at the time. So this PR adds the filter term to the projects `savedState` where some other project-specific information is already stored.

Some notes on the implementation:

1. A `specFilter` field has been added to `specs-store.ts` to preserve it when the user has left the Specs List page. This is the source of truth for the value of the filter on the client side.
2. A `useSpecFilter` composable is used in `SpecsList.vue` and `InlineSpecList.vue` which handles initializing the value in the store (from the saved data in the `currentProject.savedState.specFilter` field in gql), updating the store as the user types, and saving the search term back to gql. This reduces some of the duplication between these files and means we have a single place to, eg, set the the debounce time.
3. The existing `setPreferences` functionality has been extended to allow `global` and `project` preferences to be set. "Global" preferences are saved in the `__project__/state.json` file, and "project" preferences are saved in the `state.json` found the in the unique directory for that project. This required updating the existing usages of `setPreferences` and some related tests that stubbed it.
4. Created a new issue to clean up the confusing use of the `cache` file for project data like browser and testing type, those things probably doesn't need to live there. #23202

I looked at avoiding the store and driving this fully through graphql, but it seemed to require about the same amount of work, and using the store matches a little better with how we handle global preferences already.

We could optimize further by refactoring `setPromptShown` as well but we're about to be right back in this code, let's keep this PR from getting any bigger and make other changes in new PRs.

### Steps to test

- Open any project and search for a spec
- Refresh the page and see the search term is still there
- Navigate to other pages on the app and go back to the specs list, see the search term is still there
- Open a spec and open the inline specs list, see the search term is prefilled there
- Refresh the spec page to make sure the search term prefills based on the saved term in gql
- Type a new search in the Inline Specs List and go back to the main specs list, the new search should be there as well
- Open a different project, it should not have a saved search term.
- Search for a spec in the newly opened project, then close it and go back to to the first project, and see that the search terms saved are specific to the projects.
- Use the X in the specs list, and the clear button in the No Results state to confirm clearing the search in these ways also clears the saved search.

The Spec filter term was previously discarded when users navigated away from the page they were using to search. Now the filter is saved with the project and is only removed by user action.

### Percy

Note: changing some test nesting has changed the names of percy snapshots, so some will appear to be net-new that really aren't.

### PR Tasks

- [na] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? This isn't currently documented, so no change needed.


packages/app/src/specs/InlineSpecList.cy.tsx
```diff
@@ -28,108 +27,154 @@ describe('InlineSpecList', () => {
...
-    cy.get('li')
-    .should('exist')
-    .and('have.length', 7)
-
-    // overflow is required for the virtual list to work
-    // this test will fail if the overflow set by `useVirtualList`
-    // is overridden
-    cy.get('[data-cy="specs-list-container"]')
-    .should('have.css', 'overflow-y', 'auto')
+  describe('with no saved search term', () => {
```
marktnoonan, line 30: Most of the changes here are just nesting  according to whether a filter was saved or not.

mike-plummer (commented):
Very cool! I love the idea of consolidating the logic between the two implementations and extracting it out of the (overly-huge) specs list component. Just a couple small comments in the code and one higher-level thought below

* Since we know there might be a broader effort to provide Spec Filtering in the near future I wonder if we should structure the GQL/stateful side of things with a more specific name like "specNameFilter" so we can use "specFilter" in the future to group a collection of filters

packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -36,6 +36,7 @@ export interface ProjectApiShape {
   getCurrentBrowser: () => Cypress.Browser | undefined
   getCurrentProjectSavedState(): {} | undefined
   setPromptShown(slug: string): void
+  setSpecFilter(specFilter: string): void
```
lmiller1990, line 39: I am not sure we need new endpoints/code for this - can we just use the existing preferences code? It all ends up in the same object:

image

For example here: https://github.com/cypress-io/cypress/blob/develop/packages/data-context/src/actions/LocalSettingsActions.ts#L13-L16 (I think you implemented the original preference code too). It seems like we are adding a specific endpoint for one preference, when we can consolidate it in one place. What do you think?

lmiller1990 (commented):
Code works great, gave it a test. I think we are introducing some duplication though - we've got a pattern/code for updating preferences already (include GraphQL endpoint).

packages/data-context/src/actions/ProjectActions.ts
marktnoonan: At a glance I was expecting to do that - before picking this up we all thought this was just a trivial addition to this logic. But there are actually two different state objects and patterns: the existing `setPreference`/`localSettings` stuff manages values that apply to Cypress generally on the machine (stored in the `__global__/state.json`), and the one in your screenshot is in the project-specific `state.json`. 

We could refactor to have the same pattern for both kinds of settings, and just `setLocalPreference` and `setProjectPreference` though. I'd actually like that a lot but it seems like it would muddy this PR.

packages/server/lib/makeDataContext.ts
```diff
@@ -154,6 +154,12 @@ export function makeDataContext (options: MakeDataContextOptions): DataContext {
           },
         })
       },
+      async setSpecFilter (specFilter) {
+        await openProject.getProject()
+        ?.saveState({
+          specFilter,
+        })
```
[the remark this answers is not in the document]
marktnoonan: nah I can straighten that out, just left over from playing around with some other values in there.

packages/app/src/composables/useSpecFilter.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  watch(() => debouncedSearchString?.value, (newVal) => {
+    setSpecFilter(newVal ?? '')
+  })
+
+  // initialize spec filter in store
+  setSpecFilter(search.value)
+
+  return {
+    setSpecFilter,
```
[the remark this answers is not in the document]
marktnoonan: Good catch, initially I had exposed so that each component could use it, but then consolidated more logic into the composable, so I'll stop exposing this here.

packages/app/src/composables/useSpecFilter.ts
```diff
@@ -0,0 +1,43 @@
...
+    search,
+    debouncedSearchString,
```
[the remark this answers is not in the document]
marktnoonan: Good feedback - though I'm sure those are the ideal new names. I think `search` could maybe become `filter` (as it was changed elsewhere). `debouncedSearchValue` doesn't seem quite right since we'd go `debouncedSearchValue.value`. Maybe `specFilterModel` and `debouncedSpecFilterModel`. Gonna think on it a bit.

packages/data-context/src/actions/ProjectActions.ts
lmiller1990: It seems like this refactor would make a lot of sense though - if we don't do it now, I don't think it will ever happen, since it's very hard to justify going back and reworking code that already works great.

I'll leave this decision to you - I am not sure how much work it is, but a unified preferences API sure would be nice, instead of multiple code paths.

```js
setPreferences(json: any, type: 'global' | 'project') 
```

Or something to that meaning.

packages/app/cypress.config.ts
```diff
@@ -48,6 +48,10 @@ export default defineConfig({
         throw new Error('HTTP_PROXY_TARGET_FOR_ORIGIN_REQUESTS is missing. Close Cypress and run tests using the `yarn cypress:*` commands from the `packages/app` directory')
       }
 
+      on('test:after:run', (results) => {
+        /* ... */
```
lmiller1990, line 52: Can remove this?

lmiller1990 (commented):
One last  comment on some dead code.

I've also left a note around potentially consolidating all the preference code into a single place. It's more work but I think it's likely worth it for the long term health of the code base, but I will leave this final choice to @marktnoonan.

Once the last comment around removing some debugging code is fixed, I will :heavy_check_mark:

packages/app/cypress.config.ts
marktnoonan: lol, oh yes for sure 🤣

packages/data-context/src/actions/ProjectActions.ts
marktnoonan: After letting this sit for a bit, I started coming around as well. At a minimum, combining `setSpecFilter` and `setPromptShown` into `saveProjectPreferences' seems better than keeping them separated, since both work at the project level, and as Mike pointed out we can expect to maybe save other filtering criteria pretty soon. The only drawback is `setPromptShown` is a little more than just a simple save so we do have to shuffle things a little bit, maybe just created the object of the client instead of in `makeDataContext`.

What I really want to do is wrap this up and grab #21879, which together make a big improvement in the search experience. That pushes me towards what seems like the simpler solution, rather than fully combining both kinds of preferences... which, there's a bit to talk about there. Our names for things are a bit confusing and I'd be tempted to try to make it all make more sense.

If we come back with more things to save, we could combine them at that point. But really we don't seem to actually touch this code very much at the moment, values aren't added a lot.

marktnoonan: @mike-plummer @warrensplayer just wanted to mention I'll do this as part of the PR (should be ready to review tomorrow) and `saveProjectPreferences` should be useful for much of the ACI work, so might be worth chatting.

marktnoonan: Update, been exploring a bit around an _existing_ mutation called `setProjectPreferences`, which actually saves data in a third place, the `/Application Support/Cypress/cy/{environment}/cache` file. This file caches values for `lastBrowser` and `testingType` in objects that refer to each project by a simple string name, like `PROJECT_PREFERENCES['frontend-shared]` whereas the data in the projects folder is nested under more unique names with hashes (`frontend-shared-b1d9d76cfc89097c4beaf3412d05b7e0`), since two projects called `frontend-shared` could be in different directories and by opened by Cypress. So any cache based on project name alone is a little frail in my book.

The main purpose of that cache seems to be to maintain `USER` object with email address, auth token, and name, and the `PROJECTS` which is a list of global mode projects.

What this means:

- I'm not going to leverage that existing top-level cache, I'm going to continue to save data in the `state.json` under the `project+hash` directories where `firstOpened` etc are stored.
- Will rename `setProjectPreferences` to `setProjectPreferencesInGlobalCache` or something
- I'll make a chore ticket to move project-level saved settings out of that cache and keep it all together in the state.json (or have it explained why that's not the best solution)
- I will do Lachlan's suggestion of `setPreferences(json: any, type: 'global' | 'project')`, again to minimize confusion among all of these very related concepts.

packages/app/src/composables/useSpecFilter.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+`
+
+export function useSpecFilter (savedFilter?: string) {
+  const specStore = useSpecStore()
+  const saveSpecFilter = useMutation(SpecsList_SetSpecFilterDocument)
+
+  // prefer a filter from client side store, saved filter in gql can be stale
+  // and is only used to set the value in the store on first load
+  const initialFilter = specStore.specFilter ? specStore.specFilter : savedFilter ?? ''
```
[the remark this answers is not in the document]
marktnoonan: These were not equivalent since the `specFilter` in the store was initialized as an empty string (so, not nullish). Buuut I think it is probably better to distinguish the initial state in the store from the state set by this function, so I've set the initial state to `undefined`.

packages/app/src/composables/useSpecFilter.ts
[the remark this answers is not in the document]
marktnoonan: Went with `specFilterModel` and `debouncedSpecFilterModel`. When the time comes for more sophisticated filters `specFilterModel` will presumably become an object where the search term is just one field.

packages/app/src/specs/generators/FileChooser.vue
```diff
@@ -54,7 +54,7 @@
       <template #no-results>
         <NoResults
           empty-search
-          :search="noResults.search"
+          :search-term="noResults.search"
```
marktnoonan, line 57: Renamed this in passing since it's not always a `specFilterModel`. This component is used by several things (like file search, editor search). `searchTerm` seems a bit clearer to me than just search.

lmiller1990 (approved):

packages/graphql/schemas/schema.graphql
```diff
@@ -65,6 +65,12 @@ type CachedUser implements Node {
   id: ID!
 }
 
+"""Feature not available for subscription"""
```
mike-plummer, line 68: Ditto here with the schema changes

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
```diff
@@ -345,14 +346,15 @@ export const mutation = mutationType({
       },
     })
 
-    t.nonNull.field('setProjectPreferences', {
+    // TODO: #23202 hopefully we can stop using this for project data, and use `setPreferences` instead
+    t.nonNull.field('setProjectPreferencesInGlobalCache', {
```
mike-plummer, line 350: Worth marking this as deprecated? I believe there's a `deprecation` field we can define alongside the `description` to provide info on preferred alternative(s)

packages/graphql/schemas/schema.graphql
marktnoonan: Yeah committed here since this must have been left out of another PR in error, no need for us all to keep seeing these

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
marktnoonan: I think I'd like to actually poke at it a bit before being sure it is deprecated. Maybe there's some lifecycle reason that this is used that we can comment as the outcome of #23202.

mike-plummer (approved):

Issue #22611, opened 2022-06-30
`@cypress/vue@4.0.0` bundle is broken: `@vue/test-utils` is not externalized


0x009922:
### Current behavior

Currently `@cypress/vue` (v4) has a dependency:

```json
{
  "dependencies": {
    "@vue/test-utils": "2.0.0-rc.19"
  }
}
```

But if you explore actual `dist/cypress-vue.*.js` files, you will see that Vue Test Utils is not imported, but bundled. The only external import is `vue`.

### Desired behavior

`@cypress/vue` should import `@vue/test-utils`, not bundle it.

### Test code to reproduce

Bundle bug, not test code.

### Cypress Version

10

### Other

It is a huge problem because our tests rely heavily on setting global Vue Test Utils configuration to achieve global components registration per spec file. (https://test-utils.vuejs.org/api/#config-global)

For example, we do that trick:

```ts
import { config } from '@vue/test-utils'
import { mount } from '@cypress/vue'

before(() => {
  config.global.components = { SRadio, SRadioGroup }
})

after(() => {
  config.global.components = {}
})

it('just mounts', () => {
  cy.mount({ template: `<SRadioGroup />` })
})
```

Now it doesn't make any effect because `mount()` from `@cypress/vue` does not import `@vue/test-utils`, i.e. it is broken.

BTW, Cypress 10 is awesome!

lmiller1990:
Hi! Good point. If we import it, though, it's an additional peer dependency, which isn't really ideal.

We've got two options:

1. try to import it, then defer to bundled version.
2. bundle it and re-export it

I think 2 is the best option - this way, you just need to update to the latest `cypress` version, and you'll get all the updates for free. `cypress/vue` is generally running against the latest stable test utils version.

This would mean you'd do

```js
import { config, mount } from 'cypress/vue'
```

Also, if you are on Cypress 10, you don't even need to install `@cypress/vue` - we bundle it with Cypress, so just import it from `cypress/vue`. The goal is, instead of installing lots of different libraries, you just need to install and update Cypress, and we deal with the rest!

What do you think? Open to the first option too, but I think less peer dependencies is generally better for the end user, since you don't have to mess around with updating everything, just one dependency (cypress).

JessicaSachs:
One of the issues to consider with either solution is that `config` is a Vue 3-only concept. When we document the solution, this is only gonna be applicable to Vue 3.

[...]

PR #22757, opened 2022-07-13
feat(npm/vue): expose Test Utils API

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/22611
- Closes https://github.com/cypress-io/cypress/issues/22138

### User facing changelog

- Expose additional testing API from underlying Vue Test Utils library in the `cypress/vue` package.
- Inline `@vue/test-utils` type definitions, instead of relying on a peerDependency that might not be available, or the wrong version.

##  Expose additional testing API from underlying Vue Test Utils library in the `cypress/vue` package.
`cypress/vue` uses `@vue/test-utils` under the hood. Generally, users won't need to access this library, but there are some scenarios for advanced users where it might be useful, or where they are migrating some older jsdom + jest tests to use Cypress.

Exposing the underlying library might be useful for those scenarios. See #22611 for an examples from the community.

Generally, in Cypress, we do [global setup in a custom command](https://github.com/cypress-io/cypress/blob/1ccc90573ff0d2067db474fc6e407e446a4803e6/packages/frontend-shared/cypress/support/mock-graphql/mountFragment.ts#L23-L57). Some people may not like to use a custom command, or have some other use case in mind, so I think it's good to support alternative use cases, too.

I wouldn't consider this to be idiomatic - Cypress encourages use of custom commands where possible - so I don't think we will be recommending this in our main docs. Instead, I wrote a small note in the `@cypress/vue` README.

## Inline `@vue/test-utils` type definitions

We bundle `@vue/test-utils` entirely in the `cypress/vue` code we ship in the binary. TypeScript has no such "inline the types" feature, though, so the generated type definitions would do `import type {...} from '@vue/test-utils`.

This would give you a typing error, if you don't install a peer dependency (which we don't technically require, so it's confusing for users). In this PR, I wrote a little script to grab the types from the monorepo's `node_modules/@vue/test-utils/dist/**/*.d.ts` and paste them in `npm/vue/dist` when we build the library. This will let users get correct type definitions without needing the peer dependency. 

I also added some simple type tests to make sure it's working as expected.

### Steps to test

1. Verify problem - install cypress from npm, do `import { mount } from 'cypress/vue'` and follow it to the type def
2. You'll see they reference a library, `@vue/test-utils`, which you probably don't have installed (for example on line 3).

image

This is a problem! Verify the fix:

1. Grab this branch
2. `cd npm/vue`
3. `yarn build`
4. Look in `cli/vue/dist/index.d.ts`

Observe it's now referring to **relative** types - note the `'./` in from of `./@vue/test-utils` on line 3:

image

As for exposing the `VueTestUtils` API, you can see the test:

https://github.com/cypress-io/cypress/pull/22757/files#diff-f9a3bc77ab97b5f10f220de6ab30201be456aad25cb85e8197c52b2b90959684

### PR Tasks


npm/vue/package.json
```diff
@@ -8,24 +8,24 @@
     "cy:open": "node ../../scripts/cypress.js open --component --project ${PWD}",
     "cy:run": "node ../../scripts/cypress.js run --component --project ${PWD}",
     "build": "rimraf dist && rollup -c rollup.config.js",
-    "postbuild": "node ../../scripts/sync-exported-npm-with-cli.js",
-    "typecheck": "vue-tsc --noEmit",
+    "postbuild": "node --require @packages/ts/register ./inline-types.ts && node ../../scripts/sync-exported-npm-with-cli.js",
+    "typecheck": "yarn tsd && vue-tsc --noEmit",
     "test": "yarn cy:run",
+    "tsd": "yarn build && yarn tsc -p test-dts/tsconfig.tsd.json",
```
mike-plummer, line 14: I assume `tsd` stands for "typescript definitions", what is `dts` in the directory name meant to represent?

npm/vue/src/index.ts
```diff
@@ -1,17 +1,43 @@
...
   StyleOptions,
   getContainerEl,
   setupHooks,
 } from '@cypress/mount-utils'
 
+import * as _VueTestUtils from '@vue/test-utils'
+
+const {
+  mount: VTUmount,
+  // TODO: Should we expose shallowMount? It doesn't make much sense in the context of Cypress
```
mike-plummer, line 33: Is this really a TODO anymore? Seems like maybe just a comment is needed to explain why we're exporting even though it doesn't make a ton of sense for Cy

marktnoonan: Yeah I totally agree with exposing it for easy migration and keeping tests as close as possible before and after. Folks can modify their tests as a separate step when they have the need to see their components render fully.

npm/vue/package.json
lmiller1990: Good question, I actually have no idea, I copied it from this code base I work on a lot, where someone else set it up: https://github.com/vuejs/test-utils/tree/main/test-dts

I think `dts` is `.d.ts` as in a `foo.d.ts`, the definitions for `foo`.

I will change `test-dts` to `test-tsd` to be consistent.

npm/vue/src/index.ts
lmiller1990: Look carefully - we are **not** exposing it:

```js
const {
  mount: VTUmount,
  // TODO: Should we expose shallowMount? It doesn't make much sense in the context of Cypress
  // but it might be useful for people who like to migrate some Test Utils tests to Cypress.
  shallowMount,
  ...VueTestUtils
} = _VueTestUtils

export { VueTestUtils }
```

We destructure the properties we **do not** want to expose, and the rest are in the remaining `...VueTestUtils` object, which we export.

We should not expose `shallowMount` imo, since Cypress is supposed to encourage production-like testing, and `shallowMount` leads to all kind of weird, impossible-in-production scenarios.

We could expose it for migration purposes, I'd prefer to wait for someone to ask for this, though. I generally think we shouldn't expose it, since

- React and Angular don't have something similar
- If you do a `shallowMount`, most of Cypress (`cy.get` etc) won't work, since everything is stubbed out.

lmiller1990: @mike-plummer I will remove TODO and just say the reason we don't expose it.

lmiller1990: Done!

npm/vue/package.json
lmiller1990: Done

PR #22807, opened 2022-07-15
refactor: use Cypress design system icons in the SideBar

elevatebart:
### User-facing changelog

None (it's a refactor)

### Steps to test

- Open the app 
- Hover the Sidebar items
- Check that the icons colors change in sync with the text

https://user-images.githubusercontent.com/5592465/179316728-b2ed68d5-d55d-4825-87b1-edfff8bae0ce.mp4

Not

### PR Tasks


packages/app/src/navigation/SidebarNavigationRow.vue
```diff
@@ -23,9 +23,29 @@
...
-        class="flex-shrink-0 h-24px m-12px w-24px
-              children:transition children:duration-300"
+        v-if="active"
+        size="24"
+        stroke-color="indigo-300"
+        fill-color="indigo-700"
+        class="flex-shrink-0
+        h-24px m-12px w-24px
+        children:transition children:duration-300
+        "
```
emilyrohrbough, line 33:
```suggestion
        class="flex-shrink-0 h-24px m-12px w-24px children:transition children:duration-300"
```

packages/app/src/navigation/SidebarNavigationRow.cy.tsx
```diff
@@ -1,20 +1,22 @@
 import SidebarNavigationRow from './SidebarNavigationRow.vue'
-import BookIcon from '~icons/cy/book_x24'
-import SettingsIcon from '~icons/cy/settings_x24'
-import TestResultsIcon from '~icons/cy/test-results_x24'
-import PlaceholderIcon from '~icons/cy/placeholder_x24'
```
emilyrohrbough, line 5: Are we able to delete these svgs in the repo?

elevatebart: I meant bloat in time spent by me.
It would take long to manually find where these icons were used and check that they are not anymore.
Deleting icons could conflict with someone else's PR. They could be using the icons I deleted.
Since deleting icons is a "risky action," that would require me to ask everyone to check their use. 
I prefer to do it only once to only block everyone once. 

Sorry I was not clearer in the first comment.

elevatebart: The chance of the icons being used in another PR is small though.

I was not planning on deleting these icons just yet, but, if you prefer, I will check and delete them.

elevatebart: `test-results_x24.svg` is still used in the `EmptyGenerator.vue` component, I removed the 3 other files.

emilyrohrbough (approved):

packages/app/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "dependencies": {},
   "devDependencies": {
+    "@cypress-design/vue-icon": "^0.4.2",
```
lmiller1990, line 22: Unrelated but should this package be named `vue-icons`?

It looks like I'm importing a single icon, the Vue icon.

lmiller1990 (approved):

lmiller1990:
cc @elevatebart, need rebase 💯

packages/app/package.json
elevatebart: There are actually 2 forms of components in `vue-icon` here, and I think you are right:
We should split `vue-icon` and `vue-icons` into two separate packages.

- `vue-icon` is the non-tree-shakable prop-defined icon component.
- `vue-icons` is the set of generated components that allow us to import icon per icon. 

Since the two packages were sharing so much code internally, I made them one and the same. We could see about aliasing it in the future.

PR #22815, opened 2022-07-15
feat: Add icons to headers that have tooltips

### LINKED ISSUE #22638: Make spec data headers more obviously hoverable
### What would you like?

Update styling/content to make it more obvious that Specs Explorer headers provide tooltips with contextual actions (login, connect, request access buttons)

### Why is this needed?

Analytics indicates that users are not making use of contextual actions within Specs Explorer header tooltips (login, request access, connect project). We suspect that this is because users are unaware that the headers provide those tooltips & actions - current styling may not make it obvious that there is additional context to those items.

### Other

Ideas that have been floated:

1. Increase weight of header underline - we underline the "actionable" headers to indicate they provide tooltips, but current style is light gray and lightweight and users may not notice it

2. Add a "More Info" or "Question Mark" icon alongside header - having a "More Info" style icon (e.g., an "i" within a circle) may more clearly communicate there is more to be seen

3. Open header tooltips by default on first visit to Specs Explorer - probably the most obvious solution but has styling issues to overcome (overlap of tooltips)

4. Open header tooltips on mouseover of row content - for example, if logged out automatically open the header tooltip if user mouses over disabled latest runs status dots. This was prototyped during initial ACI dev and was shelved due to potential user annoyance concerns, but there may be alternative ways to structure this

mike-plummer:
- Closes #22638 

### User facing changelog
* Add icons to help indicate that column headers provide more information on hover

### Additional details
N/A

### Steps to test
1. Open a project in the app, navigate to the Specs Explorer
2. Validate that the "Last Updated", "Latest Runs", and "Average Duration" columns display a "question mark" icon
3. Validate that the tooltip for each column is triggered on hover of text *and* icon
4. Validate that columns do not overlap as viewport shrinks
    * Note: This depends on behavior being introduced in #22393 to automatically collapse the sidebar

https://user-images.githubusercontent.com/11482842/179297441-a1948f2d-698d-46a2-97ce-83792bc312dc.mov

### PR Tasks

- [na] Have tests been added/updated? _Covered by percy_

### DIFF
--- a/packages/app/src/specs/LastUpdatedHeader.vue
+++ b/packages/app/src/specs/LastUpdatedHeader.vue
@@ -6,10 +6,11 @@
   >
     <button
       type="button"
-      class="cursor-default decoration-dotted underline underline-gray-300 underline-offset-4 font-medium"
+      class="cursor-default flex font-medium items-center decoration-dotted underline underline-gray-300 underline-offset-4"
       data-cy="last-updated-header"
     >
       {{ t('specPage.lastUpdated.header') }}
+      <i-cy-action-question-mark-outline_x16 class="ml-1 icon-dark-gray-600" />
     </button>
     <template
       #popper
--- a/packages/app/src/specs/LastUpdatedHeader.vue
+++ b/packages/app/src/specs/LastUpdatedHeader.vue
@@ -10,13 +10,13 @@
       data-cy="last-updated-header"
     >
       {{ t('specPage.lastUpdated.header') }}
-      <i-cy-action-question-mark-outline_x16 class="ml-1 icon-dark-gray-600" />
+      <i-cy-circle-bg-question-mark_x16 class="ml-1 icon-dark-indigo-500 icon-light-indigo-100" />
     </button>
     <template
       #popper
     >
       <div
-        class="flex flex-col text-sm text-center max-w-300px p-4 items-center"
+        class="flex flex-col border-t-indigo-400 border-t-4px text-sm text-center max-w-300px p-4 items-center"
       >
         <i18n-t
           v-if="props.isGitAvailable"
--- a/packages/app/src/specs/LastUpdatedHeader.vue
+++ b/packages/app/src/specs/LastUpdatedHeader.vue
@@ -16,7 +16,7 @@
       #popper
     >
       <div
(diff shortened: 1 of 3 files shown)


ZachJW34 (approved):
🚀

marktnoonan (commented):
Screen Shot 2022-07-15 at 4 11 56 PM

It seemed like we modified the wording etc previously to keep this breakpoint looking good with the headers, should we hide these icons at small breakpoints ([percy](https://percy.io/cypress-io/cypress/builds/19667976/changed/1105038972?browser=chrome&browser_ids=23&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=500&widths=500%2C536%2C650%2C800%2C1000%2C1200%2C1280%2C1500%2C2000))? 

The only other comment I have is really a design thing, at standard zoom I can't tell that these are question marks, might just be my eyesight but the icon seems so close to the circle that it's difficult to see exactly what's going on. It's the only thing in our UI I have to zoom in to understand so figured I'd mention it.

mike-plummer:
@marktnoonan Good callouts
1. Percy is actually in a weird situation with this - in the app this section of the UI has a minimum effective width that prevents the specs list from shrinking this small (it stops around the 600px mark), so the situation displayed in Percy can't be reached at runtime. I agree that the icons cause an issue at narrower widths, but currently those widths can't be reached.
2. Design is taking another look at the icons to see if using a different variant of the question mark might be preferable

marktnoonan:
Hmm - let's remove this test from Percy then. No need to track an impossible state. Although come to think of it, users may not be able to reduce the app to this size, but they can increase the zoom in the browser which brings breakpoints like this into play even at bigger sizes. We're not systematic about handling zoom yet, so I don't mind if we remove the icon or remove the snapshot, but I'd like to do one or the other so that percy stays accurate in documenting intended states (at least in CT, where that is possible).

marktnoonan:
I think based on some light manual testing that, though I can reach this state now, I won't be able to reach it when the nav menu auto closes at those small sizes:

Screen Shot 2022-07-15 at 6 19 35 PM

So I think it's fine to remove this from Percy.

marktnoonan (approved):
Looks great to me, I like the little bit of color up there too.

PR #22822, opened 2022-07-15
chore: only run Windows on main branches

flotwig:
- For https://github.com/cypress-io/cypress-internal/issues/461

### Additional details

- Removes Windows from PRs and branch builds by default, cuz It's expensive, yo. And we don't need to do this as badly now that we aren't rapidly iterating on 10.0 pre-releases. We will be forced to reckon with Windows every 2 weeks like in the pre-10.0 days.

### PR Tasks

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -58,16 +58,16 @@ linuxWorkflowExcludeFilters: &linux-x64-workflow-exclude-filters
     - false
     # - equal: [ 'tgriesser/chore/fix-windows-build', << pipeline.git.branch >> ]
 
-# windows is slow in CI, so we only run a certain set of tests on each commit
+# windows is slow and expensive in CI, so it normally only runs on main branches
 # add your branch to this list to run the full Windows build on your PR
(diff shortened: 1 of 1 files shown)


tbiethman (approved):

circle.yml
```diff
@@ -2671,22 +2671,19 @@ windows-workflow: &windows-workflow
...
         name: windows-lint
         executor: windows
         requires:
           - windows-build
 
     - unit-tests:
-        <<: *full-windows-workflow-filters
         name: windows-unit-tests
         executor: windows
         resource_class: windows.large
```
ryanthemanuel, line 2682: Do we want to back this off to medium?  Or do we want to explore that separately?

ryanthemanuel (approved):

circle.yml
flotwig: Right now it looks like it's still too flaky :confused: https://github.com/cypress-io/cypress/pull/22821

ryanthemanuel: Sadness. Oh well this is a good option for now! 👍🏻

PR #22838, opened 2022-07-18
fix: Determine global mode based on whether there is a current project

chrisbreiding:
### User facing changelog

N/A - fixing an issue pre-release

### Additional details

[#22073](https://github.com/cypress-io/cypress/pull/22073/files#diff-a29a2817a2313d8351e2c932721e30fc122da6f8189a558bb7beadd15443172c) changes the logic for determining if the Launchpad is running in global mode, making it solely based on whether the `--global` flag is passed via the command line. In production, running `cypress open` results in a falsey value for the `--global` flag and so the Launchpad displays the welcome page instead of the global mode projects list.

Global mode should be determined by whether a project is specified or not. The `--global` flag can be used to force global mode, but if there is not a project, we can assume the Launchpad is being run in global mode.

### Steps to test

1. If you've installed a version of 10.3.1 before and have cypress installed globally, run `cypress cache clear`
2. Install globally via the latest pre-release posted in this PR (`npm install -g https://cdn.cypress.io/.../cypress.tgz`)
3. If you've installed a version of 10.3.1 before and didn't have cypress installed globally, run `cypress cache clear && cypress install` to ensure the latest binary from this PR is installed.
4. Run `cypress open`
5. The 'Add Project' or projects view should be shown

### PR Tasks

### DIFF
--- a/cli/lib/exec/open.js
+++ b/cli/lib/exec/open.js
@@ -16,7 +16,7 @@ const { exitWithError } = require('../errors')
  */
 const processOpenOptions = (options = {}) => {
   // In addition to setting the project directory, setting the project option
-  // here ultimately decides whether cypress si run in global mode or not.
+  // here ultimately decides whether cypress is run in global mode or not.
   // It's first based off whether it's installed globally by npm/yarn (-g).
   // A global install can be overridden by the --project flag, putting Cypress
   // in project mode. A non-global install can be overridden by the --global


marktnoonan (commented):
Testing the binary from this, this doesn't seem to get there.

- `cypress open` -> thinks current directory is a project always?
- downloading zip -> works great
- opening with --global flag -> fine too

packages/graphql/src/schemaTypes/objectTypes/gql-Query.ts
```diff
@@ -79,7 +79,7 @@ export const Query = objectType({
 
     t.nonNull.boolean('isGlobalMode', {
       description: 'Whether the app is in global mode or not',
```
rachelruderman, line 81: > Global mode should be determined by whether a project is specified or not. The --global flag can be used to force global mode, but if there is not a project, we can assume the Launchpad is being run in global mode.

@chrisbreiding Gah I was missing this context for global mode when I did [#22073](https://github.com/cypress-io/cypress/pull/22073/files#diff-a29a2817a2313d8351e2c932721e30fc122da6f8189a558bb7beadd15443172c). Would you mind adding your explanation to the description here so it doesn't get lost in the future? 🙏

chrisbreiding:
> Testing the binary from this, this doesn't seem to get there.
> 
>     cypress open -> thinks current directory is a project always?
>     downloading zip -> works great
>     opening with --global flag -> fine too

@marktnoonan That's odd. After installing the pre-release binary from this PR, `cypress open` works properly for me. Did you clear your cypress cache before installing the pre-release? If not, it will use the cached version if you have one installed.

marktnoonan:
@chrisbreiding it wasn't cache but it turns out I had a wrong assumption about how this was expected to work - the new comments are helpful and after retesting, yeah this is the correct behavior now and doesn't enter the in-between state that develop opens to.

marktnoonan:
Hmm, the problem this work was intended to deal with seems to have resurfaced: the "Projects" link should still be in the top left here, since we did open in global mode and users can go back there. So "global mode", as far as the UI is concerned, can't only be determined by whether there is a project at the moment - that just tells the launchpad to be on the global _page_ and as soon as a project is selected, we forget that we are running in global mode, which has effects that last long.

Screen Shot 2022-07-19 at 9 09 18 AM

I suspect this is why that `projectRootFromCI` value that was removed in [#22073](https://github.com/cypress-io/cypress/pull/22073/files#diff-a29a2817a2313d8351e2c932721e30fc122da6f8189a558bb7beadd15443172c) was introduced in the first place.

chrisbreiding:
I've fixed that issue in [`0507401` (#22838)](https://github.com/cypress-io/cypress/pull/22838/commits/0507401030a146d4e37b4e4b609d757a385463f5). Using `ctx.modeOptions.project` instead of `ctx.currentProject` keeps `isGlobalMode` from changing once a project is selected.

cli/lib/exec/open.js
```diff
@@ -15,6 +15,12 @@ const { exitWithError } = require('../errors')
  * @returns {string[]} list of CLI arguments
  */
 const processOpenOptions = (options = {}) => {
+  // In addition to setting the project directory, setting the project option
+  // here ultimately decides whether cypress si run in global mode or not.
```
marktnoonan, line 19: SUPER helpful comment, tiny typo.

```suggestion
  // here ultimately decides whether cypress is run in global mode or not.
```

marktnoonan (approved):

emilyrohrbough (approved):

PR #22856, opened 2022-07-19
chore: Make improvements to release process doc and binary upload script

chrisbreiding:
### User facing changelog

N/A - internal only

### Additional details

1. Makes the following improvements to release-process.md:
	- Moves the AWS SSO setup to prerequisites since it should only need to be done once
	- Adds a step to test out the pre-release binary before moving it to beta or publishing to npm. This is the same as the later smoke test step after publishing to npm. It can help catch any issues that might invalidate the release after it's too late because it's taken up a version on npm.
	- Adds a note that the merge from develop to master should be a merge commit and not a squash commit.
	- Replaces the release-automation set-releases step with manual GitHub release creation since the release automation for it is broken.

2. Updates the binary upload script to use the list of valid platforms+arches when purging the cache. Before, it used the user's local arch, which meant it tried and failed to purge the cache for `win32-arm64`, which is not a valid configuration.

### PR Tasks

### DIFF
--- a/guides/release-process.md
+++ b/guides/release-process.md
@@ -21,10 +21,6 @@ The `@cypress/`-namespaced NPM packages that live inside the [`/npm`](../npm) di
     [prod]
     sso_start_url = <start_url>
     sso_region = <region>
-    sso_account_id = <account_id>
-    sso_role_name = <role_name>
-    region = <region>
-    cli_pager = <cli_pager>
     aws_access_key_id = <access_key_id>
     aws_secret_access_key = <secret_access_key>
     aws_session_token = <session_token>
--- a/guides/release-process.md
+++ b/guides/release-process.md
@@ -185,7 +185,7 @@ In the following instructions, "X.Y.Z" is used to denote the [next version of Cy
 24. Check all `cypress-test-*` and `cypress-example-*` repositories, and if there is a branch named `x.y.z` for testing the features or fixes from the newly published version `x.y.z`, update that branch to refer to the newly published NPM version in `package.json`. Then, get the changes approved and merged into that project's `master`. For projects without a `x.y.z` branch, you can go to the Renovate dependency issue and check the box next to `Update dependency cypress to X.Y.Z`. It will automatically create a PR. Once it passes, you can merge it. Try updating at least the following projects:
     - [cypress-example-todomvc](https://github.com/cypress-io/cypress-example-todomvc/issues/99)
     - [cypress-example-todomvc-redux](https://github.com/cypress-io/cypress-example-todomvc-redux/issues/1)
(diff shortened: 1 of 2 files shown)


guides/release-process.md
```diff
@@ -15,6 +15,21 @@ The `@cypress/`-namespaced NPM packages that live inside the [`/npm`](../npm) di
   - Permissions for your npm account to publish the `cypress` package.
   - Permissions to update releases in ZenHub.
 
+- [Set up](https://cypress-io.atlassian.net/wiki/spaces/INFRA/pages/1534853121/AWS+SSO+Cypress) an AWS SSO profile with the [Team-CypressApp-Prod](https://cypress-io.atlassian.net/wiki/spaces/INFRA/pages/1534853121/AWS+SSO+Cypress#Team-CypressApp-Prod) role. The release scripts assumes the name of your profile is `prod`. Make sure to open the "App Developer" expando for some necessary config values. You AWS config file should end up looking like the following:
+
+    ```
```
emilyrohrbough, line 20: Mine doesn't have all of these configured and it works as expected. IMO we shouldn't add this since the Wiki pages provides it and the infra team will make updates when needed.

guides/release-process.md
```diff
@@ -148,26 +172,21 @@ In the following instructions, "X.Y.Z" is used to denote the [next version of Cy
...
 
-        ```shell
-        cd packages/issues-in-release && npm run do:comment -- --release X.Y.Z
-        ```
+    ```shell
+    cd packages/issues-in-release && npm run do:comment -- --release X.Y.Z
+    ```
 
-    - Confirm there are no issues with the label [stage: pending release](https://github.com/cypress-io/cypress/issues?q=label%3A%22stage%3A+pending+release%22+is%3Aclosed) left
+22. Confirm there are no issues with the label [stage: pending release](https://github.com/cypress-io/cypress/issues?q=label%3A%22stage%3A+pending+release%22+is%3Aclosed) left
```
emilyrohrbough, line 185: This honestly might be a good "first" step when verifying everything has been tagged for the release and wasn't missed. 🤷🏻‍♀️

emilyrohrbough (approved):

guides/release-process.md
chrisbreiding: This is covered as a first step in step #2. I think it's good to have it as a post-release step as well as it can be easy to miss something initially when there are 30 issues to go through before the release. I caught a couple today by checking the pending release tag after the release.

guides/release-process.md
emilyrohrbough: Would it be better to update the links & wiki to make it clear what was missing?

mjhenkes (approved):

PR #22314, opened 2022-06-14
feat(webpack-dev-server): add angular handler

jordanpowell88:
- Closes #22049
- Closes #22812

### User facing changelog
Adds Angular framework support for DevServer.

### Additional details
This dev-server is based on `@angular/cli` created angular projects. It sources the webpack config from the same build process used to build angular/cli projects which allows us to respect the build options that users have in their `angular.json` and tsconfig.

### Steps to test
There are system tests in `npm/webpack-dev-server` that run against an Angular 13 and 14 application. To test this on an Angular application outside of system tests, reference the `system-tests/project-fixtures/angular` angular project fixture. Since we don't have `mount` merged, there is a local version that should be used instead.

### How has the user experience changed?

https://user-images.githubusercontent.com/25158820/180316864-0a8f938b-d172-4f5a-ad61-c9a3aa98aaa3.mov

### PR Tasks

- [x] Have tests been added/updated?
- [ ] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [ ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [x] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?

PR #22858, opened 2022-07-19
feat(angular): angular mount

### LINKED ISSUE #22417: `node_modules/cypress/types/net-stubbing.ts` should be a `.d.ts` file
### Current behavior

When using Cypress as a library, there is one file called `node_modules/cypress/types/net-stubbing.ts` which only contains type definitions.
Together with latest version of API extractor, this creates the error:
```
Error: <project>/node_modules/cypress/types/net-stubbing.ts:1:1 - (ae-wrong-input-file-type) Incorrect file type; API Extractor expects to analyze compiler outputs with the .d.ts file extension. Troubleshooting tips: https://api-extractor.com/link/dts-error
```

[Error description](https://api-extractor.com/pages/messages/ae-wrong-input-file-type/) advises to report the problem to the package maintainers…

This file seems to be created during build in the [post install](https://github.com/cypress-io/cypress/blob/e6b2466f7b219a86da46c1ac720432ef75193ca4/cli/scripts/post-install.js#L76-L78) script.

The error is not detected by older versions of API extractor (not present with 7.20.1 but present with 7.25.2).
This file was already here in older versions of Cypress (I can see the error on Cypress 8.4.1).

### Desired behavior

Cypress should export its typing in `.d.ts` files rather than in `.ts` ones.

### Test code to reproduce

[Github action failing on that problem](https://github.com/Siteimprove/alfa/runs/6983182879?check_suite_focus=true#step:8:6).

### Cypress Version

8.4.1, 10.1.0

### Other

_No response_

### LINKED ISSUE #22813: Add Angular mount package
In order to support Angular for Component Testing, we need to add an Angular mounting package. We already have an `npm/angular` package, so replacing/updating this package would be best (it was never officially published).

The responsibilities of this package are:
- Mount a compiled Angular component into the DOM
- Provide options for customizing the mounted components behavior (ngModule, providers, directives etc...)

Angular users are already familiar with unit-testing Angular components. Angular provides testing support [out of the box](https://angular.io/guide/testing-components-scenarios), so leveraging the existing APIs would be great as users who are familiar with unit testing would be able to quickly pick up Cypress Angular component testing. The caveat here is that Angular uses a [custom compiler](https://github.com/angular/angular/blob/main/packages/core/testing/src/test_bed_compiler.ts), so we should investigate if using this API would produce any differences in components mounted in dev vs Cypress.

This package should be copied to the `cli` package so that users are able to import this package via `import { mount } from "cypress/angular"`.

Requirements:
- Should mount Angular components via an exported `mount` function
- Should allow Angular components to be customized in order to provide the component with the proper context (ngModule, providers, directives etc..)
- Should be copied to the `cli` package to support import from `cypress/angular`
- Should add system tests proving that Angular components can be mounted

There is an [external package](https://github.com/jordanpowell88/angular-ct) created by our very own @jordanpowell88.

There is active discussion on [versioning CT mounting libs](https://github.com/cypress-io/tech-briefs/pull/35) that will affect how we publish and version this package

jordanpowell88:
- Closes #22813
- Closes #22417

### User facing changelog
Adds `@cypress/angular` mount 

### Additional details
- This is necessary to enable component testing in Angular. 
- It is based upon [https://github.com/jordanpowell88/angular-ct](https://github.com/jordanpowell88/angular-ct)

### Steps to test
We can use system tests from #22314 to validate this work

Nothing changes from a user experience as this is a new feature

### PR Tasks
- [ na] Have tests been added/updated?
- [ na] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?


cli/angular/README.md
```diff
@@ -0,0 +1,154 @@
+> A little helper to unit test React components in the open source [Cypress.io](https://www.cypress.io/) test runner **v7.0.0+**
```
lmiller1990, line 1: We normally gitignore the `cli/*` packages, and don't commit them, since they are generated at build time. You can remove these and add them to `gitignore`.

npm/angular/src/mount.ts
```diff
@@ -0,0 +1,211 @@
...
+ *  })
+ *  cy.get('h1').contains('Hello World')
+ * })
+ * @returns Cypress.Chainable<MountResponse<T>>
+ */
+export function mount<T extends object> (
+  component: Type<T>,
+  config: TestBedConfig<T> = {},
+  autoDetectChanges = true,
+): Cypress.Chainable<MountResponse<T>> {
```
lmiller1990, line 201: Our other adapters (react, vue) both accept some additional args that let you inject some stylesheets. I wonder if this is something that would also be useful for angular?

This is implemented using `@cypress/mount-utils`.

React example: https://github.com/cypress-io/cypress/blob/7b58800166f830fd819c55fa91ba1196537928ec/npm/react/src/mount.ts#L4-L10

npm/angular/src/mount.ts
```diff
@@ -0,0 +1,211 @@
...
+): T {
+  let component: T = fixture.componentInstance
+
+  if (config?.inputs) {
+    component = Object.assign(component, config.inputs)
+  }
+
+  return component
+}
+
```
lmiller1990, line 176: How do we actually mount the component? React/Vue use a `data-cy-root` selector which they import https://github.com/cypress-io/cypress/blob/develop/npm/react/src/mount.ts#L8

I don't see how we go from component -> mount to DOM here. Am I missing it, or is that implemented in another PR?

https://github.com/cypress-io/cypress/pull/22858/files#diff-617e9faa6b88ce75b6ca3135c54ae7000965719cc2d6c9ae9813178d9da19170

lmiller1990 (commented):
Couple of small questions! Excited for this.

npm/angular/src/mount.ts
jordanpowell88: This is a behavior that you cannot do in Angular. You cannot just import a stylesheet at the component level. You have global stylesheets which will be brought in and added to the `webpack` compilation via the users `angular.json` file, and the component themselves stylesheet

cli/angular/README.md
jordanpowell88: Good catch!

npm/angular/src/mount.ts
jordanpowell88: By default Angular mounts the component for you to the DOM using the `angular.json` build.options.index property. In [THIS PR](https://github.com/cypress-io/cypress/pull/22314) we change this value from what the user has to `cypress/support/component-index.html`. It is then in that file (that we will need to generate as part of the migration work) that angular will mount the component

lmiller1990 (approved):
My questions were answered :+1: 

Did not try locally yet but code is :100:

lmiller1990:
Seems linting and a few other steps are failing, might be good to get those green before any other reviews.

circle.yml
```diff
@@ -1791,14 +1791,6 @@ jobs:
       - run:
           name: Build
           command: yarn workspace @cypress/angular build
-      - run:
```
ZachJW34, line 1794: No tests for `npm/angular`. To test this, we would need to install all of the webpack deps for Angular so all of the mounting tests are found in the `system-tests/project-fixtures/angular/src/app/mount.cy.ts` which are run in `npm-webpack-dev-server` e2e tests.

npm/angular/README.md
```diff
@@ -1,221 +1,154 @@
-# @cypress/angular
```
ZachJW34, line 1: Should clean this up to be a minimal README like the other mount packages since most of this information will be in the official docs.

npm/angular/package.json
```diff
@@ -1,95 +1,67 @@
...
-    "app-build": "ng build",
-    "test": "yarn cy:run",
-    "test-ci": "echo not testing in CI because process never exits",
-    "postinstall": "patch-package"
-  },
-  "dependencies": {
-    "@cypress/mount-utils": "0.0.0-development",
-    "debug": "^4.3.2"
+    "prebuild": "rimraf dist",
+    "build": "tsc || echo 'built, with type errors'",
```
ZachJW34, line 8: Realized that rollup is overkill for this package. The version of Angular we support expect `esm` packages so we only need to publish one entrypoint, so using `tsc` directly is sufficient.

npm/angular/src/mount.ts
```diff
@@ -0,0 +1,225 @@
+import 'zone.js'
```
ZachJW34, line 1: Might want to move side-effects into a separate file, though this would increase the complexity of scaffolding during launchpad setup.

cli/types/tslint.json
```diff
@@ -38,7 +38,10 @@
       "./mocha/*",
       "./minimatch/*",
       "./sinon/**/*",
-      "./sinon-chai/*"
+      "./sinon-chai/*",
+      // Copied from net-stubbing and renamed to a declaration file. Since it's not originally
+      // a declaration file, we need to exclude it from linting.
+      "./net-stubbing.d.ts"
```
ZachJW34, line 44: Fixed #22417 as a part of this work. Angular crawls all of the imported types and emits warnings due to this file not being included in the compilation. `.d.ts` files are fine and it's what this file should be anyway. I had to ignore this file in the `tslint.json` as `dtslint` complains. Since this file is originally a normal ts file, fixing the dtslint warnings causes ts warnings, so ignoring seemed best.

npm/angular/src/mount.ts
lmiller1990: I think this illustrates the problem of trying to have no side effects - this import makes perfect sense here, it's required for Angular, right?

An alternative would be scaffold it in `component.ts`, but then we increase the complexity of `component.js` for... not much benefit. imo. Also, is there a downside to having people import/bundle `zone.js` in E2E? (If they don't want to, they just shouldn't mix their E2E and CT support code).

ZachJW34: There shouldn't be any issues for importing `zone.js` in e2e tests. I was harping on not having side effects a while ago but now I would rather have side-effects if it means for a cleaner support file and less complexity during project setup.

jordanpowell88:
2 things I am noticing after doing some additional testing with the binary is that we don't have a good way to support the following 2 scenarios:
- content project (ie: `<ng-content></ng-content>`
- Outputs: (we need a way to spy on the emitted events)

ryanthemanuel (approved):

PR #22861, opened 2022-07-20
chore: fix test by normalizing stdout

lmiller1990:
- Closes N/A

### User facing changelog

N/A

### Additional details

Test snapshot hard codes Cypress version, so it will always fail. Now it normalizes it correctly, so it won't fail as we release new Cypress versions.

### Steps to test

Observe CI is passing.

### PR Tasks

### DIFF
--- a/packages/server/test/integration/cypress_spec.js
+++ b/packages/server/test/integration/cypress_spec.js
@@ -7,6 +7,7 @@ const Promise = require('bluebird')
 const electron = require('electron')
 const commitInfo = require('@cypress/commit-info')
 const Fixtures = require('@tooling/system-tests')
+const SystemTests = require('@tooling/system-tests/lib/system-tests')
 const snapshot = require('snap-shot-it')
 const stripAnsi = require('strip-ansi')
 const pkg = require('@packages/root')
@@ -96,7 +97,9 @@ const snapshotConsoleLogs = function (name) {
   // so must switch back to original
   process.chdir(previousCwd)
 
-  return snapshot(name, stripAnsi(args))
+  const snap = SystemTests.default.normalizeStdout(stripAnsi(args))
+
+  return snapshot(name, snap)
 }
 
 function mockEE () {
--- a/packages/server/test/integration/cypress_spec.js
+++ b/packages/server/test/integration/cypress_spec.js
@@ -7,7 +7,7 @@ const Promise = require('bluebird')
 const electron = require('electron')
 const commitInfo = require('@cypress/commit-info')
 const Fixtures = require('@tooling/system-tests')
-const SystemTests = require('@tooling/system-tests/lib/system-tests')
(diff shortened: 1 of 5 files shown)


packages/server/test/integration/cypress_spec.js
```diff
@@ -7,6 +7,7 @@ const Promise = require('bluebird')
 const electron = require('electron')
 const commitInfo = require('@cypress/commit-info')
 const Fixtures = require('@tooling/system-tests')
+const SystemTests = require('@tooling/system-tests/lib/system-tests')
```
flotwig, line 10: I think this `require` changes the process's working directory, which is why these tests are failing: https://app.circleci.com/pipelines/github/cypress-io/cypress/40774/workflows/983f4e5c-d12e-488f-9e43-c455632b2da0/jobs/1683656

The `process.cwd` call could be relocated, or `normalizeStdout` could be moved to a module that doesn't call `process.cwd`.

flotwig (changes requested):

packages/server/test/integration/cypress_spec.js
lmiller1990: Yep, was just looking into why this was failing. I'll make that refactor.

Ideally I'd have liked to have written this in `system-tests`, but it's quite hard to get the stdout to the correct state w/o a lot of stubbing, which system tests don't support for obvious reasons.

lmiller1990: Lol, hate to see it: https://github.com/cypress-io/cypress/blob/develop/system-tests/lib/system-tests.ts#L31. Always good to fix a TODO.

lmiller1990: Actually, problem is here: https://github.com/cypress-io/cypress/blob/develop/system-tests/lib/spec_helper.js (somewhere). I will find it.

lmiller1990: This is to blame for at least some of the problems: https://github.com/cypress-io/cypress/blob/develop/system-tests/lib/spec_helper.js#L83-L86

lmiller1990: I think we probably **don't** want all these side effects in `packages/server`. I'll just move the normalizeStdout function to a file, without any side effects.

system-tests/lib/normalizeStdout.ts
```diff
@@ -0,0 +1,151 @@
+import Fixtures from './fixtures'
+import _ from 'lodash'
```
lmiller1990, line 2: Copy pasted and threw in some types

packages/server/test/integration/cypress_spec.js
lmiller1990: Ok, I did it.

PR #22862, opened 2022-07-20
feat: adding svelte support for Cypress

JessicaSachs:
### Context

I was set to give a demo of component testing with Svelte tomorrow as a preview for the official Alpha release/support in late September. Is there an open issue for Svelte support already? I know it's on the roadmap.

- Closes 

### User facing changelog

Adding Alpha support for Vanilla Svelte (this does not add support for SvelteKit).

### Additional details

The Svelte Testing Library was my reference implementation.

### Steps to test

1. Create a new Svelte project
2. Try to mount and test Svelte components

We need to add gifs to the docs.

### PR Tasks

**TODOS:** We need a Svelte testing guide in the docs and we need to test vanilla Svelte against a variety of library versions using System Tests. Additionally, we need to fix the types for the mounting libraries.


cli/package.json
```diff
@@ -141,6 +142,11 @@
       "require": "./react/dist/cypress-react.cjs.js",
       "types": "./react/dist/index.d.ts"
     },
+    "./svelte": {
```
lmiller1990, line 145: I looked into this - these entries actually do not do anything from what I can tell, or, at least, they are not respected by webpack 4. I am pretty sure we don't need these at all.

What's actually happening is the bundlers look at `node_modules/cli/{react,vue}` and look inside for `package.json`, and resolve from there.

npm/svelte/cypress.config.ts
```diff
@@ -0,0 +1,24 @@
...
+  responseTimeout: 2500,
+  projectId: '134ej7',
+  experimentalFetchPolyfill: true,
+  e2e: {
+    supportFile: false,
+  },
+  component: {
+    devServer: {
+      bundler: 'vite',
+      framework: 'vue',
```
lmiller1990, line 16: `svelte`?

lmiller1990 (commented):
Thanks for this!! Someone from @cypress-io/component-testing will need to add the handler to webpack/vite, etc.

cli/package.json
JessicaSachs: They're for types and Vite. If you take them out, rebuild, and try to do a deep import from `cypress/vue`, Vite will say that it cannot find that module.

npm/svelte/cypress.config.ts
JessicaSachs: Yeah. I missed one.
```suggestion
      framework: 'svelte',
```

JessicaSachs:
Hmm, the `props` aren't getting passed through even though we're matching the API for testing-library. I'm surprised by that.

npm/svelte/src/index.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    el.append(componentNode)
+
+    // mount the component using VTU and return the component on Cypress.svelteComponent
+    const ComponentConstructor = Component.default || Component
+
+    const component = Cypress.svelteComponent = new ComponentConstructor({
+      target: componentNode,
+      options,
+    })
```
PatrickG, line 96: > Hmm, the `props` aren't getting passed through even though we're matching the API for testing-library. I'm surprised by that.

Saw your talk on SvelteSociety.
There is no `options` property on the [`options` parameter of a svelte component constructor](https://svelte.dev/docs#run-time-client-side-component-api-creating-a-component).

I guess this should be:
```ts
    const component = Cypress.svelteComponent = new ComponentConstructor({
      ...options,
      target: componentNode,
    })
```

JessicaSachs: _Thank you_!

npm/svelte/src/index.ts
```diff
@@ -0,0 +1,163 @@
...
+    componentNode.id = '__cy_svelte_root'
+
+    el.append(componentNode)
+
+    // mount the component using VTU and return the component on Cypress.svelteComponent
+    const ComponentConstructor = Component.default || Component
+
+    const component = Cypress.svelteComponent = new ComponentConstructor({
+      target: componentNode,
+      options,
```
JessicaSachs, line 95:
```suggestion
      ...options,
```

ZachJW34:
@JessicaSachs I'm picking up Svelte support! Probably going to merge this work into a feature branch at some point

Issue #22655, opened 2022-07-01
Create Flaky data tooltip component


mike-plummer:
https://www.figma.com/file/DgeFrhaX8N9ufrzvxzLnPg/app-cloud-int%3A-flake-tooltip?node-id=16635%3A148374

![Screen Shot 2022-07-01 at 11 31 35 AM](https://user-images.githubusercontent.com/11482842/176934667-90627335-49ac-42a0-aa91-85525a5f55a5.png)

**Work Items**
1. Create tooltip content layout (similar to existing layout in `SpecRunSummary.vue` for Latest Runs data)
2. Integrate "graph" icons for displaying flaky level
3. Accept data via props that can be mapped to GQL data supplied by cloud
4. Create component tests

**Acceptance Criteria**
1. Tooltip component is created and matches Figma designs
    * Note that color shading is impacted by flaky rate "level" (low/medium/high)
2. Component tests are created and passing

baus:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/22655?planning-poker) with ZenHub @astone123 @marktnoonan @mike-plummer @lmiller1990 @ZachJW34

lmiller1990:
Not sure if we've got a pre-existing popover/bubble component. It would be good to have one we can reuse.

mike-plummer:
> Not sure if we've got a pre-existing popover/bubble component. It would be good to have one we can reuse. Making the little downwards pointing triangle bit with the correct box shadow is probably harder than anyone is expecting, I don't think you can have a box shadow that follows an outline like this.

We have something very similar in the "Latest Runs" column added in ACI phase 1 using the `Menu` component from `floating-vue`. My plan is to see about abstracting that a bit to provide a common base since it's visually very similar

PR #22873, opened 2022-07-20
chore: Build Flaky Badge and tooltip content

### LINKED ISSUE #22654: Create Flaky badge component
https://www.figma.com/file/DgeFrhaX8N9ufrzvxzLnPg/app-cloud-int%3A-flake-tooltip?node-id=16539%3A146227

Screen Shot 2022-07-01 at 11 28 41 AM

Badge will display if a spec retrieves data from the cloud indicating it is flaky. It will provide a tooltip on hover/focus (tracked separately under #22655) with additional information, and should be actionable by linking directly to flaky data in the dashboard using a url provided by the cloud

**Work Items**
1. Create Flaky Badge component to match Figma designs (see above)
2. Create component tests for Flaky Badge

**Acceptance Criteria**
1. Flaky Badge component is created with appropriate component tests
4. Component matches Figma designs (see above)

mike-plummer:
* Build Flaky Badge component
* Build Flaky status summary component
* Extract common component from latest runs tooltip

- Closes #22654, #22655 

This is Part 1 of 3 to add Flaky test data to the app as part of App-Cloud Integration (ACI). We are splitting this up to avoid another "mega" PR like we had for ACI Phase 1.
1. _**(This PR)**_ Create UI components & tests using mock data
    * Leaving these components disconnected in UI so there is no production impact if we build a new release prior to all 3 parts being completed
2. Combine base components into a tooltip/popover that is integrated into the Specs Explorer
    * Splitting this into a separate PR since there are other efforts within the Specs Explorer this sprint - hopefully this minimizes potential conflicts
3. Once Cloud publishes service support for new flaky data, update GQL queries to hook up these new components and "go live"

[See the PRD](https://docs.google.com/document/d/1u5cyqB6uDcGnhDnex49XKODJzZKdBpYr1xbgovS-bC8) for more information on ACI Flake Annotation

### User facing changelog
N/A

### Steps to test
Not hooked up to UI yet, but component tests have been created for new components to validate look and feel:
* `FlakyBadge.cy.tsx`
* `FlakySpecSummary.cy.tsx`

**Flaky Badge**
Screen Shot 2022-07-21 at 9 00 42 AM

**Flaky Spec Summary**
Screen Shot 2022-07-21 at 8 57 25 AM

### PR Tasks


rachelruderman:
@mike-plummer yay a PR with UI changes!! A great opportunity to try out the new [design - engineering - product collaboration process](https://github.com/cypress-io/prod-eng-docs/blob/main/cross-functional/design-engineering-product.md) 🌟  Can you add the designer in question to this PR review?

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
```diff
@@ -0,0 +1,95 @@
...
+  specName: string
+  specExtension: string
+  severity: 'low' | 'medium' | 'high'
+  totalFlakyRuns: number
+  totalRuns: number
+  lastFlaky: number
+  dashboardUrl: string
+}>()
+
+const accentClass = ref<string | null>(null)
```
lmiller1990, line 52: Instead of explicit `null` it looks like you can do

```js
const accentClass = ref<string>()
```

Which gives the type of `string | undefined`. Bit more concise - unless there's a compelling reason to use `null`, this seems a bit better.

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
```diff
@@ -0,0 +1,95 @@
...
+  const rawRate = props.totalFlakyRuns / props.totalRuns * 100
+
+  if (rawRate > 99 && rawRate < 100) {
+    return 99
+  }
+
+  return ceil(rawRate)
+})
+
+watch(() => props.severity, (severityValue) => {
```
lmiller1990, line 70: Any reason to have all the separate `ref` variables and a `watch`? Since all the refs are derived from props, we could use a single computed property. We also would not need `{ immediate: true }` since `computed` will run immediately and on subsequent updates:

```js
const severity = computed(() => {
  switch (props.severity) {
    case 'low':
      return {
        accentClass: 'border-t-orange-400'
        textClass: 'text-orange-400'
        severityLabel: = t('specPage.flaky.severityLow')
        severityIcon: LowRateIcon
      }

      case 'medium':
        return ...
      case 'high':
         return ...
      default:
        return ...
    }
})
```

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
```diff
@@ -0,0 +1,95 @@
...
+const severityIcon = shallowRef<FunctionalComponent | null>(null)
+const flakyRate = computed(() => {
+  if (props.totalFlakyRuns <= 0 || props.totalRuns <= 0) {
+    return 0
+  }
+
+  const rawRate = props.totalFlakyRuns / props.totalRuns * 100
+
+  if (rawRate > 99 && rawRate < 100) {
+    return 99
```
lmiller1990, line 64: Wouldn't a value of 99.5 round up to 100 anyway? Why do we need this check?

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
```diff
@@ -0,0 +1,95 @@
...
+
+<script setup lang="ts">
+
+import { computed, FunctionalComponent, ref, shallowRef, watch } from 'vue'
+import { useI18n } from '@cy/i18n'
+import SpecNameDisplay from '../SpecNameDisplay.vue'
+import LowRateIcon from '~icons/cy/rate-low_x16'
+import MediumRateIcon from '~icons/cy/rate-medium_x16'
+import HighRateIcon from '~icons/cy/rate-high_x16'
+import ceil from 'lodash/ceil'
```
lmiller1990, line 38: Why not `Math.ceil`?

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
lmiller1990: Can probably make the Icon separate or use `markRaw` if you still get warnings, but I think if we use `computed` it shouldn't be an issue, since it'll just derive based on `props.severity`.

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
mike-plummer: Normally we do a round up, _unless_ the value is between 99 and 100 in which case we round down here

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
mike-plummer: I like to overcomplicate things 😆 . Will replace

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
mike-plummer: Agreed, this is cleaner.  Will update.

I think I was struggling with how to name the resulting computed prop (the hardest of all programming tasks) then subconsciously made the jump to my solution to avoid the extra name

packages/app/src/specs/SpecNameDisplay.vue
```diff
@@ -0,0 +1,17 @@
+<template>
+  <div
```
rachelruderman, line 2:
```suggestion
  <div
        :title="specFileName + specFileExtension"
```

Small accessibility win :-)

packages/app/src/specs/SpecNameDisplay.vue
```diff
@@ -0,0 +1,17 @@
...
+</template>
+
+<script lang="ts" setup>
+
+const props = defineProps<{
+  specFileName: string
+  specFileExtension: string
+}>()
+
+</script>
```
rachelruderman, line 17: The `SpecNameDisplay.vue` component is very similar to the filename logic in `SpecItem.vue` (lines 10-26). There is a bit of different functionality with Highlighter in `SpecItem.vue`, but in the areas where they have crossover, would there be benefit in sharing this logic?

packages/app/src/specs/SpecNameDisplay.vue
rachelruderman: If you add this, would be good to add a corresponding test as well

packages/app/src/specs/flaky-badge/FlakyBadge.vue
```diff
@@ -0,0 +1,17 @@
+<template>
+  <div
+    class="font-bold bg-orange-50 rounded-4px text-xs px-4px text-orange-600 items-center uppercase"
+    style="width: fit-content"
+    data-cy="flaky-badge"
```
rachelruderman, line 5: Two questions for ya!

1. Looks like we have a [Badge](https://github.com/cypress-io/cypress/blob/develop/packages/frontend-shared/src/components/Badge.vue) Vue component in `frontend-shared`; is there a need for `FlakyBadge` as a separate component? 
2. If there is a need for it to be separate, are there any other packages that might use it? In other words, would there be benefit to adding it to the `frontend-shared` package?

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
```diff
@@ -0,0 +1,93 @@
...
+
+const { t } = useI18n()
+
+const props = defineProps<{
+  specName: string
+  specExtension: string
+  severity: 'low' | 'medium' | 'high'
+  totalFlakyRuns: number
+  totalRuns: number
+  lastFlaky: number
```
rachelruderman, line 47:
```suggestion
  runsSinceLastFlake: number
```
Took me a hot second to figure out what `lastFlaky` was referring to. I don't usually go for lengthy variable names, but in this case I think it might clear up some confusion

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
```diff
@@ -0,0 +1,93 @@
...
+}>()
+
+const flakyRate = computed(() => {
+  if (props.totalFlakyRuns <= 0 || props.totalRuns <= 0) {
+    return 0
+  }
+
+  const rawRate = props.totalFlakyRuns / props.totalRuns * 100
+
+  if (rawRate >= 99 && rawRate < 100) {
```
rachelruderman, line 58:
```suggestion
  if (rawRate > 99 && rawRate < 100) {
```

`Math.ceil(99)` returns `99` so we just need `>` here

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
rachelruderman: Maybe it would be helpful to add a comment like
`Only show 100 if rawRate is actually 100`

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -151,6 +151,15 @@
         "linkText": "average spec durations"
       }
     },
+    "flaky": {
+      "badgeLabel": "Flaky",
+      "severityLow": "Low",
+      "severityMedium": "Medium",
+      "severityHigh": "High",
+      "flakyRate": "{0}% flaky rate",
+      "flakyRuns": "{0} flaky runs / {1} total",
```
rachelruderman, line 160: Do we need to take singular/plural into account here? Aka is there a case where this could render `1 flaky runs / 1 total`?

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -151,6 +151,15 @@
...
+      "lastFlaky": "Last flaky {0} runs ago"
```
rachelruderman, line 161: Do we need to take singular/plural into account here? Aka is there a case where this could render `1 flaky runs / 1 total` or `Last flaky 1 runs ago`?

packages/app/src/specs/flaky-badge/FlakyBadge.vue
mike-plummer: Was originally going to use the base `Badge`, but the designs call for _slightly_ different colors and font size than is used there. Wasn't sure whether it was better to create a duplicative component, or complicate the base with additional props to support unique styling. Thoughts?

rachelruderman: Oooooh nice call out! I think it'd be worth reaching out to the design team to get their take, it sounds like it could be an unintended inconsistency in the design. In terms of getting an answer, either @mapsandapps or @elevatebart can bring it up at the Design Review meeting on Tuesday or you can bring it up in the [#design](https://cypressio.slack.com/archives/C3WD9SGKW) channel for a faster response

If it is intended to be different, my personal preference would be to use `Badge` just so that if someone is trying to track down a badge in the code, it's a one-stop-shop. I can see both sides of the argument tho

mike-plummer: 👍 I'll reach out to design to see which direction we should go - I agree it makes the most sense to standardize, but this is a new use case for the component so it may make sense to introduce some variance.

In the short term I've updated `FlakyBadge.vue` to just be a wrapper around `Badge` to handle the necessary style customizations - simplifies it a bit and reduces duplication. When I go to hook this up in the Part 2 PR I'll see whether having this wrapper makes sense based on what design comes back with or whether I should just inline the use of `Badge` directly

elevatebart: Adding it to our next design critique 🧐.

It would really be a good idea to have this badge component in the DS.

if [this figma doc](https://www.figma.com/file/1WJ3GVQyMV5e7xVxPg3yID/Design-System?node-id=1950%3A2659) is the "Tag" your are looking for, then I believe [this ClickUp ticket](https://app.clickup.com/t/18033298/DESIGN-655) is what you want to follow.

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
lmiller1990: I don't understand why we have specific logic for > 99 and < 100 :thinking: 

why?

lmiller1990 (approved):
All my comments have been resolved (except one [outstanding question](https://github.com/cypress-io/cypress/pull/22873#discussion_r928326545) which probably isn't a big deal). :heavy_check_mark:

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
mike-plummer: It was a request from Product in order to match behavior on the dashboard ([documented in the PRD](https://docs.google.com/document/d/1u5cyqB6uDcGnhDnex49XKODJzZKdBpYr1xbgovS-bC8/edit#heading=h.shvspemfqn1w)). Not sure why it was done that way there 🤷

lmiller1990: Okay, let's just roll with it :zap:

lmiller1990:
Looks like the last comment is https://github.com/cypress-io/cypress/pull/22873#discussion_r927929926. Once this is resolved we can merge?

packages/app/src/specs/SpecNameDisplay.vue
mike-plummer: Good callout @rachelruderman. I started looking into consolidating these and ran into some performance issues - `SpecItem` uses the `HighlightedText` component to handle highlighting fuzzy search results so I had to use a `<component :is="..."` to dynamically build the inner content for the spec file name (since in my use case it's just pure text without highlighting), and this additional render step caused pretty painful delays in render on search. Maybe there's a better way to handle dynamic inner content here that I don't know about?

packages/app/src/specs/SpecNameDisplay.vue
```diff
@@ -0,0 +1,18 @@
...
+    data-cy="spec-filename"
+    :title="specFileName + specFileExtension"
+  >
+    <span class="font-semibold text-gray-800">{{ props.specFileName }}</span><span class="text-gray-600">{{ props.specFileExtension }}</span>
+  </div>
+</template>
+
+<script lang="ts" setup>
+
+const props = defineProps<{
```
ZachJW34, line 13:
```suggestion
defineProps<{
```

If you don't have any computed values/need for props in the setup script, you can remove the `props` variable and reference the props directly (which is what you are doing on line 5). Line 7 could be refactored to just use `{{ specFileName }}`.

packages/app/src/specs/flaky-badge/FlakyBadge.vue
ZachJW34: Wanted to throw out we have a Flaky badge on the runs page. Though a different context, I'd expect the styles of the badge and verbiage (All caps vs standard) to be consistent
Screen Shot 2022-07-26 at 11 23 04 AM

Might want to refactor the runs page to use the Badge component.

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
```diff
@@ -0,0 +1,94 @@
...
+  runsSinceLastFlake: number
+  dashboardUrl: string
+}>()
+
+const flakyRate = computed(() => {
+  if (props.totalFlakyRuns <= 0 || props.totalRuns <= 0) {
+    return 0
+  }
+
+  const rawRate = props.totalFlakyRuns / props.totalRuns * 100
```
ZachJW34, line 56:
```suggestion
  const rawRate = (props.totalFlakyRuns / props.totalRuns) * 100
```

A bit easier to read

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
```diff
@@ -0,0 +1,94 @@
...
+      return {
+        accentClass: 'border-t-orange-600',
+        textClass: 'text-orange-600',
+        label: t('specPage.flaky.severityHigh'),
+        icon: HighRateIcon,
+      }
+    default:
+      return undefined
+  }
+})
```
ZachJW34, line 92:
```suggestion
const SEVERITIES = {
  low: {
    label: t('specPage.flaky.severity.low'),
    icon: LowRateIcon,
    textClass: 'text-gray-600',
    accentClass: 'bg-gray-100',
  },
  medium: {
    label: t('specPage.flaky.severity.medium'),
    icon: MediumRateIcon,
    textClass: 'text-gray-600',
    accentClass: 'bg-orange-100',
  },
  high: {
    label: t('specPage.flaky.severity.high'),
    icon: HighRateIcon,
    textClass: 'text-gray-600',
    accentClass: 'bg-red-100',
  },
}

const severity = computed(() => SEVERITIES[props.severity])
```

Optional, but I don't like when a component is littered with `?.` optional access when it isn't the components job to safeguard against nulls. The component is typed such that `severity` should not be null

ZachJW34 (approved):
Few nits but none are blocking. Love the test coverage and percy snapshots!

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
mike-plummer: Regarding the `?.` usages: Normally I would agree. In this case the GQL schema is structured in a way that doesn't give us an typesafe enum for the severity value but rather just a string. Since ACI is reaching out to the cloud at runtime there's a chance that an accidental/unplanned change to the services could start sending back a value at some point in the future that's unknown to this version of the app, so I'm trying to structure this component in a way so the frontend handles it without blowing up.

I do like extracting the `switch` out to a `const` dictionary, will take a look at doing that

packages/app/src/specs/flaky-badge/FlakyBadge.vue
mike-plummer: @abelb Can you weigh in on Zach's comment above?

packages/app/src/specs/flaky-badge/FlakySpecSummary.vue
ZachJW34: In the case that the schema changes and it doesn't conform to `'low' | 'medium' | 'high'`, what does this component look like? Avoiding a runtime error is is best, but this component will display some wacky stuff. We could fallback to the low severity group if we don't match or define an undefined state, but I don't like mixing gql anomalies with a pure presentational component

mike-plummer: Yeah, falling back to the "low" state is probably a better way to handle this. Will update

packages/app/src/specs/flaky-badge/FlakyBadge.vue
mike-plummer: @ZachJW34 I'm working with @abelb on figuring out how to standardize this since it impacts the app in two places as well as the dashboard in two places, and we probably want to figure out a design standard for all four places. Will address this in Part 2 of this effort

Issue #22812, opened 2022-07-15
Add Angular handler to @cypress/webpack-dev-server


ZachJW34:
In order to support Angular for Component Testing, we need to add an Angular handler to our `npm/webpack-dev-server` package. We are targeting v13/v14 support.

The responsibilities of this handler are:
- Source the Webpack related dependencies needed to compile (webpack, webpack-dev-server, html-webpack-plugin)
- Source a Webpack config that will compile Angular components

This handler will be targeted at users that created Angular projects using the `@angular/cli` package. If a user is not using this package, they will most likely have configured their own `webpack.config.js` that they can pass in. This handler will be similar to the other handlers we have for higher-order frameworks `{create-react-app, next, nuxt, vue-cli}`.

Angular offers a Webpack plugin for compiling Angular projects: https://www.npmjs.com/package/@ngtools/webpack
The `@angular/devkit` has internal APIs for sourcing an `@angular/cli` specific project config, though it is contextual to their `builder` API so it isn't as straightforward to use these APIs: https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/build_angular/src/builders/browser/index.ts#L109

Ideally, it would be great to use the `@angular/devkit` APIs as doing so would allow us to respect the options specified by the user (normally stored in a project's `angular.json`). Otherwise, a custom solution that fits most of the needs of the user is acceptable.

### Requirements:

- Source webpack deps and webpack config that enables the compilation of Angular components
- System tests in `npm/webpack-dev-server` proving that the compilation is successful (There is an [external Angular mounting package](cypress-angular-component-testing) we can utilize while we work on official mounting support)

There is an in-progress PR for this work: #22314

Issue #22890, opened 2022-07-21
Better deps check for launchpad


lmiller1990:
### What would you like?

We should have more tight checks for CT dependencies during launchpad.

Eg: currently we look for Vite >= 2, but Vite 3 isn't officially supported and has problems. We should not use >= for dependency checking during CT installation, but only include known versions.

Users can skip through if they want, but we don't want to give them a false expectation that something is officially supported, when it's not.

lmiller1990:
Related https://github.com/cypress-io/cypress/issues/22025

PR #22897, opened 2022-07-22
feat: add angular-ct project setup

### LINKED ISSUE #22818: Add Angular CT integration in Launchpad
### What would you like?

Once we have the internals for compiling and mounting Angular components, we will need to add Angular detection and configuration into the CT Wizard setup.

We will need to:

Add Angular to the list of supported frameworks (mark it as alpha). It should autodetect Angular when possible
Screen Shot 2022-07-15 at 2 39 01 PM

Configure the required dependencies for an Angular CT App
Screen Shot 2022-07-15 at 2 39 19 PM

Configure the file scaffolding for an Angular CT App
Screen Shot 2022-07-15 at 2 39 40 PM

> Most of this configuration is stored in `packages/scaffold-config/src/frameworks.ts`.

Requirements:
- List Angular (alpha) in the CT Project Setup (with autodetection)
- Configure required dependencies
- Configure scaffolded files
- Verify flow with Angular CT setup system test

### Why is this needed?

_No response_

### Other

_No response_

ZachJW34:
- Closes #22818 

### User facing changelog
Detect and scaffold CT project files for Angular applications

### Additional details
Adding a new framework is fairly straightforward as it is all configuration based. Most of the code changes are from added system tests. The [mount](https://github.com/cypress-io/cypress/issues/22813) and [dev-server handler](https://github.com/cypress-io/cypress/issues/22812) PRs need to be merged before this can land.

### Steps to test
- Scaffold a new angular application with `ng new angular-app` (Might need to install ng globally with `npm i -g @angular/cli`)
- `yarn cypress:open` in cypress repo and point it at the new Angular application
- Walk through setup

Won't work until the Angular infra is in which should be soon. You can also run the `packages/launchpad/cypress/e2e/scaffold-component-testing.cy.ts -> angular-cli-unconfigured` test in the launchpad.

https://user-images.githubusercontent.com/25158820/180490616-643ab37e-8709-4cd5-9db0-4f615513e196.mov

### PR Tasks


ZachJW34:
Need to scaffold `specPattern` into generated config. Since we pass the specPattern into the generated tsconfig, the glob pattern must abide by the rules that the `tsconfig.include` uses. The default spec pattern does not work due to the brace sets (`{js,jsx,ts,tsx}`).

packages/scaffold-config/src/dependencies.ts
```diff
@@ -97,6 +97,51 @@ export const WIZARD_DEPENDENCY_NEXT = {
   minVersion: '>=10.0.0',
 } as const
 
+export const WIZARD_DEPENDENCY_ANGULAR_CLI = {
+  type: 'angular',
+  name: 'Angular CLI',
+  package: '@angular/cli',
+  installer: '@angular/cli',
+  description: 'CLI tool that you use to initialize, develop, scaffold, and maintain Angular applications.',
+  minVersion: '>=13.0.0',
```
lmiller1990, line 106: Do we also want to include a `maxVersion`? (not sure we actually have code path supporting this, but it would be easy to add, I think).

Also, do we know this does **not** work with any other versions (like 12 or 14)?

lmiller1990: Actually we can add a `maxVersion` here: https://github.com/cypress-io/cypress/issues/22890

Question still stands about other versions - do we have support for 12 or  14?

ZachJW34: No support for 12 but 14 works (and is proven out in the system tests)

packages/scaffold-config/src/frameworks.ts
```diff
@@ -216,5 +223,28 @@ export const WIZARD_FRAMEWORKS = [
...
+        inPkgJson(dependencies.WIZARD_DEPENDENCY_ANGULAR_CLI, projectPath),
+        inPkgJson(dependencies.WIZARD_DEPENDENCY_ANGULAR_DEVKIT_BUILD_ANGULAR, projectPath),
+        inPkgJson(dependencies.WIZARD_DEPENDENCY_ANGULAR_CORE, projectPath),
+        inPkgJson(dependencies.WIZARD_DEPENDENCY_ANGULAR_COMMON, projectPath),
+        inPkgJson(dependencies.WIZARD_DEPENDENCY_ANGULAR_PLATFORM_BROWSER_DYNAMIC, projectPath),
+      ]
+    },
+    codeGenFramework: 'angular',
+    mountModule: 'cypress/angular',
+    supportStatus: 'alpha',
```
ZachJW34, line 246:
```suggestion
    supportStatus: 'full',
```

packages/config/src/ast-utils/astConfigHelpers.ts
```diff
@@ -41,15 +42,18 @@ export interface ASTComponentDefinitionConfig {
...
+        framework: config.framework,
+        bundler: config.bundler,
       },
-    }
+      specPattern: config.specPattern,
+    },
+  }, null, 2)
+
+  return extractProperty(`
+    const toMerge = ${componentConfig}
```
lmiller1990, line 56: Hm I would expect this would have changed the indentation. Or do we handle the formatting elsewhere :thinking:

ZachJW34: Right above is the `JSON.stringigify(..., null, 2)` indents properly.

packages/config/src/ast-utils/astConfigHelpers.ts
```diff
@@ -41,15 +42,18 @@ export interface ASTComponentDefinitionConfig {
...
-          bundler: '${config.bundler}',
-        },
+  const componentConfig = JSON.stringify({
+    component: {
+      devServer: {
+        framework: config.framework,
+        bundler: config.bundler,
       },
-    }
+      specPattern: config.specPattern,
```
lmiller1990, line 51: Doesn't this mean we will be scaffolding `specPattern: undefined` for the majority of frameworks (except Angular?)

Why do we need to do this? The default is  `**/*.cy.{js,jsx,ts,tsx}`. Shouldn't that cover Angular,  too?

lmiller1990 (commented):
One question, giving it a test now.

lmiller1990:
I tried it out, works great!

image

packages/config/src/ast-utils/astConfigHelpers.ts
ZachJW34: See: https://github.com/cypress-io/cypress/pull/22897#issuecomment-1192948652
`undefiend` values get dropped when scaffolding out the config, but it is confusing that an `undefined` is equivalent to the default spec pattern. Will see if I can come up with something cleaner

lmiller1990: > Since we pass the specPattern into the generated tsconfig

I might be misunderstanding something - what does this mean? Where do we pass `specPattern` into their `tsconfig.json`?

Unfortunate that `tsconfig.json` does not support curly braces.

ZachJW34: This is an Angular dev-server implementation detail. Angular needs to know about all of the typescript files included in the compilation ahead of time, so we need to push the specPattern glob into a tsconfig that we generate and pass into the webpack configuration. [Code here](https://github.com/cypress-io/cypress/blob/84ee67ec9b5a7f4faee532f1185d9080cdcb7cd5/npm/webpack-dev-server/src/helpers/angularHandler.ts#L127)

Since the `tsconfig.includes` doesn't support curly braces, we need to have a stricter specPattern than the default, otherwise the compilation will fail.

ZachJW34: Tweaked the types in [b3f6d4c](https://github.com/cypress-io/cypress/pull/22897/commits/b3f6d4c9a060a094f9f710d6f2b2aaaff3cdd057)

lmiller1990: I think you linked the wrong commit, but sounds good.

lmiller1990 (approved):
Comments addressed and I tested it out, works great 👍

packages/scaffold-config/src/dependencies.ts
```diff
@@ -97,6 +97,51 @@ export const WIZARD_DEPENDENCY_NEXT = {
...
+} as const
```
mike-plummer, line 107: Any reason we need the const assertion here but not on the other items being added?

mike-plummer (approved):
LGTM, just one question

packages/scaffold-config/src/dependencies.ts
ZachJW34: Oversight on my part! Addressed in [be09724](https://github.com/cypress-io/cypress/pull/22897/commits/be09724cbce0eed112b16425c6acfb997cf4f7b3)

Issue #22836, opened 2022-07-18
Add 'create from Vue component' workflow


astone123:
### What would you like?

Add the workflow to allow users to create a new component test scaffold for an existing component in their Vue project. The workflow should be accessible by clicking on the "New Spec" button at the top of the spec list. 

This already exists in our Git history and was implemented at one point, but was removed before the version 10 release. Steps to walk through the existing version of this workflow:

1. `git checkout 0e48012660`
2. `yarn`
3. `yarn cypress:open`
4. Select a Vue project that has at least one component file in it
5. Select component testing and open a browser
6. Click the "+ New Spec" button at the top of the specs list
7. Click "Create from component"
8. Choose a Vue component file to create a scaffold test for

![Screen Shot 2022-07-18 at 10 24 21 AM](https://user-images.githubusercontent.com/14275198/179540362-b9ae9743-476e-486a-a9d3-2e36b702c332.png)

![Screen Shot 2022-07-18 at 10 24 37 AM](https://user-images.githubusercontent.com/14275198/179540427-26381af2-7121-477a-8240-ba9b748e7a36.png)

![Screen Shot 2022-07-18 at 10 24 53 AM](https://user-images.githubusercontent.com/14275198/179540555-c7450779-16f6-4c29-98e7-61baf0117906.png)

Ensure that the following cases are covered:
- Works for both Vue 2 and Vue 3
- Empty state for when the project has no Vue components to choose from

### Why is this needed?

This feature will greatly improve the onboarding process for new CT users. It will create a new file that mounts the given Vue component in Cypress and users can click to open the file in their IDE, never having to manually write the `mount` command or test blocks.

### Other

Make sure to get feedback from design teams before releasing the feature.

lmiller1990:
As we implement this, we should write the code in a way users can choose their own default template. The way I've solved this in my project is a gulp script:

https://github.com/lmiller1990/rhythm/blob/main/scripts/gulpfile.ts#L235-L260

If I could have some way to do this from the UI, I wouldn't need to change to my terminal. My script makes a component and a test, so my ideal end game for this feature is more like a "generate boilerplate" where I can configure what I want. Eg we might have a directory under `cypress` like `templates`, which might have

```js
- vue-component
  - Foo.vue
  - Foo.cy.js
```

A template could be:

```js
describe('{{ component }}', () => {
  it("renders", () => {
    cy.mount({{ component }})
  })
})
```

Now, I can choose what I want the template to have, preferred code style, etc.

[...]

PR #22898, opened 2022-07-22
feat: Create spec from Vue component

astone123:
- Closes https://github.com/cypress-io/cypress/issues/22836 

### User facing changelog

For Vue projects in component testing, choose to generate a spec file from a list of the components that exist in your project

### Additional details

This PR makes it so that when a user opens a Vue project for component testing and click the "+ New Spec" button, they are presented with a list of the components in their project. They can click on a component and a component spec will be created with the name of the component and placed in the directory where the component exists.

The test will be scaffolded to match the syntax of either a Vue 2 or Vue 3 app depending on which version of the framework the project is using.

### Steps to test

1. Create a new Vue (2 or 3) project that has some components
2. Configure the project for component testing
3. Open component testing in a browser
4. Click the "+ New Spec" button in the top-right of the spec list
5. Confirm that you are presented with a list of the components in your application
6. Click a component and confirm that a test is created with the correct name and template
7. Run the generated spec (it should pass unless the component has some required props that we're not passing)
8. Verify that creating empty specs still works for React projects and e2e projects

Waiting to post screenshots until we get design feedback.

### Design Questions

1. Should users be able to create from component **as well as** create empty specs in Vue projects, now that create from component is available in these projects?
2. If there is only one option, should we show the card for the option in the modal, or should we immediately go to the first step (this is what we do now, in this branch, for "create from component").
3. What should the create from component card look like (see linked issue for a screenshot of the existing one)? Do we need to make changes to the existing one? If the answer to # 1 is yes, then I think we should at least update the icon so that it looks different from the "create empty" card.

### PR Tasks

- [ ] Design approval


astone123:
@elevatebart Can you take a look at the design questions that I laid out in the PR description? We need some feedback to move forward. Thanks

elevatebart:
Thanks @astone123 I will take a look.

How urgently do you need an answer? 
Is it OK if you have them for Tuesday afternoon?

astone123:
@elevatebart Looking to have this finished by the end of the sprint. Tuesday would be great 👍🏻

elevatebart:
### Design Questions

> 1. Should users be able to create from component as well as create empty specs in Vue projects, now that create from component is available in these projects?

We did not find an easy answer yet. 
Let's involve product managers to prioritize (@pstakoun ?):
- We have to finish designing and implementing the "create form empty"
  - Give users an idea of their directory structure
  - Allow changing the default creation root directory
  - Match the default root directory to the meta-framework

> 2. If there is only one option, should we show the card for the option in the modal, or should we immediately go to the first step (this is what we do now, in this branch, for "create from component").

- What is that first step you mention?
- As is, I confirm there is no need for an extra step.

> 3. What should the create from component card look like (see linked issue for a screenshot of the existing one)? Do we need to make changes to the existing one? If the answer to # 1 is yes, then I think we should at least update the icon so that it looks different from the "create empty" card.

- We need to create an icon for the "create empty spec" if we keep the button. @ryanjwilke

ZachJW34:
@elevatebart have all the design questions been addressed?

elevatebart:
@ZachJW34 I don't think the question "should we remove the empty card?" has been answered.

Also, we still do not have an icon for the new button (the icon we use today for empty was meant for component)

ZachJW34:
@elevatebart ok, anyway we can prioritize getting the designs finalized by the start of our next sprint (next Tuesday)?

astone123:
> What is that first step you mention?

@elevatebart I was referring to the component list where the user can select a component to create a spec for

ryanjwilke:
@elevatebart I think we should discuss this with the @cypress-io/product team first as there is another button (Import from story) that was supposed to be used in it's place that was temporarily cut out in order to reduce the initial scope for `v10`. It's likely that this button will still come back in soon, although I'm not sure on the priority of that work.

I'll send a message to the product team to see what they think.

packages/app/src/specs/generators/component/ComponentGenerator.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const ComponentGenerator: SpecGenerator = {
+  card: ComponentGeneratorCard,
+  entry: ComponentGeneratorStepOne,
+  disabled: () => { },
+  show: (currentProject) => {
+    const specPattern = currentProject?.config.find((item) => item.field === 'specPattern')
+
+    if (specPattern && specPattern.from !== 'default') {
+      return false
```
astone123, line 14: This won't work with custom spec patterns right now so don't show this card if we don't have the default spec pattern. I'm going to fix this with #23071

marktnoonan:
So excited for this! @astone123 is this read to move out of draft, or are you looking for review while it is in draft, to get early feedback?

astone123:
@marktnoonan I just marked this as ready to review. I think it's good to go - just need code reviews and approval from @elevatebart and/or others on the design team

packages/app/src/specs/generators/component/ComponentGeneratorStepOne.vue
```diff
@@ -0,0 +1,200 @@
+<template>
+  <div class="flex flex-col flex-grow justify-between">
+    <template v-if="generatedSpecError">
```
ZachJW34, line 3: I looked to see if I can find anywhere that `generatedSpecError` gets returned but couldn't find any, looks like it is a holdover from https://github.com/cypress-io/cypress/pull/19862/files where if the user generated a component spec that didn't match `specPattern`, it would take them through the `EmptyGenerator` flow so that the user could input the filename manually. Since we don't have this issue with this new flow, can we remove the instances of `erroredCodegenCandidate`?

I know there is another issue where you want to tackle customSpecPatterns (which looks like what this code was trying to solve).

packages/app/src/specs/generators/component/ComponentGeneratorStepOne.vue
```diff
@@ -0,0 +1,201 @@
...
+          <p class="text-lg">
+            Loading
+          </p>
+        </div>
+        <FileChooser
+          v-else-if="!result"
+          v-model:extensionPattern="extensionPattern"
+          :files="allFiles"
+          :loading="query.fetching.value"
+          :file-match-button-disabled="true"
```
astone123, line 33: In this case, we want the file match button to be disabled so that the user can't change the glob pattern.

packages/app/src/specs/generators/empty/EmptyGenerator.tsx
```diff
@@ -7,5 +7,10 @@ export const EmptyGenerator: SpecGenerator = {
   entry: EmptyGeneratorCardStepOne,
   matches: () => true,
   disabled: () => false,
+  show: (currentProject) => {
+    const specPattern = currentProject?.config.find((item) => item.field === 'specPattern')
+
+    return currentProject?.codeGenGlobs?.component !== '*.vue' || (specPattern && specPattern.from !== 'default')
```
ZachJW34, line 13: We are keying off of the `codeGenGlobs` to determine if it's a vue app (and thus to either show empty or create-from-component). I think it makes more sense to key off of the `framework` that defined in the user's `config.component.devServer`.

packages/app/src/specs/generators/component/ComponentGenerator.tsx
```diff
@@ -0,0 +1,21 @@
+import { filters } from '../GeneratorsCommon'
+import ComponentGeneratorStepOne from './ComponentGeneratorStepOne.vue'
+import type { SpecGenerator } from '../types'
+import ComponentGeneratorCard from './ComponentGeneratorCard.vue'
+
+export const ComponentGenerator: SpecGenerator = {
+  card: ComponentGeneratorCard,
+  entry: ComponentGeneratorStepOne,
+  disabled: () => { },
```
ZachJW34, line 9: Do we ever have a disabled state? We have a lot of baggage from our previous implementation that is hanging around

packages/app/src/specs/generators/empty/EmptyGenerator.tsx
```diff
@@ -7,5 +7,10 @@ export const EmptyGenerator: SpecGenerator = {
   entry: EmptyGeneratorCardStepOne,
   matches: () => true,
   disabled: () => false,
+  show: (currentProject) => {
+    const specPattern = currentProject?.config.find((item) => item.field === 'specPattern')
```
ZachJW34, line 11: We could add a field in gql for `isDefaultSpecPattern` so we don't have t derive it on the frontend.

packages/data-context/src/codegen/spec-options.ts
```diff
@@ -18,18 +21,105 @@ interface CodeGenOptions {
...
+    const componentPath = this.relativePath()
+
+    const templateKey: TemplateKey = framework.mountModule === 'cypress/vue' ? 'vue3Component' : 'vue2Component'
+
     return {
       codeGenType: this.options.codeGenType,
-      fileName: await this.getFilename(),
+      componentName,
+      componentPath,
+      mountModule: framework.mountModule,
```
ZachJW34, line 101: Don't need to return `mountModule` since the template is handling the structure of the generated file.

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -478,6 +479,18 @@ export class ProjectDataSource {
     return preferences[projectTitle] ?? null
   }
 
+  async getCodeGenGlobs () {
+    assert(this.ctx.currentProject, `Cannot find glob without currentProject.`)
+
+    const looseComponentGlob = '*.{js,jsx,ts,tsx,.vue}'
+
+    const detected = await detectFramework(this.ctx.currentProject || '')
```
ZachJW34, line 487: The `config.component.devServer` key is a more valid indicator for what the project is using vs `detectFramework`. I don't think we need to use `detectFramework` at all as part of this work.

system-tests/projects/no-specs-vue-2/.gitignore
```diff
@@ -0,0 +1,23 @@
+.DS_Store
```
ZachJW34, line 1: We should remove any extraneous files from this system-test project, like .gitignore and README.md. Might be able to trim it down further by removing the babel and eslint depednencies/configurations.

marktnoonan:
> Click the "+ New Spec" button in the top-right of the spec list

This should be testable from opening the `app` package in component testing mode right? I get the regular "Enter the path for your new spec" prompt. Wonder if I'm doing something wrong - and I haven't tested with a fresh Vue project.

packages/app/src/specs/generators/empty/EmptyGenerator.tsx
marktnoonan: I like this. Our array-based config lookups always feels unwieldy.

astone123:
> This should be testable from opening the app package in component testing mode right?

@marktnoonan We use a custom spec pattern for the `app` package, so you won't see it there. I have https://github.com/cypress-io/cypress/issues/23071 to make this work when the project is using a custom spec pattern

marktnoonan:
> @marktnoonan We use a custom spec pattern for the `app` package, so you won't see it there. I have #23071 to make this work when the project is using a custom spec pattern

Ah, gotcha thanks. Wonder why we have any of these `.spec.tsx` files 🤔. Seems that's the reason we have a custom spec pattern, but also seems like a pretty arbitrary set of files.

Screen Shot 2022-08-04 at 1 15 31 PM

packages/app/src/specs/generators/empty/EmptyGenerator.tsx
astone123: I agree that this is the way we should do it. I don't see it anywhere in `currentProject` object here though. I wonder if there's something I need to add to the gql query to see that?

packages/data-context/src/codegen/spec-options.ts
```diff
@@ -18,18 +21,105 @@ interface CodeGenOptions {
...
+    }
+  }
+
+  private async getVueExtension (): Promise<ComponentExtension> {
+    try {
+      const fileContent = await this.ctx.fs
+      .readFile(this.options.codeGenPath)
+      .then((res) => res.toString())
+
+      return fileContent.includes('lang="ts"') ? '.cy.ts' : '.cy.js'
```
astone123, line 113: I think I can refactor this to use the `fileExtensionToUse` property instead

brian-mann:
Commenting on a few things - 

**Until officially specified** I am the acting product lead ya'll should come to for product questions.

### Design Questions

**1. Should users be able to create from component as well as create empty specs in Vue projects, now that create from component is available in these projects?**

> Yes, do not remove existing features - there is still a valid reason to create a blank spec, plus see my answer for Q3 below. 

**2. If there is only one option, should we show the card for the option in the modal, or should we immediately go to the first step (this is what we do now, in this branch, for "create from component").**

> There would be two options, but I would be okay with adding keyboard shortcuts to make this step faster. `N` for `new`, `C` or `S` for create from spec.

**3. What should the create from component card look like (see linked issue for a screenshot of the existing one)? Do we need to make changes to the existing one? If the answer to # 1 is yes, then I think we should at least update the icon so that it looks different from the "create empty" card.**

> Yes, we can add a "badge" like we did for the button styles when you pick E2E vs Component Testing. Likely want to draw attention to this button like by saying `(New)` or `(New Feature)` - something like that

brian-mann:
Also another question - when this feature previously existing there was a caret over the `*.vue` glob that allowed you to expand that in order to modify the glob search input... where did that go?

There was also a shadow on the footer to help you understand whether the list of items is "at the end" and not scrollable vs "covered up" and therefore scrollable.

/CC @ryanjwilke

astone123:
@brian-mann thanks for the answers on those questions. I'd say we should probably skip keyboard shortcuts for now but if the card badge is something that we've already defined (looks like it is) then I can add that here. Would want to get an idea of what that should look like.

> where did that go?

I disabled click events on that button and removed the caret because I didn't think it made sense to allow the user to change the glob. If they could change the glob then they could select other non-vue-component files like config files or whatever else. Handling those cases gets a little messy so I figured we should just show it so they know which files we're looking at, but they can't change it in this view.

> There was also a shadow on the footer

Not sure where this went. Is the footer shadow in the screenshot that you posted? I can't tell. I'm not sure what it looked like originally.

brian-mann:
@astone123 that makes sense why you removed the caret on the `*.vue` glob pattern. And I think that specifically for `vue` it's not as big of a problem.

Why we originally had the caret was for frameworks like `react` in which we don't know the right extension of the app components - it could be `.ts` or `.js`.

However - there is another important reason the glob pattern is required - we don't know where the source files are and you can imagine situations where monorepos or projects are using nonstandard folder structures. Given the very broad scope of the `*.vue` glob pattern could end up returning a huge number of files that are not within the right "folder scope".

By popping open the caret and modifying the search path you could then exclude/include exactly the right target folder and this value is supposed to be saved to your local app state (not cypress config) and thus carryover on subsequent "create from component" actions.

If I remember correctly, I know that @JessicaSachs or @elevatebart or @lmiller1990 or @ZachJW34 ran into this situation and could likely chime in. We built this UI directly in response to some of those real world usage patterns.

packages/app/src/specs/generators/empty/EmptyGenerator.tsx
lmiller1990: You might need to add it. I do see an edge case, though. For the "happy path" you can potentially grab the framework from the config, eg `component.devServer.framework`, and then match that up with with the correct one [in this array](https://github.com/cypress-io/cypress/blob/develop/packages/scaffold-config/src/frameworks.ts#L64). It's got a [`codeGenFramework`](https://github.com/cypress-io/cypress/blob/develop/packages/scaffold-config/src/frameworks.ts#L80) key.

The problem will be custom dev servers; we don't really have any way to handle this, since we don't know what framework they are using. I think the eventual fix might be to allow manually configuring how this feature works; someone who wants to configure a dev server themselves likely has the technical skills to configure the template they'd like to use for the scaffold feature, too.

packages/data-context/src/codegen/spec-options.ts
lmiller1990: `lang="typescript"` is also valid, too.

If we end up needing to get real crazy, the `@vue/sfc-compiler` could be used to get this info. Or, we could use a html parser - but I think this is fine for now, as long as you add a check for `lang="typescript"`.

lmiller1990:
To follow up on the above post, I think another important reason to allow narrowing the search with a glob is many React apps do not use a `jsx` extension, but just plain old `js` (even for files using JSX syntax). An example is CRA - a brand new CRA project gives you an `App.js` with JSX.

Another use case is Vue apps using TSX (you can do that). So, if we don't let them change from `vue`, they wouldn't be able to use this feature.

packages/data-context/src/codegen/spec-options.ts
```diff
@@ -18,18 +21,104 @@ interface CodeGenOptions {
 //   Button.foo.js  -> Button.foo-copy-1.js
 export const expectedSpecExtensions = ['.cy', '.spec', '.test', '-spec', '-test', '_spec']
 
+type ComponentExtension = '.cy.js' | '.cy.ts' | '.cy.jsx' | '.cy.tsx'
```
lmiller1990, line 24: Here is a neat trick:

```suggestion
type ComponentExtension = `cy.${'js' | 'ts' | 'jsx' | 'tsx'}`
```

packages/data-context/src/codegen/spec-options.ts
```diff
@@ -18,18 +21,104 @@ interface CodeGenOptions {
...
+type TemplateKey = 'e2e' | 'componentEmpty' | 'vue2Component' | 'vue3Component'
 export class SpecOptions {
   private parsedPath: ParsedPath;
+  private parsedErroredCodegenCandidate?: ParsedPath
+
+  private async getFrontendFramework () {
+    const detected = await detectFramework(this.ctx.currentProject || '')
```
lmiller1990, line 31: I think we should use `cypress.config` and get this from `component.devServer` - no need to detect, they should have told us what they are using.

packages/data-context/src/codegen/spec-options.ts
```diff
@@ -18,18 +21,104 @@ interface CodeGenOptions {
...
+      componentPath,
+      fileName: await this.buildComponentSpecFilename(await this.getVueExtension()),
+      templateKey,
+    }
+  }
+
+  private async getVueExtension (): Promise<ComponentExtension> {
+    try {
+      const fileContent = await this.ctx.fs
+      .readFile(this.options.codeGenPath)
```
lmiller1990, line 109: I think you can just do `readFile(..., 'utf8'` so you don't need `toString`

lmiller1990 (commented):
Mostly looks fine, giving it a test...

lmiller1990:
I gave it a test on one of my real projects (on a simple component, mind you) and it actually worked without a hitch. Nice!

astone123:
> @astone123 that makes sense why you removed the caret on the `*.vue` glob pattern. And I think that specifically for `vue` it's not as big of a problem as it is for react ... however there are still some real challenges with its removal.
> 
> We originally had the caret was for frameworks like `react` in which we don't know the right extension of the app components - it could be `.ts` or `.js` and therefore the user may want to scope to one or the other if they have a homogenous extension set.
> 
> However, there is another [even more] important reason the glob pattern is required - we don't know where the source files are and you can imagine situations where monorepos or projects are using nonstandard folder structures. Given the very broad scope of the `*.vue` glob pattern could end up returning a huge number of files that are not within the right "folder scope".
> 
> By popping open the caret and modifying the search path you could then exclude/include exactly the right target folder and this value is supposed to be saved to your local app state (not cypress config) and thus carryover on subsequent "create from component" actions.
> 
> If I remember correctly, I know that @JessicaSachs or @elevatebart or @lmiller1990 or @ZachJW34 ran into this situation and could likely chime in. We built this UI directly in response to some of those real world usage patterns.

Yeah, that makes sense. I wasn't thinking about that scenario. I made some updates based on your answers and updated the screenshots in the PR description to show what things look like now. Could you take a look at those and let me know what you think? Thanks

packages/data-context/src/codegen/spec-options.ts
astone123: I agree, but in this case I'm going to need to do something similar to what `detectFramework` does (look through satisfied dependencies) anyway because I need to differentiate between Vue 2 and Vue 3 (because they use different file templates). The config uses `framework: 'vue-cli'` for both Vue 2 and 3.

packages/app/src/specs/generators/empty/EmptyGenerator.tsx
astone123: https://github.com/cypress-io/cypress/pull/22898#discussion_r939162622

packages/frontend-shared/src/components/Card.vue
```diff
@@ -17,6 +17,13 @@
     >
       {{ t('versions.beta') }}
     </div>
+    <div
+      v-if="title === t('createSpec.component.importFromComponent.header')"
```
lmiller1990, line 21: This seems strange - there must be a conditional somewhere up the chain that decides `title`. Instead of doing a conditional on a string, I wonder if we should find out what the actual logic is and either pass a prop or use it here?

lmiller1990 (commented):
One additional comment

packages/data-context/src/codegen/spec-options.ts
lmiller1990: Hmm but you can use this for both v2 and v3:

```html
<script>
export default {
}
</script>

<template>
</template>
```

And it's valid. I wonder if we can really assume everyone using `script setup`? This is my original issue with not making this customizable - unless you get lucky, you'll need to change the generated content - which means the feature is probably not super useful, until you can customize the content.

Until then, can we just pick a template that works for both v2 and v3 (something simple like what's suggested above?)

astone123: @lmiller1990 I'm not sure what you mean. The templates that we have are for generating specs for vue 2 and vue 3 components with the only difference being `props: {}` vs `propData: {}` corresponding to the different Vue Test Util versions I believe. I don't think those rely on everyone using `script setup` syntax in their Vue components.

I agree that we should allow users to customize or provide their own templates in the future though

https://github.com/cypress-io/cypress/blob/99ce5c26522aebed1115cdf439730f5d412c296e/packages/data-context/src/codegen/templates/vue2-component/vue2-component.ejs

packages/frontend-shared/src/components/Card.vue
astone123: I moved this logic up out of `Card` and made it a prop that the parent component passes down. If there's badge text, then we show the badge, otherwise we don't.

packages/app/cypress/e2e/specs.cy.ts
```diff
@@ -619,6 +619,80 @@ describe('App: Specs', () => {
...
+        cy.findByText('*.vue').should('be.visible')
+        cy.findByText('2 Matches').should('be.visible')
+        cy.findByLabelText('file-name-input').type('HelloWorld')
+
+        cy.findByText('HelloWorld').should('be.visible')
+        cy.findByText('1 of 2 Matches').should('be.visible')
+        cy.findByText('App').should('not.exist')
+      })
+
+      it('shows success modal when component spec is created', () => {
```
ZachJW34, line 674: Would be cool to test that the spec gets compiled and run after the spec creation.

ZachJW34 (changes requested):
You need to add a `cypress/support/component-index.html` to the new system test project. This will clear up a lot of warnings in the terminal when executing the tests.

astone123:
> If we drop the props or propsData, would that make it easier? Not sure how strict of a requirement it is to seed with empty prop object, but without it the templates are the same and it would reduce the framework detection logic.

@ZachJW34 yeah this is a good point, we don't necessarily need to pass a blank props object, and removing this would allow us to use the same template for Vue 2 and 3. Let's remove that for now

packages/app/cypress/e2e/specs.cy.ts
astone123: It doesn't look like the test actually runs in Cypress-in-Cypress... this is what I see when I try to run the generated test. Works fine outside of the c-in-c environment

Screen Shot 2022-08-08 at 2 29 40 PM

here's the whole error

```
Error: Child compilation failed:
  Module not found: Error: Can't resolve '/private/var/folders/z_/9smbmcx124zc8d  677fn1pjm40000gq/T/cy-projects/no-specs-vue-2/cypress/support/component-index.  html' in '/private/var/folders/z_/9smbmcx124zc8d677fn1pjm40000gq/T/cy-projects  /no-specs-vue-2'
  ModuleNotFoundError: Module not found: Error: Can't resolve '/private/var/fold  ers/z_/9smbmcx124zc8d677fn1pjm40000gq/T/cy-projects/no-specs-vue-2/cypress/sup  port/component-index.html' in '/private/var/folders/z_/9smbmcx124zc8d677fn1pjm  40000gq/T/cy-projects/no-specs-vue-2'
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/w  ebpack/lib/Compilation.js:2016:28
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/w  ebpack/lib/NormalModuleFactory.js:798:13
      at eval (eval at create (/private/tmp/cy-system-tests-node-modules/no-spec  s-vue-2/node_modules/tapable/lib/HookCodeFactory.js:33:10), :10:1)
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/w  ebpack/lib/NormalModuleFactory.js:270:22
      at eval (eval at create (/private/tmp/cy-system-tests-node-modules/no-spec  s-vue-2/node_modules/tapable/lib/HookCodeFactory.js:33:10), :9:1)
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/w  ebpack/lib/NormalModuleFactory.js:434:22
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/w  ebpack/lib/NormalModuleFactory.js:120:11
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/w  ebpack/lib/NormalModuleFactory.js:670:25
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/w  ebpack/lib/NormalModuleFactory.js:855:8
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/w  ebpack/lib/NormalModuleFactory.js:975:5
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/n  eo-async/async.js:6883:13
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/w  ebpack/lib/NormalModuleFactory.js:958:45
      at finishWithoutResolve (/private/tmp/cy-system-tests-node-modules/no-spec  s-vue-2/node_modules/enhanced-resolve/lib/Resolver.js:312:11)
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/e  nhanced-resolve/lib/Resolver.js:386:15
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/e  nhanced-resolve/lib/Resolver.js:435:5
      at eval (eval at create (/private/tmp/cy-system-tests-node-modules/no-spec  s-vue-2/node_modules/tapable/lib/HookCodeFactory.js:33:10), :16:1)
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/e  nhanced-resolve/lib/Resolver.js:435:5
      at eval (eval at create (/private/tmp/cy-system-tests-node-modules/no-spec  s-vue-2/node_modules/tapable/lib/HookCodeFactory.js:33:10), :27:1)
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/e  nhanced-resolve/lib/DescriptionFilePlugin.js:87:43
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/e  nhanced-resolve/lib/Resolver.js:435:5
      at eval (eval at create (/private/tmp/cy-system-tests-node-modules/no-spec  s-vue-2/node_modules/tapable/lib/HookCodeFactory.js:33:10), :15:1)
      at /private/tmp/cy-system-tests-node-modules/no-specs-vue-2/node_modules/e  nhanced-resolve/lib/Resolver.js:435:5
  
  - Compilation.js:2016 
    [no-specs-vue-2]/[webpack]/lib/Compilation.js:2016:28
  
  - NormalModuleFactory.js:798 
    [no-specs-vue-2]/[webpack]/lib/NormalModuleFactory.js:798:13
  
  
  - NormalModuleFactory.js:270 
    [no-specs-vue-2]/[webpack]/lib/NormalModuleFactory.js:270:22
  
  
  - NormalModuleFactory.js:434 
    [no-specs-vue-2]/[webpack]/lib/NormalModuleFactory.js:434:22
  
  - NormalModuleFactory.js:120 
    [no-specs-vue-2]/[webpack]/lib/NormalModuleFactory.js:120:11
  
  - NormalModuleFactory.js:670 
    [no-specs-vue-2]/[webpack]/lib/NormalModuleFactory.js:670:25
  
  - NormalModuleFactory.js:855 
    [no-specs-vue-2]/[webpack]/lib/NormalModuleFactory.js:855:8
  
  - NormalModuleFactory.js:975 
    [no-specs-vue-2]/[webpack]/lib/NormalModuleFactory.js:975:5
  
  - async.js:6883 
    [no-specs-vue-2]/[neo-async]/async.js:6883:13
  
  - NormalModuleFactory.js:958 
    [no-specs-vue-2]/[webpack]/lib/NormalModuleFactory.js:958:45
  
  - Resolver.js:312 finishWithoutResolve
    [no-specs-vue-2]/[enhanced-resolve]/lib/Resolver.js:312:11
  
  - Resolver.js:386 
    [no-specs-vue-2]/[enhanced-resolve]/lib/Resolver.js:386:15
  
  - Resolver.js:435 
    [no-specs-vue-2]/[enhanced-resolve]/lib/Resolver.js:435:5
  
  
  - Resolver.js:435 
    [no-specs-vue-2]/[enhanced-resolve]/lib/Resolver.js:435:5
  
  
  - DescriptionFilePlugin.js:87 
    [no-specs-vue-2]/[enhanced-resolve]/lib/DescriptionFilePlugin.js:87:43
  
  - Resolver.js:435 
    [no-specs-vue-2]/[enhanced-resolve]/lib/Resolver.js:435:5
  
  
  - Resolver.js:435 
    [no-specs-vue-2]/[enhanced-resolve]/lib/Resolver.js:435:5
  
  - child-compiler.js:169 
    [no-specs-vue-2]/[html-webpack-plugin]/lib/child-compiler.js:169:18
  
  - Compiler.js:551 finalCallback
    [no-specs-vue-2]/[webpack]/lib/Compiler.js:551:5
  
  - Compiler.js:577 
    [no-specs-vue-2]/[webpack]/lib/Compiler.js:577:11
  
  - Compiler.js:1196 
    [no-specs-vue-2]/[webpack]/lib/Compiler.js:1196:17
  
  
  - Hook.js:18 Hook.CALL_ASYNC_DELEGATE [as _callAsync]
    [no-specs-vue-2]/[tapable]/lib/Hook.js:18:14
  
  - Compiler.js:1192 
    [no-specs-vue-2]/[webpack]/lib/Compiler.js:1192:33
  
  - Compilation.js:2787 finalCallback
    [no-specs-vue-2]/[webpack]/lib/Compilation.js:2787:11
  
  - Compilation.js:3092 
    [no-specs-vue-2]/[webpack]/lib/Compilation.js:3092:11
  
  
  - Hook.js:18 Hook.CALL_ASYNC_DELEGATE [as _callAsync]
    [no-specs-vue-2]/[tapable]/lib/Hook.js:18:14
  
  - Compilation.js:3085 
    [no-specs-vue-2]/[webpack]/lib/Compilation.js:3085:38
  
  
  - Compilation.js:492 fn
    [no-specs-vue-2]/[webpack]/lib/Compilation.js:492:9
  
  
  
  - task_queues:96 processTicksAndRejections
    node:internal/process/task_queues:96:5
```

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -478,6 +478,18 @@ export class ProjectDataSource {
     return preferences[projectTitle] ?? null
   }
 
+  async getCodeGenGlobs () {
+    assert(this.ctx.currentProject, `Cannot find glob without currentProject.`)
+
+    const looseComponentGlob = '*.{js,jsx,ts,tsx,.vue}'
```
tbiethman, line 484: Looks like there's an extra period for vue here:

```suggestion
    const looseComponentGlob = '*.{js,jsx,ts,tsx,vue}'
```

packages/app/src/specs/CreateSpecModal.vue
```diff
@@ -105,7 +111,15 @@ const specFileName = computed(() => {
   return getPathForPlatform(props.gql.currentProject?.defaultSpecFileName || '')
 })
 
-const filteredGenerators = getFilteredGeneratorList(props.gql.currentProject?.currentTestingType)
+const codeGenGlob = computed(() => {
+  if (!generator.value) {
+    return null
+  }
+
+  return props.gql.currentProject?.codeGenGlobs[generator.value.id]
```
tbiethman, line 119: Being unfamiliar with a lot of the `generator` design, this line threw me for a loop. Wasn't quite sure what was being looked up here. I eventually worked it out, but digging in left me wondering why we're deriving this here and passing it as a prop to the generator component? It seems like the generator component could determine this itself in a direct fashion, as it implicitly knows which type of generator it is, and leave the CreateSpecModal with less to care about. 

The same might be true for `specFileName` as well.

astone123: Yeah that was unnecessarily confusing. I updated it to just get whichever `codeGenGlob` it needs from `gql.currentProject`

packages/app/src/specs/CreateSpecModal.vue
```diff
@@ -77,6 +77,11 @@ fragment CreateSpecModal on Query {
     id
     fileExtensionToUse
     defaultSpecFileName
+    codeGenGlobs {
+      id
+      component
+    }
```
tbiethman, line 83: With your refactor, you can remove `codeGenGlobs` from the fragment.

astone123: I think that still needs to be in the query because I'm passing `gql.currentProject` via props to `ComponentGeneratorStepOne` and then using `gql.codeGenGlobs` there

tbiethman: Ah yes, I missed that the fragment in ComponentGeneratorStepOne was on CurrentProject. 

Should we import the `ComponentGeneratorStepOne_codeGenGlob` fragment here and use it in this fragment so that the prop expectations don't get out of sync? Or do the types keep these aligned for us?

packages/app/src/specs/generators/EmptyGenerator.vue
```diff
@@ -93,7 +93,7 @@
             query: {
               file: result.file.relative?.replace(/\\/g, '/')
             },
-            params: props.type === 'component'
+            params: props.type === 'component' || 'componentEmpty'
```
ZachJW34, line 96: Is the `componentEmpty` key used for both component and e2e empty spec generation? This creates an issue where the banner is shown for empty e2e specs in the runner. This banner should only be shown for component generated specs since it links to the `mount` documentation

packages/app/src/specs/CreateSpecModal.vue
lmiller1990: I think this should be the pattern. It's confusing when a component has GraphQL properties that aren't actually used there - much better to import the fragment and spread it.

I wonder if we can lint for "no unused GraphQL properties".

astone123: Okay I updated this to use the `ComponentGeneratorStepOne_codeGenGlob` fragment from `ComponentGeneratorStepOne`. Looks less confusing now

ZachJW34 (approved):
All my comments are addressed, tested and works great!

tbiethman (approved):
Works great in my local testing 👍 👍

ZachJW34 (approved):

marktnoonan:
Side note - I know this is already merged - I just finished testing this on Windows and all was good there for me 👍

JessicaSachs:
Seems like we removed the "Create from Blank Spec" button...? Bummer. I use that thing.

FYI Prior to this feature, I'd been using this nifty Code => VSCode snippet feature btw if you wanted inspiration for template syntax. Maybe you can grab the parser that this guy uses in his app and replicate it.

https://snippet-generator.app/?description=comp&tabtrigger=&snippet=import+%24%7B1%3AFileName%7D+from+%27.%2F%24%7B1%3AFileName%7D.vue%27%0A%0Adescribe%28%27%24%7B1%3AFileName%7D%27%2C+%28%29+%3D%3E+%7B%0A++it%28%27renders%27%2C+%28%29+%3D%3E+%7B%0A++++cy.mount%28%24%7B1%3AFileName%7D%0A++%7D%29%0A%7D%29&mode=vscode

astone123:
@JessicaSachs The "Create new empty spec" card should still be there along side of the "Create from component" card. Are you not seeing it?

Thanks for the info about the parser

PR #22901, opened 2022-07-22
fix: Allow typing seconds into datetime inputs

### LINKED ISSUE #22884: Type into datetime-local input: seconds not accepted
### Current behavior

When typing into an input datetime-local field, the only datetime format tolerated by Cypress doesn't accept seconds. 2022-03-04T10:11 is accepted (yyyy-MM-ddThh:mm) but 2022-03-04T10:11:12 is declined.

Although the [Cypress type page](https://docs.cypress.io/api/commands/type) doesn't give any format for datetime-local, [some Cypress runner warnings](https://github.com/cypress-io/cypress/issues/2613#issuecomment-536195767) say that yyyy-MM-ddThh:mm is supported with optional :ss or :ss.SSS suffixes, which isn't the case.

In the example code, Cypress throws the following if we want to type a DateTime with seconds:

> Typing into a datetime input with cy.type() requires a valid datetime with the format YYYY-MM-DDThh:mm, for example 2017-06-01T08:30. You passed: 2022-03-03T10:11:12 

### Desired behavior

When typing a datetime, Datetime formats should accept seconds.

### Test code to reproduce

```javascript
context('Datetime-Local picker', () => {
  it('insert a date with seconds', () => {
    cy.visit(
      'https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_type_datetime-local',
    );

    cy.get('#iframeResult', { timeout: 10000 })
      .its('0.contentDocument')
      .should('exist')
      .its('body')
      .should('not.be.undefined')
      .then((body) => {
        cy.wrap(body)
          .find('#birthdaytime')
          // Ensure that the DateTime input lets the user choose seconds
          .invoke('attr', 'value', '2022-01-02T01:02:03');
      });

    cy.get('#iframeResult', { timeout: 10000 })
      .its('0.contentDocument')
      .should('exist')
      .its('body')
      .should('not.be.undefined')
      .then((body) => {
        cy.wrap(body).find('#birthdaytime').clear().type('2022-03-03T10:11:12');
      });
  });
});
```

### Cypress Version

9.7.0

### Other

I did not try on Cypress 10.x.x because our company did not migrate to it yet. 

The real project uses React and more dependencies, as such I cannot set the input field's value using invoke.

marktnoonan:
- Closes #22884

### User facing changelog

`cy.type()` can now be used to enter datetime values with seconds, to an input of type `datetime-local`, for example `1959-09-13T10:12:13`. 

### Additional details

Seconds are valid in this datetime format, our regex was overly narrow.

### Steps to test

Run `packages/driver/cypress/e2e/commands/actions/type.cy.js` and see that the new tests are passing.

Run the test snippet provided in #22884 and see that it also passes now.

n/a

### PR Tasks


lmiller1990 (approved):

tbiethman (approved):

packages/driver/src/cy/keyboard.ts
```diff
@@ -65,7 +65,7 @@ const dateRe = /^\d{4}-\d{2}-\d{2}/
 const monthRe = /^\d{4}-(0\d|1[0-2])/
 const weekRe = /^\d{4}-W(0[1-9]|[1-4]\d|5[0-3])/
 const timeRe = /^([0-1]\d|2[0-3]):[0-5]\d(:[0-5]\d)?(\.[0-9]{1,3})?/
-const dateTimeRe = /^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}/
+const dateTimeRe = /^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}(:[0-9]{2})?(\.[0-9]{1,3})?/
```
emilyrohrbough, line 68: It's kind of odd we block keyboard input since of allowing users to ensure their input patterns are valid and handled as expected.

emilyrohrbough (approved):

packages/driver/src/cy/keyboard.ts
marktnoonan: I thought that was odd as well. I believe in this case it's because we don't actually simulate typing, we set the value for the provided datetime all at once (detecting the need to do this with [this helper here](https://github.com/cypress-io/cypress/blob/79f6651b1b5b93a88096c8d9d256a76c0b650528/packages/driver/src/dom/elements/elementHelpers.ts#L5). I guess this standardizes things since users don't ever type a raw date time, they use the browser UI to set individual parts of the date and time, or pick something from a calendar widget etc. It does seem a bit imprecise to use `type()` for this really.

PR #22912, opened 2022-07-25
chore: add check-ts command to pkg/driver

sainthkh:
- Closes 

### User facing changelog

N/A. It's an internal code fix. 

### Additional details
- Why was this change necessary? -> `pkg/driver` is really complex code blob and it won't be refactored soon. But it's a core of all Cypress commands. So, types should be checked to avoid possible bugs.
- What is affected by this change? => N/A.
- Any implementation details to explain? => N/A.

### Steps to test

Go to driver package and run `yarn check-ts`

N/A

### PR Tasks


cli/types/cypress.d.ts
```diff
@@ -3066,7 +3066,12 @@ declare namespace Cypress {
    * Takes ComponentDevServerOpts to track the signature of the devServerConfig for the provided `devServer`,
    * so we have proper completion for `devServerConfig`
    */
-  type ConfigOptions<ComponentDevServerOpts = any> = Partial<UserConfigOptions<ComponentDevServerOpts>>
+  type ConfigOptions<ComponentDevServerOpts = any> = Partial<UserConfigOptions<ComponentDevServerOpts>> & {
+    /**
+     * Hosts mappings to IP addresses.
+     */
+     hosts?: null | {[key: string]: string}
+  }
```
sainthkh, line 3074: This should be added here because we can add `hosts` to `definedConfig` like `driver/cypress.config.ts`.

packages/driver/src/cross-origin/events/socket.ts
```diff
@@ -1,4 +1,4 @@
-import { client } from '@packages/socket'
+import { client } from '@packages/socket/lib/browser'
```
sainthkh, line 1: `client` is now at `lib/browser`. It is not default exported.

packages/driver/src/cy/commands/navigation.ts
```diff
@@ -1138,6 +1138,8 @@ export default (Commands, Cypress, cy, state, config) => {
           // throw an error, else we'd be in a endless loop,
           // we also need to disable retries to prevent the endless loop
           if (previouslyVisitedLocation) {
+            // _retires is `private`. We're not using `retries()` method here because it breaks some tests.
+            // @ts-ignore
```
sainthkh, line 1142: For some reason, `retries()` method doesn't work here [for some tests](https://app.circleci.com/pipelines/github/cypress-io/cypress/40975/workflows/dbb89f4b-9b26-444a-83e9-33937f703753/jobs/1693561). I used `//@ts-ignore` to avoid this failure.

packages/driver/src/cypress/command_queue.ts
```diff
@@ -64,8 +64,18 @@ export class CommandQueue extends Queue<$Command> {
...
+
+  constructor (
+    state: StateFunc,
+    timeout: $Cy['timeout'],
+    stability: IStability,
+    cleanup: $Cy['cleanup'],
+    fail: $Cy['fail'],
+    isCy: $Cy['isCy'],
+    clearTimeout: ITimeouts['clearTimeout'],
+    setSubjectForChainer: $Cy['setSubjectForChainer'],
```
sainthkh, line 77: Added new function as a parameter to avoid type failure.

sainthkh (commented):
I added some notes on the changes.

rachelruderman (commented):
Hi @sainthkh, thank you for your contribution!! Two questions for ya

1. Can you update the description with `Steps to test`? It may be as simple as "Go to driver package and run `yarn check-ts`" but it's nice to have it explicitly stated 🙂 
2. I ran `yarn check-ts` in the driver package and it came back with 2 TypeScript errors, I believe from changes that came in after your PR was opened

cli/types/net-stubbing.ts
```diff
@@ -0,0 +1,553 @@
+// Copied from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/methods/index.d.ts
```
rachelruderman, line 1: How come we're copying instead of importing [@types/methods](https://www.npmjs.com/package/@types/methods)?

emilyrohrbough (changes requested):

packages/config/src/browser.ts
```diff
@@ -137,7 +137,13 @@ export const matchesConfigKey = (key: string) => {
...
 }
 
-export const validate = (cfg: any, onErr: (property: string) => void) => {
+interface ValidationResult {
+  key: any
+  type: any
+  value: any
+}
+
+export const validate = (cfg: any, onErr: (property: string | ValidationResult) => void) => {
```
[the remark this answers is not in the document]
sainthkh: Thanks, changed it.

cli/types/net-stubbing.ts
sainthkh: It seems that this file somehow got into this PR when I was switching between branches. Removed it.

rachelruderman (approved):
Looking good!! Thank you for your contribution and your patience @sainthkh 😸

cli/types/cypress.d.ts
BlueWinds: If we have `hosts?`, do we also need `null | {...}`. Feels like we should have undefined or a value, rather than undefined, null or a value; Is there a difference between null and undefined here? (and if there is, why would we do that >.> )

packages/driver/src/cy/commands/navigation.ts
```diff
@@ -1138,6 +1138,8 @@ export default (Commands, Cypress, cy, state, config) => {
           // throw an error, else we'd be in a endless loop,
           // we also need to disable retries to prevent the endless loop
           if (previouslyVisitedLocation) {
+            // _retires is `private`. We're not using `retries()` method here because it breaks some tests.
```
BlueWinds, line 1141:
```suggestion
            // _retries is `private`. We're not using `retries()` method here because it breaks some tests.
```

BlueWinds (approved):

cli/types/cypress.d.ts
sainthkh: I copied the type from below:

https://github.com/cypress-io/cypress/blob/3d98f9813619af5485ba6aef03ec596deaacfa6f/cli/types/cypress.d.ts#L2950-L2953

As it is inside `RuntimeConfigOptions`, I think they're the same type.

emilyrohrbough (approved):

PR #22914, opened 2022-07-25
chore: do not rely on transitive dep. from npm/react

lmiller1990:
- Closes N/A

### User facing changelog

N/A

### Additional details

I removed an unused dependency from `npm/react`, `cypress-plugin-snapshots`. Surprisingly, things broke. It turned out we were relying on a transitive dependency, which isn't ideal.

`cypress-plugin-snapshots` -> `socket.io-client` -> `engine.io-client`

I added the required dependency to `app` and `launchpad`. This still isn't ideal. We do `patch-package` in `socket`, as well as `nohoist` - I don't think there should be any socket.io related dependencies at the top level `node_modules` at all. I am unsure why there are there at all. I could not figure out why after a whole day, so I'm leaving it here - this is at least better than relying on a dependency's dependency, especially one that isn't even in `packages`.

### Steps to test

Observe CI passes - this confirms nothing is broken.

### PR Tasks

### DIFF
--- a/npm/react/package.json
+++ b/npm/react/package.json
@@ -21,7 +21,6 @@
     "@vitejs/plugin-react": "1.3.1",
     "axios": "0.21.2",
     "cypress": "0.0.0-development",
-    "cypress-plugin-snapshots": "1.4.4",
     "prop-types": "15.7.2",
     "react": "16.8.6",
     "react-dom": "16.8.6",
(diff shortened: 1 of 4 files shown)


lmiller1990:
I think I can do this as the same time as Vite 3: https://github.com/cypress-io/cypress/pull/22915

lmiller1990:
Doing this in here: https://github.com/cypress-io/cypress/pull/22915/files

Issue #22784, opened 2022-07-14
Vite 3 not working with Cypress Component Testing


lmiller1990:
### Current behavior

Vite 3 does not work.

### Desired behavior

It should work.

### Test code to reproduce

I created a PR with a failing test: https://github.com/cypress-io/cypress/pull/22748

It has some info, but there's probably more to it than the one fix suggested by the Vite team: https://github.com/vitejs/vite/issues/9059. Let's track progress here.

### Cypress Version

10.3.0

mHooshdar:
I tested it and the cypress was stuck on mounting in the component testing menu.

Do these changes fix the issue?

wopian:
> I tested it and the cypress was stuck on mounting in the component testing menu.
> 
> Do these changes fix the issue?

Adding the following to `vite.config.*` from the Vite 3 migration guide has made Cypress run without issues for me:

```ts
server: {
  port: 3000,
  host: '127.0.0.1'
}
```

lmiller1990:
Nice, we will make this change internally soon, so you don't need to 🎉

reintroducing:
There is an additional issue I found after getting tests up and running with Vite 3. I am using msw to mock all endpoints (in development and during testing) and i'm achieving that by basically pre-loading all the mock files I have defined as such:
```js
const req = import.meta.globEager('./**/index.js');
const modules = [];

Object.keys(req).forEach(key => {
    modules.push(...req[key].default);
});

export default modules;
```
As of Vite 3, `globEager` [is deprecated](https://vitejs.dev/guide/migration.html#import-meta-glob). I upgraded my code to the following:
```js
const req = import.meta.glob('./**/index.js', {eager: true});
const modules = [];

Object.keys(req).forEach(key => {
    modules.push(...req[key].default);
});

export default modules;
```
This works great in development mode, but it fails when the mocks get loaded in Cypress. I'm not sure where the issue is, but I suspect that, because it's still using Vite < 3 under the hood somewhere, its trying to load using `import.meta.glob` instead of reading the `{eager: true}` option and thus its not resolving the modules correctly.

Is this an issue with Cypress or perhaps `@cypress/vite-dev-server` (since I see that is still using v2.9.x of Vite)?

lmiller1990:
I tried updating to Vite 3.0.2 in my project, worked fine, no need for any workaround.

I'm not sure why Cypress internal tests are failing, likely due to some additional layer of complexity. 

[...]

PR #22915, opened 2022-07-25
feat: update to Vite 3

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/22784

### User facing changelog

- Update Cypress monorepo to use Vite 3
- As part of the Vite 3 update, I fixed a bug in our `vite-dev-server` integration that was introduced by a breaking change in Vite 3. The bug was Vite plugins that modify `index.html` were not correctly inserted into the `index.html` (eg `@vitejs/plugin-react`, [example issue](https://github.com/vitejs/vite/issues/9362).

### Additional details

Vite 3 came out, it had some breaking changes. [Breaking changes](https://vitejs.dev/guide/migration.html#advanced). [[#6901] fix: sequential injection of tags in transformIndexHtml](https://github.com/vitejs/vite/pull/6901) broke us, specifically.

It's fixed in this PR, [see here](https://github.com/cypress-io/cypress/pull/22915/files#diff-5b3ea0fa70e7e1e8a968123451e3b68ad11cde664c56ef07e4ccdd02769887e5R56-R94). 

Note: Percy has 51 changes - I actually **fixed** a bug, Percy was capturing the UI in the incorrect state. See [this comment](https://github.com/cypress-io/cypress/pull/22915#discussion_r929427553) on the problem. I will hit approve before we merge, once everyone is happy the PR.

### Steps to test

CI should be passing, and since we use Vite internally a lot, if it's passing, everything should be :100:. 

I'd also recommend pulling down the repo, and doing a `yarn dev` in something like `packages/app` with component testing, and making sure it all works as expected.

If you like, you could try replicate this user issue and make it it works: https://github.com/vitejs/vite/issues/9362 (we have tests for this specific case - using React and the Vite plugin - but good to be certain).

Can use Vite 3 with Cypress Component Testing.

### PR Tasks


npm/react/package.json
```diff
@@ -21,7 +21,6 @@
     "@vitejs/plugin-react": "1.3.1",
     "axios": "0.21.2",
     "cypress": "0.0.0-development",
-    "cypress-plugin-snapshots": "1.4.4",
```
lmiller1990, line 24: This was actually messing things up by introducing a transitive dependency. I wrote about it [here](https://github.com/cypress-io/cypress/pull/22914#issue-1316516145) and ultimately did the change as part of this PR.

npm/vite-dev-server/src/plugins/cypress.ts
```diff
@@ -52,21 +53,45 @@ export const Cypress = (
...
     },
-    async transformIndexHtml () {
+    async transformIndexHtml (html) {
+      // it's possibe other plugins have modified the HTML
+      // before we get to. For example vitejs/plugin-react will
+      // add a preamble. We do our best to look at the HTML we
+      // receive and inject it.
+      // For now we just grab any `<script>` tags and inject them to <head>.
+      // We will add more handling as we learn the use cases.
+      const root = parse(html)
```
lmiller1990, line 63: So, what this is doing is basically grabbing any `<script>` tags from the `html` that might have been inserted by other plugins and sticking them inside our HTML. An example is the vitejs/react-plugin, which injects this code: https://github.com/vitejs/vite/blob/main/packages/plugin-react/src/fast-refresh.ts#L32-L36. A user already reported this: https://github.com/vitejs/vite/issues/9362.

npm/vite-dev-server/src/plugins/cypress.ts
```diff
@@ -52,21 +53,45 @@ export const Cypress = (
...
+
+      const scriptTagsToInject = root.childNodes.filter((node) => {
+        // @ts-ignore
+        return node.rawTagName === 'script'
```
lmiller1990, line 67: Also, the type defs in `node-html-parser` seem to be wrong.

packages/app/vite.config.mjs
```diff
@@ -1,4 +1,4 @@
-import { makeConfig } from '../frontend-shared/vite.config'
+import { makeConfig } from '../frontend-shared/vite.config.mjs'
```
lmiller1990, line 1: So Vite is all-in on ESM it seems, we cannot use `ts` like this. We either need `type: module`, or something... I'm not sure what, I think `vite.config.mts` should also work, except we need to update TypeScript, which is pretty involved. The least invasive change is to use `mjs` here.

packages/frontend-shared/src/components/ShikiHighlight.vue
```diff
@@ -72,51 +72,12 @@ shikiWrapperClasses computed property.
   </div>
 </template>
 
-<script lang="ts">
```
lmiller1990, line 75: Moved to it's own file, Vite is more strict and complained about this export.

In general I think it's better if `vue` files only have 1 default export, anyway.

packages/frontend-shared/src/styles/shared.scss
```diff
@@ -10,17 +10,6 @@
     local("Roboto-Light"), local("DroidSans"), local("Tahoma");
 }
 
-/* Set up Roboto for modern browsers, all weights */
```
lmiller1990, line 13: This was not doing anything anyway, and it looks like something changed internally in Vite that was causing this to 404, so I just deleted it.

system-tests/projects/vite3.0.2-react/package.json
```diff
@@ -0,0 +1,8 @@
+{
```
lmiller1990, line 1: Above file has the Vite 3 test:

https://github.com/cypress-io/cypress/pull/22915/files#diff-e6bb614dd5b32e07eb125231dc627f64eb0e9a36b6c39cc7b6d23e802b9404b1

npm/vite-dev-server/src/plugins/cypress.ts
mike-plummer: I think this is because `childNodes` is an array of `Node` which is an abstract class - implementations include `TextNode` for comments (which have no tag name) and `HTMLElement` which has the `rawTagName` field. To make TS happy you'd need to do a type-check here then you could use the field name without the linter override

mike-plummer (approved):
Pulled down locally and everything seems to be working as expected. Left a couple "whatabout" comments that I'll leave up to you whether to incorporate, and one for a TS issue that might be good to fix

npm/vite-dev-server/src/plugins/cypress.ts
lmiller1990: Had a similar instinct, but I ran into a few roadblocks with this. With regards to using the `{ html, tags }` API for the HTML, by default we inject something like this:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width,initial-scale=1.0">
    <title>Components App</title>
  </head>
  <body>
    <div data-cy-root></div>
  </body>
</html>
```

Let's call this `baseHtml`. Users can customize this. We could do:

```js
return {
  html: baseHtml,
  tags: /* tags to inject */
}
```

Injecting the `script` content is pretty awkward, though, since we need to parse the `attrs` to use the `tags` API, eg:

```js
const tags = scriptTagsToInject.map<HtmlTagDescriptor>(script => ({
  tag: 'script',
  injectTo: 'head',
  attrs: {
    // is it src="..." or module="..."
  }
}))

return {
  html: newHtml,
  tags
}
```

So we need to parse the tag in even more detail. For many plugins, this API works great, like `plugin-react`, since they know they want to inject `type="module"`: [see here](https://github.com/vitejs/vite/blob/main/packages/plugin-react/src/index.ts#L390-L392). For us to use this API, we need to completely pass everything, which can include `<script>`, `<div>`, etc - this seems challenging and prone to error.

I could be missing something, it would be much nicer to use the Vite API than string substitution, definitely open to improving this if you see some alternatives. Right now there are definitely edge case, for example, if another plugin injects a `<div>` or something in `<head>` we won't grab it.

npm/vite-dev-server/src/plugins/cypress.ts
lmiller1990: Nice, I just did:

```js
const scriptTagsToInject = root.childNodes.filter((node) => {
  return node instanceof HTMLElement && node.rawTagName === 'script'
})
```

packages/app/vite.config.mjs
lmiller1990: Good idea, done: https://github.com/cypress-io/cypress/issues/22945.

I think it's probably a config and/or TS version related issus.

npm/vite-dev-server/src/plugins/cypress.ts
lmiller1990: Here: https://github.com/cypress-io/cypress/pull/22915/commits/0d12d60b24b720fa62e41a5687ce7d62a604e225

npm/react/package.json
```diff
@@ -30,7 +29,7 @@
     "rollup": "^2.38.5",
     "rollup-plugin-typescript2": "^0.29.0",
     "typescript": "^4.2.3",
-    "vite": "2.9.5",
+    "vite": "3.0.2",
```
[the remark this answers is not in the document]
lmiller1990: I updated to Vite 3.0.3! https://github.com/cypress-io/cypress/pull/22915/commits/84ee67ec9b5a7f4faee532f1185d9080cdcb7cd5

npm/vite-dev-server/src/plugins/cypress.ts
```diff
@@ -52,21 +53,45 @@ export const Cypress = (
...
+
+      let indexHtmlContent = await fs.promises.readFile(indexHtmlPath, { encoding: 'utf8' })
+
+      // Inject the script tags
+      indexHtmlContent = indexHtmlContent.replace(
+        '<head>',
+        `<head>
+        ${scriptTagsToInject.map((script) => script.toString())}
+      `,
+      )
```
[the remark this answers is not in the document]
lmiller1990: Both fine, I think I like `let` + reassign a little more than `then`, but good to know this is an option, too.

Thoughts on mixing `await` and `then`? I usually pick one or the other, but maybe combining them can make code more readable (and avoid mutation).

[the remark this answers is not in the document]
mike-plummer: I agree that it's usually best to avoid mixing `async/await` and `then` paradigms. I guess there is one other option to just wrap the await statement with parens but that isn't super-elegant either, so feel free to keep it the way it is:

```
const indexHtmlContent = (await fs.promises.readFile(indexHtmlPath, { encoding: 'utf8' }))
        // Inject the script tags
        .replace(
          '<head>',
          `<head>
          ${scriptTagsToInject.map((script) => script.toString())}
        `,
        )
```

npm/vite-dev-server/src/plugins/cypress.ts
mike-plummer: Yuck. Makes sense to just stick with the string sub for now, adding that level of additional parsing feels very brittle and error-prone

ZachJW34 (approved):
Tested it out and it works great! LGTM!!

PR #22918, opened 2022-07-25
chore: Fix glob pattern

rockindahizzy:
- Closes 

### Additional details

When running `yarn test {spec pattern}` in `system-tests` an error would occur. It looks like zsh may behave differently on m1 with how it treats cmd args.

### Steps to test

Before Change:
- On m1, run yarn test visit
- Fails to run with `The "path" argument must be of type string. Received an instance of Array`
After Change:
- On m1 run yarn test visit
- Runs successfully

### PR Tasks

### DIFF
--- a/system-tests/package.json
+++ b/system-tests/package.json
@@ -10,7 +10,7 @@
     "clean-deps": "find . -depth -name node_modules -type d -exec rimraf {} \\;",
     "preprojects:yarn:install": "yarn clean-deps",
     "projects:yarn:install": "node ./scripts/projects-yarn-install.js",
-    "test": "node ./scripts/run.js --glob-in-dir={test,test-binary}",
+    "test": "node ./scripts/run.js --glob-in-dir='{test,test-binary}'",
     "test:ci": "node ./scripts/run.js",
     "update:snapshots": "SNAPSHOT_UPDATE=1 npm run test"
   },
--- a/system-tests/package.json
+++ b/system-tests/package.json
@@ -10,7 +10,7 @@
     "clean-deps": "find . -depth -name node_modules -type d -exec rimraf {} \\;",
     "preprojects:yarn:install": "yarn clean-deps",
     "projects:yarn:install": "node ./scripts/projects-yarn-install.js",
-    "test": "node ./scripts/run.js --glob-in-dir='{test,test-binary}'",
+    "test": "node ./scripts/run.js --glob-in-dir=\"{test,test-binary}\"",
     "test:ci": "node ./scripts/run.js",
     "update:snapshots": "SNAPSHOT_UPDATE=1 npm run test"
   },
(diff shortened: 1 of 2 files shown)


system-tests/package.json
```diff
@@ -10,7 +10,7 @@
     "clean-deps": "find . -depth -name node_modules -type d -exec rimraf {} \\;",
     "preprojects:yarn:install": "yarn clean-deps",
     "projects:yarn:install": "node ./scripts/projects-yarn-install.js",
-    "test": "node ./scripts/run.js --glob-in-dir={test,test-binary}",
+    "test": "node ./scripts/run.js --glob-in-dir='{test,test-binary}'",
```
emilyrohrbough, line 13: does this impact our tests in CI on windows? I've seen glob patterns picked up differently on linux vs windows.

rockindahizzy: Good question, I hadn't thought of that! I used this as an opportunity to dig through the ci pipeline for the first time. It looks like we use the `test:ci` script in `circle.yml` which doesn't utilize in the `--glob-in-dir` option. 

It looks like we pass the file list in directly:
- [Here `@619`](https://github.com/cypress-io/cypress/blob/fix-system-test-glob-pattern/circle.yml#L619)
- [Here `@637`](https://github.com/cypress-io/cypress/blob/fix-system-test-glob-pattern/circle.yml#L637)

rockindahizzy: It still makes me wonder if I should try running on a windows box to make sure it doesn't blow up a dev.

lmiller1990: Another way to check is adding your branch name here: https://github.com/cypress-io/cypress/blob/7c3a4f8bb60fe7f5e3e6cc6719500ba9864390ba/circle.yml#L68. Then the windows tests will run on CI.

That said, we don't actually run the system tests on windows anyway, so you'd need to boot a windows machine to find out.

rockindahizzy: I tested on windows and found that `\"` gets desirable results on mac + windows. I also added my branch to the windows workflow, so we'll see how that goes!

lmiller1990 (approved):
Rocky tested on windows, he said it's working great 👍

system-tests/package.json
mirobo: > I tested on windows and found that `\"` gets desirable results on mac + windows. I also added my branch to the windows workflow, so we'll see how that goes!

@rockindahizzy Thanks for the fix of my "half" fix :D

You could probably run into the same problem in "packages\server\package.json" on M1?

    "test-integration": "node ./test/scripts/run.js --glob-in-dir=test/integration",
    "test-performance": "node ./test/scripts/run.js --glob-in-dir=test/performance",
    "test-unit": "node ./test/scripts/run.js --glob-in-dir=test/unit"

Example from README.MD:
yarn test-integration cli_spec ## shorthand, uses globbing to find spec

But I wonder if this isn't just an issue of this "minimist" package which handles the CLI args? Also I don't understand why there are almost two similar "run.js" scripts :D

PR #22922, opened 2022-07-25
fix(cypress/schematic): add --e2e option. ensure binary is installed in Angular 14

### LINKED ISSUE #21346: @cypress/schematics: cypress binary is no longer installed with CLI v14
### Current behavior

Currently, when running `ng add @cypress/schematic` the Cypress binary is automatically installed

This is no longer the case when using Angular CLI v14 (currently v14.0.0-next.13), as the CLI introduced a breaking change that no longer runs postinstall scripts

See https://github.com/angular/angular-cli/releases/tag/14.0.0-next.13 for the breaking change and a potential solution

### Desired behavior

The schematics should still install the cypress binary out of the box with CLI v14+

### Test code to reproduce

```
npm i -g @angular/cli@14.0.0-next.13
ng new binary --defaults
cd binary
ng add @cypress/schematic --defaults --skip-confirmation
ng e2e
```

throws:
```
No version of Cypress is installed in: /Users/cedric/Library/Caches/Cypress/9.6.0/Cypress.app

Please reinstall Cypress by running: cypress install

----------

Cypress executable not found at: /Users/cedric/Library/Caches/Cypress/9.6.0/Cypress.app/Contents/MacOS/Cypress

----------

Platform: darwin-x64 (21.4.0)
Cypress Version: 9.6.0
```

### Cypress Version

9.6.0

### Other

This can be workaround by installing the Cypress binary manually installed

### LINKED ISSUE #22829: Error when using --e2e with Angular CLI on command line
### Current behavior

With an `angular.json` using `@cypress/schematic:cypress` as `builder` for `e2e` tests, running `ng e2e` makes Cypress ask what type of tests should be run (e2e or component), and subsequently which browser to use. Per https://docs.cypress.io/guides/references/migration-guide#Migrating-to-Cypress-version-10-0, this can be fixed by passing `--e2e` and `--browser`. **Adding the `--e2e` results in an error**: "Error: Unknown argument: e2e"; adding `--browser` does work.

### Desired behavior

**When the `--e2e` flag is added to the `ng e2e` command, this should be used by Cypress to select e2e testing as the testing type automatically.**

This does already work when adding `"e2e": true` to the options in the `angular.json`, but with many (16 and counting) projects configured in the `angular.json`, this results in a lot of duplicate configuration. Adding `--e2e` to a script that runs `ng e2e` avoids the duplicate configuration and the need to add the configuration for every project we add.

### Test code to reproduce

Run `ng e2e --e2e` without adding `"e2e": true` to the options for "e2e" in the `angular.json`.

### Cypress Version

10.3.0

### Other

`@cypress/schematic` version: 2.0.0

admah:
## Changes/Updates
- Moved sandbox testing into `system-tests`, and fully tests schematic against Angular 13 and 14.
- Removed sandbox folders from schematic folder.
- Added `e2e` as a builder option to be passed as default (until CT arrives) or via CLI by a user.
- Updated `utilties` folder name to `utils`.

_____

- Closes #22829 
- Closes #21346


npm/cypress-schematic/sandbox13/angular.json
```diff
@@ -3,7 +3,7 @@
   "version": 1,
   "newProjectRoot": "projects",
```
lmiller1990, line 4: I wonder if we can reuse the projects in `system-tests` so we don't need to add so many test projects?

image

lmiller1990 (commented):
Build is :x: 

Left one comment about how we can reduce the number of test projects :+1:

admah:
@lmiller1990 thanks for that suggestion! I'd like to remove the extra test projects if possible.

jordanpowell88:
I agree. Some system tests would be awesome. I'm happy to help @admah

ZachJW34 (approved):
LGTM!

PR #22925, opened 2022-07-25
test: initial cypress-axe setup for e2e and CT

marktnoonan:
### User facing changelog

N/A This is not a change to Cypress itself, just a change to our own tests.

### Additional details

This will provide us with some logging of accessibility violations from cypress-axe when running a spec in open mode. Since Percy snapshots are intended to be taken of each unique state in the application, they are the perfect place to run an accessibility check, and should reach all the nooks and crannies where state might lead to different elements rendering. Due to the number of percy snapshots we currently take, the end result is about 400 places where we will now have accessibility-related information logged to the reporter.

These checks are not yet intended to generate any reporting from runs, and since running the checks in run mode seemed to produce a little bit of CI flake, we avoid running them except in open mode. More work might come about in the future to run this in CI and enable some reporting.

For now though, hopefully, as we are developing, these logs might help us notice if some alt text is missing, or a button has no label, or similar low hanging fruit that is easy to correct. Likewise it will flag if color contrast is not passing and allow us to follow up with design engineers about possibly choosing different values.

This is an example of how the logs look, reporting that in the selector playground we are missing some button labels:

Screen Shot 2022-07-25 at 12 34 41 PM

### Steps to test

Run some specs in CT or E2E that use `cy.percySnapshot()` and look at the command log.

N/A

### PR Tasks


packages/frontend-shared/cypress/support/component.ts
```diff
@@ -76,5 +76,13 @@ Cypress.on('uncaught:exception', (err) => !err.message.includes('ResizeObserver
 
 registerMountFn()
 addVueCommand()
-installCustomPercyCommand()
+installCustomPercyCommand({
+  isComponentTesting: true,
+  elementOverrides: {
+    'svg.animate-spin': ($el) => {
+      $el.attr('style', 'animation: none !important')
```
marktnoonan, line 83: Adding this to ward off Percy flake related to spinners in frontend-shared.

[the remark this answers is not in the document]
rachelruderman: Just wondering, do you have context on why we use `elementOverrides` as opposed to [percyCss](https://docs.percy.io/docs/percy-specific-css#snapshot-options--sdk-options)? percyCss seems a bit easier to work with than jquery, but they both get the job done!

packages/ui-components/cypress/support/customPercyCommand.ts
```diff
@@ -54,6 +55,54 @@ declare global {
   }
 }
 
+function setupAxeAndCheckA11y (isComponentTesting?: boolean) {
+  if (isComponentTesting) {
+    Cypress.Commands.add('injectAxe', () => {
+      // this is a work around for the issue with require.resolve in CT
+      // described here: https://github.com/component-driven/cypress-axe/issues/134
+      cy.window({ log: false }).then((window) => {
```
rachelruderman, line 63: Small insignificant nit! It would be nice if we could dynamically determine `isComponentTesting` instead of passing it around. We coullllldddddd do something like the following. I'm not familiar with the danger zones of this approach tho so take this with a grain of salt 🧂 

```
function setupAxeAndCheckA11y () {
  cy.window({ log: false }).then((window) => {
    const isComponentTesting = window.__CYPRESS_TESTING_TYPE__ === 'component'
    // rest of logic as you had it
  })
}
```

packages/frontend-shared/cypress/support/component.ts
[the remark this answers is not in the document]
rachelruderman: Omg forget about percyCSS, the [Percy media query option](https://docs.percy.io/docs/percy-specific-css#percy-css-media-query) is even slicker!!! Wonder why we're not using it?
Screen Shot 2022-07-28 at 11 14 54 AM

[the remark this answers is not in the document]
marktnoonan: From what I understand, the reason for this is so that, in open mode, we get a Cypress snapshot of the exact DOM state that will be sent to Percy when you hover over the "skipping percy snapshot in interactive mode", which is a good way of previewing what Percy will render (though, that log doesn't seem to output in CT, which I have not investigated).

packages/ui-components/cypress/support/customPercyCommand.ts
marktnoonan: Totally. The only reason not to do this is my pretty slight personal preference to avoid those global variables unless they are absolutely necessary. They are mostly there because in run mode we don't use GQL, so there are some things that typically would be read out of GQL, that get set on the window to help the Spec Runner initialize correctly in run mode.

Since the support files know what the testing type is, I figure we don't need to bind this to the global, and the function can just know what to do when it's called.

That said, who knows what a more fleshed out implementation of this will look like, might not even be in the same command, so I also see this as pretty temporary.

rachelruderman (approved):

packages/ui-components/cypress/support/customPercyCommand.ts
lmiller1990: Better yet, you can just do `Cypress.testingType`: https://docs.cypress.io/api/cypress-api/testing-type

How about using that?

lmiller1990 (commented):
One nit, I think we can use the public API `Cypress.testingType` instead of passing an argument.

packages/ui-components/cypress/support/customPercyCommand.ts
marktnoonan: one. `Cypress.testingType` feels like something that's supposed to be there. I briefly got interested in whether we can replace some of the other `__CYPRESS_TESTING_TYPE__` usages, but it wasn't an instant win so if there are some gains there we can do it in another PR.

lmiller1990 (approved):

PR #22926, opened 2022-07-25
refactor: use cp.fork to preserve stdout for packages/extension build

ZachJW34:
### User facing changelog
na

### Additional details
The extension package executes webpack inside of a gulp script. It was using a promisified `exec` script to invoke the `run-webpack.js` script and it wasn't piping any of the stdout. This PR uses `fork` to invoke webpack, making it easy to capture stdout with `stdio: 'inherit'`.

I ran into this when working through adding Angular CT support. Webpack 5 got hoisted and broke the webpack build but the error message was unclear. Now, if this script errors it will print the webpack logs.

### Steps to test
Run `yarn workspace @packages/extension build` and you'll see the webpack logs. You can tweak the src files to throw a compilation error to verify that a non-zero exit code causes the gulp task to fail.

Before:
Screen Shot 2022-07-25 at 5 53 53 PM

After:
Screen Shot 2022-07-25 at 5 47 52 PM

### PR Tasks


lmiller1990:
I ran into this exact same thing when updating to Vite 3, and could not figure out the error w/o stdout. Good fix.

packages/extension/gulpfile.ts
```diff
@@ -15,7 +16,11 @@ const manifest = () => {
 }
 
 const background = (cb) => {
-  exec('node ../../scripts/run-webpack.js').then(() => cb()).catch(cb)
```
lmiller1990, line 18: I thought `exec` had a second arg where you could pass

```js
exec('node ....', { stdout: ... }
```

But now I realize that is `execa`. Another option would be to grab the stdout from the callback:

```js
exec('node ...', (error, stdout, stderr) => {
  // stuff
})
```

It might be more idiomatic to do something like this, and do a `if (err)` check?

Either way is fine with me, I'll leave the final decision to you.

ZachJW34: I like the current implementation as it preserves the ansii highlighting, whereas console logging the `stdout` does not.

PR #22927, opened 2022-07-26
fix: Throw error when custom config file does not return expected value

### LINKED ISSUE #21948: Misconfiguring devServer for CT causes browser launching to fail silently
If you provide a value to the devServer configuration param that doesn't actually start up a dev server, you can't launch a browser from the runner, and theres no indication to the developer that anything is wrong.

```js
  component: {
    devServer: () => { }
  },
});
```

We should provide a meaningful message when the devServer doesn't start successfully so the developer can know where to investigate what might be wrong in their config.

Also, the specPattern option appears to be required even though there should be a default value set for it. If the specPattern is missing, the browser won’t launch as well.  Video example: [Loom](https://www.loom.com/share/39b94f15e213438d8c7778908c9e912f|smart-link)

https://user-images.githubusercontent.com/1271364/171253673-6dd65eb2-7ffe-4306-86a1-c3df3617df80.mp4

┆Issue is synchronized with this [Jira Bug](https://cypress-io.atlassian.net/browse/UNIFY-874) by [Unito](https://www.unito.io)
┆author: Ely Lucas
┆epic: CT Architecture
┆friendlyId: UNIFY-874
┆priority: Low
┆taskType: Bug

rockindahizzy:
- resolves  #21948

### User facing changelog

- Improve component testing configuration error messaging

Before

https://user-images.githubusercontent.com/25568640/181021302-c69156e8-aa92-4a60-b102-b6e7b6c84122.mov

After:

https://user-images.githubusercontent.com/25568640/181021356-a396c0e1-4231-4fc3-b758-f88daa5f64d7.mov

### PR Tasks


packages/launchpad/cypress/e2e/config-files-error-handling.cy.ts
```diff
@@ -294,4 +294,38 @@ describe('setupNodeEvents', () => {
...
+
+    cy.get('body')
+    .should('contain.text', cy.i18n.launchpadErrors.generic.configErrorTitle)
+    .and('contain.text', 'The returned value of the devServer function is not valid.')
+
+    cy.get('[data-cy="collapsible-header"]')
+    .should('have.attr', 'aria-expanded', 'true')
+    .contains(cy.i18n.launchpadErrors.generic.stackTraceLabel)
+
+    cy.log('Fix error and validate it reloads configuration')
```
lmiller1990, line 316: Nice, haven't seen `cy.log` used in this informative manner before, it's a good idea

system-tests/projects/dev-server-invalid/cypress/component/support.spec.ts
```diff
@@ -0,0 +1,14 @@
+describe('Support files', () => {
```
lmiller1990, line 1: I wonder if we need this? If not, might be good to remove it and set `supportFile: false` in the `cypress.config.js`, just to keep the example project and files to a minimum.

system-tests/projects/dev-server-invalid/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@test-project/vite-ct-function-api",
+  "version": "0.0.0-test",
+  "private": true,
+  "dependencies": {},
+  "devDependencies": {
+    "@cypress/vite-dev-server": "file:../../../npm/vite-dev-server",
+    "@testing-library/cypress": "8.0.2",
```
lmiller1990, line 8: Probably can make this example project just use `cypress.config.js` to keep the deps to a minimum, each project and dep makes the CI run time a little longer.

lmiller1990 (commented):
Few comments, mainly around removing extra files. Looks like this was a small fix, just knowing where to look and what to do is the hard part 😆

packages/launchpad/cypress/e2e/config-files-error-handling.cy.ts
rockindahizzy: Yea! It was in the test I stole from `it handles a configuration file error` in `error-handling.cy.ts` :)

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -233,7 +233,7 @@ export class ProjectLifecycleManager {
           const devServerOptions = await this.ctx._apis.projectApi.getDevServer().start({ specs: this.ctx.project.specs, config: finalConfig })
 
           if (!devServerOptions?.port) {
-            this.ctx.onError(getError('CONFIG_FILE_DEV_SERVER_INVALID_RETURN', devServerOptions))
+            throw getError('CONFIG_FILE_DEV_SERVER_INVALID_RETURN', devServerOptions)
```
lmiller1990, line 236: Should this be named something else? If it is the return type from the dev server function, should it be something like `devServerReturnVal`?

lmiller1990: I think this object is what we get from here: https://github.com/cypress-io/cypress/blob/develop/npm/webpack-dev-server/src/devServer.ts#L62-L76

system-tests/projects/dev-server-invalid/vite.config.js
```diff
@@ -0,0 +1 @@
+export default {}
```
lmiller1990, line 1: Can delete yarn.lock as well I think

lmiller1990 (approved):
Looks good, paired and went over it and made a few changes, mainly removing dead code.

mike-plummer (commented):
Just looking at edge cases and I'm able to get into the same "hanging, never-starting" dev server state by returning the following values.

({ port: -1 })
({ port: 'abc' })
({ port: [] })

We may want to consider expanding the validation here to properly bound allowable values for port. The bigger question I have is whether there's a way for the runner to time out after a minute or whatever if it thinks it has a good config but isn't able to confirm the server is launched, that way we don't necessarily need to validate every single edge case here and the user will eventually get meaningful feedback

rockindahizzy:
> Just looking at edge cases and I'm able to get into the same "hanging, never-starting" dev server state by returning the following values.
> 
> ({ port: -1 }) ({ port: 'abc' }) ({ port: [] })
> 
> We may want to consider expanding the validation here to properly bound allowable values for port. The bigger question I have is whether there's a way for the runner to time out after a minute or whatever if it thinks it has a good config but isn't able to confirm the server is launched, that way we don't necessarily need to validate every single edge case here and the user will eventually get meaningful feedback

That's a good point. There are also scenarios like setting valid ports. i.e. `3000` when it's really running on `3001` , which will also hang. I think in general we can really improve error messaging. Curious we should draw the line for this specific issue.

lmiller1990:
Original issue:

```js
  component: {
    devServer: () => { }
  },
});
```

Which is addressed. Definitely room for improvements around messaging if:

- the port is already used
- or you pass something invalid like "abc"

... and probably other properties we'd like to guard against, too. This requires a bit of thinking and research, and feels like a bit of scope creep over the original issue.

If someone wants to go ahead and create an issue with some thoughts, that'd be great. I am going to merge this one since as far as I can see it satisfied what we set out to do, which is ensure people return **something** from the dev server call.

PR #22940, opened 2022-07-26
fix: Infinite recursion when identical intercept routes exist

### LINKED ISSUE #22693: Infinite recursion in net-stubbing/lib/server/middleware/request.js
### Current behavior

We run our cypress tests on CircleCI, distributing the tests among several runners. Sometimes we get a failure like this one:
```
RangeError: Maximum call stack size exceeded
    at RegExp.get source [as source] (<anonymous>:null:null)
    at RegExp.toString (<anonymous>:null:null)
    at equalByTag (/root/.cache/Cypress/9.6.0/Cypress/resources/app/node_modules/lodash/lodash.js:5786:35)
    at baseIsEqualDeep (/root/.cache/Cypress/9.6.0/Cypress/resources/app/node_modules/lodash/lodash.js:3357:13)
    at baseIsEqual (/root/.cache/Cypress/9.6.0/Cypress/resources/app/node_modules/lodash/lodash.js:3316:14)
    at equalObjects (/root/.cache/Cypress/9.6.0/Cypress/resources/app/node_modules/lodash/lodash.js:5872:43)
    at baseIsEqualDeep (/root/.cache/Cypress/9.6.0/Cypress/resources/app/node_modules/lodash/lodash.js:3375:14)
    at baseIsEqual (/root/.cache/Cypress/9.6.0/Cypress/resources/app/node_modules/lodash/lodash.js:3316:14)
    at baseIsMatch (/root/.cache/Cypress/9.6.0/Cypress/resources/app/node_modules/lodash/lodash.js:3433:19)
    at /root/.cache/Cypress/9.6.0/Cypress/resources/app/node_modules/lodash/lodash.js:3603:37
    at baseFindIndex (/root/.cache/Cypress/9.6.0/Cypress/resources/app/node_modules/lodash/lodash.js:816:11)
    at Function.findIndex (/root/.cache/Cypress/9.6.0/Cypress/resources/app/node_modules/lodash/lodash.js:7322:14)
    at getRouteForRequest (/root/.cache/Cypress/9.6.0/Cypress/resources/app/packages/net-stubbing/lib/server/route-matching.js:105:77)
    at populateMatchingRoutes (/root/.cache/Cypress/9.6.0/Cypress/resources/app/packages/net-stubbing/lib/server/middleware/request.js:45:67)
    at populateMatchingRoutes (/root/.cache/Cypress/9.6.0/Cypress/resources/app/packages/net-stubbing/lib/server/middleware/request.js:50:13)
    at populateMatchingRoutes (/root/.cache/Cypress/9.6.0/Cypress/resources/app/packages/net-stubbing/lib/server/middleware/request.js:50:13)
... and on and on and on
```

### Desired behavior

Test should run without crashing

### Test code to reproduce

I haven't found a reliable way to reproduce this. It's not always the same test that fails. However, the failures I've seen so far occur in tests with a `beforeEach` function that calls a utility method, which uses `cy.intercept` for a lot of our API routes.

### Cypress Version

9.6.0

### Other

_No response_

eirikur-grid:
- Closes  #22693

### User facing changelog

 - Fixed an issue that could lead to infinite recursion and thus a crash when running tests that make use of `cy.intercept
`

### Additional details
This fixes a subtle bug in the `getRouteForRequest`  function, due to lodash's `findIndex `method performing a deep object comparison, rather than testing for reference equality. The simplest fix would be to use `Array.indexOf` instead to locate the previous route. However, I decided to turn it into a generator function instead in order to simplify the method and to increase efficiency. The time complexity of the `populateMatchingRoutes` function was _O(n*(m+1))_, where 

 - _n_=number or registered routes
 - _m_=number of matching routes

Using a generator function, the same is trivially achieved in _O(n)_ time by
```typescript
const matchingRoutes: BackendRoute[] = [...getRoutesForRequest(this.netStubbingState.routes, this.req)]
```

With that being said, it's not clear to me what could lead to identical routes being present. The infinite recursion errors occur seemingly at random in our CI runs, suggesting that there's a race condition. This won't fix the duplicate route issue, but it will prevent a crash.

### Steps to test

I've added a unit test to verify that duplicate routes will not lead to an infinite recursion. Beyond that, I think what needs to be verified is that cy.intercept hasn't regressed. I guess there are already automated tests that cover that.

N/A

### PR Tasks


packages/net-stubbing/lib/server/middleware/request.ts
```diff
@@ -41,21 +41,7 @@ export const InterceptRequest: RequestMiddleware = async function () {
...
-      return
-    }
-
-    matchingRoutes.push(route)
-
-    populateMatchingRoutes(route)
-  }
-
-  populateMatchingRoutes()
+  const matchingRoutes: BackendRoute[] = [...getRoutesForRequest(this.netStubbingState.routes, this.req)]
```
tbiethman, line 44: Very cool reduction here :+1:

packages/server/lib/server-e2e.ts
```diff
@@ -190,7 +190,9 @@ export class ServerE2E extends ServerBase<SocketE2E> {
       }
 
       // @ts-ignore
-      return !!getRouteForRequest(this.netStubbingState?.routes, proxiedReq)
+      const iterator = getRoutesForRequest(this.netStubbingState?.routes, proxiedReq)
+
+      return !iterator.next().done
```
tbiethman, line 195: Generators aren't used all that often in this code-base; a comment here explaining what this logic is validating could be helpful.

eirikur-grid: Yup, I find myself consulting MDN everytime I use the iterator API. I've added a comment and a (hopefully) reasonably named variable to make this less obscure.

packages/net-stubbing/test/unit/route-matching-spec.ts
```diff
@@ -209,15 +209,53 @@ describe('intercept-request', function () {
...
+          routeMatcher: {
+            pathname: '/foo',
+          },
+        },
+        {
+          id: '1',
+          routeMatcher: {
+            pathname: '/foo',
+          },
+        },
```
flotwig, line 236: Should `routes[0]` and `routes[1]` be the same object (by reference) for this to be a true repro of #22693?

flotwig (commented):
Had a question about the test but otherwise LGTM. Thanks for reminding me that generators exist in JS :smile: 

> With that being said, it's not clear to me what could lead to identical routes being present. The infinite recursion errors occur seemingly at random in our CI runs, suggesting that there's a race condition. This won't fix the duplicate route issue, but it will prevent a crash.

Really wish we knew WHY this was happening, but I'm glad we can at least fix this symptom... I wonder if this is causing other issues.

packages/net-stubbing/test/unit/route-matching-spec.ts
eirikur-grid: Having two identical objects was sufficient to cause problems due to how `_.findIndex` is implemented. The test runner would lock up with the original implementation in place. 
I'm still not certain whether the issue we've seen is caused by identical objects or a duplicate registration of the same object...so this repro is a bit speculative.

flotwig (approved):

tbiethman (approved):

PR #22957, opened 2022-07-27
fix: Ensure tooltips are displayed behind modals

### LINKED ISSUE #22595: Tooltip shows above overlays for login and switching testing type
Tooltip shows above overlays for login and switching testing type

mike-plummer:
- Closes #22595 

### User facing changelog
Ensure tooltips for overlaid content do not appear in front of modal dialogs

### Steps to test
Little bit tricky, but the easiest way to verify this is to get a tooltip to stick open - typically you can throw your mouse around the ACI column headers (Latest Runs & Average Duration) and eventually one will stay open after you've moused off (it'll close if you mouse back over or onto another item with a tooltip).

If you have trouble getting a tooltip to stick open you can try adding `:shown="true"` to `Tooltip.vue` around line 12 which should force at least one tooltip open without any interaction.

Once you get a tooltip stuck open, click the "New Spec" button and verify that the tooltip does not overlay the modal or its backdrop.

**Before**

Screen Shot 2022-07-27 at 11 09 53 AM

**After**

Screen Shot 2022-07-27 at 11 09 33 AM

### PR Tasks


packages/app/cypress/e2e/tooltips.cy.ts
```diff
@@ -0,0 +1,34 @@
...
+
+describe('App: Tooltips', () => {
+  beforeEach(() => {
+    cy.scaffoldProject('cypress-in-cypress')
+    cy.openProject('cypress-in-cypress')
+    cy.startAppServer('e2e')
+    cy.visitApp()
+  })
+
+  it('should layer beneath modals', () => {
```
ZachJW34, line 12: Does this make more sense as a CT test? You could mount a modal and render a tooltip inside it to verify behavior.

packages/frontend-shared/src/components/Tooltip.vue
```diff
@@ -77,6 +77,13 @@ const actualPopperClass = computed(() => {
   }
 }
 
+.v-popper__popper {
+  // Our modal overlays use a backdrop z-index of 10, but the tooltip lib defaults to 10000 which causes
+  // tooltips to overlay modals & modal backdrops. Override to set z-index so that tooltips aren't
+  // forced atop items they shouldn't be
+  z-index: 9;
```
ZachJW34, line 84: We have tooltips inside some popovers, which causes the tooltip to not overlay properly. Should this be a prop that gets passed in? We can default to a low z-index and than have an `in-modal` prop to bump the z-index up. There might be a clever way to detect the context of the tooltip and adjust

Screen Shot 2022-07-28 at 10 06 02 AM

ZachJW34 (changes requested):

packages/frontend-shared/src/components/Tooltip.vue
mike-plummer: Dang, missed that in Percy.

I started off trying to use `z-index: inherit` so the tooltip could infer an appropriate layer based on its parent, but that caused issues with other areas since the tooltips are inherently ad-hoc DOM elements attached to `body`. I'll poke at it some more to see if there's a clever way to address via CSS

mike-plummer:
Took another tack at this. Since `floating-vue` uses a fixed z-index we can't easily handle the under- and over-modal use cases without adding a new prop that users would have to provide based on where the tooltip is being used. That didn't seem ideal, so I'm instead trying to address the root issue here which is Tooltips remaining open when they shouldn't. This change has modals close any currently-open tooltips which should prevent the z-index issue from appearing.

lmiller1990 (approved):
"Just hide all the tooltips" seems like a good solution for now.

packages/frontend-shared/src/gql-components/topnav/LoginModal.vue
```diff
@@ -133,6 +134,18 @@ fragment LoginModal on Query {
 }
 `
 
+// Ensure all tooltips are closed when the modal opens - this prevents tooltips from beneath that
+// are stuck open being rendered on top of the modal due to the use of a fixed z-index in `floating-vue`
+watch(
```
ZachJW34, line 139: This is nice, `hideAllPoppers` is a good find

ZachJW34 (approved):

PR #22964, opened 2022-07-28
feat: warn when detecting unsupported dependencies for component testing

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/22025

### User facing changelog

N/A

### Additional details

Add some runtime checks for CT deps, like `vite`, `react`, etc. We do these during setup. Now, we also do it each launch. This will help warn users if they've updated a dependency (like Vite) to a new version we haven't yet got official support for.

### Steps to test

You will need a project with some mismatched dependencies. This PR includes one called `component-testing-outdated-dependencies`, you could test with that, if you want. Run yarn install inside it, fire up Cypress and choose that project. You should see the appropriate warnings.

Launchpad now shows a warning with the mismatched dependencies, what you should have provided, and what you did provide. TBA: the exact text, this screenshot is just showing how it looks right now. @christopherfrance will provide the correct text, as discussed [here](https://github.com/cypress-io/cypress/issues/22025#issuecomment-1193428399).

image

### PR Tasks


christopherfrance:
Thanks @lmiller1990 this looks like a good improvement from design perspective.

There are two changes being discussed in this thread:

DONE: Remove Restart button since it seems unlikely to work as expected.

TODO: The last line could more actionable / explain what to do next. Here is our suggestion (to
replace Cypress may not function as expected) — If you're experiencing problems, downgrade dependencies and restart Cypress.

packages/scaffold-config/src/dependencies.ts
```diff
@@ -40,7 +40,7 @@ export const WIZARD_DEPENDENCY_REACT_DOM = {
   package: 'react-dom',
   installer: 'react-dom',
   description: 'This package serves as the entry point to the DOM and server renderers for React',
-  minVersion: '>=16.x',
+  minVersion: '>=16.0.0 || >=17.0.0 || >=18.0.0',
```
ZachJW34, line 43: Should this be:
```suggestion
  minVersion: '^=16.0.0 || ^=17.0.0 || ^=18.0.0',
```

Otherwise, there is no change in behavior from `>=16.x`

packages/errors/src/errors.ts
```diff
@@ -1561,6 +1562,24 @@ export const AllCypressErrors = {
...
   },
+
+  COMPONENT_TESTING_MISMATCHED_DEPENDENCIES: (dependencies: DependencyToInstall[]) => {
+    const deps = dependencies.map<string>((dep) => {
+      if (dep.detectedVersion) {
+        return `${dep.dependency.type}. Expected ${dep.dependency.minVersion}, found ${dep.detectedVersion}.`
+      }
+
+      return `${dep.dependency.type}. Expected ${dep.dependency.minVersion} but dependency was not found.`
+    })
```
ZachJW34, line 1573: Wrapping the variables in a code block looks nicer to me
```ts
return `\`${dep.dependency.type}\`.
```

ZachJW34 (changes requested):

packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -153,8 +154,56 @@ export class ProjectConfigManager {
...
+    // and not using one of our presets. Assume they know what they are doing and
+    // what dependencies they require.
+    if (typeof this._cachedLoadConfig?.initialConfig?.component?.devServer !== 'object') {
+      return
+    }
+
+    const devServerOptions = this._cachedLoadConfig.initialConfig.component.devServer
+
+    const bundler = WIZARD_BUNDLERS.find((x) => x.type === devServerOptions.bundler)
+    const framework = WIZARD_FRAMEWORKS.find((x) => x.type === devServerOptions.framework)
```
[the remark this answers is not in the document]
lmiller1990: Hmm I was not able to reproduce this behavior:

image

Let me try a bit more.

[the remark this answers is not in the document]
lmiller1990: Argh, this is going to be a problem. Good catch.

packages/frontend-shared/src/warning/Warning.vue
```diff
@@ -68,3 +69,12 @@ let message = computed(() => {
 
 const { markdown } = useMarkdown(markdownTarget, message.value, { classes: { code: ['bg-warning-200'] } })
 </script>
+
+<style lang="scss">
+.warning-markdown {
```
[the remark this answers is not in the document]
lmiller1990: Done - I still need to do `.warning-markdown > ul` since you cannot nest selectors using Tailwind, but I changed it to use `@apply`.

The reason I did it like this is we need to style the `<ul>` which is part of the generated HTML in the `errors` package. I tried adding it with the [rest of the styles](https://github.com/cypress-io/cypress/blob/develop/packages/frontend-shared/src/composables/useMarkdown.ts#L46) but it's getting clobbered by something else and the styles aren't applied:
image

packages/scaffold-config/src/dependencies.ts
lmiller1990: Fixed!

lmiller1990:
@ZachJW34 @christopherfrance I updated the warning message and styling.

image

image

packages/errors/src/errors.ts
lmiller1990: Nice, done. See below post for screenshots.

lmiller1990: nice

lmiller1990:
@ZachJW34 updated the logic to handle the case you pointed out and added tests around it, good catch.

packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -153,8 +154,84 @@ export class ProjectConfigManager {
...
+    const frameworks = WIZARD_FRAMEWORKS.filter((x) => x.configFramework === devServerOptions.framework)
+
+    const mismatchedFrameworkDeps = new Map<typeof WIZARD_DEPENDENCIES[number]['type'], DependencyToInstall>()
+
+    let isSatisfied = false
+
+    for (const framework of frameworks) {
+      if (isFrameworkSatisfied(bundler, framework)) {
+        isSatisfied = true
+        break
```
ZachJW34, line 214: Would an early return make sense vs the break?

ZachJW34 (approved):
My comments are addressed! Tested it out and works great. One small question

rockindahizzy (approved):
LGTM!

packages/data-context/src/data/ProjectConfigManager.ts
lmiller1990: It wouldn't, because even if the framework is ok (eg, CRA), it's possible we want to warn about a mismatches bundler version (as unlikely as this seems, it's possible someone had a supported CRA version, but managed to install a old or new version of Webpack that we don't support).

ZachJW34: Forgot about the bundler, makes sense!

PR #22996, opened 2022-07-28
chore: addSelector and sample selector commands

BlueWinds:
- Related to https://github.com/cypress-io/cypress/issues/7306, but does not complete it.

### User facing changelog
No user facing changes.

Cypress.Commands.addSelector() is intentionally not documented, because it should be considered internal / unstable for now.

### Additional details
This PR is an initial implementation of Cypress.Commands.addSelector(), demonstrating the use of selector commands - commands which rely on the Cypress command queue for their retry logic, and can be retried as-needed to determine the subject of future commands.

It includes adds two new commands: `__internalSelectorGet` and `__internalSelectorShould`, which are selector-based versions of `cy.get()` and `cy.should()` respectively. It also copies most of the .get() tests to run against these new commands, demonstrating their functionality.

This functionality is *not* exercised completely enough for general use - there are more tests to write and more work to be done, but the current state of things is reviewable, and I'd like to merge it in to both avoid a long-running feature branch and also reduce the burden on reviewers as the PR as the functionality of selectors grows and evolves.

### Steps to test
Consider `querying-selector.cy.js`, and play around with `__internalSelectorGet` / `__internalSelectorShould`.

### PR Tasks


emilyrohrbough:
> This functionality is not exercised completely enough for general use - there are more tests to write and more work to be done, but the current state of things is reviewable, and I'd like to merge it in to both avoid a long-running feature branch and also reduce the burden on reviewers as the PR as the functionality of selectors grows and evolves.

Just curious, why create a PR with both in a half-finished state vs getting one command across the line and ready for full consumption? What is the next steps in getting this across the line for each command and for the func in general?

BlueWinds:
After thinking on it more, I'm going to close this PR and open a new one when I have more progress. I don't think the diff will end up too much larger than it is now.

PR #23001, opened 2022-07-28
fix: call screencastOpts as a function for starting a screencast on electron.

### LINKED ISSUE #17627: Cypress hung up trying to run a spec - won't resolve
### Current behavior

Can't reproduce this locally - but intermittently when running specs via circleci (so far 50% of the time) - one of the specs (never the same one), spins causing circleci to timeout. When I look at the run in the Cypress dashboard it shows all the rest of the specs have passed except for that one and it's just constantly saying it is running - but shows no test results. https://dashboard.cypress.io/projects/zb7814/runs/1686/overview is an example. When I rerun the tests they all pass again so it looks like there's an issue with the cypress runner. This just started happening when we upgraded to 8 - specifically 8.2

The  error  from circleci is: 
```
ERR_TIMED_OUT (-7) loading '<specUrl>'
Error: ERR_TIMED_OUT (-7) loading '<specUrl>
    at rejectAndCleanup (electron/js2c/browser_init.js:217:1457)
    at Object.failListener (electron/js2c/browser_init.js:217:1670)
    at Object.emit (events.js:315:20)
```
    
After saying it was still running for several hours on the Cypress dashboard - it finally stopped and showed this error message: 
"The run started, but never completed. This can happen when the run is canceled from CI or when Cypress crashes during running tests.

Check your CircleCI #557d91f7-1bed-4c21-9f3c-9a4850bc64ad logs for more information. Archive this run to remove it from the runs list and analytics." 

### Desired behavior

Cypress not to get hung up

### Test code to reproduce

Since it fails on a different spec each time, it doesn't seem to be related to actual test code

### Cypress Version

8.2.0

### Other

_No response_

mjhenkes:
- Closes #17627 
(Do we want to close this issue? I think there is more than one hang reported in there but this was the only reproducible example)

### User facing changelog

Fixes a hang when sending video to the dashboard from the electron browser.

### Additional details

This issue was introduced in version 8.6.0 with this change: https://github.com/cypress-io/cypress/pull/18392

`screencastOpts` was converted to a function, but not all instances were called as a function. In the electron browser we passed the function as the options object, which ended up causing a hang when running test.

Based on my testing I'm confident in this fix, but I'm looking for suggestions about how to write a test for it.

I've also added an additional debug log to for future debugging to see if we allow the browser to hang when recording.

### Steps to test

This has been tested via CI. A user was kind enough to help me get a re-creatable example setup for me to test against. Using that I was able to recreate the hang and discover the offending commit. 

The hang was introduced in build 107 (which uses cypress built from this commit: https://github.com/cypress-io/cypress/commit/b4b3d677af8e8df0df67f824cde681858383ec40 ) and corrected in build 109 (which uses cypress build from this commit: https://github.com/cypress-io/cypress/commit/3f7fc207041a9a8a41083420b61d2f75c2727b93). 108 was additional logging. 

https://app.circleci.com/pipelines/github/mjhenkes/jcontent?branch=master

This specific hang will no longer occur in CI.

### PR Tasks


packages/server/lib/modes/record.js
```diff
@@ -784,6 +784,8 @@ const createRunAndRecordSpecs = (options = {}) => {
         })
 
         if (response === responseDidFail) {
+          debug('Response did fail equals response, allowing browser to hang until it is killed: Response %o', { responseDidFail })
```
marktnoonan, line 787: Found this hard to read as a sentence, suggest keeping the variable name all together (I don't know how backticks work in `debug`, but you get the idea):

```suggestion
          debug('`responseDidFail` equals `response`, allowing browser to hang until it is killed: Response %o', { responseDidFail })
```

packages/server/lib/modes/record.js
```diff
@@ -784,6 +784,8 @@ const createRunAndRecordSpecs = (options = {}) => {
         })
 
         if (response === responseDidFail) {
+          debug('`responseDidFail` equals `response`, allowing browser to hang until it is killed: Response %o', { responseDidFail })
```
emilyrohrbough, line 787: why not just kill the process?

mjhenkes: This seems like the safest (easiest) option. I don't have context for why this was setup this way, but now we will have logs when/ if it happens.

emilyrohrbough (approved):

mschile (approved):

chrisbreiding (approved):

htekdev:
Hello @mjhenkes my organization is currently not in the position to upgrade to 10 from 9. What can we do to add a patch for 9? I actually created a fork but not sure what to do next when it comes to patches
 https://github.com/cypress-io/cypress/pull/23330

PR #23004, opened 2022-07-28
chore: remove angular integration

ZachJW34:
### User facing changelog
na

### Additional details
Product wants to push the delay of Angular CT to next release (GTM), so this PR disables all features of Angular CT. Will revert this after our next release.

### Steps to test
You can point Angular at an angular cli application. There won't be any Angular detection. Even with a pre-setup application like `system-tests/projects/angular-cli-configured`, the dev-server will throw with `no matching framework`

na

### PR Tasks

### DIFF
--- a/packages/scaffold-config/src/frameworks.ts
+++ b/packages/scaffold-config/src/frameworks.ts
@@ -217,26 +217,27 @@ export const WIZARD_FRAMEWORKS = [
     supportStatus: 'full',
     componentIndexHtml: componentIndexHtmlGenerator(),
   },
-  {
-    type: 'angular',
-    configFramework: 'angular',
-    category: 'template',
-    name: 'Angular',
-    detectors: [dependencies.WIZARD_DEPENDENCY_ANGULAR_CLI],
-    supportedBundlers: [dependencies.WIZARD_DEPENDENCY_WEBPACK],
-    dependencies: (bundler: WizardBundler['type'], projectPath: string): DependencyToInstall[] => {
-      return [
-        inPkgJson(dependencies.WIZARD_DEPENDENCY_ANGULAR_CLI, projectPath),
-        inPkgJson(dependencies.WIZARD_DEPENDENCY_ANGULAR_DEVKIT_BUILD_ANGULAR, projectPath),
-        inPkgJson(dependencies.WIZARD_DEPENDENCY_ANGULAR_CORE, projectPath),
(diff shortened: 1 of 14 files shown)


packages/scaffold-config/src/frameworks.ts
```diff
@@ -217,26 +217,27 @@ export const WIZARD_FRAMEWORKS = [
...
-    name: 'Angular',
-    detectors: [dependencies.WIZARD_DEPENDENCY_ANGULAR_CLI],
-    supportedBundlers: [dependencies.WIZARD_DEPENDENCY_WEBPACK],
-    dependencies: (bundler: WizardBundler['type'], projectPath: string): DependencyToInstall[] => {
-      return [
-        inPkgJson(dependencies.WIZARD_DEPENDENCY_ANGULAR_CLI, projectPath),
-        inPkgJson(dependencies.WIZARD_DEPENDENCY_ANGULAR_DEVKIT_BUILD_ANGULAR, projectPath),
-        inPkgJson(dependencies.WIZARD_DEPENDENCY_ANGULAR_CORE, projectPath),
-        inPkgJson(dependencies.WIZARD_DEPENDENCY_ANGULAR_COMMON, projectPath),
-        inPkgJson(dependencies.WIZARD_DEPENDENCY_ANGULAR_PLATFORM_BROWSER_DYNAMIC, projectPath),
```
brian-mann, line 233: i know this PR is reverting this code but I noticed that every single `inPkgJson` function is doing an `fs.readJsonSync` which seems unnecessary given you could just read the `package.json` once, assign it to a `const` and then traverse that object to pick out its values

lmiller1990 (commented):
Nice, good catch from @brian-mann as well, we should minimise the amount of times we read from disk where possible

lmiller1990 (approved):

mike-plummer (approved):

AtofStryker (approved):

rockindahizzy (approved):

packages/scaffold-config/src/frameworks.ts
ZachJW34: Created a follow up for addressing the sync calls in this package: https://github.com/cypress-io/cypress/issues/23016

brian-mann: also these should not be sync calls because of the `ulimit` problem... needs to use graceful-fs aka regular `fs.readJson` since at this point we are already watching files and you could have crashes.

Issue #22404, opened 2022-06-20
When you open the TestRunner on a browser such as Chrome, Firefox etc, the favicon displayed is not from the browser itself or even Cypress.


estefafdez:
### Current behavior

With the new Cypress 10 version of the Test Runner, when you open the TestRunner on a browser such as Chrome, Firefox, etc, the favicon displayed is not from the browser itself or even Cypress.

![image](https://user-images.githubusercontent.com/6990209/174639609-4aa39950-c882-4301-87d3-7688d5b9dbe4.png)

### Desired behavior

Add the correct favicon of the browser to know exactly the browser where you are running the tests and add, at least a better title with the specs instead of `test`. 

### Test code to reproduce

Any project run on Chrome or Firefox. 

### Cypress Version

10.1.0

nagash77:
@estefafdez this issue has been resolved.

Issue #22870, opened 2022-07-20
Cypress Studio Reimplementation Spike


pstakoun:
I'd like to facilitate a 2-3 point spike into the lift of reimplementing Cypress Studio into Cypress 10. The ideal outcome here would be a short and simple tech brief outlining the viability of reusing the past implementation as compared to doing it from scratch, some rough estimates (days? weeks? months?) for how much effort this would be at varying levels of parity with Cypress 9. Specifically my question is: can we reuse the Cypress 9 implementation to get it back in, or would it make more sense to rewrite this from scratch? What is the level of effort for either case? Finally, what parts of Studio are higher vs lower lift in terms of engineering complexity and product/design decisions that need to be made (e.g. the flows for creating vs extending a test)?

baus:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/22870?planning-poker) with ZenHub @astone123 @lmiller1990 @marktnoonan @mike-plummer @rockindahizzy @warrensplayer @ZachJW34

ZachJW34:
@pstakoun is there any documentation on why Studio was removed i.e. what were the pros/cons or our previous implementation.

lmiller1990:
I'm going with ~~3~~ 4 points - best part of a week to get something functioning in the new app, and idea of the complexity. This is to revive the existing implementation - most of the code is still hanging around, you just need to know where to look and what to do. This is also with the knowledge I have of the original implementation, I've got a decent idea of how/where to look. 

The main unknowns are:

1. UI. Assuming we keep the identical functionality, we still need the UI to be in line with our new design language.
2. How does it look with Component Testing? Original Studio was E2E only, it's not clear if there's any additional complexity around.
3. Tests - now we've got proper Cypress in Cypress tests, we can likely migrate the old studio tests to those. It'll take some effort, but will give us much better coverage. The original implementation had a lot of stubbing, iirc - I'd like more thorough E2E coverage around a feature as complex as Studio.

I also think we should at least consider some improvements and reflect on the experiment. If we are going to revive it (new implementation or not), it'd sure be nice to at least consider the feedback we got from the experiment - after all, that's what experimental flags are for, soliciting feedback and making improvements prior to GA.

lmiller1990:
[...]

Issue #22328, opened 2022-06-15
Empty `--spec` parameter error message is misleading for Cypress developers


sainthkh:
### Current behavior

When you don't pass anything after `--spec` like `yarn workspace @packages/launchpad cypress:run:e2e --spec`, it shows the error message like below:

```
No version of Cypress is installed in: /home/sainthkh/.cache/Cypress/0.0.0-development/Cypress

Please reinstall Cypress by running: cypress install

----------

Cypress executable not found at: /home/sainthkh/.cache/Cypress/0.0.0-development/Cypress/Cypress

----------

Platform: linux-x64 (Ubuntu - 20.04)
Cypress Version: 0.0.0-development
```

Cypress is actually installed (or cached in CI). The `--spec` is empty. But the message is really misleading. 

### Desired behavior

Maybe, the message should be changed to `No spec files provided. They should be provided after `--spec` argument`. 

### Test code to reproduce

In our project root, `yarn workspace @packages/launchpad cypress:run:e2e --spec`

### Cypress Version

10.1.0

marktnoonan:
What's happening here seems to be a quirk of how we ensure that in the Cypress repo, we run in dev mode. [This script](https://github.com/cypress-io/cypress/blob/4626f7481c9904fec484aa167a02e0197a3095c4/scripts/cypress.js#L14-L17) runs when `@packages/launchpad cypress:run:e2e` is run.

The issue is not with the `--spec` flag as such. In a regular project the message is fine. `npx cypress run --spec` leads to this:

```
error: option '-s, --spec <spec>' argument missing
```

But within the cypress repo, that last flag ends up _receiving `--dev` as the argument_. I didn't look too closely, but something in the timing of when we push it seems to be an issue. 

Swallowing the `--dev` flag means it no longer prevents the attempt to verify the Cypress binary for the version `0.0.0-development`. There's nothing in the cache to verify at that path, since we aren't working with a binary that was installed, we are working with whatever the current code is in the branch being developed. So Cypress complains, correctly, that you don't have anything at that path. The same thing happens if you leave `--browser` at the end, for example, or any flag that needs an argument. 

This definitely could be better, I'll route this to the team to improve at some point so that it works as expected when developing Cypress, just as is does when using a regular build. But at least for users of Cypress in other projects, this all seems to be working as expected.

sainthkh:
@marktnoonan I added "for Cypress Developers" for clarity.

PR #23028, opened 2022-08-01
chore: Intregrate runner packages.

### LINKED ISSUE #21689: Delete old runner code
## **Summary**
Once all the above is done and we are happy, we can remove a lot of old code:

* `packages/runner`
* `packages/runner-ct`
* Almost `packages/ui-components`, except the ones used in the Reporter
* probably `packages/runner-shared`, except a few files which can move to `packages/app`
* `LAUNCHPAD` environment variable - we will be permanently in unified mode now, so delete and associated conditionals with this
* anything that references `runner` and `runner-ct`

## **Acceptance Criteria**
* Should delete old code
* Should not break anything

┆Issue is synchronized with this [Jira Task](https://cypress-io.atlassian.net/browse/UNIFY-898) by [Unito](https://www.unito.io)
┆author: Lachlan Miller
┆epic: Chores
┆friendlyId: UNIFY-898
┆priority: Low
┆sprint: Backlog
┆taskType: Task

sainthkh:
- Related with #21689

### User facing changelog

N/A. It's an internal code cleanup.

### Additional details
- Why was this change necessary? => These are mostly-unused legacy code. 
- Any implementation details to explain? => I migrated `runner-shared` and `runner-ct` into `runner` and removed unused `scss` files and `react` components.

### Steps to test

`packages` folder structure has been changed. Run all tests to make sure that nothing has changed. 

N/A. New contributors might not be confused with the legacy codes.

### PR Tasks


packages/server/lib/controllers/runner.ts
```diff
@@ -92,8 +92,7 @@ export const runner = {
   },
 
   handle (testingType, req: Request, res: Response) {
-    // FIXME: Why is runner-ct being used for all injections? Can anyone comment here as to why this changed?
-    const pathToFile = getPathToDist(testingType === 'e2e' ? 'runner' : 'runner-ct', req.params[0])
+    const pathToFile = getPathToDist('runner', req.params[0])
```
lmiller1990, line 95: Nice, I wonder if we can just serve this from `runner-shared` instead? Ideally we should eventually eliminate both `runner` and `runner-ct` entirely.

In this case you'd probably need to change the webpack config to write the `dist` to `runner-shared`, or even move it there entirely. Thoughts?

.github/CODEOWNERS
```diff
@@ -40,8 +40,6 @@
 /packages/rewriter/  @cypress-io/end-to-end
 /packages/root/  @cypress-io/end-to-end
 /packages/runner/  @cypress-io/end-to-end
-/packages/runner-ct/  @cypress-io/component-testing
-/packages/runner-shared/  @cypress-io/end-to-end
 /packages/server/  @cypress-io/end-to-end
 /packages/socket/  @cypress-io/end-to-end
 /packages/static/  @cypress-io/end-to-end
```
lmiller1990, line 45: FYI some context - we ideally should delete ALL of runner/runner-shared, and bundle the entire app with the Vite stack.

The reason we are stuck like this is explained a little here: https://github.com/cypress-io/cypress/pull/20663#issuecomment-1073532399

Basically some of the code in driver and possibly reporter are not really ESM compatible - webpack does something different to Vite to resolve the modules, which is why it works in webpack, but we cannot import to Vite.

Not sure if this is something you are looking to investigate more @sainthkh but if you do please let me know, I'd like to follow along/help out - I'd really like to get away from this "inject webpack bundle" mess at some point, but it's quite hard. Likely making driver 100% ESM (including dependencies) would be a good first step.

lmiller1990 (commented):
Left a few comments/questions, happy to approve once they've been replied too - nice job on this, great to see some of this getting cleaned up!

.gitignore
```diff
@@ -78,8 +78,8 @@ system-tests/lib/fixtureDirs.ts
 # from npm/webpack-dev-server
 /npm/webpack-dev-server/cypress/videos
 
-# from runner-ct
-/packages/runner-ct/cypress/screenshots
+# from runner
+/packages/runner/cypress/screenshots
```
emilyrohrbough, line 82: delete. this is already included on line 40

emilyrohrbough (commented):
We think alike! I love it. I started cleaning this up [last week](https://github.com/cypress-io/cypress/compare/clean-up-runner-pck) while waiting on test results because it was bugging me! I'll drop a few comments on things I noticed while cleaning this up.

packages/runner/package.json
```diff
@@ -24,6 +24,7 @@
     "@packages/rewriter": "0.0.0-development",
```
emilyrohrbough, line 24: Need to add in the test script to run the mocha tests  that were in runner-shared

packages/server/lib/controllers/runner.ts
emilyrohrbough: I think we need to move towards removing the runner dirs all together. Ideally the app package would handle setting this up, even if both vite & webpack need to be used to bundle everything correctly. 🤷🏻‍♀️

packages/runner/package.json
```diff
@@ -36,6 +37,8 @@
     "prop-types": "15.7.2",
     "react": "16.8.6",
     "react-dom": "16.8.6",
+    "react-popper": "2.2.5",
+    "react-shadow-dom-retarget-events": "1.0.11",
     "rimraf": "3.0.2",
     "webpack": "^4.44.2",
```
emilyrohrbough, line 43: Can you update the runner's README as well?

It's doing a few more things than it currently lists with this merge:
- Contains `dom.js`, which uses proprietary webpack loaders and cannot easily be imported with Vite (dev server in `@packages/app`). Once `dom.js` is free of webpack-specific loader code, we should move it to `@packages/app`.
- Contains UI code for Cypress Studio, which was marked as experimental in Cypress 9.x and won't be part of Cypress 10.x initially. It will return at a later date. Until then, the code will be here. It's not currently used in the app.
- contains Legacy Cypress styles, most of these can likely be removed.

out of date:
- Some existing tests in `cypress/e2e` should be migrated to `@packages/app/cypress/e2e/runner`.

packages/runner-ct/webpack.config.ts
```diff
@@ -1,101 +0,0 @@
-process.env.NO_LIVERELOAD = '1'
-
-import _ from 'lodash'
-import path from 'path'
-import type webpack from 'webpack'
-import { getCommonConfig, getCopyWebpackPlugin } from '@packages/web-config/webpack.config.base'
-import * as cyIcons from '@packages/icons'
```
emilyrohrbough, line 7: We should add this to the webpack configuration in runner/webpack. I think it actually fixes #22404

packages/runner-ct/webpack.config.ts
```diff
@@ -1,101 +0,0 @@
...
-      },
-    ],
-  },
-  entry: {
-    cypress_runner: [path.resolve(__dirname, 'src/index.js')],
-  },
-  output: {
-    path: path.resolve(__dirname, 'dist'),
-    filename: '[name].js',
-    chunkFilename: '[name].[chunkhash].js',
```
emilyrohrbough, line 75: I want to call out this is the key / largest difference in webpack configuration difference between this and runner/webpack.config. I'll defer to the CT team if this is something that matters. if it does, I'd assume maybe here: https://github.com/cypress-io/cypress/pull/23028/files#diff-9af0f8f4f669446066e3dca23bb2645cb4f94dc9b11179a63f38cc657b36b30eR51

though assuming if tests are passing we are okay? cc// @lmiller1990 -- tagging you since you've already reviewed 😄

packages/server/lib/controllers/runner.ts
lmiller1990: Good point, there isn't really a compelling reason for `runner*` to exist, we may as well just move all the things to `app` (including webpack deps, config, etc). More packages just adds more misdirection.

.github/CODEOWNERS
sainthkh: It seems that we have 4 things to do to delete the `runner` pacakge.

- [ ] Migrate `dom`. -> I think `driver/src/dom` might be a good place or move it to `app` directly.
- [ ] Reimplement Studio -> #22870 
- [ ] ESM-ify `driver` -> I agree it's the biggest problem. 
- [ ] Let `reporter` compile its own styles. -> Move `runner/webpack.config.ts` to `reporter`.

packages/server/lib/controllers/runner.ts
sainthkh: When I read the current code, 

* `runner-ct` was dependent on `runner-shared` to use `dom` and `stduio`. 
* `runner` was dependent on `runner-ct` to use `unified-runner`.

So, the workflow became:

1. Migrate the `runner-shared` to `runner-ct`. (Remove unnecessary studio react files and some scss)
2. Migrate the `runner-ct` to `runner`. (Remove unnecessary scss files)

Now all the necessary codes of `runner-ct` and `runner-shared` are in `runner`.

.gitignore
sainthkh: Done.

packages/runner-ct/webpack.config.ts
lmiller1990: Yeah, I don't think we need this. We had some weird situation going on where we would conditionally load the react devtools, which we tried to embed in the runner-ct (which would be cool, if it worked as expected) but that experiment is long gone.

It sure would be nice to have the devtools embedded in the runner automatically... the main problem is the react devtools are designed to be used for 1 app and not torn down unless you refresh the page or via hot reload. When you run 100s of CT tests, we continually tear down/reset in AUT (and effectively the React app) every couple of seconds. Doing this leads to some situations where the devtools show stale components, or worse, keep a reference to components that no longer exist ... which can give us a memory leak.

packages/runner/package.json
sainthkh: Done. I changed the `UI code for Cypress Studio` with `Cypress Studio Recorder code`. Because I removed `react` UI code in this PR and `Studio Recorder` is currently injected into `app`.

.github/CODEOWNERS
lmiller1990: Dom -> are you going do to this here?
Studio -> I am looking into this now
ESMify driver -> agree, hard
Reporter styles -> do this separately I think?

lmiller1990 (approved):
Nice cleanup! Going to approve, but also happy to re-review if you move the DOM code to `app`.

Nice list of TODOs here: https://github.com/cypress-io/cypress/pull/23028#discussion_r936142417

.github/CODEOWNERS
sainthkh: I read the `dom` code and it seems that it's better to be a separate PR.

It currently renders highlight by

1. Creating the container node with jQuery.
2. Rendering the Highlight DOM with React.

I can simply migrate `dom` to `app` and use `unified-runner`. But I don't think it's not our goal. It's just hiding and postponing. 

To get migrating `dom` done, I need to do these things:

- [ ] Remove JQuery from `dom` and `dimensions`.
- [ ] Move the current JQuery version of `dimensions` to `driver/src/dom` because it's used in `driver/src/dom/blackout`.
- [ ] Migrate `Highlight` from `react` to `vue` with our `Tooltip` component. 
- [ ] Migrate `dom` functions to `app/runner/aut-iframe`. Leave `studio`-related functions.
- [ ] Clean up unified-runner `dom` and `CypressJQuery`. They're only used inside `aut-iframe`. I guess they can be removed safely from `unified-runner`.

I think it's a big change and outside the scope of this PR. 

----

As for reporter styles, it's safe to do it after ESM-ifying `driver`. We're trying to remove `webpack`. I don't think we don't need to create one more `webpack.config.ts` now.

packages/runner/package.json
sainthkh: Currently, `runner-shared` unit tests are ignored. 

https://github.com/cypress-io/cypress/blob/f69afb5aa49996cace863f8a3ae142698deff982/package.json#L53

And it fails in the current `develop` branch.

And the only the file that exists is `studio-recorder.spec.js`. It's there for reference purpose and it'll be removed soon with #22870.

----

I checked the code base and removed `network`, `socket`, and `rewriter`.

packages/runner-ct/webpack.config.ts
sainthkh: I fixed this and learned that favicon shows when the URL port is `4455`, but it fails when `5555`.

.github/CODEOWNERS
sainthkh: I opened the new PR #23187 to migrate `dom.js`. It's currently empty because I'm now reading and planning.

lmiller1990:
Are you still working on this @sainthkh or are we good to re-review and merge if green?

sainthkh:
@lmiller1990  I think it's done. I'll take care of the `dom` issue at #23187.

emilyrohrbough (approved):

lmiller1990:
run-reporter seems broken. Last blocker @sainthkh :thinking:

sainthkh:
@lmiller1990  It's happening because there are 4 test files inside `reporter`(`commands/command.cy.tsx`, `hooks/hooks.cy.tsx`, `lib/tag.cy.tsx`, `test/test.cy.tsx`), but there are 7 machines to run them. 3 of them don't have work to do. I opened an issue about this misleading message a few months ago at #22328.

I intentionally didn't fix this problem at #22326 because I thought the problem is more at the ci setting side [(comment)](https://github.com/cypress-io/cypress/pull/22326#issuecomment-1157295392).

There are a few solutions for this problem:

1. Change the CI setting from 7 to 2 or 3. -> I think it's better because we need less power and it's unlikely that we might need more than 2 or 3 machines for the tests.
2. Return nothing when there is no test file. -> We need to fix the ci script. Maybe it's a different issue and open a new PR.

lmiller1990:
Let's just take it down to `parallelism: 2` for now. For such a small number of specs, we probably don't need it at all...

PR #23034, opened 2022-08-01
chore: simplify error/warning management in data context

lmiller1990:
- resolves: https://github.com/cypress-io/cypress/issues/22030

### User facing changelog

N/A

### Additional details

As per [this todo](https://github.com/cypress-io/cypress/blob/a7b140d43894e1e90f5a29cf907abaf23e79dba9/packages/data-context/src/data/coreDataShape.ts#L151), it's much more simple if we collapse all the warnings/errors into a single place, instead of having many different properties.

The original issue was somewhat obsolete, so I repurposed it around the same generic topic of "improving the data-context". I wrote a comment to this meaning [in this issue](https://github.com/cypress-io/cypress/issues/22030#issuecomment-1200675962).

### Steps to test

No user facing change - just simplify things by consolidating error/warning into a single `diagnostics` property.

### PR Tasks


packages/data-context/src/DataContext.ts
```diff
@@ -386,14 +377,8 @@ export class DataContext {
       }
 
       this.update((d) => {
-        if (d.currentProjectData?.testingTypeData?.activeAppData) {
```
lmiller1990, line 389: This is so overly complex - it's not really obvious if there's supposed to be some specific pattern here, but I collapsed everything into a single `diagnostics` property with `error: ErrorWrapper | null, warnings: ErrorWrapper[] }` and everything seems to work just fine, so I guess it's okay? 

I think the original idea was to have different ways to handle/manage errors and warnings depending on where you are in the lifecycle, eg:

- `activeAppData` - app is open
- `testingTypeData` - launchpad, after choosing a testing type
- `currentProjectData` - launchpad, a project is chosen but no testing type selected

But there isn't any obvious value in doing this, from what I can see, other than more overhead and another undocumented pattern you need to remember. For that reason I'm thinking we just delete it, and reintroduce if/when it makes sense to do so.

packages/data-context/src/actions/ErrorActions.ts
```diff
@@ -33,34 +21,34 @@ export class ErrorActions {
    */
   clearWarning (id: string) {
     this.ctx.update((d) => {
-      const warningsIndex = d.warnings.findIndex((v) => v.id === id)
+      const warningsIndex = d.diagnostics.warnings.findIndex((v) => v.id === id)
```
emilyrohrbough, line 24: Looks like this warningIndex value will be the same for `projectWarningsIndex`, `testingTypeWarningsIndex` and `appWarningsIndex` with how this was simplified. Can we clean up some of this code or am I miss-understanding?

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -516,8 +518,8 @@ export class ProjectLifecycleManager {
       d.currentTestingType = testingType
       d.wizard.chosenBundler = null
       d.wizard.chosenFramework = null
-      if (d.currentProjectData) {
-        d.currentProjectData.testingTypeData = makeTestingTypeData(testingType)
+      if (testingType) {
```
emilyrohrbough, line 521: By resetting this error, are we potentially swallowing errors from global state and/or the initial project selected view int eh launchpad? Or are these already emitted? Not sure if we have any existing tests around the error propagation?

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -495,8 +494,11 @@ export class ProjectLifecycleManager {
...
```
emilyrohrbough, line 497: same thought as below, should we be resetting the diagnostics state here? maybe warnings for the current testing type

might be worth leaving to [setCurrentProject](https://github.com/cypress-io/cypress/pull/23034/files#diff-63757aa782fe9ed7c93e5f91cd7afdcf8f617d2b4eea12d1f51d09747fb28247R410) and [clearCurrentProject](https://github.com/cypress-io/cypress/pull/23034/files#diff-d36f6626bf805d1bbbfba4786ca3a518f0083613a9814fd037c5592808949606R84)?

packages/data-context/src/DataContext.ts
emilyrohrbough: +1 love this clean up!

packages/data-context/src/actions/ErrorActions.ts
lmiller1990: You are right, I just mindlessly refactored. This becomes

```ts
clearWarning (id: string) {
    this.ctx.update((d) => {
      const warningsIndex = d.diagnostics.warnings.findIndex((v) => v.id === id)

      if (warningsIndex !== -1) {
        d.diagnostics.warnings.splice(warningsIndex, 1)
      }
    })
  }
```

packages/data-context/src/data/ProjectLifecycleManager.ts
lmiller1990: Tests for this area:
- warnings: https://github.com/cypress-io/cypress/blob/develop/packages/launchpad/cypress/e2e/config-warning.cy.ts
errors: https://github.com/cypress-io/cypress/blob/develop/packages/launchpad/cypress/e2e/error-handling.cy.ts
I tried deleting this entire part and **not** clearing the error, and actually, everything still passes. Maybe we just remove this entirely.

If there is ever an **error** we have a pattern - show this overlay:

image

The only way out is "Try again" after you fix the error, or to close Cypress. Whether this is the ideal behavior is up for debate, but it suggests that it's fine to clobber this error.

The only time we seem to show errors is things relating to project config. I can't think of any situations where we show an error in **global** mode (eg, where you haven't chosen a project). I tried to create one, but no luck. Open mode tests are here: https://github.com/cypress-io/cypress/blob/develop/packages/launchpad/cypress/e2e/open-mode.cy.ts (none for errors, perhaps an error state in global mode is just not a possible state?)

packages/data-context/src/actions/ErrorActions.ts
lmiller1990: done

packages/data-context/src/data/ProjectLifecycleManager.ts
lmiller1990: I tried deleting this but it breaks a few things. Test case that fails is

1. open mode
2. choose project
3. config errors
4. click the "projects" breadcrumb to go back to global mode

If we do not clear the error here, you cannot get back to the projects page (error state overlay prevents you from seeing your projects).

lmiller1990:
Replied to comments and fixed requested changes @emilyrohrbough :pray:

emilyrohrbough (approved):

emilyrohrbough:
@lmiller1990 looks like we missed the second approval here...

I'm seeing unexpected warning and it's posting multiple times. 
Screen Shot 2022-08-09 at 12 06 19 PM
g

lmiller1990:
Uh oh, I will add a regression test and figure out what's going on.

lmiller1990:
@emilyrohrbough I think this is actually a **production** bug and not introduced here - another user has reported this (in 10.4, which does not have this patch): https://github.com/cypress-io/cypress/issues/23216#issuecomment-1213493767

emilyrohrbough:
@lmiller1990 nice find!

Issue #23039, opened 2022-08-01
[driver] errors thrown from test:before:run are incorrectly handled


emilyrohrbough:
### Current behavior

Errors thrown from `test:before:run` handlers are not propagated correctly and cause inconsistent behaviors in run vs open mode.

- in open mode: the error is reported for each test
- in run mode: the error is swallowed, no tests start and Cypress hangs indefinitely 

This failure occurs [here](https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cypress/runner.ts#L1548) in which the runner will catch the error [here](https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cypress/runner.ts#L1564), however since the [runnable has not been set ](https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cypress/runner.ts#L1559) the runnable (hook or test) technically hasn't officially starts which I think it causing the inconsistency with the finally handler which is in place to try to log the test results/errors and ensure the `after:test:run` and `after:test:run:async` events are emitted.

### Desired behavior

When errors thrown from `test:before:run` handlers, reports it correctly and consistently between run/open mode without hangs.

### Test code to reproduce

https://github.com/cypress-io/cypress-test-tiny/compare/verify-unhandled-err

CI job which timed out: https://ci.appveyor.com/project/cypress-io/cypress-test-tiny/builds/44331388

### Cypress Version

10.3.1

Issue #2573, opened 2018-10-08
cy.clearLocalStorage should clear all localStorage of all domains


pieterdd:
### Summary of this thread so far
`cy.clearLocalStorage` doesn't behave the way I assumed it would. It doesn't clear _all_ local storage. The docs have since been updated to reflect this.

---

### Current behavior:

I was trying to write a simple test for some to do list website when I noticed that certain state wasn't being cleared correctly between tests. So I added some sanity checks like `cy.clearCookies()` and `cy.clearLocalStorage()` - which in theory should be redundant - and then I observed this: https://youtu.be/x6Yeh6WM5fM

Oddly enough, it looks like local state is magically reappearing for reasons I don't understand. Any idea what's going on? Here's the code snippet I executed:

```javascript
describe('Hello world', () => {
    before(function () {
        cy.viewport(1200, 660)
        cy.clearCookies()
        cy.clearLocalStorage()
        cy.wait(20000)
        cy.visit('https://www.any.do/').contains('Login').click()
        cy.contains('ALREADY A MEMBER?', {
            'timeout': 10000
        }).click()
        cy.get('#emailSignin #email').type('spam@pieterdedecker.be')
        cy.get('#emailSignin #password').type('badpassword')
        cy.contains('SIGN IN').click()
    })

    it('Is a test', () => {
        cy.log('Look!')
    })
})
```

Even going into the developer tools, Application tab, clicking 'Clear storage' and then clearing everything doesn't reset it. The only workaround I have found is selecting the local storage data for https://web.any.do and clicking 'Clear All'.

### Desired behavior:

The old local storage data should not reappear when the test reaches the domain https://web.any.do.

### Steps to reproduce:

1. Run this script with valid credentials. I've changed my example account's password, but you can use anydo@mailinator.com with password 'anydone'. The script should work even without the `cy.wait`; that was just to give me time to inspect cookie/local storage beforehand.
2. The test will pass.
3. Rerun the test.
4. The test will fail because it suddenly remembers the login.

### Versions

Cypress v3.1.0, Electron 59

pieterdd:
Interesting... if this is correct it'd definitely be good to add to the documentation. But ideally I'd like _all_ local storage to be cleared by this command (or a similar one).

pieterdd:
Thank you for your analysis!

[...]

PR #23040, opened 2022-08-01
feat: add testIsolation option and support describe-only overrides

### LINKED ISSUE #22279: Add new testIsolation configuration option
### What would you like?

What I would suggest to be improved is when we are using the `cy.session()` command or to be more specific `experimentalSessionAndOrigin=true`, currently the state between the tests within one cy/spec file is cleaned and we need to reset everything, but when we do not use this flag the state is kept and we can link together some tests. 
It would be great to be able to keep the state as we do without the flag and not set the page url to `_blank_`

### Why is this needed?

This would save a lot of execution time for tests where we do need to stay on the same page, or depend on some setup from the previous test where instead of logging in, setting everything up again, we can just continue where we left off (like we do when `experimentalSessionAndOrigin` is false)

### Other

_No response_

emilyrohrbough:
- Closes #22279

- add testIsolation option - when `experimentalSessionAndOrigin=false`, the default value is `legacy`. When `experimentalSessionAndOrigin=true`, this the default value is `strict`.
   - legacy - this is the current test isolation behavior in cypress - clock, viewport, cookies, local storage, etc continue to be reset before each tests
   - strict - all rests from legacy plus clearing the page between each test by visiting `about:blank`. This ensures app state is reset/clean for each test.
- add support for config override levels - this allows each option to determine what override levels are supported. The override levels are:
  - `code` - when either the support file or spec file is loaded and config is altered with `Cypress.config()`
  - `test:before:run` - config altered during event handler with `Cypress.config()`
  - `test:before:run:async` - config altered during event handler with `Cypress.config()`
  - `suite` - configuration provided as suite override  `describe('...', { ...configOverides }, ()=>{})`
  - `test`- configuration provided as test override  `it('...', { ...configOverides }, ()=>{})`
  - `runtime` - configuration altered from runnable (hook or test execution) using Cypress.config()
  - `never` - configuration is read-only and cannot be altered without restarting Cypress

### User facing changelog

TO DO

### Additional details
- discovered `test:before:run` does not handle thrown errors correctly (#23039) so config validation is currently skipped for this hook.

- Improved error messaging
- Ability to set the `testIsolation` intensity when `experiementalSessionAndOrigin=true`. this preps for when this goes GA and users don't want to update tests.

### PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4634


packages/config/src/options.ts
```diff
@@ -181,260 +178,245 @@ const resolvedOptions: Array<ResolvedConfigOption> = [
...
-    canUpdateDuringTestTime: true,
+    overrideLevels: ALL_OVERRIDE_LEVELS,
+  }, {
+    name: 'testIsolation',
+    // TODO: When experimentalSessionAndOrigin is removed and released as GA,
+    // update the defaultValue from 'default' to 'strict' and
+    // update this code to remove the check/override specific to enable
+    // strict by default when experimentalSessionAndOrigin=true
+    defaultValue: 'default',
+    validation: validate.isOneOf('default', 'strict'),
```
emilyrohrbough, line 363: I went 'default', however writing this change summary, wondering if I should have gone 'standard'?

packages/config/src/options.ts
```diff
@@ -181,260 +178,245 @@ const resolvedOptions: Array<ResolvedConfigOption> = [
...
     defaultValue: 'cypress/fixtures',
     validation: validate.isStringOrFalse,
     isFolder: true,
-    canUpdateDuringTestTime: false,
     requireRestartOnChange: 'server',
   }, {
     name: 'excludeSpecPattern',
     defaultValue: (options: Record<string, any> = {}) => options.testingType === 'component' ? ['**/__snapshots__/*', '**/__image_snapshots__/*'] : '*.hot-update.js',
     validation: validate.isStringOrArrayOfStrings,
-    canUpdateDuringTestTime: true,
```
emilyrohrbough, line 246: looking at this....why can this update at test time?

packages/config/src/options.ts
```diff
@@ -181,260 +178,245 @@ const resolvedOptions: Array<ResolvedConfigOption> = [
...
   }, {
     name: 'nodeVersion',
     validation: validate.isOneOf('bundled', 'system'),
-    canUpdateDuringTestTime: false,
   }, {
     name: 'numTestsKeptInMemory',
     defaultValue: 50,
     validation: validate.isNumber,
-    canUpdateDuringTestTime: true,
+    overrideLevels: ALL_OVERRIDE_LEVELS,
```
emilyrohrbough, line 258: Same with this opt...this shouldn't be able to update. 😕

packages/config/src/options.ts
```diff
@@ -181,260 +178,245 @@ const resolvedOptions: Array<ResolvedConfigOption> = [
...
     defaultValue: 60000,
     validation: validate.isNumber,
-    canUpdateDuringTestTime: true,
+    overrideLevels: ALL_OVERRIDE_LEVELS,
   }, {
     name: 'port',
     defaultValue: null,
     validation: validate.isNumber,
-    canUpdateDuringTestTime: true,
+    overrideLevels: ALL_OVERRIDE_LEVELS,
```
emilyrohrbough, line 272: Same with this opt...this shouldn't be able to update. 😕

packages/config/src/options.ts
```diff
@@ -181,260 +178,245 @@ const resolvedOptions: Array<ResolvedConfigOption> = [
...
   }, {
     name: 'projectId',
     defaultValue: null,
     validation: validate.isString,
-    canUpdateDuringTestTime: true,
+    overrideLevels: ALL_OVERRIDE_LEVELS,
```
emilyrohrbough, line 277: and here??

packages/config/src/options.ts
lmiller1990: I don't think it should be able to - probably incorrect.

cli/types/cypress.d.ts
```diff
@@ -2784,6 +2784,13 @@ declare namespace Cypress {
      * @default "cypress/support/{e2e|component}.js"
      */
     supportFile: string | false
+    /**
+     * The test isolation level applied to ensure a clean slate between tests.
+     *   - default - resets/clears aliases, intercepts, clock, viewport, cookies, and local storage before each test.
+     *   - strict - all resets everything from default, plus clears the page by visiting 'about:blank' to ensure clean app state before each test.
+     * @default "default", however, when experimentalSessionAndOrigin=true, the default is "string"
```
lmiller1990, line 2791:
```suggestion
     * @default "default", however, when experimentalSessionAndOrigin=true, the default is "strict"
```

packages/config/src/browser.ts
```diff
@@ -21,10 +29,12 @@ const debug = Debug('cypress:config:browser')
 const dashesOrUnderscoresRe = /^(_-)+/
 
 // takes an array and creates an index object of [keyKey]: [valueKey]
-function createIndex<T extends Record<string, any>> (arr: Array<T>, keyKey: keyof T, valueKey: keyof T) {
+function createIndex<T extends Record<string, any>> (arr: Array<T>, keyKey: keyof T, valueKey: keyof T, defaultValueFallback?: any) {
   return _.reduce(arr, (memo: Record<string, any>, item) => {
     if (item[valueKey] !== undefined) {
       memo[item[keyKey] as string] = item[valueKey]
+    } else {
+      memo[item[keyKey] as string] = defaultValueFallback
```
lmiller1990, line 37: Seems you don't need the `as string` cast (just tried it, editor seems happy).

packages/server/lib/config.ts
```diff
@@ -215,6 +214,19 @@ export function mergeDefaults (
     throw makeConfigError(errors.get(err, ...args))
   }, testingType)
 
+  // TODO: testIsolation should equal 'strict' by default when experimentalSessionAndOrigin=true
```
lmiller1990, line 217: Will we eventually apply `strict` to everything? My understanding is we have `strict` for `cy.origin` to make it more reliable (clean slate). Since this seems ideal for all tests, is the intention to make this the default in Cypress 11?

system-tests/test/testConfigOverrides_spec.ts
```diff
@@ -33,6 +33,23 @@ describe('testConfigOverrides', () => {
     },
   })
 
+  // FIXME: bug in runner causes browser to hang in run mode when test:before:run throws an exception
```
lmiller1990, line 36: Nice, the tests throughout this PR look very thorough :100:

packages/config/test/index.spec.ts
```diff
@@ -4,6 +4,7 @@ import sinon from 'sinon'
 import sinonChai from 'sinon-chai'
 
 import * as configUtil from '../src/index'
+import { OverrideLevel, ALL_OVERRIDE_LEVELS } from '../src/options'
```
lmiller1990, line 7: Here is a neat TS trick you could add to `src/options.ts` to reduce some duplication: https://github.com/cypress-io/cypress/compare/test-isolation-config-opt-2...lmiller/test-isolation-config-opt-2-ts-update?expand=1

The trick is

```ts
const abc = ['a', 'b', 'c'] as const

type ABC = typeof abc[number] // 'a' | 'b' | 'c'
```

By casting the array `as const`, the array becomes immutable, so TS can infer that only `'a' | 'b' | 'c'` is valid. If you remove `const`, it'll become `Array<string>` instead of `Array<('a' | 'b' | 'c')>`.

lmiller1990 (commented):
Left some misc comments/questions.

packages/config/test/index.spec.ts
emilyrohrbough: I was trying to do something like this, but I'm def still a newbie at typescript. With your suggestion, I'm seeing the following: 

Screen Shot 2022-08-02 at 7 28 23 PM
Screen Shot 2022-08-02 at 7 28 04 PM


I don't think did it quite right 😅

emilyrohrbough: ah wow. re-reading, you literally provided the code....thought `typeof abc[number]` was a typo where number was meant to be string. I need to pay closer attentin.

lmiller1990: Odd, seems fine for me? https://app.circleci.com/pipelines/github/cypress-io/cypress/41239/workflows/62972cbb-f8c1-4486-baaf-27ea5c645ef5/jobs/1706374?invite=true#step-114-7

I think you missed `Readonly`

```ts
export type OverrideLevels = Readonly<Array<typeof ALL_OVERRIDE_LEVELS[number]> | 'never'>
                                               ^^^^^^^^^ this
```

packages/config/src/browser.ts
emilyrohrbough: [cypress-io/cypress@`191105b` (#23040)](https://github.com/cypress-io/cypress/pull/23040/commits/191105ba94ac63d4e1ed69e02bc83d3beab9ec12)

packages/config/test/index.spec.ts
emilyrohrbough: https://github.com/cypress-io/cypress/pull/23040/commits/191105ba94ac63d4e1ed69e02bc83d3beab9ec12

lmiller1990 (approved):
LGTM once CI is green (will leave TS to you, would be nice to get these more type safe, maybe just merge in my branch that has everything :heavy_check_mark: ?)

packages/config/src/options.ts
chrisbreiding: I agree the value name shouldn't be 'default', especially since once session/origin goes GA, it will no longer be the default option. I think we should name it 'legacy'. That's more-or-less what it is, since it's what Cypress historically has supported and 'strict' is the way forward. It's also a wink-wink, nudge-nudge that setting it to 'legacy' is not the ideal way to go.

packages/config/src/options.ts
```diff
@@ -27,32 +29,35 @@ export type BreakingOptionErrorKey =
...
   name: string
   defaultValue?: any
   validation: Function
-  isFolder?: boolean
-  isExperimental?: boolean
+  requireRestartOnChange?: 'server' | 'browser'
   /**
-   * Can be mutated with Cypress.config() or test-specific configuration overrides
+   * The list of test-time overrides levels supported by the configuration option. When undefined,
+   * it indicates the configuration value cannot be override via suite-/test-specific
```
chrisbreiding, line 46:
```suggestion
   * it indicates the configuration value cannot be overridden via suite-/test-specific
```

packages/config/src/options.ts
chrisbreiding: Maybe there's not a great reason to change this during a test, but it's still technically possible, right? I think that's the main consideration, whether you can instead of whether you should.

packages/server/lib/config.ts
chrisbreiding: `cy.origin` is tied to and dependent on the sessions feature because it enables clearing cookies/localstorage/etc for all origins, whereas historically it's only been cleared for the primary origin. Test isolation is a separate feature that, as you intuited, should benefit all e2e tests regardless of whether or not they use `cy.origin`.

system-tests/test/testConfigOverrides_spec.ts
chrisbreiding: Is there an issue for following up on this?

chrisbreiding (changes requested):

packages/driver/src/cypress/cy.ts
```diff
@@ -659,9 +658,8 @@ export class $Cy extends EventEmitter2 implements ITimeouts, IStability, IAssert
       this.queue.reset()
       this.queue.clear()
       this.resetTimer()
-      this.testConfigOverride.restoreAndSetTestConfigOverrides(test, this.Cypress.config, this.Cypress.env)
-
       this.removeAllListeners()
+      this.testConfigOverride.restoreAndSetTestConfigOverrides(test, this.Cypress.config, this.Cypress.env)
```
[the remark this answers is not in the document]
emilyrohrbough: If `restoreAndSetTestConfigOverrides` throws, the previous test's event listeners are removed -- though I'm not sure it matters. The next test should handle cleaning this up.

packages/driver/src/cypress/cy.ts
```diff
@@ -942,7 +940,6 @@ export class $Cy extends EventEmitter2 implements ITimeouts, IStability, IAssert
     this.state('promise', undefined)
     this.state('hookId', hookId)
     this.state('runnable', runnable)
-    this.state('test', $utils.getTestFromRunnable(runnable))
```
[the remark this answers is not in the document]
emilyrohrbough: I didn't remove setting it, I just moved when we set it. It is set in the reset() handler vs the setRunnable() handler so this information is available when the suite-/test-level overrides are applied.

packages/config/src/options.ts
```diff
@@ -27,32 +29,35 @@ export type BreakingOptionErrorKey =
   | 'RENAMED_CONFIG_OPTION'
   | 'TEST_FILES_RENAMED'
 
-type TestingType = 'e2e' | 'component'
+// The test-time override levels
+export const ALL_OVERRIDE_LEVELS = ['code', 'test:before:run', 'test:before:run:async', 'suite', 'test', 'runtime'] as const
```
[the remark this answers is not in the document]
emilyrohrbough: Let me think about this....they do fire at different times but are conceptually the same in regards to the override level. This is being used to accurately report in the error message but I can likely swizzle things to remove the over-specificity.

system-tests/test/testConfigOverrides_spec.ts
lmiller1990: I think this one https://github.com/cypress-io/cypress/issues/23039

AtofStryker (commented):
still going through the bulk of it but figure I will leave what I have until I can continue to look through it!

packages/driver/src/cypress.ts
```diff
@@ -220,21 +220,57 @@ class $Cypress {
     this.state = $SetterGetter.create({}) as unknown as StateFunc
     this.originalConfig = _.cloneDeep(config)
     this.config = $SetterGetter.create(config, (config) => {
-      if (this.isCrossOriginSpecBridge ? !window.__cySkipValidateConfig : !window.top!.__cySkipValidateConfig) {
-        validateNoReadOnlyConfig(config, (errProperty) => {
-          const errPath = this.state('runnable')
-            ? 'config.invalid_cypress_config_override'
-            : 'config.invalid_test_config_override'
+      let overrideLevel
```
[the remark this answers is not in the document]
emilyrohrbough: done: [cypress-io/cypress@`9848dab` (#23040)](https://github.com/cypress-io/cypress/pull/23040/commits/9848dabdbe59a5c3544aa1522e2dd8d6af7bce1c) 

additional test coverage has been added as well.

packages/config/src/options.ts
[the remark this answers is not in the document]
emilyrohrbough: [slimmed down to `event`](https://github.com/cypress-io/cypress/pull/23040/commits/88aa9e59abf776c416e5f1b1bf910ce7f657fa68#diff-63812dbb23c9353d6d1e500239c7019912de3c33fe540d2cb58768b0b858359fL33)

packages/driver/src/cypress.ts
```diff
@@ -220,21 +220,57 @@ class $Cypress {
...
+          overrideLevel = test._testConfig.applied // either suite or test
+        }
+      } else {
+        overrideLevel = 'code'
+      }
+
+      // FIXME: // bug in runner causes browser to hang in run mode when test:before:run throws an exception
+      if (overrideLevel === 'test:before:run' && this.config('isTextTerminal')) {
+        return
+      }
```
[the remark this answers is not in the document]
emilyrohrbough: good call. updated how config is set such that is skips setting it at this level: [`9848dab` (#23040)](https://github.com/cypress-io/cypress/pull/23040/commits/9848dabdbe59a5c3544aa1522e2dd8d6af7bce1c#diff-8da84c12bd9e64b84f6a0eb1db78dee3209f752e6e2e77a83d64c6a2ffca4fcaR36)

emilyrohrbough:
@chrisbreiding Added in links to the TODO comments here: https://github.com/cypress-io/cypress/pull/23040/files#diff-47d1dcf44e2076653a4be7d397029774e2b76e083b93d4deeed010a280f62af6R181

packages/driver/cypress/e2e/commands/local_storage.cy.js
```diff
@@ -42,8 +42,11 @@ describe('src/cy/commands/local_storage', () => {
         const clear = cy.spy(Cypress.LocalStorage, 'clear')
 
         Cypress.emit('test:before:run', {})
-
-        expect(clear).to.be.calledWith([])
+        if (Cypress.config('experimentalSessionAndOrigin')) {
```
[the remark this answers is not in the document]
emilyrohrbough: All of these tests are currently ran in CI with both `experimentalSessionAndOrigin` on and off so we should be good 👍🏻

packages/config/src/options.ts
emilyrohrbough: I changed to legacy and it just didn't feel right since this behavior will be around for awhile. I went with 'lax' instead. what do you think?

packages/driver/cypress/e2e/commands/local_storage.cy.js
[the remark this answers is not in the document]
AtofStryker: ah I was thinking the regular `run` that omits the origin directory 😄

packages/server/lib/config.ts
```diff
@@ -215,6 +214,20 @@ export function mergeDefaults (
     throw makeConfigError(errors.get(err, ...args))
   }, testingType)
 
+  // TODO: https://github.com/cypress-io/cypress/issues/23093
+  // testIsolation should equal 'strict' by default when experimentalSessionAndOrigin=true
+  // Once experimentalSessionAndOrigin is made GA, remove this logic  and update the defaultValue
```
AtofStryker, line 219:
```suggestion
  // Once experimentalSessionAndOrigin is made GA, remove this logic and update the defaultValue
```

packages/server/test/unit/config_spec.js
```diff
@@ -1639,6 +1645,53 @@ describe('lib/config', () => {
           })
         })
       })
+
+      it('sets testIsolation=strict by default when experimental=true and e2e testing', () => {
```
AtofStryker, line 1649:
```suggestion
      it('sets testIsolation=strict by default when experimentalSessionAndOrigin=true and e2e testing', () => {
```

packages/server/test/unit/config_spec.js
```diff
@@ -1639,6 +1645,53 @@ describe('lib/config', () => {
...
+        return config.mergeDefaults(obj, options)
+        .then((cfg) => {
+          expect(cfg.resolved).to.have.property('experimentalSessionAndOrigin')
+          expect(cfg.resolved.experimentalSessionAndOrigin).to.deep.eq({ value: true, from: 'config' })
+          expect(cfg.resolved).to.have.property('testIsolation')
+          expect(cfg.resolved.testIsolation).to.deep.eq({ value: 'strict', from: 'default' })
+        })
+      })
+
+      it('honors user config for testIsolation when experimental=true and e2e testing', () => {
```
AtofStryker, line 1672:
```suggestion
      it('honors user config for testIsolation when experimentalSessionAndOrigin=true and e2e testing', () => {
```

lmiller1990:
Dismissing my review since a lot has changed since, will re-review once conflicts are resolved.

packages/driver/src/cy/commands/local_storage.ts
```diff
@@ -21,16 +21,21 @@ const clearLocalStorage = (state, keys) => {
...
-      // window is bound to another origin
-      return clearLocalStorage(state, [])
-    } catch (error) {
-      return null
-    }
-  })
+  // TODO: Cypress sessions will clear local storage on its own before each test.
+  // Once experimentalSessionAndOrigin is made GA, remove this logic. Leave clearing
+  // session data (cookies / local storage / session storage) to reset functionality
+  if (!Cypress.config('experimentalSessionAndOrigin')) {
```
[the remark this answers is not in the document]
emilyrohrbough: Good question. session's before:test:run:async handler also clears local storage so it is always cleared regardless of `experimentalSessionAndOrigin` being set. This was just being clears 2x before this change. Here is the code that clears this:

-`before:test:run:async` handler clears all [currentSessionData](https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cy/commands/sessions/index.ts#L46)
- which clears cookies, localStorage & session storage for all domains. [see code](https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cy/commands/sessions/manager.ts#L139)

At some point, we should clean up the local_storage command to use the same APIs as sessions. It'd align func and to allow clearing local storage in all origins, like requested in https://github.com/cypress-io/cypress/issues/2573

packages/config/src/options.ts
emilyrohrbough: okay...overruled....legacy it is!

packages/config/src/options.ts
```diff
@@ -27,31 +29,35 @@ export type BreakingOptionErrorKey =
   | 'RENAMED_CONFIG_OPTION'
   | 'TEST_FILES_RENAMED'
 
-type TestingType = 'e2e' | 'component'
+// The test-time override levels
+export const ALL_OVERRIDE_LEVELS = ['code', 'event', 'suite', 'test', 'runtime'] as const
```
chrisbreiding, line 33: I think it would be useful to have the comment above this explain what each level means

packages/config/src/browser.ts
```diff
@@ -172,16 +193,24 @@ export const validateNoBreakingTestingTypeConfig = (cfg: any, testingType: keyof
   return validateNoBreakingOptions(options, cfg, onWarning, onErr, testingType)
 }
 
-export const validateNoReadOnlyConfig = (config: any, onErr: (property: string) => void) => {
-  let errProperty
+export const validateOverridableAtTestTest = (config: any, overrideLevel: Exclude<OverrideLevel, 'never'>, onErr: (result: InvalidTestOverrideResult) => void) => {
```
chrisbreiding, line 196:
```suggestion
export const validateOverridableAtTestTime = (config: any, overrideLevel: Exclude<OverrideLevel, 'never'>, onErr: (result: InvalidTestOverrideResult) => void) => {
```

Or `validateOverridableAtRuntime`?

packages/config/src/options.ts
chrisbreiding: I'm having trouble discerning the practical difference between 'code' and 'runtime'. It seems that 'code' is when setting config at the top/global level and 'runtime' is when it's inside a test, but it seems like the result is the same either way, since they're [handled the same way during validation](https://github.com/cypress-io/cypress/pull/23040/files#diff-f635b34a5d81758a8ce5c487afae2d4fb0668271687a8dd7775898afac2fd328R102). The only difference seems to be [when crafting the error message](https://github.com/cypress-io/cypress/pull/23040/files#diff-6c0564db103220b716eb75e4a31c25a8192ac068e816710298d2e14d80cf7499R274-R276), but it seems like that's just a nicety that could apply to the 'code' error as well. Is there a case where a config value would only be overrideable at 'code' level and not 'runtime' or vice versa? If not, it seems like they could be merged into the same value.

chrisbreiding (changes requested):

packages/config/src/options.ts
emilyrohrbough: So the code level is either when the support file or spec file is loaded before mocha has processed the spec to determine the suites/tests/runnable in the code. Originally I had named it "supportFile" but found it could be when either support file or spec file was loaded into the AUT.

I do think from a messaging standpoint the delineation is helpful since runTime is when a mocha runnable itself is executing, but maybe it's not? 🤷🏻‍♀️ are you confused on the difference between the two or are you wanting "slimmer" override levels and messaging?

packages/config/src/browser.ts
emilyrohrbough: oh yes. good call out. I wanted to call this testTime, since that's what it in in my head, but I went with runtime to align with the resolve level of 'runtime' we use to pass to the cloud

packages/config/src/options.ts
chrisbreiding: I understand the difference now - thanks for the clarification. If there's a use-case for having them separate, that's fine, but I would suggest renaming them to something like `runtime-global` and `runtime-test`. However, if there's not a use-case, I'd say YAGNI 😅 It's an internal API, so we can always change it later if a use-case emerges.

cli/types/cypress.d.ts
```diff
@@ -2784,6 +2784,13 @@ declare namespace Cypress {
      * @default "cypress/support/{e2e|component}.js"
      */
     supportFile: string | false
+    /**
+     * The test isolation level applied to ensure a clean slate between tests.
+     *   - legacy - resets/clears aliases, intercepts, clock, viewport, cookies, and local storage before each test.
```
mjhenkes, line 2789: > resets/clears aliases, intercepts, clock, viewport, cookies, and local storage before each test.

With our reduced scope is all this still true?

mjhenkes: actually, yes it is, i was thinking you were saying it doesn't reset those things, derp :)

cli/types/cypress.d.ts
```diff
@@ -2784,6 +2784,13 @@ declare namespace Cypress {
...
+     *   - strict - all resets everything from legacy, plus clears the page by visiting 'about:blank' to ensure clean app state before each test.
```
mjhenkes, line 2790:
```suggestion
     *   - strict - applies all resets/clears from legacy, plus clears the page by visiting 'about:blank' to ensure clean app state before each test.
```

packages/config/src/options.ts
```diff
@@ -115,42 +116,36 @@ export const defaultSpecPattern = {
...
     name: 'animationDistanceThreshold',
     defaultValue: 5,
     validation: validate.isNumber,
-    canUpdateDuringTestTime: true,
+    overrideLevel: 'any',
   }, {
     name: 'arch',
     defaultValue: () => os.arch(),
     validation: validate.isString,
-    canUpdateDuringTestTime: false,
```
mjhenkes, line 131: i like that this defaults to false now, much cleaner

packages/config/src/options.ts
mjhenkes: thats funny, what would overriding this even do?

packages/driver/src/util/config.ts
```diff
@@ -60,3 +72,57 @@ export const preprocessConfig = (config: Cypress.Config) => {
...
+// Configuration can be override at multiple run-time levels. Ensure the configuration keys can
+// be override and that the provided override values are the correct type.
+//
+// The run-time override levels (listed in order applied):
+//   fileLoad -  config override via Cypress.config() when either loading the supportFile or specFile in the
+//                 browser (this is before mocha as process the spec
+//   suite    - config override via describe('', {...}, () => {})
+//   test     - config override via it('', {...}, () => {})
+//   event    - config override via Cypress.config() in test:before:runner or test:before:runner:async event
+//   runtime  - config override via Cypress.config() when the test callback is executed
```
chrisbreiding, line 95: This comment should be updated

chrisbreiding (changes requested):

cli/types/cypress.d.ts
emilyrohrbough: It DOES reset those things, but that is the current behavior in Cypress

packages/driver/src/util/config.ts
emilyrohrbough: I left it initially to document the order these overrides are applied since the aren't documented and not easy to find in the code base without stepping through the runner code.

chrisbreiding (approved):

packages/server/lib/util/spec_writer.ts
```diff
@@ -385,6 +385,7 @@ export const createFile = (path: string) => {
   return fs.writeFile(path, newFileTemplate(path))
 }
 
+// current un-used with 10.x release
```
mjhenkes, line 388: nit:

```suggestion
// currently un-used post 10.x release
```

mjhenkes (approved):

system-tests/test/testConfigOverrides_spec.ts
```diff
@@ -61,7 +61,7 @@ describe('testConfigOverrides', () => {
       spec: 'testConfigOverrides/invalid.js',
       snapshot: true,
       browser: browserList,
-      expectedExitCode: 8,
+      expectedExitCode: 14,
```
AtofStryker, line 64: dare I ask what exit code `14` is?

AtofStryker (approved):

system-tests/test/testConfigOverrides_spec.ts
emilyrohrbough: Yup! there are 14 failing tests in the invalid.js file

PR #23048, opened 2022-08-01
fix: vite v3 not working with node >=17

### LINKED ISSUE #23042: CT + Vite 3.0 does not work with Node >= 17
### Current behavior

A Cypress CT project using the new Vite 3.0 support does not work out of the box if the user is using a node version >=17. There is a workaround by explicitly setting the `host` to `127.0.0.1` in the `vite.config.js` (or the overrides in `cypress.config.js`), but we should not have to rely on user workarounds.

For details, see the [vite migration guide](https://vitejs.dev/guide/migration.html#dev-server-changes) regarding dev-server changes.

### Desired behavior

Cypress CT projects using Vite 3.0.0 and Node version >=17 should work without user provided workarounds.

### Test code to reproduce

1. `nvm install 18`
2. `npx init vite` => I chose a react js app
3. cd into project and run `yarn add -D https://cdn.cypress.io/beta/npm/10.4.0/darwin-x64/develop-4d3ad9edaa9b4b19ed10094279f72ed5657b851c/cypress.tgz` (Vite 3.0 hasn't been released yet so I'm using a binary from develop)
4. `yarn cypress open` => walk through CT setup
5. Create a CT test => I created `src/App.cy.jsx` with contents:
```
import App from "./App";

it("should render", () => {
  cy.mount(<App />);
});

```
7. Click on CT test, runner will hang

https://user-images.githubusercontent.com/25158820/182201979-30c9ddf2-8cc1-4785-ad7c-8f68e61c1591.mov

### Cypress Version

10.4.0 (binary from develop)

### Other

We did the same change for webpack a while back, so it's makes sense to do it here as well.
https://github.com/cypress-io/cypress/pull/21430

ZachJW34:
- Closes #23042 

### User facing changelog
CT projects using Vite v3 work with Node versions >=17, and a comma is removed from react + vite 3 apps in index.html

### Additional details
We now force the host to be `127.0.0.1` when spinning up the vite-dev-server. We did the same fix for webpack a while ago (https://github.com/cypress-io/cypress/pull/21430) and Vite v3 needs a similar fix.

I also fixed an issue that is only prevalent when using vite + react where a comma was being inserted into the served `index.html` due to how we were injecting scripts.

I added a vite setup for the `simple-ct` system test proejct and added it to the binary tests to verify behavior across node versions.

### Steps to test
This issue can be seen inside our own codebase, just switch to a node version >= 17 and try running any our our vite component tests. See the original issue for repro steps, you can now point the locally running cypress at a vite 3 project using a node version >= 17 and it will now work.

Before: (Notice the comma on the second load when I switch back to Node 16)

https://user-images.githubusercontent.com/25158820/182245411-6edafed7-8aa3-4f7a-a1a8-4e8cc25dff2f.mov

After:

https://user-images.githubusercontent.com/25158820/182245539-9a2a029f-0c10-4663-a5e3-446b451e82f7.mov

### PR Tasks


astone123 (approved):
I tested this by running some CT tests in the `app` package. Those wouldn't run before this change when using Node 18 and now they do.

Thanks 🎉

ZachJW34:
Binary tests I added aren't too happy ☹️

lmiller1990:
Vite was working on Node.js 17 just fine, in my case.

I wonder if it's either

1) only Node.js 18+ with the problem
2) linux vs macos difference?

npm/vite-dev-server/src/plugins/cypress.ts
```diff
@@ -76,7 +76,7 @@ export const Cypress = (
       indexHtmlContent = indexHtmlContent.replace(
         '<head>',
         `<head>
-        ${scriptTagsToInject.map((script) => script.toString())}
+        ${scriptTagsToInject.map((script) => script.toString()).join('')}
```
lmiller1990, line 79: Is this essential or just nice to have? Just curious

ZachJW34: Check out the before video (near the end), it was actually rendering a comma that was visible. I don't think Percy would have caught this since we don't snapshot our system tests. Going to look into this a bit more, since like you said I shouldn't have seen the comma if it was in the `<head>` so somehow it's making it's way to the `body`

AtofStryker, line 79: Is the `toString` still necessary here since [join](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join#description) should call `toString` on the contents?

```suggestion
        ${scriptTagsToInject.join('')}
```

ZachJW34: That's clean I'll throw this in

ZachJW34: After digging into this, I discovered that if we insert text into the head that doesn't parse correctly (in this case, we had the comma due to the stringified array), it will get shifted into the body.
Before:
Screen Shot 2022-08-02 at 12 13 04 PM

After:
Screen Shot 2022-08-02 at 12 12 25 PM

So this fix is more important that I thought.

lmiller1990: Why do browsers do things like "move text from `<head>` into `<body>`???? :-1:

lmiller1990: Does not seem to impact percy what-so-ever though! Weird.

Issue #22657, opened 2022-07-01
Integrate Flaky Badge & Tooltip into Specs List


mike-plummer:
https://www.figma.com/file/DgeFrhaX8N9ufrzvxzLnPg/app-cloud-int%3A-flake-tooltip?node-id=2456%3A49923

Display Flaky Badge & Tooltip as inline row data within Specs Explorer list

**Work Items**
1. Combine Flaky Badge and Flaky Badge tooltip content into a component that can be rendered as an actionable popper
2. Update Spec List props & data flow to determine whether to render flaky indicator (see #22656)
3. Display flaky badge when spec row is indicated as flaky
4. Ensure offline state is handled (no errors when cloud data not available)

**Acceptance Criteria**
1. Validate accessibility and keyboard navigation
2. Validate contrast, spacing, and responsiveness at smaller viewport widths
3. Flaky badge is only displayed for flaky spec rows
5. Tooltip displays on hover of badge and displays appropriate content
6. Badge & tooltip update accordingly when cloud recordings update (including hiding/showing badge dynamically as flaky status changes)
7. Badge and tooltip are both navigable - take user to flaky portion of dashboard
8. Component **and** E2E tests are updated to validate behavior for non-flaky and flaky spec combinations

baus:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/22657?planning-poker) with ZenHub @astone123 @lmiller1990 @marktnoonan @mike-plummer @ZachJW34

mike-plummer:
Resolved by #23049

PR #23049, opened 2022-08-01
feat: Flake Indicator ACI

### LINKED ISSUE #22656: Integrate new Flaky API data via GQL
**CLOUD dependencies to provide the following data**

Data for indicator: [CLOUD-691](https://cypress-io.atlassian.net/browse/CLOUD-691)
* "isFlaky" attribute for spec

Data for tooltip: [CLOUD-692](https://cypress-io.atlassian.net/browse/CLOUD-692)
* "Flaky Rate" - low, medium, high and percentage
* "Flaky vs total runs" totals (2 flaky runs / 50 total)
* "Last Flaky" (Last flaky 2 runs ago)
* URL to take user directly to dashboard content

**Work Items**
1. GQL schema in app is updated
2. New "flaky" components are updated to pull data from GQL
3. GQL subscription(s) are updated to ensure Flaky fields are updated by existing ACI spec polling

**Acceptance Criteria**
1. Flaky Badge is displayed at runtime for applicable specs
2. Flaky Tooltip displays correct cloud-supplied data
3. Flaky badge/tooltip is actionable to take user to dashboard page
4. Flaky badge & tooltip are updated automatically by background polling as new run data is recorded

mike-plummer:
- Closes #22656 , #22657 

### User facing changelog
* Display a "flaky test" indicator in the Specs Explorer for specs that are known to be flaky in the Cypress Dashboard

### Additional details
* Building new data adapter component to manage fetch/re-fetch of data for the tooltip
* Lots of cloud data management logic got added to `SpecsList` during initial ACI work - I took this opportunity to extract it out to a new composable. This helps simplify `SpecsList` so it can move back towards being a presentational component and helps organize some of the cloud spec data flows.
* Adding support for `RemoteFetchable*` types in our mock gql layer so we can leverage these in tests (like in `mountFragment`)

### Steps to test
1. Download and extract this example project: 
[flaky-e2e.zip](https://github.com/cypress-io/cypress/files/9237525/flaky-e2e.zip)
2. Open project in the app - ensure you are *logged out*
3. Verify that Specs Explorer does not display Flake Indicator, no user-noticeable changes to the UI
4. Log in to the app with your Staging Dashboard account, ACI data should populate
5. Verify that multiple specs are marked with a "Flaky" indicator (note - if other devs have recorded new runs as part of their testing the number of specs marked as flaky may change)
6. Mouse over Flaky indicator, verify tooltip opens and displays data re: flaky state of the test
    * If you want to validate the data you can review the project in the [Staging Dashboard](https://dashboard-staging.cypress.io/projects/ji3i6b/analytics/flaky-tests)
7. Verify that clicking on the indicator *or* on the tooltip content opens the dashboard to the flaky data section
8. Extra credit: Initiate a new recording from CLI, verify that the app refreshes to reflect in-progress and final data once the recording is complete (approx 30 second polling interval). Verify that flaky tooltips update with new data.

https://user-images.githubusercontent.com/11482842/182251591-1e20f5d9-7b57-4e7e-a34c-0d656d055079.mov

### PR Tasks


lmiller1990:
It works!

image

I have some UI notes (although I know you didn't design the UI, just implement it):

1. It was not immediately obvious I could hover the "flaky" badge
2. I had to hover for around 500ms - I think the tooltip should render faster, if not instantly, to make this more discoverable. Right now this feature requires to to hover a tooltip that isn't obviously hover-able, then wait for quite a bit of time (500ms? longer than a normal tooltip) to render.

I think the badge should change color when I hover it - maybe go darker. Or, the tooltip should render immediately (no delay).

I will look at the code now.

packages/app/cypress/e2e/specs_list_actual_git_repo.cy.ts
```diff
@@ -89,7 +89,7 @@ describe('Spec List - Last updated with git info', () => {
     cy.get('[data-cy-row="foo.spec.js"] [data-cy="git-info-row"] svg')
     .trigger('mouseenter')
 
-    cy.get('.v-popper__popper--shown').contains('Created')
+    cy.get('.v-popper__popper--shown').should('be.visible').and('contain.text', 'Created')
```
lmiller1990, line 92: I wonder if we should add a custom command for this, like `cy.assertPopper` 🤔

packages/app/cypress/e2e/specs_list_flaky.cy.ts
```diff
@@ -0,0 +1,104 @@
...
+    cy.loginUser()
+
+    cy.withCtx((ctx, o) => {
+      // Must have a cloud project ID in order to fetch flaky data
+      o.sinon.stub(ctx.project, 'projectId').resolves('abc123')
+      // Must have an active Git branch in order to fetch flaky data (see @include($hasBranch) restriction)
+      o.sinon.stub(ctx.lifecycleManager.git!, 'currentBranch').value('fakeBranch')
+    })
+
+    cy.remoteGraphQLIntercept(async (obj) => {
```
lmiller1990, line 15: Fine for now but these huge intercepts are so ugly and make the tests really verbose. After the ACI project, I wonder if we can do a retro and explore better patterns for GraphQL mocks? Or do you find these are okay to work with?

We don't have any actual e2e tests that hit a real cloud deployment, do we? I'm always a bit uneasy about that - dashboard could be down completely, and these would all still pass... we'd have no idea.

packages/app/src/composables/useCloudSpecData.ts
```diff
@@ -0,0 +1,119 @@
...
+  isOffline: Ref<boolean>,
+  projectId: string | null | undefined,
+  mostRecentUpdate: Ref<string | null>,
+  displayedSpecs: Ref<(SpecsListFragment | undefined)[]>,
+  allSpecs: (SpecsListFragment | undefined)[],
+) {
+  const refetchMutation = useMutation(CloudData_RefetchDocument)
+
+  const isCloudSpecOlderThan = (item: NonNullCloudSpec, comparisonDttm: string | null) => {
+    if (item.data?.__typename !== 'CloudProjectSpec' && item.data?.__typename !== 'CloudProjectSpecNotFound') {
```
lmiller1990, line 38: We also have `dayjs` in the dependencies, not sure if we need it for this particular comparison, but wanted to note this for the future if we need anything more complex w.r.t day/time.

packages/app/src/composables/useCloudSpecData.ts
```diff
@@ -0,0 +1,119 @@
...
+    // nothing new, no need to refetch
+    return false
+  }
+
+  /**
+   * Refetch any displayed RemoteFetchable entries that are older than the latest cloudProject update
+   * or that have not yet been fetched
+   */
+  const fetchDisplayedCloudData = async () => {
+    const cloudSpecIdsToRefetch = displayedSpecs.value
```
lmiller1990, line 74: Should we combine these into a single `reduce` so we don't need to loop so much? Small optimization but always good to reduce iteration (imo).

packages/app/src/composables/useCloudSpecData.ts
```diff
@@ -0,0 +1,119 @@
...
+  const refetchFailedCloudData = async () => {
+    const latestRunsIds = allSpecs
+    .map((s) => s?.cloudSpec)
+    .filter((cloudSpec): cloudSpec is NonNullCloudSpec => Boolean(cloudSpec?.fetchingStatus === 'ERRORED'))
+    .map((cloudSpec) => cloudSpec.id) ?? []
+
+    await refetchMutation.executeMutation({ ids: [...latestRunsIds] })
+  }
+
+  const displayedSpecIds = computed(() => displayedSpecs.value.map((v) => v?.cloudSpec?.id).join('|'))
```
lmiller1990, line 97: What is `v?.cloudSpec` is null? Do we need to filter out those potentially? I guess since we just use this for a kind of caching comparison, it's not a big deal?

packages/app/src/composables/useCloudSpecData.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  /*
+  Automatically trigger refresh & purge on the following:
+  - Set of displayed specs changes (scroll thru virtualized list)
+  - Network connectivity state changes
+  - Project connectivity state changes
+  - Latest update timestamp for cloud project changes
+  */
+  watch(
+    [debouncedDisplayedSpecIds, isOffline, isProjectDisconnected, mostRecentUpdate],
```
lmiller1990, line 108: This is interesting, I never thought about watching a debounced value. The way I've done this previously is to do something [this like](https://github.com/lmiller1990/vuejs-composition-api-v3/blob/master/src/components/PostWriter.vue#L44-L52) where you debounce the watch callback.

This is interesting too - if it works, it works! I like this composable - pretty nice way to do a kind of "if any of these things change, re-run the fetch".

packages/app/src/specs/flaky-badge/FlakyInformation.vue
```diff
@@ -0,0 +1,107 @@
...
+}
+`
+
+const props = defineProps<{
+  projectGql: FlakyInformationProjectFragment | null | undefined
+  specGql: FlakyInformationSpecFragment | null | undefined
+  cloudSpecGql: FlakyInformationCloudSpecFragment | null | undefined
+}>()
+
+const isFlaky = computed(() => props.cloudSpecGql?.data?.__typename === 'CloudProjectSpec' && !!props.cloudSpecGql?.data?.isConsideredFlaky)
```
lmiller1990, line 87: Not really any error on your end but it drives me crazy just how many null checks we need (mainly due to GraphQL having a "all the things can be nullable" philosophy).

I know we can't really do much about it, but does anyone know if there's any other patterns? How do gigantic GraphQL apps with hundreds of fields handle this problem, do they all just write the code with lots of `?` fields?

packages/app/src/specs/flaky-badge/FlakyInformation.vue
```diff
@@ -0,0 +1,107 @@
...
+  return getUrlWithParams({
+    url: flakyStatus?.dashboardUrl || '#',
+    params: {
+      utm_medium: 'Specs Flake Annotation Badge',
+      utm_campaign: 'Flaky',
+    },
+  })
+})
+
+const projectId = computed(() => props.projectGql?.projectId)
```
lmiller1990, line 101: Any reason to have all these computed instead of just passing `:props?.gql?.branch` straight to the components? Is this just for readability?

lmiller1990 (commented):
Left some comments and questions - the feature seems to work great! I also made some UI recommendations in another comment. I can approve once my comments are responded/resolved, looks great.

packages/app/cypress/e2e/specs_list_flaky.cy.ts
mike-plummer: Completely agree, I really don't like this. It's not too bad if you're just mocking out a single query/type, but when we've got multiple remote queries in a single spec and/or different permutations of data it gets ugly fast. I actually tried a few approaches to structuring this differently but ran into issues, leading to this monster. Unfortunately, since the interceptor is shipped off and run server-side we aren't able to reference anything outside function scope which precludes pulling mock data construction out into util functions/external variables and means we can't use something like `cy.fixture`

The only workaround/hack I was able to figure out was to marshal my mock data in the `beforeEach` and dump it onto `testState` using a `cy.withCtx`. The testState is available as a second interceptor arg so I could use it there and adjust as necessary with dynamic IDs and generated dates. This worked but felt like an abuse of testState, and also probably adds perf overhead.

I see a couple potential improvements for this capability:
1. Add a remote intercept command that allows specifying a discriminator/filter so we can set up multiple instances within a given test flow, rather than setting up one mega instance that covers all possible scenarios
2. Figure out a way to allow `cy.fixture` data to be resolved and passed into the intercept functions to encourage managing mock data "the cypress way"


To your second question, no, all of the remote data fetches are stubbed out to use data from the `StubGQLRegistry`. The design for all cloud features is that they should gracefully handle offline/missing data conditions, so there should be no impact (other than individual feature loss) if dashboard services are down

packages/app/src/composables/useCloudSpecData.ts
mike-plummer: Good callout - nil values wouldn't break anything in this case but could trigger unnecessary refetch calls. Will filter to truthy cloudSpec ids

packages/app/src/specs/flaky-badge/FlakyInformation.vue
mike-plummer: `<sigh />`

No good reason whatsoever, just for readability. I swear I'm trying to wean myself of this habit but my brain works in mysterious ways. Will simplify

mike-plummer:
> 1. It was not immediately obvious I could hover the "flaky" badge

Will pass this along to UX and get their thoughts

> 2. I had to hover for around 500ms - I think the tooltip should render faster, if not instantly, to make this more discoverable. Right now this feature requires to to hover a tooltip that isn't obviously hover-able, then wait for quite a bit of time (500ms? longer than a normal tooltip) to render.

This has been tricky - originally I had the tooltip pop immediately and display a default state while we fetched data, but that caused a render flicker that wasn't super attractive. Second approach was to pop immediately but render a loading spinner during fetch, but it seems like our Tooltip library doesn't handle child component swapouts very well because it caused the tooltip to jump around briefly which looked even worse (seemed like the Tooltip was trying to auto-place itself again despite size/shape remaining constant). So this was the third solution - delay rendering tooltip until data is fetched which appeared fine on my local but I imagine Australia has higher latency to staging that I do 😆 . I'll take another look because I agree that is too long of a wait - maybe I missed something in the Tooltip or CSS structure that will strike a balance

packages/app/src/composables/useCloudSpecData.ts
mike-plummer: I agree this would minimize iteration, but I feel like the `reduce` would be more difficult to grok. Since the array length here should be <100-ish (only displayed rows) I think increased readability outweighs the potential perf gains. Also (and I may be overthinking this) would the `.map/.filter` approach be more efficient for building an array than repeated `.push` calls like the reduce would call for?

```
const cloudSpecIdsToRefetch = displayedSpecs.value
    .reduce((values, spec) => {
      if (spec?.cloudSpec?.id && shouldRefetch(spec.cloudSpec)) {
        values.push(spec.cloudSpec.id)
      }

      return values
    }, [] as string[])
```

packages/app/cypress/e2e/specs_list_actual_git_repo.cy.ts
mike-plummer: Are you thinking something like `cy.get('my-thing').assertPopper('be.visible')` to validate 'my-thing' contains a visible popper element? Or something more like `cy.get('my-thing').findPopperContent().should('be.visible')`? The approach might be impacted by the way the poppers are structured - the trigger is on the thing you're looking at, but the actual popper content is appended to `body` so a normal "find within" context wouldn't make sense

lmiller1990: I didn't have a specific implementation in mind, just saw the duplication and thought a command might be nice. Off the top of my head, maybe something simple:

```js
Cypress.Commands.add('popperContains', (text) => {
  return cy.get('.v-popper__popper--shown').should('be.visible').and('contain.text', text)
})
```

You can dynamically register it for this spec only, it wouldn't need to be globally available. I don't think it's a requirement for this PR, but I wonder if we should explore more of these utility commands for single spec files that contain a lot of duplication.

packages/app/src/composables/useCloudSpecData.ts
lmiller1990: Fair enough... whenever I see `map` and `filter` combos my intuition is to merge them. To your point; we should just benchmark these things and find out exactly which one is faster and by how much. I'm guessing it's negligent in this case.

:+1:

packages/app/src/specs/flaky-badge/FlakyInformation.vue
lmiller1990: Ohhh I like the adapter pattern you are suggesting. The ugliness still exists, but you only need to look at it if you are actually working on it - right now, if you touch any UI component with graphql in a drive-by fashion, you have to see all this and grok what the heck is going on.

I'd like to see some experimentation with this, maybe in a non trivial and non time sensitive feature in the future.

lmiller1990:
Re: latency - yes I think the issue is I'm almost as far as geographically possible and my internet is not exactly the fastest. If it feels good on your end, that's fine.

lmiller1990 (approved):
Added/responded to a few more comments but I don't see any blockers in the code. Some UX discussion ongoing in Slack but I'll leave resolving that to the relevant individuals. This feature is great!

packages/app/src/composables/useCloudSpecData.ts
marktnoonan: I looked into a lot of these when I was spiking into specs list perf more generally. I agree `reduce` is less readable at a glance, it's fine but I generally find other ways of achieving the same thing clearer. 

I do think `map -> filter -> map` situations can be optimized to drop the first map, so here we could examine the `cloudSpec` property directly in the `filter` instead of doing an iteration to extract that one property, another to examine it. 

That said I did some pretty unscientific logging to figure out where time was being spent in dealing with the specs list and this stuff was not definitely not the bottleneck. I refactored a few instances of "extra" loops and it didn't make any difference, even in larger lists.

tbiethman (approved):
I tested with the listed steps, it works great across our supported browsers. 

I do still notice a bit of judder when opening the tooltip, due to the change in width going from the loading -> loaded state and the tooltip re-centering of itself at that new width. Not a deal breaker, but we could set a constant width (between the current 250 min and 400 max) and eliminate it, unless we really need to be dynamic here.

PR #23098, opened 2022-08-03
refactor(angular): handle mounting teardown before tests

### LINKED ISSUE #23096: Teardown Angular Components Before Each Test
### Current behavior

Currently the previous component that was mounted is only torn down once the next `mount` function is called. This is problematic as it leaks state from the previous test into the next.

By default angular tears down tests after each test is run. The problem with this behavior is that we want the component to remain mounted so you can see the component, and interact with it after the test has run. Therefore we use the following flag when we call `initTestEnvironment()`

```ts
teardown: { destroyAfterEach: false }
```

However, this causes the state leak described above

### Desired behavior

We need a way to manually call `teardownTestingModule` & `resetTestingModule` on `TestBed` before every tests runs. We should be able to call `setupHooks()` from `@cypress/mount-utils` to achieve this behavior

### Test code to reproduce

if you add this to a spec file you will see each screenshot shows the previous tests component on the DOM

```ts
beforeEach(() => {
  cy.snapshot()
})
```

### Cypress Version

10.5

### Other

_No response_

jordanpowell88:
- Closes #23096 

### User facing changelog
Teardown the previous component from the DOM before every test is run

### Additional details
This was pointed out by @brian-mann 

### Steps to test
run the `angular.cy.ts` system test

Tests were setup to take a screenshot before each test...

Before:

https://user-images.githubusercontent.com/25158820/182721424-410b9e91-d26b-4c8e-aef6-f4f6582934f3.mov

After:

https://user-images.githubusercontent.com/25158820/182721441-ef4fde16-dbf9-49d3-a647-748d82fdcc88.mov

### PR Tasks

- [X ] Have tests been added/updated?
- [na ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -27,7 +27,7 @@ mainBuildFilters: &mainBuildFilters
     branches:
       only:
         - develop
-        - zachw/angular-ct-teardown
+        - tbiethman/electron-19
 
 # usually we don't build Mac app - it takes a long time
 # but sometimes we want to really confirm we are doing the right thing
@@ -36,7 +36,7 @@ macWorkflowFilters: &darwin-workflow-filters
   when:
     or:
     - equal: [ develop, << pipeline.git.branch >> ]
-    - equal: [ 'zachw/angular-ct-teardown', << pipeline.git.branch >> ]
+    - equal: [ 'tbiethman/electron-19', << pipeline.git.branch >> ]
     - matches:
           pattern: "-release$"
           value: << pipeline.git.branch >>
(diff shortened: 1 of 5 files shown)


lmiller1990 (approved):

lmiller1990:
Code seems correct - not sure what's up with CI?

BlueWinds:
> Code seems correct - not sure what's up with CI?

https://github.com/yargs/yargs-parser/issues/452 is currently causing the `test-types-cypress-and-jest` job to fail. Can't speak to the rest of it.

ZachJW34 (approved):

PR #23104, opened 2022-08-04
feat: experimental single tab run mode for component testing

lmiller1990:
- Resolves https://github.com/cypress-io/cypress/issues/22353

### User facing changelog

Introduce a new Cypress experiment, `experimentalSingleTabRunMode`, for Component Testing. All specs will execute in a single tab, similar to the Component Testing alpha in Cypress 9.

### Useful Linkd
- [Technical brief](https://github.com/cypress-io/tech-briefs/pull/39/files#diff-f3d6492f15a74d4320ff63d7c537a57327aef8d71cd8e5327c3b553e25f8f098) (Cypress team only)
- [Documentation Update](https://github.com/cypress-io/cypress-documentation/pull/4638/files)
- [Binaries](https://github.com/cypress-io/cypress/commit/293a0f489bbbeae9ecc6d76053fa24eef77fc2de#comments) you can test

In this PR, I added an experimental flag that will reverted back to the Cypress 9 behavior, which will run all the specs for CT in a single tab. It's much faster (see the associated issue for numbers). There are some trade-offs, as noted in the technical brief, which is why this is behind an experimental flag.

This experiment was driven by feedback in https://github.com/cypress-io/cypress/issues/22353, and also from our dashboard users, noting running all CT specs was up to 50% slower. 

Ideally, when we work on the Run All specs feature, we will explore and find out how to best isolate specs running in a single tab. This work will likely make this experiment obsolete, and we'd just deprecate it, in favor of the new run all specs feature. This is a significant project and unlikely to land anytime soon, thus this experiment. 

### Steps to test

- pull this branch down
- go in `packages/launchpad`, which enables the experiment
- `yarn cypress:run:ct --browser chrome --config video=false`. 55s on my machine. :zap: 
- observe it's faster, and it runs all specs in 1 tab. You can add `--headed` to see it running in a single tab.
- remove experimental flag from `cypress.config.ts`
- `yarn cypress:run:ct --browser chrome --config video=false`. 91.9s on my machine. :snail: 

Component Tests run much more quickly with this experiment enabled.

### PR Tasks


packages/server/lib/modes/run.js
```diff
@@ -1105,6 +1111,16 @@ module.exports = {
       return this.currentSetScreenshotMetadata(data)
     }
 
+    if (options.experimentalSingleTabRunMode && options.testingType === 'component' && !options.isFirstSpec) {
+      // reset browser state to match e2e behavior when opening/closing a new tab
+      return openProject.resetBrowserState().then(() => {
```
lmiller1990, line 1116: We reset the state here, much like we do between tabs, so this minimizes the chance of negatively impacting spec isolation.

cli/types/cypress.d.ts
```diff
@@ -3055,6 +3055,11 @@ declare namespace Cypress {
   interface ComponentConfigOptions<ComponentDevServerOpts = any> extends Omit<CoreConfigOptions, 'baseUrl' | 'experimentalSessionAndOrigin'> {
     devServer: DevServerFn<ComponentDevServerOpts> | DevServerConfigOptions
     devServerConfig?: ComponentDevServerOpts
+    /**
+     * Enables running all specs in a single tab in run mode for faster run mode execution.
```
emilyrohrbough, line 3059:
```suggestion
     * 'Runs all component specs in a single tab, trading performance for spec isolation.
```

packages/server/lib/experiments.ts
```diff
@@ -56,6 +56,7 @@ const _summaries: StringValues = {
   experimentalSessionAndOrigin: 'Enables cross-origin and improved session support, including the `cy.origin` and `cy.session` commands.',
   experimentalModifyObstructiveThirdPartyCode: 'Applies `modifyObstructiveCode` to third party `.html` and `.js`, removes subresource integrity, and modifies the user agent in Electron.',
   experimentalSourceRewriting: 'Enables AST-based JS/HTML rewriting. This may fix issues caused by the existing regex-based JS/HTML replacement algorithm.',
+  experimentalSingleTabRunMode: 'Runs all component specs in a single tab, trading performance for spec isolation.',
```
emilyrohrbough, line 59:
```suggestion
  experimentalSingleTabRunMode: 'Runs all component specs in a single tab, trading spec isolation for faster run mode execution.',
```

packages/server/lib/modes/run.js
```diff
@@ -776,6 +776,12 @@ module.exports = {
 
   displayRunStarting,
 
+  navigateToNextSpec (spec) {
+    debug('navigating to next spec')
```
emilyrohrbough, line 780:
```suggestion
    debug('navigating to next spec %s', spec)
```

packages/server/lib/modes/run.js
```diff
@@ -1105,6 +1111,16 @@ module.exports = {
       return this.currentSetScreenshotMetadata(data)
     }
 
+    if (options.experimentalSingleTabRunMode && options.testingType === 'component' && !options.isFirstSpec) {
+      // reset browser state to match e2e behavior when opening/closing a new tab
```
emilyrohrbough, line 1115: Its not just e2e, it's the default cypress run mode behavior

packages/config/src/options.ts
```diff
@@ -225,6 +225,13 @@ const resolvedOptions: Array<ResolvedConfigOption> = [
     isExperimental: true,
     canUpdateDuringTestTime: false,
     requireRestartOnChange: 'server',
+  }, {
```
emilyrohrbough, line 228: Our config isn't in the greatest state. As is, e2e testers could set these even if the server didn't honor it. it'd be better to add this to the list of breaking e2e test configuration changes that are lower in this file.

system-tests/test/component_testing_spec.ts
```diff
@@ -106,3 +106,21 @@ describe(`React major versions with Webpack`, function () {
     })
   }
 })
+
+describe('experimentalSingleTabRunMode', function () {
+  systemTests.setup()
+
+  systemTests.it('executes all specs in a single tab', {
```
emilyrohrbough, line 113: I'm assuming no, but with our current system tests, this there any way to benchmark the performance gains from this flag?

system-tests/__snapshots__/component_testing_spec.ts.js
```diff
@@ -564,3 +564,144 @@ exports['React major versions with Webpack executes all of the tests for React v
 
 
 `
+
+exports['experimentalSingleTabRunMode / executes all specs in a single tab'] = `
+We detected that you have versions of dependencies that are not officially supported:
+
+ - \`webpack\`. Expected >=4.0.0 || >=5.0.0 but dependency was not found.
```
emilyrohrbough, line 571: Does this mean our system test isn't in a great state if we arn't testing against our supported versions?

packages/server/lib/modes/run.js
emilyrohrbough: maybe this doesn't matter since this is experimental, but if something goes wrong, this loses the test retries for CT.

packages/server/lib/modes/run.js
```diff
@@ -1244,17 +1260,16 @@ module.exports = {
...
+      if (testingType === 'e2e' || !config.experimentalSingleTabRunMode) {
+        debug('attempting to close the browser tab')
+
+        await openProject.resetBrowserTabsForNextTest(shouldKeepTabOpen)
 
-      debug('resetting server state')
-      openProject.projectBase.server.reset()
+        debug('resetting server state')
+
+        openProject.projectBase.server.reset()
```
emilyrohrbough, line 1271: we might be dropping the server reset(). does it make single tab move if we reset this regardless of the flag?

system-tests/test/component_testing_spec.ts
lmiller1990: They are fairly simple at this point - I don't think the will reveal and isolation issues, but having some that do sure would be nice. I'm not entirely sure of the best example of one, though - something that modifies `window.top`? :thinking:

packages/server/lib/modes/run.js
lmiller1990: Oh true, this seems like a problem - something we will definitely need to highlight in the notes, unless we can find a fix prior to releasing this experiment (if that's even acceptable, will need to update the tech brief with this drawback... good point).

lmiller1990:
@emilyrohrbough thanks for the review. @brian-mann said he's going to propose an alternative to what I'm doing in this PR, so I'll wait until that comes before going too deep into any of your recommended changes.

system-tests/__snapshots__/component_testing_spec.ts.js
lmiller1990: What's happening is we are falling back to the bundled version of webpack that we ship with Cypress. In open mode we actually show a warning in the launchpad. In run mode, we don't do that.

We could add `webpack` to the `package.json` for this test project, if we'd like to get rid of this warning.

packages/server/lib/modes/run.js
lmiller1990: Actually @emilyrohrbough I don't get it - I tried to use `retries` but it works okay: https://github.com/cypress-io/cypress/pull/23104/commits/4c99b4d380725d913db7a3e005d0ff77e8c34644#diff-ff7a203eea50ba8d34ac184c54310e07e4f279447a923ba9bec878e643b9e289R699-R700

Is there another way to configuring this or an edge case that I'm missing?

system-tests/test/component_testing_spec.ts
lmiller1990: As for performance, and tracking regressions, no... I think we should look into this, probably as a fast follow. One idea I had was we could just run `app-component-tests` twice using the system test infrastructure, once with the experimental flag and once without, then print a report and make some assertions. It might cost us a bit of extra $, will need to look into this more closely.

I think there's some talk of using Graphana in #team-e2e-testing but this seems like it won't land anytime soon.

I'll try to add something small in here to at least verify it is faster with the experimental flag.

packages/config/src/options.ts
lmiller1990: Good idea. Also a good chance to solicit feedback from CT users. What do you think of this?

image

packages/config/src/options.ts
```diff
@@ -4,28 +4,34 @@ import path from 'path'
...
-  | 'EXPERIMENTAL_SESSION_SUPPORT_REMOVED'
-  | 'EXPERIMENTAL_SHADOW_DOM_REMOVED'
-  | 'EXPERIMENTAL_STUDIO_REMOVED'
-  | 'FIREFOX_GC_INTERVAL_REMOVED'
-  | 'NODE_VERSION_DEPRECATION_SYSTEM'
-  | 'NODE_VERSION_DEPRECATION_BUNDLED'
-  | 'PLUGINS_FILE_CONFIG_OPTION_REMOVED'
-  | 'RENAMED_CONFIG_OPTION'
-  | 'TEST_FILES_RENAMED'
+const BREAKING_OPTION_ERROR_KEY: Readonly<AllCypressErrorNames[]> = [
```
lmiller1990, line 9: Just making things a bit more strict - now you cannot add a breaking key without an associated error message defined in `packages/errors`.

packages/server/lib/modes/run.js
lmiller1990: What do you mean "move"?

Also I don't exactly remember why we reset here at all - I'm not sure how this impacts single tab run mode uniquely, I'm guessing anything that impacts regular run mode will also impact single tab run mode.

packages/app/src/runner/aut-iframe.ts
```diff
@@ -35,6 +35,14 @@ export class AutIframe {
     return $iframe
   }
 
+  destroy () {
```
lmiller1990, line 38: This is new for single tab - we tear down the AUT between specs, right after we stop recording, but before we start the next spec.

For comparison of before/after, see: https://whimsical.com/spec-isolation-XiJyCyVUJmWQYRQbysgpP1

packages/errors/src/errors.ts
```diff
@@ -1085,6 +1085,12 @@ export const AllCypressErrors = {
 
         You can safely remove the ${fmt.highlight(`experimentalStudio`)} configuration option from your config.`
   },
+  EXPERIMENTAL_SINGLE_TAB_RUN_MODE: () => {
+    return errTemplate`\
+        The ${fmt.highlight(`experimentalSingleTabRunMode`)} experiment is currently only supported for Component Testing.
+
+        If you are enjoying the experiment with Component Testing, please join the discussion here: http://on.cypress.io/experimental-single-tab-run-mode`
```
ryanthemanuel, line 1092: Maybe have a more open ended message? They may be enjoying it, but they may not be? Either way we would want feedback.

packages/server/lib/modes/run.js
ryanthemanuel: The retries are only for the initial browser launch right?  I'm not sure we'd need retries for when we're navigating in this new way

packages/server/lib/modes/run.js
```diff
@@ -1218,6 +1234,12 @@ module.exports = {
         .catch(warnVideoRecordingFailed)
       }
 
+      const usingExperimentalSingleTabMode = testingType === 'component' && config.experimentalSingleTabRunMode
+
+      if (usingExperimentalSingleTabMode) {
```
ryanthemanuel, line 1239: Any reason we're doing this here rather than at line 1269? Seems like it might make sense to destroy the Aut at the same place that we were previously closing the tab.

astone123:
Running the CT specs in launchpad with this experiment went from ~46s to ~28s on my machine 👍🏻

packages/errors/src/errors.ts
lmiller1990: https://github.com/cypress-io/cypress/pull/23104/commits/d7aa638467cdc1054541051038285a741387c583

packages/server/lib/modes/run.js
lmiller1990: Good point, I'll move it down near there, near where we close/change tab. https://github.com/cypress-io/cypress/pull/23104/commits/e83e56cba120f1950e5f2138a1dbb1cb9cf1e35e

lmiller1990:
@ryanthemanuel thanks for the quick review! Updated.

packages/app/src/runner/event-manager.ts
```diff
@@ -54,6 +55,8 @@ export class EventManager {
   studioRecorder: any
   selectorPlaygroundModel: any
   cypressInCypressMochaEvents: CypressInCypressMochaEvent[] = []
+  // Used for testing the experimentalSingleTabRunMode experiment. Ensures AUT is correctly destroyed between specs.
+  autDestroyedCount = 0
```
emilyrohrbough, line 59: Any way we can setup this up in the e2e/ct support files so we don't have this hanging around in prod code?

lmiller1990: Sure thing, nice idea. Done: https://github.com/cypress-io/cypress/pull/23104/commits/3cb9ab34205e892cca64c453ff92a1cdf15d208d

ryanthemanuel (approved):
Looks good. Nice work!

astone123 (approved):

emilyrohrbough (approved):

brian-mann:
@lmiller1990 
 can you upload a screenshot of what the settings UI looks like for E2E vs CT that indicate which experiments are available?

I would expect that `experimentalSingleTabRunMode` experiment option is displayed as being available to set in CT, but not in E2E. Additionally because this is a `testingType` specific option, I would expect the docs to explain this and the option to guide the user towards setting it in the right place (not the top level).

PR #23142, opened 2022-08-05
fix: remove dependency causing semantic-release to fail

ZachJW34:
### User facing changelog
na

### Additional details
Semantic release has been failing, possibly due to the [semantic-release bump](https://github.com/cypress-io/cypress/pull/22238) a few weeks ago. Hard to verify since we haven't had a passing `npm-release` job in a while. The problem was due to peer-deps from the `@cypress/code-coverage` package in `npm/vue`. Removing this allowed the `npm-release` job to move pass the error occurring in the `@semantic-release/npm prepare` plugin (I verified by sshing into the last master `npm-release` job and running `yarn npm-release`. The tokens were out of date so no packages were published).

We weren't verifying the results of `code-coverage` for the vue cypress tests, and it makes more sense in a system test anyway. In fact, we aren't even running the `npm/vue` cypress tests on CI right now. Going to follow up on this.

### Steps to test
No easy way to test this, since running `yarn npm-release` locally is mostly stubbed. I had to ssh into circleci to verify the behavior.

na

### PR Tasks

### DIFF
--- a/npm/angular/package.json
+++ b/npm/angular/package.json
@@ -2,6 +2,7 @@
   "name": "@cypress/angular",
   "version": "0.0.0-development",
   "description": "Test Angular Components using Cypress",
+  "private": true,
   "main": "dist/index.js",
   "scripts": {
     "prebuild": "rimraf dist",
--- a/npm/vue/cypress.config.ts
+++ b/npm/vue/cypress.config.ts
@@ -1,5 +1,4 @@
 import { defineConfig } from 'cypress'
-import { devServer } from '@cypress/webpack-dev-server'
 
 export default defineConfig({
   'viewportWidth': 500,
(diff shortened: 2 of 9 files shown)


npm/angular/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@cypress/angular",
   "version": "0.0.0-development",
   "description": "Test Angular Components using Cypress",
+  "private": true,
```
ZachJW34, line 5: Should have been marked as private since we haven't released yet 😅

admah: Good catch!

admah (approved):
Looks good!

tbiethman (approved):

PR #23163, opened 2022-08-05
fix: hitting enter in the  'Create Blank Spec' dialog will create the spec

### LINKED ISSUE #21815: 'Create Blank Spec' dialog should be a form. Hitting "Enter" should behave like a form submit button.
Currently, the “Create New Blank Spec” modal is functionally a form, but does not behave like one. The “Create Spec” button does nothing when Enter is pressed. This is a form, but doesn’t behave like one. For a11y reasons, we should be treating this flow as a form submission.
On mac: create new empty spec ~~> name file (or not)~~ > {press enter}

Screen Shot 2022-04-04 at 5 06 18 PM

┆Issue is synchronized with this [Jira Task](https://cypress-io.atlassian.net/browse/UNIFY-1516) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/11698">Screen Shot 2022-04-04 at 5.06.18 PM.png</a>
┆author: Jess Sachs
┆epic: UI/UX Improvements
┆friendlyId: UNIFY-1516
┆priority: Lowest
┆sprint: Backlog
┆taskType: Task

warrensplayer:
Fixes #21815

- Closes  https://github.com/cypress-io/cypress/issues/21815

### User facing changelog
In the "Create Blank Spec" dialog, pressing enter or return when the focus is on the spec path input field will create the blank spec.

### Additional details
* This appears to the first use of the `form` tag in the Cypress app.  
* The change was isolated to just wrap the one input field that makes up the form in this modal.  
* Currently, the `Button.vue` component (packages/frontend-shared/src/components/Button.vue) does not have a `type` attribute.  All buttons in the app default to a type of "submit".  This is not the correct semantics for buttons that are not part of a form.  I did not want to make any changes to such a commonly used component for this simple issue.  Additional uses of the `form` tag will want to consider updating the "Button.vue" component to allow for setting the type appropriately.

### Steps to test

Click the "New Spec" button
1_click_new_spec

This opens the new blank spec modal if your project already has specs in it
2_new_spec_modal

Type a name for the spec file or leave as the default
3_type_spec_name

Hit the "Enter" or "Return" key and your new spec will be created just like clicking the "Create spec" button
4_new_blank_spec_created

### PR Tasks


packages/app/src/specs/CreateSpecModal.cy.tsx
```diff
@@ -62,6 +62,34 @@ describe('<CreateSpecModal />', () => {
...
+        return defineResult({ matchesSpecPattern: variables.specFile !== '' })
+      })
+
+      //try to submit an empty path which is invalid
+      cy.get('input')
+      .clear()
+      .type('{enter}')
+
+      //should stay on current state
+      cy.contains('h2', 'Enter the path for your new spec')
```
ZachJW34, line 89: Would be more meaningful to expect that the `Great! The spec was successfully added` does not exist.

ZachJW34 (approved):
Looks great! One Ux improvement could be to auto focus the `Okay, run the spec` button on the next modal. If the user is using the keyboard to submit, they'd probably expect to be able to hit `enter/return` again and be dropped into the runner.

Also, sounds like it will be worth it to create a followup issue for fixing the `Button.vue` type attribute issue.

lmiller1990 (approved):
Code seems fine - another alternative for testing, instead of adding a component test, would be to change [this test](https://github.com/cypress-io/cypress/blob/develop/packages/app/cypress/e2e/specs.cy.ts#L273-L275) to do an `{enter}` instead of a click, but I think both are fine.

packages/app/src/specs/CreateSpecModal.cy.tsx
```diff
@@ -62,6 +62,34 @@ describe('<CreateSpecModal />', () => {
...
+
+      //should switch to success state
+      cy.contains('h2', 'Great! The spec was successfully added')
+      .should('be.visible')
+    })
+
+    it('enter should not call create spec function if spec file path is invalid', () => {
+      cy.stubMutationResolver(EmptyGenerator_MatchSpecFileDocument, (defineResult, variables) => {
+        //mocking one pattern use case
+        return defineResult({ matchesSpecPattern: variables.specFile !== '' })
```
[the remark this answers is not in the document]
warrensplayer: @ZachJW34 I had just a `false` originally, but the test would still pass if the `.clear()` was removed.

packages/app/src/specs/CreateSpecModal.cy.tsx
warrensplayer: I was trying to figure out the best way to indicate that the state of the modal did not change (i.e. clicking the button had no effect).  What do you think about grabbing the text of the H2 at the beginning of the test, then doing the `clear()` and `.type('{enter}')`, then testing if the H2 was the same as what was captured in the beginning?

lmiller1990: What do you think about: https://github.com/cypress-io/cypress/pull/23205

Since the goal is to assert the form is not valid, we can use the fact the button is disabled to infer this.

lmiller1990 (approved):
Feature works great, made a recommendation for the spec in https://github.com/cypress-io/cypress/pull/23205 but not really a blocker on merging. I'll leave the rest to @warrensplayer :+1:

Issue #22086, opened 2022-06-03
"Still waiting to connect to Firefox" issue in Cypress 10.0.2 on Windows


laerteneto:
### Current behavior

The Firefox doesn't load Cypress in both open and headless mode
![image](https://user-images.githubusercontent.com/9399511/171828216-c7ad1cf1-db54-4572-a20f-091f406d3536.png)

### Desired behavior

Cypress must be loaded within the Firefox browser

### Test code to reproduce

In the open mode:
- cypress open
- choose Firefox

Headless:
 cypress run --browser firefox

### Cypress Version

10.0.2

### Other

Windows 10
Firefox 101

lmiller1990:
Thanks for the bug report, and upgrading to Cypress 10 so quickly. I reproduced this - Firefox is timing out for me, too. We are investigating and will publish a patch when we find the cause and solution.

stacks13:
It's timing out on Cypress 8 and 9 as well.

lmiller1990:
Oh, I am also seeing the same problem on 9.7.0 as pointed out by @stacks13. I guess this is not a Cypress 10 specific regression? Needs more investigation.

roshan04:
@lmiller1990 I am facing the same issue with firefox 101 on cypress 9.5.1. The browser is getting launched but still showing the error: **Waiting to connect to firefox**. Attaching the screenshots. Can you help here.
<img width="1234" alt="Screenshot 2022-06-03 at 4 06 40 PM" src="https://user-images.githubusercontent.com/47280057/172102173-8720bfdc-3869-41a9-8f95-1e666e520456.png">
<img width="1243" alt="Screenshot 2022-06-03 at 4 07 16 PM" src="https://user-images.githubusercontent.com/47280057/172102295-df90aba4-ce7d-42ff-aeb8-7a96a09bb8a8.png">

lmiller1990:
Thanks for sharing - I reproduced and can clarify it's windows + Firefox only.

I am unsure why this is occurring, and if it's two different, unrelated regressions or the same. We will investigate.

fanksin:
Check if you get any error on 127.0.0.1

If you use any kind of local development environment software like xampp or mamp, etc, be sure they are properly closed. If not, start your local server and stop it after everything works properly. 

Besides you can run the command **pnpx kill port 51827** which kills the port I came across that was being used by cypress to run Firefox on 127.0.0.1:51827

laerteneto:
[...]

PR #23164, opened 2022-08-05
chore: adding granular Firefox browser validation

tbiethman:
We recently ran into [an issue](https://github.com/cypress-io/cypress/issues/22086) where Cypress failed to connect to Firefox on Windows. A bug was identified within Firefox and a fix was released with v103, but v101 and v102 were not patched. Those versions will never work with Cypress on Windows.

I wanted to ensure we could communicate this incompatibility through the app, rather than have users run into this and have to be redirected to the closed issue. However, there wasn't a clean way of marking these specific version/architecture combinations as unsupported with our current logic. So I did a bit of refactoring to allow our default `browsers` objects to have an associated `validator` function that can more dynamically determine Cypress support.

### User facing changelog

* Stable Firefox versions 101 and 102 are now explicitly rendered as unsupported in Windows due to an unpatched browser incompatibility in those versions.

### Steps to test

We have pretty good test coverage around our minimum version checks. The new Firefox/Windows checks can be tested by installing [101](https://support.mozilla.org/en-US/kb/install-older-version-firefox) (or 102) and launching Cypress. You should see this:

Screen Shot 2022-08-03 at 5 00 33 PM

103 should render as normal.

### PR Tasks


packages/launcher/lib/browsers.ts
```diff
@@ -1,14 +1,161 @@
...
+  }
+
+  return {
+    isSupported: true,
+  }
+}
 
 /** list of the browsers we can detect and use by default */
 
+export const browsers: Browser[] = [
```
tbiethman, line 37: With the additional logic in play here, it didn't feel right keeping the default browsers defined within the `types` package. It was a little weird to have them there in the first place?

packages/launcher/lib/browsers.ts
```diff
@@ -1,14 +1,161 @@
...
+
+      if (platform === 'win32' && browser.majorVersion && ['101', '102'].includes(browser.majorVersion)) {
+        return {
+          isSupported: false,
+          warningMessage: `Cypress does not support running ${browser.displayName} version ${browser.majorVersion} on Windows due to an unpatched defect. To use ${browser.displayName} with Cypress on Windows, install version 103 or newer.`,
+        }
+      }
+
+      return validateMinVersion(browser)
+    },
```
tbiethman, line 97: I did not utilize the same validator for the dev/nightly channels as the impacts are much more undefined; some 100/101/102/103/104 versions will be impacted, some won't.

packages/launcher/lib/detect.ts
```diff
@@ -25,20 +26,24 @@ type HasVersion = Omit<Partial<FoundBrowser>, 'version' | 'name'> & {
...
-  if (unsupportedVersion) {
-    foundBrowser.unsupportedVersion = true
-    foundBrowser.warning = `Cypress does not support running ${browser.displayName} version ${majorVersion}. To use ${browser.displayName} with Cypress, install a version of ${browser.displayName} newer than or equal to ${browser.minSupportedVersion}.`
+  if (isSupported) {
+    return
   }
 
-  return foundBrowser
+  browser.unsupportedVersion = true
+  browser.warning = warningMessage
```
tbiethman, line 46: I'm noticing now that I abort early if isSupported is truthy; anyone see value 'it is supported but still has a warning' use case?

packages/launcher/test/unit/detect_spec.ts
```diff
@@ -110,7 +145,7 @@ describe('browser detection', () => {
             info: 'Loaded from /foo/bar/browser',
             custom: true,
             version: '9001.1.2.3',
-            majorVersion: 9001,
+            majorVersion: '9001',
```
tbiethman, line 148: We were previously setting the majorVersion as an integer version if it could be parsed in [`setMajorVersion`](https://github.com/cypress-io/cypress/pull/23164/files#diff-3c69c99db158ca335f24134e072fa8e4b28c55cec925262ce1959aca87c986aaL30-L34). But it's defined as a string on the FoundBrowser type and looks like it should to be (sometimes the version contains chars). So I normalized the type being set, and updated these tests.

packages/launcher/lib/browsers.ts
```diff
@@ -1,14 +1,161 @@
...
+    minSupportedVersion: MIN_FIREFOX_VERSION,
+    validator: (browser: FoundBrowser, platform: NodeJS.Platform): BrowserValidatorResult => {
+      // Firefox 101 and 102 on Windows features a bug that results in Cypress being unable
+      // to connect to the launched browser. A fix was first released in stable 103.
+      // See https://github.com/cypress-io/cypress/issues/22086 for related info.
+
+      if (platform === 'win32' && browser.majorVersion && ['101', '102'].includes(browser.majorVersion)) {
+        return {
+          isSupported: false,
+          warningMessage: `Cypress does not support running ${browser.displayName} version ${browser.majorVersion} on Windows due to an unpatched defect. To use ${browser.displayName} with Cypress on Windows, install version 103 or newer.`,
```
flotwig, line 92: specificity?

```suggestion
          warningMessage: `Cypress does not support running ${browser.displayName} version ${browser.majorVersion} on Windows due to a blocking bug in Firefox. To use ${browser.displayName} with Cypress on Windows, install version 103 or newer.`,
```

flotwig (approved):
Nice, tested that this blocks FF 101 on Windows and not on Linux, looks good.

Noticed a new defect though in the process :smile:  #23200

packages/launcher/lib/browsers.ts
flotwig: :100:

packages/launcher/test/unit/browsers_spec.ts
```diff
@@ -14,4 +15,179 @@ describe('browsers', () => {
...
+          ...firefoxBrowser,
+          version: '102.1.0',
+          majorVersion: '102',
+          path: '/path/to/firefox',
+        } as FoundBrowser
+
+        const result = firefoxBrowser.validator(foundBrowser, 'win32')
+
+        expect(result.isSupported).to.be.false
+        expect(result.warningMessage).to.contain('Cypress does not support running Firefox version 102 on Windows due to an unpatched defect.')
```
tbiethman, line 50:
```suggestion
        expect(result.warningMessage).to.contain('Cypress does not support running Firefox version 102 on Windows due to a blocking bug in Firefox.')
```

packages/launcher/lib/browsers.ts
marktnoonan: In theory we could use the `info` property to set more a of "ymmv" version of the message for those channels, but it is not actually rendering in the UI at the moment: https://github.com/cypress-io/cypress/issues/21769

This is early in my list to ask to prioritize after the search work is done.

packages/launcher/lib/detect.ts
```diff
@@ -25,20 +26,24 @@ type HasVersion = Omit<Partial<FoundBrowser>, 'version' | 'name'> & {
...
-  const majorVersion = parseInt(ver) || browser.version
-
-  const unsupportedVersion = browser.minSupportedVersion && majorVersion < browser.minSupportedVersion
+export const getMajorVersion = (version: string): string => {
+  return version.split('.')[0]
+}
 
-  const foundBrowser = extend({}, browser, { majorVersion })
+// Determines if found browser is supported by Cypress. If found to be
+// unsupported, the the browser will be unavailable for selection and
```
ryanthemanuel, line 34:
```suggestion
// unsupported, the browser will be unavailable for selection and
```

ryanthemanuel (approved):
One small comment. Tested this and it is working well. Nice work!

packages/launcher/lib/browsers.ts
tbiethman: Thanks for the heads up Mark, using `info` (when it works) would be perfect

Issue #23032, opened 2022-08-01
Better messaging about React 18 support


lmiller1990:
### What would you like?

Right now if you use `cypress/react` with React 18, React gives you a warning in your console. The next release of Cypress ships `cypress/react18`. We should detect if a user is on React 18, and if they `import { mount } from 'cypress/react'`, we should print a warning telling them to update their import to from `cypress/react` to `cypress/react18`.

PR landing React 18 is here: https://github.com/cypress-io/cypress/pull/22876

### Why is this needed?

So users know that we've got support for the version of React they are using.

PR #23182, opened 2022-08-08
chore: add warning encouraging react users to update mount function

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/23032

### User facing changelog

N/A

### Additional details

Add a useful warning for users on the old React API, `cypress/react`, who are using React 18. They can now take advantage of the React 18 integration, `cypress/react18`.

### Steps to test

It's a trivial change. I tested it out, here's how it looks:

## Before (warning in console and command log)

image

## After (no, warning in console or command log)

image

### PR Tasks


astone123 (approved):
Looks good, nice improvement 👍🏻

npm/react/package.json
```diff
@@ -14,10 +14,14 @@
     "test": "yarn cy:run",
     "watch": "yarn build --watch --watch.exclude ./dist/**/*"
   },
+  "dependencies": {
+    "semver": "^7.3.2"
```
ZachJW34, line 18: Noticed this is getting bundled in the rollup output. It's not a huge amount of code, but since the `cli` package has `semver` listed as a dependency we could mark this package as "external" and save some bundle size.

lmiller1990: Had this thought too, but what about people using via npm? I think we should make the adapters as self-contained as possible?

Also, since we do a deep import, the bundle size barely changes - we only grab the `major` function.

ZachJW34: If it's listed as a dependency, then users who download it via npm will have semver downloaded for them. I'm not saying remove it from `packageJson.dependencies`, just have it added to `rollupConfig.externals`.

lmiller1990: Right, gotcha.

lmiller1990:
I will get this green and merge it up.

ZachJW34 (approved):

PR #23187, opened 2022-08-08
chore: migrate runner/dom.js

sainthkh:
- Part 2 of #21689

### User facing changelog

### Additional details

- Why was this change necessary? => Some legacy codes are left inside `runner`. This PR migrates `dom` codes. 

To migrate `runner/dom` to `pkg/app`, these things had to be done:

  * Migrate runner/dom.js
  * Mgirate Highlight React component to Vue. 
  * Migrate AssertionsMenu React component to Vue. 
  * Remove JQuery as much as possible. 

The result is like this:

* `src/runner/dom.ts` for common codes for `Highlight` and `AssertionsMenu`.
* If it's related to placing `Highlight` correctly, I moved it inside `aut-iframe`.
* I moved `Highlight` to `runner/selector-playground`.
* I moved `AssertionsMenu` to `runner/studio`. And I splitted the file to 3 component files. 

I used `floating-ui` rather than our `Tooltip` because `floating-vue` doesn't support shadow dom. It seems that it's fixed after floating-ui@1.x.

### Steps to test

* Check if Selector Playground's `Highlight` works correctly.
* Check if Studio's Assertions menu works correctly.

N/A.

### PR Tasks


sainthkh:
I checked the `cy.origin` test in my local Ubuntu 20.04. It hangs forever in the current `develop` branch.

emilyrohrbough (commented):
@sainthkh Love the changes. This is doing a bit more than just migrating runner/dom.js. Any way we can split this into three PRs? One for migrating dom.js, one for migrating selector-playground and one for migrating studio? It'd be easier to review with smaller changes.

packages/frontend-shared/package.json
```diff
@@ -20,6 +20,7 @@
   "devDependencies": {
     "@antfu/utils": "^0.3.0",
     "@cypress-design/css": "^0.5.2",
+    "@floating-ui/dom": "1.0.1",
```
emilyrohrbough, line 23: why is this added to this package vs the app package?

sainthkh: It's my mistake and I'll fix this my `Highlight` PR.

sainthkh:
As this PR is too big, it'll be break into 3 PRs.

* showing-Highlight-related `dom.js` functions: #23792
* Highlight component: #23973
* AssertionsMenu component: #24286

PR #23204, opened 2022-08-08
fix(react18): unmount component with react18 API

### LINKED ISSUE #23081: The React18 support in Component Testing doesn't work
### Current behavior

Currently when running a CT using the new entry "cypress/react18" I get the following error:

image

Which is thrown from https://github.com/cypress-io/cypress/blob/c058bf287861f36371ee2c589dd1e08213e5dfa9/npm/react/src/createMount.ts#L172 because the lastMountedReactDom is react-dom/client which only exports createRoot and hydrateRoot (https://github.com/facebook/react/blob/9fcaf88d58cfd942e2fdd303ae8291dbf4828969/packages/react-dom/client.js#L25)

### Desired behavior

_No response_

### Test code to reproduce

~I can't provide a reproduction at the moment but will provide one soon. Using the react18 entry in a CT suite should cause it immediately.~
https://github.com/joshwooding/cypress-react18-repro

### Cypress Version

10.4.0

### Other

_No response_

rockindahizzy:
- Closes #23081

### User facing changelog
- Fixes issue in component testing when using React 18 mount

### Steps to test
- Build `npm/react` and `npm/react18` projects
- Copy the `dist` folders from the projects and place them in a React project under their respective React folders in `node_modules/cypress`
- Run cypress in the React project

### PR Tasks

### DIFF
--- a/npm/react/src/createMount.ts
+++ b/npm/react/src/createMount.ts
@@ -137,7 +137,7 @@ export const makeMountFn = (
  */
 export const makeUnmountFn = (options: UnmountArgs) => {
   return cy.then(() => {
-    const wasUnmounted = mountCleanup()
+    const wasUnmounted = mountCleanup?.()
 
     if (wasUnmounted && options.log) {
       Cypress.log({
--- a/npm/react/src/createMount.ts
+++ b/npm/react/src/createMount.ts
@@ -1,5 +1,4 @@
 import * as React from 'react'
-import ReactDOM from 'react-dom'
 import getDisplayName from './getDisplayName'
 import {
   injectStylesBeforeElement,
@@ -7,7 +6,7 @@ import {
   ROOT_SELECTOR,
   setupHooks,
 } from '@cypress/mount-utils'
-import type { InternalMountOptions, InternalUnmountOptions, MountOptions, MountReturn, UnmountArgs } from './types'
+import type { InternalMountOptions, MountOptions, MountReturn, UnmountArgs } from './types'
 
 /**
  * Inject custom style text or CSS file or 3rd party style resources
@@ -20,7 +19,7 @@ const injectStyles = (options: MountOptions) => {
   }
 }
 
-export let lastMountedReactDom: (typeof ReactDOM) | undefined
+export let mountCleanup: InternalMountOptions['cleanup']
 
 /**
  * Create an `mount` function. Performs all the non-React-version specific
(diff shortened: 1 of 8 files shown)


npm/react/src/createMount.ts
```diff
@@ -136,40 +135,33 @@ export const makeMountFn = (
...
-                description: 'Unmounts React component',
-                parent: $el[0],
-                home: 'https://github.com/cypress-io/cypress',
-              }
-            },
-          })
-        }
-      }
-    })
+    const wasUnmounted = mountCleanup()
```
lmiller1990, line 140: Can this be undefined? I'm surprised we don't need a null check, since the definition is `let mountCleanup: InternalMountOptions['cleanup']`

Should we add a null check? Could just be `mountCleanup?.()`.

npm/react/src/mount.ts
```diff
@@ -1,34 +1,44 @@
...
   makeUnmountFn,
-  lastMountedReactDom,
 } from './createMount'
 import type {
   MountOptions,
   InternalMountOptions,
-  InternalUnmountOptionsReact,
 } from './types'
 
+let lastReactDom: any
```
lmiller1990, line 13: Could be `typeof ReactDOM`?

npm/react18/src/index.ts
```diff
@@ -23,20 +32,12 @@ export function mount (jsx: React.ReactNode, options: MountOptions = {}, rerende
...
-  const internalOptions: InternalUnmountOptionsReact18 = {
-    // type is ReturnType<typeof ReactDOM.createRoot>
-    unmount: (): boolean => {
-      root.unmount()
-
-      return true
-    },
-  }
-
-  return makeUnmountFn(options, internalOptions)
```
lmiller1990, line 41: The file below this (`_snapshots__`) has a surprisingly huge diff with a lot of red. Is this right? It seems to be missing the "summary" part of the CLI output.

lmiller1990:
I tried it out (build the new react modules and stuck them in an existing project's node_modules/cypress) and it's working great, nice job :100:

lmiller1990 (approved):

ZachJW34 (approved):

lmiller1990:
Nice, second latest job went all green: https://app.circleci.com/pipelines/github/cypress-io/cypress?branch=zachw%2Ffix-react18-unmount

Not sure why the info in the checkbox is showing red, maybe because I canceled the later build (due to the previous one passing). Let's ship it.

lmiller1990 (approved):

PR #23205, opened 2022-08-09
chore: improve spec for #21815

lmiller1990:
Potential improvement to spec for #21815.

### DIFF
--- a/packages/app/src/specs/generators/EmptyGenerator.vue
+++ b/packages/app/src/specs/generators/EmptyGenerator.vue
@@ -56,6 +56,7 @@
       >
         <Button
           size="lg"
+          type="submit"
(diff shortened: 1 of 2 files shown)


packages/app/src/specs/generators/EmptyGenerator.vue
```diff
@@ -56,6 +56,7 @@
       >
         <Button
           size="lg"
+          type="submit"
```
warrensplayer, line 59: I did not realize you could do fallthrough atrributes like this with Vue (https://vuejs.org/guide/components/attrs.html), but learned something new today!

warrensplayer (approved):
I like the addition!  Does this type of PR into another branch need another review?

PR #23226, opened 2022-08-09
fix: cypress hang caused in run mode when test:before:run handler throws

### LINKED ISSUE #23039: [driver] errors thrown from test:before:run are incorrectly handled

emilyrohrbough:
- Closes #23039 

When in runMode, and a test fails, we will generate the before & after snapshots. We are relying on the reporter's UI to update to indicate when the before snapshot has been taken. Because `before:test:run` has yet to start the runnable, no logs have been updated and the reporter UI is essentially static and won't be updated, which means the snapshot callback was never executed, which caused the hang.

### User facing changelog
Fixed an edge-case bug where Cypress would hang in run-mode if an error was through from a `test:before:run` event handler. Fixed [#23039](https://github.com/cypress-io/cypress/issue/23039).

### PR Tasks

- [n/a Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 


packages/driver/src/cypress/runner.ts
```diff
@@ -1426,23 +1434,27 @@ export default {
 
         const isHook = runnable.type === 'hook'
 
+        let runnableName = 'test'
+        let runnableId = runnable.id
+
+        if (isHook) {
```
emilyrohrbough, line 1440: this was more clean-up....

packages/reporter/src/test/test-model.ts
```diff
@@ -163,6 +163,10 @@ export default class Test extends Runnable {
   }
 
   @action update (props: UpdatableTestProps, cb: UpdateTestCallback) {
+    if (this.state === 'processing' && !props.state) {
```
emilyrohrbough, line 166: this was the root-cause of this hang.

packages/driver/src/cypress/runner.ts
```diff
@@ -1534,6 +1546,23 @@ export default {
...
+        // then override the test function to associate the error to the test
+        const handleBeforeTestEventError = (err: Error): boolean => {
+          const { fn } = runnable
+
+          const restore = () => {
+            return runnable.fn = fn
+          }
+
+          runnable.fn = () => {
+            restore()
```
mjhenkes, line 1559: Could this line just be resetting runnable.fn to fn instead of calling the restore function?
```suggestion
            runnable.fn = fn
```

mjhenkes (approved):

packages/driver/src/cypress/runner.ts
```diff
@@ -1534,6 +1546,23 @@ export default {
...
 
+        // if either the TEST_BEFORE_RUN_EVENT or TEST_BEFORE_RUN_ASYNC_EVENT throws
+        // then override the test function to associate the error to the test
+        const handleBeforeTestEventError = (err: Error): boolean => {
+          const { fn } = runnable
+
+          const restore = () => {
+            return runnable.fn = fn
+          }
+
```
emilyrohrbough, line 1557:
```suggestion
```

AtofStryker (approved):

PR #23251, opened 2022-08-10
feat: scaffold react18 projects in component setup

### LINKED ISSUE #23033: Seed `react18` for new React 18 projects
### What would you like?

Cypress ships support for React 18 under `cypress/react18`. When you set up Cypress with a React project using the launchpad, it seeds `import { mount } from 'cypress/react'` in your supportFile, which isn't really ideal.

We should detect the React version and give them the correct mounting function. 

### React <= 17

```js
```

### React 18
```js
import { mount } from 'cypress/react18'
```

The code for generating the `supportFile` is [here](https://github.com/cypress-io/cypress/blob/develop/packages/scaffold-config/src/supportFile.ts).

To see an example of the code we generate, the easiest way is to make a new React app and add Cypress!

- create with Vite: `yarn create vite react-app --template react`
- go in there and install deps
- commit your changes
- install cypress from prod: `yarn add cypress --dev`
- `yarn cypress open`
- choose component testing
- follow prompts
- git status and see what was generated (all inside `cypress`. 
  - Notice `cypress/support/component.js` does `import { mount } from 'cypress/react'`. 
  - Ideally we should detect if they are using React 18 (this template does) and do `import { mount } from 'cypress/react18'`).

### Why is this needed?

It's useful to seed the the correct mounting function that corresponds to the React version so the test code matches the production code, which would be using the new `createRoot` API from` react-dom/client`.

### Other

_No response_

ZachJW34:
- Closes #23033

### User facing changelog
React 18 projects will now be scaffolded with the `cypress/react18` mount registration during component setup

### Additional details
Refactored the `mountModule` framework property to be async, allowing the detection of the installed react version to determine `cypress/react` vs `cypress/react18`. React is the only framework that needs the detection, but I've refactored the other frameworks so they have the same function signature

### Steps to test
Create a project using React 18, for example `npm init vite` -> choose React

Open project with local Cypress, verify `support/component.js` contains `import { mount } from 'cypress/react18'`

https://user-images.githubusercontent.com/25158820/183974906-51679798-bee6-4f1e-920a-b771c92c56bf.mov

### PR Tasks


package.json
```diff
@@ -50,7 +50,7 @@
     "stop-only": "npx stop-only --skip .cy,.publish,.projects,node_modules,dist,dist-test,fixtures,lib,bower_components,src,__snapshots__ --exclude cypress-tests.ts,*only.cy.js",
     "stop-only-all": "yarn stop-only --folder packages",
     "pretest": "yarn ensure-deps",
-    "test": "yarn lerna exec yarn test --scope cypress --scope \"'@packages/{config,errors,data-context,electron,extension,https-proxy,launcher,net-stubbing,network,proxy,rewriter,socket}'\"",
+    "test": "yarn lerna exec yarn test --scope cypress --scope \"'@packages/{config,errors,data-context,electron,extension,https-proxy,launcher,net-stubbing,network,proxy,rewriter,scaffold-config,socket}'\"",
```
ZachJW34, line 53: Unit tests for `scaffold-config` were not being run in CI

packages/scaffold-config/src/frameworks.ts
```diff
@@ -61,6 +61,22 @@ function getBundlerDependency (bundler: WizardBundler['type'], projectPath: stri
...
+
+export type WizardMountModule = typeof WIZARD_MOUNT_MODULES[number]
+
+const mountModule = (mountModule: WizardMountModule) => async (projectRoot: string) => Promise.resolve(mountModule)
+
+const reactMountModule = async (projectPath: string) => {
+  const reactPkg = await isDependencyInstalled(dependencies.WIZARD_DEPENDENCY_REACT, projectPath)
+
+  try {
+    return semver.major(reactPkg.detectedVersion!) === 18 ? 'cypress/react18' : 'cypress/react'
```
ZachJW34, line 74: `semver.major` can throw

packages/scaffold-config/test/unit/detect.spec.ts
```diff
@@ -76,7 +76,7 @@ describe('detectFramework', () => {
     const projectPath = await scaffoldMigrationProject('create-react-app-unconfigured')
 
     fakeDepsInNodeModules(projectPath, [{ dependency: 'react-scripts', version: '5.0.0' }])
-    const actual = detectFramework(projectPath)
+    const actual = await detectFramework(projectPath)
```
ZachJW34, line 79: All of the test in this file were failing, the `detectFramework` was changed to be async recently but since we were never running on CI we didn't catch this

scripts/gulp/tasks/gulpCypress.ts
```diff
@@ -172,7 +172,7 @@ export async function startCypressWatch () {
   }
 
   const watcher = chokidar.watch([
-    'packages/{graphql,data-context}/src/**/*.{js,ts}',
+    'packages/{graphql,data-context,scaffold-config}/src/**/*.{js,ts}',
```
ZachJW34, line 175: I was annoyed that making changes to `scaffold-config` never triggered the `Server update` toast so I'm including that watching here

packages/scaffold-config/src/frameworks.ts
```diff
@@ -61,6 +61,22 @@ function getBundlerDependency (bundler: WizardBundler['type'], projectPath: stri
   }
 }
 
+export const WIZARD_MOUNT_MODULES = ['cypress/react', 'cypress/react18', 'cypress/vue', 'cypress/vue2', 'cypress/angular'] as const
+
+export type WizardMountModule = typeof WIZARD_MOUNT_MODULES[number]
```
mike-plummer, line 66: Any reason we need this in array format? Or could we simplify to something like this?
```
export type WizardMountModule = 'cypress/react' | 'cypress/react18' | 'cypress/vue' | 'cypress/vue2' | 'cypress/angular'
```

packages/scaffold-config/src/frameworks.ts
```diff
@@ -61,6 +61,22 @@ function getBundlerDependency (bundler: WizardBundler['type'], projectPath: stri
...
+
+const mountModule = (mountModule: WizardMountModule) => async (projectRoot: string) => Promise.resolve(mountModule)
+
+const reactMountModule = async (projectPath: string) => {
+  const reactPkg = await isDependencyInstalled(dependencies.WIZARD_DEPENDENCY_REACT, projectPath)
+
+  try {
+    return semver.major(reactPkg.detectedVersion!) === 18 ? 'cypress/react18' : 'cypress/react'
+  } catch (e) {
+    return 'cypress/react'
```
mike-plummer, line 76: Worth dumping out the error here to `debug`?

packages/scaffold-config/src/frameworks.ts
```diff
@@ -61,6 +61,22 @@ function getBundlerDependency (bundler: WizardBundler['type'], projectPath: stri
   }
 }
 
+export const WIZARD_MOUNT_MODULES = ['cypress/react', 'cypress/react18', 'cypress/vue', 'cypress/vue2', 'cypress/angular'] as const
+
+export type WizardMountModule = typeof WIZARD_MOUNT_MODULES[number]
+
+const mountModule = (mountModule: WizardMountModule) => async (projectRoot: string) => Promise.resolve(mountModule)
+
+const reactMountModule = async (projectPath: string) => {
```
mike-plummer, line 70: /nitpick: Could we standardize these function signatures to both use `projectRoot` or `projectPath`?

mike-plummer (commented):
Tested it out with Vue, React17, and React18 - all working as expected. Just a couple small comments, will approve once those are answered/addressed

packages/scaffold-config/test/unit/detect.spec.ts
lmiller1990: Argh

packages/scaffold-config/test/unit/supportFile.spec.ts
```diff
@@ -1,170 +1,223 @@
...
+          // 'supportFile' configuration option.
+          //
+          // You can read more here:
+          // https://on.cypress.io/configuration
+          // ***********************************************************
+
+          // Import commands.js using ES2015 syntax:
+          import './commands'
+
+          // Alternatively you can use CommonJS syntax:
```
lmiller1990, line 30: Unrelated to this PR, but we should probably consider removing this line, eventually. I am not even sure if Vite 3 works with `require` out of the box (maybe it does, I haven't tried). We don't have any tests for it, though, so we probably don't want to recommend something unless we know it works 100% of the time.

packages/scaffold-config/src/frameworks.ts
lmiller1990: Weird, I guess it throws if the string we pass isn't valid. Instead of `!` should we just do a check using the `semver.valid` function? https://github.com/npm/node-semver#usage

```js
const vers = semver.valid(reactPkg.detectedVersion)

if (!vers) {
  throw Error(`Could not find valid version. Got ${reactPkg.version}`)
}

return semver.major(vers) === 18 ? ...
```

I like to error when things are unexpected, or at least log it - this seems better than `!`. You also won't need the `try/catch` anymore.

packages/scaffold-config/src/frameworks.ts
```diff
@@ -61,6 +61,22 @@ function getBundlerDependency (bundler: WizardBundler['type'], projectPath: stri
   }
 }
 
+export const WIZARD_MOUNT_MODULES = ['cypress/react', 'cypress/react18', 'cypress/vue', 'cypress/vue2', 'cypress/angular'] as const
+
+export type WizardMountModule = typeof WIZARD_MOUNT_MODULES[number]
+
```
lmiller1990, line 67: I think you can remove `async` here, implied by the `Promise.resolve`

packages/scaffold-config/src/frameworks.ts
```diff
@@ -61,6 +61,22 @@ function getBundlerDependency (bundler: WizardBundler['type'], projectPath: stri
...
+const mountModule = (mountModule: WizardMountModule) => async (projectRoot: string) => Promise.resolve(mountModule)
```
lmiller1990, line 68:
```suggestion
const mountModule = (mountModule: WizardMountModule) =>  (projectRoot: string) => Promise.resolve(mountModule)
```

I think you can remove `async` here, implied by the `Promise.resolve`

lmiller1990 (commented):
Some general questions, will drop an approval after the update (if needed).

packages/scaffold-config/src/frameworks.ts
ZachJW34: Addressed in [143a9d1](https://github.com/cypress-io/cypress/pull/23251/commits/143a9d171d0389a749eac7907842e2d59390a6cf)

I don't want to throw an error here, since we do offer the `skip` button on the Install screen. Also, throwing an error during the file scaffolding would cause some weird behavior, like writing 3/4 files to disk and having the user stuck in a weird spot. I'm falling back to `cypress/react` if the dep isn't installed.

packages/scaffold-config/src/frameworks.ts
ZachJW34: I believe I was planning to use the array in the test but ended up scrapping it. Good catch, addressed in [143a9d1](https://github.com/cypress-io/cypress/pull/23251/commits/143a9d171d0389a749eac7907842e2d59390a6cf)

packages/scaffold-config/src/frameworks.ts
ZachJW34: Went with `projectPath` to make it consistent: [143a9d1](https://github.com/cypress-io/cypress/pull/23251/commits/143a9d171d0389a749eac7907842e2d59390a6cf)

packages/scaffold-config/src/frameworks.ts
ZachJW34: Removed async in [143a9d1](https://github.com/cypress-io/cypress/pull/23251/commits/143a9d171d0389a749eac7907842e2d59390a6cf)

packages/scaffold-config/src/frameworks.ts
ZachJW34: Refactored to not throw an error by calling `semver.valid` first

packages/scaffold-config/test/unit/supportFile.spec.ts
ZachJW34: Created an issue: https://github.com/cypress-io/cypress/issues/23287

mike-plummer (approved):

lmiller1990 (approved):

lmiller1990:
Gave it another test, looks good!

Issue #22935, opened 2022-07-26
Dashboard Integration Smart Record Run Notification


pstakoun:
This is an info banner as per the [designs in Figma](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/V0lCq4uBplf9J3GZRiD02m/Cypress-App?node-id=12097%3A68131).

The banner should show the record command.
If the banner is dismissed, it should never be shown again.

The conditions for this banner to be visible are:
- The user is logged in
- The user has at least 1 organization
- The current project is connected (i.e. has valid projectId)
- The user has been using Cypress for at least 4 days (at least 96 hours since first open)
  - The automatic docs prompts should not pop up within 24 hours of the banners, the same way the docs do not pop up within 24 hours of each other (the banners get priority, so the docs prompts only pop up if no banners have been shown for 24 hours) - we will likely have to make this change in `shouldShowPrompt` in `HeaderBarContent.vue`
- There are no error, warning, or higher priority info banners present
- The banner has never been dismissed by the user

The copy for this banner should be the "A" variant described in [this doc](https://docs.google.com/document/d/1PFN79d3deDyqDVFmVIBe4cVMzWiEll0J6XCpesBZrSg/edit).

Please reference [this sheet](https://docs.google.com/spreadsheets/d/1_ejTVOv8zA9XAH69Eqa7bBrKqoQY__WeQ3Gqsk7fU0w/edit) for info about all relevant states.

mike-plummer:
@pstakoun Other "smart notification" issues don't reference the "only show if it's been 2 days" behavior - is that intended to be applied to those as well? Going to need a bit more info about how this applies - is the 2-day restriction based on showing a notification that "comes before" this one in the setup workflow, or does it also apply to itself (only show this prompt if it's been 2 days since we last showed it)? And I assume the dismiss behavior overrides the 2 day restriction so we never show it again regardless of time period?

pstakoun:
@mike-plummer I would say no, only this one should have the "2 days" behavior, but I've been debating that a bit. You are correct in assuming the dismiss behavior overrides the 2 day restriction. Anything else unclear here?

abelb:
[Figma Link](https://www.figma.com/file/V0lCq4uBplf9J3GZRiD02m/ACI-Fast-Follows?node-id=18093%3A153388)

![image](https://user-images.githubusercontent.com/108173/182229069-94ade710-3ee9-4278-843b-2c29a9cc93dd.png)

[...]

Issue #22934, opened 2022-07-26
Dashboard Integration Smart Organization Creation Notification


pstakoun:
This is an info banner as per the [designs in Figma](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/V0lCq4uBplf9J3GZRiD02m/Cypress-App?node-id=12097%3A68131).

The create organization CTA should send the user to `/organizations` in the Dashboard, the same way the "Manage organizations" link does. UTM parameters should be included.
If the banner is dismissed, it should never be shown again.

The conditions for this banner to be visible are:
- The user is not logged in
- The user has 0 affiliated organizations (i.e. 0 memberships)
- The user has been using Cypress for at least 4 days (at least 96 hours since first open)
  - The automatic docs prompts should not pop up within 24 hours of the banners, the same way the docs do not pop up within 24 hours of each other (the banners get priority, so the docs prompts only pop up if no banners have been shown for 24 hours) - we will likely have to make this change in `shouldShowPrompt` in `HeaderBarContent.vue`
- There are no error, warning, or higher priority info banners present
- The banner has never been dismissed by the user

The copy for this banner should be the "A" variant described in [this doc](https://docs.google.com/document/d/1PFN79d3deDyqDVFmVIBe4cVMzWiEll0J6XCpesBZrSg/edit).

Please reference [this sheet](https://docs.google.com/spreadsheets/d/1_ejTVOv8zA9XAH69Eqa7bBrKqoQY__WeQ3Gqsk7fU0w/edit) for info about all relevant states.

abelb:
https://app.clickup.com/t/364rtqz

baus:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/22934?planning-poker) with ZenHub @astone123 @lmiller1990 @marktnoonan @mike-plummer @rockindahizzy @warrensplayer @ZachJW34

abelb:
[Figma Link](https://www.figma.com/file/V0lCq4uBplf9J3GZRiD02m/ACI-Fast-Follows?node-id=18326%3A163203)

![image](https://user-images.githubusercontent.com/108173/182231526-aad42486-9c0a-4866-8b1c-f429c6415891.png)

mike-plummer:
Completed in #23256

Issue #22798, opened 2022-07-14
Dashboard Integration banner state for "no projectId"/"connect your project" state


pstakoun:
This is an info banner as per the [designs in Figma](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/V0lCq4uBplf9J3GZRiD02m/Cypress-App?node-id=12097%3A68131).

The connect project CTA should send the user to the "connect project modal".
If the banner is dismissed, it should never be shown again.

The conditions for this banner to be visible are:
- The user is not logged in
- The user has at least 1 organization
- The current project is not connected (i.e. no projectId)
- The user has been using Cypress for at least 4 days (at least 96 hours since first open)
  - The automatic docs prompts should not pop up within 24 hours of the banners, the same way the docs do not pop up within 24 hours of each other (the banners get priority, so the docs prompts only pop up if no banners have been shown for 24 hours) - we will likely have to make this change in `shouldShowPrompt` in `HeaderBarContent.vue`
- There are no error, warning, or higher priority info banners present
- The banner has never been dismissed by the user

The copy for this banner should be the "A" variant described in [this doc](https://docs.google.com/document/d/1PFN79d3deDyqDVFmVIBe4cVMzWiEll0J6XCpesBZrSg/edit).

Please reference [this sheet](https://docs.google.com/spreadsheets/d/1_ejTVOv8zA9XAH69Eqa7bBrKqoQY__WeQ3Gqsk7fU0w/edit) for info about all relevant states.

abelb:
https://app.clickup.com/t/364ru66

baus:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/22798?planning-poker) with ZenHub @astone123 @lmiller1990 @marktnoonan @mike-plummer @rockindahizzy @warrensplayer @ZachJW34

abelb:
[Figma Link](https://www.figma.com/file/V0lCq4uBplf9J3GZRiD02m/ACI-Fast-Follows?node-id=18091%3A152804)

![image](https://user-images.githubusercontent.com/108173/182229903-9f5e706a-5593-4476-940b-09e4ec587405.png)

marktnoonan:
This one is hard to estimate by itself, it really forms part of a larger system of managing what we show when:

> There are no error, warning, or higher priority info banners present

[...]

Issue #23072, opened 2022-08-02
Banner Framework


baus:
Requirements for banners here.

A/B testing
Closed state
Analytics
etc.

PR #23256, opened 2022-08-10
feat: ACI Notification Banners

### LINKED ISSUE #22933: Dashboard Integration Smart Login Notification
This is an info banner as per the [designs in Figma](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/V0lCq4uBplf9J3GZRiD02m/Cypress-App?node-id=12097%3A68131).

The login CTA should send the user through the login flow. UTM parameters should be included.
If the banner is dismissed, it should never be shown again.

The conditions for this banner to be visible are:
- The user is not logged in
- The user has been using Cypress for at least 4 days (at least 96 hours since first open)
  - The automatic docs prompts should not pop up within 24 hours of the banners, the same way the docs do not pop up within 24 hours of each other (the banners get priority, so the docs prompts only pop up if no banners have been shown for 24 hours) - we will likely have to make this change in `shouldShowPrompt` in `HeaderBarContent.vue`
- There are no error, warning, or higher priority info banners present
- The banner has never been dismissed by the user

The copy for this banner should be the "A" variant described in [this doc](https://docs.google.com/document/d/1PFN79d3deDyqDVFmVIBe4cVMzWiEll0J6XCpesBZrSg/edit).

Please reference [this sheet](https://docs.google.com/spreadsheets/d/1_ejTVOv8zA9XAH69Eqa7bBrKqoQY__WeQ3Gqsk7fU0w/edit) for info about all relevant states.

mike-plummer:
- Closes 
#22933 
#22934 
#22798 
#22935 

### User facing changelog
* Add new banners in the app to help get your project connected to the dashboard and complete a first run recording

### Additional details
Four main areas here:
1. Adding new section of project preferences to track the "last shown" and "dismissed" timestamps for each banner being added here
2. Create new `TrackedBanner` component to allow reuse of storing "last shown" and "dismissed" info for banners
3. Add four new "Smart Notifications" that trigger based on users progress towards connecting their project to the dashboard
	* Login
	    * Has been using Cypress for at least 4 days
	    * Is **_not_** logged in
	    * Not previously dismissed in this project
	* Create Organization
	    * Has been using Cypress for at least 4 days
	    * Is logged in
	    * Is **_not_** a member of any cloud organizations
	    * Not previously dismissed in this project
	* Connect Project
	    * Has been using Cypress for at least 4 days
	    * Is logged in
	    * Is member of at least one cloud organization
	    * Project is **_not_** successfully connected to dashboard
	    * Not previously dismissed in this project 
	* Record First Run 	
	    * Has been using Cypress for at least 4 days
	    * Is logged in
	    * Is member of at least one cloud organization
	    * Project is successfully connected to dashboard
	    * Project does **_not_** have any cloud recording runs
	    * Not previously dismissed in this project
4. Connect new banners to SpecsList in the `SpecsListBanners` component - this is currently serving as a "banner controller" of sorts to control which banners to display and when (a refinement of this construct is coming in future work related to A/B testing of banner content)

### Steps to test
**Note: To fully test it will be necessary to either create a new user or have an existing user removed from all organizations. You may want to create a new "throwaway" user rather than remove yourself from the shared Cypress org in staging.**

**Note: launch with CYPRESS_INTERNAL_ENV=staging for staging**.

1. Create a new empty Cypress project (name it something distinctive) with at least one spec (content is unimportant)
2. Open the project in Cypress. Ensure you are logged out. Validate that "Smart Login Banner" does not display.
3. Close the project and find its `state.json` file. On Mac this is stored in a project-specific directory under `~/Library/Application Support/Cypress/cy/staging/projects/`
4. Modify `state.json` to have a `firstOpened` value more than 4 days in the past: `Date.now() - 8.64e+7 * 4`
5. Reopen Cypress, validate the "Smart Login Banner" now displays
6. Dismiss the banner by clicking the "X" button. Refresh app and validate banner does not display
7. Reopen `state.json`, observe an entry for the login banner has been created. Remove it, and reopen the project.
```
    "aci_082022_login": {
      "lastShown": 1660075007022,
      "dismissed": 1660075012749
    },
```
9. Observe login banner once again displays. Use banner to execute login flow, validate it succeeds.
   * Note: Next three prompts use same logic for respecting dismissal
11. "Smart Create Organization Banner" should now display (if your user has been removed from all cloud orgs). Use banner to create an organization, validate UI updates to no longer show create organization banner
12. "Smart Connect Project Banner" should now display. Use banner to create new cloud project, validate UI updates to no longer show connect project banner
13. "Smart Record Banner" should now display. Validate record command looks correct (should include `--component` flag for CT, valid record key) and can be copy-pasted
14. _Optional:_ Record a cloud run, validate that "Smart Record Banner" does not display on next app launch

**Login Banner**

https://user-images.githubusercontent.com/11482842/183981545-0f78a87c-6ad9-4c9f-a8a0-4e2becd2a1be.mov

**Create Org Banner**

https://user-images.githubusercontent.com/11482842/183981573-78b599bd-f2a9-48b8-b686-1c57930742e7.mov

**Connect Project Banner**

https://user-images.githubusercontent.com/11482842/183981601-0dade474-003a-4838-afdf-77aae5fca8ed.mov

**Record Banner**

https://user-images.githubusercontent.com/11482842/183981630-8233f4a1-a229-4890-8405-591b30d8d895.mov

### PR Tasks


packages/app/src/specs/SpecsListBanners.vue
```diff
@@ -96,7 +96,7 @@
     </Button>
   </Alert>
   <Alert
-    v-if="showProjectRequestAccess"
+    v-else-if="showProjectRequestAccess"
```
lmiller1990, line 99: I think many v-else-if combination is fine, but one thing I started using in my personal projects lately is `v-switch`, I wonder if it would be useful in Cypress, too: https://www.npmjs.com/package/@lmiller1990/v-switch

The source code is only about 10 lines, we could just copy-paste and register as a global component.

packages/app/src/specs/SpecsListBanners.vue
```diff
@@ -175,4 +237,41 @@ watch(
...
+ * once login state resolves
+ */
+debouncedWatch(
+  cloudData,
+  ([cloudViewer, currentProject]) => {
+    const isLoggedIn = !!cloudViewer?.id
+    // Need to be able to tell whether the lack of `firstOrganization` means they don't have an org or whether it just hasn't loaded yet
+    // Not having this check can cause a brief flicker of the 'Create Org' banner while org data is loading
+    const isOrganizationLoaded = !!cloudViewer?.firstOrganization
+    const isMemberOfOrganization = (cloudViewer?.firstOrganization?.nodes?.length ?? 0) > 0
```
lmiller1990, line 255: Just to keep the thoughts around making GraphQL more readable front-and-center of mind ... I think we might want to start exploring some kind of pattern to model this logic, hopefully independently of the component that consumes it. 

I wonder if we can rip this logic out and put it into a separate model that we could potentially test in isolation? It's pretty hard to grok the component when there's all this complexity sitting inside a giant `watch`. Some ideas that come to mind for what might be easier to understand in the future:

```js
const { showRecordBanner } = useBanners(props.gql)
// or
const bannerModel = createBannerModel(props.gql.currentProject)

watch(bannerModel, newVal => {
  showRecordBanner.value = newVal.showRecordBanner
})
```

When I did React/Redux, this separation was more simple with `connect` used to separate container/components. With the move towards hooks/composables, it's not so clear anymore.

Just sparking some discussion - what does you think? If we are happy with the "put it all in the component" pattern, that's fine, too. Commenting primarily to ensure we keep thinking about this as we learn more about how best to utilise GraphQL as we more deeply integrate the Cloud and App, since I do see this getting out of hand eventually.

packages/app/src/specs/banners/ConnectProjectBanner.vue
```diff
@@ -0,0 +1,76 @@
...
+      @click="handleButtonClick"
+    >
+      {{ t('specPage.banners.connectProject.buttonLabel') }}
+    </Button>
+
+    <CloudConnectModals
+      v-if="isProjectConnectOpen && cloudModalsQuery.data.value"
+      :gql="cloudModalsQuery.data.value"
+      @cancel="handleModalClose"
+      @success="handleModalClose"
```
lmiller1990, line 30: Nice, I really like the `handleXXX` or `onXXX` for event handles. Not sure if we can lint for this, but I am a fan of this convention.

packages/app/src/specs/banners/ConnectProjectBanner.vue
```diff
@@ -0,0 +1,76 @@
...
+
+gql`
+query ConnectProjectBanner {
+  ...CloudConnectModals
+}
+`
+
+withDefaults(defineProps<{
+  modelValue: boolean
+}>(), {})
```
lmiller1990, line 54: Should this have a default true/false?

Or do we even need `withDefaults`?

packages/app/src/specs/banners/RecordBanner.vue
```diff
@@ -0,0 +1,78 @@
...
+const emit = defineEmits<{
+  (e: 'update:modelValue', value: boolean): void
+}>()
+
+const { t } = useI18n()
+
+const query = useQuery({ query: RecordBannerDocument })
+
+const firstRecordKey = computed(() => {
+  return query.data?.value?.currentProject?.cloudProject?.__typename === 'CloudProject' && query.data.value.currentProject.cloudProject.recordKeys?.[0]?.key
```
lmiller1990, line 68:
```suggestion
  return query.data?.value?.currentProject?.cloudProject?.__typename === 'CloudProject' && query.data.value.currentProject.cloudProject.recordKeys?.[0]?.key ?? '<record-key>'
```

Maybe, unless I missed something (or you think the ternary is more readable).

packages/app/src/specs/banners/TrackedBanner.vue
```diff
@@ -0,0 +1,79 @@
...
+
+function handleBannerDismissed (visible: boolean) {
+  if (!visible) {
+    updateBannerState('dismissed')
+  }
+
+  emit('update:modelValue', visible)
+}
+
+async function updateBannerState (field: 'lastShown' | 'dismissed') {
```
lmiller1990, line 71: Seems we are not awaiting this anyway, so can drop `async` and just return the final value instead of `await`

packages/app/src/specs/banners/TrackedBanner.vue
```diff
@@ -0,0 +1,79 @@
...
+`
+
+const props = withDefaults(defineProps<TrackedBannerComponentProps>(), {})
+
+const emit = defineEmits<TrackedBannerComponentEmits>()
+
+const stateQuery = useQuery({ query: TrackedBanner_ProjectStateDocument })
+const setStateMutation = useMutation(TrackedBanner_SetProjectStateDocument)
+
+watch(
```
lmiller1990, line 53: I think we can use `watchEffect`, looks similar to `watch` but `immediate: true` by default. https://vuejs.org/api/reactivity-core.html#watcheffect

Interestingly enough there is also `watchPostEffect` and `watchEffectSync` which I haven't used before. Good to know about, I think it's like React's `flushSync`, [docs](https://reactjs.org/docs/react-dom.html#flushsync).

```js
watchEffect(() => {
  if (props.modelValue) {
    updateBannerState('lastShown')
   }
})
```

packages/app/src/specs/banners/index.ts
```diff
@@ -0,0 +1,10 @@
+export const BannerIds = {
+  ACI_082022_LOGIN: 'aci_082022_login',
+  ACI_082022_CREATE_ORG: 'aci_082022_createOrganization',
+  ACI_082022_CONNECT_PROJECT: 'aci_082022_connectProject',
+  ACI_082022_RECORD: 'aci_082022_record',
+}
```
lmiller1990, line 6: Can mark as `as const` for (potentially) better type inference.

packages/app/src/specs/banners/index.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export { default as LoginBanner } from './LoginBanner.vue'
+
+export { default as CreateOrganizationBanner } from './CreateOrganizationBanner.vue'
```
lmiller1990, line 10: I also really like this pattern where we have one directory of something (eg component, concern, whatever) and mark the "public" API by an `index.ts` that exports things. I have been [trying it in my projects](https://github.com/lmiller1990/rhythm/tree/main/packages/frontend/src/screens/gameplay/components/Gameplay) and it's going well... I think we do want to start to explore more rigid architectural patterns (including source files and directories) as we grow the project.

Any thoughts on this pattern?

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -245,7 +246,7 @@ const userData = computed(() => {
 })
 
 const savedState = computed(() => {
-  return props.gql?.currentProject?.savedState
+  return props.gql?.currentProject?.savedState as AllowedState
```
lmiller1990, line 249: Curious why we cannot get the types via GraphQL (didn't look too closely)

packages/types/src/config.ts
```diff
@@ -45,3 +45,12 @@ export interface SettingsOptions {
   testingType?: 'component' |'e2e'
   args?: AllModeOptions
 }
+
+export type BannerState = {
+  lastShown?: number | null
```
lmiller1990, line 50: Should we just drop the `?`? Or, make the type `| null` - otherwise `undefined` and `null` are both okay, maybe it's good to have one way to show false-y-ness.

Could be more to it, not a big deal, just call this out whenever I see it.

packages/types/src/config.ts
```diff
@@ -45,3 +45,12 @@ export interface SettingsOptions {
...
+  dismissed?: number | null
+}
+
+export type BannersState = {
+  [bannerId: string]: BannerState | null
```
lmiller1990, line 55: `bannerId` is always known, right? I wonder if we can type this - it'll be very hard to understand what banners are available down the track.

If you make `BannerIds` like this

```ts
export const BannerIds = {
  ACI_082022_LOGIN: 'aci_082022_login',
  ACI_082022_CREATE_ORG: 'aci_082022_createOrganization',
  ACI_082022_CONNECT_PROJECT: 'aci_082022_connectProject',
  ACI_082022_RECORD: 'aci_082022_record',
} as const
```

(note `as const`) I think we should be able to type this more strongly (not just `string`).

Bit of a nitpick but having things typed properly will probably help when we have a lot more banners in the future and we've all forgotten the different types that exist, etc.

lmiller1990 (commented):
Did a basic review and ask a few questions, sparked some conversation... I will give it a proper test now and report back.

lmiller1990:
> Create a new empty Cypress project (name it something distinctive) with at least one spec (content is unimportant)

Done

> Open the project in Cypress. Ensure you are logged out. Validate that "Smart Login Banner" does not display.

image

> Close the project and find its state.json file. On Mac this is stored in a project-specific directory under ~/Library/Application Support/Cypress/cy/staging/projects/

Here it is:

```
{
  "firstOpened": 1660256845067,
  "lastOpened": 1660269565222,
  "lastProjectId": null,
  "specFilter": ""
}
```

> Modify state.json to have a firstOpened value more than 4 days in the past: Date.now() - 8.64e+7 * 4

```
{
  "firstOpened": 1659924157665,
  "lastOpened": 1660269565222,
  "lastProjectId": null,
  "specFilter": ""
}
```

> Reopen Cypress, validate the "Smart Login Banner" now displays

The docs dropdown also showed, not sure if this was a bug or my moue was just in the wrong place and the wrong time?

image

It then disappeared (apparently on it's own, I think). It looks like there is a `banners.ci1` added to my `state.json`. I wonder if your ACI feature is also timed similar to another prompt? 

Anyway, the correct banner (also) showed up. The banner didn't show up **instantly** - that said I am

1. on staging (it is rate limited?)
2. generally slow internet 
3. in Australia

So the request took a bit of time. I wonder if we should be doing the fetching on the server so it doesn't have that "janky" feeling and cause a repaint.

Either way, it did show after around 1.5s. Here it is:

image

> Dismiss the banner by clicking the "X" button. Refresh app and validate banner does not display

It did not come back.

> Reopen state.json, observe an entry for the login banner has been created. 

It was. I also got `promptsShown.ci1`.

```
{
  "firstOpened": 1659924157665,
  "lastOpened": 1660269841047,
  "lastProjectId": null,
  "specFilter": "",
  "promptsShown": {
    "ci1": 1660269851159
  },
  "banners": {
    "aci_082022_login": {
      "lastShown": 1660269851539,
      "dismissed": 1660269961623
    }
  }
}
```

> Remove it, and reopen the project. Observe login banner once again displays. 

It's back.

> Use banner to execute login flow, validate it succeeds.

I actually got stuck on 
image

Because I **created a new account**. I don't think that's a bug introduced here - we probably don't handle the "new account" workflow, only the "login with existing account".

Edit: I'm stuck, I cannot seem to log in. I'll see if this is only on my machine, or something else. I've tried both `staging` and `production` for `CYPRESS_INTERNAL_ENV`.

TODO: The rest...

Note: Next three prompts use same logic for respecting dismissal
"Smart Create Organization Banner" should now display (if your user has been removed from all cloud orgs). Use banner to create an organization, validate UI updates to no longer show create organization banner
"Smart Connect Project Banner" should now display. Use banner to create new cloud project, validate UI updates to no longer show connect project banner
"Smart Record Banner" should now display. Validate record command looks correct (should include --component flag for CT, valid record key) and can be copy-pasted
Optional: Record a cloud run, validate that "Smart Record Banner" does not display on next app launch

lmiller1990:
Recording:

Notice I needed to restart after creating and org - it didn't immediately prompt me to add my project. Also it looks like there was some stale data towards the end after I'd recorded - the record banner showed up briefly, before disappearing.


https://user-images.githubusercontent.com/19196536/184281787-97cbf610-42eb-44e1-88da-d1f68124f155.mp4

packages/app/src/specs/SpecsListBanners.vue
mike-plummer: Absolutely agree, I don't feel this is elegant or a long-term solution by any means. The reason I went this way is that there is an upcoming epic (#23072) to build a framework for user messaging in which a more holistic pattern for all our banners will be constructed. It will handle hooking up to state, managing preconditions, A/B testing, etc

packages/app/src/specs/banners/index.ts
mike-plummer: Yes! I started building a barrel file here then apparently got lost halfway through. If nothing else, they really help shorten up import statements. Will finish this one and fix the imports

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
mike-plummer: The field is currently defined this way, so it appears to be deliberate

```
t.json('savedState', {
      description: 'Project saved state',
        return ctx.project.getCurrentProjectSavedState()
      },
    })
```

Perhaps this is a safety measure since theoretically the content on disk can be manually mutated in ways that cause a schema mismatch?

packages/types/src/config.ts
mike-plummer: Hmm, I'll give this some more thought. My original thought was to mimic the GQL way of "anything can be undefined or null by default", but since we aren't directly querying this in GQL (just the encapsulating JSON object) I guess that isn't set in stone

packages/types/src/config.ts
mike-plummer: 🤔 Good callout. I probably can't directly use the referenced `BannerIds` struct here since that would add a dependency on `@cypress/app` to the `@cypress/types` package, but I'll see about adding a new `BannerId` union type that can be used both here and on the struct

mike-plummer:
@lmiller1990 Thank you for the thorough testing & review! Sorry for the miss on the "login without org", that was an oversight in the procedure on my part - glad you were able to get it sorted, hope it didn't waste too much of your time

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -305,8 +306,9 @@ function shouldShowPrompt (prompt: { slug: string, noProjectId: boolean, interva
   }
 
   const now = Date.now()
-  const timeSinceOpened = now - savedState.value?.firstOpened
+  const timeSinceOpened = now - (savedState.value?.firstOpened ?? now)
   const allPromptShownTimes: number[] = Object.values(savedState.value?.promptsShown ?? {})
+  const bannersLastShown = Object.values(savedState.value?.banners ?? {}).map((banner) => banner?.lastShown).filter((val): val is number => !!val)
```
warrensplayer, line 311: This would be for another issue, but it would be nice to move this "shouldShowPrompt" logic out of the HeaderBarContent component. I know the requirements for this issue call for this additional dependency, but it seems like a strange coupling in the code.

packages/app/src/specs/SpecsListBanners.vue
```diff
@@ -16,7 +16,7 @@
     <p>{{ t('specPage.noSpecError.explainer') }}</p>
   </Alert>
   <Alert
-    v-if="showOffline"
+    v-else-if="showOffline"
```
warrensplayer, line 19: I know the new banners have mutually exclusive logic, but how about adding a test to confirm the offline or fetch error banner take precedence and prevent the new banners from showing?

warrensplayer (commented):
I am still figuring out the code patterns, etc. so just adding a few comments.  I would like to look to see if there are ways to make the testing of this easier when coming back to address the A/B testing addition, but not sure if that is possible.  I was able to get all banners to appear, but not sure if I was able to exercise all possible scenarios depending on how a project was created, etc.  Overall, great job and getting these different states to behave.  I liked your use of functions to encapsulate the repeating parts of the tests for each banner in the SpecsListBanners.cy.tsx

warrensplayer:
@marktnoonan If you have a chance, can you look at the parts of this PR as it relates to saving the preferences since you have been in that part of the code recently?

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
mike-plummer: Agreed; my thoughts are that these cross-cutting concerns would get bundled up as part of the Banner Framework so that a shared composable could be used. Didn't want to take that on in this work since we hadn't done any planning for what that framework needed to look like

packages/app/src/specs/SpecsListBanners.vue
lmiller1990: Sounds good, let's deal with it there.

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
lmiller1990: Not sure, maybe you cannot type `t.json` (limitation).

We can find out more at some point, had a quick look but seems like it's not a simple "just add the types" change and this PR is getting fairly beefy.

packages/types/src/config.ts
lmiller1990: Right, I see... I guess that makes sense, `null` means "It's not defined" and `undefined` means "it's not here since you didn't request it" - similar to JSON.

That said, not sure we want GraphQL leaking into places like `packages/types`, wonder if it makes sense to do `lastShown?: number`, then in the GraphQL resolver, coerce using `?? null` (since the `null` is a GraphQL concern).

Not a big deal, I really like to keep concerns nice and compartmentalized but since we can't realistically lint for this it's not exactly practical to do, even if we did agree on this pattern.

packages/types/src/config.ts
lmiller1990: Hmm should we just move the `BannerIds` to `@packages/types`? I know it feels kind of bad to just dump whatever we need in `packages/types` but I don't find it so bad once you get over the fact it's sort of a melting pot for anything and everything. Then `packages/app` can depend on `packages/types` - not a big deal.

lmiller1990 (approved):
Still a couple of outstanding comments/thoughts from my review but I wouldn't call them blockers on shipping... going to approve this.

I'd still like to see a response + resolution on the remaining discussions @mike-plummer :pray:

packages/types/src/config.ts
mike-plummer: I like it - I hadn't considered moving the `const` over to the typings. Done in [`da3c2e3`](https://github.com/cypress-io/cypress/pull/23256/commits/da3c2e3be43d84ccb4816bc2a08b341c394462d3)

packages/types/src/config.ts
mike-plummer: You've convinced me 😁 . Done in [`da3c2e3`](https://github.com/cypress-io/cypress/pull/23256/commits/da3c2e3be43d84ccb4816bc2a08b341c394462d3)

packages/app/src/specs/SpecsListBanners.vue
mike-plummer: Added tests to `SpecsListBanners.cy.tsx` to validate that each smart notification is preempted by each of the 5 higher-priority banners [`da3c2e3`](https://github.com/cypress-io/cypress/pull/23256/commits/da3c2e3be43d84ccb4816bc2a08b341c394462d3)

marktnoonan:
Still reviewing code and tests but wanted to share that functionality all works for me! 

@pstakoun Product feedback - I wasn't quite expecting the "guided tour" feeling where if I don't dismiss the banner, its content changes on the next step - eg, "Login" content -> "Connect" content -> "Record" content. UX wise the internal content of a banner changing like this feels like it could use some sort of "journey" UI ("Step 1 of 3") or some element that visibly "progresses" as well as simply the text content being updated. I think the "guided" feeling is good & leaning into it a bit could be a future enhancement.

pstakoun:
@marktnoonan totally agree with you and that feedback resonates with me. I think banners are not the form-factor that this experience will be in the near future - an onboarding checklist similar to that in the Dashboard is also under consideration.

marktnoonan:
Screen Shot 2022-08-16 at 10 46 09 AM

We have some overflow issues anyway on this page but maybe we could still confine this messaging to max size of "viewport width minus nav size" or something?

astone123 (approved):
I tested each of the banner situations with staging. Each one worked correctly for me. One thing was that in order to get into the state where I was logged in but wasn't in an org, I had to delete my existing org. Not sure how a user would get into this state since I think you need to be in an org to complete the login flow, but good that we're handling that either way.

Nice work 🎉 I'm excited to see these banners get more styled in the future!

packages/app/src/specs/SpecsListBanners.cy.tsx
```diff
@@ -48,6 +55,75 @@ describe('<SpecsListBanners />', () => {
...
+              dismissed: Date.now(),
+            },
+          },
+        })
+
+        cy.get(`[data-cy="${bannerTestId}"]`).should('not.exist')
+      })
+    })
+
+    it('should not render when using cypress < 4 days', () => {
```
marktnoonan, line 120: Wonder if we should we wrap this in its own `context`, or move it to the top. It ends up jumping to the top in the UI anyway, which was just a little confusing when reviewing the tests.

Screen Shot 2022-08-16 at 11 50 46 AM

mike-plummer:
> We have some overflow issues anyway on this page but maybe we could still confine this messaging to max size of "viewport width minus nav size" or something?

@marktnoonan I don't think we should add custom styling for these at this time. From a UX perspective the app has a hard minimum width, and I think it would look kinda funky for parts of the UI to try to be responsive while the majority overflows. I think we *should* be more responsive, but it should be tackled holistically so there is a consistent experience

packages/app/src/specs/SpecsListBanners.vue
marktnoonan: good tests!

packages/app/src/specs/SpecsListBanners.vue
```diff
@@ -30,7 +30,7 @@
     </p>
   </Alert>
   <Alert
-    v-if="showFetchError"
+    v-else-if="showFetchError"
     v-model="showFetchError"
```
marktnoonan, line 34: Hmm. The combination of `v-if` and `v-model` pointing to the same value seems like a hint that `v-model` is not doing what we want and hiding/showing the Alert on its own?

packages/app/src/specs/SpecsListBanners.vue
```diff
@@ -110,6 +110,22 @@
...
+    v-else-if="showConnectBanner"
+    v-model="showConnectBanner"
+  />
+  <CreateOrganizationBanner
+    v-else-if="showCreateOrganizationBanner"
+    v-model="showCreateOrganizationBanner"
+  />
+  <LoginBanner
+    v-else-if="showLoginBanner"
+    v-model="showLoginBanner"
```
marktnoonan, line 127: Similar comment here.  Seems like we could push the `if` inside these components (or use a wrapper) and drive the presence/absence of these things from the state being tracked via the `v-model`.

packages/app/src/specs/banners/CreateOrganizationBanner.vue
```diff
@@ -0,0 +1,79 @@
...
+      {{ t('specPage.banners.createOrganization.content') }}
+    </p>
+
+    <ExternalLink
+      class="border rounded mx-auto outline-none bg-indigo-500 border-indigo-500 text-white max-h-60px py-6px px-16px inline-block hocus-default hocus:no-underline"
+      :href="createOrganizationUrl"
+      :include-graphql-port="true"
+      data-cy="create-organization-button"
+    >
+      <i-cy-office-building_x16 class="inline-block icon-dark-white" />
```
marktnoonan, line 23: The alignment of this icon could use a little tweaking - looks too close to text text and a bit too low. Though we may not need to deal with the styling here. Our `Button` component knows how to render an `ExternalLink` styled like a button. It might even properly bind the `include-graphql-port` prop out of the box, would need to experiment a bit. 

Screen Shot 2022-08-16 at 12 23 21 PM

packages/app/src/specs/SpecsListBanners.vue
mike-plummer: This was an attempt to simplify the "only show one banner" concern within `SpecsListBanners.vue`. Currently, `showFetchError` just cares about whether the conditions for its display are met independently, but if we remove the `v-if/v-else-if` then I would need to add conditions to check whether any higher-priority banners are shown when determining the model value. This felt like a cleaner approach

packages/app/src/specs/SpecsListBanners.cy.tsx
mike-plummer: Moved tests to better align with execution order

packages/app/src/specs/SpecsListBanners.vue
marktnoonan: Yeah that makes sense, at least for this iteration. I know more is coming 👍

packages/app/src/specs/banners/CreateOrganizationBanner.cy.tsx
```diff
@@ -0,0 +1,21 @@
...
+      __typename: 'CloudUser',
+      id: 'test123',
+      cloudOrganizationsUrl: linkHref,
+    } as any
+
+    cy.mount({ render: () => <CreateOrganizationBanner modelValue={true} /> })
+
+    cy.get('a')
+    .should('have.attr', 'href')
+    .and('contain', linkHref)
```
marktnoonan, line 17: Should we assert the `specPage.banners.Join or create an organization in Cypress Dashboard to access your projects and recorded test runs` here? Since grabbing that content is also a responsibility of this component.

packages/app/src/specs/banners/LoginBanner.cy.tsx
```diff
@@ -0,0 +1,9 @@
+import LoginBanner from './LoginBanner.vue'
+
+describe('<LoginBanner />', () => {
+  it('should render expected content', () => {
+    cy.mount({ render: () => <LoginBanner modelValue={true} /> })
+
+    cy.percySnapshot()
```
marktnoonan, line 7: Similar to an earlier comment, I'd like this test to fail if the content were missing. Percy will catch this too but is a lot slower/sometimes gets ignored.

Or maybe we have an e2e test or higher level component test that checks the contents, and I haven't see it yet.

packages/frontend-shared/src/gql-components/HeaderBarContent.cy.tsx
```diff
@@ -378,6 +378,10 @@ describe('<HeaderBarContent />', { viewportWidth: 1000, viewportHeight: 750 }, (
             },
             render: (gqlVal) => <div class="border-current border-1 h-700px resize overflow-auto"><HeaderBarContent gql={gqlVal} show-browsers={true} allowAutomaticPromptOpen={true} /></div>,
           })
+
+          cy.tick(2000)
```
marktnoonan, line 382: Could we add a comment with the intention behind this tick? Would tests fail if we removed it?

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -299,7 +300,20 @@ const prompts = sortBy([
   },
 ], 'interval')
 const isForceOpenAllowed = ref(true)
+const isOpenDelayElapsed = ref(false)
+
+onMounted(() => {
+  setTimeout(() => isOpenDelayElapsed.value = true, 2000)
```
marktnoonan, line 306: nice, this makes sense of the `tick` in the test code for mounting.

marktnoonan: Sidenote: we could simplify a little and remove the mounted hook updating the ref with `useTimeout` (https://vueuse.org/shared/usetimeout/)

marktnoonan (approved):

PR #23257, opened 2022-08-10
feat: add packherd to monorepo

### LINKED ISSUE #22979: Move Packherd to the Monorepo
### What would you like?

As part of the v8 snapshots feature, we need access to a new package called `packherd` which was created by @thlorenz partly to support the v8 snapshot functionality. In order to utilize it in cypress, I want to move `packherd` from being a [standalone project](https://github.com/cypress-io/packherd) to the monorepo.  

### Why is this needed?

By moving this project to the monorepo, it will allow for faster integration, more consistent project structure, and no new release effort.

ryanthemanuel:
- Closes #22979

### User facing changelog

n/a

### Additional details

As part of the v8 snapshots feature, we need access to a new package called `packherd` which was created previously partly to support the v8 snapshot functionality. In order to utilize it in the Cypress app, I want to move packherd from being a [standalone project](https://github.com/cypress-io/packherd) to the monorepo. The code has mostly stayed the same as what written originally. I just cleaned up a few things and standardized it to be in line with the monorepo.

### Steps to test

For now this package is completely self contained. Future PRs will incorporate this into V8 snapshots. I would recommend checking out the `ReadMe` and documentation in the code to follow what is happening with it.

n/a

### PR Tasks


rachelruderman:
How would you recommend we go about reviewing? Perhaps a good opp for a group code review party?

ryanthemanuel:
> How would you recommend we go about reviewing? Perhaps a good opp for a group code review party?

@rachelruderman I do realize there's a lot here. The best place to start is probably the ReadMe. Thorsten did a good job of describing what all is done in this package. Each file itself also has pretty good doc with a step by step description of how things are being looked up, etc.

packages/packherd/README.md
```diff
@@ -0,0 +1,140 @@
+## @packages/packherd
+
+Herds all dependencies reachable from an entry and packs them.
```
BlueWinds, line 3: Reading through this readme, I kept thinking: "isn't this a description of vite/webpack"?

Vite in particular already uses esbuild as a backend, and it feels like it should do most of what packherd is doing. Can you walk me through the differences between packherd and vite? What features do we need that vite doesn't provide, and have we considered using vite to provide much of this functionality out-of-the-box and reduce our maintenance burden?

ryanthemanuel: There are a few things that are provided by `packherd` that are fairly unique to what we're doing here (or going to be doing once `v8-snapshot` gets moved over as well).

When generating the bundle, we need a way to swap out how the bundle is actually created (https://github.com/cypress-io/cypress/blob/bb89994af7e119236bd263234294d325439abce2/packages/packherd/src/packherd.ts#L25). In the v8 snapshot case, this will utilize `snapbuild` which is our custom version of esbuild that returns warnings that allow us to use our [snapshot doctor](https://github.com/cypress-io/tech-briefs/tree/main/briefs/startup-performance#alternatives-to-existing-tooling) to properly create the snapshot bundle.

In addition to generating a bundle, the other big part of `packherd is the ability to provide the hooks that we need to load our snapshotted exports and definitions (described here: https://github.com/cypress-io/cypress/tree/ryanm/feat/add-packherd-to-monorepo/packages/packherd#loading-bundledsnapshotted-modules-with-packherd).

circle.yml
```diff
@@ -1317,7 +1352,7 @@ jobs:
             # run type checking for each individual package
             - run: yarn lerna run types
             - verify-mocha-results:
-                expectedResultCount: 10
+                expectedResultCount: 14
```
ryanthemanuel, line 1355: We weren't verifying on all of the mocha unit tests previously. This takes care of the ones we weren't verifying + the new package.

lmiller1990:
Thanks for this - I'll try to get this reviewed in the next 24 hours. Also https://github.com/cypress-io/tech-briefs/tree/main/briefs/startup-performance might be useful for reviewing.

packages/server/package.json
```diff
@@ -98,6 +98,7 @@
     "ospath": "1.2.2",
     "p-defer": "^3.0.0",
     "p-queue": "6.1.0",
+    "performance-now": "^2.1.0",
```
flotwig, line 101: Is this used anywhere? Node has `performance.now()` :thinking:

flotwig (commented):
Looks good, couple questions

packages/packherd-require/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "dependencies": {
+    "convert-source-map": "^1.7.0",
+    "debug": "^4.3.1",
+    "esbuild": "^0.15.3",
+    "source-map-js": "^0.6.2"
+  },
+  "devDependencies": {
+    "@packages/ts": "0.0.0-development",
+    "mocha": "7.0.1",
+    "spok": "^1.3.0"
```
tbiethman, line 27: Doesn't look like `spok` is used by `packherd-require` after the split.

tooling/packherd/test/packherd.spec.ts
```diff
@@ -0,0 +1,92 @@
+import {
+  CreateBundle,
+  CreateBundleOpts,
+  CreateBundleResult,
+  CreateBundleOutputFile,
+  packherd,
+} from '../src/packherd'
+import { expect } from 'chai'
+import spok from 'spok'
```
tbiethman, line 9: Any thoughts on not introducing a `spok` dependency here? The assertions here might have more boilerplate without it, but there's not that many of them.

CONTRIBUTING.md
```diff
@@ -252,32 +252,47 @@ Cypress is a large open source project. When you want to contribute to Cypress,
...
 
 Here is a list of the core packages in this repository with a short description, located within the [`packages`](./packages) directory:
 
  | Folder Name                           | Package Name            | Purpose                                                                      |
  | :------------------------------------ | :---------------------- | :--------------------------------------------------------------------------- |
  | [cli](./cli)                          | `cypress`               | The command-line tool that is packaged as an `npm` module.                   |
+ | [app](./packages/app)                 | `@packages/app`         | The front-end for the Cypress App                                            |
```
tbiethman, line 262: Thanks for getting this up to date

system-tests/lib/dep-installer/index.ts
```diff
@@ -148,7 +148,7 @@ async function makeWorkspacePackagesAbsolute (pathToPkgJson: string): Promise<st
  * specified in the project's `package.json`. No-op if no `package.json` is found.
  * Will use `yarn` or `npm` based on the lockfile present.
  */
-export async function scaffoldProjectNodeModules (project: string, updateLockFile: boolean = !!process.env.UPDATE_LOCK_FILE): Promise<void> {
+export async function scaffoldProjectNodeModules (project: string, updateLockFile: boolean = !!process.env.UPDATE_LOCK_FILE, forceCopyDependencies: boolean = false): Promise<void> {
```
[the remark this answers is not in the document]
ryanthemanuel: The issue here was around usability/readability of the tests. By leaving the node modules in the temp directory sym linked to the fixture project, I end up getting a lot of paths that are relative from the fixture project to the temp directory rather than within the project itself.

e.g. https://github.com/cypress-io/cypress/blob/ryanm/feat/add-packherd-to-monorepo/tooling/packherd/test/entry-generator.spec.ts#L33-L36

ends up being:

```
"../../../../../../../../../tmp/cy-system-tests-node-modules/v8-snapshot/minimal/node_modules/isobject/index.cjs.js"
"../../../../../../../../../tmp/cy-system-tests-node-modules/v8-snapshot/minimal/node_modules/tmpfile/index.js"
```

instead of:

```
"node_modules/isobject/index.cjs.js"
"node_modules/tmpfile/index.js"
```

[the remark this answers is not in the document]
ryanthemanuel: Good call on the arguments. Will update those

tooling/packherd/test/packherd.spec.ts
ryanthemanuel: Updated here: [`c3e4634` (#23257)](https://github.com/cypress-io/cypress/pull/23257/commits/c3e46342d4b31d7bf8f5b89ccd5c512ac4e53a19)

(though it's going to be harder on subsequent PRs 😄)

system-tests/lib/dep-installer/index.ts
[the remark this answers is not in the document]
ryanthemanuel: Updated here: [`c3e4634` (#23257)](https://github.com/cypress-io/cypress/pull/23257/commits/c3e46342d4b31d7bf8f5b89ccd5c512ac4e53a19)

packages/server/package.json
ryanthemanuel: Good call. This was from a previous PR (I forgot to add the dependency). A lot of the benchmarking that had been done with v8-snapshots previously used `performance-now`, but I think I don't see a reason not to use `performance.now()`. Updated to use `performance.now()` here: [`c3e4634` (#23257)](https://github.com/cypress-io/cypress/pull/23257/commits/c3e46342d4b31d7bf8f5b89ccd5c512ac4e53a19)

packages/packherd-require/package.json
ryanthemanuel: Removed: [`a6f143f` (#23257)](https://github.com/cypress-io/cypress/pull/23257/commits/a6f143f6b012f6fc2fb6c7670d7ad851d5f08fd0)

packages/packherd-require/src/loader.ts
```diff
@@ -0,0 +1,1107 @@
+import debug from 'debug'
+import Module from 'module'
+import path from 'path'
+import type {
+  ModuleBuiltin as ModuleBuiltin,
```
lmiller1990, line 5:
```suggestion
  ModuleBuiltin,
```

lmiller1990:
Still internalising and wrapping my head around the whole snapshot project, I don't see any obvious problems (with the caveat if I there are, I probably missed it or it's still outside my understanding).

Excited to keep following this project and understanding the intricaties/tricks to speed up Electron apps, and by the looks of what I've read about snapshots, Node.js apps in general.

flotwig (approved):

PR #23259, opened 2022-08-10
fix: Add delay to showing ACI header tooltips

### LINKED ISSUE #23115: Tooltips can cover the "New Spec" button
### Current behavior

Currently there's a mildly annoying situation where users intending to click "New Spec" on the Spec Lists have to be mindful of avoiding the Column heading beneath it with their pointer so that the tooltip doesn't cover the button.

https://user-images.githubusercontent.com/8340719/182914074-d622ff37-aaf8-4235-b567-d28e57293884.mov

### Desired behavior

A small delay before tooltips open would avoid this. I'd suggest we increase the `show` delay above 0 in the main `Tooltip` Vue component so that simply passing over and element doesn't unintentionally trigger tooltips. A very unscientific test just now suggests 150 ms is a good starting point, and this is a good signal generally that the pointer has come to rest on an element, so it avoids a lot of scenarios where currently moving the mouse around can open many tooltips at once.

### Test code to reproduce

N/A

### Cypress Version

10.4.0

### Other

_No response_

mike-plummer:
- Closes #23115 

### User facing changelog
* Add delay to header tooltips so they don't pop open unless hovered on

### Additional details
Had to make two fixes here:
1. Add a "show-delay" to both header components and pass thru to base Tooltip
2. `floating-vue` by default keeps tooltip content in the DOM for 5 seconds after the tooltip closes before disposing of it. This caused the delay to be ignored if you moused back over the tooltip within that 5 second period. The disposal period is a global conflg for the library - I don't foresee any issues with reducing it to immediately dispose (it may actually clean up some of the weird tooltip behaviors we have currently) but let me know if you have concerns

### Steps to test
1. Open a project in the app
2. From the specs list, move your mouse to the "New Spec" button (crossing over the "Average Duration" header in the process). Observe the header tooltip does not pop open, blocking the button
3. Hover over the "Last Updated", "Latest Runs" and "Average duration" headers, observe tooltips display after a brief delay
4. Validate that other uses of `Tooltip` in the app are unaffected (collapsed sidebar icons, selector playground, last updated row elements, etc)

https://user-images.githubusercontent.com/11482842/184014634-9a3f76ca-82b6-43de-8e59-ca82e2230f11.mov

### PR Tasks

### DIFF
--- a/packages/app/src/specs/LastUpdatedHeader.cy.tsx
+++ b/packages/app/src/specs/LastUpdatedHeader.cy.tsx
@@ -3,6 +3,8 @@ import LastUpdatedHeader from './LastUpdatedHeader.vue'
 import { defaultMessages } from '@cy/i18n'
 
 describe('<LastUpdatedHeader />', () => {
+  const popperContentSelector = '.v-popper__popper--shown'
+
   function mountWithProps (isGitAvailable: boolean) {
     cy.mount(() => <div class="flex justify-around"><LastUpdatedHeader isGitAvailable={isGitAvailable} /></div>)
   }
@@ -15,7 +17,7 @@ describe('<LastUpdatedHeader />', () => {
     const expectedTooltipText = defaultMessages.specPage.lastUpdated.tooltip.gitInfoAvailable
     .replace('{0}', defaultMessages.specPage.lastUpdated.tooltip.gitStatus)
 
-    cy.get('.v-popper__popper--shown').should('have.text', expectedTooltipText)
+    cy.get(popperContentSelector).should('have.text', expectedTooltipText)
 
     cy.percySnapshot()
   })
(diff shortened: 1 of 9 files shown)


packages/app/src/specs/LastUpdatedHeader.cy.tsx
```diff
@@ -28,8 +30,17 @@ describe('<LastUpdatedHeader />', () => {
...
     .replace('{0}', defaultMessages.specPage.lastUpdated.tooltip.gitInfo)
 
-    cy.get('.v-popper__popper--shown').should('have.text', expectedTooltipText)
+    cy.get(popperContentSelector).should('have.text', expectedTooltipText)
 
     cy.percySnapshot()
   })
+
+  it('delays popping tooltip', () => {
+    cy.clock()
```
lmiller1990, line 39: Nice usage of clock - I don't see this API too often, but it's neat.

lmiller1990 (approved):
I was just thinking this would be nice to fix - didn't realize we already had a ticket for it. Looks great!

astone123 (approved):
Nice, this has been really annoying for me recently 😄 🎉

PR #23261, opened 2022-08-10
chore: adding more required fields to issue creation template in github

nagash77:
### User facing changelog
n/a

### Additional details
Adding fields to our issue creation templates to assist in faster and more accurate triage of issues coming into the Cypress repo 

### Steps to test
Create an bug or installation issue in the cypress repo to see new fields.

there are more required fields to enter a bug or installation issue
### PR Tasks

- [ ] Have tests been added/updated? n/a
- [ ] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only) n/a
- [ ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)?  n/a
- [ ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)? n/a


.github/ISSUE_TEMPLATE/2-install-issue.yml
```diff
@@ -53,6 +60,14 @@ body:
         - other
     validations:
       required: true
+  - type: input
+    id: os-version
+    attributes:
+      label: Operating System Version
+      description: What version of your operating system are you running Cyrpess on?
```
rachelruderman, line 67:
```suggestion
      description: What version of your operating system are you running Cypress on?
```

.github/ISSUE_TEMPLATE/1-bug-report.yml
```diff
@@ -35,6 +35,30 @@ body:
...
+      required: true
+  - type: input
+    id: os
+    attributes:
+      label: Operating System
+      description: What operating system is Cypress running on when you are encountering your problem?  The more specific the better  ie macOS 12.4 or Windows 10.0.19044.1889
+      placeholder: ex. macOS 12.4
+    validations:
+      required: true
+  - type: textarea
```
BlueWinds, line 54: I'm not a fan of this input. In most cases we care more about a reproduction than we do about the debug logs; They're a heavy club of last resort when all else fails, and I don't want to make them required for opening an issue.

BlueWinds (changes requested):

.github/ISSUE_TEMPLATE/2-install-issue.yml
```diff
@@ -42,6 +42,13 @@ body:
         - other
     validations:
       required: true
+  - type: input
+    id: package-manager-version
+    attributes:
+      label: Package Manager Version
+      description: What version of your chosen package manager
```
rockhold, line 49: It's minor, but there is some inconsistency in how we're wording these descriptions. For the Cypress Version attribute above, the description is a statement. For this one and the below, it's a question.

For this one in particular, it also seems to be an incomplete thought, though. I'd personally prefer consistency across all the questions, but at a minimum, I think we could fix up the wording on this one.

nagash77: good call out.  I will update

.github/ISSUE_TEMPLATE/1-bug-report.yml
```diff
@@ -35,6 +35,30 @@ body:
...
+      label: Node version
+      description: What version of node.js are you using to run Cypress?
+      placeholder: ex. v16.16.0
+    validations:
+      required: true
+  - type: input
+    id: os
+    attributes:
+      label: Operating System
+      description: What operating system is Cypress running on when you are encountering your problem?  The more specific the better  ie macOS 12.4 or Windows 10.0.19044.1889
```
rockhold, line 50: Just a heads up that it looks like GitHub will not render double white space between sentences. I noticed it as an inconsistency in the doc here, but GitHub ignores it, so 🤷🏼 

Also, more minor stuff here, but we have a mix of using ending punctuation for these descriptions across the doc. I may be the only one that notices stuff like that (b/c my brain apparently likes to notice things like that), but in case you care and want to fix it, I thought I'd just mention it 😄

.github/ISSUE_TEMPLATE/1-bug-report.yml
tbiethman: I wouldn't go so far as to say they're a last resort, but I agree that they're not needed/helpful for every class of issue. We could leave this as an optional field, unless we think its presence will cause undue friction. 

If we do keep it, this link (https://docs.cypress.io/guides/references/troubleshooting#Print-DEBUG-logs) is probably more applicable. The current link is targeted towards debugging specs within the browser more-so than Cypress itself.

nagash77: I agree it is a heavier lift than most of the information.  But based on the tickets I have seen coming in over the last month or two, we need to raise the bar on expectations of the community when entering a ticket if for no other reason than we simply don't have the resources to triage in an effective way on a majority of issues entered.  By asking for this information up front we can more quickly dive deeper into issues without slow back and forth async communication where we are never sure if someone is going to ever get back to us (they often don't).  I am hoping with clearer details such as debug logs from the start we can get through more issues and ultimately have time to fix more issues with the limited resources we have as a consequence.

BlueWinds: I agree about raising the bar - but I'd rather do that by asking for reliable reproductions, rather than debug logs.

Perhaps instead, a required link to "Repository demonstrating the issue, or CI build Cypress team can run"? Something like that. I'd be fine with this as an optional field, but as a required one it feels like it's just adding friction to the process and not actually providing what we need - which is a way to reproduce issues and more quickly locate the heart of the problem.

BlueWinds: Debug logs put all the effort on us, to read through the logs and make sense of them, when what we actually care about is asking users to put in more effort.

rockhold: I like asking for it up front, but perhaps not requiring it. We could update the description to indicate that we may have to ask for it later on, though, and that it could delay investigating their issue.

Beyond that, though, I'm concerned that without the additional `render` attribute, these logs could make scrolling through the issue incredibly painful. It looks like this attribute will not expand the field to show the entire content (if I'm understanding the [textarea render attribute doc](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#attributes-1) correctly). Perhaps we add that attribute with a value of `Text` for language?

nagash77: @BlueWinds @rockhold I have set required to false based on general feedback.  

@rockhold good point on the rendering of the textarea for debug logs.  I have added render: shell based on some examples I could find online.  We can change to text if this parsing isn't optimal in everyday life.

@tbiethman thank for the more relevant debug log link.  I have updated the URL.

I also tried to go through and standardize wording on fields and punctuation but that is not my strong suit so if you all notice any other inconsistencies worth fixing let me know.

BlueWinds (approved):

tbiethman (approved):

tbiethman (approved):

mjhenkes (approved):

rockhold (approved):
Thanks for the updates to the language! Looks good to me.

ashikkumar23:
Hi there, 

I was trying to create an issue using the bug template mentioned here, https://github.com/cypress-io/cypress/issues/new?assignees=&labels=&template=1-bug-report.yml

However, I am unable to submit an issue successfully! Results in `Comment is too long (maximum is 65536 characters)` nevertheless I have not crossed the limit!!

image

Can someone please help with this?

Thanks.

PR #23283, opened 2022-08-11
feat: Support dependencies in cy.origin() callback

### LINKED ISSUE #19974: [Cross-origin]: Support dependencies in cy.origin() callback
Enable utilizing external dependencies within the `cy.origin()` callback.

Dependencies are different files, either 3rd party node packages or local files referenced with a relative path.

This will enable including a file that defines custom commands, so they can be shared between all origins.

One possible implementation:
- Pass a require function to the multi-domain callback
- Run the multi-domain callback through webpack to support requiring files inside of it

```js
cy.origin('foobar.com', (args, { require }) => {
  const utils = Cypress.require('../support/uitls')
  Cypress.require('../support/command') // defines custom commands like cy.login()

  utils.stringify()
  cy.login()
})
```

chrisbreiding:
- Closes #19974

### User facing changelog

- Added support for requiring dependencies within the `cy.origin()` callback. See [insert docs link here] for more information.

### Additional details

Enables support for requiring dependencies within the `cy.origin()` callback using the `Cypress.require()` method.

```js
  const _ = Cypress.require('lodash')

  expect(_.get({ foo: 'foo' }, 'foo')).to.equal('foo')
})
```

It's implemented by adding a loader to the webpack-preprocessor that extracts the callback when it contains a reference to `Cypress.require()`. It replaces the callback with an object that is used at runtime to reference and call into the extracted callback. It replaces any `Cypress.require()` uses with standard `require()`s and then runs the callback through webpack again so that it has its own bundle with its dependencies. 

One caveat of this is that the latest webpack preprocessor is required for this functionality. If it's not used or if an older version without the loader is used, `Cypress.require()` will not be replaced. Its default functionality is to error with a message that the latest webpack preprocessor is required.

### Steps to test

- Run Cypress with `experimentalSessionAndOrigin: true`
- Run a test like the following (this one will work in the driver):

```js
  const _ = Cypress.require('lodash')

  expect(_.get({ foo: 'foo' }, 'foo')).to.equal('foo')
})
```

Users can now use 1st and 3rd party dependencies in the `cy.origin()` callback and more easily share common code and custom commands between cross-origin tests.

### PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4681


npm/webpack-preprocessor/deferred.ts
```diff
@@ -1,17 +0,0 @@
-import * as Promise from 'bluebird'
```
chrisbreiding, line 1: Moved into `webpack-preprocessor/lib/utils.ts`

npm/webpack-preprocessor/index.ts
```diff
@@ -1,19 +1,33 @@
...
+import _ from 'lodash'
+import * as events from 'events'
+import * as path from 'path'
+import webpack from 'webpack'
+import utils from './lib/utils'
+import { crossOriginCallbackStore } from './lib/cross-origin-callback-store'
 import { overrideSourceMaps } from './lib/typescript-overrides'
+import { compileCrossOriginCallbackFiles } from './lib/cross-origin-callback-compile'
 
-import * as Promise from 'bluebird'
```
mjhenkes, line 3: +1 for being explicit about if it's a regular or a bluebird promise

chrisbreiding: Plays better with typescript this way too, since it's not shadowing the global 'Promise'

packages/driver/src/cross-origin/origin_fn.ts
```diff
@@ -126,7 +132,20 @@ export const handleOriginFn = (Cypress: Cypress.Cypress, cy: $Cy) => {
     })
 
     try {
-      const value = window.eval(`(${fn})`)(args)
+      let value
+
+      if (_.isString(fn)) {
```
mjhenkes, line 137: Nice, we only bundle up origin blocks with cypress.require in them. That should save on network calls.

mjhenkes:
To double check I'm understanding how this works:

User writes a test using Cypres.require in cy.origin

```js
it('', () => {
  cy.visit('localhost')
  cy.origin('foobar.com', () => {
    const dep = Cypress.require('someDep')
    dep()
  })
})
```

During the webpack stage as the last loader we use babel to parse the AST for cy.origin blocks that contain Cypress.require statements. If and only if we find a Cypress.require statement, we extract the cy.origin block, replacing it with an object containing the uuid file name and the file location. We swap Cypress.require for require and then run that block through webpack temporarily storing the file in the (virtual?) filesystem for later retrieval.

We will have a bundle per origin block containing a Cypress.require method. (Is that right?) I wonder we could eventually reduce bundle size by implementing chunking to share chunks between bundles.

After cypress loads when the origin block is executed, we send the object to the spec bridge which then loads and evals the bundle.

chrisbreiding:
> During the webpack stage as the last loader we use babel to parse the AST for cy.origin blocks that contain Cypress.require statements. If and only if we find a Cypress.require statement, we extract the cy.origin block, replacing it with an object containing the uuid file name and the file location. We swap Cypress.require for require and then run that block through webpack temporarily storing the file in the (virtual?) filesystem for later retrieval.

Yes, that's right. The extracted callback is virtual/in-memory (the input to the subsequent webpack compilation), but the output bundle is saved to disk. 

> We will have a bundle per origin block containing a Cypress.require method. (Is that right?) I wonder we could eventually reduce bundle size by implementing chunking to share chunks between bundles.

Yes, right again. That's a good idea about utilizing chunks to share the common dependencies. We don't even do that currently between support file and spec file, so we could save some bundle size between them too. I'll draft an issue on our board for that.

mjhenkes (approved):
Seems like a solid approach to me 👍

npm/webpack-preprocessor/lib/cross-origin-callback-compile.ts
```diff
@@ -0,0 +1,76 @@
...
+    const inputPath = path.join(dir, inputFileName)
+
+    memo.entry[inputFileName] = inputPath
+    memo.virtualConfig[inputPath] = source
+
+    return memo
+  }, { entry: {}, virtualConfig: {} })
+}
+
+const getWebpackOptions = ({ webpackOptions, entry, virtualConfig, outputDir }) => {
```
AtofStryker, line 25:
```suggestion
const getWebpackOptions = ({ webpackOptions, entry, virtualConfig, outputDir }: {
  webpackOptions: webpack.Configuration
  entry: { [key: string]: string }
  virtualConfig: { [key: string]: string }
  outputDir: string
}) => {
```

npm/webpack-preprocessor/index.ts
```diff
@@ -223,6 +243,21 @@ const preprocessor: WebpackPreprocessor = (options: PreprocessorOptions = {}): F
     })
     .value() as any
 
+    crossOriginCallbackLoaderAdded = false
+
+    if (!crossOriginCallbackLoaderAdded && cypressConfig.experimentalSessionAndOrigin) {
```
AtofStryker, line 248: Is there a case where `!crossOriginCallbackLoaderAdded` doesn't evaluate to `true` here since `crossOriginCallbackLoaderAdded` is set to `false` right about it? I think I am lacking some knowledge here 😅

npm/webpack-preprocessor/lib/cross-origin-callback-compile.ts
```diff
@@ -0,0 +1,76 @@
+import _ from 'lodash'
+import Debug from 'debug'
+import * as path from 'path'
+import webpack from 'webpack'
+import { CrossOriginCallbackStoreFile } from './cross-origin-callback-store'
+
+const VirtualModulesPlugin = require('webpack-virtual-modules')
+
+const debug = Debug('cypress:webpack')
+
```
AtofStryker, line 10: I feel documenting the methods in this file would go a long way to make it a bit clearer what is happening. I had to give the tests in the driver a run to get a decent handle on that, but I also feel like that is expected?

AtofStryker (changes requested):
Super exciting to see this work 🎉 !

I wonder if you can leverage the [cy-in-cy](https://github.com/cypress-io/cypress/blob/develop/packages/app/cypress/e2e/cypress-origin-communicator.cy.ts#L111) tests to test `Cypress.require` hot reload when the origin block changes, or even better when the dependency changes. 

Not sure why, but I can't seem to get the debug logs to print when running the unit tests when inside `cross-origin-callback-compile`, even though I can see `compileCrossOriginCallbackFiles`. Is it stubbed and I am just missing that?

AtofStryker:
> To double check I'm understanding how this works:
> 
> User writes a test using Cypres.require in cy.origin
> 
> ```js
> it('', () => {
>   cy.visit('localhost')
>   cy.origin('foobar.com', () => {
>     const dep = Cypress.require('someDep')
>     dep()
>   })
> })
> ```
> 
> During the webpack stage as the last loader we use babel to parse the AST for cy.origin blocks that contain Cypress.require statements. If and only if we find a Cypress.require statement, we extract the cy.origin block, replacing it with an object containing the uuid file name and the file location. We swap Cypress.require for require and then run that block through webpack temporarily storing the file in the (virtual?) filesystem for later retrieval.
> 
> We will have a bundle per origin block containing a Cypress.require method. (Is that right?) I wonder we could eventually reduce bundle size by implementing chunking to share chunks between bundles.
> 
> After cypress loads when the origin block is executed, we send the object to the spec bridge which then loads and evals the bundle.

Can we capture this in documentation somewhere? Maybe a `README.md`?

npm/webpack-preprocessor/index.ts
```diff
@@ -1,19 +1,33 @@
...
+interface CypressConfigPartial {
+  experimentalSessionAndOrigin: boolean
+}
+
+declare global {
+  // this could be undefined if used with an older version of Cypress or if
+  // it's removed in a future version. this global should have the entire
+  // config, but only declaring the properties used for brevity's sake
+  // eslint-disable-next-line no-var
+  var __CYPRESS_CONFIG__: CypressConfigPartial | undefined
```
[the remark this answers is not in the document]
chrisbreiding: I thought about that, but I don't think it's worth it for one property.

npm/webpack-preprocessor/index.ts
chrisbreiding: Ah whoops, I think I added that when debugging and forgot to remove it. Good catch!

npm/webpack-preprocessor/index.ts
```diff
@@ -295,29 +330,68 @@ const preprocessor: WebpackPreprocessor = (options: PreprocessorOptions = {}): F
...
+          debug('no cross-origin callback files')
+
+          return resolveAllBundles()
+        }
+
+        compileCrossOriginCallbackFiles(crossOriginCallbackStore.getFilesFor(sourceFiles), {
+          originalFilePath: filePath,
+          webpackOptions,
+        })
+        .then(() => {
```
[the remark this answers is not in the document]
chrisbreiding: No, the outputDir is encapsulated in the object we replace the callback function with. That's where we get it from to get the file at runtime. So it's okay to ignore it here.

npm/webpack-preprocessor/index.ts
chrisbreiding: Addressed in [`b05864a` (#23283)](https://github.com/cypress-io/cypress/pull/23283/commits/b05864acb22eb9a00efec5093ccfae270780ebc6)

npm/webpack-preprocessor/lib/cross-origin-callback-compile.ts
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/23283/commits/a3826c68871558f1c1808fc8ad17bc63d532ad3d

npm/webpack-preprocessor/lib/cross-origin-callback-compile.ts
```diff
@@ -0,0 +1,76 @@
...
+const getConfig = ({ files, originalFilePath }) => {
```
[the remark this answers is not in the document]
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/23283/commits/a3826c68871558f1c1808fc8ad17bc63d532ad3d

npm/webpack-preprocessor/lib/cross-origin-callback-compile.ts
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/23283/commits/a3826c68871558f1c1808fc8ad17bc63d532ad3d

npm/webpack-preprocessor/lib/cross-origin-callback-compile.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  return modifiedWebpackOptions
+}
+
+interface CompileOptions {
+  originalFilePath: string
+  webpackOptions: webpack.Configuration
+}
+
+export const compileCrossOriginCallbackFiles = (files: CrossOriginCallbackStoreFile[], options: CompileOptions) => {
```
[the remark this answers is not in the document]
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/23283/commits/a3826c68871558f1c1808fc8ad17bc63d532ad3d

npm/webpack-preprocessor/lib/cross-origin-callback-compile.ts
```diff
@@ -0,0 +1,76 @@
...
+  const outputDir = path.dirname(files[0].outputFilePath)
+  const { entry, virtualConfig } = getConfig({ files, originalFilePath })
+  const modifiedWebpackOptions = getWebpackOptions({
+    webpackOptions,
+    entry,
+    virtualConfig,
+    outputDir,
+  })
+
+  return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/23283/commits/a3826c68871558f1c1808fc8ad17bc63d532ad3d

npm/webpack-preprocessor/test/unit/cross-origin-callback-loader.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+
+      expectAddFileSource(store).to.equal(stripIndent`
+        __cypressCrossOriginCallback = () => {
+          const someVar = 'someValue';
+
+          const result = require('./fn')(someVar);
+
+          expect(result).to.equal('mutated someVar');
+        }`)
+    })
```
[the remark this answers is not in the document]
chrisbreiding: Added that test in https://github.com/cypress-io/cypress/pull/23283/commits/86225f2367a95f85dcc3b6dab017c8c96794e099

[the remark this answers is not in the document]
chrisbreiding: The example

```js
cy.origin('http://foobar.com:3500', { args: undefined}}, (foo = Cypress.require('foo')) => {})
```

isn't valid because the `Cypress.require()` isn't inside the callback itself.

A different scenario that's not currently handled is assigning the callback to a variable and passing that into `cy.origin()`:

```js
const callback = () => {
  Cypress.require('foo')
}

cy.origin('http://foobar.com:3500', callback)
```

It _might_ be possible to replace it in that case, but I think that should be a later enhancement.

chrisbreiding:
> Can we capture this in documentation somewhere? Maybe a README.md?

Added in https://github.com/cypress-io/cypress/pull/23283/commits/2111612a074c10e4158da9a4494a89f0cc014bef

npm/webpack-preprocessor/index.ts
```diff
@@ -295,29 +328,78 @@ const preprocessor: WebpackPreprocessor = (options: PreprocessorOptions = {}): F
...
+          // resolve with the outputPath so Cypress knows where to serve
+          // the file from
+          deferred.resolve(outputPath)
+        })
+
+        bundles[filePath].deferreds.length = 0
+      }
+
+      // the cross-origin-callback-loader extracts any cy.origin() callback
+      // functions // that contains Cypress.require() and stores their sources
```
AtofStryker, line 342:
```suggestion
      // functions that contains Cypress.require() and stores their sources
```

npm/webpack-preprocessor/test/unit/cross-origin-callback-loader.spec.ts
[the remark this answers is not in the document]
AtofStryker: I think that is completely reasonable. Especially if a user is doing something like that the likelihood of a workaround is extremely high.

AtofStryker (approved):

packages/server/lib/plugins/child/run_plugins.js
```diff
@@ -125,6 +125,9 @@ class RunPlugins {
     })
     .then((modifiedCfg) => {
       debug('plugins file successfully loaded')
+      // provide the config globally so plugins can access it
+      global.__CYPRESS_CONFIG__ = modifiedCfg
```
emilyrohrbough, line 129: Oof....this is quite the obj to put on the process specifically for this change. Once we go GA, this won't necessarily be needed anymore either. Any reason you didn't add as `__CYPRESS_EXPERIMENAL_SESSION_AND_ORIGIN__`? and/or a configuration option that could be pulled when both experimental mode & GA mode will need to be supported?

Also, does the modifiedCfg included the resolved configuration as well? `config.resolved`? if so, we should remove that

packages/driver/cypress/support/defaults.js
```diff
@@ -31,3 +31,9 @@ beforeEach(() => {
     $(cy.state('window')).off()
   } catch (error) {} // eslint-disable-line no-empty
 })
+
```
emilyrohrbough, line 34: Just curious, if someone used `Cypress.require()` at the root supportFile level, what is the error messaging?

packages/driver/src/cross-origin/cypress.ts
```diff
@@ -102,6 +103,15 @@ const setup = (cypressConfig: Cypress.Config, env: Cypress.ObjectLike) => {
   // @ts-ignore
   Cypress.isCy = cy.isCy
 
+  // this is valid inside the cy.origin() callback, but it should be replaced
+  // by the webpack preprocessor with an actual require() before the spec code
+  // is run in the browser. if it's not, it means the user isn't using the
+  // webpack preprocessor or is using an older version of it. this error guides
```
emilyrohrbough, line 109: can you add the version of webpack preprocessor which introduces support for this?

packages/driver/src/cypress/error_messages.ts
```diff
@@ -1454,6 +1454,17 @@ export default {
     },
   },
 
+  require: {
+    invalid_outside_origin: {
+      message: `${cmd('Cypress.require')} can only be used inside the ${cmd('origin')} callback.`,
+      docsUrl: 'https://on.cypress.io/origin',
+    },
+    invalid_inside_origin: {
+      message: `Importing dependencies with ${cmd('Cypress.require')} requires using the latest version of \`@cypress/webpack-preprocessor\`.`,
```
emilyrohrbough, line 1463:
```suggestion
      message: `Importing dependencies with ${cmd('Cypress.require')} requires using \`@cypress/webpack-preprocessor\` version ^{VERSION_NUMER}.`,
```

npm/webpack-preprocessor/lib/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+    promise,
+  }
+}
+
+function hash (contents: string) {
+  return md5(contents)
+}
+
+function tmpdir () {
+  return path.join(os.tmpdir(), 'cypress', 'webpack-preprocessor')
```
emilyrohrbough, line 27: With this being a tempDir, what type of file permission testing / issue could we hit? any risk webpack will fail here?

npm/webpack-preprocessor/lib/cross-origin-callback-store.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  getFilesFor (sourceFiles: string[]) {
+    return Object.keys(this.files).reduce((files, sourceFilePath) => {
+      return sourceFiles.includes(sourceFilePath) ? files.concat(this.files[sourceFilePath]) : files
+    }, [] as CrossOriginCallbackStoreFile[])
+  }
+
+  reset (sourceFilePath: string) {
+    this.files[sourceFilePath] = []
```
emilyrohrbough, line 25: Should this also delete the files from the file system or when will those be cleaned up?

npm/webpack-preprocessor/lib/cross-origin-callback-store.ts
```diff
@@ -0,0 +1,29 @@
...
+  inputFileName: string
+  outputFilePath: string
+  source: string
+}
+
+export class CrossOriginCallbackStore {
+  private files: { [key: string]: CrossOriginCallbackStoreFile[] } = {}
+
+  addFile (sourceFilePath: string, file: CrossOriginCallbackStoreFile) {
+    this.files[sourceFilePath] = (this.files[sourceFilePath] || []).concat(file)
```
emilyrohrbough, line 11: Just curious, with our current file watchers in the data-context, we are watching more than we were in 9.x in the `cypress` directory. If `Cypress.require()` is used on a file within the dir, is replacing those contents in this list or appending it even it it prev existed?

npm/webpack-preprocessor/lib/cross-origin-callback-loader.ts
```diff
@@ -0,0 +1,165 @@
...
+      //   // ... webpack stuff stuff ...              ┛
+      //   __cypressCrossOriginCallback = (args) => {  ┓ extracted callback
+      //     const dep = require('../path/to/dep')     ┃
+      //     // ... test stuff ...                     ┃
+      //   }                                           ┛
+      //   // ... webpack stuff stuff ...              ┓ added by webpack
+      // }())                                          ┛
+      // __cypressCrossOriginCallback(args)            】added at runtime
+      //
+      const callbackName = '__cypressCrossOriginCallback'
```
emilyrohrbough, line 120: Once we start on the iframe handling, will we be reusing this?

packages/server/lib/plugins/child/run_plugins.js
emilyrohrbough: Could align with naming you used in the preprocessor: `__supportCypressCrossOriginCallback`

lmiller1990 (commented):
I just saw this PR floating around and got a bit curious - I didn't look deeply at the code yet, but at a cursory glance it looks like using the `Cypress.require` feature is tightly coupled to webpack-preprocessor? If you are not using webpack-preprocessor, but another preprocessor, does that mean that `Cypress.require` won't work?

packages/server/lib/plugins/child/run_plugins.js
chrisbreiding: That's a good point and it helped me recognize a few issues with this approach. 

I refactored things in [`90648a0` (#23283)](https://github.com/cypress-io/cypress/pull/23283/commits/90648a01f517b743f79e9bd3cc2afbc7859b3283) so it no longer sets the config as a global. Instead, the Cypress binary sets a list of commands based on the flag. This accomplishes the following:

- It greatly lightens the size of the value.
- Once we remove the experimental flag, it can just always set the value. Otherwise, we would have still had to set the experimental flag on this global object so it would be backwards-compatible.
- Once we add iframe support, we can expand the list of commands, adding whatever name we decide on for it.

packages/driver/cypress/support/defaults.js
chrisbreiding: Any usage of it outside the `cy.origin()` callback results in [this error](https://github.com/cypress-io/cypress/blob/90648a01f517b743f79e9bd3cc2afbc7859b3283/packages/driver/src/cypress/error_messages.ts#L1458-L1461).

```
`Cypress.require()` can only be used inside the `cy.origin()` callback.
```

packages/driver/src/cross-origin/cypress.ts
chrisbreiding: I can, but it will need to wait until closer to release time in order to determine what version that will be, since there could be versions of the preprocessor released between now and when this is released.

npm/webpack-preprocessor/lib/utils.ts
chrisbreiding: We use `os.tmpdir()` for [installing the Cypress binary](https://github.com/cypress-io/cypress/blob/7f2522f85769b9b72a159db6cf39807a036fba83/cli/lib/tasks/install.js#L345-L347), so if it's not writeable, the user wouldn't be able to get this far.

npm/webpack-preprocessor/lib/cross-origin-callback-store.ts
chrisbreiding: This store is used to track files that will _eventually_ be created when the loader finds uses of `Cypress.require()`. It's completely in-memory. Nothing is written to disk until webpack runs through the files referenced and compiles them. It doesn't currently add any more watching, though that is something that will need to be implemented. Even then, we won't need to worry about it since webpack will handle it for us.

npm/webpack-preprocessor/lib/cross-origin-callback-loader.ts
chrisbreiding: Yes, we will.

chrisbreiding:
> I just saw this PR floating around and got a bit curious - I didn't look deeply at the code yet, but at a cursory glance it looks like using the Cypress.require feature is tightly coupled to webpack-preprocessor? If you are not using webpack-preprocessor, but another preprocessor, does that mean that Cypress.require won't work?

@lmiller1990, yes that's correct. It requires using webpack-preprocessor.

npm/webpack-preprocessor/lib/cross-origin-callback-store.ts
chrisbreiding: Realized we have a perfectly good place to clean them up: when the preprocessor receives the 'close' event. Addressed in [`95e7c65` (#23283)](https://github.com/cypress-io/cypress/pull/23283/commits/95e7c65080fc66040bdea8f1c8d8ba5e101c8ef7)

lmiller1990:
> > I just saw this PR floating around and got a bit curious - I didn't look deeply at the code yet, but at a cursory glance it looks like using the Cypress.require feature is tightly coupled to webpack-preprocessor? If you are not using webpack-preprocessor, but another preprocessor, does that mean that Cypress.require won't work?
> 
> @lmiller1990, yes that's correct. It requires using webpack-preprocessor.

Okay, good to know. 

I'm super deep in firewatch and not sure I have time to do a more thorough review at the point. I'm going to remove my review request - definitely interested in digging into this more down the road, but I won't block merging. I imagine this was pretty complex to implement, great job!

circle.yml
```diff
@@ -479,7 +479,7 @@ commands:
               # internal PR
               if <<parameters.experimentalSessionAndOrigin>>; then
                 CYPRESS_RECORD_KEY=$MAIN_RECORD_KEY \
-                yarn cypress:run-experimentalSessionAndOrigin --record --parallel --group 5x-driver-<<parameters.browser>>-experimentalSessionAndOrigin --browser <<parameters.browser>>
+                DEBUG=cypress:webpac* yarn cypress:run-experimentalSessionAndOrigin --record --parallel --group 5x-driver-<<parameters.browser>>-experimentalSessionAndOrigin --browser <<parameters.browser>>
```
emilyrohrbough: still need this debug log?

npm/webpack-preprocessor/package.json
```diff
@@ -22,6 +22,8 @@
   "dependencies": {
     "bluebird": "3.7.1",
     "debug": "^4.3.2",
+    "del": "^6.1.1",
```
emilyrohrbough: Well has quite a few dependencies to only use it to delete the temp dir. Why did you choose `del` vs `fs-extra` (fewer dependencies) or just using fs provided by node?

packages/server/lib/plugins/child/run_plugins.js
```diff
@@ -145,6 +151,19 @@ class RunPlugins {
...
+    // prefer the modified value if it's specified
+    if (
+      typeof modifiedConfig === 'object'
+      && typeof modifiedConfig.experimentalSessionAndOrigin === 'boolean'
+    ) {
+      return modifiedConfig.experimentalSessionAndOrigin
+    }
+
+    // otherwise, use the initial value
+    return initialConfig.experimentalSessionAndOrigin
```
emilyrohrbough, line 164: Where you finding it was unset in modifiedCfg?

chrisbreiding: `modifiedConfig` is what the user returns from `setupNodeEvents`, so it could be undefined or not include the value.

circle.yml
chrisbreiding: Nope - good catch! Removed in https://github.com/cypress-io/cypress/pull/23283/commits/dad0523765bc531abf2e30330133fc266e1fef3c

packages/server/lib/plugins/child/run_plugins.js
emilyrohrbough: ohhh right!

npm/webpack-preprocessor/package.json
chrisbreiding: Node's `fs` doesn't have support for recursively removing directories until 14.4.0 and we need to support Node 12+. Good call on using `fs-extra` instead of `del`. Looks like it [it is smaller](https://bundlephobia.com/package/fs-extra@10.1.0) than [del](https://github.com/sindresorhus/del). Changed it in https://github.com/cypress-io/cypress/pull/23283/commits/51492f3d3ef2fe7799b5c5ec5be4ef938602bea9

npm/webpack-preprocessor/package.json
```diff
@@ -22,7 +22,7 @@
   "dependencies": {
     "bluebird": "3.7.1",
     "debug": "^4.3.2",
-    "del": "^6.1.1",
+    "fs-extra": "^10.1.0",
```
emilyrohrbough, line 25: Lets remove `"fs-extra": "9.1.0",` from the devDependencies list

emilyrohrbough (approved):

npm/webpack-preprocessor/package.json
chrisbreiding: Good call. Addressed in https://github.com/cypress-io/cypress/pull/23283/commits/6bea1fc483c4228c4c11bd4d02585576316fed90

npm/webpack-preprocessor/package.json
```diff
@@ -46,7 +50,7 @@
     "eslint-plugin-mocha": "8.1.0",
     "fast-glob": "3.1.1",
     "find-webpack": "1.5.0",
-    "fs-extra": "9.1.0",
+    "md5": "2.3.0",
```
KittySparkles, line 53: Shouldn’t this be a production dependency? See https://github.com/cypress-io/cypress/issues/24051.

chrisbreiding: Yes indeed! Thanks for bringing this to my attention as I had missed the issue for it. Will be fixed by https://github.com/cypress-io/cypress/pull/24098.

PR #23295, opened 2022-08-11
feat(app): show "connect project" after login if project is not connected

### LINKED ISSUE #23022: Show "Connect project" CTA after login if projectId is unset
After being redirected from the login flow, if there is no projectId in the config, show a "Connect project" button instead of a "Continue" button that takes you to the connect project modal. In global mode where there is no active project, show "Continue" regardless.

marktnoonan:
- Closes #23022

### User facing changelog

If the user logs in from the App, but the current project does not have a project ID, the "Continue" button will we replaced with a "Connect project" button in the login confirmation modal. Clicking the "Connect project" modal will move the user to the Connect Project flow. Closing the modal with the X or by clicking away will _not_ have this effect, the user will be back in the app as before.

### Additional details

This change helps users of the Dashboard successfully connect their project after they have logged into the Dashboard into the app, since logging in does not provide a benefit unless the project is actually connected.

### Notes on implementation

This is is mostly driven by props and events. The Login Modal is a shared component between the App and the Launchpad, but the "Connect Project" flow is only currently possible in the App, so we are using a prop to tell the Login Modal whether it is allowed to show the "Connect project" button at that end, and letting the HeaderBarContent component determine whether it is in the app or the launchpad - which is something it already infers in deciding what content to show in the header.

This could be cleaned up more (little bit of "prop drilling"), but the next step for this feature is to move the Connect Project modal and related code into `frontend-shared` the props will go away, making it possible to connect a project from anywhere if a project is open. As it stands, the props/events make this easy to test.

### Alternatives considered

Simply placing the existing CloudConnectButton into the LoginModal and moving everything to shared seemed like it could be a quick win, but there ends up being a circular dependency since the CloudConnect flow also contains a LoginModal, so this led to errors.

The next ticket for this behavior will be to sort that out into more of a system, since really, if the LoginModal moves you to connect if necessary, it's possible to combine these components. Just a bigger change than needed for the immediate goal, so I'll make a new ticket for it. 

### Steps to test

- Open any Cypress project
- Remove the `projectId` from `cypress.config.ts` if there is one
- Log out if you are logged in
- Log in through the **header in the app**
  - See the "Connect project" button
  - Click the button and see the "create project" modal open
- Log out, log back in through the tooltips on the **specs list page**
  - See the "Connect project" button
  - Click the button and see the "create project" modal open
- Log out, log back in through **the Runs page** "Log in to the Cypress Dashboard" button
  - See the "Connect project" button
  - Click the button and see the "create project" modal open
- Re-add the `projectId` or add a new one in your config
- Log in through the same places and see that the button says Continue
- Log in through the Launchpad and see that the button never says "Connect project" even if you don't have a project id set. 

New state after login if a project is not connected:

Screen Shot 2022-08-11 at 10 26 34 PM

### PR Tasks


packages/app/src/specs/SpecsList.vue
```diff
@@ -153,7 +153,9 @@
     v-model="isLoginOpen"
     :gql="props.gql"
     :utm-medium="loginUtmMedium"
-    @loggedin="refreshPage"
+    show-connect-button-after-login
+    @loggedin="handleLoggedin"
+    @connect-project="isProjectConnectOpen = true"
```
lmiller1990, line 158: I am surprised this works - should this not be a callback? `() => isProjectConnectOpen = true`?

marktnoonan: I've always just thought of these inline expressions as idiomatic Vue (like this [`count++` example](https://vuejs.org/guide/essentials/event-handling.html#inline-handlers) in the docs) - same syntax is valid for `onclick` attribute in HTML. 

We handle events this way plenty already (eg `@show-spec-pattern-modal="showSpecPatternModal = true"`, and several similar things). There are a couple reasons I prefer the non-callback version but they are all pretty marginal, we could standardize on something else.

packages/app/src/specs/SpecsList.vue
```diff
@@ -153,7 +153,9 @@
     v-model="isLoginOpen"
     :gql="props.gql"
     :utm-medium="loginUtmMedium"
-    @loggedin="refreshPage"
+    show-connect-button-after-login
```
marktnoonan, line 156: Going to test this logic a bit more myself, initially I figured just always showing it when you've logged in via the tooltip makes sense, but actually should probably still confirm no project ID, it's valid to log in from the tooltip into a connected project

packages/frontend-shared/src/gql-components/Auth.vue
```diff
@@ -88,6 +91,7 @@ const props = defineProps<{
   showRetry?: boolean
   showLogout?: boolean
   utmMedium: string
+  showConnectButtonAfterLogin?: boolean
```
mike-plummer, line 94: I'm definitely over-thinking this, but from a prop structure perspective we could make this a bit more flexible (and shorter name) by defining more like `postLoginAction: 'none' | 'connect-project'`. I wouldn't be surprised if we want to support A/B flows through login in the future, so having the more flexible prop structure might come in handy

packages/frontend-shared/src/gql-components/HeaderBarContent.vue
```diff
@@ -273,6 +275,16 @@ const props = defineProps<{
   allowAutomaticPromptOpen?: boolean
 }>()
 
+const emit = defineEmits<{
+  (event: 'connect-project'): void
+}>()
+
+const isApp = window.__Cypress__
```
mike-plummer, line 282: TIL 👍

packages/app/src/specs/SpecsList.vue
```diff
@@ -153,7 +153,9 @@
     v-model="isLoginOpen"
     :gql="props.gql"
     :utm-medium="loginUtmMedium"
-    @loggedin="refreshPage"
+    :show-connect-button-after-login="!cloudProjectId"
```
marktnoonan, line 156: I need to add a test for this specs list tooltip flow.

packages/app/src/specs/SpecsList.vue
```diff
@@ -307,6 +309,10 @@ const emit = defineEmits<{
   (e: 'showCreateSpecModal'): void
 }>()
 
+const cloudProjectId = computed(() => {
+  return props.gql?.currentProject?.config?.find((item: { field: string }) => item.field === 'projectId')?.value
```
mike-plummer, line 313: Could we just add the `projectId` field to the fields being queried under `currentProject` and reference that?

packages/app/src/runs/CloudConnectButton.vue
```diff
@@ -63,4 +69,8 @@ function openConnection () {
     isProjectConnectOpen.value = true
   }
 }
+
+const cloudProjectId = computed(() => {
+  return props.gql?.currentProject?.config?.find((item: { field: string }) => item.field === 'projectId')?.value
```
mike-plummer, line 74: Just query for `projectId` on `currentProject`?

packages/frontend-shared/src/gql-components/Auth.vue
```diff
@@ -202,6 +213,26 @@ const handleCancel = () => {
   emit('continue', true)
 }
 
-const { t } = useI18n()
+const buttonText = computed(() => {
+  const strings = {
+    login: t('topNav.login.actionLogin'),
+    connectProject: t('runs.connect.modal.selectProject.connectProject'),
+    continue: t('topNav.login.actionContinue'),
+  }
```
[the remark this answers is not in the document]
marktnoonan: I just found it a bit more expressive to read the returns & the logic if they weren't these long paths, if it's just "condition -> name of string that matches condition"

packages/app/src/specs/SpecsList.vue
marktnoonan: let me double check, I have a terrible feeling that that project ID is a different thing than this

packages/frontend-shared/src/gql-components/Auth.vue
marktnoonan: Yeah this can change quite a bit in the next sprint, since we'll be making it work from the launchpad as well. It's possible the component will end up just knowing how to check this on its own and not use a prop at all, and be somewhat combined with CloudConnect stuff, since all logins will then potentially lead to CloudConnect. So a whole bunch of stuff can simplify then I think. I don't know if I'd worry about it in this PR.

packages/app/src/specs/SpecsList.vue
marktnoonan: Sweet, yeah this is an improvement, I'll make the change. We've definitely overburdened the words "project" and "id". I feel like there is a 3rd thing that's not in this screenshot that represents the unique ID of a local project, as in a "directory we've opened in Cypress" not a project in the cloud sense.

Screen Shot 2022-08-12 at 2 44 50 PM

packages/app/src/layouts/default.vue
```diff
@@ -48,6 +49,13 @@
           <component :is="Component" />
         </transition>
       </router-view>
+      <CloudConnectModals
+        v-if="showConnectDialog && query.data.value"
+        :show="showConnectDialog"
+        :gql="query.data.value"
+        @cancel="showConnectDialog = false"
+        @success="showConnectDialog = false"
+      />
```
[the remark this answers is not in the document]
mike-plummer: Ah, looked at this a bit deeper and it wasn't quite doing everything I thought it was - we've still got multiple instances of `CloudConnectModals`. Just thinking out loud here, no action expected -I wonder whether it would be viable to consolidate to the single instance and control via new global state as part of the upcoming work we're doing where we'll be touching these modals again, or whether that would make things a bit too tightly coupled thru state 🤔

packages/frontend-shared/src/gql-components/Auth.vue
```diff
@@ -177,6 +182,12 @@ const loginMutationIsPending = computed(() => {
 })
 
 const handleLoginOrContinue = async () => {
+  if (cloudViewer.value && props.showConnectButtonAfterLogin) {
```
mike-plummer, line 185: /nitpick: We could eliminate the duplicate `cloudViewer.value` check here

```
if (cloudViewer.value) {
  if (props.showConnectButtonAfterLogin) {
    emit('connect-project')
  } else {
    emit('continue', true)
  }
  return
}
```

mike-plummer (approved):
Works as expected for me for all the scenarios I was able to think of. Nice!

packages/frontend-shared/src/gql-components/Auth.vue
marktnoonan: this is style preference for me, I avoid nesting conditionals out of habit if it's not a lot of trouble, but yeah this may be fine

marktnoonan: updated this and one other place, all goo

packages/frontend-shared/src/graphql/urqlClient.ts
```diff
@@ -108,6 +108,7 @@ declare global {
       namespace: AutomationElementId
       hideCommandLog: boolean
     }
+    __Cypress__: boolean
```
ZachJW34, line 111: Not related to this PR but it feels weird that our global Window definitions are in a `graphql` specific file. `__Cypress__` is derived from the `index.html` file.

packages/app/src/specs/SpecsList.vue
```diff
@@ -487,6 +489,15 @@ function refreshPage () {
   location.reload()
 }
 
+const handleLoggedin = () => {
```
ZachJW34, line 492: Why is the `location.reload` only necessary on the `SpecsList` page? You can log in on the runs page that has cloud data, does that not need a refresh?

ZachJW34 (approved):
Tested a works like a charm. Got minor questions that are non-blocking

mike-plummer:
CI is green and this is looking good in my local testing, going to go ahead and merge to get this in for next release. @marktnoonan Zach had a few non-blocking questions on this PR, could you answer those when you have a chance?

packages/frontend-shared/src/graphql/urqlClient.ts
marktnoonan: Yeah it's weird. I wouldn't object to moving this to anywhere that's more universal, I'd like to keep it all together though.

packages/app/src/specs/SpecsList.vue
marktnoonan: This is a good point and I don't really know the answer - I believe this was added with the first pass at the ACI work. I'd say this should be checked more thoroughly and maybe we are missing something - or maybe we don't really need the refresh. Logging in from the header on the Specs List page does _not_ trigger a refresh, only logging in from the tooltip does, which doesn't make sense if the refresh is needed.

I think next sprint we should consolidate Login and Cloud Connect into a single top-level component in App opened/closed via a store, that behaves the same way no matter where you triggered it from. I'm going to make that ticket and I'll add this "confirm when and why we need to reload the page for the user" point & probably get with @mike-plummer about the context there.

mike-plummer: @marktnoonan @ZachJW34 I don't have a firm answer to why this was initially added - my suspicion is that either `SpecHeaderCloudDataTooltip.vue` or `SpecsList.vue` needs a `cloudViewerChange` subscription (see `CloudConnectModals.vue`) so that the tooltip state is properly updated after login. This `reload` is very likely a workaround to get data looking "correct" by forcing a larger re-query

marktnoonan: oh cool. when moving this to frontend-shared we can perhaps get the subscription wrapped up in the overall login process, or at least lifted up in App such that it would trigger from logging in on the runs page via the header, etc.

PR #23298, opened 2022-08-11
feat: Enable create from Vue component for projects with custom spec patterns

### LINKED ISSUE #23071: Add 'create from Vue component' card to projects with custom spec patterns
When #22836 is completed, users will be able to generate basic component tests from a list of their Vue components. However, this isn't enabled on projects that have a custom `specPattern` specified in their Cypress configuration.

Add support for custom spec patterns.

This should be completed before the next release so that support all goes out at once.

astone123:
- Closes #23071

### User facing changelog

This is a part of the create from Vue component effort. It shouldn't have its own changelog entry

### Additional details

In the last create from Vue component PR, I excluded projects with custom spec patterns to separate out complexity. In order to release the feature, this also needs to work for projects with custom spec patterns.

### Steps to test

1. Open a Vue project that has a custom spec pattern defined for CT
2. If your project has specs, click on the "+ New Spec" button
3. Verify that you can see the "Create from component" card
4. Click on the card and then choose a Vue component from the list.
5. Verify that the spec file that is created matches your project's spec pattern and that the component you selected is imported correctly into the spec file (best way to verify this is to run the new spec and ensure that the spec mounts correctly).

Users will now see the "Create from component" generator card when adding a new spec to a Vue project using component testing **when the project has a custom spec pattern** defined in its configuration.

### PR Tasks


system-tests/projects/no-specs-custom-pattern-vue/babel.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
astone123, line 1: The diff looks really big because I added a new project to system-tests. I needed a project that had a custom spec pattern but was a Vue project.

You should be safe to ignore any changed files under `system-tests/projects/`

packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -531,4 +539,52 @@ export class ProjectActions {
     // @ts-ignore - because of the conditional above, we know that devServer isn't a function
     return WIZARD_FRAMEWORKS.find((framework) => framework.configFramework === config?.component?.devServer.framework)
   }
+
+  async getDefaultSpecFileName (
```
astone123, line 543: I moved this from `ProjectDataSource` to here and passed anything that we needed from context as props instead. This way I could reuse the logic for determining the correct file path for a spec.

ZachJW34: This function doesn't rely on `ctx` (or rather `this.ctx`). Refactoring this into a utils file would allow the `SpecOptions` class to simply import it and you wouldn't have to pass a function in.

packages/data-context/src/codegen/spec-options.ts
```diff
@@ -3,11 +3,14 @@ import type { CodeGenType } from '@packages/graphql/src/gen/nxs.gen'
...
 import path from 'path'
+import type { FoundSpec, TestingType } from '@packages/types'
 
 interface CodeGenOptions {
   codeGenPath: string
   codeGenType: CodeGenType
   isDefaultSpecPattern: boolean
+  specPattern: string[]
+  currentProject: string | null
   erroredCodegenCandidate?: string | null
```
ZachJW34, line 14: Now that we support custom spec pattern, can we confirm that there is no use for `erroredCodegenCandidate`? I'd really like to see this removed as it's pretty confusing.

ZachJW34 (commented):
Gave it a thorough test and it worked great! Just a few comments

packages/data-context/src/codegen/spec-options.ts
astone123: Yeah honestly I've held off on removing the logic surrounding that because I'm not sure what it does. I'll take a look at it now and see about removing it

packages/app/cypress/e2e/create-from-component.cy.ts
```diff
@@ -0,0 +1,189 @@
...
+
+        cy.findByRole('link', { name: 'Okay, run the spec' })
+        .should('have.attr', 'href', `#/specs/runner?file=src/components/HelloWorld.cy.js`).click()
+      })
+
+      cy.findByText('<HelloWorld ... />').should('be.visible')
+    })
+  })
+
+  context('project with custom spec pattern', () => {
```
mike-plummer, line 97: This feels like we're duplicating the same tests, just with a different specPattern arg. Could we extract them to a reusable function or within a loop so they can be run with variable setup/args?

```
const validateComponentCard = (configFile: string, expectedLocation: string) => {
  beforeEach(() => {
      cy.scaffoldProject('no-specs-vue-2')
      cy.openProject('no-specs-vue-2', ['--config-file', configFile])
      cy.startAppServer('component')
      cy.visitApp()

      cy.findAllByTestId('card').eq(0).as('ComponentCard')
    })

    it('Shows create from component card for Vue projects with default spec patterns', () => {
...
}

context('project with default spec pattern', () => {
  validateComponentCard('configFile1', 'my-awesome-location-1')
})

context('project with custom spec pattern', () => {
  validateComponentCard('configFile2', 'my-awesome-location-2')
})
```

packages/data-context/src/codegen/spec-options.ts
```diff
@@ -67,29 +66,45 @@ export class SpecOptions {
...
 
-    const componentPathRelative = path.relative(this.parsedPath.dir, this.parsedErroredCodegenCandidate.dir)
+      const componentPath = path.join(componentPathRelative, this.parsedPath.base)
 
-    const componentPath = path.join(componentPathRelative, this.parsedErroredCodegenCandidate.base)
+      return componentPath.startsWith('.') ? componentPath : `./${componentPath}`
+    }
 
-    return componentPath.startsWith('.') ? componentPath : `./${componentPath}`
+    return `./${this.parsedPath.base}`
```
mike-plummer, line 78: Would `./` cause any issues for Windows?

rockindahizzy: Would using`path.join('./', componentPath)`resolve any issue with os specifics?

ZachJW34: This is for the component import path in the component e.g. `import App from '<relative-path>'.js`. These should always be forward slashes, but still something we should check on Windows since we might be putting backslashes in the path where they shouldn't be.

lmiller1990: I think handling the module import paths is actually a concern of the **bundler** (eg, webpack and vite) so likely it's not something we need to think about.

lmiller1990: Something similar here: https://github.com/cypress-io/cypress/blob/develop/packages/data-context/src/sources/migration/codegen.ts#L269-L275

packages/data-context/src/codegen/spec-options.ts
```diff
@@ -112,23 +127,21 @@ export class SpecOptions {
...
     const foundSpecExtension = expectedSpecExtensions.find((specExtension) => {
       return this.parsedPath.base.endsWith(specExtension + this.parsedPath.ext)
     })
 
     return foundSpecExtension || ''
   }
 
-  private async buildComponentSpecFilename (specExt: string) {
-    const { dir, base, ext } = this.parsedPath
+  private async buildComponentSpecFilename (specExt: string, filePath?: ParsedPath) {
```
lmiller1990, line 137: No need to mark as `async`

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -134,7 +135,7 @@ export function getLongestCommonPrefixFromPaths (paths: string[]): string {
   return lcp.slice(0, endIndex).join(path.sep)
 }
 
-export function getPathFromSpecPattern (specPattern: string, testingType: TestingType, fileExtensionToUse?: 'js' | 'ts') {
+export function getPathFromSpecPattern (specPattern: string, testingType: TestingType, fileExtensionToUse?: 'js' | 'ts', name = '') {
```
lmiller1990, line 138: When we have this many positional arguments, some required, some not, what do you think about moving to an object of options?

```js
interface GetSpecOptions {
  specPattern: string
  testingType: TestingType
  fileExtensionToUse?: 'js' | 'ts'
  name?: string
}
```

Edit: Looks like we've basically got this [here](https://github.com/cypress-io/cypress/pull/23298/files#diff-c351c3ec7e69c56e8c57fdf49babbece37086dbb43b9fc07a386f3cd33e508c9R17), should we just extract an interface and share it?

Also is `fileExtensionToUse` even nullable? It doesn't look like it should be.

packages/data-context/src/sources/migration/utils.ts
```diff
@@ -7,3 +12,53 @@ export const isDefaultSupportFile = (supportFile: string) => {
...
+export async function getDefaultSpecFileName (
+  { currentProject, testingType, fileExtensionToUse, specPattern, specs = [], name }:
+  { currentProject: string | null, testingType: TestingType | null, fileExtensionToUse: 'js' | 'ts', specPattern: string[], specs?: FoundSpec[], name?: string },
+): Promise<string> {
+  const debug = Debug('cypress:data-context:sources:migration:utils')
+
+  const defaultFilename = `${name ? name : testingType === 'e2e' ? 'spec' : 'ComponentName'}.cy.${fileExtensionToUse}`
+  const defaultPathname = path.join('cypress', testingType ?? 'e2e', defaultFilename)
+
+  if (!currentProject || !testingType) {
```
lmiller1990, line 25: Another more concise way that seems more in line with TypeScript's recommendations is an [assert function](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions). Example 

https://github.com/cypress-io/cypress/blob/develop/packages/data-context/src/actions/ProjectActions.ts#L159-L160

```ts
import assert from 'assert'

assert(currentProject, 'currentProject is required')
```

etc.

That said, do we even need this? Instead of throwing and returning here, should we just do

```ts
if (!currentProject || !testingType) {
    debug('currentProject or testingType undefined. Error intelligently detecting default filename, using safe default %o', err)
    return defaultPathname
}
```

lmiller1990 (changes requested):
I found an inconsistency: it's using `"` for the component import but `'` for the describe/it blocks. Feels a bit bad, quick fix? :pray: 

image

packages/data-context/src/sources/ProjectDataSource.ts
astone123: I ended up not using that interface in both places because the types were slightly different

ZachJW34 (changes requested):
Found a bug while creating a spec from a component that already has an existing spec when using a custom spec pattern. The `copy` gets appended after the `.cy`, causing the spec to not match.


https://user-images.githubusercontent.com/25158820/184662128-9d3812fb-7f65-41b4-a7e6-22da6c38158e.mov

packages/data-context/src/codegen/spec-options.ts
ZachJW34: The complications that arise from a custom spec-pattern make me think that dumping the user into the empty-generator flow might be the best way to solve this problem

marktnoonan: Empty-generator is what's happening now, so that sounds OK.

astone123:
@ZachJW34 it looks like that issue already exists currently for create empty spec. If you try to create a file with the same name, we append "copy" and it ends up not matching the spec pattern. We don't give a warning there either.

I wonder if we could push this bug fix off into a separate issue since it already exists and is common logic between the two cards.

ZachJW34:
> @ZachJW34 it looks like that issue already exists currently for create empty spec. If you try to create a file with the same name, we append "copy" and it ends up not matching the spec pattern. We don't give a warning there either.
> 
> I wonder if we could push this bug fix off into a separate issue since it already exists and is common logic between the two cards.

What is the reproduction for the empty spec bug? I gave it a try and it seemed to be working. I ran through empty-spec generation twice for `cypress/component/ComponentName.cy.js` (the default) and got:
- src/ComponentName.cy.js
- src/ComponentName-copy-1.cy.js

This is the expected output and it matches the specPattern. For the create-from-component, the output was `src/App.cy-copy-1.js` where the `cy` and `copy` are swapped.

Also, I posted a comment regarding Windows import paths: https://github.com/cypress-io/cypress/pull/23298#discussion_r945873250

astone123:
@ZachJW34 Oh right, you wouldn't be able to see the bug with that spec pattern. I was using `'src/specs-folder/*-cypress-spec.js'` as my spec pattern, and with both methods of generation I saw the issue.

So in order to see it with empty, you need a custom suffix for your file names. I see that it's a different issue for create from component though, so I'll check it out

packages/data-context/src/codegen/spec-options.ts
astone123: > Gave it a try on my Windows and a spec that is generated using a custom spec-pattern

@ZachJW34 yeah I think this is happening because when I use `path` in `spec-options` it takes the OS into account, so the path it gives me has backslashes in it on Windows 🤔 

Kind of hacky, but what do we think about just replacing `\` with `/` for these relative paths? I still need to use Node `path` to get the relative path of the component module compared to the spec directory, but I don't want it to use OS-specific slashes because this is a relative import being written to a file

lmiller1990 (approved):

lmiller1990:
My changes were responded to!

packages/data-context/src/codegen/spec-options.ts
lmiller1990: Maybe we can use this function https://github.com/cypress-io/cypress/blob/c40a120aa7e3a0c52016c89ab316ce1e5eb8fa69/packages/data-context/src/util/file.ts#L3-L5

astone123: Thanks Lachlan, [this](https://github.com/cypress-io/cypress/pull/23298/commits/ce4c883c0b174ce66ecfe7b2656ec5274fbcb7b5) should fix the issue on Windows

ZachJW34: Yeah the replacement makes sense, we have a few examples of `posixify`-ing paths in the codebase

lmiller1990:
I found another edge case:

1. Go to `frontend-shared`
2. Change `specPattern` to `**/*.test.js`
3. Go to open mode
4. It should show no specs found (since you changed the pattern)
5. Create a spec for Alert.vue
6. It creates a spec in `Alert.test.js` (top level) but the UI shows "Created in `src/components/Alert.test.js`.

lmiller1990:
Dismissing my review since I found a bug.

lmiller1990 (changes requested):
https://github.com/cypress-io/cypress/pull/23298#issuecomment-1215981425

May need a mini brief around the edge cases, I am not sure the expected filename and path is clear (above behavior is expected based on current code, but certainly not what you'd want).

lmiller1990:
I found an edge case. Here's my expected output when creating a new spec:

| specPattern | existing specs | component     | expected new spec |
| ---------------- | ------------------- | ------------------ | -------------------------- | 
| `src/**/*.js`   | n/a                   |  src/Foo.vue   | src/Foo.cy.js            |
| `**/*.js`         | n/a                   |  src/Foo.vue   | src/Foo.cy.js           |
| `**/*.js`         | src/Bar.cy.js     |  src/Foo.vue   | src/Foo.cy.js           |
| `**/*.js`         | src/Foo.cy.js     |  src/Foo.vue   | src/Foo-copy-1.cy.js  |


Case 2 is problematic. I think it's because we use https://github.com/cypress-io/cypress/pull/23298/files#diff-c351c3ec7e69c56e8c57fdf49babbece37086dbb43b9fc07a386f3cd33e508c9R52. When I created a spec in a new project where the spec pattern is `**/*.cy.ts`, I chose `src/Alert.vue` for the spec, and it's creating it at the top level - `Alert.cy.tsx`, which is not really what you'd expect.

packages/data-context/test/unit/codegen/spec-options.spec.ts
```diff
@@ -76,49 +75,65 @@ describe('spec-options', () => {
...
-          codeGenType: 'component',
-          isDefaultSpecPattern: false,
-          framework: WIZARD_FRAMEWORKS[1],
-          specPattern,
+      context('custom spec pattern', () => {
+        afterEach(function () {
+          sinon.restore()
+        });
+
+        [{ testName: 'src/specs-folder/*.cy.{js,jsx}', componentPath: 'ComponentName.vue', specs: [], pattern: 'src/specs-folder/*.cy.{js,jsx}', expectedPath: 'src/specs-folder/ComponentName.cy.js' },
```
lmiller1990, line 103: Nice way to cover all the cases

lmiller1990:
Just checking if we need a re-review or should this go back to in progress? Also might be good to clarify the test plan for the various edge cases (eg what we will do for `**/*.test.js` for CT projects without any specs, etc). :pray:

astone123:
Okay, after looking at this some more I updated the assertions for the unit tests and fixed a couple things with extensions. This doesn't match the table that @lmiller1990 put in an earlier comment, but these do match the default file name for the empty spec generator. 

I'm hoping that this is good enough for now, as the plan for this wasn't to change the underlying `defaultSpecFileName` logic that we use. If we want to change our default spec name logic then I think we should follow up with another issue, or we should update this generator to allow the user to input their own file path for the spec.

To get an idea of the different cases that we have test coverage for and which file paths they result in, check out this [test block](https://github.com/cypress-io/cypress/pull/23298/commits/04d865620859e70577f5043ea7a30b28cac5dead#diff-675a2b4d978410448068ca3ce0cedab44ff671f179c85883733f1430570049c7R121)

@ZachJW34 can you verify that you can create duplicate files and the `-copy` doesn't get added to your file extension now?

ZachJW34 (approved):
My comments have been addressed!

One last thing I'd like to throw out, would it make sense to create the sibling path first and check to see if it matches the specPattern before calling the `getDefaultSpecFileName`? This would reduce the likelihood that a spec gets generated away from it's sibling component when using a specPattern like `**/*.cy.js`

Just a thought, going to ✅

lmiller1990:
I also verified, copy naming bug is gone, I think this is good for now - I do expect this feature to receive lots of updates moving forward, we can address/rejig/rethink more when we get the next requirement/stories. Nice job, this one sure turned out to have some hidden complexity.

lmiller1990 (approved):
LGTM

Will leave merging to @astone123 since I think there's one more potential (optional) change that @ZachJW34 pointed out that you may or may not want to do - I'm thinking we ship this and focus on Studio/other stuff for now, though?

astone123:
> My comments have been addressed!
> 
> One last thing I'd like to throw out, would it make sense to create the sibling path first and check to see if it matches the specPattern before calling the `getDefaultSpecFileName`? This would reduce the likelihood that a spec gets generated away from it's sibling component when using a specPattern like `**/*.cy.js`
> 
> Just a thought, going to ✅

Yeah, I started on this but it ended up being more complicated than I originally thought, so I decided that for now we should just go with the behavior of the existing spec generator card. Will open an issue to improve this

PR #23299, opened 2022-08-11
fix: remove webpack entrypoints from for ct webpack projects

### LINKED ISSUE #23224: Components testing - Webpack entry feature causes confusion
### Current behavior

when loading app's webpack config cypress will build the app using the `entry` configuration of webpack. Then the bundled JS is injected in cypress when loading a component.

The issue is that with many default webpack configs will create errors in this configuration because the bundle will be a JS file that tries to render the app in the html. For example `ReactDom.render(<App/>, someElement);`. That is not a desired behavior for cypress components testing.

### Desired behavior

@ZachJW34 proposed:

> I think there could be a configuration option that would make sense for this. I imagine most people who provide a custom webpack config would want their apps entrypoint pruned

---

I would add that I understand the need to give an opportunity to inject global styles and global js stuff globally but I think it should at the minimum be mentioned in the doc and in the getting started guide to make sure people don't spend a long time figuring out why it's not working

### Test code to reproduce

Clone that repo: https://gitlab.com/souf/react-webpack-playground/-/blob/cypress-webpack-entry/webpack.config.js using branch `cypress-webpack-entry`.

Run `yarn install` then `yarn cypress:open`. Open the component testing for the component `App`. Observe the error in the console triggering because the app tries to render in cypress's frame without a match DOM element for the given id.

image

### Cypress Version

10.4.0

### Other

_No response_

ZachJW34:
- Closes #23224

### User facing changelog
Webpack entrypoints are no longer preserved for CT Webpack projects

### Additional details
Fixes a regression from https://github.com/cypress-io/cypress/pull/22314 where webpack entrypoints were being preserved. This is mostly a revert of the changes made to entrypoints in that PR. It made sense for frameworks as their was logic added for removing entrypoints that would cause problems in each handler, but users who pass in a custom webpack config (don't use a higher-order framework like Next or create-react-app) were facing issues with their entrypoint being included and executed.

### Steps to test
Checkout the repo provided in the original issue: https://gitlab.com/souf/react-webpack-playground/-/tree/cypress-webpack-entry

Install Cypress and walk through CT setup.

Create a simple spec:

```js
// src/components/App.cy.jsx
import React from "react";
import App from "./App";

it("should render", () => {
  cy.mount(<App />);
});
```

Then run the spec. Nothing will load, and the console will display:

Screen Shot 2022-08-11 at 5 09 20 PM

Checkout my branch and run `yarn workspace @cypress/webpack-dev-server build && yarn cypress:open`. Open the same project and no more error!

https://user-images.githubusercontent.com/25158820/184251849-0a3145ca-9b7e-472d-b920-be401c79c4c7.mov

### PR Tasks

### DIFF
--- a/npm/webpack-dev-server/src/helpers/vueCliHandler.ts
+++ b/npm/webpack-dev-server/src/helpers/vueCliHandler.ts
@@ -17,8 +17,6 @@ export function vueCliHandler (devServerConfig: WebpackDevServerConfig): PresetH
 
     debug('webpack config %o', webpackConfig)
 
-    delete webpackConfig.entry
-
     return { frameworkConfig: webpackConfig, sourceWebpackModulesResult }
   } catch (e) {
     console.error(e) // eslint-disable-line no-console
--- a/npm/webpack-dev-server/src/helpers/createReactAppHandler.ts
+++ b/npm/webpack-dev-server/src/helpers/createReactAppHandler.ts
@@ -77,8 +77,6 @@ function loadWebpackConfig (devServerConfig: WebpackDevServerConfig): Configurat
       webpackConfig = webpackConfig('development')
     }
 
-    delete webpackConfig.entry
-
     return webpackConfig
   } catch (err) {
     throw new Error(`Failed to require webpack config at ${webpackConfigPath} with error: ${err}`)
--- a/npm/webpack-dev-server/src/helpers/nextHandler.ts
+++ b/npm/webpack-dev-server/src/helpers/nextHandler.ts
@@ -72,8 +72,6 @@ async function loadWebpackConfig (devServerConfig: WebpackDevServerConfig): Prom
     },
   )
 
-  delete webpackConfig.entry
-
   return webpackConfig
 }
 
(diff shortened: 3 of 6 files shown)


npm/webpack-dev-server/src/helpers/vueCliHandler.ts
```diff
@@ -17,8 +17,6 @@ export function vueCliHandler (devServerConfig: WebpackDevServerConfig): PresetH
 
     debug('webpack config %o', webpackConfig)
 
-    delete webpackConfig.entry
```
lmiller1990, line 20: Instead of deleting it in many places, and then having to remember this if we implement a new webpack handler, any chance we could lift this up higher and do something like

```js
if (!angular) {
  delete webpackConfig.entry
}
```

Thoughts?

lmiller1990 (commented):
One question/comment. This must go in the next release!

npm/webpack-dev-server/src/helpers/vueCliHandler.ts
ZachJW34: This PR is doing exactly that. In #22314 I moved entry handling to each handler. This PR reverts those changes so that the handlers don't have to worry about entrypoints since they will be replaced with the Cypress entry [here](https://github.com/cypress-io/cypress/pull/23299/files#diff-7031da07fd51256fb2dc14db39454b4e8ec0f8f90ab283511c2bf847e790eead)

The line you referenced is a removal, not an addition (unless I'm misunderstanding your comment)

astone123 (approved):
I tested this with the example project and these changes resolve the webpack entrypoint issue 👍🏻

rockindahizzy (approved):

PR #23325, opened 2022-08-12
fix: match full spec path and improve search results

### LINKED ISSUE #21879: Fuzzy search results on specs list are not what is expected
## **Current Behavior**
The fuzzy search in the specs list makes absolutely no sense for what I’m actually looking for. 

If I search for ‘test’, the result brings up ne*t_work_r_e_que_st*s.cy.js. 

I would never want this or expect this in my search and it muddies the results. 

Screen Shot 2022-05-02 at 9 38 30 AM

## **Desired Behavior**
* We should use the same search algorithm we had in 9.x as it was fine and no one opened issues about it and you could always find your specs. 

## **Reproducible Steps**
1. Scaffold kitchensink files
1. Search for ‘test

┆Issue is synchronized with this [Jira Bug](https://cypress-io.atlassian.net/browse/UNIFY-1702) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/11920">Screen Shot 2022-05-02 at 9.38.30 AM.png</a>
┆author: Jennifer Shehane
┆epic: 9.x Parity
┆friendlyId: UNIFY-1702
┆priority: High
┆taskType: Bug

### LINKED ISSUE #22463: Search cannot find paths with filenames
### Current behavior

When searching in the Cypres UI for a file, if the full path is used, Cypress cannot find the file. When just the folder is searched, Cypress can find the files. Also, if just the filename is used as a search parameter, then Cypress can find the test file.

For example

`cypress/e2e/spec.cy.js` -> will not be found

`cypress/e2e/` -> will not find any results

`cypress/e2e` -> will find the files in the e2e folder

`spec.cy.js` -> will find all tests with the same name

### Desired behavior

Cypress should allow searching with the full path of a file and be able to find any test file if the filename+full relative path is passed as a search parameter.

### Test code to reproduce

https://github.com/DanielYang-AT/cypress-test-tiny
(this is a fork of the cypress-test-tiny with 0 changes)

* Run Cypress Open
* Select e2e testing
* Open with chrome or electron
* search for a spec by pasting in `cypress/e2e/spec.cy.js`
* no results are found when the example spec should be shown.

### Cypress Version

10.2.0

### Other

Cypress 9.x.x and below were all able to find the files as expected. This was a change after the upgrade to 10.x.x

ZachJW34:
- Closes #22463 
- Closes #21879 

### User facing changelog
Allows full path matching in specs list and reduces the "fuzziness" of the search

### Additional details
We couldn't match against the full path due to the fuzzy-sort keys being bound to "directory" and "baseName". A full path will only match the "relative". I've switched it to "relative" and "basename". Keeping the "basename" in the search allows the filenames to be highlighted even if the directory contains matching characters, similar to VS code directory search. I've also limited the `threshold` of the fuzzysort to reduce the noisy results returned. There is still some fuzziness, but drastically reduced.

I refactored how the indexes are calculated to be an up-front cost. This might slow down the tree-building a bit but reduces the cost of calculating them per row render which I think is a net win and is in sync with how we would tackle search if/when we move it to be server-side.

### Steps to test
Pull up a project with a decent amount of tests and start typing in the search to get a feel for the new behavior. `packages/app` CT and `packages/driver` e2e have a decent amount of tests.

Before:

https://user-images.githubusercontent.com/25158820/184447613-09219177-47b9-405b-937c-041ade25fca5.mov

After:

https://user-images.githubusercontent.com/25158820/184447643-2e3b55e6-b951-4287-a996-242ea92afd6b.mov

### PR Tasks


ZachJW34:
@mapsandapps Wanted to add you to the review since I'm changing the highlighting behavior of the spec list. Overall, it is more inline with VS Code

packages/app/src/specs/SpecItem.vue
```diff
@@ -28,11 +28,16 @@
...
 
-withDefaults(defineProps<{
+const props = withDefaults(defineProps<{
   fileName: string
   extension: string
   indexes?: number[]
 }>(), { indexes: () => [] })
+
+const refs = toRefs(props)
+const split = useSplitIndexes(refs.fileName, refs.indexes)
```
lmiller1990, line 42: It seems unusual for a `useXXX` function to return a `computed` property. You needed to do `props` -> `toRefs` -> `computed`. Also, it leads to the logic (when to split/join) mixing with the UI concerns (the reactivity layer).

Instead, why not just:

```js
const split = computed(() => {
  return props.indexes.reduce((acc, curr) => ...)
})
```

Since you want to share the logic and re-use it, just make a stateless function (no need to leak reactivity into it):

```js
const split = computed(() => {
  return deriveIndexes(props.fileName, props.indexes)
})
```

What do you think? Extra bonus of this is, if you decided you did want to unit test the `deriveIndexes` (eg, it became more complex) you wouldn't need a UI framework or reactivity system to do so.

lmiller1990 (commented):
One comment on the separation of concerns - logic/reactivity layer. I tried it out and the new sorting feels much better, I was able to find my file more easily using the same style of search I do in VS Code (partially typing folder + filename, and it's right there) :tada:.

packages/app/src/specs/SpecItem.vue
ZachJW34: Your suggestion is exactly what I was looking for! Going to implement

ZachJW34: Refactored in [cb98862](https://github.com/cypress-io/cypress/pull/23325/commits/cb9886239b6523f78bdd8f29e5d8f03b473bf5c1)

lmiller1990 (approved):
My changes were responded to, approved.

:heavy_check_mark:

packages/app/src/specs/InlineSpecList.cy.tsx
```diff
@@ -49,11 +49,16 @@ describe('InlineSpecList', () => {
...
-      cy.get('input').type('scome', { force: true })
+      cy.get('input').type('compspec', { force: true })
 
       cy.get('li').should('have.length', 4)
       .should('contain', 'src/components')
       .and('contain', 'Spec-A.spec.tsx')
+
+      // Don't want the search to be too fuzzy
+      // sr => src, co => components, sp => spec
+      cy.get('input').clear().type('srcose', { force: true })
```
astone123, line 60: Just for my knowledge - why do we have to use `force` here when typing into this input?

astone123 (approved):
This works much better than before and returns results more in line with what I would expect 👍🏻 just had one question

warrensplayer (approved):
The functionality works great and I like the algorithm being used to match.  My only request (but does not hold up my approval) would be to provide some method comments explaining the logic being used in new methods like `getHighlightIndexes` and `deriveIndexes` in `spec-utils.ts`.  I was able to debug through them to get an idea of what was going on, but documenting the inputs and basic logic would be helpful when someone is not familiar with the code.

packages/app/src/specs/InlineSpecList.cy.tsx
```diff
@@ -49,11 +49,16 @@ describe('InlineSpecList', () => {
...
       mountInlineSpecList()
-      cy.get('input').type('scome', { force: true })
+      cy.get('input').type('compspec', { force: true })
 
       cy.get('li').should('have.length', 4)
       .should('contain', 'src/components')
       .and('contain', 'Spec-A.spec.tsx')
+
+      // Don't want the search to be too fuzzy
+      // sr => src, co => components, sp => spec
```
mike-plummer, line 59: comment doesn't seem to quite line up with what's being typed - "sp" vs "se"

mike-plummer (approved):
Validated locally & looks good. Also ran this against a couple large projects with ACI data (since loading ACI data can trigger tree rebuilds on scroll), feels smooth.

One question (not for this PR) - have we ever considered doing a `trim` on search input? As I was trying to break search I noticed that even if I have the exact filepath in the search, I get no results if I prefix/suffix with a single space. Same is true if a space is in the middle of the string, but that's at least visually noticeable

packages/app/src/specs/InlineSpecList.cy.tsx
ZachJW34: Addressed in [06a2c94](https://github.com/cypress-io/cypress/pull/23325/commits/06a2c946f798336fdd3672c7b124c8ceaa872909)

packages/app/src/specs/InlineSpecList.cy.tsx
ZachJW34: Not sure, but I removed it and it's happy [06a2c94](https://github.com/cypress-io/cypress/pull/23325/commits/06a2c946f798336fdd3672c7b124c8ceaa872909)

ZachJW34:
@warrensplayer added some docs to the new/tweaked functions

@mike-plummer I considered adding the trim but it was going to tweak something that uses saved state and didn't want to blow out the scope of the PR. Worth adding an issue for trimming the search?

packages/app/src/specs/InlineSpecList.cy.tsx
mike-plummer: I suspect this is kinda-sorta my fault. The search input used to rely on absolute positioning to get the prefix & suffix elements aligned properly which resulted in DOM elements overlaying the `input`. I refactored that when adding the "X" button to clear the search input, but didn't notice these `force` calls could be removed at the same time

mike-plummer:
@ZachJW34 I'll get an issue written up for the trim

PR #23329, opened 2022-08-13
feat: Expose setSystemTime on clock object

### LINKED ISSUE #15424: Expose setSystemTime from lolex
### What would you like?
I would like to be able to do:

```javascript
cy.clock().then(clock => clock.setSystemTime(12345));
```

### Why is this needed?
I am testing a timer, and I realized that when I wanted to test it also displays hours correctly, I did cy.tick(4*60*60*1000) and Cypress froze because it was running thousands of setInterval callbacks in a row. `setSystemTime` allows me to skip forward in time, and I can still use tick for small increment testing.

I'm happy to do the implementing

emilgoldsmith:
- Closes #15424

### User facing changelog

Expose `clock.setSystemTime` from the underlying time mocking library. This allows you to skip ahead by large amounts of time without running what could be thousands of timer callbacks hogging your CPU. Addresses #15424.

### Additional details

See #15424 for justification. 

### Steps to test

You could implement an html page with a single element with a number that gets updated with the current time in a `requestAnimationFrame` callback and then copy the test from the example in the documentation and assert that it jumps to an hour.

They now have the ability to use `clock.setSystemTime`

### PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4680/files


emilyrohrbough:
@emilgoldsmith Thank you for the contribution! Let's get this change tested & documented. There is not need for it to be a "hidden" feature 😄

emilgoldsmith:
Sounds good to me @emilyrohrbough. 

As I mentioned in the PR description I think that it also makes a lot of sense to expose the `now` meta data as well as I think it's a common use case that you would like to fast forward the time by let's say 1 hour, which is currently exposed as a "hidden feature" in the `details` object that it looks like you use for testing.

I guess my personal suggestion would be to also expose that as a similar separate key on the clock object, and document and add types to that, and then just keep the details for your internal testing needs? I don't know if you'd like the tests that access `now` to be migrated to using that newly exposed property or if it's a fine clean way to just keep `details` as a special testing property on the object and not worry about `now` being able to be accessed in two different ways?

When we're fully agreed on the spec I'll try and get the change implemented within a few days to a week :)

emilyrohrbough:
@emilgoldsmith Lets keep this change scoped to `setSystemTime()`

cli/types/cypress.d.ts
```diff
@@ -5783,6 +5783,26 @@ declare namespace Cypress {
      *   cy.clock().invoke('restore')
      */
     restore(): void
+    /**
+     * Change the time without invoking any timers.
+     *
+     * Default value with no argument or undefined is 0.
+     *
+     * This can for example be useful if you need to change the time by an hour
```
emilyrohrbough, line 5791:
```suggestion
     * This can be useful if you need to change the time by an hour
```

emilgoldsmith:
Thank you @emilyrohrbough 

I’ll write documentation tomorrow, and also I couldn’t get the tests to run locally, it looks like the scope command excludes the driver package by default so ‘yarn test-unit —scope @packages/driver’ didn’t work, I assume I’m missing something from not being familiar with the code base

emilgoldsmith:
I also just remembered one more test I’d thought about writing, asserting that setSystemTime doesn’t invoke any timers, I’ll see if I can add that tomorrow too

BlueWinds:
> Thank you @emilyrohrbough
> 
> I’ll write documentation tomorrow, and also I couldn’t get the tests to run locally, it looks like the scope command excludes the driver package by default so ‘yarn test-unit —scope @packages/driver’ didn’t work, I assume I’m missing something from not being familiar with the code base

The driver is a bit annoying because changes you make to code aren't immediately reflected - the driver is pre-compiled. You can watch for your local changes via either

```
// Yes, the runner is the package that recompiles the driver, confusing. ^^;;
cd packages/runner
yarn watch

// In separate tab open the driver tests
cd packages/driver
yarn cypress:open
```

or

```
yarn dev --project packages/driver
```

which will watch the entire cypress repo for changes (takes longer to start up than `yarn watch` in the runner) and open cypress so you can run your spec.

emilgoldsmith:
Thanks @BlueWinds  that worked :).

I think we're nearly there, as I have fixed tests, and added 2 more, made the documentation change, and updated the PR template in the description so I now should have most stuff covered there.  I just have a few final questions I need help with though:

1. Which branch is it I should target with the documentation PR? Right now I have closed the issue as I accidentally did it against master which I know is wrong, but I can't see a `10.7` which I guess would be the next one right? The closed PR which is also the one I have referenced in the PR template as of now is https://github.com/cypress-io/cypress-documentation/pull/4680/files
2. I see that `clock.tick` has a nice error message, not just the underlying one from the `sinonjs` library, should we also implement this for `setSystemTime`? The reason I'm in doubt is because it seems more obvious to make one for something that is also a `cy.` command, so maybe it's not as important for us?
3. On that note, I can see that `restore` also logs a small `restore` message. Should we also implement logging for this command? That would be consistent with restore, and then I guess we should also test it
4. Finally as I was looking into those things I noticed that both `clock.restore` and `clock.tick` support an `options` argument, but it's not shown in the `cypress.d.ts` types. I guess that should only be done in a subsequent PR but is a quick fix so just thought I'd ask if I should quickly fix that? It's not mentioned in the documentation either though so maybe those two things should be done together in separate PRs?

emilgoldsmith:
Hi! I’m ready to wrap this pr up if we can just get the questions above cleared up :)

BlueWinds:
Ah, sorry, forgot to respond here, thank you for your patience with us!

>     1. Which branch is it I should target with the documentation PR? Right now I have closed the issue as I accidentally did it against master which I know is wrong, but I can't see a `10.7` which I guess would be the next one right? The closed PR which is also the one I have referenced in the PR template as of now is https://github.com/cypress-io/cypress-documentation/pull/4680/files

Go ahead and target master, and we'll merge that in when we do the release.

You'll probably also want to add a "History" block to the docs PR, a'la https://github.com/emilgoldsmith/cypress-documentation/blob/ead4b539f33ef9761fb733262807ee97239e9e2a/content/api/commands/readfile.md#history, showing the version 10.7 that the new feature was added in.

>     2. I see that `clock.tick` has a nice error message, not just the underlying one from the `sinonjs` library, should we also implement this for `setSystemTime`? The reason I'm in doubt is because it seems more obvious to make one for something that is also a `cy.` command, so maybe it's not as important for us?

>     3. On that note, I can see that `restore` also logs a small `restore` message. Should we also implement logging for this command? That would be consistent with restore, and then I guess we should also test it

I'm inclined to leave this as simple as can be. If you wanted to add more complex logging / error messages I don't think anyone would object, but I don't think it's necessary for the first itteration / just getting it working.

>     4. Finally as I was looking into those things I noticed that both `clock.restore` and `clock.tick` support an `options` argument, but it's not shown in the `cypress.d.ts` types. I guess that should only be done in a subsequent PR but is a quick fix so just thought I'd ask if I should quickly fix that? It's not mentioned in the documentation either though so maybe those two things should be done together in separate PRs?

Yeah, I'd be inclined to push that off to a separate PR. Our types are... not terribly accurate, really, and improvements are always welcome. Again you could do it here and no one would object, but not necessary to push this one through.

emilgoldsmith:
Great, thank you @BlueWinds! Well in that case I think this is ready for review and potentially release if everything checks out under review.

I just added the history block to the documentation that you recommended and reopened the pull request that goes to master. The only thing I haven't done on the PR checklist is tagging it in Zenhub which I'm assuming that you all can handle for me?

BlueWinds:
> I just added the history block to the documentation that you recommended and reopened the pull request that goes to master. The only thing I haven't done on the PR checklist is tagging it in Zenhub which I'm assuming that you all can handle for me?

Yep, zenhub is our internal tracker; in this case, adding it to a release just makes sure we don't forget it when writing the changelog.

@emilyrohrbough, @mike-plummer , looks like this is ready for another look.

BlueWinds:
> I just added the history block to the documentation that you recommended and reopened the pull request that goes to master. The only thing I haven't done on the PR checklist is tagging it in Zenhub which I'm assuming that you all can handle for me?

Yep, zenhub is our internal tracker; in this case, adding it to a release just makes sure we don't forget it when writing the changelog.

@emilyrohrbough, @mike-plummer , looks like this is ready for another look.

emilyrohrbough (approved):

mike-plummer (approved):

cli/types/tests/cypress-tests.ts
```diff
@@ -682,6 +682,24 @@ namespace CypressClockTests {
...
+  // setting system time with date object
+  cy.clock().then(clock => {
+    clock.setSystemTime(new Date(2019, 3, 2));
+  })
+  // setting system time with no argument and shortcut
+  cy.clock().invoke('setSystemTime')
+  // setting system time with timestamp and shortcut
+  cy.clock().invoke('setSystemTime', 1000)
+  // setting system time with date object and shortcut
+  cy.clock().invoke('setSystemTime', new Date(2019, 3, 2));
```
emilyrohrbough, line 702:
```suggestion
  cy.clock().invoke('setSystemTime', new Date(2019, 3, 2))
```

cli/types/tests/cypress-tests.ts
```diff
@@ -682,6 +682,24 @@ namespace CypressClockTests {
...
+  cy.clock().then(clock => {
+    clock.setSystemTime()
+  })
+  // setting system time with timestamp
+  cy.clock().then(clock => {
+    clock.setSystemTime(1000)
+  })
+  // setting system time with date object
+  cy.clock().then(clock => {
+    clock.setSystemTime(new Date(2019, 3, 2));
```
emilyrohrbough, line 695:
```suggestion
    clock.setSystemTime(new Date(2019, 3, 2))
```

cli/types/cypress.d.ts
```diff
@@ -5784,6 +5784,26 @@ declare namespace Cypress {
...
+     *   cy.clock()
+     *   cy.visit('/')
+     *   ...
+     *   cy.clock().then(clock => {
+     *     clock.setSystemTime(60 * 60 * 1000)
+     *   })
+     *   // or use this shortcut
+     *   cy.clock().invoke('setSystemTime', 60 * 60 * 1000)
+     */
+    setSystemTime(now?: number | Date | undefined): void
```
emilyrohrbough, line 5806:
```suggestion
    setSystemTime(now?: number | Date): void
```

PR #23330, opened 2022-08-13
fix: call screencastOpts as a function for starting a screencast on electron



htekdev:
- Closes 

### PR Tasks

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cypress",
-  "version": "9.7.0",
+  "version": "9.7.1",
   "description": "Cypress.io end to end testing tool",
   "private": true,
   "scripts": {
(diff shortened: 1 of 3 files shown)


mjhenkes (changes requested):
We haven't traditionally done patch fixes on prior major versions of Cypress. We'll need to discuss internally if/how we want to approach this. Marking this PR as Request changes to ensure it isn't merged until we're ready.

emilyrohrbough:
@htekdev At this time, we are not interested in creating a patch release for 9.x. Historically we have not done patch releases for previous major versions. This isn't to say we wouldn't consider it in the future, but at this time we will make that call on a case-by-case basis.

Is there a reason you are unable to update to Cypress 10.0 where this issue might be fixed and/or this change could be applied there?

nagash77:
@htekdev We haven't heard from you in awhile so I am going to close this PR.  We would love to hear more about why you are unable to upgrade your Cypress to the latest version, so please do respond to this ticket if you can.  If there are blockers we would like to understand what they are.

PR #23332, opened 2022-08-14
fix: Address windows test failures introduced by gql query changes

mike-plummer:
Addressing Windows e2e failure introduced by #23295 

### Additional details
New GQL fields were added to existing queries in some base components which introduced delays/failures in e2e tests. Happened to get green CI for the original PR but [CI builds on develop surfaced issues on the Windows runs](https://app.circleci.com/pipelines/github/cypress-io/cypress/41828/workflows/b91a66aa-6ba1-4147-9ef9-435b0c991a4c/jobs/1734268).

GQL fields for cloud connect modals (some of which are remote query fields) were added to the app default layout which caused failures in `authChange-subscription.cy.ts`. Updating to run the query on-demand instead so it doesn't impact majority of e2e tests.

Also reverting how two components are resolving `projectId` to how Mark originally had them. There is some delay/instability introduced by querying `projectId` directly that was causing some e2e failures and I wasn't able to track down precisely what was causing it. Given time we should be able to track down the proper way to stub/mock this out in the tests so we don't have to do this manually, but I didn't want to hold up the release since it wasn't obvious to me how to resolve

### Steps to test
Use test procedure from #23295

### PR Tasks

### DIFF
--- a/packages/app/src/layouts/default.vue
+++ b/packages/app/src/layouts/default.vue
@@ -16,7 +16,7 @@
       :page-name="currentRoute.name?.toString()"
       data-cy="app-header-bar"
       :allow-automatic-prompt-open="true"
-      @connect-project="showConnectDialog = true"
+      @connect-project="handleConnectProject"
     />
     <div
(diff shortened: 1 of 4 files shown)


lmiller1990 (approved):

marktnoonan (approved):
Tested this out locally, everything still looks good to me

Issue #23216, opened 2022-08-09
Next.js CT apps have "Component Testing Mismatched Dependencies" warning for deps it doesn't need.


ZachJW34:
### Current behavior

When booting up a Next.js app for CT testing, a warning for `Component Testing Mismatched Dependencies` is displayed on the "Open Browser" page. This warning says that the application is missing "webpack", but Next.js applications don't require webpack since Next bundles it internally.

![Screen Shot 2022-08-09 at 10 28 34 AM](https://user-images.githubusercontent.com/25158820/183699351-91c8b5d6-f726-45b6-8b34-f85ad36c7b0c.png)

### Desired behavior

The warning should not show when launching Next.js Component Testing

### Test code to reproduce

1. `npx create-next-app my-next-app && cd my-next-app`
2. `npm i -D cypress`
3. `npx cypress open`
4. Walk through CT setup, "Choose a browser" page will show misleading warning.

### Cypress Version

10.4.0

baus:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/23216?planning-poker) with ZenHub @astone123 @lmiller1990 @marktnoonan @mike-plummer @rockindahizzy @warrensplayer @ZachJW34

marktnoonan:
Looks like a pretty small change. But I'm not totally sure at a glance where `getBundlerDependency` is being called when the `nextjs` `dependencies` function doesn't actually call it, and already ignores the `bundler` argument. Which would have been my naive assumption about the change needed here.

mike-plummer:
My guess is this is the line giving the error: `ProjectConfigManager.ts` -`const result = await isDependencyInstalled(bundler, this.options.projectRoot)` - it treats the bundler as a required dependency even if it isn't listed in the dependencies. Would need to rework the wizard deps logic to remove this condition and then ensure other frameworks have their bundler listed as a dependency

JasonFairchild:
Just wanted to add that I have experienced this. I thought it might be a bug, so this is reassuring to find. But to be clear, if I'm using nextjs with 'latest' version, then I shouldn't expect to see any mismatched dependency warning, right? I can also confirm that I've run into no problems so far, but I've only written a few component tests, so not super in depth yet from that perspective.

[...]

PR #23339, opened 2022-08-15
fix: do not duplicate warnings when saving `cypress.config`

### LINKED ISSUE #23341: Warnings duplicated if you save `cypress.config` on the browser launch screen
### Current behavior

Reported: https://github.com/cypress-io/cypress/issues/23216#issuecomment-1213493767

If you go to the browser launch screen with a warning, and save your `cypress.config`, the warning is duplicated.

### Desired behavior

Just one of the warning, the latest one.

### Test code to reproduce

Repro:

1. Open packages/driver in open mode, go to browser launch screen
2. Put an invalid `baseUrl` in config and save
3. Put another invalid `baseUrl` in config and save
4. Whamo blamo you got two of the same warning

### Cypress Version

10.4

### Node version

16

### Operating System

All of them

### Debug Logs

_No response_

### Other

I fixed it already just need some reviewers https://github.com/cypress-io/cypress/pull/23339

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/23341

### User facing changelog

Fix bug where warnings are  duplicated if you save `cypress.config` while a warning is showing.

Eg: https://github.com/cypress-io/cypress/issues/23216#issuecomment-1213493767
Eg: https://github.com/cypress-io/cypress/pull/23034#issuecomment-1209643883

### Additional details

We should store warnings by the type which is unique, not some arbitrary id we generate. We don't want to two show two of the same warnings; we should only show one, the latest one.

Before (observe test on right; it changes the `baseUrl`, and we duplicate a warning):

image

After (observe test passes - asserts length of warning is 1):

image

### Steps to test

1. Get this repo
2. `yarn`
3. open some project like `driver`
4. go e2e and go to browser screen
5. change `baseUrl` in `cypress.config.js` to something like `localhost:40404` that does not exist
6. change it to some other ports - it should update the warning with the correct port, and not duplicate the warning
7. change back to `localhost:3500`
8. warning is gone now

See above.

### PR Tasks


packages/data-context/src/DataContext.ts
```diff
@@ -398,7 +398,14 @@ export class DataContext {
       }
 
       this.update((d) => {
-        d.diagnostics.warnings.push(warning)
+        const warningsIndex = d.diagnostics.warnings.findIndex((warning) => warning.cypressError.type === err.type)
+
+        if (warningsIndex !== -1) {
+          // older warning is superceeded by newer one (eg - warning same, but the details of the stack trace could be different)
+          d.diagnostics.warnings.splice(warningsIndex, 1)
```
emilyrohrbough, line 405: Is this logic need to be merged with the [issuedWarnings ](https://github.com/cypress-io/cypress/blob/develop/packages/config/src/browser.ts#L74)logic in `@packages/config`?

ZachJW34 (approved):
Works great! Though I am curious about what @emilyrohrbough pointed out, seems like we are storing warnings in two places (issuedWarnings stores key, d.diagnostics store entire error message)

lmiller1990:
Woah, I didn't know about `issuedWarnings`!

Should we just delete all of that? I haven't looked enough - I assume they are the same kind of warnings in both cases (the yellow overlay one)?

This is probably the root source of the duplication 🤔

lmiller1990:
@emilyrohrbough @ZachJW34 thoughts on https://github.com/cypress-io/cypress/pull/23339/commits/54cae3e799a57df836f8cbed14fe93efd44cb0f5

I merged the logic but there's some conflicting logic - we have a test that says once you've dismissed a warning issued from `packages/config`, it should not re-warn. Eg: 

Before: we warn about Cypress Studio deprecation. If the user dismisses this, then saves their `cypress.config.js`, we do NOT re-warn.
After: we re-warn every time we reload the `cypress.config.js`. This seems sensible to me - most of time, we want to re-warn (eg - the `baseUrl` server isn't available, etc. This means until they delete the `experimentalStudio` from `cypress.config.js`, they'll see a warning after dismissing it, and reloading their config.

This is a slight change in behavior (that was likely never well defined in the first place; thus the two sources of truth for warnings).

Thoughts? This reduces the surface area for warnings to a single location (good) and looses the coupling between `packages/config` and `packages/data-context`. Seems like it's probably sensible to warn every time they reload their config - or do we want special logic for `EXPERIMENTAL_` warnings? We could do this, but I am not sure how valuable it is.

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
```diff
@@ -11,6 +11,9 @@ import { Query } from './gql-Query'
 import { ScaffoldedFile } from './gql-ScaffoldedFile'
 import { WIZARD_BUNDLERS, WIZARD_FRAMEWORKS } from '@packages/scaffold-config'
 import { ErrorTypeEnum } from '../enumTypes'
+import debugLib from 'debug'
+
+const debug = debugLib('cypress:graphql:mutation')
```
emilyrohrbough, line 16: is this being used?

packages/launchpad/cypress/e2e/config-warning.cy.ts
```diff
@@ -98,7 +112,7 @@ describe('experimentalStudio', () => {
     cy.get('[data-cy="warning-alert"]').contains('Warning: Experimental Studio Removed')
   })
 
-  it('should not continually show experimentalStudio warning in the same project', () => {
+  it('should show experimentalStudio warning in the same project', () => {
```
emilyrohrbough, line 115: This test was added specifically to ensure warnings are clearable and they don't persist when switching testing types. See https://github.com/cypress-io/cypress/pull/21630.

I think we'll want to figure out what we need to hook into to maintain this functionality.

emilyrohrbough (changes requested):

packages/launchpad/cypress/e2e/config-warning.cy.ts
lmiller1990: Right, I looked into this a bit more. There's actually two kinds of warnings.

The edge case here is *global config* warnings. Eg consider

```ts
export default defineConfig({
  experimentalStudio: true,
  e2e: {
    baseUrl: 'http://this-does-not-exist:2222'
  }
})
```

Since we only execute `e2e` and `component` after selecting a testing type, and we clear those warnings when we change testing types, it's fine - the behavior is as expected.

The difference with `experimentalStudio` (which I'm calling a "global" config) is those warnings show up on the "selecting testing type" screen. So we need some kind of way to handle those. The majority (well, all?) of the deprecated global config result in an error (since we've got the migration UI, it's rare users will hit this):

image

I think `experiementalStudio` is the only *warning* (not error) that is global. 

Either way good catch - I'll look into some logic to cover this case.

packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -86,10 +86,10 @@ export class ProjectActions {
...
       d.activeBrowser = null
       d.currentProject = null
-      d.diagnostics = {
-        error: null,
-        warnings: [],
-      }
+
+      d.diagnostics.error = null
+      d.diagnostics.warnings = []
+      d.diagnostics.globallyIssuedWarnings = new Set()
```
lmiller1990, line 92: We bust this warning cache when we navigate out of the project (eg go to global mode) same as prior to this PR. I just moved the source of truth from `config` to `data-context`.

packages/launchpad/cypress/e2e/config-warning.cy.ts
lmiller1990: Done... https://github.com/cypress-io/cypress/pull/23339/commits/081cba8154c7785a3570abd09b3cf9d4198d4617 🙏

lmiller1990: I think the root of double source of truth is these function signatures...

https://github.com/cypress-io/cypress/blob/develop/packages/config/src/browser.ts#L78


```js
const validateNoBreakingOptions = (
  breakingCfgOptions: Readonly<BreakingOption[]>,
  cfg: any, 
  onWarning: ErrorHandler, 
  onErr: ErrorHandler, 
  testingType?: TestingType) => {
  // ...
}
```

We should not be passing `onWarning` and `onErr` from data-context to config. Instead, `validateNoBreakingOptions` should be `get BreakingOptions`, which would be dealt with in data-context:

```js
const { errors, warnings } = get BreakingOptions(config, testingType)

if (warnings) {
  onWarning(warnings)
)

if (errors) {
  onError(errors)
}
```

To better represent the separation of responsibilities. Anyway, this is a big lift, but good to know how this initial spreading of truth was introduced and a nice drive-by refactor if we are in this part of the code base again in the future. I really think config should *not* have side effects.

packages/data-context/src/DataContext.ts
lmiller1990: Done, see below comment/discussion thread.

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
lmiller1990: Yes it is used here https://github.com/cypress-io/cypress/blob/develop/packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts#L668

lmiller1990:
Moving to draft while I figure out the failing tests.

emilyrohrbough:
@lmiller1990 Do you expect to work on this any time soon or should we close till you have time to focus on it?

lmiller1990:
Let's close it off, this seems like a pretty low priority right now.

PR #23347, opened 2022-08-15
fix: persist registered spec sessions

### LINKED ISSUE #22381: No session is defined with the name
### Current behavior

Keep running into behavior where a recently declared session is no longer available.

It works on the first test, but fails on the second. Appears to be an issue with how we utilize `clearAllSavedSessions` to reset all sessions.

### Desired behavior

Declared sessions persist.

### Test code to reproduce

```ts
beforeEach(() => {
  cy.wrap(Cypress.session.clearAllSavedSessions());
});

it('a', () => {
  cy.session('INDY', () => {});
  cy.session('CLIENT', () => {});

  cy.session('INDY');
});

it('b', () => {
  cy.session('INDY', () => {});
  cy.session('CLIENT', () => {});

  cy.session('INDY');
});
```

### Cypress Version

10.1.0

### Other

Screen Shot 2022-06-17 at 10 28 45 AM

emilyrohrbough:
- Closes #22381
Ensure unique session ids are used in the spec and an previously used session id cannot be overridden with new arguments.

This is invalid:
```js
it('a', () => {
  cy.session('INDY', () => {
    'a';
  });
  cy.session('INDY');
});

it('b', () => {
  cy.session('INDY', () => { // will now throw as needs to be a unique session id
    'b'; 
  }); 
});
```

This now works
```js
it('a', () => {
  cy.session('INDY', () => {
    'a';
  });
  cy.session('INDY');
});

it('b', () => {
  cy.session('INDY', () => {
    'a';
  });
  // this was previously throwing needs setup func when it should have been registered
  cy.session('INDY');
});
```

### User facing changelog

- Fixed an issue where the unique session id logic was not persisting registered session ids which incorrectly allows session ids to override previously used session ids with different setups. Fixed [#22381](https://github.com/cypress-io/cypress/issues/22381).

This ensure unique session ids are used in the spec.

**Note:** if consumers are running `Cypress.session.clearAllSavedSessions()` between tests and are re-using an existing session id with a new setup function, they will now receive the following error:
```js
You may not call `cy.session()` with a previously used name and different options. If you want to specify different options, please use a unique name other than <session_id>.
```

I don't anticipate this o be hit by many users as re-using the session should restore the previous session. If a different setup function is used, it should already throw this error or they are restoring the session without a setup function, which won't hit this error.

### PR Tasks


AtofStryker (approved):

system-tests/projects/session-and-origin-e2e-specs/cypress/e2e/session/restores_saved_session.cy.js
```diff
@@ -21,3 +21,26 @@ it('t2', () => {
 
   cy.log('after')
 })
+
+// https://github.com/cypress-io/cypress/issues/22381
+describe('test sessionid integrity is maintained', () => {
```
chrisbreiding, line 26: Why system tests and not driver or cy-in-cy tests?

emilyrohrbough: These are tested as [cy-in-cy tests](https://github.com/cypress-io/cypress/pull/23347/files#diff-4b8a975eae3aec8fa6cff9bdfb404a1c0dac802061c61205f8791166e54c8ac9R186). All of the cy-in-cy tests that I've looked at are loading a project from system-test's to run tests against.

mjhenkes (approved):

system-tests/projects/session-and-origin-e2e-specs/cypress/e2e/session/restores_saved_session.cy.js
chrisbreiding: Ah my bad. I didn't realize that's how cy-in-cy tests worked. Interesting...

chrisbreiding (approved):

system-tests/projects/session-and-origin-e2e-specs/cypress/e2e/session/restores_saved_session.cy.js
emilyrohrbough: No worries at all. I personally don't love the setup...it seems we've started to abuse the system-test setup by copying these projects around and pulling these in to our integration tests. It seems like they should be fixtures instead 🤷🏻‍♀️

PR #23354, opened 2022-08-15
chore: Refactor assertion logging

BlueWinds:
### User facing changelog
No change in behavior. Refactoring only.

### Additional details
As part of the Detached DOM work, I've run into a lot of issues with the way assertion logging currently works. This PR simplifies our use of the "onBeforeLog" hook, moving the logic for "is this the same log message" into easier to find and understand places.

### Steps to test
The automated tests are pretty thorough in this regard. Feel free to play around with any existing project, or especially `assertions.cy.ts` to verify that nothing has changed.

### PR Tasks


packages/driver/cypress/e2e/commands/assertions.cy.js
```diff
@@ -790,19 +789,18 @@ describe('src/cy/commands/assertions', () => {
       cy.noop({}).should('have.property', 'foo')
     })
 
-    it('ends and snapshots immediately and sets child', (done) => {
+    it('snapshots immediately and sets child', (done) => {
```
BlueWinds, line 792: This PR does introduce one test change - the assertion log is in a `pending` state when initially added, and is `passed` asynchronously as part of the `finishAssertions()` function, which happens slightly later in the lifecycle of the command.

Other tests verify that it moves from the `pending` state to `resolved` before the next command executes - it's just slightly later in resolving the same promise chain than `log:added`. I don't believe this is a meaningful difference, and so this can remain a chore even though I had to update one test. :woman_shrugging:

packages/driver/src/cy/commands/asserting.ts
```diff
@@ -58,28 +96,38 @@ export default function (Commands, Cypress, cy, state) {
...
+          if (_.isFunction(cmd)) {
+            try {
+              return cmd.apply(memo, args)
+            } catch (err: any) {
+              // if we made it all the way to the actual
+              // assertion but its set to retry false then
+              // we need to log out this .should since there
+              // was a problem with the actual assertion syntax
+              if (err.retry === false) {
+                return throwAndLogErr(err)
```
lmiller1990, line 121: How can I get in here? I tried a few things, but not entirely clear how to reach this branch.

lmiller1990: Actually this one does the job: https://github.com/cypress-io/cypress/blob/develop/packages/driver/cypress/e2e/commands/assertions.cy.js#L379-L384

packages/driver/src/cypress/log.ts
```diff
@@ -348,10 +348,10 @@ export class Log {
       return this
     }
 
-    _.defaults(options, {
-      at: null,
-      next: null,
-    })
+    if (this.get('next')) {
```
lmiller1990, line 351: Any context on why we do this? I know this wasn't changed as part of this PR, but when I removed it and ran the tests, only navigation.cy.ts failed, specifically " it.only('logs before + after', () => {". The test doesn't exactly shine much light on the actual reason we do this, though.

lmiller1990 (commented):
All tests pass and I don't see any obvious problems, so looks good. Asked a few questions.

packages/driver/src/cypress/log.ts
lmiller1990: Right, it's the label for `before/after`. Without it, you get the id:

image

lmiller1990:
Actually, this might be incorrect? I ran a spec in app - `cypress/e2e/cypress-in-cypress-e2e.cy.ts`. I ran the first test using `.only`. I'm seeing a lot of extra logs.

image

lmiller1990 (changes requested):
Possible regression, going to request changes until we verify it's not a regression.

packages/driver/src/cypress/log.ts
BlueWinds: Yep, just so. This was fairly arcane to me as well, escaped notice until the most recent commits.

packages/driver/src/cy/commands/asserting.ts
BlueWinds: I'm going to be moving this file around and refactoring it a bunch more as part of Detached DOM, so I didn't spend a ton of time commenting every detail. I think the upcoming changes should make this area a bit more straightforward, separating out "command errors" from "execution errors".

BlueWinds:
> Actually, this might be incorrect? I am seeing a lot of red in the command log.

> I ran a spec in app - cypress/e2e/cypress-in-cypress-e2e.cy.ts. I ran the first test using .only. I don't see this in develop on the same test. cc @BlueWinds

Oof, yeah, I see the same. Investigating, will fix and add another driver test once I figure out what's going on here. Wondering if it has to do with cypress-in-cypress.

BlueWinds:
> Possible regression, going to request changes until we verify it's not a regression.

Ah, figured it out. `.should('not.exist')` was getting duplicated when it retried; there's no assertion on the number of logs for any of the `not.exist` tests.

I've fixed the issue - the 'deduplicate logs' logic just needed to be moved slightly higher in the function - and added an assertion to an existing test to prevent future regressions in the area. See https://github.com/cypress-io/cypress/pull/23354/commits/6849134c4a4b27ba72d06e0da2112eae322361dd for the exact changes, and good catch!

lmiller1990 (approved):
Regression was fixed and comments responded too. I am not driver expert yet, so defs need that second 👀 , but seems like things behave the same as they did prior to and after this refactor 👍

packages/driver/cypress/e2e/commands/assertions.cy.js
```diff
@@ -790,19 +793,18 @@ describe('src/cy/commands/assertions', () => {
...
       cy.on('log:added', (attrs, log) => {
-        if (attrs.name === 'assert') {
-          cy.removeAllListeners('log:added')
+        if (attrs.name !== 'assert') {
+          return
+        }
 
-          expect(log.get('ended')).to.be.true
-          expect(log.get('state')).to.eq('passed')
```
emilyrohrbough, line 799: Are theses no longer true?

packages/driver/src/cypress/log.ts
```diff
@@ -624,13 +598,53 @@ class LogManager {
...
+            return
+          }
+        }
+
+        command.log(log)
+      }
+
+      // if snapshot was passed
+      // in, go ahead and snapshot
+      if (log.get('snapshot') && !log.get('snapshots')) {
```
emilyrohrbough, line 638:
```suggestion
      if (log.get('snapshot')) {
```

packages/driver/src/cypress/log.ts
```diff
@@ -624,13 +598,53 @@ class LogManager {
...
       }
 
+      if (command) {
+        // See commands/asserting.ts for a detailed explanation of commandLogId.
+        // Short version: If onBeforeLog sets the same commandLogId as an existing log on the current command,
+        // then we want to merge the current instance into the existing one, rather than display a new log to
+        // the user.
+        let commandLogId = log.get('commandLogId')
+
+        if (commandLogId != null) {
```
emilyrohrbough, line 615: This logic is specific to `commands/asserting.ts` and commands are responsible for managing their own logs. Theres no need to move this logic here. This also puts additional attributes on the log that don't seem necessary which are stored on on the runner and the command.

emilyrohrbough (changes requested):

packages/driver/cypress/e2e/commands/assertions.cy.js
BlueWinds: Correct. I put an explanation of the change here: https://github.com/cypress-io/cypress/pull/23354#discussion_r946210351

packages/driver/src/cypress/log.ts
BlueWinds: Ah, yep. This was a remnant from earlier attempts to rein in the number of duplicate snapshots, but no longer necessary now that previousLogInstance logic is working correctly.

flotwig (approved):
Nice!

packages/driver/src/cypress/log.ts
```diff
@@ -648,13 +622,24 @@ class LogManager {
         }
       }
 
-      // set the log on the command
-      const current = state('current')
+      let command = state('current')
```
emilyrohrbough, line 625:
```suggestion
      const command = state('current')
```

packages/driver/src/cypress/log.ts
```diff
@@ -667,9 +652,8 @@ class LogManager {
 
       this.triggerLog(log)
 
-      // if not current state then the log is being run
-      // with no command reference, so just end the log
-      if (!current) {
+      // if the log isn't associated with a command, then we know it won't be retrying and we should just end it.
+      if (!state('current') || log.get('end')) {
```
emilyrohrbough, line 656:
```suggestion
      if (!command || log.get('end')) {
```

packages/driver/src/cypress/log.ts
```diff
@@ -355,15 +355,7 @@ export class Log {
...
     if (options.next && shouldRebindSnapshotFn) {
-      const originalLogSnapshotFn = this.snapshot
-
-      this.snapshot = function () {
-        // restore the original snapshot function
-        this.snapshot = originalLogSnapshotFn
-
-        // call orig fn with next as name
-        return originalLogSnapshotFn.call(this, options.next)
-      }
```
emilyrohrbough, line 366: We need to maintain this for cross-origin snapshots.

BlueWinds: I merged in develop, which included a bunch of cross-origin changes, and fixed several merge conflicts; the tests were initially failing, but after some playing around I got them passing again.

Do you have a specific instance in mind where it's failing to snapshot cross-origin in this branch, so I can add a test?

packages/driver/src/cy/commands/asserting.ts
```diff
@@ -7,6 +7,30 @@ import $errUtils from '../../cypress/error_utils'
 const reExistence = /exist/
 const reHaveLength = /length/
 
+const onBeforeLog = (log, command, commandLogId) => {
+  log.set('commandLogId', commandLogId)
+
+  if (command) {
```
emilyrohrbough, line 13: When would command be undefined here?

packages/driver/src/cypress/log.ts
emilyrohrbough: @AtofStryker ^^

packages/driver/src/cy/commands/asserting.ts
BlueWinds: It wouldn't be here. Remnant of when this was in log.ts, where there might not be a current command. Will remove the check.

packages/driver/src/cy/commands/asserting.ts
```diff
@@ -7,6 +7,30 @@ import $errUtils from '../../cypress/error_utils'
 const reExistence = /exist/
 const reHaveLength = /length/
 
+const onBeforeLog = (log, command, commandLogId) => {
+  log.set('commandLogId', commandLogId)
```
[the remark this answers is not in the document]
BlueWinds: I tried to use the log id initially, you can't set a log's ID:

```
    // dont ever allow existing id's to be mutated
    if (this.attributes.id) {
      delete this.obj.id
    }
```

because the ID is meant to be *completely unique* - to identify a log instance, across the entire test. Updating it in the log would also require updating the mapping of id -> Log instance that the `LogManager` maintains, which begins to look like the same level of complexity as having its own attribute.

Also important to note that in the current implementation, the commandLogId isn't globally unique - it's unique (and reproducible) only within the context of a single command. Different instances of `.should()` can generate the same commandLogId. This isn't a problem in the current implementation, because we only search `command.get('logs')` for a matching commandLogId.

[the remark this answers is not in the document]
BlueWinds: Basically, this just serves a different purpose than the log's id. It's a bit of state-tracking for an instance of `.should()`.

packages/driver/src/cypress/log.ts
AtofStryker: @emilyrohrbough @BlueWinds I don't believe there is an impact here to cross origin snapshots. 

In the case of asking the cross origin spec bridge for a snapshot, this should only happen in the case that something in the primary instance of cypress is requesting a snapshot while a `cy.origin` block is active or is about to be active, which means it's highly unlikely to be a command asking for the snapshot (ex: XHR's and fetch requests that ask for snapshots as a generated log that aren't direct command invocations). The tests in [snapshots.cy.ts](https://github.com/cypress-io/cypress/blob/develop/packages/driver/cypress/e2e/e2e/origin/snapshots.cy.ts) should cover this for known cases. 

In the case were we DO need to rebind the original snapshot function, this usually happens on interaction events which should almost always be commands. Commands that are run within `cy.origin` take snapshots within the spec bridge itself, so no asking spec bridges for snapshots should be occuring. This should be covered by the various [command](https://github.com/cypress-io/cypress/tree/develop/packages/driver/cypress/e2e/e2e/origin/commands) tests that assert on `consoleProps` and other log attributes to verify the snapshot is correct.

Since we don't have tests that verify snapshots interactively (can't leverage the system test and cy-in-cy doesn't support snapshots), I ran this manually to double check and everything looks good from what I can see 😄 .

BlueWinds: Thank you for double checking! I'd hate to have accidentally broken something you merged in just a few days ago.

AtofStryker: Of course! Honestly wish we had a better way to test snapshots moving forward. Getting snapshot testing into `cy-in-cy` I think would cover a lot of ground.

packages/driver/src/cypress/log.ts
BlueWinds: Applied.

packages/driver/src/cypress/log.ts
BlueWinds: Applied.

packages/driver/src/cy/commands/asserting.ts
```diff
@@ -7,6 +7,30 @@ import $errUtils from '../../cypress/error_utils'
...
+      // copy over 'snapshots' beforehand so that existing snapshots aren't lost
+      // in the merge operation.
+      log.set('snapshots', previousLogInstance.get('snapshots'))
+      previousLogInstance.merge(log)
+
+      if (previousLogInstance.get('end')) {
+        previousLogInstance.end()
+      }
+
+      return false
```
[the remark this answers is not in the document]
BlueWinds: Wrote a slightly different comment (with the same meaning).

emilyrohrbough (approved):

Issue #23357, opened 2022-08-16
@cypress/react18 does not properly rerender


hiebj:
### Current behavior

- The `rerender` function returned by `@cypress/react18` creates a new root on every render, forcibly recreating the component tree on rerender.
- The `rerender` functionality does not work (a mounted component's state can't be persisted from one render to the next).

### Desired behavior

- The `rerender` function returned by `@cypress/react18` should reuse the existing root if it has already been created but not yet cleaned up.
- The `rerender` functionality should work (a mounted component's state should be persisted from one render to the next).

### Test code to reproduce

The `rerender/input-accumulator` test from the adjacent `react` package fails. Also,

```ts
import React from 'react';

import { mount } from '@cypress/react18';

describe('rerender', () => {
  it('should persist state between rerender() calls', () => {
    mount(<StatefulComponent foo="baz" />).then(({ rerender }) => {
      cy.get('button').should('have.text', 'baz 0');
      cy.get('button').click().should('have.text', 'baz 1');
      rerender(<StatefulComponent foo="baz" />);
      cy.get('button').should('have.text', 'baz 1');
    });
  });
});

const StatefulComponent: React.FC<{ foo: string }> = ({ foo }) => {
  const [bar, setBar] = React.useState(0);
  return (
    <button
      onClick={() => {
        setBar(bar + 1);
      }}
    >
      {foo} {bar}
    </button>
  );
};

```

### Cypress Version

10.4.0

### Node version

v14.19.1

### Operating System

macOS 12.4

### Other

In the console, the following message is printed from React:

```
react-dom.development.js?61bb:86 Warning: You are calling ReactDOMClient.createRoot() on a container that has already been passed to createRoot() before. Instead, call root.render() on the existing root instead if you want to update it.
```

I believe I know exactly what the issue is and how to fix it, I am creating this issue to accompany a PR.

baus:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/23357?planning-poker) with ZenHub @amehta265 @astone123 @lmiller1990 @marktnoonan @mike-plummer @rockindahizzy @warrensplayer @ZachJW34

PR #23360, opened 2022-08-16
fix: cypress/react18 rerender

hiebj:
- Closes [issue-23357](https://github.com/cypress-io/cypress/issues/23357)

### User facing changelog
Fix rerender in @cypress/react18

### Additional details
The `rerender()` function returned by `mount()` creates a new `ReactDOM.Root` every time it is invoked.  This results in state being lost between rerenders, and all lifecycle hooks being rerun.

This PR fixes that by only creating a new root if one has not been created yet, or the previous one has been cleaned up.

### Steps to test
I ran the tests from the neighboring `react` project against `react18`. I have elected not to add tests to the `react18` package as it does not currently have them and that would make this PR much more substantial. I suggest finding some way to run the `react` tests on the `react18` project rather than copying or re-implementing the tests.

`rerender()` works now.

### PR Tasks

- [ na ] Have tests been added/updated? **SEE Steps to test ABOVE**
- [ na ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [ na ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?


lmiller1990 (commented):
Thanks for this, we definitely need the tests but someone who knows the code base well can probably pick it up.

For whoever does, here's what we should do:

We have `react18` in `system-tests`: https://github.com/cypress-io/cypress/blob/develop/system-tests/projects/react18/package.json#L11 which has `"projectFixtureDirectory": "react"` inside `package.json. This means when we run the system test, it inherits all the code from https://github.com/cypress-io/cypress/tree/develop/system-tests/project-fixtures/react (including the tests).

So, you could add your test in [here](https://github.com/cypress-io/cypress/tree/develop/system-tests/project-fixtures/react), then all react projects get it. You'll need to update a ton of snapshots, with `cd system-tests && SNAPSHOT_UPDATE=1 yarn test component_testing_spec`.

That should give us the coverage we need across all react versions.

cc @rockindahizzy, I think you worked on React 18 so this might be relevant to you. Alternatively if @rachelruderman would like to work on the system tests with either of us, that might be a good way to share how we test the `cy.mount` adapters for Component Testing (or in this case, how we missed coverage for re-rendering).

hiebj:
I can take a shot at implementing the tests the way you describe if one of your gurus doesn't get to it first.

lmiller1990:
Hi @hiebj I will try to get this tested and merged up by the end of the week - sorry for the delay.

lmiller1990:
Hi @hiebj I will try to get this tested and merged up by the end of the week - sorry for the delay.

hiebj:
@lmiller1990 No problem. I was considering taking a shot at it myself but I have no idea which snapshots to update. I tried running it with * and of course it took forever and some of the tests failed for reasons I did not understand. I was going to come back to it in the next day or two, but if you are on it I will defer.

lmiller1990:
> @lmiller1990 No problem. I was considering taking a shot at it myself but I have no idea which snapshots to update. I tried running it with * and of course it took forever and some of the tests failed for reasons I did not understand. I was going to come back to it in the next day or two, but if you are on it I will defer.

All good... I will write the process I follow to update everything as well, for future reference. Running the entire suite locally takes hours and hours, don't try that 😓 it's massively parallelized on CI so it takes ~20 minutes.

system-tests/project-fixtures/react/vite.config.js
```diff
@@ -1,5 +1,11 @@
 const {defineConfig} = require('vite')
 
 module.exports = defineConfig({
+  resolve: {
```
astone123, line 4: I had to add aliases for these packages, otherwise the React 17 tests that use Vite would fail when I called `React.useState` in my test.

system-tests/projects/react17/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "dependencies": {
     "react": "17.0.2",
-    "react-dom": " 17.0.2"
```
astone123, line 4: Drive-by fix. Not sure why there was a space here

ZachJW34 (approved):
Looks great! `unit-tests` is failing but not due to your changes https://github.com/cypress-io/cypress/issues/23623

LGTM!

astone123 (approved):

astone123:
This is ready to merge, just trying to get the CI green

lmiller1990 (approved):

PR #23296, opened 2022-08-11
fix: remove working directory from `--spec` glob patterns

tbiethman:
- Closes #22272

### User facing changelog

* Fixed an issue where usage of the `--spec` option would result in no specs found if the absolute project path included `glob` pattern characters.

### Additional details

We document the [`--spec`](https://docs.cypress.io/guides/guides/command-line#cypress-run-spec-lt-spec-gt) run mode option as supporting:
* Relative paths
* Absolute paths
* Glob patterns
* Comma-delimited lists of the above

Given this range of support, we normalize the provided value into an absolute path during option processing. This value eventually is given to the FileDateSource for matching against the filesystem using globby. However, the absolute path can contain glob pattern syntax (e.g. `C:\Program Files (x86)\my-project\...`) that is parsed by globby and can cause our lookups to fail. Today, users have little recourse but to move their project into a different directory, which isn't always possible.

My initial thought was to stop converting the option value into an absolute path altogether, but there were test cases featuring relative pathing that made me question the passivity of such a solution in a patch release. I instead updated the logic around our globbing to ensure that the working directory is never present in the patterns we supply to globby. We limit the search space to the current working directory using separate globby options, so it is not needed within the pattern itself. I also added a few unit tests around the updated logic and an additional system test with a project that includes these troublesome characters.

These issues are somewhat related to those I raised around our expected `supportFile` values [here](https://github.com/cypress-io/cypress/issues/22255), namely that it is hard for us to support absolute paths, relative paths, and glob patterns through a single API, especially across our supported architectures. Simplifying these options down to supporting only glob patterns, relative to the project root, would allow us to simplify a lot of internal transforms we're doing and leave us with these potential values:

```js
const supportedPatterns = [
  'cypress/e2e/*.cy.js',                 // standard patterns
  'cypress/e2e/spec.cy.js' ,             // patterns to match single files
  'cypress/e2e/spec \\(1\\).cy.js',      // patterns with user-escaped glob chars
  '**/spec-1.cy.js,**/spec-2.cy.js',     // we could still support multiple patterns if we want to...
  '**/spec-(1|2).cy.js',                 // ...but having defined glob expectations might negate the need
]

const unsupportedPatterns = [
  'cypress\e2e\*.cy.js',      // on windows, users would still use forward-slashes (backwards are reserved for escaping)
  'cypress/e2e/spec (1).js'.  // this will not match against an explicit file including '()', chars must be escaped.
]
```  
  
### Steps to test

[...]

PR #23368, opened 2022-08-16
fix: ensure absolute directories in glob patterns are leading prior to removal

### LINKED ISSUE #23361: 10.5.0 - "Your supportFile is missing or invalid" even when hardcoded in the config
### Current behavior

When we upgrade to Cypress 10.5.0 and run our E2E test suite through our pipelines, we get an error on each of docker containers:

> cypress_1  | Your project does not contain a default supportFile. We expect a file matching cypress/support/e2e.{js,jsx,ts,tsx} to exist.
> cypress_1  | 
> cypress_1  | If a support file is not necessary for your project, set supportFile to false.
> cypress_1  | 
> cypress_1  | https://on.cypress.io/support-file-missing-or-invalid

Even when we hardcode the location of the support file into the config, we get a similar error:

> cypress_2  | Your supportFile is missing or invalid: cypress/support/e2e.js
> cypress_2  | 
> cypress_2  | The supportFile must be a .js, .ts, .coffee file or be supported by your preprocessor plugin (if configured).
> cypress_2  | 
> cypress_2  | Fix your support file, or set supportFile to false if a support file is not necessary for your project.
> cypress_2  | 
> cypress_2  | If you have just renamed the extension of your supportFile, restart Cypress.
> cypress_2  | 
> cypress_2  | https://on.cypress.io/support-file-missing-or-invalid

When we downgrade to Cypress 10.4.0 - works again

### Desired behavior

Tests should just run as they did for 10.4.0

### Test code to reproduce

Not sure I can provide code - we didn't change anything. Initially, our configs did NOT have the supportFile location set

This is how our directory is structured - which is the default way, I believe:

Screenshot 2022-08-16 at 10 51 49

### Cypress Version

10.5.0

### Node version

16.16.0

### Operating System

Docker - running the Cypress included image for 10.5.0

### Debug Logs

_No response_

### Other

_No response_

tbiethman:
Closes #23361

### User facing changelog
 

* Fixes an issue where absolute path glob manipulation was incorrectly breaking supportFile/spec glob patterns that contained deep instances of the working directory.

### Additional details

The logic introduced with https://github.com/cypress-io/cypress/pull/23296 to broadly remove the project directory from our glob patterns was a little over-zealous. If the project directory was sufficiently simple and also found within the projects subdirectories (e.g. project located at `/e2e`, with a `--spec` value of `/e2e/cypress/e2e/*.cy.js`), the logic would behave incorrectly by replacing the nested instance of the directory (`cypress/e2e/*.cy.js` -> `cypress/.*.cy.js`). The same could be true for supportFile lookups, which go through the same logic.

This change ensures that we only remove the working directory from the glob pattern if it is in the leading position.

### PR Tasks


emilyrohrbough (approved):

AtofStryker (approved):

packages/data-context/src/sources/FileDataSource.ts
```diff
@@ -36,7 +36,11 @@ export class FileDataSource {
       // The working directory path may include characters that conflict with glob
       // syntax (brackets, parentheses, etc.) and cause our searches to inadvertently fail.
       // We scope our search to the working directory using the `cwd` globby option.
-      return globPattern.replace(cwd, '.')
+      if (globPattern.startsWith(cwd)) {
+        return globPattern.replace(cwd, '.')
+      }
```
mjhenkes, line 41: A root directory of `/cypress/e2e/`, would still break right? `/cypress/e2e/cypress/e2e/*.cy.js` -> `*.cy.js`

mjhenkes: Is there any way you can just replace the first instance of the cwd?

tbiethman: Using `replace` (vs. `replaceAll`), only the first instance will be removed: `/cypress/e2e/cypress/e2e/*.cy.js` -> `./cypress/e2e/*.cy.js`

mjhenkes: So how did we see the issue initially? does this get called multiple times?

tbiethman: So in the issue, there's a working directory of `/e2e`, and a glob pattern of `cypress/support/e2e.{js,ts}`. Without the startsWith check, we get `cypress/support..{js.ts}`, which is not helpful.

mjhenkes (approved):

PR #23376, opened 2022-08-16
chore: SEC-228 | Adding security scans

cypresschris:
- Closes SEC-228

### User facing changelog
n/a

### Additional details
This PR adds security controls in support of vulnerability scanning for third party packages as well as first party code.  Snyk is already implemented, but these actions bring the Cypress App into alignment with our vulnerability management strategy.  The actions that this pull request introduces send the scan results back to the Snyk dashboard.  The Snyk dashboard is then coupled with branch protection rules that dictate our levels of protection by blocking PRs where this git action fails as a check.  

Once the solution is fully implemented, the failures you see on the PR related to this would prevent merges of the code to the 'develop' and 'master' branches.  The exit status it failed with today indicates that there is a `critical` vulnerability in the code base.  This PR aims to increase this type of visibility.

### Steps to test
n/a

n/a
### PR Tasks

- [n/a ] Have tests been added/updated?
- [n/a ] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [n/a ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [n/a ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?


.github/workflows/static_analysis.yaml
```diff
@@ -0,0 +1,28 @@
+name: Snyk Static Analysis Scan
+# This git workflow leverages Snyk actions to perform a Static Application 
+# Testing scan (SAST) on our first-party code upon Pull Requests to Main & 
+# Develop branches.  We use this as a control to prevent vulnerabilities 
+# from being introduced into the codebase. 
+on:
+  pull_request:
+    branches: 
+      - main
```
emilyrohrbough, line 9:
```suggestion
      - master
```

emilyrohrbough:
@cypresschris Mind updating the summary to include more details on what to expect from this change and what we should be verifying? How how we'd handle a failure when on occurs? I see this build is failing, but not sure what the action would be if this was my PR.

We already receive weekly Synk reports via email & we have https://github.com/cypress-io/cypress/security/dependabot. Is this tool replacing those?

cypresschris:
> 

Hi Emily, I have made the requested changes to the branch name for `master` as well as updated the summary.  There is an internal call to discuss all of this tomorrow, I will invite you :)

emilyrohrbough:
@cypresschris Thank you!! Are we able to log the list of dependencies with errors instead of the exit code?
Screen Shot 2022-08-16 at 1 39 01 PM

mjhenkes:
@cypresschris is the cypress repo already in compliance or will adding this change prevent any further updates to cypress until we correct any security problems

How Synk collect its list of vulnerabilities? Is it similar to npm audit? Is there a way to flag false positive vulnerabilities?

cypresschris:
> @cypresschris is the cypress repo already in compliance or will adding this change prevent any further updates to cypress until we correct any security problems
> 
> How Synk collect its list of vulnerabilities? Is it similar to npm audit? Is there a way to flag false positive vulnerabilities?

Hello @mjhenkes 
- The repo is compliant, and this change will not prevent anything.  This change introduces automatic vulnerability scanning whenever there is a PR opened against Master or Develop.
- Snyk analyses the yarn.lock to make it's assessment of vulnerabilities.  The false positives can certainly be flagged and this is a part of the triage process.

nagash77 (approved):

.github/workflows/sca_scan.yaml
```diff
@@ -0,0 +1,32 @@
...
+  pull_request:
+    branches: 
+      - master
+      - develop
+jobs:
+  Snyk_SCA_Scan:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [14.x]
```
AtofStryker, line 16: Any reason we need to use node 14 here? Might be good to use node 16 to be in parity with the rest of the repo, or even better `16.14.2`.
```suggestion
        node-version: [16.x]
```

.github/workflows/sca_scan.yaml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - uses: actions/checkout@master
+      - name: Setting up Node
+      #- name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Installing snyk-delta and dependencies
+        run: npm i -g snyk-delta
+      - uses: snyk/actions/setup@master
```
mschile, line 26: Do we want to reference master or a particular version?

.github/workflows/sca_scan.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Perform SCA Scan
+        continue-on-error: false
+        run: |
+          snyk test --yarn-workspaces --strict-out-of-sync=false --detection-depth=6 --exclude=docker,Dockerfile --severity-threshold=critical
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
```
mschile, line 32: Secrets (with the exception of `GITHUB_TOKEN`) aren't passed when the PR is created from a forked repository. Does this matter?

https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow
> Note: With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

.github/workflows/sca_scan.yaml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  Snyk_SCA_Scan:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [14.x]
+    steps:
+      - uses: actions/checkout@master
+      - name: Setting up Node
+      #- name: Use Node.js ${{ matrix.node-version }}
```
AtofStryker, line 20:
```suggestion
```

AtofStryker (commented):
Looks like we are scanning anything with a `package.json` within the repository workspaces? IMO we shouldn't scan `system-test` project directories for vulnerabilities and fail PRs as those dependencies don't get shipped into the final project and are only used under test. We could scan them, but it seems weird we would fail a contributor PR based on this. 

Honestly, if we aren't scanning `lock` files, I feel like the scan is only the tip of the iceberg and could possibly project vulnerabilities that aren't existent. Is the goal for us to iterate with this over time to improve this before the solution is fully implemented and PRs fail?

Right now the job is optional, but once considered fully implemented, any PR with a vulnerability will fail. Do I understand that correctly? What is the process for fixing a vulnerability?

.github/workflows/sca_scan.yaml
cypresschris: Master

.github/workflows/sca_scan.yaml
cypresschris: Correct.  I like that security model, and I don't think there's a work-around.  They'll have to be working from the repo.

mschile: Just to be clear, no contributor PRs will have the security scan run then.

cypresschris:
> 

@AtofStryker we have the ability to omit certain workspaces, I'm just not aware of which ones.  If you get me a list, I'm happy to exclude them.

AtofStryker:
> @AtofStryker we have the ability to omit certain workspaces, I'm just not aware of which ones. If you get me a list, I'm happy to exclude them.

From the job, it looks like its just `system-tests`. Omitting that I think gets the expected result. 

 > Also this change does not cause a PR to fail. The branch protection which relies on this would ultimately do that. But this action is intended to raise visibility. 

In other words, the job isn't required yet, right? 

> The process for fixing was discussed during our call. I can fill you in offline, but there is a triage and timeline in place for the fixing of vulns.

I remember we had some discussions about it, but I am wondering if contributors are able to see a process online at some point to rectify issues that might exist in their PR. Then again, if forked PRs don't run the action, would this even be a problem? Do we want this to run on contributor PRs?

.github/workflows/sca_scan.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - master
+      - develop
+jobs:
+  Snyk_SCA_Scan:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [16.x]
+    steps:
+      - uses: actions/checkout@master
```
mschile, line 18: In the `static_analysis` action we use `@v2` instead of `@master`. I would prefer we reference versions wherever possible.

.github/workflows/static_analysis.yaml
```diff
@@ -0,0 +1,27 @@
+name: Snyk Static Analysis Scan
```
mschile, line 1: These actions both execute on the same triggers. Any reason to not combine them?

.github/workflows/sca_scan.yaml
```diff
@@ -0,0 +1,31 @@
+name: Snyk Software Composition Analysis Scan
```
mschile, line 1: Suggest renaming the file to something like `snyk_scan.yaml`.

.github/workflows/sca_scan.yaml
cypresschris: Thanks, I've updated

.github/workflows/static_analysis.yaml
cypresschris: I have them in separate jobs because they are two different scans.  They may grow in complexity as our needs evolve so I just thought it was simpler this way.

.github/workflows/sca_scan.yaml
cypresschris: Are you suggesting prepending `snyk_` to the name?  This is the naming convention we're using in the other repos as well since there are two different Snyk scans.

mschile: Yes, I think prefixing the filename with `snyk_` would help.

mschile (approved):

mjhenkes (approved):

.github/workflows/snyk_static_analysis_scan.yaml
```diff
@@ -0,0 +1,29 @@
+name: Snyk Static Analysis Scan
+# This git workflow leverages Snyk actions to perform a Static Application 
+# Testing scan (SAST) on our first-party code upon Pull Requests to Main &
```
AtofStryker, line 3:
```suggestion
# Testing scan (SAST) on our first-party code upon Pull Requests to Master & 
```

.github/workflows/snyk_sca_scan.yaml
```diff
@@ -0,0 +1,33 @@
+name: Snyk Software Composition Analysis Scan
+# This git workflow leverages Snyk actions to perform a Software Composition 
+# Analysis scan on our Opensource libraries upon Pull Requests to Main &
```
AtofStryker, line 3:
```suggestion
# Analysis scan on our Opensource libraries upon Pull Requests to Master & 
```

.github/workflows/snyk_sca_scan.yaml
```diff
@@ -0,0 +1,33 @@
+name: Snyk Software Composition Analysis Scan
+# This git workflow leverages Snyk actions to perform a Software Composition 
+# Analysis scan on our Opensource libraries upon Pull Requests to Main & 
+# Develop branches.  We use this as a control to prevent vulnerable packages
```
AtofStryker, line 4:
```suggestion
# Develop branches. We use this as a control to prevent vulnerable packages 
```

.github/workflows/snyk_static_analysis_scan.yaml
```diff
@@ -0,0 +1,29 @@
+name: Snyk Static Analysis Scan
+# This git workflow leverages Snyk actions to perform a Static Application 
+# Testing scan (SAST) on our first-party code upon Pull Requests to Main & 
+# Develop branches.  We use this as a control to prevent vulnerabilities
```
AtofStryker, line 4:
```suggestion
# Develop branches. We use this as a control to prevent vulnerabilities 
```

AtofStryker (commented):
I don't see the snyk check in the PR. Am I missing it?

chrisbreiding (approved):

PR #20375, opened 2022-02-26
fix: shortcuts conflict with browser native shortcuts

majidzeno:
- Closes [#20222](https://github.com/cypress-io/cypress/issues/20222) 

### User facing changelog
Cypress shortcuts updated to not conflict with the browser native shortcuts

### PR Tasks

- [ ] Have tests been added/updated?
- [ ] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [ ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [ ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [ ] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

PR #23384, opened 2022-08-16
fix: add keyboard shortcut for focusing specs search (9.x parity)

### LINKED ISSUE #21836: Implement Cmd + K for searching Specs
## **Summary**
In 9.x, the specs list search can be focused using `Cmd`/`Ctrl`

* `F`, and this is indicated to the user in the Search field.

To have parity with 10.0 we should have a shortcut key for focusing the search, but it shouldn’t be Cmd+F anymore, since that’s the shortcut for _browser_ search and we’ve already made clear elsewhere that we do not want to clobber this with our own shortcuts. It worked fine in the Electron app because this conflict didn’t exist there.

The most popular choices for this in other web apps seem to be `/` (same key as ``?``) and `Cmd`/`Ctrl` + `K`.

Since our docs use `Cmd`/`Ctrl`

* `K`, would suggest we match that as people are likely to hop back and forth.

## **Refinement TODOs**
Some considerations around implementing this would be:

* the need to show the keyboard shortcut in the Keyboard Shortcut modal, which currently only applies to hotkeys that are used in the Runner
* specifying the behavior of the button when used on the Spec Runner page - would it expand the specs list and focus the search? Should that have the knock on effect of stopping the running test? Should we just avoid having the shortcut work at all on the runner page? That seems weird if we have the specs list open and a test is not running, when it would be perfectly safe to hit a shortcut.

So a bit of work to do to confirm what shortcut to use, and where to show it, and what is the ideal UX on the spect runner page.

┆Issue is synchronized with this [Jira Task](https://cypress-io.atlassian.net/browse/UNIFY-1586) by [Unito](https://www.unito.io)
┆author: Mark Noonan
┆epic: 9.x Parity
┆friendlyId: UNIFY-1586
┆priority: Medium
┆sprint: Backlog
┆taskType: Task

ZachJW34:
- Closes #21836

### User facing changelog
Adds back in the keyboard shortcut for searching specs in the runner

### Additional details
We already have a system in place for shortcuts so rather than implementing it in Vue-land I decided to use the existing system. I initially tried it out in Vue (using `useMagicKeys`) but since the `isSpecsListOpen` property is disconnected from the app and the reporter, the state of the `Toggle Specs List` button in the reporter was falling out of sync. Initiating the flow from the reporter and using the event emitter to communicate back kept everything in sync.

The original issue has an [Update text label](https://github.com/cypress-io/cypress/issues/21836#issuecomment-1203145980) task for updating the placeholder of the search to include the shortcut (from `Search Specs` => `Search Specs (Cmd + K)`) but adding the extra text to the label is not ideal since there is little space to work with. Most of the time, the text would be truncated and the user would see `Search Specs...`.

@mapsandapps tagging you to see what we should add to surface this shortcut. @marktnoonan you stated in the original issue to skip adding to the "Keyboard Shortcuts Modal" but in my mind that seems like the best place to put it.

### Steps to test
Open up any app and navigate to the runner (click on a spec). Press `Cmd + k` or `Ctrl + k` to have the Search Specs input focused. It will auto-open the specs list if it is closed.

https://user-images.githubusercontent.com/25158820/184994119-a88fb9ec-204d-452c-a51e-f094c2d5021f.mov

### PR Tasks

### DIFF
--- a/packages/app/src/runner/event-manager.ts
+++ b/packages/app/src/runner/event-manager.ts
@@ -255,6 +255,10 @@ export class EventManager {
       this.saveState(state)
     })
 
+    this.reporterBus.on('focus:search', () => {
+      this.localBus.emit('focus:app:search')
+    })
+
     this.reporterBus.on('clear:session', () => {
       if (!Cypress) return
 
--- a/packages/app/src/specs/InlineSpecListHeader.vue
+++ b/packages/app/src/specs/InlineSpecListHeader.vue
@@ -10,7 +10,7 @@
         class="flex h-full inset-y-0 w-32px absolute items-center pointer-events-none"
       >
         <i-cy-magnifying-glass_x16
-          :class="inputFocused ? 'icon-dark-indigo-300' : 'icon-dark-gray-800'"
+          :class="runnerUiStore.isSpecsListInputFocused ? 'icon-dark-indigo-300' : 'icon-dark-gray-800'"
           class="icon-light-gray-1000"
         />
       </div>
@@ -26,22 +26,22 @@
           placeholder-gray-700
           text-gray-500
         "
-        :class="inputFocused || props.specFilterModel.length ? 'w-full' : 'w-16px'"
+        :class="runnerUiStore.isSpecsListInputFocused || props.specFilterModel.length ? 'w-full' : 'w-16px'"
         :value="props.specFilterModel"
         type="search"
         minlength="1"
         autocapitalize="off"
         autocomplete="off"
         spellcheck="false"
-        @focus="inputFocused = true"
-        @blur="inputFocused = false"
+        @focus="handleFocus"
+        @blur="handleBlur"
         @input="onInput"
       >
       <label
         for="inline-spec-list-header-search"
         class="cursor-text font-light bottom-4px left-24px text-gray-500 pointer-events-none absolute"
         :class="{
-          'sr-only': inputFocused || props.specFilterModel
+          'sr-only': runnerUiStore.isSpecsListInputFocused || props.specFilterModel
         }"
       >
         {{ t('specPage.searchPlaceholder') }}
@@ -56,7 +56,7 @@
       >
         <i-cy-delete_x16
           class="icon-light-gray-1000 group-hocus:icon-dark-indigo-300"
-          :class="inputFocused ? 'icon-dark-indigo-300' : 'icon-dark-gray-800'"
+          :class="runnerUiStore.isSpecsListInputFocused ? 'icon-dark-indigo-300' : 'icon-dark-gray-800'"
         />
       </button>
     </div>
@@ -90,8 +90,9 @@
 </template>
 
 <script lang="ts" setup>
-import { ref } from 'vue'
+import { ref, watchEffect } from 'vue'
 import { useI18n } from '@cy/i18n'
+import { useRunnerUiStore } from '../store'
 
 const { t } = useI18n()
 const props = defineProps<{
@@ -104,8 +105,19 @@ const emit = defineEmits<{
   (e: 'newSpec'): void
 }>()
 
-const inputFocused = ref(false)
+const runnerUiStore = useRunnerUiStore()
+
 const input = ref<HTMLInputElement>()
+
+watchEffect(() => {
+  if (runnerUiStore.isSpecsListInputFocused) {
+    // Wrapping in timeout since the inline spec tree auto-focuses
+    // the active spec after mount.
+    setTimeout(() => {
+      input.value?.focus()
+    }, 100)
+  }
+})
 
 const onInput = (e: Event) => {
   const value = (e.target as HTMLInputElement).value
@@ -115,6 +127,14 @@ const onInput = (e: Event) => {
 
 const clearInput = (e: Event) => {
   emit('update:specFilterModel', '')
+}
+
+const handleFocus = () => {
+  runnerUiStore.setPreference('isSpecsListInputFocused', true)
+}
+
+const handleBlur = () => {
+  runnerUiStore.setPreference('isSpecsListInputFocused', false)
 }
 </script>
 
--- a/packages/reporter/cypress/e2e/shortcuts.cy.ts
+++ b/packages/reporter/cypress/e2e/shortcuts.cy.ts
@@ -161,5 +161,25 @@ describe('shortcuts', function () {
         expect(runner.emit).not.to.have.been.calledWith('focus:tests')
       })
     })
+
+    context('focus search', () => {
+      ['{ctrl+k}', '{meta+k}'].forEach((shortcut) => {
+        it(`emits "focus:search" and toggles spec list with ${shortcut}`, () => {
+          cy.get('body').type(shortcut).then(() => {
+            expect(runner.emit).to.have.been.calledWith('save:state')
+            expect(runner.emit).to.have.been.calledWith('focus:search')
+            cy.contains('button', 'Specs').should('have.attr', 'aria-expanded', 'true')
+
+            // Reset call history to verify we don't toggle spec list if it's already open
+            runner.emit.resetHistory()
+          })
+
+          cy.get('body').type(shortcut).then(() => {
+            expect(runner.emit).to.not.have.been.calledWith('save:state')
+            expect(runner.emit).to.have.been.calledWith('focus:search')
+          })
+        })
+      })
+    })
   })
 })
--- a/packages/reporter/src/lib/shortcuts.ts
+++ b/packages/reporter/src/lib/shortcuts.ts
@@ -19,7 +19,26 @@ class Shortcuts {
     const isTextLike = $dom.isTextLike(event.target)
     const isAnyModifierKeyPressed = event.altKey || event.ctrlKey || event.shiftKey || event.metaKey
 
-    if (isAnyModifierKeyPressed || isTextLike) return
+    if (isTextLike) return
+
+    if (isAnyModifierKeyPressed) {
+      switch (event.key) {
+        case 'k': {
+          if (event.ctrlKey || event.metaKey) {
+            action('toggle:spec:list', () => {
+              if (!appState.isSpecsListOpen) {
+                appState.toggleSpecList()
+                events.emit('save:state')
+              }
+
+              events.emit('focus:search')
+            })()
+          }
+
+          break
+        } default: return
+      }
+    }
 
     switch (event.key) {
       case 'r': !appState.studioActive && events.emit('restart')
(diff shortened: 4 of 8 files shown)


marktnoonan:
Sweeet. A couple things:

> The original issue has an https://github.com/cypress-io/cypress/issues/21836#issuecomment-1203145980 task for updating the placeholder of the search to include the shortcut (from Search Specs => Search Specs (Cmd + K)) but adding the extra text to the label is not ideal since there is little space to work with. Most of the time, the text would be truncated and the user would see Search Specs....

> @mapsandapps tagging you to see what we should add to surface this shortcut. @marktnoonan you stated in the original issue to skip adding to the "Keyboard Shortcuts Modal" but in my mind that seems like the best place to put it.

I was really thinking of the label in the main Specs List. Makes sense that this doesn't fit in the Inline Specs list, I hadn't tried it there. You are probably right that we might as well put it in the keyboard modal (even though not all shortcuts are there, and it is not itself super discoverable).  

At a glance, since I am about to log off, it looks like this only implements the shortcut on the Inline Specs List, which is awesome, but we should have it on the main specs list also. That was where it was in 9.x (in the old Electron app) and is the thing we are trying to restore with this issue. I should have been more clear about that in the requirements when it was clear I wasn't going to be the one doing this. That's where the VueUse stuff would come in handy - the runner stuff using `shortcuts.ts` seems great.

Sadly hardly anybody knows about the Inline Specs List itself, so there's not a great way to surface it there. I was figuring having the shortcut in the UI of the main specs list would encourage folks to try the same shortcut on a different screen and be pleasantly surprised when it worked, and they see the inline specs list for (quite possibly) the first time.

Let's review this one with Brian before merge as well in case there are any reservations about new keyboard shortcuts in the runner. It's a variation on what does `f` does, so I think it's not really a new concept, but still couldn't hurt.

packages/app/src/specs/InlineSpecListHeader.vue
```diff
@@ -104,9 +105,20 @@ const emit = defineEmits<{
...
-const inputFocused = ref(false)
+const runnerUiStore = useRunnerUiStore()
+
 const input = ref<HTMLInputElement>()
 
+watchEffect(() => {
+  if (runnerUiStore.isSpecsListInputFocused) {
+    // Wrapping in timeout since the inline spec tree auto-focuses
+    // the active spec after mount.
+    setTimeout(() => {
```
lmiller1990, line 116: Guessing you might have tried this but should we find the logic for "focus `activeSpec`", and do this same check there? Eg - something more reliable than just a `setTimeout` and hope that whatever milliseconds is enough.

Looks like that happens here https://github.com/cypress-io/cypress/blob/develop/packages/frontend-shared/src/composables/useVirtualListNavigation.ts#L10

So maybe we could pass a `shouldFocusItem` arg here? https://github.com/cypress-io/cypress/blob/develop/packages/app/src/specs/InlineSpecListTree.vue#L24

Then instead of doing the whole "focus the active spec then focus the input" we could just skip the active spec part, and we would not need the `setTimeout` work around.

Just an idea, not sure on the logistics or complexity of this, lmk what you think - can re-review/approve if this approach has some flaw.

packages/app/src/specs/InlineSpecListHeader.vue
```diff
@@ -116,6 +128,14 @@ const onInput = (e: Event) => {
 const clearInput = (e: Event) => {
   emit('update:specFilterModel', '')
 }
+
+const handleFocus = () => {
+  runnerUiStore.setPreference('isSpecsListInputFocused', true)
```
lmiller1990, line 133: Hmm I think this is actually the wrong place for this - looking at the other preferences, they are all things that are saved in the appData - like reporterWidth, reporterHeight - things that persist between sessions. 

It looks like `setPreferences` is coded in a way to just update any key in the store but I think this is actually implemented for the app preferences specifically, it's only used here right now: https://github.com/cypress-io/cypress/blob/develop/packages/app/src/composables/usePreferences.ts#L21 which is mixed in with all the GraphQL etc.

I think we might want to create a separate setter for this, like `runnerUiStore.setShouldSpecsListInputFocused` to better separate the concerns. 

Also looking *more* closely it looks like `runnerUiStore` should likely be two separate stores - `runnerUiStore` (for things like this) and `runnerPreferences` (for appData and things that are persisted).

lmiller1990 (commented):
Cool, we definitely need this for 9.x parity. Looking though this made a few things stand out - nothing you changed, but some existing debt that we likely want to attack in the near future.

Not entirely sure if/when to make the UI store change I suggested (runner UI vs appDataPreferences, I'm sure you see the concern).

More curious about your thoughts on the focus logic - `setTimeout` vs another approach.

mapsandapps:
> @mapsandapps tagging you to see what we should add to surface this shortcut.

I would add it to the modal, yeah. It also needs to be added to [the docs](https://docs.cypress.io/guides/core-concepts/cypress-app#Keyboard-shortcuts)

packages/reporter/src/lib/shortcuts.ts
```diff
@@ -19,7 +19,26 @@ class Shortcuts {
     const isTextLike = $dom.isTextLike(event.target)
     const isAnyModifierKeyPressed = event.altKey || event.ctrlKey || event.shiftKey || event.metaKey
 
-    if (isAnyModifierKeyPressed || isTextLike) return
+    if (isTextLike) return
+
+    if (isAnyModifierKeyPressed) {
+      switch (event.key) {
```
emilyrohrbough, line 25: None of our current or previous keyboard short cuts use [`ctrl` to trigger the shortcut](https://github.com/cypress-io/cypress/commit/a80f2ba255ec51c240b4987559bbc60527f9f1c6#diff-0c07826c542debabd81f3ca774e5dfee0280f24a69aa09bb1fb4d206f5bfc4d4R24)....do we need to introduce it for this shortcut?

packages/reporter/cypress/e2e/shortcuts.cy.ts
```diff
@@ -161,5 +161,25 @@ describe('shortcuts', function () {
         expect(runner.emit).not.to.have.been.calledWith('focus:tests')
       })
     })
+
+    context('focus search', () => {
```
emilyrohrbough, line 165: Can you split this test up and/or add a log to  indicate which part is testing the spec list is open or not? It's not obvious looking through this without comparing your changes

emilyrohrbough (changes requested):

packages/reporter/src/lib/shortcuts.ts
marktnoonan: I think we should. The main intention of this ticket was to restore the shortcut into the main Specs List (which is not yet implemented in this PR) - In Cy 9.x this was `Ctrl + F`, which worked since it was in the Electron app, in 10 `Ctrl + F` would step on the browser search, so `Ctrl + K` is the next best option (and matches our docs site).

Single-key shortcuts have some accessibility drawbacks (voice control-related iirc) and in the long run it would be good to move away from them.

emilyrohrbough: >  The main intention of this ticket was to restore the shortcut into the main Specs List (which is not yet implemented in this PR)

There is a shortcut for[ toggling the spec list already `f`](https://github.com/cypress-io/cypress/pull/23384/files/056df50d3d9a83180bf432d09b8d0507f5836b69#diff-0c07826c542debabd81f3ca774e5dfee0280f24a69aa09bb1fb4d206f5bfc4d4R48) ...its just a bit different since 9.x moved you back to the "launchpad" essentially.

emilyrohrbough: I will say, in 9.x it was just `f` though, which out the `ctrl+f` because a user specially added in the logic to prevent our shortcuts on `ctrl+f`: https://github.com/cypress-io/cypress/pull/20375

marktnoonan: What I mean is, on the equivalent of this page in Cypress 9, you could press `Cmd + F` to focus the search bar. The goal is to get a keyboard shortcut for this into Cy 10. The implications for "what should `Cmd + F` do in the runner?" are more like an after thought - since we have a specs list there as well, the same keyboard shortcut should work both places.

Screen Shot 2022-08-17 at 11 32 22 AM

marktnoonan: > I will say, in 9.x it was just f though, which out the ctrl+f because a user specially added in the logic to prevent our shortcuts on ctrl+f: #20375

Right, and I agree with that, we shouldn't step on browser search. We're talking about 2 different things though. "Go to the specs list" in 9 vs "Focus the search bar" in 9.

"Go to specs list" became "open inline specs list" and "focus the search bar" got dropped as a feature.

packages/app/src/runner/event-manager.ts
```diff
@@ -255,6 +255,10 @@ export class EventManager {
       this.saveState(state)
     })
 
+    this.reporterBus.on('focus:search', () => {
```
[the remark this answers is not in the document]
ZachJW34: Didn't realize that the `focus input` was needed in the main specs list, so I'll look into moving to a vue-only solution, only thing that was stopping me was syncing the state of the specs-list toggle between the app and the reporter but we _might_ be able to communicate from the app to the reporter and ensure it stays synced.

As for the `NO_COMMAND_LOG` yeah it wouldn't work unless the keyboard listener was hoisted up to the Vue app

packages/app/src/specs/InlineSpecListHeader.vue
ZachJW34: That's a good idea, will see if it's an easy add

packages/app/src/specs/InlineSpecListHeader.vue
ZachJW34: It did feel off using the `setPreferences`, it would make more sense if the mutation for saving a preference was in the `setPreference` action rather than having to do:
```
runnerUiStore.setPreference(preference, value)
```

packages/reporter/cypress/e2e/shortcuts.cy.ts
ZachJW34: Can do!

ZachJW34:
@marktnoonan I probably should have realized it was for both 😅, will get it added. Will probably move a bit slow on this one but will have it ready for our next sync with Brian

packages/app/src/specs/InlineSpecListHeader.vue
lmiller1990: Sounds good, separate action works for me.

emilyrohrbough:
Closing as stale since it's been a while since this branch has been updated.

PR #23385, opened 2022-08-16
feat(cypress/schematic): add support for component testing

admah:
This adds Angular CT support to the Cypress Schematic.
 
### User facing changelog

- Users can setup CT via `ng-add`
- Users can scaffold component tests on install via `ng-add`
- Users can generate e2e and component tests via `ng-generate`
- Users can generate component tests alongside ALL project components via `ng-generate`

More details are in the README


npm/cypress-schematic/README.md
```diff
@@ -223,12 +242,22 @@ In order to prevent the application from building, add the following to the end
...
 
+### Specify Testing Type 
+
+The default generated spec is E2E.  In order to generate a component test you can run:
+
+```shell script
+ng generate @cypress/schematic:spec --name=button -t component
+```
+
+`-t` is an alias for `testing-type`. It accepts `e2e` or `component` as arguments. If you are using the CLI tool, a prompt will appear asking which spec type you want to generate.
```
jordanpowell88, line 253: not a huge deal but the cypress app uses `--component` & `--e2e` as there flags. What do you think about using that instead of `-t`?

npm/cypress-schematic/README.md
```diff
@@ -238,17 +267,33 @@ This will create a new spec file named `login.cy.ts` in the default Cypress fold
 Add a `--project=` flag to specify the project:
 
 ```shell script
-ng generate @cypress/schematic:e2e --name=login --project=sandbox
+ng generate @cypress/schematic:spec --name=login --project=sandbox
```
jordanpowell88, line 270: a future idea would be to look for the component name matching the `--name` and generating the component there instead of the default. In e2e the default is perfect but in CT we wouldn't really want to generate those there

npm/cypress-schematic/src/ct.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+const ANGULAR_PROJECTS: ProjectFixtureDir[] = ['angular-13', 'angular-14']
+
+describe('ng add @cypress/schematic / e2e and ct', function () {
+  this.timeout(1000 * 60 * 4)
+
+  for (const project of ANGULAR_PROJECTS) {
+    it('should install ct files with option and no component specs', async () => {
+      const projectPath = await scaffoldAngularProject(project)
+
+      await runCommandInProject(`yarn add @cypress/schematic@file:${cypressSchematicPackagePath}`, projectPath)
```
jordanpowell88, line 36: does this need an `ng` command? Not sure how its running as is

jordanpowell88 (approved):
LGTM. Just a few minor pieces of feedback

npm/cypress-schematic/src/builders/cypress/schema.json
```diff
@@ -87,6 +92,10 @@
       "type": "boolean",
       "description": "Recompile and run tests when files change.",
       "default": false
+    },
+    "testingType": {
+      "type": "string",
+      "description": "Specify the type of tests to execute; either e2e or component. Defaults to e2e."
```
ZachJW34, line 98: We can add `"enum": ["e2e, "component"]`

npm/cypress-schematic/src/schematics/utils/index.ts
```diff
@@ -46,3 +61,47 @@ export function getLatestNodeVersion (packageName: string): Promise<NodePackage>
...
+    return tree.create(`${appPath}/${componentFilename}.component.cy.ts`, ctSpecContent({ component: componentName ? componentName[0] : componentFilename, file: componentFilename }))
+  }
+}
+
+export function getDirectoriesAndCreateSpecs ({ appPath, tree }: { appPath: string, tree: Tree}) {
+  let components = []
+  let directories = []
+
+  const projectPath = resolve(getSystemPath(normalize('')))
+  const contents = readdirSync(resolve(`${projectPath}/${appPath}`), { withFileTypes: true })
```
ZachJW34, line 83: There is a `tree.readDir` method which could be useful as it will read the files/dirs that are in the virtual tree (if any are added)

npm/cypress-schematic/src/schematics/utils/index.ts
```diff
@@ -46,3 +61,47 @@ export function getLatestNodeVersion (packageName: string): Promise<NodePackage>
...
+
+    return tree.create(`${appPath}/${componentFilename}.component.cy.ts`, ctSpecContent({ component: componentName ? componentName[0] : componentFilename, file: componentFilename }))
+  }
+}
+
+export function getDirectoriesAndCreateSpecs ({ appPath, tree }: { appPath: string, tree: Tree}) {
+  let components = []
+  let directories = []
+
+  const projectPath = resolve(getSystemPath(normalize('')))
```
ZachJW34, line 82: I believe Angular normalizes all file-paths behind the scene so you should only have to use posix paths

npm/cypress-schematic/src/schematics/utils/index.ts
```diff
@@ -1,13 +1,28 @@
...
 
 import { getPackageJsonDependency } from './dependencies'
+import { JSONFile } from './jsonFile'
 
 export interface NodePackage {
   name: string
   version: string
 }
 
+const ctSpecContent = ({ component, file }: {component: string, file: string}): string => {
```
ZachJW34, line 15: Can we use the `files/ct` template to generate the spec?

npm/cypress-schematic/src/schematics/ng-generate/cypress-test/index.ts
```diff
@@ -53,14 +53,18 @@ export default function (options: Schema): Rule {
...
-    const templateSource = apply(url('../files/__path__'), [
+    console.log(`Creating new ${options.testingType} spec named: ${options.name}`)
+
+    const templatePath = options.testingType === 'component' ? '../files/ct/__path__' : '../files/e2e/__path__'
+    const templateSource = apply(url(templatePath), [
       applyTemplates({
         classify: strings.classify,
         dasherize: strings.dasherize,
-        name: options.name,
+        name: options.testingType === 'component' ? `${options.name.toUpperCase()}Component` : options.name,
```
ZachJW34, line 66: `toUpperCase` results in:

```ts
import { MYSPECComponent } from './MySpec.component'

describe('MYSPECComponent', () => {
  it('should mount', () => {
    cy.mount(MYSPECComponent)
  })
})
```

We are using classify in the template, so we should be able to remove the uppercase transform

npm/cypress-schematic/README.md
```diff
@@ -57,11 +59,28 @@ If you have chosen to add or update the `ng e2e` command, you can also run Cypre
...
+If you have chosen to add the `ng ct` command, you can also run Cypress component testing in `open` mode using this:
+
+```shell script
+ng run {project-name}:ct
+```
+
 To generate new e2e spec files:
 
 ```shell script
+<<<<<<< HEAD
```
ZachJW34, line 71: Bad merge?

npm/cypress-schematic/src/ct.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+      await runCommandInProject('yarn ng add @cypress/schematic --e2e --ct --add-ct-specs false', projectPath)
+      await runCommandInProject('yarn ng run angular:ct --watch false', projectPath)
```
ZachJW34, line 38:
```suggestion
      await runCommandInProject('yarn ng run angular:ct --watch false --spec src/app/app.component.cy.ts', projectPath)
```

so as to not run more tests than necessary. Bonus points if we remove the component and then invoke the component generator and run the test that is created

npm/cypress-schematic/src/e2e.spec.ts
```diff
@@ -26,16 +26,16 @@ const cypressSchematicPackagePath = path.join(__dirname, '..')
...
 
   for (const project of ANGULAR_PROJECTS) {
-    it('should', async () => {
+    it('should install e2e files by default', async () => {
       const projectPath = await scaffoldAngularProject(project)
 
       await runCommandInProject(`yarn add @cypress/schematic@file:${cypressSchematicPackagePath}`, projectPath)
-      await runCommandInProject('yarn ng add @cypress/schematic --e2eUpdate', projectPath)
-      await runCommandInProject('yarn ng e2e angular --watch false', projectPath)
+      await runCommandInProject('yarn ng add @cypress/schematic --e2e --ct false --add-ct-specs false', projectPath)
```
ZachJW34, line 37: The `by default` would mean we don't need the extra flags so `yarn ng add @cypress/schematic` should be sufficient

npm/cypress-schematic/src/schematics/ng-generate/cypress-ct-tests/schema.json
```diff
@@ -0,0 +1,25 @@
...
+  "properties": {
+    "project": {
+      "type": "string",
+      "description": "The name of the project.",
+      "alias": "p",
+      "$default": {
+        "$source": "projectName"
+      }
+    },
+    "createSpecs": {
```
ZachJW34, line 15: What is the purpose of this arg? If they say no, the schematic doesn't do anything

npm/cypress-schematic/src/schematics/ng-generate/cypress-ct-tests/index.spec.ts
```diff
@@ -30,11 +31,9 @@ describe('@cypress/schematic:e2e ng-generate', () => {
...
   })
 
-  it('should create cypress spec file', async () => {
-    return schematicRunner.runSchematicAsync('e2e', { name: 'foo', project: 'sandbox' }, appTree).toPromise().then((tree) => {
-      const files = tree.files
-
-      expect(files).to.contain('/projects/sandbox/cypress/e2e/foo.cy.ts')
-    })
+  it('should create cypress component tests alongside components', () => {
+    schematicRunner.runSchematicAsync('specs-ct', { createSpecs: true, project: 'sandbox' }, appTree).pipe(
```
ZachJW34, line 35:
```suggestion
    await schematicRunner.runSchematicAsync('specs-ct', { createSpecs: true, project: 'sandbox' }, appTree).toPromise()
```

I've had to utilize the `done` mocha callback in the past when working with observables to ensure the test executes properly. Using the `toPromise` makes the test lest nested and you don't have to worry about explicitly calling `done`

Same comment for the other schematic unit tests

npm/cypress-schematic/src/schematics/ng-add/files-core/cypress.config.ts
```diff
@@ -0,0 +1,19 @@
+import { defineConfig } from 'cypress'
```
ZachJW34, line 1: Can we suffix all template files with `.template` or `__template__` (not sure which), then we wouldn't have to worry about eslint ignores or tsconfig excludes

ZachJW34 (changes requested):

elylucas:
Giving it a try and noticed what might be a bug. During installation, I selected 'no' to not add CT, but then it continued to prompt me asking if it should add ct specs alongside. I don't think the last prompt should show if the user answers no to the previous:

image

npm/cypress-schematic/src/schematics/utils/index.ts
admah: The issue is that `getSystemPath` is looking for a `Path` not a `string`.  I'm using `normalize` to convert the string to that Path type.

npm/cypress-schematic/src/ct.spec.ts
admah: This command is to add the dependency itself from the local file using `yarn`.

npm/cypress-schematic/src/schematics/ng-generate/cypress-ct-tests/schema.json
admah: Good catch. I'm removing this.

npm/cypress-schematic/src/schematics/utils/index.ts
admah: I didn't find `tree.readDir` to be a good substitute for `readdirSync` here.

npm/cypress-schematic/src/schematics/utils/index.ts
admah: I would really like to do that as well, but can't find a way to implement it.

npm/cypress-schematic/src/schematics/ng-generate/cypress-test/index.ts
admah: This has been removed. Good catch!

npm/cypress-schematic/src/schematics/ng-add/files-core/cypress.config.ts
admah: Yes. I added `.template` and then updated the eslint ignore file.

npm/cypress-schematic/README.md
admah: You can currently pass a `--path` argument to save files in a specific location relative to the project root.

npm/cypress-schematic/README.md
admah: Yes.

npm/cypress-schematic/src/builders/cypress/schema.json
admah: Done!

npm/cypress-schematic/src/schematics/ng-generate/cypress-ct-tests/index.spec.ts
admah: I left one of these tests as-is because it fails using `toPromise` because of how the files are generated by the schematic. I talked through it with @jordanpowell88. All of the other tests were fine.

npm/cypress-schematic/README.md
admah: Good call. I removed `-t` in favor of a default for `e2e` or  `--component` for generating component tests.

PR #23408, opened 2022-08-17
fix: Add component testing to package description

marktnoonan:
Noticed we were not mentioning Component Testing in our description and keywords! Any reason not to make this change?

### User facing changelog

N/A

### PR Tasks

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "cypress",
   "version": "10.6.0",
-  "description": "Cypress.io end to end testing tool",
+  "description": "Cypress.io End-to-End and Component testing tool",
   "private": true,
   "scripts": {
     "binary-build": "node ./scripts/binary.js build",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "cypress",
   "version": "10.6.0",
-  "description": "Cypress.io end to end testing tool",
+  "description": "Cypress.io End-to-End and Component testing tool",
```
ZachJW34, line 4: I'm not against this, but wouldn't adding a description and keywords to the `cli/package.json` make more sense?

emilyrohrbough: pulled from https://docs.cypress.io/guides/overview/why-cypress#In-a-nutshell
> Cypress.io front end testing tool built for the modern web. 

Might reach out to marketing about updating the cypress.io home page which called out
> A complete end-to-end testing experience.

Issue #23336, opened 2022-08-15
Wire Up Cypress Studio in Cypress 10


lmiller1990:
In Cypress 9 we had an experimental studio feature. We are looking to bring this back to Cypress 10.

Spike branch: https://github.com/cypress-io/cypress/pull/23181

AC:

- Can use Cypress Studio with existing UI and E2E
- Wire up Cypress 10 UI to existing Studio hooks
- Write tests for it

Out of scope:

- UI doesn't need to look good (deal with this in https://github.com/cypress-io/cypress/issues/23337)
- Not expected to work with CT (Studio experiment is only defined for E2E - this needs product input)

Issue #23338, opened 2022-08-15
Reintroduce `experimentalStudio` flag


lmiller1990:
Part of the Studio epic: https://github.com/cypress-io/cypress/issues/23230

AC:

- add an `experimentalStudio` flag nested under **e2e**
- when enabled, the Cypress Studio feature should work
- should throw error for CT, informing the user Studio only works with e2e right now, and CT support will come in a future version of Cypress

## Example

Enable studio:

```ts
export default defineConfig({
  e2e: {
    experimentalStudio: true
  }
})
```

This should throw an error. Here's a similar example: https://github.com/cypress-io/cypress/pull/23104/files#r941966899

```ts
export default defineConfig({
  component: {
    experimentalStudio: true
  }
})
```

This is also not allowed:

```ts
export default defineConfig({
  experimentalStudio: true
})
```

Make sure you update the type definitions.

We have a similar pattern for experiments that only work with one testing type. An example is `experimentalSingleTabRunMode`. That PR is [here](https://github.com/cypress-io/cypress/pull/23104/). You could do a similar pattern for this one.

This is where we configure it to error for `experimentalSingleTabRunMode`:  https://github.com/cypress-io/cypress/pull/23104/files#diff-63812dbb23c9353d6d1e500239c7019912de3c33fe540d2cb58768b0b858359fR660-R665.

PR #23413, opened 2022-08-18
chore: wire up Cypress Studio 

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/23336

**Note** targeting a feature branch, `feat/studio-cypress-10`.

### User facing changelog

N/A - internal for now

### Additional details

Wire up Cypress Studio in Cypress 10. There's a bit going on, and it's not fully complete but works at a basic level. It's going into a feature branch, so things like tests, some cleanup etc will come before we merge the whole thing into develop.

Learn about Studio: https://docs.cypress.io/guides/references/cypress-studio#What-you-ll-learn

### Steps to test

WIP, there's probably lots of edge cases that are kind of busted so I wouldn't go too far. Here's a demo of how it works, you are welcome to play around and try it out!

The final review will be towards the end once we've got a proper UI in place. It's hard to really review without the full Studio UI in place - that's why it's targeting a feature branch.

Here's a quick demo (not sure why the test fails, but the wiring is working - you can see it generated a spec from my actions).

https://user-images.githubusercontent.com/19196536/185599079-8505f97c-a7fd-45ef-ab18-fccc9a500754.mp4

Any small bugs or minor fixes will be done in https://github.com/cypress-io/cypress/issues/23461, which is the feature branch. The final review of this is where we need to be more thorough and deliberate.

### PR Tasks


packages/app/cypress/e2e/cypress-in-cypress-e2e.cy.ts
```diff
@@ -148,7 +148,7 @@ describe('Cypress In Cypress E2E', { viewportWidth: 1500, defaultCommandTimeout:
   })
 
   it('shows a compilation error with a malformed spec', { viewportHeight: 596, viewportWidth: 1000 }, () => {
-    const expectedAutHeight = 500 // based on explicitly setting viewport in this test to 596
+    const expectedAutHeight = 456 // based on explicitly setting viewport in this test to 596
```
lmiller1990, line 151: I changed the `<a>` for the AUT URL to be an `<input`> so we can type into it to navigate to a URL for Cypress Studio, and for some reason the wrapping break points for `flex-wrap` is _slightly_ different. I have no idea why and was quite stuck on this, but for all intensive purposes the UI is the same, except things wrap in the AUT header slightly earlier if you have your Cypress window really small. 

We can probably revisit this in the final pass if needed, but I don't think it makes any difference.

packages/app/cypress/e2e/cypress-in-cypress.cy.ts
```diff
@@ -243,7 +243,7 @@ describe('Cypress in Cypress', { viewportWidth: 1500, defaultCommandTimeout: 100
       cy.visitApp()
       cy.contains('dom-content.spec').click()
 
-      cy.contains('http://localhost:4455/cypress/e2e/dom-content.html').should('be.visible')
+      cy.findByTestId('aut-url-input').invoke('val').should('contain', 'http://localhost:4455/cypress/e2e/dom-content.html')
```
lmiller1990, line 246: Bit more specific about what's going on, just a general improvement.

packages/app/cypress/e2e/reporter_header.cy.ts
```diff
@@ -19,18 +19,20 @@ describe('Reporter Header', () => {
     it('filters the list of specs when searching for specs', () => {
       cy.get('body').type('f')
 
-      cy.get('input').type('dom', { force: true })
+      cy.findByTestId('specs-list-panel').within(() => {
```
lmiller1990, line 22: Again more specific, we've got another `<input>` now for the Studio URL to visit for the test, so we need to make this more specific.

packages/app/src/runner/SpecRunnerHeaderOpenMode.vue
```diff
@@ -23,14 +23,16 @@
         >
           <i-cy-crosshairs_x16 :class="[selectorPlaygroundStore.show ? 'icon-dark-indigo-500' : 'icon-dark-gray-500']" />
         </Button>
-        <a
+        <input
```
lmiller1990, line 26: Here is the `<input>` I was talking about.

packages/config/__snapshots__/index.spec.ts.js
```diff
@@ -7,7 +7,6 @@ exports['config/src/index .getBreakingKeys returns list of breaking config keys
   "experimentalRunEvents",
```
lmiller1990, line 7: **NOTE** above file is the store for all studio related things, straight copy paste from https://github.com/cypress-io/cypress/pull/23413/files#diff-c72280ee321ab528ae0e55a0f0d353cece65c1c115c60d89bf83e126b9fb3996

Feel free to take a look but I didn't implement this originally - happy to make improvements if you spot them, though. There's quite a bit going on, once this Studio feature is done we can spend some time on general "how it works", docs etc.

lmiller1990 (commented):
some notes for reviewers

packages/reporter/src/test/test.tsx
```diff
@@ -192,15 +192,13 @@ class Test extends Component<TestProps> {
...
-    if (appState.studioActive) {
-      controls.push(
-        <Tooltip key={`studio-command-${this.props.model}`} placement='right' title='Add Commands to Test' className='cy-tooltip'>
-          <a onClick={this._launchStudio} className='runnable-controls-studio'>
-            <WandIcon />
-          </a>
-        </Tooltip>,
-      )
-    }
+    controls.push(
```
lmiller1990, line 195: Confusingly these are **always** rendered, but hidden by CSS  by default, then un-hidden [here](https://github.com/cypress-io/cypress/blob/develop/packages/reporter/src/runnables/runnables.scss#L88) if `experimentalStudio` is enabled.

Not ideal but this is how it was in 9.x - I'm thinking instead of changing too much, I'm just going to re-introduce Studio with as little changes a possible. We can improve the implementation prior to removing the `experimental` flag - the goal right now is just to re-enable it for users as it was before.

packages/runner/unified-runner.tsx
```diff
@@ -5,7 +5,6 @@ import { Reporter } from '@packages/reporter/src/main'
 import shortcuts from '@packages/reporter/src/lib/shortcuts'
```
lmiller1990, line 5: Nuked above spec file - there's a ton of stubbing and whatnot, I'm not a huge fan of unit testing stores anyway... we will def need coverage, but now we've got Cypress in Cypress that will help a lot with Studio tests. I will also look to add some for the more complex logic once we've got the whole UI implemented.

packages/app/cypress/e2e/cypress-in-cypress-e2e.cy.ts
lmiller1990, line 151: Change `<a>` to `<input>` for the Studio URL input - this caused the breakpoint for `flex: flex-wrap` to change slightly. Seems like it's not a big deal.

packages/app/src/store/studio-store.ts
```diff
@@ -0,0 +1,907 @@
...
+        id: this.testId,
+        err: {
+          ...err,
+          message: saveErrorMessage(err.message),
+          docsUrl: 'https://on.cypress.io/studio-beta',
+        },
+      }
+    },
+
+    setFileDetails (fileDetails) {
```
ZachJW34, line 426: I'm not able to save the file as it fail with error:

```
ENOENT: no such file or directory, open '/Users/zachjw/work/component-test-repos/create-next-app/create-next-app/cypress/e2e/passing.cy.js'
```

Notice the projectName is being appended twice `create-next-app/create-next-app`. I traced it down to [stripCustomProtocol](https://github.com/cypress-io/cypress/blob/b0299c38009792df98ecd706dab29c04f24490dc/packages/driver/src/cypress/stack_utils.ts#L260), might need to tweak the regex

packages/reporter/src/test/test.tsx
```diff
@@ -192,15 +192,13 @@ class Test extends Component<TestProps> {
...
+      <Tooltip key={`studio-command-${this.props.model}`} placement='right' title='Add Commands to Test' className='cy-tooltip'>
+        <a onClick={this._launchStudio} className='runnable-controls-studio'>
+          <WandIcon />
+        </a>
+      </Tooltip>,
+    )
```
ZachJW34, line 201: Note for `_launchStudio` below:

I wasn't able to get studio to launch unless I clicked the icon, navigated away from the runner and then back in. I refactored the function to set the `studioActive` property and it started working

```ts
_launchStudio = (e: MouseEvent) => {
    e.preventDefault()
    e.stopPropagation()

    const { model, events } = this.props

    action('studio:init:test', () => {
      appState.setStudioActive(true)
      events.emit('studio:init:test', model.id)
    })()
}
```

ZachJW34 (approved):
Looks like a good start! Didn't do too deep a dive but it looks like your goal of getting most of it wired up is working. Just a few comments, since it's WIP I'll leave it up to you to address them or not

packages/app/src/runner/event-manager.ts
```diff
@@ -150,7 +147,7 @@ export class EventManager {
     })
 
     this.ws.on('watched:file:changed', () => {
-      this.studioRecorder.cancel()
+      useStudioStore().cancel()
```
mike-plummer, line 150: I'm not very familiar with `pinia` so maybe this is not an issue, but I notice other `use*Store` usages in this file set the store into a shared const rather than making repeated `use*Store` calls. Just a style preference, or is there a functional difference?

mike-plummer (approved):
Had to debug around a bit like Zach to get studio to activate, but once I figured out the trick I was able to record & save a couple specs without issue. Awesome work getting this much working so quickly!

packages/app/src/runner/event-manager.ts
lmiller1990: I just did a full text search + replace - `s/this.studioRecorder/useStudioStore()/g`. Was trying to minimise code changes, but since this seems like a fairly straight forward and obvious improvement, I'll go ahead and make it 👍

packages/app/src/runner/useEventManager.ts
```diff
@@ -37,10 +37,22 @@ export function useEventManager () {
...
+
+      getAutIframeModel().reattachStudio()
+    })
+
     eventManager.on('visit:blank', ({ type }) => {
       getAutIframeModel().visitBlank({ type })
     })
 
+    eventManager.on('run:end', () => {
+      getAutIframeModel().startStudio()
```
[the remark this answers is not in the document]
lmiller1990: You'd certainly think so, but here's 9.x:

- https://github.com/cypress-io/cypress/blob/v9.7.0/packages/runner/src/iframe/iframes.jsx#L80
- Calls https://github.com/cypress-io/cypress/blob/v9.7.0/packages/runner-shared/src/iframe/aut-iframe.js#L448-L452

We do the exact same thing: https://github.com/cypress-io/cypress/pull/23413/files#diff-533ffcf1ce12c3f364ebb004a571fce7c2e34670aa16e6b7ac36a1a5b2cd45b9R484-R488

This implementation has quite a few glaring issues (I mean the original 9.x one). I'm lifting everything as closely as possible in a 1:1 fashion in this PR - we can definitely make some improvements as we go, though. Just sharing this for context on why we do **not** do a check here (we do it in the `startStudio` method, using `isLoading` as the flag).

I agree it's confusing - I'll go ahead and lift the check out of `startStudio` into here.

packages/reporter/src/test/test.tsx
lmiller1990: Good catch - I'd been testing with a spec containing `it('is empty', () => {})`. I'd missed this case. I'll fix it.

packages/app/src/store/studio-store.ts
lmiller1990: Weird, I am **not** seeing this. I may need help to reproduce this one.

lmiller1990:
@mike-plummer @ZachJW34 it sounds like you've both encountered a bug that I am not able to reproduce. I'm going to address all the other comments for now (mainly refactors/tweaks) and we can sync up so I can see how to reproduce the "cannot activate studio" problem you've both encountered.

packages/app/src/runner/useEventManager.ts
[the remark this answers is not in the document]
lmiller1990: Done.

lmiller1990:
New demo: 

https://user-images.githubusercontent.com/19196536/185846703-89f03092-2954-4544-a3af-9a04d415237b.mp4

packages/app/src/runner/event-manager.ts
lmiller1990: Done. Thanks :+1:

mike-plummer:
@lmiller1990 It's working for me after pulling down your latest commits 🥳 . I didn't dig too far into it on Friday but I think the issue was something around the `showedStudioModal` state flag - it looked like there used to be a "first studio run" init modal that isn't there anymore and maybe we were falling into that? (event-manager.ts, line 285) In any event, seems to be working now but I'll let Zach confirm

packages/app/src/runner/SpecRunnerHeaderOpenMode.vue
emilyrohrbough: Why did we need to change this to an input? I think I missed the reason

packages/reporter/src/test/test.cy.tsx
```diff
@@ -16,7 +16,11 @@ describe('test/test.tsx', () => {
...
     cy.mount(<div className="runnable suite">
-      <Test model={model} appState={appState} />
+      <Test
+        model={model}
+        appState={appState}
+        experimentalStudioEnabled={false}
+      />
     </div>)
 
     cy.percySnapshot()
```
emilyrohrbough, line 26: I'd expect tests added for when studio is enabled. When is that work planned if it's not happening here? There are quite a skipped tests in the reporter atm for studio. Here's a few I know off-hand

- https://github.com/cypress-io/cypress/blob/develop/packages/reporter/cypress/e2e/runnables.cy.ts#L174
- https://github.com/cypress-io/cypress/blob/develop/packages/reporter/cypress/e2e/tests.cy.ts#L137
- https://github.com/cypress-io/cypress/blob/develop/packages/reporter/cypress/e2e/suites.cy.ts#L133

packages/app/src/runner/StudioControls.vue
```diff
@@ -0,0 +1,118 @@
+<template>
+  <div v-if="!studioStore.url && studioStore.isActive">
```
emilyrohrbough, line 2: We are missing tests for this component.

emilyrohrbough (changes requested):
We have very little test coverage for studio and we should add these as cypress-in-cypress tests before this is released. We are missing coverage in the app, launchpad and server (no socket tests for studio). We should also add tests to verify the studio json stats are correctly being [read/written with file paths/OS](https://github.com/cypress-io/cypress/blob/develop/packages/server/lib/util/spec_writer.ts#L389).

emilyrohrbough:
Tried out your branch. It wasn't working for me:

https://user-images.githubusercontent.com/14099737/185947140-430a21f8-1768-49b7-b389-822eb53a3ed2.mov

Also noticing we don't have the experiment listed in the settings page and the config option is not shown in the resolved settings:
Screen Shot 2022-08-22 at 9 27 05 AM
Screen Shot 2022-08-22 at 9 27 17 AM

We are also missing the[ setting's localized strings](https://github.com/cypress-io/cypress/blob/develop/packages/frontend-shared/src/locales/en-US.json#L471) as well. The last two experiments we releases had them missing 😅

astone123:
> Also noticing we don't have the experiment listed in the settings page and the config option is not shown in the resolved settings

@emilyrohrbough I think we're going to handle adding the flag back with [this issue](https://github.com/cypress-io/cypress/issues/23338)

packages/reporter/src/runnables/runnable-and-suite.tsx
```diff
@@ -86,7 +93,9 @@ class Runnable extends Component<RunnableProps> {
         })}
         data-model-state={model.state}
       >
-        {model.type === 'test' ? <Test model={model as TestModel} /> : <Suite model={model as SuiteModel} />}
+        {model.type === 'test'
+          ? <Test model={model as TestModel} experimentalStudioEnabled={experimentalStudioActive} />
+          : <Suite model={model as SuiteModel} experimentalStudioActive={experimentalStudioActive} />}
```
astone123, line 98: I think we should use `experimentalStudioEnabled` for this prop name. This property describes whether or not Studio is enabled in the Cypress configuration, not whether Studio is currently active. We should be careful not to conflate the two

ZachJW34:
@lmiller1990 UI is now working for me. Also, the problem with saving the file seems to be only related to Next.js apps, it's working on other applications just fine now.

What I'm seeing for Next is that the `originalFile` that comes from the stack details is `webpack://create-next-app/./cypress/e2e/passing.cy.js`, whereas a normal webpack app is `webpack:///./cypress/e2e/commands/actions/a.cy.js`. Can you try saving a studio generated spec within a Next.js app and see if you run into the problem?

Not a blocker since I know your goal is to try and get close to a 1-to-1 implementation of the previous studio but would be good to followup on this.

packages/app/src/runner/SpecRunnerHeaderOpenMode.vue
marktnoonan: This is here: https://github.com/cypress-io/cypress/pull/23413#discussion_r949976076

I think it was an input in Cypress 9, which I always didn't really love - I'm hoping when we clean up the UX for this, this will only render an input when it is actually editable in studio mode, and will render a plain old link the rest of the time.

packages/launchpad/cypress/e2e/config-warning.cy.ts
```diff
@@ -87,7 +87,9 @@ describe('experimentalSingleTabRunMode', () => {
   })
 })
 
-describe('experimentalStudio', () => {
+// TODO: Figure out this. experimentalStudio is back, but must be nested under E2E? Or, does
```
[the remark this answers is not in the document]
marktnoonan: We should add the issue number to this TODO: https://github.com/cypress-io/cypress/issues/23338

lmiller1990:
> We have very little test coverage for studio and we should add these as cypress-in-cypress tests before this is released. We are missing coverage in the app, launchpad and server (no socket tests for studio). We should also add tests to verify the studio json stats are correctly being [read/written with file paths/OS](https://github.com/cypress-io/cypress/blob/develop/packages/server/lib/util/spec_writer.ts#L389).

Yep, we have an entire ticket for writing a bunch of tests before moving the feature branch to review: https://github.com/cypress-io/cypress/issues/23461

(Note this branch is targeting said feature branch).

packages/reporter/src/test/test.cy.tsx
lmiller1990: Oh I did not realize we had these skipped - I'll go back and add them in now in https://github.com/cypress-io/cypress/issues/23461

packages/app/src/runner/StudioControls.vue
lmiller1990: Yep, thanks for calling this out -- DM'd you, gonna merge this to the feature branch and add a bunch of tests here:  https://github.com/cypress-io/cypress/issues/23461

lmiller1990:
@ZachJW34 thanks for calling out Next.js weirdness, going to triage this separately -- I think 9.x suffers the same problem.

packages/launchpad/cypress/e2e/config-warning.cy.ts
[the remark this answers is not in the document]
lmiller1990: There is https://github.com/cypress-io/cypress/issues/23338

packages/reporter/src/runnables/runnable-and-suite.tsx
```diff
@@ -63,6 +69,7 @@ const Suite = observer(({ eventManager = events, model }: SuiteProps) => {
 export interface RunnableProps {
   model: TestModel | SuiteModel
   appState: AppState
+  experimentalStudioActive: boolean
```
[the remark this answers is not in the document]
lmiller1990: Hmm good point, maybe we just name it `studioEnabled`?

I'll think about this a bit more.

Hmm actually, maybe we just do a check against `Cypress.config('experimentalStudio')` and drop this prop entirely? Let me mull over this.

packages/reporter/src/test/test.tsx
```diff
@@ -106,6 +106,7 @@ interface TestProps {
   runnablesStore: RunnablesStore
   scroller: Scroller
   model: TestModel
+  experimentalStudioEnabled: boolean
```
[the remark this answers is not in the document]
lmiller1990: I would have thought that as long as we set the top level prop in `<Reporter />` it'll get passed down - eg, since it's not optional, we don't need a default value? And TS **should** complain if we forgot to pass it down somewhere -- or maybe the TS config for `reporter` isn't strict enough?

I'll take a look but I don't think setting a default value for a non nullable boolean prop makes sense, does it?

Will definitely get a chance to re-assess once thing final PR opens, but it seems like we should only need default values for nullable props.

packages/app/src/runner/SpecRunnerHeaderOpenMode.vue
lmiller1990: I guess `<input>` as a link is not really the best a11y. We could do a fancy switcharoo depending if studio is active? Or, hack `<a>` to be editable (but this seems not ideal).

I had a ton of problems with tailwind doing weird things - for some reason the sizing changes between `<a>` and `<input>` and messed up the flow of the flex-wrap, even though tailwind claims to do a full CSS reset. Probably need to jam with a CSS-minded individual to sort this out.

packages/reporter/src/instruments/instrument-model.ts
```diff
@@ -1,5 +1,5 @@
 import { observable } from 'mobx'
-import { TestState } from '../test/test-model'
+import { Instrument, TestState } from '@packages/types'
```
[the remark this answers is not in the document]
lmiller1990: Done, thanks

packages/reporter/src/runnables/runnable-and-suite.tsx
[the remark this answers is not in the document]
lmiller1990: I changed it to `studioEnabled`.

packages/reporter/src/runnables/runnable-and-suite.tsx
lmiller1990: Done, globally renamed to `studioEnabled`. I also dropped the `experimental` prefix, which doesn't add much value internally as pointed out by https://github.com/cypress-io/cypress/pull/23413/files#r951467056

lmiller1990:
I unskipped the skipped tests in `reporter`:

- https://github.com/cypress-io/cypress/pull/23413/commits/9c24014b00a42125ae4209a224fa6ab5db97ec20
- https://github.com/cypress-io/cypress/pull/23413/commits/c7d66f5ab140f171b126b7cf10a804317267e42e

In https://github.com/cypress-io/cypress/pull/23413/commits/c7d66f5ab140f171b126b7cf10a804317267e42e there was some race condition I think - I just moved the tests around (reorder) and it's working fine now. I am not sure if this was originally flaky or something changed. Either way, these tests are only really unit/integration - I'm excited to add some real E2E tests next using Cypress in Cypress.

lmiller1990:
Ok, gonna merge this up and continue working on tests in #23461.

packages/launchpad/cypress/e2e/config-warning.cy.ts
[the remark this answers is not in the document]
lmiller1990: It's done here https://github.com/cypress-io/cypress/pull/23506

packages/reporter/src/test/test.cy.tsx
lmiller1990: Ok un-skipped.

I will write E2E tests in a separate PR once the UI stuff is done, and you'll be tagged to review 💯 

https://github.com/cypress-io/cypress/pull/23413/commits/9c24014b00a42125ae4209a224fa6ab5db97ec20
https://github.com/cypress-io/cypress/pull/23413/commits/c7d66f5ab140f171b126b7cf10a804317267e42e

Issue #23333, opened 2022-08-14
Remove Nullable type from globals


dartess:
### Current behavior

Cypress adds `Nullable` global type in https://github.com/cypress-io/cypress/blob/56234e52d6d1cbd292acdfd5f5d547f0c4706b51/cli/types/cypress-type-helpers.d.ts .

### Desired behavior

Cypress does not add unexpected and unnecessary global types that are not clearly Cypress related.
If it is an internal type, please export and import it explicitly.

### Test code to reproduce

`type A = Nullable<string>`

### Cypress Version

10.4.0

### Node version

any

### Operating System

any

nagash77:
@dartess Can you please elaborate on why you think this is a problem?  Can you also provide a reproducible example where we can see the problem you are describing.

dartess:
Ok

1. Your type is called `Nullable`. It looks like this type works exactly the opposite of TypeScript's built-in `NonNullable` type. But it is not. `NonNullable` exclude `null` _and_ `undefined`, but your `Nullable` add only `null`. It's not obvious.

2. Cypress is not modular by design (like all other modern software) and I put up with it. But global variables and Cypress types are at least obviously related to it. Guessing from the code that the Nullable type is not a built-in type in the typescript, but a global type from Cypress, can be difficult. This is how you mislead users. They may start relying on undocumented global Cypress types without even knowing it.

3. By the way, where in the documentation does it say that Cypress will add such a global type to your code?

In general, there is no reason why Cypress should add such non-obvious global types to its users' code base. Moreover, Cypress itself uses this type literally in one place. And nothing prevents it from being imported and exported once.

nagash77:
@dartess after internal review, the team agrees with your perspective.  I will route it to the appropriate team, however due to current backlog lengths I would not expect a fix for this in the near future from our team.  I have marked it as a 'Good First Issue' to hopefully encourage community contributions to address this issue.  Thanks for your work in reporting it!

PR #23439, opened 2022-08-18
fix: prevent deliver global Nullable type

dartess:
- Closes https://github.com/cypress-io/cypress/issues/23333

### User facing changelog

Cypress no longer adds `Nullable` internal type to the global scope.

### Additional details

This type could be falsely assumed to be a built-in TypeScript type, because it was available to all Cypress users globally..

### Steps to test

N/A

Users no longer have access to a global type intended for internal use.

### PR Tasks


dartess:
I'm not very familiar with the concept of namespaces and Triple-Slash Directives. But in general, I liked the occasional `Cy` prefix on types.

I see three solutions:

1. current. wrap helper types in a namespace.
2. rename the type to CyNullable (looks weird)
3. delete this type. I don't really understand how it's useful - `Nullable<string>` is more cumbersome than `string | null`.

cli/types/cypress-type-helpers.d.ts
```diff
@@ -1,2 +1,4 @@
 // type helpers
-type Nullable<T> = T | null
+declare namespace CyTypeHelpers {
```
mike-plummer, line 2: Generally we use `Cypress` as our namespace in our other `d.ts` files, can we do the same here?

mike-plummer (commented):
Hi @dartess , thanks for opening a PR! Just one request to align with our existing namespacing

cli/types/cypress.d.ts
```diff
@@ -2876,7 +2877,7 @@ declare namespace Cypress {
      * To enable test retries only in runMode, set e.g. `{ openMode: null, runMode: 2 }`
      * @default null
      */
-    retries: Nullable<number | { runMode?: Nullable<number>, openMode?: Nullable<number> }>
+    retries: Cypress.Nullable<number | { runMode?: Cypress.Nullable<number>, openMode?: Cypress.Nullable<number> }>
```
emilyrohrbough, line 2880: honestly, this isn't hardly used. IMO it'd be similar to remove it all together:

```suggestion
    retries: number | { runMode?: number | null, openMode?: number | null } | null
```

emilyrohrbough (approved):

cli/types/cypress.d.ts
dartess: I am not a cypress developer and do not have voting rights. but I came here just because in our project we came to a consensus that the Nullable construction is redundant and does not add readability. Right now I'm cleaning up the code.

Should I remove this type here as well? Or does it require further discussion?

mike-plummer (approved):
LGTM, thanks for the contribution! There's a couple failing tests on CI (not your fault) that I need to get green, then I'll get this merged

PR #23441, opened 2022-08-18
feat: Add svelte detection to wizard

### LINKED ISSUE #23432: Svelte CT setup
### What would you like?

Svelte apps should be autodetected during CT setup and scaffolded with Svelte specific files

### Why is this needed?

To support Svelte Component Testing

### Other

The flow should be similar to how we detect/scaffold other frameworks for CT support:

Screen Shot 2022-08-18 at 10 32 06 AM

Requirements:

- Autodetects Svelte during CT setup
- Lists required dependencies
- Scaffolds Svelte specific CT files
- System tests verifying Svelte setup flow

rockindahizzy:
- Closes #23432

### Steps to test

1. Open launchpad
1. Select a svelte project
1. Select `Component Testing`
1. Wizard should detect Svelte and correct bundler
1. Click `Next Step`
1. Wizard should correctly validate that required dependencies are installed
2. Click `Continue`
3. The following criteria should be met in the `Configuration Files`
    - `cypress.config.js` should have a devServer set up as shown below:
    ```js
    devServer: {
        framework: "svelte",
        bundler: "vite",
      },
    ```
    - `cypress/support/component.js` should include the correct mount
      ```js 
      import { mount } from 'cypress/svelte'`
      ```

### PR Tasks


packages/scaffold-config/src/dependencies.ts
```diff
@@ -142,6 +142,15 @@ export const WIZARD_DEPENDENCY_ANGULAR_PLATFORM_BROWSER_DYNAMIC = {
   minVersion: '>=13.0.0',
 } as const
 
+export const WIZARD_DEPENDENCY_SVELTE = {
```
rockindahizzy, line 145: Curious what this is used for. I couldn't find a reference in the project

packages/scaffold-config/src/frameworks.ts
```diff
@@ -61,9 +61,16 @@ function getBundlerDependency (bundler: WizardBundler['type'], projectPath: stri
   }
 }
 
-export const WIZARD_MOUNT_MODULES = ['cypress/react', 'cypress/react18', 'cypress/vue', 'cypress/vue2', 'cypress/angular'] as const
+export const WIZARD_MOUNT_MODULES = [
```
rockindahizzy, line 64: Curious what this is used for. I couldn't find a reference in the project

packages/scaffold-config/src/frameworks.ts
```diff
@@ -266,4 +273,23 @@ export const WIZARD_FRAMEWORKS = [
...
+    dependencies: (bundler: WizardBundler['type'], projectPath: string): Promise<DependencyToInstall[]> => {
+      return Promise.all([
+        getBundlerDependency(bundler, projectPath),
+        isDependencyInstalled(dependencies.WIZARD_DEPENDENCY_SVELTE, projectPath),
+      ])
+    },
+    codeGenFramework: 'svelte',
+    glob: '*.svelte',
+    mountModule: mountModule('cypress/svelte'),
+    supportStatus: 'alpha',
```
mike-plummer, line 292: Didn't we have a conversation around not using `alpha` for our newly-supported frameworks? Or was that just for Angular?

packages/scaffold-config/test/unit/detect.spec.ts
```diff
@@ -219,6 +219,34 @@ describe('detectFramework', () => {
...
+      { dependency: 'vite', version: '3.0.0' },
+    ])
+
+    const actual = await detectFramework(projectPath)
+
+    expect(actual.framework?.type).to.eq('svelte')
+    expect(actual.bundler?.type).to.eq('vite')
+  })
+
+  it(`Svelte and Vite`, async () => {
```
mike-plummer, line 236: _*webpack*_

packages/scaffold-config/test/unit/detect.spec.ts
```diff
@@ -219,6 +219,34 @@ describe('detectFramework', () => {
...
+    const projectPath = await scaffoldMigrationProject('svelte-webpack-unconfigured')
+
+    fakeDepsInNodeModules(projectPath, [
+      { dependency: 'svelte', version: '3.0.0' },
+      { dependency: 'webpack', version: '5.0.0' },
```
mike-plummer, line 241: Do we need a webpack 4 vs 5 test?

rockindahizzy: That's a good question. I'm not sure how testing different versions of the frameworks add value for the scaffolding functionality.

ZachJW34: We can add a test for both versions but I don't think it adds much value

ZachJW34: Maybe we should shift away from `alpha` to `experimental`, since the alpha tag doesn't really mean much (we aren't going to support Svelte Alpha -> Svelte Beta -> Svelte Stable). Would be a good thing to bring up next ct-product meeting

packages/scaffold-config/src/frameworks.ts
ZachJW34: I don't think we ever came to a good conclusion on why we would add `alpha`. My take it that we would do it as a safeguard to allow us to make breaking changes to the framework without forcing a breaking change of Cypress itself.

packages/scaffold-config/src/frameworks.ts
ZachJW34, line 64: I thought I deleted this. Should be unused so you can remove if unused.

system-tests/projects/svelte-webpack-unconfigured/README.md
```diff
@@ -0,0 +1,64 @@
+# svelte app
```
ZachJW34, line 1: Let's remove this file and the `.gitignore` to reduce the file bloat. This goes for both the vite and webpack app, so any files that aren't actively used should be removed.

ZachJW34 (changes requested):
Can you add tests in `packages/launchpad/cypress/e2e/scaffold-component-testing.cy.ts` for Svelte (Webpack + Vite)

packages/scaffold-config/test/unit/detect.spec.ts
mike-plummer: > We can add a test for both versions but I don't think it adds much value

Sounds good, just curious since we already had a webpack 5-specific test in this file

packages/scaffold-config/src/frameworks.ts
```diff
@@ -61,9 +61,16 @@ function getBundlerDependency (bundler: WizardBundler['type'], projectPath: stri
...
+  'cypress/react',
+  'cypress/react18',
+  'cypress/vue',
+  'cypress/vue2',
+  'cypress/angular',
+  'cypress/svelte',
+] as const
 
-export type WizardMountModule = 'cypress/react' | 'cypress/react18' | 'cypress/vue' | 'cypress/vue2' | 'cypress/angular'
+export type WizardMountModule = 'cypress/react' | 'cypress/react18' | 'cypress/vue' | 'cypress/vue2' | 'cypress/angular' | 'cypress/svelte'
```
lmiller1990, line 73: We can infer it to reduce duplication:

```ts
export type WizardMountModule = Awaited<ReturnType<typeof WIZARD_FRAMEWORKS[number]['mountModule']>>
```

🎉

packages/scaffold-config/src/frameworks.ts
lmiller1990: I am fairly sure I remember this, too.

My recollection (and I share this opinion) is we just drop alpha and beta entirely - I think we can ship without a label.

We did this with Angular, too - I think Svelte should follow suit?

The only public API we've committed to across frameworks is `mount()` - and I'm sure this won't change.

packages/scaffold-config/test/unit/supportFile.spec.ts
```diff
@@ -220,4 +220,88 @@ describe('supportFileComponent', () => {
       })
     }
   })
+
+  context('svelte', () => {
+    for (const mountModule of ['cypress/svelte'] as const) {
+      it(`handles ${mountModule} and JS`, () => {
```
lmiller1990, line 226: Any reason to use a `for` loop for an array with a single value - do we expect to add more to this array?

system-tests/projects/svelte-vite-unconfigured/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Logs
```
lmiller1990, line 1: Do we need this? I think all of these are ignored in the top level `.gitignore`?

lmiller1990 (commented):
Couple of small comments/questions. Main q we should answer is are we doing this as alpha/beta/experimental/no label?

packages/scaffold-config/src/frameworks.ts
rockindahizzy: I was having issues trying to implement this. Seems like `Awaited` is implemented in `4.5` we are on  `4.4.4`

lmiller1990: Hmm can we just update to TS 4.5.0? I think it's pretty easy from 4.4.4.

Alternatively I think you can implement `Awaited` with TS - there's probably an implementation you can grab for TS <= 4.5. But updating to TS 4.5 seems like a better option?

ZachJW34: I've created a (draft) PR here. Decent amount of changes so didn't want to change the scope of this PR

rockindahizzy: Going to let this change be handled in that PR, thanks @ZachJW34

mike-plummer:
Does there need to be an update to the types in `cypress.d.ts` for `DevServerConfigOptions` for the webpack and vite cases?

rockindahizzy:
> @rockindahizzy Does there need to be an update to the types in `cypress.d.ts` for `DevServerConfigOptions` for the webpack and vite cases, or is that being handled by another segment of this work?

@mike-plummer, good call out. Looks like that is happening [here](https://github.com/cypress-io/cypress/pull/23519/files#diff-9fde1fb2f7eaba5ea49b585c811461495bfbba88e614e900a96863ecfe6b980a)

packages/scaffold-config/src/frameworks.ts
rockindahizzy: Based on our CT sync with Brian today, it seems like Alpha is the right move. Going to add this back

lmiller1990 (commented):
Noted here https://cypressio.slack.com/archives/C011BAPC614/p1661385240122829

Code is fine but `zachw/svelte-ct` which you are targeting has conflicts with develop - make sure to resolve those then backmerge  `zachw/svelte-ct` into here before merging. Going to leave a "request changes" to ensure this happens - other than that ✅

packages/scaffold-config/src/frameworks.ts
lmiller1990: Nice, `Awaited` is super nice

rockindahizzy (commented):
Review after resolving conflicts seems good

ZachJW34 (approved):

PR #23442, opened 2022-08-18
feat(webkit): fix multidomain driver tests in WebKit

### LINKED ISSUE #23832: expecting an array or an iterable object but got [object Null] using playwright-webkit
### Current behavior

I am experiencing a couple of errors just using the new browser experimental feature.

* TypeError
expecting an array or an iterable object but got [object Null] Because this error occurred during a `before each` hook we are skipping all of the remaining tests.

* TypeError
expecting an array or an iterable object but got [object Null]

### Desired behavior

Should run my tests without problems.

### Test code to reproduce

https://github.com/JoanEsquivel/cypress-webkit

1. Just install dependencies.
2. Open the test runner
3. Run for example:
* cypress/e2e/1-helloworld.ct.ts
* cypress/e2e/10-checkboxes.cy

### Cypress Version

10.8.0

### Node version

v16.13.0

### Operating System

Windows 10

### Debug Logs

```shell
TypeError
expecting an array or an iterable object but got [object Null]
View stack trace
 Print to console
TypeError: expecting an array or an iterable object but got [object Null]
    automation@http://uitestingplayground.com/__cypress/runner/cypress_runner.js:153782:63
    @[native code]
```

### Other

I want to say that **this could be my mistake doing a wrong usage of typescript or some library**. However, it is weird that this issue is only happening for webkit.

Also I see the same error from "bestu" in the discord channel General/general.

image

Appreciate your help.

### LINKED ISSUE #24116: Cy.origin fails gracefully for webkit until full support is ready

flotwig:
- Fixes #24116
- Fixes #23832

### User facing changelog

- Adds support for `cy.session()` to experimental WebKit.
- Ensures that with `experimentalSessionAndOrigin` enabled, tests are still able to run against webkit.

### Additional details

- [x] Depends on #23579 

### PR Tasks


packages/driver/cypress/e2e/e2e/origin/basic_login.cy.ts
```diff
@@ -1,4 +1,4 @@
-describe('basic login', () => {
+describe('basic login', { browser: '!webkit' }, () => {
```
mjhenkes, line 1: Skipping cy.origin tests for webkit

packages/driver/cypress/e2e/e2e/origin/origin.cy.ts
```diff
@@ -224,7 +224,9 @@ describe('cy.origin', () => {
     describe('errors', () => {
       it('propagates secondary origin errors to the primary that occur within the test', (done) => {
         cy.on('fail', (err) => {
-          expect(err.message).to.include('variable is not defined')
+          const undefinedMessage = Cypress.isBrowser('webkit') ? 'Can\'t find variable: variable' : 'variable is not defined'
```
mjhenkes, line 227: while this isn't used right now, it will be after we enable cy.origin for webkit.

packages/driver/cypress/e2e/webkit.cy.ts
```diff
@@ -9,17 +9,7 @@ describe('WebKit-specific behavior', { browser: 'webkit' }, () => {
     cy.origin('foo', () => {})
   })
 
-  it('cy.session() is disabled', (done) => {
```
mjhenkes, line 12: Removing this test since we're enabling cy.session for webkit.

packages/driver/src/cypress/cy.ts
```diff
@@ -581,6 +565,22 @@ export class $Cy extends EventEmitter2 implements ITimeouts, IStability, IAssert
 
         // If the runner can communicate, we should setup all events, otherwise just setup the window and fire the load event.
         if (isRunnerAbleToCommunicateWithAUT) {
+          if (this.Cypress.isBrowser('webkit')) {
```
mjhenkes, line 568: Moved this logic to only run if the runner is able to communicate with the AUT.

We may want to add this patch to the injection code for cy.origin when we enable it.

packages/server/lib/browsers/webkit-automation.ts
```diff
@@ -30,22 +30,25 @@ function convertSameSiteExtensionToCypress (str: CyCookie['sameSite']): 'None' |
   return str ? extensionMap[str] : undefined
 }
 
-const normalizeGetCookieProps = (cookie: any): CyCookie => {
-  if (cookie.expires === -1) {
-    delete cookie.expires
+const normalizeGetCookieProps = ({ name, value, domain, path, secure, httpOnly, sameSite, expires }: playwright.Cookie): CyCookie => {
```
mjhenkes, line 33: Most changes to this file are around typing.

packages/server/lib/browsers/webkit-automation.ts
```diff
@@ -256,15 +299,34 @@ export class WebKitAutomation {
...
+   * Clear all cookies
+   * @param cookies for compatibility reasons return the passed in cookies.
+   * @returns cookies cleared
+   */
+  private async clearCookies (cookies: [CookieFilter?]): Promise<CyCookie[]> {
+    const allCookies = await this.getCookies()
+
+    await this.context.clearCookies()
+
+    return allCookies
```
mjhenkes, line 330: clear cookies must return cookies or else there is a type error upstream.

packages/driver/cypress/e2e/commands/sessions/sessions.cy.js
```diff
@@ -24,6 +24,27 @@ const clearAllSavedSessions = () => {
   })
 }
 
+// In webkit, the clear page and clear cookies, etc log messages may be reversed. This isn't an issue, but we just want to test we have both messages.
```
emilyrohrbough, line 27: Do we know why this is the case? The clear page should be awaited before clearing cookies.

https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cy/commands/sessions/index.ts#L360
```js
          await navigateAboutBlank()
          await sessions.clearCurrentSessionData()
```

mjhenkes: My assumption is that it is a timing issue with the logs and that we don't await every action in navigateAboutBlank()

packages/driver/src/cross-origin/events/socket.ts
```diff
@@ -4,7 +4,15 @@ export const handleSocketEvents = (Cypress) => {
...
     path: Cypress.config('socketIoRoute'),
     transports: ['websocket'],
-  }).connect()
+  })
+
+  webSocket.on('connect_error', () => {
+    // fall back to polling if websocket fails to connect (webkit)
+    // https://github.com/socketio/socket.io/discussions/3998#discussioncomment-972316
+    webSocket.io.opts.transports = ['polling', 'websocket']
+  })
```
flotwig: It turns out that this fall-back approach causes issues. On some projects (like `cypress-documentation`), the websocket will take the maximum time to error out (20 seconds, as defined by socket.io), which results in a 20-second delay before the socket.io connection gets established. This socket.io connection code should be updated to explicitly use `Cypress.isBrowser('webkit')` to avoid the possible 20-second delay. Here's how it's done outside of x-origin:

https://github.com/cypress-io/cypress/blob/26a71fe0193f74758e5f03b141ee76aafa367c17/packages/app/src/runner/index.ts#L34-L42

This could be a good chance to unify these two areas as well, so the implementations don't get further out of sync.

flotwig:
This PR seems to fix #23992 even for non-multidomain users - @mjhenkes do you know which change in particular fixed this? Nothing's leaping out at me from the diff.

mjhenkes:
@flotwig, the root of the problem is that somewhere down the line clear all cookies is expected to return the cookies it cleared. I don't know exactly where the expectation is set, but retuning an array of cookies we clear resolves this problem and brings us inline with the return type for CDP, even if that array is empty.

in webkit-automation: https://github.com/cypress-io/cypress/blob/webkit-multidomain/packages/server/lib/browsers/webkit-automation.ts#L328

in cdp-automation: https://github.com/cypress-io/cypress/blob/webkit-multidomain/packages/server/lib/browsers/cdp_automation.ts#L367 (the cookies map into an array)

Without this fix, enabling sessions breaks webkit completely.

packages/driver/src/cross-origin/events/socket.ts
mjhenkes: Unified here: https://github.com/cypress-io/cypress/pull/23442/commits/d3085db57ebbdb61da73c4dba3ed3999400a6ade

I moved the socket creation logic into the socket package because the driver package cannot depend on the app package.

.circleci/config.yml
```diff
@@ -2437,11 +2438,10 @@ linux-x64-workflow: &linux-x64-workflow
         context: test-runner:cypress-record-key
         requires:
           - build
-    # TODO: Implement WebKit network automation to fix the majority of these tests before re-enabling
-    # - driver-integration-tests-webkit-experimentalSessionAndOrigin:
-    #     context: test-runner:cypress-record-key
-    #     requires:
-    #       - build
+    - driver-integration-tests-webkit-experimentalSessionAndOrigin:
```
ryanthemanuel, line 2441: 🎉

packages/driver/cypress/e2e/e2e/origin/basic_login.cy.ts
ryanthemanuel, line 1: Is there a github issue to track on enabling all of these tests in webkit? I just want to make sure nothing gets lost in the shuffle

ryanthemanuel (approved):

packages/driver/cypress/e2e/e2e/origin/basic_login.cy.ts
mjhenkes: Yes, here is the issue for enabling cy.origin for webkit https://github.com/cypress-io/cypress/issues/23799

packages/driver/cypress/e2e/commands/cookies.cy.js
```diff
@@ -2,6 +2,25 @@ const { assertLogLength } = require('../../support/utils')
...
 
+describe('src/cy/commands/cookies - no stub', () => {
+  it('clears all cookies', () => {
+    cy.setCookie('foo', 'bar')
+    cy.getCookies().should('have.length', 1)
+    cy.clearCookies()
+    cy.getCookies().should('have.length', 0)
+  })
+
+  it('clears a single cookie', () => {
```
emilyrohrbough, line 13: What is the behavior when you clear a specific cookie and it does not exist?

packages/server/lib/browsers/webkit-automation.ts
```diff
@@ -30,22 +30,25 @@ function convertSameSiteExtensionToCypress (str: CyCookie['sameSite']): 'None' |
...
-    delete cookie.expires
+const normalizeGetCookieProps = ({ name, value, domain, path, secure, httpOnly, sameSite, expires }: playwright.Cookie): CyCookie => {
+  const cyCookie: CyCookie = {
+    name,
+    value,
+    domain,
+    path,
+    secure,
+    httpOnly,
+    // Use expirationDate instead of expires 🤷‍♀️
```
emilyrohrbough: CyCookie was modeled off of Chrome.cookie from what I can tell. maybe drop the shrugging lady? 😄

packages/server/lib/browsers/webkit-automation.ts
```diff
@@ -30,22 +30,25 @@ function convertSameSiteExtensionToCypress (str: CyCookie['sameSite']): 'None' |
...
+    ...expires !== -1 ? { expirationDate: expires } : {},
+  } as CyCookie
```
emilyrohrbough: You defined this object type as CyCookie above with `cyCookie: CyCookie` so no need to cast it here
```suggestion
  }
```

packages/server/lib/browsers/webkit-automation.ts
```diff
@@ -88,17 +91,33 @@ let requestIdCounter = 1
 const requestIdMap = new WeakMap<playwright.Request, string>()
 let downloadIdCounter = 1
 
+type WebKitAutomationOpts = {
```
emilyrohrbough, line 95: 💚

packages/server/lib/browsers/webkit-automation.ts
```diff
@@ -256,15 +299,33 @@ export class WebKitAutomation {
 
     return normalizeGetCookieProps(cookie)
   }
-
-  private async clearCookie (filter: CookieFilter) {
+  /**
+   * Clears one specific cookie
```
emilyrohrbough, line 305: nice: we don't have docs like this in many places

packages/socket/lib/browser.ts
```diff
@@ -5,3 +5,11 @@ export type { Socket } from 'socket.io-client'
 export {
   io as client,
 }
+
+export function createWebsocket ({ path, browserFamily }: { path: string, browserFamily: string}) {
+  return io({
+    path,
+    // TODO(webkit): the websocket socket.io transport is busted in WebKit, need polling
```
emilyrohrbough, line 12: What is the to-do around this? Wait for a new version of Webkit? Is there something special we need to do in Cypress to align these? or is this just a webkit thing and this comment could be updated to reflect that?

packages/server/lib/browsers/webkit-automation.ts
```diff
@@ -127,6 +146,8 @@ export class WebKitAutomation {
 
     let promises: Promise<any>[] = []
 
+    if (this.shouldMarkAutIframeRequests) promises.push(this.markAutIframeRequests())
```
emilyrohrbough, line 151: Should we add a TODO to remove this if check once session/origins goes GA? We'd should also be able to drop the param to create this. Might also be worth passing experimentalSessionAndOrigin through to make this obvious it needs cleaned up

packages/server/lib/browsers/webkit-automation.ts
mjhenkes: IMO we should do this more often, it's hard to get a read on function expectations without some sort of doc.

packages/socket/lib/browser.ts
mjhenkes: There's not a TODO, But the comment should remain.

packages/server/lib/browsers/webkit-automation.ts
emilyrohrbough: 💯

packages/socket/lib/browser.ts
emilyrohrbough:
```suggestion
    // the websocket socket.io transport is busted in WebKit, so need polling instead
```

flotwig: https://github.com/cypress-io/cypress/issues/23807 is the follow-up for this issue. It needs to be fixed before stable WebKit, the Network tab gets very noisy with polling.

```suggestion
    // TODO(webkit): the websocket socket.io transport is busted in WebKit, need polling
    // https://github.com/cypress-io/cypress/issues/23807
```

packages/server/lib/browsers/webkit-automation.ts
mjhenkes: added

packages/socket/lib/browser.ts
mjhenkes: updated

emilyrohrbough (approved):

PR #23446, opened 2022-08-18
fix: Detect user-configured browsers

ryanjduffy:
- Closes 

### User facing changelog

Adding a custom browser configuration will no longer emit a warning when trying to use that browser

### Additional details

If you specify a custom browser (either directly in `setupNodeEvents` or via a plugin) and try to use that browser via `--browser`, Cypress logs a warning

image

I expect this isn't the correct fix for this problem hence the draft status but hoping to get some feedback. The issue seems to be that the logic within `makeDataContext` only resolves the machine browsers based on the known browsers and ignores any user-configured browsers. Perhaps I'm just configuring things wrong and there's another way to do it.

### Steps to test

* Add a new browser like the Brave example on https://docs.cypress.io/guides/guides/launching-browsers#Customize-available-browsers
* Launch `npx cypress run --browser Brave` (or whatever you've called the custom browser*

Expected:
- Tests are run

Actual:
- Test are run but a warning is issued beforehand

### PR Tasks


packages/server/lib/makeDataContext.ts
```diff
@@ -46,9 +47,20 @@ export function makeDataContext (options: MakeDataContextOptions): DataContext {
       close: browsers.close,
       getBrowsers,
       async ensureAndGetByNameOrPath (nameOrPath: string) {
+        const projectConfig: FullConfig = await (ctx as DataContext).project.getConfig()
```
ryanjduffy, line 50: This type assertion seems like it should be unnecessary (and isn't used below) but I couldn't avoid the `any` without it. Improvements welcome!

packages/server/lib/makeDataContext.ts
```diff
@@ -46,9 +47,20 @@ export function makeDataContext (options: MakeDataContextOptions): DataContext {
...
+        const declaredBrowsers = projectConfig.browsers?.map<FoundBrowser | undefined>((b) => {
+          if (!b.path) return
```
ryanjduffy, line 52: It seems that only browsers injected by the user or plugins have a `path` set so I'm using it as the proxy for custom browsers. Seem reasonable?

ryanjduffy (commented):
I think this is ready-ish. Left a couple discussion points below. Could also use some advice on adding tests for this.

packages/server/lib/makeDataContext.ts
emilyrohrbough: @ryanjduffy Thank you for opening a change to update improve this check! Considering where this is used, it;'d likely be cheaper and cleaner from a start-up perspective to update `ensureAndGetByNameOrPath` to be a pass through to the imported `ensureAndGetByNameOrPath`

```diff
  getBrowsers,
- async ensureAndGetByNameOrPath (nameOrPath: string) {
-    ...
- },
+ ensureAndGetByNameOrPath
```

and instead update where this is called in the data-context package to supply the list of browsers.
Jumping through the flow:
- we [set the active browser](https://github.com/cypress-io/cypress/blob/develop/packages/data-context/src/data/ProjectLifecycleManager.ts#L267) after we finish loading the config
- this will then call [`ensureAndGetByNameOrPath`](https://github.com/cypress-io/cypress/blob/develop/packages/data-context/src/data/ProjectLifecycleManager.ts#L284) if a browser was provided via the CLI. Seems like we should be able to pull up the machine browsers check and pass it in here.

packages/server/lib/makeDataContext.ts
lmiller1990: Hi 👋 

You can probably do `const projectConfig = await getCtx().project.getConfig()` to get the correct type inference (which will return a reference to `ctx`).

packages/server/lib/makeDataContext.ts
```diff
@@ -46,9 +47,20 @@ export function makeDataContext (options: MakeDataContextOptions): DataContext {
...
+            path: b.path,
+            version: b.version,
+            majorVersion: b.majorVersion,
+            custom: true,
+          }
+        }).filter<FoundBrowser>((b): b is FoundBrowser => b != null) || []
         const browsers = await ctx.browser.machineBrowsers()
 
-        return await ensureAndGetByNameOrPath(nameOrPath, false, browsers)
+        return await ensureAndGetByNameOrPath(nameOrPath, false, [...declaredBrowsers, ...browsers])
```
lmiller1990, line 63: I wonder if we want to do a check to make sure there isn't any duplicates here - it shouldn't really happen, but if someone adds a default browser via `cypress.config.ts` for whatever reason, it'd show up twice.

ryanjduffy: What would you de-dupe on? family / major version / minor version / path?

ryanjduffy: Just ran across a `removeDuplicates` method elsewhere which uses name and version
```
  const removeDuplicates = (val) => {
    return _.uniqBy(val, (browser: FoundBrowser) => {
      return `${browser.name}-${browser.version}`
    })
  }
```

lmiller1990: If this is the pattern, maybe we just go with that. I wonder if we can make that a reused utility? Thoughts?

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -296,7 +296,7 @@ export class ProjectLifecycleManager {
 
     // lastBrowser is cached per-project.
     const prefs = await this.ctx.project.getProjectPreferences(path.basename(this.projectRoot))
-    const browsers = await this.ctx.browser.machineBrowsers()
+    const browsers = await this.ctx.browser.allBrowsers()
```
ryanjduffy, line 299: This also fixes restoring the last browser if it's added via the config

packages/server/lib/makeDataContext.ts
ryanjduffy: I've consolidated this into `BrowserDataSource` and exported it from there.

packages/server/lib/makeDataContext.ts
ryanjduffy: @emilyrohrbough Appreciate the feedback! It's helped me track how this data is used. I've taken a slightly different approach by consolidating both browser lookups (machine and user) into `BrowserDataSource` and then exposed an `allBrowsers` method to fetch both. I like that this approach keeps the "concerns" of the browser config in one spot and makes the diff more clear.

The added bonus of this was that it allowed me to fix `cypress open` to remember the last browser when it was a user browser since that logic can now rely on `allBrowsers`.

packages/data-context/src/sources/BrowserDataSource.ts
```diff
@@ -58,6 +72,37 @@ export class BrowserDataSource {
...
+  /**
+   * Gets the browsers from the project config, caching the Promise on the
+   * coreData so we only look them up once
+   */
+  userBrowsers () {
+    if (!this.ctx.coreData.userBrowsers) {
+      const p = this.ctx.project.getConfig()
+
+      this.ctx.coreData.userBrowsers = p.then(async (cfg) => {
+        return cfg.browsers?.map<FoundBrowser | undefined>((b) => {
```
emilyrohrbough, line 84: at this point, won't the `cfg.browsers` include the found machine browsers as well as the user-defined browsers?

ryanjduffy: It does! I didn't catch that. What is the authoritative source for "browsers from the config file" (is this a thing?) vs "browsers added via a plugin" vs "browsers on the machine" vs "resolved browsers from the full config" ?

I'll push a commit that tries to "4" - "3" but it seems suspect.

ryanjduffy: @emilyrohrbough - following up to see if you have feedback on the updates I pushed. Thanks!!

packages/data-context/src/sources/BrowserDataSource.ts
```diff
@@ -22,6 +23,14 @@ if (os.platform() === 'win32') {
 
 const platform = os.platform()
 
+function getBrowserKey<T extends {name: string, version: string | number}> (browser: T) {
+  return `${browser.name}-${browser.version}`
+}
+
+export function removeDuplicateBrowsers (browsers: FoundBrowser[]) {
+  return _.uniqBy(browsers, getBrowserKey)
```
lmiller1990, line 31: Oh neat, I forgot about `uniqBy`!

packages/data-context/src/sources/BrowserDataSource.ts
```diff
@@ -33,6 +42,41 @@ export interface BrowserApiShape {
 export class BrowserDataSource {
   constructor (private ctx: DataContext) {}
 
+  /**
+   * Gets the browsers from the machine and the project config
+   */
+  allBrowsers () {
```
lmiller1990: I think we can move this to use the `async/await` style - I know you are just matching the code style, but in general I think we are moving away from big nested `then` chains in favor of the more modern async/await. Thoughts?

packages/data-context/src/sources/BrowserDataSource.ts
```diff
@@ -33,6 +42,41 @@ export interface BrowserApiShape {
...
+    if (!this.ctx.coreData.allBrowsers) {
+      const p = this.ctx.project.getConfig()
+      const machineBrowsers = this.machineBrowsers()
+
+      this.ctx.coreData.allBrowsers = Promise.all([p, machineBrowsers]).then(async ([cfg, m]) => {
+        if (!cfg.browsers) return m
+
+        const userBrowsers = cfg.browsers.map<FoundBrowser | undefined>((b) => {
```
lmiller1990: Whenever I see a `map` then `filter` combo I like to opt for `reduce` to decrease the number of iterations. Unless you think that would be more difficult to read?

packages/data-context/src/sources/BrowserDataSource.ts
```diff
@@ -33,6 +42,41 @@ export interface BrowserApiShape {
...
+            majorVersion: String(b.majorVersion),
+            custom: true,
+          }
+        }).filter<FoundBrowser>((b): b is FoundBrowser => b != null) || []
+
+        return _.concat(m, userBrowsers)
+      }).catch((e) => {
+        this.ctx.update((coreData) => {
+          coreData.allBrowsers = null
+          coreData.diagnostics.error = e
```
lmiller1990: Nice job figuring out where/how to propagate this error, the data-context is pretty confusing. Did you actually verify this error manifests in the UI/CLI?

I'm also not entirely clear on the condition that would lead to this - is there one, or did you just put it here for good measure?

lmiller1990 (commented):
Looks good, just some nits. Also we don't appear to have any coverage around this - I'm still thinking about how best to add that. It might be quicker if someone internally picks that up, just so we can get this over the line.

Curious about the `catch` you've added - can you think of a specific example we might need to test for that where the logic you've added might error?

packages/data-context/src/sources/BrowserDataSource.ts
lmiller1990: Gonna need a quick test for both cases, I don't think we have much coverage around detecting the browsers in general, I could pick this up.

packages/data-context/src/sources/BrowserDataSource.ts
ryanjduffy: Fine with me. As you guessed, I was just matching style

packages/data-context/src/sources/BrowserDataSource.ts
ryanjduffy: Not a strong opinion. Some folks find reduce to be less readable. If there's a preference for reduce, I can make that change.

packages/data-context/src/sources/BrowserDataSource.ts
ryanjduffy: I don't think the new logic will likely error. `machineBrowsers()` re-throws so that'll be caught here but really you're just getting two copies of the same error then. Lemme experiment a bit to see if I can break it. ;)

ryanjduffy: Looks like the config validation logic would catch the cases in which this code would fail so I can remove the error handling.

ryanjduffy:
@lmiller1990 - Pushed a commit to address you feedback. Thanks!

lmiller1990 (approved):
Looks good, I wrote some tests https://github.com/cypress-io/cypress/pull/23446/commits/41c296d6f39129b0f7f4891f1fb7a0024e7dc817 then made a small refactor (code style - I really prefer return early than nesting lots of logic in an `if` statement, fairly confident everything still works since I wrote the tests before the refactor :+1: )

Not related to this PR but I *think* we should not be mutating coreData in `*DataSource`, but in `*Actions` (so in `BrowserActions.ts`). I think changing this is a big outside the scope of this, so going to approve this for now - code does what it's supposed to.

lmiller1990:
@emilyrohrbough WDYT? Any other comments?

lmiller1990:
Oops, CI is failing - lemme take a look.

https://github.com/cypress-io/cypress/pull/23446/commits/9ce44b4dcc7a33e960b024c8fa4c9ff08b36f973 should fix it.

lmiller1990:
Just need one more :heavy_check_mark:

ryanjduffy:
image

ZachJW34 (approved):
Works great! Let's get CI to run and then get this merged

lmiller1990:
I will get it merged

lmiller1990:
Hmm I think the remaining failures are not unique to this branch. Let me take a look.

lmiller1990:
I pushed a commit make the timeout for the npm/schematic test longer.

lmiller1990:
This is passing on CI https://app.circleci.com/pipelines/github/cypress-io/cypress/44027/workflows/3c5e7436-590d-4360-afcb-114d18b56c53

I don't know why the status did not update. I'm going to perform an admin merge, since I've visually inspected this is all passing on CI, and I've also run the tests locally that are related to this code change.

Merged... took a while, but we did it :+1:

PR #23494, opened 2022-08-22
fix(webpack-dev-server): add custom project config to handler

### LINKED ISSUE #23161: Allow Angular CT configuration to accept custom project configurations
### What would you like?

Angular CT setup should support custom build options to accommodate projects that don't follow the Angular CLI conventions.

### Why is this needed?

To support projects that use Angular but don't adhere to Angular CLI conventions (particularly apps created with [@nrwl/nx](https://github.com/nrwl/nx))

### Other

We developed Angular CT for projects utilizing `@angular/cli`. There are conventions with Angular CLI created projects that we rely on, such as sourcing a project's build options via their `angular.json`. Nx is a popular monorepo tooling solution that supports Angular projects but don't follow the same conventions. The build options are the same, but their source can differ (configurations can be stored in `workspace.json` or project level `project.json` configuration files).

We can extend support by allowing these build options to be passed in explicitly (in the `cypress.config.ts`). Support could look like:

```ts
export const defineConfig({
  component: {
    devServer: {
      framework: 'angular',
      bundler: 'webpack',
      // New Options object
      options: {
        projectConfig: { ... }
      },
    },
  },
})
```

We will probably have to normalize what the projectConfig contains. For Angular CLI projects, the type we expect is:

```ts
export type AngularJsonProjectConfig = {
  projectType: string
  root: string
  sourceRoot: string
  architect: {
    build: {
      options: { [key: string]: any } & { polyfills?: string }
      configurations?: {
        [configuration: string]: {
          [key: string]: any
        }
      }
    }
  }
}
```

We need the `root`, `sourceRoot`, `options` and `configuration['<key>']` information, but the json for Nx's `workspace.json/project.json` might differ so we can condense it to:

```ts
type ProjectConfig = {
  root: string
  sourceRoot: string
  buildOptions: Record<key, any>
}
```

where we assume the `buildOptions` conform to the Angular builder schema. Rough outline but should fit the bill.

We also assume a `tsconfig.json` that we extend from, but this might need to be configurable as well.

jordanpowell88:
- Closes #23161 

### User facing changelog
Adds support for passing a custom `projectConfig` to the webpack dev server.

### Additional details
The current Angular CT support only works for `angular CLI` projects. Frameworks like Nx are therefore not currently supported. This adds the option for Nx or other frameworks to pass in the project configuration details we need to generate the Angular Webpack config for the user. This would also extend the current behavior for users with multiple applications in the angular cli project could specify which application they want to target.

### PR Tasks

- [X] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?


npm/webpack-dev-server/src/devServer.ts
```diff
@@ -24,6 +24,7 @@ export type WebpackDevServerConfig = {
 } & {
   framework?: typeof ALL_FRAMEWORKS[number] // Add frameworks here as we implement
   webpackConfig?: unknown // Derived from the user's webpack
+  options?: ProjectConfig
```
ZachJW34, line 27: I'd like this to be:
```
options?: {
  projectConfig: ProjectConfig
}
```

That way we can encapsulate all of the project config options into one property. If we want to add another option in the future, the interfaces don't have to merge.

Also, we will need to work on types a bit since this type should only apply to projects that have specified `angular`.

npm/webpack-dev-server/src/helpers/angularHandler.ts
```diff
@@ -28,14 +30,53 @@ type AngularJson = {
   }
 }
 
+export type ProjectConfig = {
+  root: string
+  sourceRoot: string
+  buildOptions: BuildOptions
```
ZachJW34, line 36: If a user is explicitly supplying `buildOptions` we don't need to take into account the `configurations` since we will still have the problem of assuming `configurations.development` which might not exist. I think this option should only contain the browser builder options and we let the user determine what configuration they'd like to merge options in, something like:
```js
const { projects } = require('./angular.json')

const projectConfig = {
  root: projects.custom.root,
  sourceRoot: projects.custom.sourceRoot,
  buildOptions: {
    ...projects.custom.architect.build.options
    ...projects.custom.architect.build.configurations.test
  }
}

module.exports = { 
  component: {
    devServer: {
      framework: 'angular',
      bundler: 'webpack',
      options: {
        projectConfig
      }
    }
  }
}
```

Since this is an escape hatch, I like the idea of no longer relying on convention (like the expected shape of a proejct config target)

npm/webpack-dev-server/src/devServer.ts
jordanpowell88: Great idea! I'll add this in

Sathish787:
Thanks for implementing this. Would be great help for us as our application is split into multiple npm packages and angular.json configuration are quite different. I don't want to change in the actual angular.json file instead I wanted to use customized angular.json file for cypress unit tests.

When can we expect this change available?

jordanpowell88:
> Thanks for implementing this. Would be great help for us as our application is split into multiple npm packages and angular.json configuration are quite different. I don't want to change in the actual angular.json file instead I wanted to use customized angular.json file for cypress unit tests.
> 
> When can we expect this change available?

We are hoping this lands with the next release, as it is currently implemented in the PR it would allow you to pass in the `projectConfig` you want to use as part of the webpack devserver

cli/types/cypress.d.ts
```diff
@@ -3047,16 +3047,29 @@ declare namespace Cypress {
 
   type PickConfigOpt<T> = T extends keyof DefineDevServerConfig ? DefineDevServerConfig[T] : any
 
+  type ProjectConfig = {
```
ZachJW34, line 3050: Let's make this type more specific, like `AngularDevServerProjectConfig`, or you can inline the type and extract it in the dev-server. `ProjectConfig` sounds like it might apply to other Cypress types

npm/webpack-dev-server/src/helpers/angularHandler.ts
```diff
@@ -5,19 +5,22 @@ import { pathToFileURL } from 'url'
...
 
+type BuildOptions = Record<string, any>
+
+type Configurations = {
+  configurations?: {
+    [configuration: string]: BuildOptions
+  }
+}
+
+export type MyType = Extract<Cypress.DevServerConfigOptions, { framework: 'angular'}>
```
ZachJW34, line 16: Unused

system-tests/test/component_testing_spec.ts
```diff
@@ -123,6 +123,14 @@ describe(`Angular CLI major versions`, () => {
       expectedExitCode: 0,
     })
   }
+
+  systemTests.it('angular 14 custom config with mount tests', {
```
ZachJW34, line 127: Rename to `angular 14 custom config` since it isn't running the mount tests.

system-tests/projects/angular-custom-config/src/app/my-component.cy.ts
```diff
@@ -0,0 +1,8 @@
+import { MyComponent } from './my-component'
+
+describe('MyComponent', () => {
+  it('should mount with an h1 tag with a class of very-red', () => {
+    cy.mount(MyComponent)
+    cy.get('h1').should('have.css', 'color', 'rgb(255, 0, 0)')
```
ZachJW34, line 6:
```suggestion
    // Proves out the Angular specific `projectConfig` option. The global styles in `custom-styles.scss` would not normally be applied with how this project's `angular.json`
    // is configured. By utilizing the `projectConfig` option in `component.devServer.options`, we can change the
    // build options to allow this global style to be loaded.
    cy.get('h1').should('have.css', 'color', 'rgb(255, 0, 0)')
```

cli/types/cypress.d.ts
```diff
@@ -3047,16 +3047,29 @@ declare namespace Cypress {
...
   } | {
     bundler: 'vite'
     framework: 'react' | 'vue'
     viteConfig?: Omit<Exclude<PickConfigOpt<'viteConfig'>, undefined>, 'base' | 'root'>
+  } | {
+    bundler: 'webpack',
+    framework: 'angular',
+    webpackConfig?: PickConfigOpt<'webpackConfig'>,
+    options?: {
+      projectConfig: AngularDevServerProjectConfig
```
mike-plummer, line 3071: Should this be optional? Looks like it's declared that way in `devServer.ts`

npm/webpack-dev-server/src/helpers/angularHandler.ts
```diff
@@ -138,8 +122,8 @@ export async function generateTsConfig (devServerConfig: WebpackDevServerConfig,
     includePaths.push(toPosix(cypressConfig.supportFile))
   }
 
-  if (projectConfig.architect.build.options.polyfills) {
-    const polyfills = getProjectFilePath(projectConfig.architect.build.options.polyfills)
+  if (buildOptions?.polyfills) {
+    const polyfills = getProjectFilePath(buildOptions?.polyfills)
```
ZachJW34, line 126:
```suggestion
  if (buildOptions.polyfills) {
    const polyfills = getProjectFilePath(buildOptions.polyfills)
```

No need for the optional access here

ZachJW34 (approved):
Couple nits and @mike-plummer comment about the optional `projectConfig` property, works great!

cli/types/cypress.d.ts
jordanpowell88: @mike-plummer sorry, I thought you were referring to the change that @ZachJW34 proposed. This commit 133d6aa1c40b0505cff1eded103d11ccdc6b389a should have the change you requested. Great catch!

mike-plummer (approved):

astone123:
@jordanpowell88 are there any docs going in around this `options` config key? Might be worth adding a section about it [here?](https://docs.cypress.io/guides/references/configuration#devServer)

npm/webpack-dev-server/src/helpers/angularHandler.ts
```diff
@@ -30,21 +33,7 @@ type AngularJson = {
...
-async function getAngularCliWebpackConfig (devServerConfig: WebpackDevServerConfig) {
-  const { projectRoot } = devServerConfig.cypressConfig
-
-  const {
-    generateBrowserWebpackConfigFromContext,
-    getCommonConfig,
-    getStylesConfig,
-  } = await getAngularCliModules(projectRoot)
-
+export async function getProjectConfig (projectRoot: string): Promise<Cypress.AngularDevServerProjectConfig> {
```
astone123, line 36: Might it be possible to throw a config error in this function (if we can't resolve the config with the defaults) that has some info about configuring your Angular project using `devServer.options.projectConfig`?

jordanpowell88:
> @jordanpowell88 are there any docs going in around this `options` config key? Might be worth adding a section about it [here?](https://docs.cypress.io/guides/references/configuration#devServer)

Great idea. I will work on adding some documentation around that now

Issue #23337, opened 2022-08-15
Implement Studio UI


lmiller1990:
Part of the Cypress Studio Epic: https://github.com/cypress-io/cypress/issues/23230

AC:

- Ensure Studio UI renders correct in the new Cypress 10 app UI. 
- Use existing UI elements where possible (either from Cypress 9's original experimental Studio feature, or Cypress 10's new design system, whatever is easiest and looks best)
- Make it look decent/usable (need not be perfect on the first pass, since we want to collaborate with design on a unified look/feel. That's part of #23410).

Out of scope:

- Redesign the UI using our new design system

## Screenshots

You can see Studio in action by creating a new project with Cypress 9:

```
mkdir foo
cd foo
yarn init -y
yarn add cypress@9.7.0
echo '{"experimentalStudio": true}' > cypress.json
yarn cypress open
```

Make sure you do the above and you must do `experimentalStudio: true` in the `cypress.json`.

Next, create an empty spec, eg `it('works', () => {})`.

Click the little wand (hover the test description) to activate studio:

<img width="724" alt="Screen Shot 2022-08-17 at 4 10 42 pm" src="https://user-images.githubusercontent.com/19196536/185047306-1f33009e-2bc8-400b-a0ab-6d2876df0eb6.png">

More info here: https://docs.cypress.io/guides/references/cypress-studio#Using-Cypress-Studio

We will _eventually_ reskin it to look like the rest of the App UI. It is still experimental for now, so it's fine to just make it look good enough to ship - we will rework the UI before moving out of experimental.

Here's the specific UI elements we need (it should mostly exist in the code base, to find specifics you might need to check out Cypress 9. You can do it with `git checkout v9.7.0` and look around there).

<img width="1537" alt="Screen Shot 2022-08-17 at 4 00 43 pm" src="https://user-images.githubusercontent.com/19196536/185046460-e52e22a1-30d4-4e45-ba31-df6e99554903.png">

<img width="1537" alt="Screen Shot 2022-08-17 at 4 01 43 pm" src="https://user-images.githubusercontent.com/19196536/185046488-90f4b801-3dba-4b12-a7c9-345a791ec742.png">

<img width="1537" alt="Screen Shot 2022-08-17 at 4 02 21 pm" src="https://user-images.githubusercontent.com/19196536/185046501-19023e90-94a9-4d26-be85-614c3b9147df.png">

[...]

PR #23506, opened 2022-08-23
feat: add experimentalStudio flag back

### LINKED ISSUE #23338: Reintroduce `experimentalStudio` flag

lmiller1990:
- resolves #23338

**NOTE** targeting `feat/studio-cypress-10` feature branch, not develop.

- Add back `experimentalStudio` flag.
- Only works for E2E for now.
- You can set it globally, too - this way, anyone on 9.x waiting for Studio before updating to Cypress 10 can upgrade and continue w/ Studio as-in, without changing their config.
- CT throws an informative error.


cli/types/cypress.d.ts
```diff
@@ -2866,10 +2866,15 @@ declare namespace Cypress {
      */
     experimentalModifyObstructiveThirdPartyCode: boolean
     /**
-     * Generate and save commands directly to your test suite by interacting with your app as an end user would.
```
lmiller1990, line 2869: This description is clearly incorrect, drive-by fix.

mike-plummer (approved):

mike-plummer:
@lmiller1990 Not sure whether this would be in scope for this PR or is being tackled elsewhere - looks like there's missing i18n strings under the Project Settings view:

Screen Shot 2022-08-23 at 9 23 19 AM

astone123:
> @lmiller1990 Not sure whether this would be in scope for this PR or is being tackled elsewhere - looks like there's missing i18n strings under the Project Settings view:
> 
> Screen Shot 2022-08-23 at 9 23 19 AM

Nice catch, Mike 👍🏻 I just added them

lmiller1990:
Thanks for the catch @mike-plummer and thanks for adding that @astone123.

Good point on the studio icons for CT when `experimentalStudio` is root - I'll leave it to you @astone123 in https://github.com/cypress-io/cypress/issues/23337 as suggested.

Issue #6422, opened 2020-02-12
Safari (WebKit) Support


amirrustam:
> **Current Status** - _last updated August 4, 2022_
> 
> ✅ Prioritized and on the roadmap
> ✅ Doing exploratory work to assess requirements and timelines
> ✅ Under active development
> ℹ️ _Cypress currently supports [Firefox](https://cypress.io/blog/2020/02/06/introducing-firefox-and-edge-support-in-cypress-4-0/) and Chrome-family browsers (including Edge and Electron)_.

**Objective**
---
Add support for running Cypress tests within WebKit, which is the engine used by the [Safari](https://www.apple.com/safari/) web browser.

**Task List**
---

- [x] Initial PR to add development-only support: https://github.com/cypress-io/cypress/pull/15533
- [ ] Fix required driver/other functionality
	- [x] Network functionality https://github.com/cypress-io/cypress/pull/23232
	- [x] Keyboard tests (`type*.cy.js`) https://github.com/cypress-io/cypress/pull/23522
	- [x] Input tests (`click.cy.js`, focus tests...) https://github.com/cypress-io/cypress/pull/23522
	- [x] Stack traces https://github.com/cypress-io/cypress/pull/23730
	- [x] Check component testing
	- [x] Video recording #23579 
	- [x] Handle downloads https://github.com/cypress-io/cypress/pull/23662
	- [x] Fix global mode detection of npm module https://github.com/cypress-io/cypress/pull/23711
	- [ ] "Network" GraphQL error fix - wip
	- [x] [Fix open mode close](https://github.com/cypress-io/cypress/pull/15533#issuecomment-1211506734) https://github.com/cypress-io/cypress/pull/23662
	- [x] Disable `forceNetworkError` https://github.com/cypress-io/cypress/pull/23711
	- [x] Fix cross-origin unhandledRejection events: https://github.com/cypress-io/cypress/pull/23588
- [ ] Possibly fix less critical functionality
	- [ ] Multi-domain implementation https://github.com/cypress-io/cypress/pull/23442
	- [ ] Fix websocket polling/proxying/allowedSockets
	- [ ] [Fix devicePixelRatio](https://github.com/cypress-io/cypress/pull/15533#issuecomment-1210169765)
	- [x] Fix/skip focus from app
- [x] Documentation - limitations, Linux dependencies https://github.com/cypress-io/cypress-documentation/pull/4710
[...]

PR #23522, opened 2022-08-23
chore(webkit): driver updates for clicking/typing actions and related tests

tbiethman:
- Related: https://github.com/cypress-io/cypress/issues/6422

### User facing changelog

N/A

### Additional details

This PR re-enables driver tests related to clicking and typing for the WebKit browser. Included are some WebKit-specific tweaks to the driver for these actions. In-line doc is included where we deviate, please call out anything that is unclear or could otherwise be expounded on. 

### Steps to test

Review `driver-integration-tests-webkit` job results in CircleCI for the successful execution of the following specs:
* `type.cy.js`
* `type_errors.cy.js`
* `type_special_chars.cy.js`
* `click.cy.js`

### PR Tasks


packages/driver/cypress/e2e/commands/actions/click.cy.js
```diff
@@ -4064,12 +4062,28 @@ describe('mouse state', { browser: '!webkit' }, () => {
           y: 10,
         }
 
+        const coordsWebKit = {
```
tbiethman, line 4065: The coordinates vary per-browser due to differences in the default stylesheet.

packages/app/src/runner/logger.ts
```diff
@@ -118,43 +118,46 @@ export const logger = {
...
 
-  _logTable (consoleProps: any) {
-    if (isMultiEntryTable(consoleProps.table)) {
-      _.each(
-        _.sortBy(consoleProps.table, (val, key) => key),
-        (table) => {
-          return this._logTable({ table })
-        },
-      )
+  _logTables (consoleProps: any) {
```
tbiethman, line 121: We currently log tables for keyboard and mouse events, and we were creating slightly different data structures within the consoleProps for each. These differences made the existing logic complicated (and made parsing out the elements to avoid the WebKit hang difficult).

So I normalized the inputs and simplified the parsing here a bit, while ensuring that the elements don't make it into the table for WebKit. I'm sure we could simplify this further.

packages/driver/cypress/e2e/e2e/origin/commands/actions.cy.ts
```diff
@@ -455,14 +455,14 @@ context('cy.origin actions', () => {
...
+
+        expect(KeyboardEventsTable.data[1]).to.have.property('Details').that.equals('{ code: KeyO, which: 79 }')
+        expect(KeyboardEventsTable.data[1]).to.have.property('Typed').that.equals('o')
 
         expect(KeyboardEventsTable.data[2]).to.have.property('Details').that.equals('{ code: KeyO, which: 79 }')
         expect(KeyboardEventsTable.data[2]).to.have.property('Typed').that.equals('o')
-
-        expect(KeyboardEventsTable.data[3]).to.have.property('Details').that.equals('{ code: KeyO, which: 79 }')
-        expect(KeyboardEventsTable.data[3]).to.have.property('Typed').that.equals('o')
       })
```
tbiethman, line 466: This difference is a result of switching from a number-keyed object (starting with key 1) with an array (indexed at 0).

packages/driver/cypress/e2e/commands/actions/type_special_chars.cy.js
```diff
@@ -999,7 +1003,10 @@ describe('src/cy/commands/actions/type - #type special chars', { browser: '!webk
       cy.get('input[type="number"]:first')
       .invoke('val', '12.34')
       .type('{uparrow}{uparrow}')
-      .should('have.value', '14')
+      // WebKit does not round to the step value when calling stepUp/stepDown,
+      // as we do here for the ArrowUp handler.
+      // https://bugs.webkit.org/show_bug.cgi?id=244206
+      .should('have.value', Cypress.isBrowser('webkit') ? '14.34' : '14')
```
tbiethman, line 1009: The bug I logged is getting a little traction, but I'm not sure how quick these things get turned around. Do we think it's worth adding our own logic here for WebKit to bring it to parity, or just documenting the issue as a current limitation?

packages/driver/cypress/e2e/commands/actions/type_special_chars.cy.js
```diff
@@ -309,8 +314,8 @@ describe('src/cy/commands/actions/type - #type special chars', { browser: '!webk
       attachKeyListeners({ input })
 
       cy.get(':text:first').invoke('val', 'ab')
-      .then(($input) => $input[0].setSelectionRange(0, 0))
       .focus()
+      .then(($input) => $input[0].setSelectionRange(0, 0))
```
tbiethman, line 318: Setting the selection range prior to focusing the input like we did previously will fail, due to the default selection Webkit performs (and that we're working around [here](https://github.com/cypress-io/cypress/pull/23522/files#diff-0ee5140d59a5ee5cef7c7b5cfbe41509b1a08f9a4930d46b7e1746cbdac9ad9cR144)). 

The ordering here for these commands seemed pretty arbitrary given what is asserted, and it generally seems appropriate to focus an input prior to setting a selection within it, so I didn't have too many concerns with this change. But it is technically a difference in behavior.

packages/app/src/runner/logger.ts
```diff
@@ -118,43 +118,46 @@ export const logger = {
...
+    const logTable = ({ name, data, columns }) => {
+      let tableData = data
 
-      return
-    }
+      if (Cypress.isBrowser('webkit')) {
+        // WebKit will hang when we attempt to log element references
```
lmiller1990, line 126: Woah, major bug in webkit.

packages/app/src/runner/logger.ts
```diff
@@ -118,43 +118,46 @@ export const logger = {
...
+        // within a table. We replace the element with a simplified display
+        // string in this case.
+        // https://bugs.webkit.org/show_bug.cgi?id=244100
 
-    const table = this._getTable(consoleProps)
+        const getSimplifiedElementDisplay = (element: Element) => {
```
lmiller1990, line 131: This seems similar to https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/dom/elements/utils.ts#L45 - or are they different enough not to warrant reuse?

packages/driver/cypress/e2e/commands/actions/type_special_chars.cy.js
lmiller1990: Option 3 ... Fix WebKt ourselves! But we'd still need something in the (probably long) interim. 

If this is a WebKit bug, what you've done make sense to me - if a user runs a test expecting 14, it *should* fail - since it will fail in production for real users, too.

packages/driver/cypress/e2e/e2e/origin/commands/actions.cy.ts
lmiller1990: Where is the source code you changed that led to the need to change this test? I don't see it.

lmiller1990 (approved):
No major blocker, just some questions.

packages/driver/cypress/e2e/e2e/origin/commands/actions.cy.ts
tbiethman: The directly relevant change is here: https://github.com/cypress-io/cypress/pull/23522/files#diff-c0edb6eb8f7d6c9276eae18e47f83859d84a34dafaf05ec2a97b6fd5478983a5R91

packages/driver/cypress/e2e/commands/actions/type_special_chars.cy.js
tbiethman: > if a user runs a test expecting 14, it should fail - since it will fail in production for real users, too

The behavior we're simulating here is a user tapping the up/down arrows inside the input, which _does_ work appropriately in WebKit. The problem is that the `stepUp`/`stepDown` functions we call to simulate those presses don't work the same way. So this would definitely need to be a documented limitation if we left things as-is.

I did write up the algorithm to handle the step logic ourselves. It's not too intensive, but it'll need a new suite of tests associated to it. I want to make sure we have other critical issues wrapped up before spending much more time on it (unless we consider this critical?).

packages/app/src/runner/logger.ts
tbiethman: They are pretty close. I wrote the current version to match what Chrome renders in its tables, the util would produce slightly different results. But we could reuse if there are strong feelings either way.

packages/driver/cypress/e2e/commands/actions/type.cy.js
```diff
@@ -1794,33 +1814,36 @@ describe('src/cy/commands/actions/type - #type', { browser: '!webkit' }, () => {
     })
 
     // https://github.com/cypress-io/cypress/issues/7088
+    // In WebKit, setting the inputType is not supported by the InputEvent constructor.
+    // This results in the inputType being unset in any of the simulated beforeinput events.
+    // https://bugs.webkit.org/show_bug.cgi?id=170416
```
flotwig, line 1819: Should this limitation be updated to also mention the missing `inputType` prop? https://app.clickup.com/t/18033298/PM-1796?block=block-3bf03878-604a-44e1-84bb-e5d26fc7f749

packages/driver/cypress/e2e/commands/actions/type_special_chars.cy.js
flotwig: The cross-section of users who use `.type('{uparrow}')` on a number input who will be using experimental WebKit is probably pretty small, I'd agree that focus should be put elsewhere. How large is the change to fix this?

packages/driver/cypress/e2e/commands/actions/type_special_chars.cy.js
```diff
@@ -1736,7 +1745,8 @@ describe('src/cy/commands/actions/type - #type special chars', { browser: '!webk
         })
       })
 
-      it('will not submit the form', function (done) {
+      // WebKit still emits the submit event on the form in this configuration.
```
flotwig, line 1748: Is this just how WebKit works? It acts the same if a user presses "Enter" on a form with multiple submits?

packages/driver/src/dom/selection.ts
```diff
@@ -575,19 +575,25 @@ const _moveSelectionTo = function (toStart: boolean, el: HTMLElement, options =
...
-      // Sometimes, selection.rangeCount is 0 when there is no selection.
-      // In that case, it fails in Chrome.
-      // We're creating a new range and add it to the selection to avoid the case.
-      if (selection.rangeCount === 0) {
-        range = doc.createRange()
-        selection.addRange(range)
+      // WebKit must use selectAllChildren to ensure selections are made in all use cases.
+      // Some content, like HTML markup, cannot be selected using a range when inside a contenteditable field.
+      if (Cypress.isBrowser('webkit')) {
+        selection.selectAllChildren(el)
```
flotwig, line 581: Interesting, I wonder why we aren't just using `selectAllChildren` in all browsers, did you check if this works for Cr/FF?

flotwig (approved):
Nice! I appreciate the thorough comments, these will help a lot down the road. Had a few Q's but LGTM.

packages/driver/cypress/e2e/commands/actions/type_special_chars.cy.js
tbiethman: The algorithm isn't very complicated, there are just a lot of bound checks/edge cases that I'd want to write thorough tests for. I'd hate to write a custom handler that introduces yet more bugs :bug:

I think we could wrap it up prior to the initial release if we start looking good on the other issues.

packages/driver/cypress/e2e/commands/actions/type_special_chars.cy.js
tbiethman: Correct, I recreated this scenario outside of Cypress and found the form was still submitted in this case in WebKit, whereas its not in Chrome/FF.

packages/driver/src/dom/selection.ts
tbiethman: Good call, it does work for the current versions of Chrome/Firefox and Chromium 74, which is the earliest I can grab. Minimum supported Chrome is 64, so there's still a range of supported versions I can't validate.

flotwig: MDN says it's supported since uh.. Chrome... *squints* Chrome 1. https://developer.mozilla.org/en-US/docs/Web/API/Selection/selectAllChildren#browser_compatibility So it should be safe if the tests pass?

tbiethman: Wow, straight from the primordial soup. Well we know it won't throw at least! I'll get that change made and verified in CI.

packages/driver/cypress/e2e/e2e/origin/commands/actions.cy.ts
lmiller1990: I didn't even know we could log a table like this, neat.

packages/driver/cypress/e2e/commands/actions/type_special_chars.cy.js
lmiller1990: Sounds good, I say the sooner we ship this, the better - little fixes like this will come up organically, if we don;'t get to them (and we might have other edges cases we don't even know about yet).

packages/app/src/runner/logger.ts
lmiller1990: Up to you, I do not have strong feelings about this 💯

PR #23523, opened 2022-08-23
chore: update typescript

ZachJW34:
### User facing changelog
n/a

### Additional details
Update to latest version of typescript. Had to bump `ts-loader` as `@packages/extension` was failing with the new version of TS.

### Steps to test
Counting on CI here, everything worked well locally.

na

### PR Tasks

### DIFF
--- a/packages/driver/src/dom/elements/nativeProps.ts
+++ b/packages/driver/src/dom/elements/nativeProps.ts
@@ -238,7 +238,7 @@ export const getNativeProp = function<T, K extends keyof T> (obj: T, prop: K): T
   if (!nativeProp) {
     const props = _.keys(nativeGetters).join(', ')
 
-    throw new Error(`attempted to use a native getter prop called: ${prop as string}. Available props are: ${props}`)
+    throw new Error(`attempted to use a native getter prop called: ${String(prop)}. Available props are: ${props}`)
   }
 
   let retProp = nativeProp.call(obj, prop)
--- a/packages/driver/src/dom/elements/nativeProps.ts
+++ b/packages/driver/src/dom/elements/nativeProps.ts
@@ -258,7 +258,7 @@ export const setNativeProp = function<T, K extends keyof T> (obj: T, prop: K, va
   if (!nativeProp) {
     const fns = _.keys(nativeSetters).join(', ')
 
-    throw new Error(`attempted to use a native setter prop called: ${prop as string}. Available props are: ${fns}`)
+    throw new Error(`attempted to use a native setter prop called: ${String(prop)}. Available props are: ${fns}`)
   }
 
   let retProp = nativeProp.call(obj, val)
--- a/packages/driver/src/dom/elements/nativeProps.ts
+++ b/packages/driver/src/dom/elements/nativeProps.ts
@@ -12,7 +12,7 @@ const descriptor = <T extends keyof Window, K extends keyof Window[T]['prototype
   const desc = Object.getOwnPropertyDescriptor(window[klass].prototype, prop)
 
   if (desc === undefined) {
-    throw new Error(`Error, could not get property descriptor for ${klass}  ${prop}. This should never happen`)
+    throw new Error(`Error, could not get property descriptor for ${klass}  ${prop as string}. This should never happen`)
   }
 
   return desc
@@ -238,7 +238,7 @@ export const getNativeProp = function<T, K extends keyof T> (obj: T, prop: K): T
   if (!nativeProp) {
     const props = _.keys(nativeGetters).join(', ')
 
-    throw new Error(`attempted to use a native getter prop called: ${prop}. Available props are: ${props}`)
+    throw new Error(`attempted to use a native getter prop called: ${prop as string}. Available props are: ${props}`)
   }
 
   let retProp = nativeProp.call(obj, prop)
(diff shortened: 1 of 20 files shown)


packages/scaffold-config/src/frameworks.ts
```diff
@@ -61,11 +61,9 @@ function getBundlerDependency (bundler: WizardBundler['type'], projectPath: stri
   }
 }
 
-export const WIZARD_MOUNT_MODULES = ['cypress/react', 'cypress/react18', 'cypress/vue', 'cypress/vue2', 'cypress/angular'] as const
+export type WizardMountModule = Awaited<ReturnType<typeof WIZARD_FRAMEWORKS[number]['mountModule']>>
```
ZachJW34, line 64: Example using the new `Awaited` utility type.

packages/driver/src/dom/elements/nativeProps.ts
```diff
@@ -12,7 +12,7 @@ const descriptor = <T extends keyof Window, K extends keyof Window[T]['prototype
   const desc = Object.getOwnPropertyDescriptor(window[klass].prototype, prop)
 
   if (desc === undefined) {
-    throw new Error(`Error, could not get property descriptor for ${klass}  ${prop}. This should never happen`)
+    throw new Error(`Error, could not get property descriptor for ${klass}  ${prop as string}. This should never happen`)
```
lmiller1990, line 15:
```suggestion
    throw new Error(`Error, could not get property descriptor for ${klass}  ${String(prop)}. This should never happen`)
```

packages/driver/src/dom/elements/nativeProps.ts
```diff
@@ -238,7 +238,7 @@ export const getNativeProp = function<T, K extends keyof T> (obj: T, prop: K): T
   if (!nativeProp) {
     const props = _.keys(nativeGetters).join(', ')
 
-    throw new Error(`attempted to use a native getter prop called: ${prop}. Available props are: ${props}`)
+    throw new Error(`attempted to use a native getter prop called: ${prop as string}. Available props are: ${props}`)
```
lmiller1990, line 241:
```suggestion
    throw new Error(`attempted to use a native getter prop called: ${String(prop)}. Available props are: ${props}`)
```

packages/driver/src/dom/elements/nativeProps.ts
```diff
@@ -258,7 +258,7 @@ export const setNativeProp = function<T, K extends keyof T> (obj: T, prop: K, va
   if (!nativeProp) {
     const fns = _.keys(nativeSetters).join(', ')
 
-    throw new Error(`attempted to use a native setter prop called: ${prop}. Available props are: ${fns}`)
+    throw new Error(`attempted to use a native setter prop called: ${prop as string}. Available props are: ${fns}`)
```
lmiller1990, line 261:
```suggestion
    throw new Error(`attempted to use a native setter prop called: ${String(prop)}. Available props are: ${fns}`)
```

packages/scaffold-config/src/frameworks.ts
lmiller1990: 🎉

lmiller1990: For fun, if you do "go to definition" on `Awaited` you can see how they built it:

```ts
type Awaited<T> =
    T extends null | undefined ? T : // special case for `null | undefined` when not in `--strictNullChecks` mode
        T extends object & { then(onfulfilled: infer F): any } ? // `await` only unwraps object types with a callable `then`. Non-object types are not unwrapped
            F extends ((value: infer V, ...args: any) => any) ? // if the argument to `then` is callable, extracts the first argument
                Awaited<V> : // recursively unwrap the value
                never : // the argument to `then` was not callable
        T; // non-object or non-thenable
```

lmiller1990 (approved):
Nice, couple of minor improvements, won't block though.

packages/scaffold-config/src/frameworks.ts
ZachJW34: 🤯 would take me a whole day to try and cook up something like that from scratch

emilyrohrbough:
Should we go ahead and update the typescript versions in our system tests?

ZachJW34:
@emilyrohrbough I wouldn't want to update the system-tests since they were created to represent a project in a given state. For example, `angular-13` has `"typescript": "~4.6.2"` and might not work with a bump. There are probably a few I could update but not sure if there is any improvement in doing so.

emilyrohrbough:
Gottchya 👍🏻  is there a strategy on knowing when it's appropriate to bumping these system-test dependencies for component testings? or not really?
I had assumed they were the version used when added and we just haven't necessarily been bumping deps due to other work

emilyrohrbough (approved):

ZachJW34:
We don't have a strategy, but for component testing I'd say it would make sense to simply remove system tests once we drop support for that version, so if we drop `angular@13` support or `nuxt@2` then they should be deleted.

packages/scaffold-config/src/frameworks.ts
lmiller1990: It's quite cool, notice they use `Awaited` in the definition? Recursive types!

This is actually a great type to study... you can learn

- extends
- conditionals
- infer
- recursive types
- never (equivalent of `throw Error('this is impossible')` for types

PR #23535, opened 2022-08-24
fix: filtering with `--spec` option for projects in Docker container root

### LINKED ISSUE #23380: Using `--spec` option results in "no specs found" when project is located at Docker root directory
### Current behavior

image

All that changed is the version number in my dockerfile

If I downgrade from there to 10.4.0 I have no issues.

### Desired behavior

Tests should run with no errors

### Test code to reproduce

Build cypress/included:10.5.0

Try doing a `cypress run`

### Cypress Version

10.5.0

### Node version

16.16.0

### Operating System

Linux

### Debug Logs

```shell
Your supportFile is missing or invalid: support/e2e.ts

The supportFile must be a .js, .ts, .coffee file or be supported by your preprocessor plugin (if configured).

Fix your support file, or set supportFile to false if a support file is not necessary for your project.

If you have just renamed the extension of your supportFile, restart Cypress.

https://on.cypress.io/support-file-missing-or-invalid
```

### Other

_No response_

tbiethman:
- Closes #23380

### User facing changelog

* Fixed an issue where filtering with the `--spec` run mode option would not find any specs if the project was located at the root directory of a file system.

### Additional details

This logic has been changed a [few](https://github.com/cypress-io/cypress/pull/23296) [times](https://github.com/cypress-io/cypress/pull/23368) lately to ensure that we are only globbing user-provided patterns. These changes introduced a regression where, with a working directory of root (`/`), the glob pattern would be incorrectly manipulated:

working dir of `/` + glob pattern of `/foo/bar/*.cy.js` => `.foo/bar/*.cy.js` // this won't match anything

This typically only occurred with our docker images, where users may mount project content directly to the container's root. We correct this issue by having better checks around our absolute glob pattern detection and deriving implicit relative patterns (`foo/bar/*.cy.js` rather than `./foo/bar/*.cy.js`) for the glob patterns that we must manipulate.

### Steps to test

The added unit tests now show the root working directory scenarios being properly handled. 

I also created a validation repo here: https://github.com/tbiethman/verify-23535. Steps to validate are included on its README.

### PR Tasks


packages/data-context/src/sources/FileDataSource.ts
```diff
@@ -9,6 +9,10 @@ import { toPosix } from '../util/file'
 
 const debug = Debug('cypress:data-context:sources:FileDataSource')
 
+export const matchGlobs = async (globs: string | string[], globbyOptions?: GlobbyOptions) => {
+  return await globby(globs, globbyOptions)
+}
```
tbiethman, line 14: Our pinned version of globby exports the [function itself](https://github.com/sindresorhus/globby/blob/v11.1.0/index.js#L114) as the module export, making it difficult to stub calls to it for unit testing purposes. 

So I added a little wrapper here so I could stub these calls appropriately, but I can remove this if I'm missing an obvious way to stub the module directly.

packages/data-context/test/unit/sources/FileDataSource.spec.ts
```diff
@@ -5,127 +5,273 @@ import path from 'path'
...
+
+      it('matches absolute patterns when working directory is root', async () => {
+        const files = await fileDataSource.getFilesByGlob(
+          '/',
+          '/cypress/e2e/**.cy.js',
+        )
+
+        expect(files).to.eq(mockMatches)
+        expect(matchGlobsStub).to.have.been.calledWith(
+          ['cypress/e2e/**.cy.js'],
```
tbiethman, line 182: In 10.6.0, this test case would produce `.cypress/e2e/**.cy.js`, which is not a valid pattern.

packages/data-context/src/sources/FileDataSource.ts
```diff
@@ -32,12 +36,14 @@ export class FileDataSource {
...
       // The working directory path may include characters that conflict with glob
       // syntax (brackets, parentheses, etc.) and cause our searches to inadvertently fail.
       // We scope our search to the working directory using the `cwd` globby option.
-      if (globPattern.startsWith(cwd)) {
-        return globPattern.replace(cwd, '.')
+      if (globPattern.startsWith(workingDirectoryPrefix)) {
+        return globPattern.replace(workingDirectoryPrefix, '')
       }
 
       return globPattern
```
flotwig, line 49: Would `return path.relative(workingDirectoryPrefix, globPattern)` also work? Might not even need line 39 at that point

flotwig (approved):
LGTM, thanks for adding the unit tests here!

packages/data-context/src/sources/FileDataSource.ts
tbiethman: I think `path.relative` has the potential to mutate the glob patterns to be relative to the current execution directory, which may or may not equal the given `cwd` argument. In those cases, the pattern would no longer be representative of what the user expects.

packages/data-context/src/sources/FileDataSource.ts
```diff
@@ -9,6 +9,10 @@ import { toPosix } from '../util/file'
 
 const debug = Debug('cypress:data-context:sources:FileDataSource')
 
+export const matchGlobs = async (globs: string | string[], globbyOptions?: GlobbyOptions) => {
```
BlueWinds, line 12: An async function with `return await` is equivalent to a non-async function that just returns the promise.

BlueWinds (approved):

PR #23536, opened 2022-08-24
chore(server): convert modes/run to typescript

flotwig:
- Closes 

### User facing changelog

n/a

### Additional details

- Was working on refactoring the video recording code for #6422 and was finally driven insane by the number of floating untyped variables in `modes/run`.
- I added types with minimal refactoring - it's still a mess, but it's a typed mess.
- *GitHub's built in diff tool fails to recognize the rename of `run.js` -> `run.ts` if you view all changes together.* To make it easier to review, here are the distinct commits and what they do:
  - 11ecebd238b4f8ca65bd787e0c7db2d681314d66 adds types and fixes type errors
  - 98e69d0d1ce3c96760e7bf8b9606cb031418dffd removes a unit-level `run` spec that has not been running for a while now, and has been superseded by e2e/integration tests
  - [d06d529](https://github.com/cypress-io/cypress/pull/23536/commits/d06d529904aef8e85d3494bd3858900cf69eaa6d) removes all `run` exports except `run()` itself. moving everything off of `module.exports` and `this` caused new type errors since more type checking took place, this commit fixes those errors.

### PR Tasks


packages/server/test/integration/cypress_spec.js
```diff
@@ -338,8 +339,11 @@ describe('lib/cypress', () => {
       await clearCtx()
 
       sinon.stub(electron.app, 'on').withArgs('ready').yieldsAsync()
-      sinon.stub(runMode, 'waitForSocketConnection').resolves()
-      sinon.stub(runMode, 'listenForProjectEnd').resolves({ stats: { failures: 0 } })
+      globalThis.CY_TEST_MOCK = {
```
flotwig, line 342: So, I introduced this pattern because we have a lot of integration test coverage here that does deep mocking of (what was previously on) `module.exports`. I know this is an unfortunate way to handle mocking, but I felt it was the lesser of evils... the other options being to export everything to make it stubbable and call methods off `module.exports`, or some kinda proxyquire hackery. Open to better ideas here.

packages/server/test/integration/cypress_spec.js
```diff
@@ -1159,43 +1163,28 @@ describe('lib/cypress', () => {
 
       sinon.stub(electron.app, 'on').withArgs('ready').yieldsAsync()
       sinon.stub(browsers, 'open')
-      sinon.stub(runMode, 'waitForSocketConnection').resolves()
-
-      sinon.stub(runMode, 'waitForBrowserToConnect').resolves({
-        stats: {
```
flotwig, line 1165: this object stub to have been a typo, which is why it was removed

packages/server/lib/modes/run.ts
```diff
@@ -0,0 +1,1661 @@
...
+    testId: data.testId,
+    testAttemptIndex: data.testAttemptIndex,
+    takenAt: resp.takenAt,
+    path: resp.path,
+    height: resp.dimensions.height,
+    width: resp.dimensions.width,
+  }
+}
+
+async function runSpecs (options: { config: Cfg, browser: Browser, sys: any, headed: boolean, outputPath: string, specs: SpecWithRelativeRoot[], specPattern: string | RegExp | string[], beforeSpecRun?: BeforeSpecRun, afterSpecRun?: AfterSpecRun, runUrl?: string, parallel?: boolean, group?: string, tag?: string, testingType: TestingType, quiet: boolean, project: Project, onError: (err: Error) => void, exit: boolean, socketId: string, webSecurity: boolean, projectRoot: string } & Pick<Cfg, 'video' | 'videoCompression' | 'videosFolder' | 'videoUploadOnPasses'>) {
```
lmiller1990, line 1290: Ahhh yes awesome, we need this so desperately. Every time I tried to make this into TS, I realized how much work it was and abandoned it - definitely time well spent to figure out what these types are.

flotwig: Yeah, there's still about 4 billion de-structured variables too many in this file, but at least this will enable us to more confidently refactor in the future.

packages/server/test/integration/cypress_spec.js
lmiller1990: I really question the value of some of these tests, where we mock 90% of the world to test some random method... I almost wonder if we collapse ~100 unit tests into ~5 system tests and just aggressively delete some of these. I think a lot of these are from the days prior to `system-tests` existing? 

If we have really complex unit level logic, stick it in a pure function and unit test that, so we don't need so much mocking/stubbing.

packages/types/src/config.ts
```diff
@@ -30,7 +30,7 @@ export interface FullConfig extends Partial<Cypress.RuntimeConfigOptions & Cypre
...
-  & Pick<Cypress.ResolvedConfigOptions, 'chromeWebSecurity' | 'supportFolder' | 'experimentalSourceRewriting' | 'fixturesFolder' | 'reporter' | 'reporterOptions' | 'screenshotsFolder' | 'supportFile' | 'baseUrl' | 'viewportHeight' | 'viewportWidth' | 'port' | 'experimentalInteractiveRunEvents' | 'userAgent' | 'downloadsFolder' | 'env' | 'excludeSpecPattern' | 'specPattern' | 'experimentalSessionAndOrigin' | 'experimentalModifyObstructiveThirdPartyCode'> // TODO: Figure out how to type this better.
+  & Pick<Cypress.ResolvedConfigOptions, 'chromeWebSecurity' | 'supportFolder' | 'experimentalSourceRewriting' | 'fixturesFolder' | 'reporter' | 'reporterOptions' | 'screenshotsFolder' | 'supportFile' | 'baseUrl' | 'viewportHeight' | 'viewportWidth' | 'port' | 'experimentalInteractiveRunEvents' | 'userAgent' | 'downloadsFolder' | 'env' | 'excludeSpecPattern' | 'specPattern' | 'experimentalSessionAndOrigin' | 'experimentalModifyObstructiveThirdPartyCode' | 'video' | 'videoCompression' | 'videosFolder' | 'videoUploadOnPasses' | 'resolvedNodeVersion' | 'resolvedNodePath' | 'trashAssetsBeforeRuns'> // TODO: Figure out how to type this better.
```
lmiller1990, line 33: One day this will be `Pick<all the things>`. I wonder if this is useful? I think I wrote this originally, when it had like 2-3 properties in `Pick`... it's growing with every PR! 

I think the original intent was "These are the options we use in this part of the code" but we might have outgrown it.

flotwig: Yeah, I was questioning this also. It doesn't seem like we would have only some ResolvedConfigOptions and not others at any point.

lmiller1990 (commented):
Thanks for this - I know it's still draft but going to follow along, this will have very high long term value - working in `run.js` is kind of like stumbling around in the dark at the moment :|

packages/server/test/integration/cypress_spec.js
flotwig: Agreed. I didn't want to have to dive into figuring out what edge cases are/aren't covered in these integration tests versus other tests for this PR, but we do have way too many tests like this.

tbiethman (approved):
LGTM. I'll second the thanks for this, it will be a big help going forward.

packages/server/lib/modes/run.ts
```diff
@@ -0,0 +1,1661 @@
...
+import type { Cfg } from '../project-base'
+import type { Browser } from '../browsers/types'
+
+type Screenshot = {
+  width: number
+  height: number
+  path: string
+  specName: string
+}
+type SetScreenshotMetadata = (data: Screenshot) => void // ?
```
emilyrohrbough, line 42:
```suggestion
type SetScreenshotMetadata = (data: Screenshot) => void
```

packages/server/lib/modes/run.ts
```diff
@@ -0,0 +1,1661 @@
...
+    colorIf(results.totalPending, 'cyan'),
+    colorIf(results.totalSkipped, 'blue'),
+  ]
+}
+
+const formatSymbolSummary = (failures) => {
+  return getSymbol(failures)
+}
+
+const macOSRemovePrivate = (str) => {
```
emilyrohrbough, line 158:
```suggestion
const macOSRemovePrivate = (str: string): string => {
```

packages/server/lib/modes/run.ts
```diff
@@ -0,0 +1,1661 @@
...
+const colorIf = function (val, c) {
+  if (val === 0 || val == null) {
+    val = '-'
+    c = 'gray'
+  }
+
+  return color(val, c)
+}
+
+const getSymbol = function (num) {
```
emilyrohrbough, line 79:
```suggestion
const getSymbol = function (num?: number) {
```

packages/types/src/server.ts
```diff
@@ -46,7 +47,7 @@ export interface AutomationMiddleware {
   onBeforeRequest?: OnRequestEvent | null
   onRequest?: OnRequestEvent | null
   onResponse?: NullableMiddlewareHook
-  onAfterResponse?: NullableMiddlewareHook
+  onAfterResponse?: ((eventName: string, data: any, resp: any) => void) | null
```
emilyrohrbough, line 50: I'd be nice if we someday remove the null states and leveraged undefined

packages/server/lib/modes/run.ts
```diff
@@ -0,0 +1,1661 @@
...
+}
+
+export async function run (options, loading: Promise<void>) {
+  if (require('../util/electron-app').isRunningAsElectronProcess({ debug })) {
+    const app = require('electron').app
+
+    // electron >= 5.0.0 will exit the app if all browserwindows are closed,
+    // this is obviously undesirable in run mode
+    // https://github.com/cypress-io/cypress/pull/4720#issuecomment-514316695
+    app.on('window-all-closed', () => {
```
emilyrohrbough, line 1648: I wonder if this is still true

emilyrohrbough (approved):

packages/server/lib/modes/run.ts
flotwig: Yeah, it should be. This was an intentional change, since most Electron apps should have 0 processes left when all the windows are closed. We're the exception to their rule

tbiethman: Here's the underlying Electron code that handles this, for reference: https://github.com/electron/electron/blob/d44a187d0b226800fe0cb4f7a0d2b36c871b27cd/lib/browser/init.ts#L172

PR #23537, opened 2022-08-24
chore: Add Studio UI to Cypress 10

### LINKED ISSUE #23337: Implement Studio UI

astone123:
- Closes #23337

### User facing changelog

This PR adds the Cypress studio UI back into the app with some styling changes. Note that this change will be merged into the `feat/studio-cypress-10` branch to be reviewed all together and merged at a later time.

### Steps to test

1. Add `experimentalStudio: true` your project's Cypress configuration under `e2e`
2. Launch e2e testing for with your project
3. Run a spec
4. Hover over a test in the runner command log and click the magic wand icon
5. Click through the Studio UI and ensure that it is consistent with [the mockup](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/gMEcNAtUnMXhR4tyBLOBVx/Cypress-App?node-id=16373%3A174947)

Screen Shot 2022-08-24 at 4 37 57 PM

Screen Shot 2022-08-24 at 4 38 26 PM

Screen Shot 2022-08-24 at 4 39 35 PM

Screen Shot 2022-08-24 at 4 38 56 PM

### PR Tasks
Deleting the PR checklist since we'll be handling more testing, documentation, etc when we merge the feature branch.


packages/app/src/runner/studio/StudioUrlPrompt.vue
```diff
@@ -0,0 +1,83 @@
...
+
+onMounted(() => {
+  if (props.autUrlInputRef) {
+    props.autUrlInputRef.focus()
+  }
+})
+
+</script>
+
+<style lang="scss">
```
lmiller1990, line 59: We should use one of the coolest Vue features, scoped styling.

https://vue-loader.vuejs.org/guide/scoped-css.html

Just add `<style lang="scss" scoped>` and the CSS will only be applied inside this component.

packages/app/src/runner/studio/StudioUrlPrompt.vue
```diff
@@ -0,0 +1,83 @@
...
+      content: " ";
+      position: absolute;
+      bottom: 100%;
+      left: 50%;
+      margin-left: -9px;
+      border-width: 9px;
+      border-style: solid;
+      border-color: transparent transparent #1B1E2E transparent;
+    }
+  }
```
lmiller1990, line 82: Assuming this is too specific to do with Tailwind?

packages/app/src/store/studio-store.ts
```diff
@@ -206,6 +212,8 @@ export const useStudioStore = defineStore('studioRecorder', {
         }
       }
 
+      this.setUrl(getAutUrl())
```
lmiller1990, line 215: No need to do `getAutUrl` like this, we should be able to do

```js
this.setUrl(autStore.url)
```

Although I think there's a bug around this anyway, as Mike pointed out in Slack, there's some weirdness around the URL logic.

packages/frontend-shared/src/components/StandardModal.vue
```diff
@@ -77,7 +77,7 @@ const props = withDefaults(defineProps<{
 }>(), {
   modelValue: false,
   helpText: `${defaultMessages.links.needHelp}`,
-  helpLink: 'https://on.cypress.io',
+  helpLink: '',
```
lmiller1990, line 80: Any reason to make this default value `''`?

lmiller1990 (commented):
Some misc comments, nothing big -- I think the URL logic still has some bugs, I'm going to work on those in the meantime. New styles look 💯

packages/reporter/src/test/test.tsx
```diff
@@ -193,7 +193,7 @@ class Test extends Component<TestProps> {
       )
     }
 
-    if (this.props.studioEnabled) {
+    if (this.props.studioEnabled && !appState.studioActive) {
```
lmiller1990, line 196: Seems this is actually breaking some tests in reporter - do we need this check at all?

lmiller1990:
This modal seems kind of ugly to me:

image

1) do we need the wand in the title? Could be just me but I don't like icons that much unless they add value or meaning, this just seems like it's a decoration for no reason - not sure if this was a designer recommendation or not, no strong opinions
2) the black circle around the :white_check_mark: looks kind of weird to me - again I wonder if we need the icon at all? :thinking:

lmiller1990:
@astone123 I need these changes to work on the E2E tests, and it looks like we've got a few failures. I'm going to go ahead and get CI green and merge this -- we will solicit a full review by EOD tomorrow, I'm thinking since the PR is pretty big, we get specific people to do specific things - eg, someone from CT would look at the UI components, someone from E2E the tests/wiring, etc.

packages/frontend-shared/src/components/StandardModal.vue
lmiller1990: Right I see - to hide the link for modals where a "Help Link" doesn't make sense.
Quick refactor to support this and to get CI green https://github.com/cypress-io/cypress/pull/23545

packages/app/src/runner/studio/StudioUrlPrompt.vue
astone123: yeah, I figured this was better than adding it to the already cluttered `class` attribute on the outer container. Definitely open to suggestions for how to better deal with this though

packages/frontend-shared/src/components/StandardModal.vue
astone123: Yeah, this just makes it so that if you don't pass a link then we don't show the link section in the modal header

PR #23542, opened 2022-08-24
chore(server): remove dead runner code 

emilyrohrbough:
chore: remove dead code -- lives in [data-context/src/sources/HtmlDataSource.ts](https://github.com/cypress-io/cypress/blob/develop/packages/data-context/src/sources/HtmlDataSource.ts#L64)


packages/server/lib/controllers/runner.ts
```diff
@@ -1,101 +1,11 @@
...
-  return res.render(runnerPath, {
-    base64Config,
-    projectName: config.projectName,
-    namespace: config.namespace,
-  })
-}
+import { getPathToDist } from '@packages/resolve-dist'
 
 export const runner = {
-  serve (req, res, runnerPkg: RunnerPkg, options: ServeOptions) {
```
lmiller1990, line 50: Did not realize this was unused... I wonder if this *should* live in data-context? Server seems like a better place for, well, serving things. The line between server/data-context is really blurred, though. I think the goal was incrementally moving things so we could strongly type them, but a lot of things ended up just getting dumped there (probably since it has more strict types, so it's easier to work in).

lmiller1990 (approved):
Good to delete dead code - definitely need to 

1) get server properly type safe
2) figure out what exactly goes in data-context and server, and how the seam between the two should be decided

packages/server/lib/controllers/runner.ts
emilyrohrbough: yeah the line between the data-context & server is very blurred. I think intention was to separate server vs config/setup type concepts but we didn't get all the way there how intertwined some of this is

lmiller1990:
Going to go ahead and merge this up -- it's dead code, all CI is passing, I don't see much risk and there's not any additional manual testing required.

PR #23555, opened 2022-08-25
refactor: move console logging out of run.ts

flotwig:
- Closes 

### Additional details

- Makes `run.ts` easier to work on by factoring out all of the complicated display logic. Now `run.ts` is a measly ~1100 lines long, down from ~1700!

### PR Tasks

	- Existing snapshot coverage should not change.

### DIFF
--- a/packages/server/lib/util/print-run.ts
+++ b/packages/server/lib/util/print-run.ts
@@ -1,0 +1,536 @@
+/* eslint-disable no-console */
+import _ from 'lodash'
+import logSymbols from 'log-symbols'
+import chalk from 'chalk'
+import human from 'human-interval'
+import pkg from '@packages/root'
+import humanTime from './human_time'
+import duration from './duration'
+import newlines from './newlines'
+import env from './env'
+import terminal from './terminal'
+import * as experiments from '../experiments'
+import type { SpecFile } from '@packages/types'
+import type { Cfg } from '../project-base'
+import type { Browser } from '../browsers/types'
+
+type Screenshot = {
+  width: number
+  height: number
+  path: string
+  specName: string
+}
+
+function color (val, c) {
+  return chalk[c](val)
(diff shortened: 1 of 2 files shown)


packages/server/lib/util/print-run.ts
```diff
@@ -0,0 +1,536 @@
...
+  if (n) {
+    let nameWithNewLines = newlines.addNewlineAtEveryNChar(name, n)
+
+    return `${color(nameWithNewLines, colour)}`
+  }
+
+  return `${color(name, colour)}`
+}
+
+function formatNodeVersion ({ resolvedNodeVersion, resolvedNodePath }: Pick<Cfg, 'resolvedNodeVersion' | 'resolvedNodePath'>, width) {
```
emilyrohrbough, line 156: Can we split out / add the appropriate unit tests for this? The snapshots exist but are more a side-effect of other aspects vs explicitly testing these.

Also is it possible to fully type this new file verse adding partial types?

flotwig: The goal of this PR is to just split out some LoC from `run.ts` to make it easier to work in.

All of the `modes/run.js` unit tests were removed during unification, so there is not currently any unit-level testing for this logic. There's a ton of logic in here that would be helpful to have unit tests for, but since we already have coverage for this in the form of system-test snapshots, I don't feel it's particularly important to add in this PR, since it has survived without unit tests for this long.

I added more types to make this area easier to work in: e0aeb06ef2 I started trying to type the `results` object, but it turned into a rabbithole, it's overloaded with 3 separate result types (Mocha end, early exit, and dashboard skip), properly typing it would require either a refactor or a lot of falsy checks.

tbiethman (approved):
👍 👍

rachelruderman (approved):

Issue #22451, opened 2022-06-22
Cannot read properties of undefined (reading 'currentlyAttachedTarget')


koooge:
Hi there,

### Current behavior

```
Timed out waiting for the browser to connect. Retrying...
Cannot read properties of undefined (reading 'currentlyAttachedTarget')
TypeError: Cannot read properties of undefined (reading 'currentlyAttachedTarget')
    at Object.connectToNewSpec (/path/to/.cache/Cypress/10.2.0/Cypress/resources/app/packages/server/lib/browsers/chrome.js:431:48)
    at Object.connectToNewSpec (/path/to/.cache/Cypress/10.2.0/Cypress/resources/app/packages/server/lib/browsers/index.js:139:27)
    at /path/to/.cache/Cypress/10.2.0/Cypress/resources/app/packages/server/lib/open_project.js:148:47
    at tryCatcher (/path/to/.cache/Cypress/10.2.0/Cypress/resources/app/packages/server/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/path/to/.cache/Cypress/10.2.0/Cypress/resources/app/packages/server/node_modules/bluebird/js/release/promise.js:547:31)
    at Promise._settlePromise (/path/to/.cache/Cypress/10.2.0/Cypress/resources/app/packages/server/node_modules/bluebird/js/release/promise.js:604:18)
    at Promise._settlePromiseCtx (/path/to/.cache/Cypress/10.2.0/Cypress/resources/app/packages/server/node_modules/bluebird/js/release/promise.js:641:10)
    at _drainQueueStep (/path/to/.cache/Cypress/10.2.0/Cypress/resources/app/packages/server/node_modules/bluebird/js/release/async.js:97:12)
    at _drainQueue (/path/to/.cache/Cypress/10.2.0/Cypress/resources/app/packages/server/node_modules/bluebird/js/release/async.js:86:9)
    at Async._drainQueues (/path/to/.cache/Cypress/10.2.0/Cypress/resources/app/packages/server/node_modules/bluebird/js/release/async.js:102:5)
    at Immediate.Async.drainQueues (/path/to/.cache/Cypress/10.2.0/Cypress/resources/app/packages/server/node_modules/bluebird/js/release/async.js:15:14)
    at processImmediate (node:internal/timers:464:21)
    111
```

### Desired behavior

no error

### Test code to reproduce

[...]

PR #23556, opened 2022-08-25
chore(server): convert remaining browsers code to typescript

flotwig:
- Closes 

### Additional details

- This PR converts `packages/server/lib/browsers/index.js` to `.ts` and adds types for the `BrowserLauncher` implementations
- Also converts `electron.js` to `.ts`, although I did not spend as much time adding types here
- Converted Bluebird usage to native async/await as convenient
- With these changes, everything in server/lib/browsers is finally TSified
- [x] Depends on #23555

### PR Tasks


packages/server/lib/browsers/electron.ts
```diff
@@ -449,85 +450,87 @@ module.exports = {
...
-      // get our electron default options
-      // TODO: this is bad, don't mutate the options object
-      options = this._defaultOptions(projectRoot, state, options, automation)
+    debug('received saved state %o', state)
 
-      // get the GUI window defaults now
-      options = Windows.defaults(options)
+    // get our electron default options
+    // TODO: this is bad, don't mutate the options object
+    options = this._defaultOptions(projectRoot, state, options, automation)
```
lmiller1990, line 474: I agree mutation is the main thing that makes server so damn hard to work in :(

packages/server/lib/browsers/electron.ts
```diff
@@ -449,85 +450,87 @@ module.exports = {
...
 
-        win.once('closed', () => {
-          debug('closed event fired')
+    // cause the webview to receive focus so that
+    // native browser focus + blur events fire correctly
+    // https://github.com/cypress-io/cypress/issues/1939
+    tryToCall(win, 'focusOnWebView')
 
-          Windows.removeAllExtensions(win)
+    const events = new EE
```
lmiller1990, line 503:
```suggestion
    const events = new EE()
```

Trivial but I find `new` with `()` is so weird to look at :shrug: nitpicking, not a biggie

packages/server/lib/browsers/index.ts
```diff
@@ -62,32 +71,31 @@ const setFocus = async function () {
...
-const getBrowserLauncher = function (browser) {
+function getBrowserLauncher (browser: Browser, browsers: FoundBrowser[]): BrowserLauncher {
   debug('getBrowserLauncher %o', { browser })
-  if (!isBrowserFamily(browser.family)) {
-    debug('unknown browser family', browser.family)
-  }
 
   if (browser.name === 'electron') {
-    return require('./electron')
+    return require('./electron') as typeof import('./electron')
```
lmiller1990, line 78: Not sure if this breaks things but should be able to do something like (I think)

```js
return await import('./electron')
```

And get typing without `as typeof`. Also good to standardize on one syntax. Then `getBrowserLauncher` will return a `Promise` and well need to be awaited, though.

Or, we can just do

```ts
import chrome from './chrome'
```

Etc all at the top - not sure if there's any performance implication here, but moving towards proper ESM seems like a good idea, so maybe one day we can ship native ESM instead of CommonJS (although this is likely many years away, it would still be nice to work towards it).

packages/server/lib/browsers/types.ts
```diff
@@ -9,5 +10,29 @@ export type Browser = FoundBrowser & {
...
+   */
+  browser?: Browser
+  /**
+   * If set, the browser is currently in the process of exiting due to the parent process exiting.
+   * TODO: remove need for this
+   */
+  isProcessExit?: boolean
+}
+
+export type BrowserLauncher = {
```
lmiller1990, line 31: Yay, standardize browser launching interface, nice,

packages/server/lib/browsers/webkit.ts
```diff
@@ -18,7 +19,11 @@ export async function connectToNewSpec (browser: Browser, options, automation: A
   await wkAutomation.reset(options.url)
 }
 
-export async function open (browser: Browser, url, options: any = {}, automation: Automation): Promise<BrowserInstance> {
+export async function connectToExisting () {
```
lmiller1990, line 22:
```suggestion
export function connectToExisting () {
```

lmiller1990: Or should we type connectToExisting as

```ts
connectToExisting: (browser: Browser, options: BrowserLaunchOpts, automation: Automation) => Promise<void>
```

(and not allow a `void` return type)?

lmiller1990 (commented):
Cool, glad to see more TS, I hope we can get more and more strict. In general

- seems some un-necesary `async` usage
- consider `assert` over `throw new Error` (more concise and seems more idiomatic in my experience)

:100:

packages/server/lib/browsers/electron.ts
```diff
@@ -89,16 +93,16 @@ const _getAutomation = async function (win, options, parent) {
   return automation
 }
 
-const _installExtensions = function (win, extensionPaths = [], options) {
+async function _installExtensions (win: BrowserWindow, extensionPaths: string[], options) {
   Windows.removeAllExtensions(win)
 
-  return Bluebird.map(extensionPaths, (extensionPath) => {
+  return Promise.all(extensionPaths.map((extensionPath) => {
```
ryanthemanuel, line 99: 👍🏻

packages/server/lib/browsers/chrome.ts
```diff
@@ -679,21 +696,7 @@ export = {
 
     const pageCriClient = await browserCriClient.attachToTargetUrl('about:blank')
 
-    await this._setAutomation(pageCriClient, automation, browserCriClient.resetBrowserTargets, options)
```
ryanthemanuel, line 682: Nice deduping of this code here. Good clean up

ryanthemanuel (approved):
Nice work. Lots of good clean up here.

packages/frontend-shared/cypress/e2e/support/e2eSupport.ts
```diff
@@ -294,6 +294,7 @@ function startAppServer (mode: 'component' | 'e2e' = 'e2e', options: { skipMocki
         if (!ctx.lifecycleManager.browsers?.length) throw new Error('No browsers available in startAppServer')
 
         await ctx.actions.browser.setActiveBrowser(ctx.lifecycleManager.browsers[0])
+        // @ts-expect-error this interface is strict about the options it expects
         await ctx.actions.project.launchProject(o.mode, { url: o.url })
```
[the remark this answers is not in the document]
flotwig: It's only `OpenProjectLaunchOpts` at this point, which doesn't have `url`. This is an area that should be refactored, we're putting a lot of functionality into `project.launchProject` that makes it difficult to reason with.

packages/server/lib/browsers/chrome.ts
```diff
@@ -551,43 +549,60 @@ export = {
     return args
   },
 
-  async connectToNewSpec (browser: Browser, options: CypressConfiguration = {}, automation: Automation) {
+  async connectToNewSpec (browser: Browser, options: BrowserNewTabOpts, automation: Automation) {
     debug('connecting to new chrome tab in existing instance with url and debugging port', { url: options.url })
 
     const browserCriClient = this._getBrowserCriClient()
+
+    if (!browserCriClient) throw new Error('Missing browserCriClient in connectToNewSpec')
```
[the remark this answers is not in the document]
flotwig: `assert` doesn't seem to provide the type guarding needed, which is why I added this conditional. With `assert`, I still get `browserCriClient is possibly 'undefined'`

packages/server/lib/browsers/chrome.ts
```diff
@@ -551,43 +549,60 @@ export = {
...
+
     const pageCriClient = browserCriClient.currentlyAttachedTarget
 
+    if (!pageCriClient) throw new Error('Missing pageCriClient in connectToNewSpec')
+
+    if (!options.url) throw new Error('Missing url in connectToNewSpec')
+
+    await this.attachListeners(browser, options.url, pageCriClient, automation, options)
```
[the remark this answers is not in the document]
flotwig: Readability-wise, I prefer `await`ing a Promise over `return`ing it, because it makes it obvious you're doing an asynchronous operation at the "end" of the function.

packages/server/lib/browsers/electron.ts
```diff
@@ -37,12 +41,12 @@ const tryToCall = function (win, method) {
 }
 
 const _getAutomation = async function (win, options, parent) {
-  const sendCommand = Bluebird.method((...args) => {
+  async function sendCommand (method: string, data?: object) {
```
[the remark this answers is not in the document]
flotwig: Good call, made this change in a few places. Cleans up `cri-client` quite a bit: ce151923dab8d8e535bcc4a7be87f49de71db95

packages/server/lib/browsers/electron.ts
```diff
@@ -89,16 +93,16 @@ const _getAutomation = async function (win, options, parent) {
   return automation
 }
 
-const _installExtensions = function (win, extensionPaths = [], options) {
+async function _installExtensions (win: BrowserWindow, extensionPaths: string[], options) {
```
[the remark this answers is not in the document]
flotwig: ce151923dab8d8e535bcc4a7be87f49de71db95

packages/server/lib/browsers/index.ts
flotwig: Since each browser requires different dependencies, it's done this way for performance. Whether or not it makes a big (or indeed, positive) impact... I'm not sure, but I'm not going to refactor the lazy-loading in this PR. I did switch to await import: ce151923dab8d8e535bcc4a7be87f49de71db95

packages/server/lib/browsers/webkit.ts
flotwig: Made your suggested change in ce151923dab8d8e535bcc4a7be87f49de71db95

flotwig:
> with the new additional throw statements, do they fail gracefully or crash the server completely?

@ryanthemanuel All the new `throw`s are there for type guarding, where the type isn't providing a sufficient guarantee that a required property is defined in these places. Since our tests don't fail after these changes, I'm thinking it's unlikely users will see these errors. Server error handling should capture these as gracefully as any other browser launching error.

packages/server/lib/browsers/index.ts
```diff
@@ -129,89 +121,81 @@ exports = {
...
-          // TODO: make this a required property
-          if (!options.onBrowserOpen) throw new Error('onBrowserOpen did not exist in interactive mode')
-
-          options.onBrowserOpen()
-          ctx.browser.setBrowserStatus('open')
-
-          return instance
-        })
-      })
+  async open (browser: Browser, options: BrowserLaunchOpts, automation: Automation, ctx) {
```
emilyrohrbough, line 140:
```suggestion
  async open (browser: Browser, options: BrowserLaunchOpts, automation: Automation, ctx): BrowserInstance {
```

packages/server/lib/open_project.ts
```diff
@@ -20,7 +20,7 @@ const debug = Debug('cypress:server:open_project')
 
 export class OpenProject {
   private projectBase: ProjectBase<any> | null = null
-  relaunchBrowser: ((...args: unknown[]) => Bluebird<void>) | null = null
+  relaunchBrowser: (() => Promise<any>) | null = null
```
emilyrohrbough, line 23:
```suggestion
  relaunchBrowser: (() => Promise<BrowserInstance>) | null = null
```

flotwig: 5e743998a8

packages/server/lib/browsers/index.ts
flotwig: Currently, these 3 methods can actually resolve with `null | BrowserInstance`, since the browser can theoretically exit and null out `instance`. Added explicit types here: 5e743998a8

packages/server/lib/browsers/chrome.ts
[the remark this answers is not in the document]
emilyrohrbough: This should actually close https://github.com/cypress-io/cypress/issues/22451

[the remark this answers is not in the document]
flotwig: Yeah, at least it will give a clearer error message.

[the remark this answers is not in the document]
emilyrohrbough: Well...it doesn't fix it but it just give a better warning. I remember talking with @ryanthemanuel about this 
> we close the browser on a retry:
https://github.com/cypress-io/cypress/blob/develop/packages/server/lib/modes/run.js#L1127
but likely don't flip the indicator that tells it to try and launch the browser from scratch

lmiller1990 (approved):

PR #23579, opened 2022-08-27
chore(webkit): add video recording

flotwig:
- For #6422 

### User facing changelog

n/a

### Additional details

- Moves browser-specific video recording logic out of `run.ts` and in to the `browsers/*` drivers.
	- This happened in https://github.com/cypress-io/cypress/pull/23579/commits/e3b197998b34efc110a6ca087ada8beb0a0838a5
	- Browsers now control the video recording more closely, and only expose a small API to `run.ts` (start/end/restart control)
- Adds WebKit video recording.
	- This necessitated the above refactor - WebKit only supports recording an entire Page session at a time, and it streams to disk, so we can't use the ffmpeg method we regularly use with WK.
- Adds a `webkit-system-tests` job. 60 `.it`s are skipped in this PR since they don't work yet. It breaks down like this:
	- 6 skipped due to needing multidomain support
	- 8 skipped due to missing before:browser:launch support
	- 22 skipped due to broken stack traces
	- 24 other
- But importantly for this PR, the `video_compression` system-test spec is unskipped and passing in WebKit
- [x] Depends on #23587

### Steps to test

Run a test project that records video in WebKit, for example:

```
yarn cypress:run --project ../cypress-test-tiny --browser webkit --config video=true
```

(the project needs to have `playwright-webkit` installed)

Video should be recorded to the videosFolder after doing this. No guarantee that the project won't fail for other reasons since WebKit is WIP.

### PR Tasks


packages/types/src/server.ts
```diff
@@ -4,11 +4,57 @@ import type { PlatformName } from './platform'
 
 export type WriteVideoFrame = (data: Buffer) => void
 
+export type VideoRecording = {
```
flotwig, line 7: I'm super open to input on these APIs, especially around naming and structure. I know it's currently not the smoothest experience. They're structured the way they are to avoid a complete refactor of this part of `run.ts`, where two complex, asynchronous code paths that are interdependent require different parts of the `BrowserVideoController` at different times:

https://github.com/cypress-io/cypress/blob/4ad2279d0fbe3c9e37bc5a886a6e48819f207834/packages/server/lib/modes/run.ts#L848-L879

packages/server/test/integration/cypress_spec.js
```diff
@@ -1024,7 +1024,6 @@ describe('lib/cypress', () => {
               browser: 'electron',
               foo: 'bar',
               onNewWindow: sinon.match.func,
-              writeVideoFrame: sinon.match.func,
```
flotwig, line 1027: Removed this unit assertion since it's complex to mock now and is covered by other Electron tests that record video.

packages/types/src/video.ts
```diff
@@ -0,0 +1,60 @@
...
+  outputOptions?: string[]
+  videoFilters?: string
+}
+
+export type WriteVideoFrame = (data: Buffer) => void
+
+export type VideoRecording = {
+  api: RunModeVideoApi
+  controller?: BrowserVideoController
+}
```
flotwig, line 19: I'm super open to input on these APIs, especially around naming and structure. I know it's currently not the smoothest experience. They're structured the way they are to avoid a complete refactor of this part of `run.ts`, where two complex, asynchronous code paths that are interdependent require different parts of the `BrowserVideoController` at different times:

https://github.com/cypress-io/cypress/blob/4ad2279d0fbe3c9e37bc5a886a6e48819f207834/packages/server/lib/modes/run.ts#L848-L879

weyert:
Nice, great job, that’s a nice approach to add video capturing for Playwright :)

rachelruderman:
@flotwig can you update the PR description with Steps to Test? 🙏

flotwig:
> can you update the PR description with Steps to Test? pray

@rachelruderman done

rachelruderman (approved):
Tried and tested!! Great job!!

lmiller1990:
I did not review this one yet (was not assigned, but can take a look this week) - but I did *test* it out as part of https://github.com/cypress-io/cypress/pull/23662 and the recording feature is working as expected, at least based on a few tests. I did not do more in depth testing yet.

Should we be able to see recordings in the dashboard from WK with this PR @flotwig? My understanding is WK is only installed during local dev, so I guess not?

packages/server/lib/browsers/firefox.ts
```diff
@@ -529,7 +536,7 @@ export async function open (browser: Browser, url: string, options: BrowserLaunc
     // user can overwrite this default with these env vars or --height, --width arguments
     MOZ_HEADLESS_WIDTH: '1280',
     MOZ_HEADLESS_HEIGHT: '721',
-  }) as LaunchedBrowser & { browserCriClient: BrowserCriClient}
+  }) as unknown as BrowserInstance
```
tbiethman, line 539: I haven't seen this before, tricky.

packages/types/src/video.ts
```diff
@@ -0,0 +1,60 @@
...
+ * Interface passed to the browser to enable capturing video.
+ */
+export type RunModeVideoApi = {
+  onError: (err: Error) => void
+  videoName: string
+  compressedVideoName: string
+  /**
+   * Create+use a new VideoController that uses ffmpeg to stream frames from `writeVideoFrame` to disk.
+   */
+  newFfmpegVideoController: (opts?: { webmInput?: boolean}) => Promise<BrowserVideoController>
```
tbiethman, line 51: This does stand out a bit. Using `new` here makes me think of instantiation, but users of this API shouldn't really care much about that level of granularity. It might help to rename this to `setFfmpegVideoController()` to better indicate that there's more than just instantiation going on? I guess it could be unintuitive to return a Promise from what looks like a setter....maybe `useFfmpegVideoController()`? Just spitballing here.

Doing so might also help bring some parity to the `setVideoController`, which seems mutually exclusive with `newFfmpegVideoController`. Could similarly rename it to `useVideoController` and match the return values so that it's more obvious that one or the other API should be used.

tbiethman (approved):

flotwig:
> Should we be able to see recordings in the dashboard from WK with this PR? My understanding is WK is only installed during local dev, so I guess not?

@lmiller1990 in the monorepo, it's a `devDependency`, we run most of our tests using the dev version of Cypress so we do have video: https://dashboard.cypress.io/projects/ypt4pf/runs/38732/test-results/abbf0e35-9475-4a03-a446-743cb06c5e22/video

packages/types/src/video.ts
flotwig: I like that, I made the change in [45e9b8f](https://github.com/cypress-io/cypress/pull/23579/commits/45e9b8f8e04b075f37966ef157f6d8b2f7020a2c)

packages/server/lib/browsers/firefox.ts
flotwig: Yeah... I don't know the "right" way to change the type of an object, type-safely, without using `class ... extends ...`. If there even is a way without `as unknown`. Would be nice to eventually remove the need for this.

lmiller1990: This one seems a quick fix https://github.com/cypress-io/cypress/pull/23693

I think what you are referring to is something like

```ts
function foo<T extends SomeBaseClass>(arg: T): T {
  // 
}
```

Now as long as `arg` extends `SomeBaseClass` you should get type safety (error if you don't extend the base class).

lmiller1990: Maybe like this? https://www.typescriptlang.org/play?#code/IYIwzgLgTsDGEAJYBthjAgagUwBYEsVsEBvAKAUoVEhngQHNtF8ATBACgEoEAuBWvgB2DMmKoIA7rmzZkYfkICuAWxDYoFKrAD2Q2kvg6oHabPmLV6qD3ISJEAmAB0ZuRgC8Ume62UAvmSBZChoGADCwFAI2AAeENhCrBg4BESkfghsCF4A5LBRuZmJDMLY-Lly2PBQhLkIAD4IuQAOzFA6yEWZuvrQhhDGHCVlFVU1dY3NbdCdubaZEmBKbSYALFyLVI74LiNCxF772Jn+CEFioegIAEL4ANbEcQlJKXiEyMR2VNl5IA-YIoSFQ6VhyCogFSxepNXIgpRCCDAYQw5odYCsboSXoGIwmEFg5AQqGouE6BFIlFTXLozELexUZarDgAJk2DMoOxcBLkOQQPOQp3OwQAZgj4Pg9IwoLIIAAeAAqMXiiWSWHeRAAfBwWsAAJ7IHQY-gK+lIPRgTrYZyGhgcAAGCr1bSyGAAJCRdQajaxnGwzpJ8I4EB6vYaMa4fPIzso1BoEDoRd5zC57ZtggwZcwOAdJAhIiZKp8JrB5ptM7Kc9g83dHhwyRTkUIy2QgA

Also works with `interface` and plain objects ... not just limited to `class`.

Issue #22932, opened 2022-07-26
Improved Dashboard Integration column hover interactions


pstakoun:
Improve experience of hovering over and interacting with the Latest runs and Average duration columns.

[Figma Prototype](https://www.figma.com/proto/PAOozuCxtfetqP8fUVvg0y/branch/V0lCq4uBplf9J3GZRiD02m/Cypress-App?page-id=16373%3A174946&node-id=18481%3A162161&viewport=-3529%2C-2113%2C0.5&scaling=min-zoom&starting-point-node-id=18481%3A162161)

**AC:**
- Replace empty run dots with dashes when user is not logged in or does not have a valid project connected
- User hovering over dashes, will see one of the following on hover, depending on the state:
  - Connect (for log in flow)
  - Connect project (for connect project flow, if no valid project)
  - Request access (for request access flow, if no access)
- When offline, no hover behavior, just show the --
- No connection to smart prompts, but the CTAs should map to those in the header tooltips
- Include UTM params for login

Here are the [states in Figma](https://www.figma.com/file/V0lCq4uBplf9J3GZRiD02m/ACI-Fast-Follows?node-id=18493%3A162104)

[ClickUp](https://app.clickup.com/t/18033298/PM-2476)

abelb:
Background hover triggers tooltip
https://app.clickup.com/t/18033298/DESIGN-1233

baus:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/22932?planning-poker) with ZenHub @astone123 @lmiller1990 @marktnoonan @mike-plummer @rockindahizzy @warrensplayer @ZachJW34

abelb:
Design: WIP

[Figma Link](https://www.figma.com/file/V0lCq4uBplf9J3GZRiD02m/ACI-Fast-Follows?node-id=18272%3A162529)

lmiller1990:
@abelb what are we supposed to be looking at? I just see

![image](https://user-images.githubusercontent.com/19196536/182265557-cf8adef1-1554-4ef9-bcf4-451990dadbe4.png)

It's not clear on what is required for delivery of this ticket.

marktnoonan:
I'd agree, there is not currently a task here that can be estimated.

pstakoun:
Added a few notes but need to sync with @abelb on the final iteration of this

[...]

Issue #23379, opened 2022-08-16
feat(launchpad): show "connect project" after login if project is not connected


marktnoonan:
### What would you like?

In #23295 we set up the login flow to move into the "Connect Project" flow. This can also reasonably happen from the launchpad. We didn't include the launchpad in the initial PR because it involved some refactoring that would take more time, so we split this into 2 items.

The refactoring that is needed would be:

- Move all of the "Cloud Connect" components to the frontend-shared package
- Standardize the login process to always flow into "connect project" if you log in, when a project is open that has no `projectId` set.
- Initially this will only work when logging in _after_ a testing type has been chosen, since that is when the config has been parsed. In the future we could grab the config earlier in the process.

It seems possible to group all of this into a single component that can manage both login and cloud-connect flows, since the Cloud Connect modal already prompts you to log in at certain times.

One other thing to be aware of is potentially conflicting changes happening to put a variant of the Cloud Connect flow into the Specs List table rows as part of ACI work in this sprint.

PR #23584, opened 2022-08-28
feat: show cloud connect options when hoving spec list fields

warrensplayer:
- Closes https://github.com/cypress-io/cypress/issues/22932

### User facing changelog

Hovering over the "Latest runs" or "Average duration" columns in the Specs list will show buttons to help connect a project to the Cypress dashboard.

#### Implementation details
1. Added additional Button variant to match the styles in the design called "LinkBold"
2. Rendered new Button variant in component test plus added other existing variants that were not represented so they would be caught in the Percy snapshot
3. Added a new component called `SpecListCloudButton` to encapsulate the logic for the different states of a project being connected to the dashboard and show an appropriate CTA to help the user continue the workflow.  This logic to calculate this cloud connection status for the project was repeated from other components such as the `SpecHeaderCloudDataTooltip` and the `SpecsListBanners`.  This logic can be extracted to use the work being performed in #23379 that will encapsulate this cloud status in one place.
4. Added a new component called 'SpecListHoverCell' to encapsulate the logic for hovering over a cell in the Specs list to show the new button.  This logic does have some location dependent details in it related to styling to make it work in the Specs list.  To use this in a different location in the app, those details would need to be refactored into props.
    * I was not able to use the TailwindCss `group-hover` because the Spec row is already using `group` to change the icon when hovering the row.  Used the `useElementHover` from VueUse instead.
6. Changed grid columns for Specs list columns to use a flexible first column for the spec name and path and fixed widths for the last three columns since the width of the data for those columns is known.  This was done in order to make sure there was enough space for the new buttons that appear in the column on hover.  The columns do reduce in width and the button text for the new buttons is shortened at smaller breakpoints.
7. Updated the placeholder in the "Latest runs" column when the project is not connected to show two dashes (i.e. "--") just like the "Average duration" column

_Question for any reviewers_

- In order to defeat the hover functionality in the SpecListHoverCell component when not needed, I am used a 'prop' that just returns out of the `watch` function if set to true.  This value is calculated by the parent component.  Would it have been better to just conditionally replace that component with a plain `div`?

### Steps to test

Testing this feature will be similar to testing the Spec list banners ( #23256 ) and for the tooltips shown for the "Latest runs" and "Average duration" columns.  For each state shown in the table, confirm that the appropriate button is shown and clicking it performs the appropriate action.

| Cloud state | Button shown when hovering | Button action |
| -- | -- | -- |
| not logged in | "Connect" with user icon | opens login modal |
| logged in and project not connected | "Connect project" with  link icon | opens connect project modal |
| logged in and  project connected but no user access | "Request access" with play icon | requests access for the current user |
| logged in and request access has been sent | "Request sent" with play icon and button is disabled | No action for button |
| Project is fully connected to the cloud | No button shown when hovering | | 

At smaller breakpoints when the column headers for "Latest runs" and "Average duration" are shortened, then the button text will also be shortened to one word:

| Long text | Shortened text |
| -- | -- |
| Connect | _not shortened_ |
| Connect project | Connect |
| Request access | Request |
| Request sent | Sent |

#### Column widths for Specs list have been updated to fit content better

SpecList_NewColumnWidths

#### New buttons shown when hovering columns in Specs list

| When not logged in |
| ---- |
| SpecListHoverButton_NotLoggedIn |

| When not connected to a project  |
| ---- |
|  SpecListHoverButton_ConnectProject |

| When needing to request access to a project |
| -- |
| SpecListHoverButton_RequestAccess |

| When request for access has been sent |
| -- |
| SpecListHoverButton_RequestSent |

### PR Tasks

- [NA] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?


lmiller1990:
I'll divide my review and testing separately. Taking screenshots of each step to verify and keep track of my testing (there's quite a bit involved, you need two accounts to do the request + grant workflow).

Not connected

image

Request Access

image

Requested Access

image

I had to restart Cypress to get it to the "Accepted" state - I guess this is expected at this point in time, since we are not polling to see if an access request was granted.

packages/app/src/specs/SpecsList.vue
```diff
@@ -123,15 +126,26 @@
...
             <div
-              class="h-full grid justify-items-end items-center"
+              class="h-full grid justify-items-end items-center relative"
             >
               <RunStatusDots
                 v-if="row.data.isLeaf && row.data.data && (row.data.data.cloudSpec?.data || row.data.data.cloudSpec?.fetchingStatus !== 'FETCHING')"
                 :gql="row.data.data.cloudSpec ?? null"
                 :spec-file-extension="row.data.data.specFileExtension"
                 :spec-file-name="row.data.data.fileName"
+                @hasRuns="(v) => hasRuns = v"
```
lmiller1990, line 148: I found this a bit confusing - having both an event named `hasRuns` and a variable named `hasRuns`. I wonder if we can name this `setHasRuns`?

Also might be useful: `@hasRuns` compiles to `onHasRuns` - you can think of these like passing a `onHasRuns` callback in React. You can see what Vue compiles to using the [SFC Playground](https://sfc.vuejs.org/#eNpNj8FOxDAMRH/F8mVBYhPOVVrBL3D2pZSU7Yo4UeyWQ9V/x2FXiJvHnnka7/haitvWiB0GmepSFCTqWgbiJZVcFXaocYYD5poTnMx6IiaeMovCZZS3lQX65nl4fiQO/kaxvAmNqXyNGk0BhI9lg5d7pCfcoB/+ETbCIXjzmDn4vyQ+4a3IOY3FXSWzVd0bj+4HIezgd9N2VrBpwotqkc57maf24FVcrp/eJldX1iVFFyWd32v+llgNTNgQB/GBxw+mkF/r). This is why you can easily use JSX with Vue - the SFC (`.vue` file) compiles to something that basically maps 1:1 with React's JSX `h` function.

Eg: 
image

packages/app/src/specs/SpecsList.vue
```diff
@@ -210,6 +226,8 @@ const isOffline = ref(false)
 
 watch(isOnline, (newIsOnlineValue) => isOffline.value = !newIsOnlineValue, { immediate: true })
 
+const tableGridColumns: string = 'grid-cols-[1fr,135px,130px] md:grid-cols-[1fr,135px,130px,130px] lg:grid-cols-[1fr,160px,160px,180px]'
```
lmiller1990, line 229:
```suggestion
const tableGridColumns = 'grid-cols-[1fr,135px,130px] md:grid-cols-[1fr,135px,130px,130px] lg:grid-cols-[1fr,160px,160px,180px]'
```

packages/app/src/specs/SpecsList.vue
```diff
@@ -413,6 +433,10 @@ const { refetchFailedCloudData } = useCloudSpecData(
   props.gql.currentProject?.specs as SpecsListFragment[] || [],
 )
 
+const hasRuns = ref(false)
+
+watch([list], () => hasRuns.value = false)
```
lmiller1990, line 438:
```suggestion
watch(list, () => hasRuns.value = false)
```

lmiller1990: Also I don't fully understand the logic here - when list changes, we set `hasRuns` to `false` - why do we do that?

packages/app/src/specs/SpecsListCloudButton.cy.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+describe('<SpecsListCloudButton>', { viewportWidth: 300, viewportHeight: 400 }, () => {
+  function mountWithStatus (
+    status: 'NOT_FOUND' | 'LOGGED_OUT' | 'CONNECTED' | 'NOT_CONNECTED' | 'UNAUTHORIZED' | 'ACCESS_REQUESTED',
+  ) {
+    cy.mountFragment(SpecCloudDataHoverButtonFragmentDoc, {
+      onResult: (result) => {
+        set(result, 'cloudViewer', { __typename: 'CloudUser', id: 'abc123' })
+
+        switch (status) {
```
lmiller1990, line 13: Nice,  very concise :+1:

packages/app/src/specs/SpecsListCloudButton.vue
```diff
@@ -0,0 +1,141 @@
...
+      }
+    }
+  }
+  ...Auth
+  ...CloudConnectModals
+  ...RequestAccessButton
+}
+`
+
+const projectConnectionStatus = computed(() => {
```
lmiller1990, line 56: Nice use of `computed` here.

packages/app/src/specs/SpecsListCloudButton.vue
```diff
@@ -0,0 +1,141 @@
...
+    text: t('specPage.hoverButton.requestAccess'),
+    text_short: t('specPage.hoverButton.requestAccessShort'),
+    icon: PlayIcon,
+    emits: 'requestAccess',
+  },
+  ACCESS_REQUESTED: {
+    text: t('specPage.hoverButton.requestSent'),
+    text_short: t('specPage.hoverButton.requestSentShort'),
+    icon: PlayIcon,
+  },
```
lmiller1990, line 110: We don't even need to render this entire component if the state is `CONNECTED` (which is what `v-if="!!buttonOptions.text"` is doing, but in a not-so-obvious way).

Instead of rendering this component, a bunch of times (one for each row) and doing all these computations, should we just do something like this in the parent:

```html
<SpecsListCloudButton 
  v-if="connectionStatus !==  'CONNECTED'"
  :connectionStatus="connectionsStatus" />

<script>
const connectionStatus = computed(() => {
  // move logic from this component to parent - then we only need to do this calculation **once**
  // not every time we render this component, which is **a lot** of times
})
</script>
```

It would also simplify the code here and be more type safe, since we won't need all the `?` checks here, since `VALUES` will contain all possible values of `connectionStatus`.

Thoughts? This seems better both in terms of performance and type safety. The code and logic won't really change, it'll just be in a different location.

packages/app/src/specs/SpecsListHoverCell.cy.tsx
```diff
@@ -0,0 +1,61 @@
...
+    .should('be.visible')
+    .get(hoverSelector)
+    .should('not.exist')
+    .log('hover content to show hover slot')
+    .get(contentSelector)
+    .trigger('mouseenter')
+    .should('not.exist')
+    .get(hoverSelector)
+    .should('be.visible')
+    .log('hover slot should stay visible if hovering the hover slot')
```
lmiller1990, line 27: Oh, this is neat, cool way to document your test.

packages/app/src/specs/SpecsListHoverCell.vue
```diff
@@ -0,0 +1,73 @@
...
+
+const columnRef = ref()
+const hoveredItem = ref()
+
+const shouldShowHover = ref(false)
+
+const isHoveredColumn = useElementHover(columnRef)
+const isHoveredItem = useElementHover(hoveredItem)
+
+let controls
```
lmiller1990, line 47:
```suggestion
let controls: Stoppable
```

Can grab the type from `@vueuse/core`, I found it by hovering over the `useTimeoutFn` in VS Code.

packages/app/src/specs/SpecsListHoverCell.vue
```diff
@@ -0,0 +1,73 @@
...
+    isHoveredColumn.value = false
+  }
+
+  if (isHoveredColumn.value || isHoveredItem.value) {
+    if (shouldShowHover.value) return
+
+    controls = useTimeoutFn(() => {
+      shouldShowHover.value = true
+    }, 200)
+  } else if (!isHoveredColumn.value && !isHoveredItem.value) {
```
lmiller1990, line 68: Still thinking about this implementation, I wonder if we will have any perf/memory issues w/ constantly creating these hover event listeners.

Not sure if there is a better way, or if this even has any perf implications. I will think about it a little more. As an aside, why do we wait for 200 ms before showing the content?

packages/app/src/specs/SpecsListRowItem.vue
```diff
@@ -2,7 +2,8 @@
   <div data-cy="specs-list-row">
     <component
       :is="isLeaf ? 'RouterLink' : 'div'"
-      class="h-full outline-none ring-inset grid pr-20px grid-cols-7 group md:grid-cols-9 focus:outline-transparent focus-within:ring-indigo-300 focus-within:ring-1 children:cursor-pointer"
+      class="h-full outline-none ring-inset grid pr-20px group focus:outline-transparent focus-within:ring-indigo-300 focus-within:ring-1 children:cursor-pointer"
+      :class="[gridColumns]"
```
lmiller1990, line 6:
```suggestion
      :class="gridColumns"
```

packages/app/src/specs/SpecsListRowItem.vue
```diff
@@ -20,28 +20,54 @@
...
         </div>
-        <div>
-          <slot name="latest-runs" />
-        </div>
-        <div class="hidden md:col-span-2 md:block">
-          <slot name="average-duration" />
-        </div>
+        <SpecsListHoverCell
+          data-cy="specs-list-row-latest-runs"
+          :is-hover-disabled="hasRuns"
```
lmiller1990, line 28: Nice, I like this `is` convention for booleans.

packages/app/src/specs/SpecsListRowItem.vue
```diff
@@ -20,28 +20,54 @@
...
+        >
+          <template #content>
+            <slot name="latest-runs" />
+          </template>
+          <template #hover>
+            <slot
+              name="connect-button"
+              :utmMedium="'Specs Latest Runs Empty State'"
```
lmiller1990, line 36: Since it's a string that won't change we can do 
```suggestion
              utmMedium="Specs Latest Runs Empty State"
```

Vue can optimize this more now, as well, since it's never going to change.

packages/app/src/specs/SpecsListRowItem.vue
```diff
@@ -20,28 +20,54 @@
...
+          :is-hover-disabled="hasRuns"
+          class="hidden md:block"
+        >
+          <template #content>
+            <slot name="average-duration" />
+          </template>
+          <template #hover>
+            <slot
+              name="connect-button"
+              :utmMedium="'Specs Average Duration Empty State'"
```
lmiller1990, line 51:
```suggestion
              utmMedium="Specs Average Duration Empty State"
```

lmiller1990 (changes requested):
Nice job -- although you've been working in our stack for weeks, the vast majority of the code is idiomatic and looks like it was written by someone who's worked in the code base for quite a bit longer. 

Good job figuring out the GraphQL stuff, all the code-gen certainly makes for a pretty steep learning curve. Couple of small things.

As far as functionality, I tested it as best I can, seems like it works as expected.

packages/app/src/specs/RunStatusDots.vue
```diff
@@ -131,8 +142,20 @@ const props = defineProps<{
...
 
+const emits = defineEmits<{
+  (eventName: 'hasRuns', value: boolean): void
+}>()
+
 const runs = computed(() => {
-  return props.gql?.data?.__typename === 'CloudProjectSpec' ? props.gql?.data?.specRuns?.nodes ?? [] : []
+  const runs = props.gql?.data?.__typename === 'CloudProjectSpec' ? props.gql?.data?.specRuns?.nodes ?? [] : []
+
+  emits('hasRuns', !!runs)
```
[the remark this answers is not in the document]
warrensplayer: Good catch. Now that you called it to my attention, I did see it mentioned in the Vue docs to avoid side effects in `computed` properties.  I think I did get a linting hint in another place when trying to change an external value inside a computed property.  I will fix.

packages/app/src/specs/RunStatusDots.vue
```diff
@@ -1,63 +1,74 @@
...
+        #popper
+      >
+        <ExternalLink
+          v-if="latestRun"
+          :href="dashboardUrl"
+          :use-default-hocus="false"
+        >
+          <SpecRunSummary
+            :run="latestRun"
+            :spec-file-no-extension="props.specFileName"
```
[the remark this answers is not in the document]
warrensplayer: I have noticed those two patterns.  I have tried to match the existing pattern in what ever code I was working in unless doing a larger refactor. In this instance, this was existing code.

packages/app/src/specs/SpecsListCloudButton.cy.tsx
warrensplayer: Thanks, but I think I stole it from somewhere else!

packages/app/src/specs/SpecsListCloudButton.vue
warrensplayer: Also stole this pattern, but it is nice for sure.

packages/app/src/specs/SpecsListHoverCell.vue
warrensplayer: The delay was to prevent showing the buttons unless the user paused briefly over the cell.  This was intended to prevent the buttons from showing as a user quickly moved the mouse across the table.  

I share your concerns with perf/memory, but did some basic profiling and saw no major impacts.  If the specs list is changed from a virtual list that only shows a handful of things at a time to the full list in the DOM and only hides rows, then may be more of a concern.

I am currently reevaluating when these get shown based on your other related comments.

packages/app/src/specs/SpecsListRowItem.vue
warrensplayer: 👍

packages/app/src/specs/SpecsListHoverCell.cy.tsx
warrensplayer: 👍

packages/app/cypress/e2e/specs_list_latest_runs.cy.ts
```diff
@@ -111,15 +111,14 @@ function simulateRunData () {
 }
 
 function allVisibleSpecsShouldBePlaceholders () {
-  cy.findAllByTestId('run-status-dot-0').should('have.class', 'icon-light-gray-300')
-  cy.findAllByTestId('run-status-dot-1').should('have.class', 'icon-light-gray-300')
-  cy.findAllByTestId('run-status-dot-2').should('have.class', 'icon-light-gray-300')
-  cy.findAllByTestId('run-status-dot-latest')
-  .should('not.have.class', 'animate-spin')
-  .and('have.attr', 'data-cy-run-status', 'PLACEHOLDER')
+  // cy.findByTestId('run-status-dot-0').should('have.class', 'icon-light-gray-300')
```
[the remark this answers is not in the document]
warrensplayer: Removed

packages/app/src/specs/SpecsList.vue
```diff
@@ -28,17 +28,18 @@
     />
     <div
       v-if="specs.length"
-      class="mb-4 grid grid-cols-7 md:grid-cols-9 children:font-medium children:text-gray-800"
+      class="mb-4 grid children:font-medium children:text-gray-800"
       :style="`padding-right: ${scrollbarOffset + 20}px`"
+      :class="[tableGridColumns]"
```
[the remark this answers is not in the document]
warrensplayer: Corrected

packages/app/src/specs/SpecsListRowItem.vue
warrensplayer: Corrected

packages/app/src/specs/SpecsListRowItem.vue
warrensplayer: Fixed

packages/app/src/specs/SpecsListRowItem.vue
warrensplayer: Fixed

packages/app/src/composables/useRequestAccess.ts
```diff
@@ -0,0 +1,24 @@
...
+    ... on CloudProjectUnauthorized {
+      message
+      hasRequestedAccess
+    }
+  }
+}
+`
+
+export function useRequestAccess () {
+  const requestAccessMutation = useMutation(RequestAccessComposable_RequestAccessDocument)
```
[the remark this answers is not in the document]
warrensplayer: I am still finding my way around the code gen part of the graphql and the different patterns in play here.  @marktnoonan  is working on another issue in which he is pulling the calculation of the cloud status logic out to a reusable component.  The "request access" logic is part of that.  Definitely interested in pushing down business logic where appropriate.  I agree that the A/B testing will introduce more complexity that will need to be managed.

marktnoonan:
Not blocking, but looking at [this diff in Percy](https://percy.io/cypress-io/cypress/builds/21068456/changed/1176171912?browser=chrome&browser_ids=23&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=original&width=2000&widths=300%2C500%2C650%2C800%2C1200%2C2000)

Screen Shot 2022-08-29 at 10 30 48 AM

Do we want any sort of max-width on the first column in the specs list? On wider screens is puts all the content really far away from the spec names.

I like the new empty state a lot.

packages/app/src/specs/RunStatusDots.vue
```diff
@@ -131,10 +142,22 @@ const props = defineProps<{
...
 const runs = computed(() => {
   return props.gql?.data?.__typename === 'CloudProjectSpec' ? props.gql?.data?.specRuns?.nodes ?? [] : []
 })
 
+const isRunsLoaded = computed(() => {
+  return !!props.gql?.data
+})
+
+watchEffect(() => {
+  emits('hasRuns', isRunsLoaded.value)
```
marktnoonan, line 158: This pattern overall feels like something we could compute at the level above. It might not be doing exactly what you expect:

The `:has-runs="hasRuns"` in `SpecsList.vue` is binding _all rows_ to a single `hasRuns` ref in the scope of `SpecList.vue.` Each `RunStatusDots` component instance will set the value of `hasRuns` when it is rendered, which will affect all rows, making the last `RunStatusDots` that renders the "winner" in what gets set.

I suspect this is not exactly what you intended, and that we want each row to know if _that spec_ has any runs? Or more likely it's enough to know if _any_ spec in the project has runs, which would mean the project is connected and we don't need messaging about connecting it. But that's hard to derive without rendering all specs at once.

marktnoonan (commented):
Review in comments, sharing progress so far as I have to drop for a meeting. One question about `emits('hasRuns'` that might be an issue? Nothing else blocking so far, just little suggestions.

packages/app/src/specs/RunStatusDots.cy.tsx
```diff
@@ -28,6 +28,16 @@ function mountWithRuns (runs: Required<CloudSpecRun>[]) {
   })
 }
 
+function mountWithNoData () {
+  cy.mount(() => {
```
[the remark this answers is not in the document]
warrensplayer: I will just leave this one as is for now then since it is the same pattern as the other tests.  Is there a "best-practices" guide for how we should write our tests here that gets into bullet points like you are pointing out?  Something like:
> Do not use `beforeEach` or `afterEach` if there is only one test.  Just include the content in the test itself.

packages/app/src/specs/SpecsListHoverCell.vue
```diff
@@ -0,0 +1,73 @@
...
+  <div
+    ref="columnRef"
+    class="relative"
+  >
+    <slot
+      v-if="!shouldShowHover"
+      name="content"
+    />
+    <div
+      v-if="shouldShowHover"
```
marktnoonan, line 11: If we change this to a `v-else` we can wrap both of these in our `TransitionQuickFade` component which could be nicer than just popping these in.

packages/app/src/specs/SpecsList.vue
warrensplayer: Removed the hasRuns by pulling up the logic into this file

packages/app/src/specs/RunStatusDots.cy.tsx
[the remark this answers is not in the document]
marktnoonan: This guid would probably be the place: https://github.com/cypress-io/cypress/blob/develop/guides/testing-strategy-and-styleguide.md

Though I think there are _some_ valid use cases for the beforeEach with a single test, I don't remember the details.

packages/app/src/specs/SpecsListRowItem.cy.tsx
```diff
@@ -8,6 +8,7 @@ describe('SpecItem', () => {
       <SpecsListRowItem
         isLeaf={true}
         route={{ path: '/specs/runner', query: { file: '' } }}
+        gridColumns="grid-cols-[1fr,1fr]"
```
marktnoonan, line 11: Gotta say these seem so much better now when resizing 👍

packages/app/src/specs/SpecsListHoverCell.vue
warrensplayer: Done

packages/app/src/specs/RunStatusDots.vue
warrensplayer: Removed unused code

packages/app/src/specs/SpecsListRowItem.cy.tsx
warrensplayer: I agree!

packages/app/src/specs/SpecsListCloudButton.vue
```diff
@@ -53,76 +55,59 @@ fragment SpecCloudDataHoverButton on Query {
...
 const buttonOptions = computed(() => {
   const options: ButtonOptions = {
-    text: VALUES[projectConnectionStatus.value]?.text,
-    textShort: VALUES[projectConnectionStatus.value]?.text_short,
-    icon: VALUES[projectConnectionStatus.value]?.icon,
-    emits: VALUES[projectConnectionStatus.value]?.emits,
+    text: VALUES[props.projectConnectionStatus]?.text,
+    textShort: VALUES[props.projectConnectionStatus]?.textShort,
+    icon: VALUES[props.projectConnectionStatus]?.icon,
+    emits: VALUES[props.projectConnectionStatus]?.emits,
```
lmiller1990, line 103: Can get rid of `?` now

lmiller1990 (approved):

packages/app/src/specs/SpecsListCloudButton.vue
warrensplayer: Done

packages/app/src/specs/SpecsListRowItem.vue
```diff
@@ -20,28 +20,54 @@
...
+          data-cy="specs-list-row-latest-runs"
+          :is-hover-disabled="isProjectConnected"
+        >
+          <template #content>
+            <slot name="latest-runs" />
+          </template>
+          <template #hover>
+            <slot
+              name="connect-button"
+              utmMedium="Specs Latest Runs Empty State"
```
marktnoonan, line 36: Are we testing that these UTM medium values get used?

marktnoonan: Looks like we would want to stub/spy on the mutation called from Auth.vue. Looking around now to see if we have similar coverage.

marktnoonan (approved):
I have a comment about testing the utm medium for login, but since we are about to rework all of that and consolidate, we can add the testing for this in the upcoming sprint.

packages/app/src/specs/SpecsListCloudButton.vue
warrensplayer: Done

packages/app/src/specs/SpecsListRowItem.vue
warrensplayer: Added test to make sure correct `utmMedium` parameter  is passed to slot differently depending on which row you hover over.

PR #23588, opened 2022-08-29
chore(webkit): working around WebKit bug by patching Error

tbiethman:
- Related to #6422 

### User facing changelog

N/A

### Additional details

There were a few test cases within `uncaught_errors.cy.js` that were failing for the WebKit browser. The errors that were being thrown and uncaught within a Promise chain were not being received by Cypress. The root cause can be found documented within a [WebKit bug report here](https://bugs.webkit.org/show_bug.cgi?id=187822); it shows a minimal reproduction, which I can confirm with my own findings.

A good bit of debugging and trial and error resulted in the work-around in this PR. It was found that any sort of operation executed after the Error was thrown would ensure that the proper error handlers would fire. A no-op microtask was chosen for this purpose, as it seems to have the smallest chance of impacting other functionality.

### Steps to test

With this PR, all uncaught_errors specs will pass without any modification across our supported browsers.

### PR Tasks


packages/driver/src/cypress/cy.ts
```diff
@@ -550,6 +550,22 @@ export class $Cy extends EventEmitter2 implements ITimeouts, IStability, IAssert
...
+          // WebKit's unhandledrejection event will sometimes not fire within the AUT
+          // due to a documented bug: https://bugs.webkit.org/show_bug.cgi?id=187822
+          // To ensure that the event will always fire (and always report these
+          // unhandled rejections to the user), we patch the AUT's Error constructor
+          // to enqueue a no-op microtask when executed, which ensures that the unhandledrejection
+          // event handler will be executed if this Error is uncaught.
+          let originalError = autWindow.Error
+
+          autWindow.Error = function (...args) {
+            autWindow.queueMicrotask(() => {})
```
flotwig, line 563: Looking at this now, I wondered if this could backfire and cause an infinitely expanding queue, but I can't think of a scenario where that would happen where the user wouldn't already be doing infinite recursion anyways. So this seems safe enough...

packages/driver/src/cypress/cy.ts
```diff
@@ -550,6 +550,22 @@ export class $Cy extends EventEmitter2 implements ITimeouts, IStability, IAssert
...
         setWindowDocumentProps(autWindow, this.state)
 
+        if (this.Cypress.isBrowser('webkit')) {
+          // WebKit's unhandledrejection event will sometimes not fire within the AUT
+          // due to a documented bug: https://bugs.webkit.org/show_bug.cgi?id=187822
+          // To ensure that the event will always fire (and always report these
+          // unhandled rejections to the user), we patch the AUT's Error constructor
+          // to enqueue a no-op microtask when executed, which ensures that the unhandledrejection
+          // event handler will be executed if this Error is uncaught.
+          let originalError = autWindow.Error
```
flotwig, line 560:
```suggestion
          const originalError = autWindow.Error
```

flotwig (approved):

packages/driver/src/cypress/cy.ts
flotwig: Ah, I thought of one - if `window.queueMicrotask` is monkey-patched to throw an Error itself, this would infinitely recurse. But that seems unlikely to happen in the wild. Maybe we could name this function something like `__CyWebKitError` so that it's at least identifiable in a stack trace in case someone does run in to this?

tbiethman: That does seem unlikely, but naming the function is a good idea.

flotwig: Could even `try...catch` the `queueMicrotask` call... but... then that could also lead to really hard-to-debug issues ("why does my queueMicrotask monkey-patch make unhandled rejections not get caught by Cypress???"), so I think the way you have it is good for now. Let's see if any bug reports roll in.

ryanthemanuel (approved):

PR #23593, opened 2022-08-29
fix: angular 14.2 mount compilation error

### LINKED ISSUE #23585: Angular Component Test broken by changes in Angular 14.2.0 (tearDownTestingModule)
### Current behavior

Running Angular component tests fails:
```
2022-08-29 08:24:14: webpack compiled in 29272 ms (23942f0303cba40d)

  ExampleComponent
    1) "before each" hook for "mounts"

  0 passing (183ms)
  1 failing

  1) ExampleComponent
       "before each" hook for "mounts":
     TypeError: _angular_core_testing__WEBPACK_IMPORTED_MODULE_3__.TestBed.tearDownTestingModule is not a function

Because this error occurred during a `before each` hook we are skipping all of the remaining tests.
      at http://localhost:8080/__cypress/src/cypress_support_component_ts.js:645:62
      at $Cypress.<anonymous> (http://localhost:8080/__cypress/src/cypress_support_component_ts.js:433:73)
      at ../driver/node_modules/eventemitter2/lib/eventemitter2.js.EventEmitter.emit (http://localhost:8080/__cypress/runner/cypress_runner.js:99774:22)
      at parent.<computed> [as emit] (http://localhost:8080/__cypress/runner/cypress_runner.js:17[111](https://github.com/xxxxx/runs/8066523883?check_suite_focus=true#step:5:112)2:31)
      at $Cypress.action (http://localhost:8080/__cypress/runner/cypress_runner.js:163775:14)
      at fire (http://localhost:8080/__cypress/runner/cypress_runner.js:173427:18)
      at http://localhost:8080/__cypress/runner/cypress_runner.js:174835:13
      at tryCatcher (http://localhost:8080/__cypress/runner/cypress_runner.js:[113](https://github.com/xxxxx/runs/8066523883?check_suite_focus=true#step:5:114)18:23)
      at Promise.attempt.Promise.try (http://localhost:8080/__cypress/runner/cypress_runner.js:8592:29)
      at Object.onRunnableRun (http://localhost:8080/__cypress/runner/cypress_runner.js:174828:67)
```  

Angular (latest) version 14.2.0 version changed something in that part of their code:
https://github.com/angular/angular/blob/main/CHANGELOG.md#1420-2022-08-25
core - fix	align TestBed interfaces and implementation (#46635)
https://github.com/angular/angular/commit/796840209cd38aacc5061a31701efe7eda1f6587
https://github.com/angular/angular/pull/46635
_The change affected the public API surface, but in practice those types were already exposed via [the TestBedStatic interface](https://angular.io/api/core/testing/TestBedStatic), so those APIs were already public._

I think Cypress Angular Component Tests needs to update how they use that mentioned API.

### Desired behavior

_No response_

### Test code to reproduce

Update Angular to (latest) version 14.2.0 to reproduce.

### Cypress Version

v10.6.0

### Node version

v16.16.0

### Operating System

Windows 10.0.19043 Build 19043

### Debug Logs

_No response_

### Other

_No response_

ZachJW34:
- Closes #23585

### User facing changelog
Fix Angular 14.2 mount compilation error

### Additional details
Angulars 14.2 update changed the API we relied on for tearing down the component between tests. Before 14.2, `tearDownTestingModule` was a static method on the `TestBed` class but was changed to be a method on the instance. Converting all `TestBed` references to the `getTestBed()` factory function gives us the proper instance of TestBed with the methods we need and reatins backwards compatibility.

### Steps to test
Running the system tests in `system-tests/test/component_testing_spec.ts` is the best way to test. The `angular-14` project has had it's deps bumped. Without the changes to `mount` this test will fail.

https://user-images.githubusercontent.com/25158820/187244307-ef6d0a5b-a42c-423e-af5e-0795e6d08239.mov

### PR Tasks

### DIFF
--- a/npm/angular/src/mount.ts
+++ b/npm/angular/src/mount.ts
@@ -13,6 +13,7 @@ import {
   ComponentFixture,
   getTestBed,
   TestModuleMetadata,
+  TestBed,
 } from '@angular/core/testing'
 import {
   BrowserDynamicTestingModule,
@@ -171,7 +172,9 @@ function createComponentFixture<T> (
   component: Type<T> | string,
 ): Type<T | WrapperComponent> {
   if (typeof component === 'string') {
-    getTestBed().overrideTemplate(WrapperComponent, component)
+    // getTestBed().overrideTemplate is available in v14+
+    // The static TestBed.overrideTemplate is available across versions
+    TestBed.overrideTemplate(WrapperComponent, component)
 
     return WrapperComponent
   }
--- a/npm/angular/src/mount.ts
+++ b/npm/angular/src/mount.ts
@@ -12,7 +12,6 @@ import { Component, EventEmitter, Type } from '@angular/core'
 import {
   ComponentFixture,
   getTestBed,
-  TestBed,
   TestModuleMetadata,
 } from '@angular/core/testing'
 import {
@@ -144,12 +143,12 @@ function initTestBed<T> (
 
   const componentFixture = createComponentFixture(component) as Type<T>
 
-  TestBed.configureTestingModule({
+  getTestBed().configureTestingModule({
     ...bootstrapModule(componentFixture, configRest),
   })
 
   if (providers != null) {
-    TestBed.overrideComponent(componentFixture, {
+    getTestBed().overrideComponent(componentFixture, {
       add: {
         providers,
       },
@@ -172,7 +171,7 @@ function createComponentFixture<T> (
   component: Type<T> | string,
 ): Type<T | WrapperComponent> {
   if (typeof component === 'string') {
-    TestBed.overrideTemplate(WrapperComponent, component)
+    getTestBed().overrideTemplate(WrapperComponent, component)
 
     return WrapperComponent
   }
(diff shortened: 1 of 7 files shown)


jordanpowell88 (approved):

npm/angular/src/mount.ts
```diff
@@ -309,6 +308,6 @@ getTestBed().initTestEnvironment(
 
 setupHooks(() => {
   // Not public, we need to call this to remove the last component from the DOM
-  TestBed['tearDownTestingModule']()
-  TestBed.resetTestingModule()
+  getTestBed()['tearDownTestingModule']()
```
mike-plummer, line 311: So this function isn't part of the public API for `TestBed`? Would it make sense to guard it with a `try/catch` to keep us from blowing up, or would that just break tests in a less-obvious way?

ZachJW34: An error thrown here makes sense, since otherwise it would lead to some really strange behavior (like not removing the previously mounted component). Since it threw, it was very easy to track down what was going on and get a fix for it.

It is marked as a [public property](https://github.com/angular/angular/pull/46635/files#diff-2bd12fb2a5773e584a5fa36c468bbdb2a76747088b0444af9ff1a3034d6c562cR678), just not exposed on the interface for some reason.

mike-plummer (approved):

Issue #23486, opened 2022-08-22
Cypress UI no longer displays spec files to run


NUlmerTS:
### Current behavior

From version 10.5.0 onwards, the specs section no longer displays any specs in the file.
![image](https://user-images.githubusercontent.com/89112159/185836821-d3eebfee-a8c7-43d6-b602-df09ca50253e.png)

However, when reviewing the project settings spec patterns, nine files can be found
![image](https://user-images.githubusercontent.com/89112159/185836878-389ddc0e-a9b8-4121-a8d3-6cc02cd743ec.png)

It should be noted that when the tests are run in headless mode, they continue to work as expected.

The current workaround is to downgrade to 10.4.0.

### Desired behavior

After opening the cypress to use the UI format.  Then selecting a  browser types the spec files should display.

![image](https://user-images.githubusercontent.com/89112159/185837905-8844ad0a-96ae-4b23-b812-07e638852a7d.png)

### Test code to reproduce

const { defineConfig } = require('cypress')

module.exports = defineConfig({
  projectId: 'f2y8fb',
  chromeWebSecurity: false,
  responseTimeout: 60000,
  blockHosts: '*.parsely.com',
  pageLoadTimeout: 90000,
  e2e: {
    setupNodeEvents(on, config) {
      require("cypress-grep/src/plugin")(config);
    },
    baseUrl: 'https://www.teachstarter.com/',
    //baseUrl: 'https://staging-prexit.teachstarter.com/',
    specPattern: 'cypress/e2e/teachStarterWeb/**/*.{js,jsx,ts,tsx,spec}',
  },

### Cypress Version

10.5.0

### Node version

v14.17.5

### Operating System

Windows 11:22000.856

### Debug Logs

```shell
No logs displayed
```

lmiller1990:
Same as https://github.com/cypress-io/cypress/discussions/23460

lmiller1990:
Hi @NUlmerTS 👋We both previously worked at @doseme, small world!

Is your `specPattern` correct? `cypress/e2e/teachStarterWeb/**/.{js,jsx,ts,tsx,spec}` does not look correct - it it supposed to be `cypress/e2e/teachStarterWeb/**/*.{js,jsx,ts,tsx,spec}` (note the extra `*` in `**/*`).

In your second screenshot, it **does** have the `**/*` glob. Can you please clarify the `specPattern`?

[...]

PR #23595, opened 2022-08-29
fix: Issue 23486

amehta265:
- Closes https://github.com/cypress-io/cypress/issues/23486

### User facing changelog
- When using the Cypress UI, if the dev server is not set up correctly in the cypress config, users should still be able to see list of spec files they can run.

### Additional details

- The error stems from a missing null check for the component.devServer in the config. This could have happened if component testing was partially set up or devServer dependency was removed after the set up.
- Adding optional chaining to check the existence of the framework on the devServer circumvents the "GraphQL error [GraphQL] Cannot read properties of undefined (reading 'framework')". With optional chaining we will still be checking if the devServer is defined in the framework and if it is undefined it will not throw the error.
- The cypress-in-cypress.cy.ts test manually changes the config file removing the devServer from the component and then checks if the spec list still shows up.

### Steps to test

1. Go to cypress/packages/data-context/src/actions/ProjectActions.ts. On line 534 remove the '?' character between devServer and framework such that it reads as devServer.framework
2. Run the cypress-in-cypress.cy.ts test within the app folder of packages.
3. The 'checks that specs load when devServer configuration is not set in open mode for e2e/CT' test should fail as the spec list is not visible in the DOM.
4. Repeat step 1 BUT add back the '?' character so it reads as devServer?.framework
5. Repeat step 2 and the test should pass now.

Before (When config file does not have a devServer):
Screen Shot 2022-08-29 at 12 12 12 PM

After bug fix:
Screen Shot 2022-08-29 at 12 12 28 PM

### PR Tasks


packages/app/cypress/e2e/cypress-in-cypress.cy.ts
```diff
@@ -253,6 +253,42 @@ describe('Cypress in Cypress', { viewportWidth: 1500, defaultCommandTimeout: 100
...
+        await ctx.actions.file.writeFileInProject('cypress.config.js', config)
+
+        o.sinon.stub(ctx.actions.browser, 'closeBrowser')
+        o.sinon.stub(ctx.actions.browser, 'relaunchBrowser')
+      })
+
+      cy.get('[data-cy="loading-spinner"]').should('be.visible')
+      cy.get('[data-cy="loading-spinner"]').should('not.exist')
+
+      // We navigate to another page to allow enough time for the specs page to re-render with the updated changes in cypress config
```
marktnoonan, line 284:
```suggestion
      // We navigate to another page to ensure the specs page to re-renders with the updated changes in cypress config
```

marktnoonan (approved):
Suggested update to the comment but otherwise looks good!

astone123:
Nice! 👍🏻 @marktnoonan should we update the config types? If you look at `cypress.d.ts`, it looks like `devServer` is required

```
  interface ComponentConfigOptions<ComponentDevServerOpts = any> extends Omit<CoreConfigOptions, 'baseUrl' | 'experimentalSessionAndOrigin' | 'experimentalStudio'> {
    devServer: DevServerFn<ComponentDevServerOpts> | DevServerConfigOptions
    /**
     * Runs all component specs in a single tab, trading spec isolation for faster run mode execution.
     * @default false
     */
    experimentalSingleTabRunMode?: boolean
  }
```
it seems like it should be optional `devServer?: DevServerFn<ComponentDevServerOpts> | DevServerConfigOptions`

packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -531,6 +531,6 @@ export class ProjectActions {
     }
 
     // @ts-ignore - because of the conditional above, we know that devServer isn't a function
-    return WIZARD_FRAMEWORKS.find((framework) => framework.configFramework === config?.component?.devServer.framework)
+    return WIZARD_FRAMEWORKS.find((framework) => framework.configFramework === config?.component?.devServer?.framework)
```
ZachJW34, line 534: Rather than add the optional here, I think it's better to tweak the guard above:

```ts
if (!config?.component?.devServer || typeof config?.component?.devServer === 'function') {
  return undefined
}
...
```

ZachJW34:
@astone123 we shouldn't tweak the devServer types. If they have a `component` field in the Cypress config, they should have a `component.devServer` property otherwise they shouldn't have CT configured at all.

marktnoonan:
The ideal thing to do will be to correctly show a config validation error if `devServer` is not present. We need a follow up issue for that actually, this just stops the bug.

packages/data-context/src/actions/ProjectActions.ts
astone123: I like this suggestion 👍🏻

astone123 (approved):
Looks good 🎉

PR #23596, opened 2022-08-29
fix(angular): call ngOnChanges after mount

### LINKED ISSUE #23591: Angular Component Testing: `ngOnChanges` does not fire when using Class syntax
### Current behavior

When using the Class syntax in a Cypress Component test, `ngOnChanges` does not fire.,

### Desired behavior

ngOnChanges should always fire when an input property is set.

### Test code to reproduce

Create a component which implements OnChanges and make it set a flag which we display in the template:

``` typescript
import { Component, Input, OnChanges, OnInit, SimpleChanges } from '@angular/core';

@Component({
  selector: 'app-simple',
  template: `<div data-cy="simple">Hi {{ name }}. ngOnChanges fired: {{ ngOnChangesFired }}</div>`,
})
export class SimpleComponent implements OnChanges {
  @Input() name: string | undefined;

  ngOnChangesFired = false;

  ngOnChanges(changes: SimpleChanges): void {
    this.ngOnChangesFired = true;
  }
}
``` 

Create a Cypress component test and assert the output:
``` typescript
import { SimpleComponent } from './simple.component';

describe(`${SimpleComponent.name}`, () => {

  // This succeeds
  it('should fire ngOnChanges when using template syntax', () => {
    cy.mount(`<app-simple [name]="name"></app-simple>`, {declarations: [SimpleComponent], componentProperties: {name: 'Henk'}});

    cy.get('[data-cy=simple]').should('have.text', 'Hi Henk. ngOnChanges fired: true');
  });

  // This fails
  it('should fire ngOnChanges when using class syntax', () => {
    cy.mount(SimpleComponent, {componentProperties: {name: 'Henk'}});

    cy.get('[data-cy=simple]').should('have.text', 'Hi Henk. ngOnChanges fired: true');
  });

})
````

The first test succeeds (template syntax), and the second test fails (class syntax). 

### Cypress Version

10.6.0

### Node version

16.15.1

### Operating System

Windows 10

### Debug Logs

_No response_

### Other

Angular 13.3.11

### PR DESCRIPTION
**Pull request by jordanpowell88**
- Closes  #23591 

### User facing changelog
Fix that calls ngOnChanges after mounting an angular component

### PR Tasks

### CODE REVIEW

File: npm/angular/src/mount.ts [hunk 1]
```diff
@@ -230,6 +229,10 @@ function setupComponent<T> (
     })
   }
 
+  if (component.ngOnChanges) {
+    component.ngOnChanges()
```
mike-plummer, line 233: My Angular is a little rusty, but doesn't `ngOnChanges` expect to receive an object with the prev/new value for all the changed props? Wouldn't we need to construct an appropriate arg for this function?

Also, the original issue showed that components mounted using template strings *were* properly calling `ngOnChanges` - would this result in a double-call in those situations?

ZachJW34: Agree we should include a parameter that matches the `ngOnChanges` input. Here is how [atl](https://github.com/testing-library/angular-testing-library/blob/36648c4b431a9e17efa1f59cc2d83710c17a42ee/projects/testing-library/src/lib/testing-library.ts#L291) is doing it, we don't support `rerender` so it doesn't have to be too fancy.

@mike-plummer the component created via the use of templates will never have an `ngOnChanges` hook so no need to worry about the double call.

mike-plummer: @ZachJW34 I may be misunderstanding, but I don't think that's quite accurate. The template string example provided in the [original issue](https://github.com/cypress-io/cypress/issues/23591) *does* have an `ngOnChanges` that is properly firing ATM, so I think this change would cause it to double-fire on mount. Were you referring to the fact that you shouldn't expect it to fire *after* initial mount since there's no way to update the props?

Waterstraal: If I'm not mistaken, `ngOnChanges` only fires when a data-bound input property changes. So I think we should only call it when `componentProperties` is set.

system-tests/project-fixtures/angular/src/app/mount.cy.ts
```diff
@@ -164,7 +165,28 @@ describe("angular mount", () => {
...
+    cy.mount('<app-lifecycle [name]="name"></app-lifecycle>', {
+      declarations: [LifecycleComponent],
+      componentProperties: {
+        name: 'Angular'
+      }
+    })
+
+    cy.get('p').should('have.text', 'Hi Angular. ngOnInit fired: true and ngOnChanges fired: true')
+  })
+
```
Waterstraal, line 189: I would add 2 testcases (template syntax, class syntax) without `componentProperties` set where we assert that `ngOnChanges` was not fired.

npm/angular/src/mount.ts
ZachJW34: @mike-plummer I should have been more clear, when using a template, the component passed in via the template is rendered via a Proxy component called `WrapperComponent`. Then reason `ngOnChanges` works with the template is due to the fact that changing properties on the ProxyComponent (set via `componentProperties`) on first mount gets detected by change detection, and due to the template binding calls the child components `ngOnChanges` lifecycle event.

So there wouldn't be a double fire because the explicit call of `ngOnChanges` for template examples would be trying to fire the Proxy component's `ngOnChanges` which is a noop since it won't be defined.

Waterstraal:
@jordanpowell88 @mike-plummer @ZachJW34 Any idea when this PR will be merged? Thanks!

npm/angular/src/mount.ts
```diff
@@ -235,6 +234,20 @@ function setupComponent<T> (
...
+  if (component.ngOnChanges && config.componentProperties) {
+    const { componentProperties } = config
+    let simpleChanges: SimpleChanges
+
+    Object.keys(componentProperties).forEach((key: string, index: number, keys: string[]) => {
+      simpleChanges = {
+        ...simpleChanges,
+        [key]: new SimpleChange(null, componentProperties[key], true),
+      }
+    })
```
mike-plummer: Could streamline this a tiny bit with `reduce`:

```ts
const simpleChanges: SimpleChanges = Object.entries(componentProperties).reduce((acc, [key, value]) => {
  acc[key] = new SimpleChange(null, value, true)

  return acc
}, {})
```

npm/angular/src/mount.ts
```diff
@@ -235,6 +234,20 @@ function setupComponent<T> (
     })
   }
 
+  if (component.ngOnChanges && config.componentProperties) {
```
mike-plummer, line 240: From the contract it looks like `componentProperties` could be an empty object. Do we need to exclude that case from firing `ngOnChanges`?

jordanpowell88: Fixed in 045f59ec470dd921f2a8dc5cebae0e6098bab232

mike-plummer (approved):

system-tests/project-fixtures/angular/src/app/components/lifecycle.component.ts
```diff
@@ -0,0 +1,23 @@
...
+  ngOnChangesFired = false
+  conditionalName = false
+
+  ngOnInit(): void {
+    this.ngOnInitFired = true
+  }
+  
+  ngOnChanges(changes: SimpleChanges): void {
+    this.ngOnChangesFired = true;
+    if (changes['name'].currentValue === 'CONDITIONAL NAME') {
```
ZachJW34, line 19: This is exactly what we talked about, thanks for adding a test!!!

ZachJW34 (approved):

rujorgensen:
Will this also fix later @Input() changes (potential issue [here](https://github.com/cypress-io/cypress/issues/24198)) 🙂? It looks to me like this fix, and its tests, is only concerned with what happens upon mounting.

PR #23598, opened 2022-08-29
fix: don't show deps warning for projects that don't require bundler

### LINKED ISSUE #23216: Next.js CT apps have "Component Testing Mismatched Dependencies" warning for deps it doesn't need.

ZachJW34:
- Closes #23216

### User facing changelog
Remove misleading dependency warning in Launchpad for Next.js applications

### Additional details
We were including the bundler configured in the users `cypress.config` as a dependency to validate against. Next bundles it's own version of Webpack internally, meaning the user won't have a `node_modules/webpack` dep installed. The only dependencies we should validate against are the dependencies listed in the frameworks `dependencies` field.

For most frameworks, the `dependencies` list already includes a bundler. For projects that don't require `webpack/vite` to be explicitly installed (such as projects like Next.js that bundle it internally or `react-scripts` in which it is a transient dependency), it should not be listed. I've updated the list to remove the bundler from projects that had it listed when it shouldn't be.

### Steps to test
- Create a next application with `npx create-next-app@latest my-app` (don't put this project into a folder where webpack can be found up recursively...)
- cd into application and run `npm i -D cypress`
- Open cypress `npx cypress open`
- Walk through CT setup, browser screen should show `webpack` dep warning
- Kill Cypress and run local Cypress `yarn cypress:open`
- Add new next.js application and click on CT testing, no warning

https://user-images.githubusercontent.com/25158820/187308042-0d025aaf-4975-43ad-9fd4-bc2ecc5b2ab5.mov

### PR Tasks


packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -191,12 +191,6 @@ export class ProjectConfigManager {
       return
     }
 
-    const result = await isDependencyInstalled(bundler, this.options.projectRoot)
```
ZachJW34, line 194: `await framework.dependencies(...)` will return all the deps __including__ the bundler (if required) so no need to duplicate the bundler check.

packages/scaffold-config/src/frameworks.ts
```diff
@@ -86,7 +86,6 @@ export const WIZARD_FRAMEWORKS = [
     dependencies: (bundler: WizardBundler['type'], projectPath: string): Promise<DependencyToInstall[]> => {
       return Promise.all([
         isDependencyInstalled(dependencies.WIZARD_DEPENDENCY_REACT_SCRIPTS, projectPath),
-        isDependencyInstalled(dependencies.WIZARD_DEPENDENCY_WEBPACK, projectPath),
```
ZachJW34, line 89: `webpack` is a transient dependency of `react-scripts` so we don't need to list it here, same for `vue-cli`

lmiller1990 (approved):

lmiller1990:
@ZachJW34 looks like there might be some flake. Code seems fine :heavy_check_mark: 

I unskipped https://github.com/cypress-io/cypress/blob/565d22b5505a5638496ef1599e74cb402f6b721b/packages/launchpad/cypress/e2e/config-warning.cy.ts#L159 locally and it passes - I hope we can re-introduce this test soon, since it tests the opposite case (should show warning).

astone123 (approved):
Looks good 👍🏻

ZachJW34:
@lmiller1990 I unskipped it as well and it was working perfectly, but the failure seems to be related to a timeout of the node_modules scaffold so we need to figure out why it's timing out/find a way to speed up the install, could just be an expensive operation that hangs on underpowered machines (like the CI machines)

system-tests/projects/experimentalSingleTabRunMode/webpack.config.js
```diff
@@ -1 +1,3 @@
-module.exports = {}
+module.exports = {
+  stats: 'errors-warnings',
```
ZachJW34, line 2: Webpack compilation will log the number of modules generated e.g. `30 modules`, but where this is logged is non-deterministic so I've disabled the logging here. I kept errors-warnings in case something happens in the future that we would like to see.

PR #23601, opened 2022-08-29
test: add e2e test confirming login/connect button displays on hover

marktnoonan:
Here's a way to test this from e2e to ensure the component appears in the right place on hover. @warrensplayer feel free to merge to your branch or just use this as an example.

### DIFF
--- a/packages/app/cypress/e2e/specs_list_latest_runs.cy.ts
+++ b/packages/app/cypress/e2e/specs_list_latest_runs.cy.ts
@@ -200,6 +200,29 @@ describe('App/Cloud Integration - Latest runs and Average duration', { viewportW
 
       cy.findByTestId('average-duration-header').trigger('mouseleave')
     })
+
+    it('shows login/connect button in row when hovering', () => {
+      cy.get('[data-cy="spec-list-file"] [data-cy="specs-list-row-latest-runs"]')
+      .eq(0)
+      .as('latestRunsCell')
+      .trigger('mouseenter')
+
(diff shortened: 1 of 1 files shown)


packages/app/cypress/e2e/specs_list_latest_runs.cy.ts
```diff
@@ -200,6 +200,29 @@ describe('App/Cloud Integration - Latest runs and Average duration', { viewportW
...
       cy.findByTestId('average-duration-header').trigger('mouseleave')
     })
+
+    it('shows login/connect button in row when hovering', () => {
+      cy.get('[data-cy="spec-list-file"] [data-cy="specs-list-row-latest-runs"]')
+      .eq(0)
+      .as('latestRunsCell')
+      .trigger('mouseenter')
+
+      cy.contains('[data-cy="specs-list-row-latest-runs"] [data-cy="cloud-button"]', 'Connect').should('be.visible')
```
marktnoonan, line 210: Here we include the `latest-runs` so that the button is found in the right column

packages/app/cypress/e2e/specs_list_latest_runs.cy.ts
```diff
@@ -200,6 +200,29 @@ describe('App/Cloud Integration - Latest runs and Average duration', { viewportW
...
+
+      cy.get('@latestRunsCell').trigger('mouseleave')
+
+      cy.contains('[data-cy="cloud-button"]', 'Connect').should('not.exist')
```
marktnoonan, line 214: Here we leave out the `latest-runs` because no connect button should be shown _anywhere_ at that point.

Issue #22390, opened 2022-06-18
🐛 dev server vite 2.2.3 dynamic import are not using path aliases


hinogi:
### Current behavior

`importsToLoad` dynamic import of the spec file fails because it does not replace the path alias given by vite in the dynamic import.

https://github.com/cypress-io/cypress/blob/e5385fe413ae866985c04dfc991f8378673c6159/npm/vite-dev-server/client/initCypressTests.js#L9

This results in a TypeError when it resolves.

### Desired behavior

When vite is used and path aliases are given, check all kinds of imports or references to files with the according path alias.

### Test code to reproduce

`src/components/QuasarButton.vue`
```html
<template>
  <input
    data-cy="button"
    label="test emit"
    @click="$emit('test')"
  />
</template>

<script>
import { defineComponent } from 'vue';

export default defineComponent({
  name: 'QuasarButton',
  emits: ['test'],
});
```

`src/components/__test__/QuasarButton.spec.ts`
```
// __test__/QuasarButton.spec.ts
```
and a path alias for `src` pointing to src

### Cypress Version

10.1.0

### Other

The content does not really matter because cypress will not reach any of it because vite has found spec files and is providing the paths with knowing the path alias for it but not resolving it but initCypresssTests is blind to it and will fail trying to dynamically import from the path given by vite.

lmiller1990:
Hi, thanks for the bug report. Quasar is super nice to work with.

Just to clarify, you want to do something like:

```ts
export default defineConfig({
  plugins: [vue()],
  resolve: {
    alias: {
      '@': 'src'
    }
  }
})
```

And then type `import QuasarButton from '@components/QuasarButton.vue'`?

hinogi:
The path alias yes but the spec files are found by cypress itself, it is about spec files

IlCallo:
Adding a bit more context: @hinogi is trying to add support for Cypress 10 into our Cypress App Extension
You can find the draft PR here: https://github.com/quasarframework/quasar-testing/pull/259

lmiller1990:
Hello @IlCallo! Long time no talk.

I looked at the draft PR but I think I'm still missing something.

> The path alias yes but the spec files are found by cypress itself, it is about spec files

Can you point me some code that doesn't work (or even make a minimal reproduction)? I looked at the draft PR but I do not see any obvious `@` alias that isn't working. Also, the initial post doesn't include enough:

> src/components/__test__/QuasarButton.spec.ts
[...]

PR #23605, opened 2022-08-30
feat: allow vite/webpack config to be an async function

### LINKED ISSUE #23302: Allow webpack/vite config to be async function
### What would you like?

Right now you can pass a webpack/vite config to `component.devServer`:

```js
defineConfig({
  component: {
    devServer: {
      viteConfig: require('./vite.config.js')
    }
  }
})
```

It would help third parties integrate with Cypress if this could be an `async function` that receives the base Vite config as the first arg:

```js
defineConfig({
  component: {
    devServer: {
      viteConfig: async (baseConfig) => {
        const newConfig = await something()
        return {...baseConfig, ...newConfig}
      })
    }
  }
})
```

Context: https://github.com/cypress-io/cypress/issues/22390#issuecomment-1211523927

### Why is this needed?

It will help tools that want to ship pre-configured Cypress for Component Testing. Example: Quasar framework. Context: https://github.com/cypress-io/cypress/issues/22390#issuecomment-1211523927

### Other

_No response_

lmiller1990:
- Closes #23302

### User facing changelog

Add support for advanced dev server configuration via an async function that can optionally modify the dev server config.

### Additional details

Mainly useful for third party tooling that wants to ship a built in or pre-configured solution for Component Testing using Cypress. Case study and info here: https://github.com/cypress-io/cypress/issues/22390#issuecomment-1211523927

### Steps to test

Not much to do - just look at the test I added and see that they are passing.

Please review the docs PR, too: https://github.com/cypress-io/cypress-documentation/pull/4699

No change.

### PR Tasks


npm/vite-dev-server/src/resolveConfig.ts
```diff
@@ -90,7 +90,15 @@ export const createViteDevServerConfig = async (config: ViteDevServerConfig, vit
     ].filter((p) => p != null),
   }
 
-  const finalConfig = vite.mergeConfig(viteBaseConfig, viteOverrides as Record<string, any>)
+  let resolvedOverrides: UserConfig = {}
+
+  if (typeof viteOverrides === 'function') {
+    resolvedOverrides = await viteOverrides(viteBaseConfig)
```
ZachJW34, line 96: Would it make sense to utilize the [loadConfigFromFile](https://github.com/vitejs/vite/blob/8f315a20a7c3a33c72e7993860b6b69f5d6f0e05/packages/vite/src/node/config.ts#L872) function before executing the viteConfig function? That way, the config passed to the function would have the resolved values from the `configFile` that they could modify.

lmiller1990: Cool idea, but the `loadConfigFromFile` function is new in Vite 3; since we support Vite 2, we won't be able to use it if we want to have the same experience across both versions.

ZachJW34 (approved):
Looks good! One thing I would like to see is some unit tests in `npm/webpack-dev-server/test/makeWebpackConfig.spec.ts` and `npm/vite-dev-server/test/resolveConfig.spec.ts`.

npm/webpack-dev-server/src/devServer.ts
```diff
@@ -27,12 +27,16 @@ type FrameworkConfig = {
   }
 }
 
+export type ConfigHandler = Partial<Configuration>
+| ((baseConfig: Configuration) => Promise<Configuration>)
+| ((baseConfig: Configuration) => Configuration)
```
mike-plummer: Thoughts on consolidating the two function types and just vary the return type?

```
(baseConfig: Configuration) => (Configuration | Promise<Configuration>)
```

Side note: Should the return types be `Promise<Partial<Configuration>>` and `Partial<Configuration`?

mike-plummer (approved):
Looks good, just one tiny question/nit

npm/webpack-dev-server/src/devServer.ts
lmiller1990: Oh nice, this is obviously better than what I had - I don't know why I wrote it like this :man_facepalming:

lmiller1990: I think we don't need `Partial` at all - every property in `Configuration` is nullable already (via webpack's type defs) but I'll add it for good measure, since we seem to do it in a few other places.

lmiller1990:
Thanks for the reviews!

@mike-plummer @ZachJW34 updates with tests and improved types: https://github.com/cypress-io/cypress/pull/23605/commits/723e306a7b28ed8aa281138d6468c8dc205fd20c

Also I found some edges cases and decided we should slightly simplifying this.

## Before

```js
component: {
  devServer: {
    webpackConfig: async (baseConfig) => {
      return /* whatever you want */
    }
  }
}
```

## After


```js
component: {
  devServer: {
    webpackConfig: async () => {
      return /* whatever you want */
    }
  }
}
```
 
The change is we do not pass a baseConfig to the callback. It's not even clear what it should be - the CRA/Next.js config (if we found one? Should we pass that?) We probably do NOT want to pass in our internal baseConfig, that'll give the users opportunities to remove our plugins and break things by accident.

Also there is many edge cases we'd need to consider we *do* try to pass in their config we found. Vite has a `loadConfigFromFile` API in v3, but not in v2, so we'd need to implement our own. Looking at the Vite implemenation, it's really complex, since they consider `.mjs`, `mts`, etc etc. Instead of going too far down this path, the most simple solution is to just merge whatever the user returns from the callback using the idiomatic way for the bundler, which is:

- Vite: `vite.mergeConfig(cypressBase, userlandConfig)`
- webpack: `webpack.merge(cypressBase, userlandConfig)`

Which is what we already do.

This should still solve the problems we wanted to with this API - support for asynchronously modifying the config - as well as avoid needing to do the whole Node.js module dance, which is a mess (not just here, but in the entire ecosystem - even Vite has some problems with certain module combinations when you throw TS into the mix).

I updated the docs: https://github.com/cypress-io/cypress-documentation/pull/4699 to reflect this.

I'll run it by Brian next meeting but I think this should be fine - if anything, we should do the minimum changes to support the ecosystem and community, we can revisit passing a baseConfig to the callback if it's needed.

lmiller1990:
Merging is blocked until this is reviewed and finalized by product (that would be Brian).

PR #23624, opened 2022-08-30
feat: Add support for optional env key to browser launch options

ryanjduffy:
### User facing changelog

n/a

### Additional details

Related PR: https://github.com/cypress-io/cypress/pull/22250

In building out support to [integrate replay with cypress](https://github.com/replayio/replay-cli/tree/main/packages/cypress),  I discovered there isn't a way to set environment variables for the browser without putting the onus on the end user. Further, since the browser launch seems to happen in a different process than the plugin, I couldn't inject those values into `process.env` and have them picked up by the existing logic.

This adds support for an optional `env` key on the launch options passed to `before:browser:launch` which is then merged with the `defaultBrowserEnv` and `process.env`. I placed it between those two so the default browser family config (`defaultBrowserEnv`) would be overwritten by browser instance config (`browser.env`) which would be overwritten by the current `process.env`.

### Steps to test

* Create a new cypress project
* Add a `before:browser:launch` handler in `setupNodeEvents` to add an env variable
* Verify that the browser receives that env variable. I added a debug entry to log the environment which is probably sufficient for testing this change.

### PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4730


ryanjduffy:
@emilyrohrbough - Finally got back to this! Let me know what y'all think when you get to it. It'll go a long way to improve the DX for [replay.io](https://replay.io) users integrating with cypress.io!

emilyrohrbough:
@ryanjduffy Thank you so much! I will add this to my list to review and get back to you shortly. Hopefully by end of day today, if not by EOD tomorrow!

packages/launcher/lib/browsers.ts
```diff
@@ -181,7 +182,9 @@ export function launch (
 
   // allow setting default env vars such as MOZ_HEADLESS_WIDTH
   // but only if it's not already set by the environment
-  const env = Object.assign({}, defaultBrowserEnv, process.env)
+  const env = Object.assign({}, defaultBrowserEnv, browser.env, launchOptionsEnv, process.env)
```
emilyrohrbough, line 185: @ryanjduffy It looks like this change is doing more than accepting env values form the launchOptions. I wouldn't expect envs set on the found browser to be used in the node process that starts the browser. In the previous PR you had opened, you had linked to a few code snipped in reply & you're plugin. Where are these ENV values actually getting applied?

ryanjduffy: Not sure I follow but I'll share some context here in hopes that clarifies!

Here's the current code in our plugin. It's of course not setting env variables yet because it can't! 

https://github.com/replayio/replay-cli/blob/c90c03f08d9ccfcb9bc9a4a6a7c07246ff1b07f5/packages/cypress/src/index.ts#L11-L14

The eventual implementation after this landed would be:

```js
  on("before:browser:launch", (browser, launchOptions) => {
    selectedBrowser = browser.family;
    reporter.onTestSuiteBegin(undefined, "CYPRESS_REPLAY_METADATA");

    launchOptions.env = {
      RECORD_ALL_CONTENT: 1,
      RECORD_REPLAY_METADATA_FILE: '/path/to/file.json'
    };

    return launchOptions;
  });
```

The goal here being that we need to configure the replay browser with `RECORD_ALL_CONTENT` to instruct the firefox fork to record everything (unlike the desktop experience in which the user clicks the Rec button) and where to find metadata for the recording which is how we link up the recording with the test result.

This is all happening from the plugin because that's the point at which we have the test context to add the metadata. IIRC, I also didn't find another way to set an environment variables for the launcher node process beyond requiring the user to set them before invoking cypress.

Hope that helps! Happy to have a quick video call to chat if that's useful too.

ryanjduffy: Also, i'm in the cypress community discord as ryanjduffy if you want to chat there!

emilyrohrbough: @ryanjduffy This is great insights! Thank you!

emilyrohrbough, line 185:
```suggestion
  const env = Object.assign({}, defaultBrowserEnv, launchOptionsEnv, process.env)
```

packages/types/src/browser.ts
```diff
@@ -54,6 +54,8 @@ export type Browser = {
   minSupportedVersion?: number
   /** if set, is called to determine if found browser is supported by Cypress */
   validator?: BrowserValidator
+  /** optional additional environment variables set when launching the browser process */
+  env?: NodeJS.ProcessEnv
```
emilyrohrbough, line 58: We shouldn't need to update this

lmiller1990 (commented):
Is there a good way to test this manually? Eg - should we set up Replay locally? Considering this is a new feature, after all, I think we probably should do some testing on our end before we review + merge it.

packages/launcher/lib/browsers.ts
ryanjduffy: Looks like `browsers.env` was just cruft from my prior impl. @emilyrohrbough 's suggestions to drop those two diffs should do the trick!

emilyrohrbough (approved):

ryanjduffy:
Force pushed a fresh commit to rebase on `develop` and match up code style.

ryanjduffy:
> Is there a good way to test this manually? Eg - should we set up Replay locally? Considering this is a new feature, after all, I think we probably should do some testing on our end before we review + merge it.

Definitely! I'm walking through the steps myself now to verify it all one more time. Our [Cypress instructions](https://docs.replay.io/docs/cypress-instructions-370835875f044b71b41d88e30bedda20) cover most of it. I can push an experimental version of our plugin that uses this feature so others can try it as well.

ryanjduffy:
Manual testing steps:

1. Clone `replayio/replayable` and checkout `demo-cypress-env`. You could also follow the [instructions](https://docs.replay.io/docs/cypress-instructions-370835875f044b71b41d88e30bedda20) and set up your own project for extra credit
    ```
    git clone git@github.com:replayio/replayable && cd replayable && git checkout demo-cypress-env && npm i && npm run dev
    ```
2. From cypress, checkout this branch
    ```
    git fetch upstream +refs/pull/23624/merge && git checkout -qf FETCH_HEAD
    ````
4. Run cypress in dev mode against the replayable repo
    ```
    yarn cypress:run --project /path/to/replayable --browser "Replay Firefox"
    ````
5. Verify Replay made local recordings. There should be three entries.
    ```
    npx @replayio/replay ls
    ````

packages/launcher/lib/browsers.ts
```diff
@@ -166,6 +166,7 @@ export function launch (
   debuggingPort: number,
   args: string[] = [],
   defaultBrowserEnv = {},
+  launchOptionsEnv = {},
```
flotwig, line 169: This is too many parameters for this function, can you either remove the need for this parameter or move the parameters for `.launch` into an `options` object?

cli/types/cypress.d.ts
```diff
@@ -5480,6 +5480,7 @@ declare namespace Cypress {
     extensions: string[]
     preferences: { [key: string]: any }
     args: string[]
+    env: { [key: string]: any }
```
flotwig, line 5483: This is essentially a no-op in Electron, should we add an error or warning if someone sets .env with Electron?

flotwig (changes requested):
This needs a docs PR, this file should be edited: https://github.com/cypress-io/cypress-documentation/blob/master/content/api/plugins/browser-launch-api.md

packages/launcher/lib/browsers.ts
ryanjduffy: :+1: I'll collapse the two env members into one.

cli/types/cypress.d.ts
ryanjduffy: 👍 I noticed that `args` seems to be the same so I've added a warning for both of those. Let me know if I've misread and `args` is used or if there's a different way that is preferred for warning.

ryanjduffy:
> This needs a docs PR, this file should be edited: https://github.com/cypress-io/cypress-documentation/blob/master/content/api/plugins/browser-launch-api.md

Opened https://github.com/cypress-io/cypress-documentation/pull/4723

lmiller1990:
@ryanjduffy works great for me, I followed the steps and got it:

```
ID                                    Status  Title                          Created At
790783cd-f2c1-4c02-891c-fef7e2223767  onDisk  cypress/e2e/view-issues.cy.js  2022-09-11T23:15:34.279Z
054c8d69-fb1b-4429-a9da-6de10c48d738  onDisk  cypress/e2e/navigation.cy.js   2022-09-11T23:15:27.215Z
0943eec6-b68f-440b-86c1-341e161c7b76  onDisk  Replay of localhost:3000       2022-09-11T23:15:11.757Z
```

Left a quick comment on the docs PR, I'll wait for @flotwig to loop back and approve this PR, but looks good to me.

packages/server/lib/browsers/electron.ts
```diff
@@ -465,6 +465,18 @@ export = {
...
 
+  validateLaunchOptions (launchOptions: typeof utils.defaultLaunchOptions) {
+    const options: string[] = []
+
+    if (Object.keys(launchOptions.env).length > 0) options.push('env')
+
+    if (launchOptions.args.length > 0) options.push('args')
+
+    if (options.length > 0) {
+      errors.warning('BROWSER_UNSUPPORTED_LAUNCH_OPTION', 'electron', options)
```
lmiller1990, line 476: @ryanjduffy sorry last comment here before approval... how can I actually pass something to this? I want to verify this error actually shows up, but I couldn't figure out how.

lmiller1990 (commented):
Sorry for the delay, one last question!

packages/server/lib/browsers/electron.ts
ryanjduffy: Something like this should do it!

```
      on("before:browser:launch", (browser, launchOptions) => {
        launchOptions.env = {
          DEBUG: "whatever",
        };

        return launchOptions;
      });
```

image

ZachJW34 (commented):
Changes look good! I reran CI and we have a few build failures, lmk if you need help fixing those up. A few are snapshot related, but we did recently add webkit support so looks like we need to make some tweaks there.

ryanjduffy:
> Changes look good! I reran CI and we have a few build failures, lmk if you need help fixing those up. A few are snapshot related, but we did recently add webkit support so looks like we need to make some tweaks there.

Looks like it. Fixing up now

lmiller1990 (approved):
Pushed a commit ensuring the warning shows up correctly :+1: 

https://github.com/cypress-io/cypress/pull/23624/commits/62ad6d3468d7973918a92797cb1acacc68ffb3bb

lmiller1990:
Seems I broke something when I pushed a commit, reverting + fixing...

flotwig (approved):

flotwig:
Looks like the `deprecated_spec` snapshot needs to be updated, then this should be good to go:

```
cd system-tests
SNAPSHOT_UPDATE=1 yarn test deprecated_spec --browser electron
```

should do the trick.

lmiller1990:
Looks like a few more needed updating, I ran + pushed. :crossed_fingers:

ZachJW34:
Looks like we need to tweak one more system-test `using non-deprecated API - no warning`. The snapshot needs to be browser specific, since `args` is not supported for Electron the snapshot will differ. Looks like the pattern is to add:
```ts
onRun: (exec, browser) => {
  if (browser === 'electron') {
    return exec({ originalTitle: `deprecated before:browser:launch args / using non-deprecated API - no warning - [electron]` })
  }

  return exec({ originalTitle: `deprecated before:browser:launch args / using non-deprecated API - no warning - [firefox,chromium]` })
},
```

I'd push but my local is a bit borked. Could you add those changes and run `cd system-tests && SNAPSHOT_UPDATE=1 yarn test deprecated_spec`

ZachJW34:
LGTM!

PR #23643, opened 2022-08-31
fix: Improve document.cookie patch

### LINKED ISSUE #23531: Fix issue with the timing of setting cookies in the browser
Currently, if a cross-origin request sets a cookie and the app attempts to read the cookie via `document.cookie` synchronously on page load, the `document.cookie` value will not be correct / up-to-date.

Spec:

```js
it('document.cookie includes cookie set by http request', () => {
  cy.get('[data-cy="cookie-login-land-on-document-cookie"]').click()
  cy.origin('http://foobar.com:3500', { args: { username } }, ({ username }) => {
    cy.get('[data-cy="username"]').type(username)
    cy.get('[data-cy="login"]').click()
  })

  cy.origin('http://idp.com:3500', { args: { username } }, ({ username }) => {
    cy.get('[data-cy="document-cookie"]').invoke('text')
    .should('include', `user=${username}`)
  })
})
```

App:

```html
<p>
  <strong>document.cookie:</strong>
  <span data-cy="document-cookie"></span>
</p>
<script>
  document.querySelector('[data-cy="document-cookie"]').textContent = document.cookie;
</script>
```

chrisbreiding:
- Closes #23531

### User facing changelog

- Fixed an issue where, in certain circumstances, `document.cookie` would not reflect the correct value in cross-origin tests

### Additional details

Previously, the document.cookie patch would have to wait ~250ms to be up-to-date with cookies stored in the browser because it polls automation for them. This could result in document.cookie not being accurate if a page read its value on page load.

This fixes the issue by updating the document.cookie patch's value before the page loads by sending the cookies down from the proxy when the response is intercepted. This makes the correct value available on page load.

### PR Tasks


packages/driver/src/cross-origin/patches/cookies.js
```diff
@@ -1,3 +1,46 @@
+const parseCookieString = (cookieString) => {
```
mjhenkes, line 1: This file is going to be interesting to convert into an injection time patch when the spec bridge isn't attached yet.

mjhenkes, line 1: Is this a replacement for the tough cookie parsing or are they parsing different cookie string formats?

packages/driver/src/cross-origin/patches/cookies.js
```diff
@@ -13,43 +56,59 @@
...
 //   updates to the cookie jar (via http requests, cy.setCookie, etc) are
 //   reflected in the document.cookie value.
-export const patchDocumentCookie = (Cypress, window) => {
-  const setAutomationCookie = (toughCookie) => {
-    const { superDomain } = Cypress.Location.create(window.location.href)
-    const automationCookie = Cypress.Cookies.toughCookieToAutomationCookie(toughCookie, superDomain)
+export const handleDocumentCookie = (Cypress) => {
+  // if cookies were sent before the spec bridge was created, they'll come
+  // through as state('crossOriginCookies'), so use them then reset them
+  const cookies = Cypress.state('crossOriginCookies')
```
mjhenkes, line 62: is this state unique per spec bridge or could we get this state from the primary cypress instance through a post message?

packages/driver/src/cross-origin/patches/cookies.js
```diff
@@ -77,4 +136,30 @@ export const patchDocumentCookie = (Cypress, window) => {
...
+    updateCookies([cookie])
+  })
+
+  Cypress.on('clear:cookie', (name) => {
+    documentCookieValue = clearCookie(documentCookieValue, name)
+  })
+
+  Cypress.on('clear:cookies', () => {
+    reset()
+  })
```
mjhenkes, line 164: It looks like these could wait to be hooked up until a spec bridge is attached.

packages/driver/src/cross-origin/patches/cookies.js
```diff
@@ -77,4 +136,30 @@ export const patchDocumentCookie = (Cypress, window) => {
   }
 
   window.addEventListener('unload', onUnload)
+
+  Cypress.on('window:before:load', patch)
```
mjhenkes, line 140: Are we patching here to ensure that state is setup first? We will have to come up with something else if there isn't a spec bridge attached.

packages/driver/src/cross-origin/patches/cookies.js
```diff
@@ -77,4 +136,30 @@ export const patchDocumentCookie = (Cypress, window) => {
...
+
+  Cypress.specBridgeCommunicator.on('cross:origin:cookies', (cookies) => {
```
mjhenkes, line 142: This updates the cookies in the cross origin aut with changes pushed from the primary cypress instance? This should be pretty easy to not rely on cypress for.

packages/driver/src/cross-origin/patches/cookies.js
mjhenkes: instead of passing state here, could we call out to the primary cypress instance to retrieve these cookies?

packages/driver/src/cross-origin/patches/cookies.js
```diff
@@ -62,7 +121,7 @@ export const patchDocumentCookie = (Cypress, window) => {
```
AtofStryker, line 121: I think we need the full `hostname` here since `Cypress.Location` omits the subdomain. If we have cookies set on `www.example.com` and we call the automation API to get cookies for `.example.com`, are the cookies from `www.example.com` omitted? 

UPDATE: looks like cookies from `.www.example.com` are included with super domain `.example.com`. My guess is that this would also mean that other subdomains might get applied to `document.cookie` when we might not want them to, like `app.example.com`.

If we inject the AUTWindow into the patch, we should be able to use `window.location.hostname` to get the domain with the subdomain in it.

packages/driver/src/cross-origin/patches/cookies.js
```diff
@@ -13,43 +56,59 @@
...
+        const cookie = Cypress.Cookies.parse(newValue)
 
-    set (newValue) {
-      const cookie = Cypress.Cookies.parse(newValue)
+        // If cookie is undefined, it was invalid and couldn't be parsed
+        if (!cookie) return documentCookieValue
 
-      // If cookie is undefined, it was invalid and couldn't be parsed
-      if (!cookie) return documentCookieValue
+        documentCookieValue = mergeCookies(documentCookieValue, [{
```
AtofStryker, line 93: I don't think this currently preserves duplicates, such as 

```
const merged = mergeCookies('foo=bar; foo=bar', [{
  name: 'foo',
  value: 'baz',
}])

// merged is 'foo=baz', where I would expect something like 'foo=bar; foo=baz'
```

I played around with some cookies in chrome and firefox to spoof the following





I think we need more to go off of to set the cookie here properly, such as cookie domain and path, which it looks like we get from the `get:cookies` automation response.

Screen Shot 2022-09-08 at 4 25 46 PM

I wonder if we can take the cookies from the `setInterval` , i.e `Cypress.automation('get:cookies', { domain })` and store those within the `handleDocumentCookie` closure similar to how we store the `documentCookieValue`. Then, when a user gets `document.cookie`, we just stringify the cookies from the automation response. When a user sets a cookie, we can take the cookies from the variable that has the reference to the `get:cookies` stored variable and make sure we overwrite the correct cookie. The default should be the full `hostname` with path `/`, which I think `tough-cookie` handles but I can't remember off the top of my head.

packages/driver/src/cross-origin/patches/cookies.js
```diff
@@ -77,4 +136,30 @@ export const patchDocumentCookie = (Cypress, window) => {
...
+  Cypress.on('set:cookie', (cookie) => {
+    updateCookies([cookie])
+  })
+
+  Cypress.on('clear:cookie', (name) => {
+    documentCookieValue = clearCookie(documentCookieValue, name)
+  })
+
+  Cypress.on('clear:cookies', () => {
+    reset()
```
AtofStryker, line 163: I wonder if we still need to do this if we store the reference to `get:cookies`, or just clear the reference out?

packages/proxy/lib/http/response-middleware.ts
```diff
@@ -485,11 +485,13 @@ const CopyCookiesFromIncomingRes: ResponseMiddleware = async function () {
     return this.next()
   }
 
-  this.serverBus.once('cross:origin:automation:cookies:received', () => {
+  const origin = cors.getOriginPolicy(this.req.proxiedUrl)
```
AtofStryker, line 488: I think when it comes to `getOriginPolicy` for cookies, we need to factor in the subdomain. I think this method was originally intended for the `document.domain` matching to prevent AUT reloads, but we might need a different method here all together, as well as rename this method. I am going to try some sorts of a rename effort in #23728

AtofStryker (changes requested):
any idea if this fixes the issues inside the `cookie_behavior` tests with trying to get/set cookies inside `cy.origin` using the cypress api?

I think we might need to change the patch a little bit to work with cookies of the same key, but different domains/paths (which are apparently more common than I would have thought), but I think there is something we can try that I suggested below!

packages/driver/src/cross-origin/patches/cookies.js
chrisbreiding: The trick is we need them synchronously on page load. Post message won't work because it's async. I think what we'll need to do is hold on to them server-side and include them directly in the injection when we render it.

packages/driver/src/cross-origin/patches/cookies.js
chrisbreiding: It's here because it needs to be done on before the page loads. If we move this into the injection it will just work since the injection is the first script run.

packages/driver/src/cross-origin/patches/cookies.js
chrisbreiding: This ultimately comes from the proxy. If the spec bridge is already set up, it sends this event to pass in the cookies to the patch before the page load that might potentially use them via document.cookie.

packages/driver/src/cross-origin/patches/cookies.js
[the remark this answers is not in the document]
chrisbreiding: I had converted it to TypeScript and then converted it back when it became evident it would need to go into the injection, thinking it couldn't be TypeScript for that. Looking into it now, it seems that's not an actual limitation of code going into the injection, so I'll convert it to TypeScript once again.

packages/driver/src/cross-origin/patches/cookies.js
mjhenkes: We can do that. I already do it for modify obstructive code config.

packages/driver/src/cross-origin/patches/cookies.js
mjhenkes: lucky for us, once we move the patch to the injection window, will be the AUT window!

packages/driver/src/cross-origin/patches/cookies.js
mjhenkes: oh, that makes sense.

packages/driver/src/cross-origin/patches/cookies.js
chrisbreiding: Essentially they're different. Tough cookie will only parse a single cookie string, namely the one in `Set-Cookie` response headers. This is for parsing the `document.cookie` string, which includes multiple cookies in the format `key1=value1; key2=value2`.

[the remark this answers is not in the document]
chrisbreiding: Addressed in [`e728aed` (#23643)](https://github.com/cypress-io/cypress/pull/23643/commits/e728aed89f661e9ebaf7936e0899669e9214f028)

packages/driver/src/cross-origin/patches/cookies.js
```diff
@@ -13,43 +56,59 @@
...
+  Cypress.state('crossOriginCookies', [])
 
-    Cypress.automation('set:cookie', automationCookie)
-    .catch(() => {
+  const patch = (window) => {
+    // Cookies added to the server-side cookie jar are optimistically
+    // added here so that if a cross-origin request sets cookies, they're
+    // available via document.cookie synchronously on page load
+    const setAutomationCookie = (toughCookie) => {
+      const { superDomain } = Cypress.Location.create(window.location.href)
```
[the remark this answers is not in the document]
chrisbreiding: Now using [`location.hostname` instead](https://github.com/cypress-io/cypress/pull/23643/files#diff-0e98d845aa0ebdbe6c8b598f18dc81cc8f0ea4c09e7894ea32000e0ba4d116daR59).

packages/driver/src/cross-origin/patches/cookies.js
chrisbreiding: Now using [`location.hostname` instead](https://github.com/cypress-io/cypress/pull/23643/files#diff-0e98d845aa0ebdbe6c8b598f18dc81cc8f0ea4c09e7894ea32000e0ba4d116daR59).

packages/driver/src/cross-origin/patches/cookies.js
chrisbreiding: Fixed and [added a test](https://github.com/cypress-io/cypress/pull/23643/files#diff-6bf45b7f9903569c8f01830b58bfeac19264dc6b4eee767155d1b6484492b25fR854). Now use a `tough-cookie` cookie jar to track the cookies, so it applies the correct logic for handling duplicate cookie keys.

packages/driver/src/cross-origin/patches/cookies.js
chrisbreiding: This is to optimistically clear out the cookies when a user calls `cy.clearCookies()` so there's no delay/race condition waiting to sync with the automation cookies. There's actually no longer any polling to sync with automation cookies. It just caused issues and now all cases of cookies changing in automation are taken care of in a more deterministic fashion

packages/proxy/lib/http/response-middleware.ts
chrisbreiding: No longer using the origin here, so this shouldn't be a concern.

packages/proxy/lib/http/response-middleware.ts
```diff
@@ -372,10 +373,23 @@ const MaybePreventCaching: ResponseMiddleware = function () {
   this.next()
 }
 
+const setSimulatedCookies = (ctx: HttpMiddlewareThis<ResponseMiddlewareProps>) => {
+  const defaultDomain = (new URL(ctx.req.proxiedUrl)).hostname
+  const allCookiesForRequest = ctx.getCookieJar()
```
AtofStryker, line 380: should be put these variable declarations inside the `if` block since they are a no-op without it?

packages/proxy/lib/http/util/rewriter.ts
```diff
@@ -16,35 +17,56 @@ export type InjectionOpts = {
...
   wantsInjection: CypressWantsInjection
   wantsSecurityRemoved: any
+  simulatedCookies: AutomationCookie[]
 }
 
-const doctypeRe = /(<\!doctype.*?>)/i
-const headRe = /(<head(?!er).*?>)/i
-const bodyRe = /(<body.*?>)/i
-const htmlRe = /(<html.*?>)/i
+const doctypeRe = /<\!doctype.*?>/i
```
AtofStryker, line 23: Did removing the capture group here have any side effects / why did we need to remove it? Just redundant?

AtofStryker: I think I answered my own question after seeing the refactor for `insertBefore`/`insertAfter` 😄

packages/runner/injection/patches/cookies.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    return {
+      name: name.trim(),
+      value: value.trim(),
+      domain: location.hostname,
+      expiry: null,
+      httpOnly: false,
+      maxAge: null,
+      path: null,
+      sameSite: 'lax',
```
AtofStryker, line 23: This is done synchronously to seed the cookie jar with the simulated cookies on the initial response? I think ideally we could get these cookies out of CDP and set them here, but we need to do that async, and the only thing I think we get out of it is accurate `path`, `sameSite`, etc... I think these should be sane defaults, right?

packages/proxy/lib/http/response-middleware.ts
AtofStryker: or is there a side effect here?

packages/driver/src/cross-origin/cypress.ts
```diff
@@ -165,21 +165,18 @@ const attachToWindow = (autWindow: Window) => {
 
   const cy = Cypress.cy
 
+  // this communicates to the injection code that Cypress is now available so
+  // it can safely subscribe to Cypress events, etc
+  // @ts-ignore
+  autWindow.__attachToCypress(Cypress)
```
mjhenkes, line 171: lol, I have basically this same thing in my PR: https://github.com/cypress-io/cypress/pull/24243

packages/driver/src/cross-origin/communicator.ts
```diff
@@ -113,6 +113,10 @@ export class PrimaryOriginCommunicator extends EventEmitter {
     debug('Unexpected postMessage:', data)
   }
 
+  isConnectedToSpecBridge (originPolicy: string) {
```
mjhenkes: I think we just want to call this origin @AtofStryker 

```suggestion
  isConnectedToSpecBridge (origin: string) {
```

packages/runner/injection/cross-origin.js
```diff
@@ -92,13 +94,28 @@ const timers = createTimers()
 
 timers.wrap()
 
-const Cypress = findCypress()
+const attachToCypress = (Cypress) => {
```
mjhenkes, line 97: I borrowed your function name but have a slightly different spin on the function which i've setup to self destruct and remove itself from window after it's called.

packages/driver/src/cross-origin/communicator.ts
chrisbreiding: Realized I don't even use this method anymore so I removed it in https://github.com/cypress-io/cypress/pull/23643/commits/12277d6678d964cc1bf759ff834d7aeed13c0649

packages/driver/src/cross-origin/cypress.ts
chrisbreiding: Great minds and all 🧠

packages/proxy/lib/http/response-middleware.ts
chrisbreiding: Good call. Addressed in https://github.com/cypress-io/cypress/pull/23643/commits/5b7822e6544df861848f07ef711c272b04df06c3

packages/driver/src/cross-origin/events/cookies.ts
```diff
@@ -0,0 +1,47 @@
...
+  let cookiesToSend: AutomationCookie[] = []
+  let waitingToSend = false
+
+  Cypress.on('cross:origin:cookies', (cookies: AutomationCookie[]) => {
+    Cypress.backend('cross:origin:cookies:received')
+
+    cookiesToSend = cookiesToSend.concat(cookies)
+
+    const onBeforeStabilityRelease = (stable: boolean) => {
+      if (!stable) return
```
mjhenkes: It's so weird that 'before:stability:release' can get called with false stability, i would have thought that this action would be called when stability is false but just before it is set to true.

packages/runner/injection/patches/cookies.ts
chrisbreiding: Right, it's to grab any value currently in document.cookie so if a user reads it, it's still there. If it's originally read from `document.cookie`, its path/sameSite/etc must match already, so it makes sense for it to be read from `document.cookie` subsequently, right?

packages/driver/src/cross-origin/events/cookies.ts
chrisbreiding: Good point. Changed it to behave like that: https://github.com/cypress-io/cypress/pull/23643/commits/f1441f0c59522b19d6b9a05863f852fc2bf2c7e1

packages/driver/src/cross-origin/events/cookies.ts
```diff
@@ -0,0 +1,45 @@
...
+        // errors here can be ignored as they're not user-actionable
+      })
+    }
+
+    if (!waitingToSend) {
+      waitingToSend = true
+      // this event allows running a handler before stability is released.
+      // this prevents subsequent commands from running until the cookies
+      // are set via automation
+      Cypress.on('before:stability:release', onBeforeStabilityRelease)
```
mjhenkes: With your changes i think you could make this a `Cypress.once` and remove the `Cypress.off` call to clean it up inside the function.

chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/23643/commits/8c9d60f077eb12e18de078ffec11ff122494ba2b

packages/runner/injection/patches/cookies.ts
```diff
@@ -0,0 +1,133 @@
+import { AutomationCookie } from '@packages/server/lib/automation/cookies'
```
mjhenkes: Could we break automationCookie out into its own file? This import is pulling in an awful lot just to grab the interface.

```js
import _ from 'lodash'
import Debug from 'debug'
import extension from '@packages/extension'
import { isHostOnlyCookie } from '../browsers/cdp_automation'
```

chrisbreiding: Changed it to use `import type`, so it shouldn't pull anything in to the bundle. Addressed in https://github.com/cypress-io/cypress/pull/23643/commits/7b91ba5cde2a1fb690c3e08642e7909c5fd4522a

packages/runner/injection/patches/cookies.ts
```diff
@@ -0,0 +1,133 @@
...
+// if the user's app was run in top.
+//
+// The general strategy is:
+// - Keep the document.cookie value (`documentCookieValue`) available so
+//   the document.cookie getter can synchronously return it.
+// - Optimistically update that value when document.cookie is set, so that
+//   subsequent synchronous calls to get the value will work.
+// - On an interval, get the browser's cookies for the given domain, so that
+//   updates to the cookie jar (via http requests, cy.setCookie, etc) are
+//   reflected in the document.cookie value.
```
mjhenkes: Does this comment need to be updated? I don't think we still get cookies on an interval

chrisbreiding: Moved most of this comment into separate ones closer to where they're relevant: https://github.com/cypress-io/cypress/pull/23643/commits/1c91af77313d0e558d05668625c33088272ae8bb

packages/runner/injection/patches/cookies.ts
```diff
@@ -0,0 +1,127 @@
...
+
+      return getDocumentCookieValue()
+    },
+  })
+
+  const reset = () => {
+    cookieJar.removeAllCookies()
+  }
+
+  const bindCypressListeners = (Cypress: Cypress.Cypress) => {
```
mjhenkes, line 108: Do we need to listen for cookies changes even before a cypress instance attaches?

for example, would the following test work? (I haven't run this so it probably doesn't work without modifications)

```js
it('might need to listen without a cypress instance', () => {
  visit('localhost')
  visit('http://www.foobar.com:3500')
  cy.setCookie('foo', 'bar', {domain: foobar.com})
  cy.origin('foobar.com:3500', () => {
    cy.window((win) => {
      expect(win.document.cookies).to.eq('foo=bar')
    }
  })
}
```

packages/runner/injection/patches/cookies.ts
```diff
@@ -0,0 +1,127 @@
...
+    },
+  }, '*')
+}
+
+// document.cookie monkey-patching
+// -------------------------------
+// We monkey-patch document.cookie when in a cross-origin injection, because
+// document.cookie runs into cross-origin restrictions when the AUT is on
+// a different origin than top. The goal is to make it act like it would
+// if the user's app was run in top.
```
mjhenkes, line 47: In the case outside of cy.origin where we are same-site but cross-origin, would we want to apply this patch?
We may need to check on this for #24275

@AtofStryker

packages/runner/injection/patches/cookies.ts
chrisbreiding: Currently that fails with `The command was expected to run against origin http://localhost:3500 but the application is at origin http://www.foobar.com:3500`. I think it makes sense that the user should put it inside the origin callback in that case and then the listeners will be set up as well.

packages/runner/injection/patches/cookies.ts
chrisbreiding: Good to keep in mind. I'll defer to Bill on that one.

mjhenkes (approved):

packages/runner/injection/patches/cookies.ts
mjhenkes: I think we may want to revisit throwing that error:
>The command was expected to run against origin http://localhost:3500 but the application is at origin http://www.foobar.com:3500

I added the limitation that cy.cookies could only be run when the aut was same-site as top, I don't think thats correct anymore since we will be allowing cy.cookies to be set on any domain (i think it can be done now?) and we should probably think about changing this behaviour in #24264

packages/runner/injection/patches/cookies.ts
AtofStryker: I believe so, at least for `hostname` specific cookies

packages/runner/injection/patches/cookies.ts
AtofStryker: I think so. Does this get updated to the known values once the automation cookies sync in?

AtofStryker (approved):
Do we know if this fixes some of cookie assertions in `cookie_behavior.cy.ts`? I think most of those are fixed with https://github.com/cypress-io/cypress/issues/24264?

packages/runner/injection/patches/cookies.ts
chrisbreiding: It no longer polls to sync with the automation cookies, since that leads to race conditions and was pretty non-deterministic.

chrisbreiding:
> Do we know if this fixes some of cookie assertions in cookie_behavior.cy.ts? I think most of those are fixed with https://github.com/cypress-io/cypress/issues/24264?

I don't think so because this only fixes timing issues with `document.cookie`, not with cookies being available via automation. To fix those issues, we probably need to move Firefox to use CDP for cookies and ensure they're set before an cookie-setting requests are allowed to proceed.

packages/runner/injection/patches/cookies.ts
AtofStryker: oh duh 🤦🏻 . I wonder in the future if we can get automation cookies w/ simulated cookies to know these values upon injection. Otherwise though I feel like the defaults are sane and fine.

PR #23650, opened 2022-08-31
fix: Refetch latest cloud data when opening cloud connect modals

### LINKED ISSUE #23538: Refetch orgs and projects on connection modal open
If a project is created in the Dashboard before the project connection modal is opened, then it is not shown in the list of existing projects to connect. Similarly, if the modal is used to create a new project, then the projectId is deleted, it is not possible to then reconnect that same project.

A user's organizations and projects should be refetched whenever that modal is opened, so that the most up-to-date orgs and projects are shown as options to connect.

mike-plummer:
- Closes #23538 

### User facing changelog
* Fixes an issue where outdated organization and project information could be shown in dialogs to connect a project to the Cypress Dashboard.

### Additional details
* Adding a refetch when opening any of the Cloud Connect modals (Create/Select organization, Create/Select project) to reflect any org/project changes that have occurred outside the Cypress app since the user last logged in/opened the app
* Fixed an issue in the Select Project dialog that would continue to use old data in select menus after an updated `gql` prop was passed in

### Steps to test
This procedure involves creating/deleting cloud data. We point to Prod for cloud data by default now, if you would be more comfortable testing against Staging you can do so by launching like so `CYPRESS_INTERNAL_ENV=staging yarn dev`

1. Open a project that is not connected to the cloud (no `projectId` config entry) in the app, CT or E2E
2. Ensure you are logged in.
3. Open the dashboard in your browser (`dashboard.cypress.io` or `dashboard-staging.cypress.io` depending on how you launched). Make note of your organization memberships and what projects are in each org.
4. In the app, when mousing over the "Latest runs" and "Average duration" header tooltips you should get a "Connect project" action. Click to open the connect project modal, verify list of orgs and projects matches what is in the dashboard. Close the modal without taking any action.
5. In the dashboard, create a new project and name it something distinctive
6. In Cypress, re-open the Connect Project modal. Verify newly-created project is reflected in the options. Close the modal without taking any action
7. In the dashboard, create a new organization and name it something distinctive
8. In Cypress, re-open the Connect Project modal. Verify newly-created organization is reflected in the options.

**[Regression Test]**
9. Verify select menus within the modal update appropriately when selecting different organizations (projects list should change, previous project selection should be wiped out).
10. Select an org and a project and save selection using "Connect project" button. Verify correct project ID is saved into project config.

https://user-images.githubusercontent.com/11482842/187937181-4e4daa5b-37a5-4198-a9a7-8c90ec6071d1.mov

### PR Tasks

### DIFF
--- a/packages/app/src/runs/modals/CloudConnectModals.spec.tsx
+++ b/packages/app/src/runs/modals/CloudConnectModals.spec.tsx
@@ -6,8 +6,13 @@ import { CloudConnectModalsFragmentDoc } from '../../generated/graphql-test'
 import CloudConnectModals from './CloudConnectModals.vue'
 import cloneDeep from 'lodash/cloneDeep'
 
+type MountOptions = {
+  hasOrg: boolean
+  hasProjects: boolean
+}
+
 describe('<CloudConnectModals />', () => {
-  function mountDialog (hasOrg: boolean, hasProjects: boolean) {
+  function mountDialog ({ hasOrg, hasProjects }: MountOptions) {
     cy.mountFragment(CloudConnectModalsFragmentDoc, {
       onResult: (result) => {
         result.currentProject = {
@@ -41,7 +46,7 @@ describe('<CloudConnectModals />', () => {
 
   context('has no organization', () => {
     beforeEach(() => {
-      mountDialog(false, false)
+      mountDialog({ hasOrg: false, hasProjects: false })
     })
 
     it('shows the create/select org modal when orgs are added', () => {
@@ -54,7 +59,7 @@ describe('<CloudConnectModals />', () => {
   context('has organizations', () => {
     context('with no projects', () => {
       beforeEach(() => {
-        mountDialog(true, false)
+        mountDialog({ hasOrg: true, hasProjects: false })
       })
 
       it('shows the select project modal with create new project action', () => {
@@ -68,7 +73,7 @@ describe('<CloudConnectModals />', () => {
 
     context('with projects', () => {
       beforeEach(() => {
-        mountDialog(true, true)
+        mountDialog({ hasOrg: true, hasProjects: true })
       })
 
       it('shows the select project modal with list of projects', () => {
--- a/packages/app/src/runs/modals/CloudConnectModals.spec.tsx
+++ b/packages/app/src/runs/modals/CloudConnectModals.spec.tsx
@@ -4,9 +4,10 @@ import { CloudUserStubs,
 } from '@packages/graphql/test/stubCloudTypes'
 import { CloudConnectModalsFragmentDoc } from '../../generated/graphql-test'
 import CloudConnectModals from './CloudConnectModals.vue'
+import cloneDeep from 'lodash/cloneDeep'
 
 describe('<CloudConnectModals />', () => {
-  function mountDialog (noOrg = false) {
+  function mountDialog (hasOrg: boolean, hasProjects: boolean) {
     cy.mountFragment(CloudConnectModalsFragmentDoc, {
       onResult: (result) => {
         result.currentProject = {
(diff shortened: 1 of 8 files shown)


packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
```diff
@@ -644,9 +644,9 @@ export const mutation = mutationType({
       },
     })
 
-    t.field('refreshOrganizations', {
+    t.field('refreshCloudViewer', {
```
mike-plummer, line 647: Renaming this to make it clear this invalidates more than just the organizations data

astone123 (approved):
Works well for me 👍🏻

packages/app/src/runs/modals/CloudConnectModals.spec.tsx
```diff
@@ -18,7 +19,16 @@ describe('<CloudConnectModals />', () => {
 
         result.cloudViewer = {
           ...CloudUserStubs.me,
-          organizations: noOrg ? null : CloudOrganizationConnectionStubs,
+          organizations: hasOrg ? cloneDeep(CloudOrganizationConnectionStubs) : null,
```
lmiller1990, line 22: Why do we need to cloneDeep here?

packages/app/src/runs/modals/CloudConnectModals.spec.tsx
```diff
@@ -29,10 +39,43 @@ describe('<CloudConnectModals />', () => {
...
+      cy.contains(defaultMessages.runs.connect.modal.createOrg.button).should('be.visible')
+
+      cy.percySnapshot()
+    })
+  })
+
+  context('has organizations', () => {
+    context('with no projects', () => {
+      beforeEach(() => {
+        mountDialog(true, false)
```
lmiller1990, line 57: `mountDialog(true, false)` seems kind of hard to understand - I wonder if this should be an object of options?

packages/app/src/runs/modals/SelectCloudProjectModal.vue
```diff
@@ -285,31 +285,41 @@ const isInternalServerError = ref(false)
...
-const pickedProject = ref<CloudProjectNodeFragment>()
+  return sortBy(options || [], 'name')
+})
+const newProject = ref(projectOptions.value.length === 0)
+const pickedProject = ref<typeof projectOptions.value[number]>()
 
-watchEffect(() => {
-  if (pickedOrganization.value?.projects?.nodes?.length === 1) {
-    pickedProject.value = pickedOrganization.value.projects.nodes[0]
+watch([projectOptions], () => {
```
lmiller1990, line 316: How come we changed to this syntax? I think `watch` + `immediate: true` is the same as `watchEffect`?

Also I think this is now actually doing something a little different - if you use `watch`, you need to get the new value from the callback - [docs](https://vuejs.org/guide/essentials/watchers.html#basic-example). Eg:

```js
watch(projectOptions, (newProjectOptions) => {
                       ^^^^^^^ this is the latest value
  // stuff
})
```

lmiller1990 (commented):
Quick few comments, main query is around `watch` - I think the code change makes the behavior slightly different which might have been unintended?

packages/app/src/runs/modals/CloudConnectModals.spec.tsx
mike-plummer: The GQL stub isn't reset between tests since it's just a simple object import, since I'm mutating the data on lines 26-31 I didn't want my changes leaking into other tests

packages/app/src/runs/modals/SelectCloudProjectModal.vue
mike-plummer: Good call. My refactor meandered through a state where I had an reference loop between two `watchEffect`s hence the change to this format, but has since changed further so I can switch back to `watchEffect`

ZachJW34 (approved):
Works great, LGTM!

packages/app/src/runs/modals/CloudConnectModals.spec.tsx
mike-plummer: @lmiller1990 Going to merge this PR so it doesn't sit too long, but let me know if you have an issue with this pattern or know of a better way to address it - happy to revisit this

PR #23659, opened 2022-09-01
chore: valid config files should be collapsed during onboarding

### LINKED ISSUE #23260: When config files are created during onboarding, keep them collapsed
During onboarding, when config files are created, keep them collapsed on the "Configuration Files" screens. If they have warnings (changes required, etc.), then start with them open.

Screen Shot 2022-08-10 at 4 55 42 PM

amehta265:
- Closes #23260

### User facing changelog
- During Cypress onboarding, when config files are created, all valid, skipped, error config files will remain collapsed on default while warning files will be open on default.

### Additional details

- This reduces the information shown to the user on a 'config success' screen making the page more easier to read.
- In order to allow for more visual accessibility I changed the color contrast of the header background and header description such that the description is now text-gray-700 from the previous text-gray-600. I was wondering if this is okay @mapsandapps 

### Steps to test

- All tests are inside launchpad folder
1. Run yarn dev from the root folder (cypress)
2. Add the launchpad as a project on the Cypress UI
3. Go to e2e testing and run the project-setup.cy.ts spec file. This contains all updated e2e tests that check the default collapsible functionality.
4. Change to Component Testing and run the FileRow.cy.ts spec file. This contains all the component tests for FileRow where I made the changes.

Before:
Screen Shot 2022-09-01 at 11 40 57 AM
After:
Screen Shot 2022-09-01 at 11 39 26 AM

### PR Tasks

### DIFF
--- a/packages/launchpad/cypress/e2e/project-setup.cy.ts
+++ b/packages/launchpad/cypress/e2e/project-setup.cy.ts
@@ -84,7 +84,7 @@ describe('Launchpad: Setup Project', () => {
       cy.containsPath('cypress/support/e2e.js')
       cy.containsPath('cypress/support/commands.js')
       cy.containsPath('cypress/fixtures/example.json')
-      cy.get('[data-cy=collapsible-header]').invoke('attr', 'aria-expanded').should('eq', 'false')
+      cy.get('[data-cy=collapsible-header]').should('have.attr', 'aria-expanded', 'false')
     })
 
     verifyScaffoldedFiles('e2e')
@@ -240,7 +240,7 @@ describe('Launchpad: Setup Project', () => {
           cy.containsPath('cypress/support/e2e.js')
           cy.containsPath('cypress/support/commands.js')
           cy.containsPath('cypress/fixtures/example.json')
-          cy.get('[data-cy=collapsible-header]').invoke('attr', 'aria-expanded').should('eq', 'false')
+          cy.get('[data-cy=collapsible-header]').should('have.attr', 'aria-expanded', 'false')
         })
 
         verifyScaffoldedFiles('e2e')
@@ -307,7 +307,7 @@ describe('Launchpad: Setup Project', () => {
           cy.containsPath('cypress/support/e2e.js')
           cy.containsPath('cypress/support/commands.js')
           cy.containsPath('cypress/fixtures/example.json')
-          cy.get('[data-cy=collapsible-header]').invoke('attr', 'aria-expanded').should('eq', 'false')
+          cy.get('[data-cy=collapsible-header]').should('have.attr', 'aria-expanded', 'false')
         })
 
         verifyScaffoldedFiles('e2e')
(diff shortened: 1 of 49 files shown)


packages/launchpad/cypress/e2e/project-setup.cy.ts
```diff
@@ -84,6 +84,7 @@ describe('Launchpad: Setup Project', () => {
       cy.containsPath('cypress/support/e2e.js')
       cy.containsPath('cypress/support/commands.js')
       cy.containsPath('cypress/fixtures/example.json')
+      cy.get('[data-cy=collapsible-header]').invoke('attr', 'aria-expanded').should('eq', 'false')
```
ZachJW34, line 87:
```suggestion
      cy.get('[data-cy=collapsible-header]').should('have.attr', 'aria-expanded', 'false')
```

I'm not a big fan of invoke when you can use something more standard. This comment applies to all other cases where invoke has been added.

ZachJW34 (approved):
Works great! One small nit but not blocking

astone123 (approved):
Looks good 👍🏻

rockindahizzy (approved):
Nice!

packages/launchpad/cypress/e2e/project-setup.cy.ts
```diff
@@ -305,6 +307,7 @@ describe('Launchpad: Setup Project', () => {
           cy.containsPath('cypress/support/e2e.js')
           cy.containsPath('cypress/support/commands.js')
           cy.containsPath('cypress/fixtures/example.json')
+          cy.get('[data-cy=collapsible-header]').should('have.attr', 'aria-expanded', 'false')
```
lmiller1990, line 310: I don't think this is fully doing what you might expect:

image

As long as **one** is `false`, this will pass. I've got a valid one open here, and it isn't failing the test. 

We might want to add some extra data selectors, so something like `data-cy=valid`, than assert all those are collapsed. It might look like

```js
cy.findByTestId('valid').should('have.attr', 'aria-expanded', 'false')
```

lmiller1990 (changes requested):
Just want to make sure the test is doing what it should be - I think this will always pass no matter what. I could be wrong, if so please lmk. Thanks!

packages/launchpad/cypress/e2e/project-setup.cy.ts
amehta265: This was a great catch! As you mentioned, the `cy.get('[data-cy=collapsible-header]').should('have.attr', 'aria-expanded', 'false')` would pass as long as at least 1 collapsible-header was closed which does not test all of them being closed.

After some troubleshooting and with the help of Mark we came up with a solution:
`cy.get('[data-cy=valid] [data-cy=collapsible-header]').each((element) => {
          cy.wrap(element).should('have.attr', 'aria-expanded', 'false')
        })`
This individually asserts the collapsible headers of a certain type. I made the updates to all the tests to reflect this change

lmiller1990 (approved):

PR #23662, opened 2022-09-01
chore(webkit): add before:browser:launch and download support

flotwig:
- For #6422 

### User facing changelog

n/a

### Additional details

- Adds `before:browser:launch` support. `launchOptions.extensions` isn't supported since WK doesn't take extensions.
- Adds support for `config.downloadsFolder` and download events.
- Also fixes the issue where Cypress would reopen in interactive mode after `cypress run` ended, since it was making local system-tests fail.
- [x] Depends on #23579 

### Steps to test

Refer to #23579 testing steps for instructions on installing WebKit.

1. Run a project with `before:browser:launch` using WebKit, observe that the event is hit and that the launchOptions can be edited.
2. Run a project that downloads files, note that they are saved to `downloadsFolder` when using WebKit.
3. Exit a Cypress process after launching WebKit, note that Cypress no longer re-opens itself back up in `open` mode.

### PR Tasks


packages/server/lib/browsers/webkit.ts
```diff
@@ -53,7 +83,7 @@ export async function open (browser: Browser, url: string, options: BrowserLaunc
     }
 
     async kill () {
-      debug('closing pwBrowser')
+      debug('killing pwServer process tree')
```
flotwig, line 86:
```suggestion
      debug('closing pwBrowser')
```

packages/server/lib/browsers/webkit.ts
```diff
@@ -26,22 +31,47 @@ export function connectToExisting () {
...
+  launchOptions.preferences.args = [...launchOptions.args, ...(launchOptions.preferences.args || [])]
+
+  const pwServer = await pw.webkit.launchServer(launchOptions.preferences)
+
+  /**
+   * Playwright adds an `exit` event listener to run a cleanup process. It tries to use the current binary to run a Node script by passing it as argv[1].
+   * However, the Electron binary does not support an entrypoint, leading Cypress to think it's being opened in global mode (no args) when this fn is called.
+   * Solution is to filter out the problematic function.
+   * TODO(webkit): do we want to run this cleanup script another way?
+   * @see https://github.com/microsoft/playwright/blob/7e2aec7454f596af452b51a2866e86370291ac8b/packages/playwright-core/src/utils/processLauncher.ts#L191-L203
```
lmiller1990, line 60: Ahh this must have been tricky to debug. Nice

lmiller1990:
I used a somewhat spammy test download site for my testing:

```js
describe('empty spec', () => {
  it('passes', () => {
    cy.visit('https://file-examples.com/index.php/sample-documents-download/sample-rtf-download/')
    cy.get('a').contains('Download sample rtf file').click()
  })
})
```

It did work - the file was downloaded. Due to the websites's jankiness, it somehow injected a Google Ad into the command log (wtf?). Also it shows an error:

image

lmiller1990 (approved):
Code seems fine, I did some testing - see my comment - I'm not sure if the second test sample is/how it is supposed to work. It works in a my real webkit browser, though - so I think it probably should?

tbiethman (approved):
Can confirm Cypress no longer relaunches after exiting in run mode.

packages/server/lib/browsers/webkit.ts
flotwig: Super tricky. The stdout/stderr isn't inherited and it doesn't pass `--inspect-brk` so it was like a ghost process. I eventually figured out what was going on by adding logging using `fs.createWriteStream` and `tail`'ing the file.

Issue #23652, opened 2022-09-01
alias returns undefined when there is a customized get command for the alias and I visit the page again


haipinggong:
### Current behavior

1. Get an element by using a customized command called cy.get2 chained by a children command, then save it as an alias: 'container'.
2. Visit the page again.
3. Call cy.get('@container').

Cypress throws an error saying the cy.get2 returns undefined.

![image](https://user-images.githubusercontent.com/20938090/187851325-d28c95e6-6fb7-4bcf-87be-9d59a8c1a0c7.png)

### Desired behavior

In this situation, cy.get should still get the same element as I get in step 1.
The script worked on 10.4.0.

### Test code to reproduce

https://github.com/haipinggong/cypress-test-tiny/tree/alias-bug

```
describe("page", () => {
  it("works", () => {
    cy.visit("https://example.cypress.io");
    cy.get("body").children(".container").as("container");
    cy.visit("https://example.cypress.io");

    cy.get("@container");
  });

  it("does not work", () => {
    // The test case works in Cypress 10.4.0
    // The test case does not work in Cypress 10.5.0+

    cy.visit("https://example.cypress.io");
    cy.get2("body").children(".container").as("container");
    cy.visit("https://example.cypress.io");

    cy.get("@container");
  });
});
```

support/command.js file:

```
Cypress.Commands.add("get2", (selector, ...options) => {
  return cy.get(selector, ...options);
});
```

### Cypress Version

10.5.0+

### Node version

16.15.0

### Operating System

Windows 11

emilyrohrbough:
@haipinggong Thank you for the reproducible example! I tried it out and it does appear that this is a regression in 10.5.0.

BlueWinds:
That is a lovely short little test demonstrating the issue, was able to figure it out almost immediately. Opened a PR to fix it.

Was a regression introduced during some of my work related to https://github.com/cypress-io/cypress/issues/7306, that went out in 10.5.0.

PR #23663, opened 2022-09-01
fix: Aliases properly retry commands from inside custom commands

BlueWinds:
- Closes https://github.com/cypress-io/cypress/issues/23652

### User facing changelog
Fix: Mixing custom commands and aliases now works again. Fixes regression introduced between 10.4.0 and 10.5.0.

### Additional details
When a custom command invoked cypress commands, these 'internal' commands' return values are properly used as the subject for future commands, via the 'subjectLinks' internal Cypress state.

However, as of 10.5.0, aliases were not respecting these links, meaning they would end up with 'undefined' for a subject instead of the proper one (which would come from replaying the 'internal' commands).

See the test added by this PR for a more concrete example.

### Steps to test
Added automated test demonstrating the issue. Also see the repro in https://github.com/cypress-io/cypress/issues/23652 for external validation.

### PR Tasks

### DIFF
--- a/packages/driver/cypress/e2e/commands/aliasing.cy.js
+++ b/packages/driver/cypress/e2e/commands/aliasing.cy.js
@@ -101,6 +101,15 @@ describe('src/cy/commands/aliasing', () => {
           expect(this.input).to.eq(obj)
         })
       })
+
+      it('retries previous commands invoked inside custom commands', () => {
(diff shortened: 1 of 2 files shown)


emilyrohrbough (approved):

rachelruderman (approved):
Excelente! Tested locally with the OG issue and worked like a charm 🌟

Issue #22587, opened 2022-06-29
Callback form of should chained off cy.window receives null when retried


david-sabata:
### Current behavior

My goal is to have a retriable check of a `window` property that my app exposes. To achieve that, I'm using [the callback form](https://docs.cypress.io/guides/core-concepts/retry-ability#Use-should-with-a-callback) of `should`.

The window object is correctly passed into the callback only for the first time. If the assertions fail and trigger a retry, the callback argument is `null`. (See the repro code)

Maybe I missed something, but I went through the docs and I don't think this is expected behavior.

<img width="418" alt="image" src="https://user-images.githubusercontent.com/305251/176432700-497b3a45-0513-4dbb-b8ee-7a3157bc9037.png">

### Desired behavior

The callback of `should` always receives the `window` object

### Test code to reproduce

```js
it('retries should on window object', () => {
    cy.window().should(win => {
      // this logs <window> on the first run, but null on the next
      cy.log('window =', win).then(() => {
        // failing to trigger a retry
        expect(true).to.equal(false)
      })
    })
})
```

### Cypress Version

10.3.0

david-sabata:
Possibly related to #8346 ?

emilyrohrbough:
@david-sabata The issue you have logged is indeed a duplicate of #8346 (thank you for linking!) The `cy.log()` has a bug because it logs twice.

With the example you linked out, because the `.then()` command fails, it won't actually trigger a retry.

I assume your example is a rough high-level example of the behavior you are trying to accomplish? For the `cy.window()` command to re-poll when the `.should()` command fails, you'll want to stick with sync assertions and avoid non-cypress commands.

BlueWinds:
Very interesting, thanks for the precise repro. I'm currently working in this area - how subjects are determined and how commands are retried - in relation to https://github.com/cypress-io/cypress/issues/7306.

The short version is that `should` is not really designed for this use case, and we don't provide anything that *is* designed for this use case either. ^^;; I'm aiming to address that in my work. I'm going to assign myself here so I remember to follow up on this specific case.

A temporary workaround would be to ensure that you don't have any cypress commands inside your `should` callback:

```
it.only('retries should on window object', () => {
      let count = 0
      cy.window().should(win => {
        count++
        if (count < 4) { throw new Error('foobar') }
        expect(win).to.haveOwnProperty('location')
      })
    })
```

This passes; it's only when the should callback contains other cypress commands that cy gets confused over what the subject should be during retries.

[...]

PR #23665, opened 2022-09-01
feat: _addQuery()

### LINKED ISSUE #23550: Detached DOM: Commands._addQuery()
### What would you like?

As part of the Detached DOM work, we need to implement "queries", which are commands that function in a different way from existing cypress commands - they are synchronous, idempotent, and aren't responsible for their own retry loop.

I'm starting work with `.get()` and `.as()` - these are some of the most complicated commands in Cypress (encompassing as they do "getting elements from the DOM" and aliases, which are strangely complex for a simple-sounding concept).

### Why is this needed?

Have to start somewhere. I started at the deep end, so that I can encounter any issues with my approach immediately rather than later in the process.

### Other

This PR is again the `develop` branch - unlike the previous iteration of this work, it should be safe to release to users as-is (not a breaking change).

BlueWinds:
- Closes https://github.com/cypress-io/cypress/issues/23550

### User facing changelog
This PR contains three user-facing changes:

- The addition of Cypress.Commands.addSelector(), in packages/driver/src/cypress/commands.ts and packages/driver/src/cypress/command_queue.ts. This currently an internal API - it will be added to the documentation and finalized as part of upcoming work.
- Migrating .get() to be the first selector.
- Moving Aliases to use concept of "subject chains" rather than re-running commands via the command queue, in packages/driver/src/cy/commands/querying/querying.ts and packages/driver/src/cy/commands/aliasing.ts.

### Additional details
TODO

### Steps to test
TODO

.get() should no longer be susceptible to Detached DOM errors. I intend to write additional tests around it, but please note the one I had to delete in within.cy.ts - it no longer fails, because addSelector() isn't susceptbile to that issue!

### PR Tasks

- [x] Have tests been added/updated? Some have been added, but I would like to add more before merging, especially around custom commands and selectors.
- [n/a] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? API is not public facing, and one query alone has no behavioral changes (they only start occurring when queries interact with each other)


packages/driver/cypress/e2e/commands/assertions.cy.js
```diff
@@ -820,23 +820,8 @@ describe('src/cy/commands/assertions', () => {
         done()
       })
 
-      cy.get('body').then(() => {
-        expect(cy.currentSubject()).to.match('body')
-      })
-    })
-
-    it('sets type to child when subject matches', (done) => {
```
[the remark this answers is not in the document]
flotwig: `log:added` is in our public API in case anyone reading this was wondering: https://docs.cypress.io/api/events/catalog-of-events#Cypress-Events But I agree this is extremely unlikely to cause breakage.

packages/driver/src/cy/aliases.ts
```diff
@@ -16,16 +20,19 @@ const aliasDisplayName = (name) => {
...
 
-    const remoteSubject = cy.getRemotejQueryInstance(subject)
-
-    ctx[alias] = remoteSubject ?? subject
+    Object.defineProperty(ctx, alias, {
+      configurable: true,
+      get () {
+        return $utils.getSubjectFromChain(aliasObj.subjectChain, cy)
+      },
+    })
```
flotwig, line 35: Won't this break if a user is enumerating aliases on `this`? Or any use case where the user does `.as('foo')` and then later does `this.foo = ...`?

Like this: image

packages/driver/src/cy/commands/querying/querying.ts
```diff
@@ -21,6 +160,21 @@ interface InternalContainsOptions extends Partial<Cypress.Loggable & Cypress.Tim
...
+   *
+   * This is because other commands in the driver rely on the original .get() implementation, via
+   * `cy.now('get', selector, getOptions)`.
+   *
+   * The key is that cy.now() relies on `cy.commandFns[name]` - which addAll() sets, but overwriteQuery() does not.
+   *
+   * The upshot is that any test that relies on `cy.get()` is using the query-based implementation, but various
+   * driver commands have access to the original implementation of .get() via cy.now(). This is a temporary state
+   * of affairs while we refactor other commands to also be queries - we'll eventually be able to delete this
+   * original version of .get() entirely.
```
flotwig, line 176: Yeesh..... yes please. Glad you're getting to the bottom of this.

packages/driver/src/cypress/command.ts
```diff
@@ -33,9 +33,39 @@ export class $Command {
...
+        log.snapshot()
+      }
+    })
+  }
+
   finishLogs () {
-    // finish each of the logs we have
-    return _.invokeMap(this.get('logs'), 'finish')
+    // TODO: Investigate whether or not we can reuse snapshots between logs
+    // that snapshot at the same time
```
flotwig, line 46: Interesting idea. How would this work? My first thought is use a MutationObserver to prevent duplicating snapshots when nothing has changed in the DOM?

BlueWinds: I was thinking about doing it a little bit less intelligently, and just noting if we're taking multiple snapshots synchronously - in a couple of places we do `command.get('logs').forEach()` sorts of things, where we iterate through commands and / or logs and snapshot all of them.

If we're looping like that without a chance for external JS code to execute, then we know the snapshots can be shared. Mutation observer would be a more intelligent way to do it - just keep a pointer to the most recent snapshot, and invalidate it whenever the DOM mutates.

Dunno, just a thing I was thinking about as I looked at this code. It's come up a couple of times (with Emily looking at memory usage, and also Brian brought up the idea at one point).

packages/driver/src/cy/aliases.ts
BlueWinds: Yeah; I actually worked around this in one of our very own tests, where we have `.as('win'); cy.then(() => { this.win = ...override the alias we just made... })`.

This particular change isn't necessary - I could leave it as 'set on context once when command executes', but it seemed like an improvement. Setting aliases as static properties on the context suffers from the same "can be out of date" issue as any other cypress command.

packages/driver/src/cypress/command.ts
lmiller1990: Also interested in solving the "snapshots are big and memory hungry" problem. Another dumb way (I guess someone thought of this and dismissed it for reasons I hope I'm about to learn) is just to serialize the snapshots and compare them before/after, eg [using XMLSerializer](https://developer.mozilla.org/en-US/docs/Web/API/XMLSerializer).

As an aside, when we talk about memory usage, is the issue primarily **browser** memory usage (this is my understanding). If so, the other solution I've seen floated (also talked about) would just be saving them in some database (sqlite?) on the server and shuttling them back/forth as needed via a websocket. 

The obvious benefit of this is less complexity - just consumes a lot more data. Potential positive would be you can keep the snapshots for use later on (could be useful for many cool things, like replaying something that happened on CI in the Cloud or locally...).

lmiller1990 (commented):
Some early comments, I will play with this for a bit and probably make some more, I'm still internalizing things.

packages/driver/src/cy/commands/querying/querying.ts
```diff
@@ -156,7 +310,8 @@ export default (Commands, Cypress, cy, state) => {
       }
 
       if (aliasObj) {
-        let { subject, alias, command } = aliasObj
+        let { alias, command } = aliasObj
+        let subject = $utils.getSubjectFromChain(aliasObj.subjectChain, cy)
```
lmiller1990, line 314: I was trying to find out where this was used so I pushed a commit throwing an error here: https://github.com/cypress-io/cypress/compare/issue-7306-addSelector-redux...lmiller1990/issue-7306-addSelector-redux?expand=1

Surprisingly only tests related to `type()` broke - is this what you'd expect? Why would this be case?

CI: https://app.circleci.com/pipelines/github/cypress-io/cypress?branch=lmiller1990%2Fissue-7306-addSelector-redux

lmiller1990 (commented):
I left some more questions/comments :thumbsup:

packages/driver/src/cy/aliases.ts
flotwig: Would something like this work?

```suggestion
    let hasSetValue = false
    let ctxAliasValue: any = undefined
    Object.defineProperty(ctx, alias, {
      configurable: true,
      enumerable: true,
      get () {
        return hasSetValue ? ctxAliasValue : $utils.getSubjectFromChain(aliasObj.subjectChain, cy)
      },
      set (value) {
        hasSetValue = true
        ctxAliasValue = value
      }
    })
```

I also agree the spirit of this change is an improvement in the direction of detached DOM, but I feel like this should be removed for now, given the high potential to cause breakage. And we can release this improvement in a breaking version.

BlueWinds: I'll remove this from this PR and add it to the next one I'm working on (which will target 11.x rather than develop).

Good thought about accounting for overwrites, will include something to that effect.

packages/driver/src/cypress/utils.ts
```diff
@@ -399,4 +399,28 @@ export default {
   isPromiseLike (ret) {
     return ret && _.isFunction(ret.then)
   },
+
+  /* Given a chain of functions, return the actual subject. `subjectChain` might look like any of:
+   * [<input>]
+   * ['foobar', f()]
+   * [undefined, f(), f()]
+   */
+  getSubjectFromChain (subjectChain, cy) {
```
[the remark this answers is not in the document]
BlueWinds: Added some types to the PR. SubjectChain looks like so:

```
type QueryFunction = (any) => any
export type SubjectChain = [any, ...QueryFunction[]];
```

`any` in this context is a complete description - any value is valid for the first entry, including `undefined`, and QueryFunctions can take anything (one argument) and return anything (one argument).

```
Commands.addQuery (name: string, fn: () => QueryFunction)
```

Also added these types: a query is a function that returns a QueryFunction (department of redundancy department is redundant :D ).

packages/driver/src/cypress/command.ts
BlueWinds: I'd prefer to move in the opposite direction, to shuttle *less* over the air between app <-> server, rather than more.

Serializing is expensive, even if we don't retain it in memory. Repeatedly snapshotting and throwing away the snapshots if they're identical is a great way to keep CPU manufacturers in business. :P

packages/driver/src/cy/commands/querying/querying.ts
```diff
@@ -21,6 +160,21 @@ interface InternalContainsOptions extends Partial<Cypress.Loggable & Cypress.Tim
...
+   * via Commands.overwriteQuery(), which overwrites it.
+   *
+   * This is because other commands in the driver rely on the original .get() implementation, via
+   * `cy.now('get', selector, getOptions)`.
+   *
+   * The key is that cy.now() relies on `cy.commandFns[name]` - which addAll() sets, but overwriteQuery() does not.
+   *
+   * The upshot is that any test that relies on `cy.get()` is using the query-based implementation, but various
+   * driver commands have access to the original implementation of .get() via cy.now(). This is a temporary state
+   * of affairs while we refactor other commands to also be queries - we'll eventually be able to delete this
```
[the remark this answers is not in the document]
BlueWinds: The key is that that `Commands.add()` registers the function in `cy.commandFns[name]`, while `addQuery` doesn't. So we register it as a command, creating  `cy.get()` and `cy.commandFns.get`, then register it as a query, overriding `cy.get()`.

https://github.com/cypress-io/cypress/pull/23791/files#diff-231d74866a51e1411d172ab6ca0658f7fa712e9676f216ba797c544c47dcbf94 contains the actual fix - registering queries separately (without a wrapping promise) and then updates our other commands to use `cy.now('get')` properly when .get is a query - but I didn't have that ready when I made this PR, and I think it makes sense to merge in separately regardless (since it's a breaking change, needs to go in 11.x branch and not develop).

packages/driver/src/cy/commands/querying/querying.ts
BlueWinds: It's because the type() tests invoke `cy.getAll()`, which a custom command registered in `support/utils.ts` and uses `cy.now('get', '@stuff')` - which is the only way to hit this block of code.

Normal users use `cy.get('@foo')` (using the selector), and most of our commands use `cy.get('button')` (which isn't an alias and therefore misses this code block about aliases). Only the combination of aliasing and .now() can end up here, and .type() seems to be the only place in our code base that does so. :woman_shrugging:

flotwig (approved):

mikelhamer:
So hype for this fix!!! Keep up the great work :)

packages/driver/src/cypress/utils.ts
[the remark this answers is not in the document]
lmiller1990: I think it's still useful to know that something returns itself, even if it feels a bit redundant. Eventually it would be nice to support something like

```ts
cy.get<HTMLButtonElement>('#button'). [only show relevant methods to HTMLButtonElement here]
```

Gotta start somewhere.

packages/driver/src/cy/commands/querying/querying.ts
lmiller1990: Hm ok - we will eventually refactor `type` to not go here, then we can delete this block (only used for one thing, which is `type`, which is hard to know by looking at this)?

lmiller1990 (approved):
Looks good, responded to a couple more things for my own clarification - I wouldn't consider any of my existing comments to be outright blockers at this point. I also played with the branch locally. I didn't see any regressions. I understand the intentional of changes now, (mostly) how it works, and where it's leading. Great!!

packages/driver/src/cy/commands/querying/querying.ts
BlueWinds: Yeah. This entire block - the whole `add('get')` section - will be removed in my next PR, which I'm currently working on and should have ready today or tomorrow. This is just a messy temporary state because I wanted to limit the scope of things going in at once.

packages/driver/src/cypress/utils.ts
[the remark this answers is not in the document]
BlueWinds: Individual commands can potentially be more strictly typed - but I think you'll quickly find that most Cypress commands are so versatile that their types end up horribly messy. `.get()`, for example, can return an `HTMLElement`... or the result of an alias, which can be literally anything. '.first()`, which looks like a method for traversing HTML collections, can also accept `any[]`, or `JQuery<any>` as a previous subject. :woman_shrugging:

BlueWinds:
Spent a bit of time profiling this change. I tried several different tests, but ultimately settled on this one, executed in open mode:

```
it('should go fast', () => {
  // packages/driver/cypress/fixtures/dom.html

  for (let i = 0; i < 500; i++) {
    cy.get('#button').as('a' + i)
  }
  for (let i = 0; i < 500; i++) {
    cy.get('@a' + i).should('exist')
  }
})
```

This results in a huge number of log messages, DOM snapshots, and command executions. I ran this test in Firefox and recorded the time from the runner, and memory usage from the Firefox web process 15s after the test had ended (when FF had time to run several garbage collection cycles).

`develop` branch:
- 68s, 1021mb
- 70s, 1400mb
- 74s, 1600mb

`issue-7306-addSelector-redux` branch:
- 67s, 840mb
- 69s, 1200mb
- 70s, 1600mb

The execution times and memory usage are comparable; At the very least this branch isn't slowing anything down significantly; it might be speeding things up slightly, but the difference is small enough it could be luck.

Note the climbing memory usage across runs - I'm pretty sure we have a memory leak between tests, and this is something I want to look into - but it's preexisting, not introduced by this PR.

I also ran this test with most logs disabled:
```
  it('should go fast', () => {
    for (let i = 0; i < 500; i++) {
      cy.get('#button', {log: false}).as('a' + i)
    }
    for (let i = 0; i < 500; i++) {
      cy.get('@a' + i, {log: false}).should('exist')
    }
  })
```

In this case, runs were consistently 9s regardless of branch, and memory usage went roughly 500 -> 550 -> 600mb. This test does suggest that our logs are extremely resource intensive, which is something *else* I want to look into, unrelated to the detached DOM work.

BlueWinds:
Also went and ran this test in `run` mode, and the results are consistent.

`develop`:
- 96491ms
- 95579ms

`issue-7306-addSelector-redux`:
- 94935ms
- 94621ms

Having done this several times in a couple of different ways, I'm now confident that `.get()` as a query *is* faster. Only by a small margin, but it's consistently faster across runs and modes.

packages/driver/src/cypress/utils.ts
```diff
@@ -399,4 +403,28 @@ export default {
...
     return ret && _.isFunction(ret.then)
   },
+
+  /* Given a chain of functions, return the actual subject. `subjectChain` might look like any of:
+   * [<input>]
+   * ['foobar', f()]
+   * [undefined, f(), f()]
+   */
+  getSubjectFromChain (subjectChain: SubjectChain, cy) {
+    // If we're getting the subject of a previous command, then any log messages have already
```
emilyrohrbough, line 413: Shouldn't the existing log messages be updated with the new subject value so it's reflected correctly in the console log when clicking a command in the reporter?

packages/driver/src/cypress/error_messages.ts
```diff
@@ -865,6 +861,10 @@ export default {
       message: '`Cypress.Commands.add()` cannot create a new command named `{{name}}` because that name is reserved internally by Cypress.',
       docsUrl: 'https://on.cypress.io/custom-commands',
     },
+    invalid_new_query: {
+      message: '`Cypress.Commands.addQuery()` is used to create new query commands, but `{{name}}` is an existing Cypress command or is reserved internally by Cypress.\n\n If you want to override an existing command, use `Cypress.Commands.overrideQuery()` instead.',
+      docsUrl: 'https://on.cypress.io/custom-commands',
+    },
```
emilyrohrbough: addCommand will through the `reserved_command` error  if the command is reservered internally. Shouldn't we do the same for addQuery() so users understand how to handle this error?

packages/driver/src/cypress/error_messages.ts
```diff
@@ -865,6 +861,10 @@ export default {
       message: '`Cypress.Commands.add()` cannot create a new command named `{{name}}` because that name is reserved internally by Cypress.',
       docsUrl: 'https://on.cypress.io/custom-commands',
     },
+    invalid_new_query: {
```
emilyrohrbough, line 864:
```suggestion
    invalid_add_query: {
```

packages/driver/src/cypress/error_messages.ts
```diff
@@ -1803,6 +1834,18 @@ export default {
   },
 
   subject: {
+    is_null (obj) {
+      return stripIndent`\
+        ${cmd(obj.name)} failed because it requires a subject, but didn't receive one. This usually happens because you're trying to use a child query as if it were a parent.`
```
emilyrohrbough: I think we should consider aligning this verbiage with https://github.com/cypress-io/cypress/blob/issue-7306-addSelector-redux/packages/driver/src/cypress/error_messages.ts#L874 ('oops, looked like ...') and provide pusedo code on how to fix this. If someone isn't as familiar with parent vs child command this likely won't make sense
```suggestion
        ${cmd(obj.name)} failed because it requires a subject, but it didn't receive one. This usually happens because you're trying to use a child query command as if it were a parent command`
```

packages/driver/src/cypress/error_messages.ts
```diff
@@ -1803,6 +1834,18 @@ export default {
...
+    },
+    not_null (obj) {
```
emilyrohrbough: I assume this only occurs when a command cannot be chained? If so, using that verbiage

packages/driver/src/cypress/cy.ts
```diff
@@ -244,13 +246,14 @@ export class $Cy extends EventEmitter2 implements ITimeouts, IStability, IAssert
...
     this.addCommand = this.addCommand.bind(this)
+    this.addQuery = this.addQuery.bind(this)
     this.now = this.now.bind(this)
-    this.replayCommandsFrom = this.replayCommandsFrom.bind(this)
     this.onBeforeAppWindowLoad = this.onBeforeAppWindowLoad.bind(this)
     this.onUncaughtException = this.onUncaughtException.bind(this)
     this.setRunnable = this.setRunnable.bind(this)
     this.cleanup = this.cleanup.bind(this)
     this.setSubjectForChainer = this.setSubjectForChainer.bind(this)
+    this.currentSubject = this.currentSubject.bind(this)
```
emilyrohrbough, line 254: Do we need to add this to the global object?  If so, we should double check this is getting set to null on a refresh / when cypress is town down when navigating between specs in open mode so the browser will gc.

packages/driver/src/cypress/cy.ts
```diff
@@ -343,13 +346,16 @@ export class $Cy extends EventEmitter2 implements ITimeouts, IStability, IAssert
     this.ensureNotHiddenByAncestors = ensures.ensureNotHiddenByAncestors
     this.ensureExistence = ensures.ensureExistence
     this.ensureElExistence = ensures.ensureElExistence
+    this.ensureNull = ensures.ensureNull
+    this.ensureNotNull = ensures.ensureNotNull
```
emilyrohrbough: same question, do we need to add on the global?

packages/driver/src/cy/commands/misc.ts
```diff
@@ -24,22 +23,6 @@ export default (Commands, Cypress, cy, state) => {
...
-      if (state('current').get('injected')) {
-        const restoreCmdIndex = state('index') + 1
-
-        cy.queue.insert(restoreCmdIndex, $Command.create({
-          args: [cy.currentSubject()],
-          name: 'log-restore',
-          fn: (subject) => subject,
-        }))
-
-        state('index', restoreCmdIndex)
```
emilyrohrbough, line 40: does this change fix the https://github.com/cypress-io/cypress/issues/22587?

packages/driver/src/cypress/command.ts
brian-mann: You'd track the diffs using MutationObserver and only serialize the patches that describe the minimum set of changes, and then rebuild them on demand to a particular state / time

packages/driver/cypress/e2e/commands/agents.cy.js
```diff
@@ -323,10 +323,6 @@ describe('src/cy/commands/agents', () => {
           expect(cy.state('aliases').myStub).to.exist
         })
 
-        it('stores the agent as the subject', function () {
```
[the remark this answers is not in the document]
emilyrohrbough: @BlueWinds Can you link that test?

packages/driver/cypress/e2e/commands/aliasing.cy.js
```diff
@@ -34,11 +26,13 @@ describe('src/cy/commands/aliasing', () => {
...
-    it('stores the resulting subject as the alias', () => {
-      const $body = cy.$$('body')
-
+    it('stores the resulting subject chain as the alias', () => {
       cy.get('body').as('b').then(() => {
-        expect(cy.state('aliases').b.subject.get(0)).to.eq($body.get(0))
+        const { subjectChain } = cy.state('aliases').b
+
+        expect(subjectChain.length).to.eql(2)
+        expect(subjectChain[0]).to.be.undefined
```
emilyrohrbough, line 34: Why is the first subject chain undefined? Shouldn't the only subject chain be `.get()`? Shouldn't we assert the second subject chain is actually `.get()`?

packages/driver/cypress/e2e/commands/agents.cy.js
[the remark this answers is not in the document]
BlueWinds: The rest of this file is asserting on the output of the alias (eg, that we can store and retrieve aliases). It's not a specific test, but the whole file. :woman_shrugging:

packages/driver/cypress/e2e/commands/aliasing.cy.js
BlueWinds: Subject chains always start with a "primitive" value, and then append any number of functions that then chain off this value.

Consider these two cases:

```
cy.get('button').contains('Login')

cy.get('button').click().contains('Login')
```

In the first case, we make a subject chain like `[undefined, getFn(), containsFn()]`. In the second one, we start with that same value, but .click() *replaces* the subject chain rather than adds to it - so we end up with `[<button>, containsFn()]`.

I start subject chains with 'undefined' so that they always have a consistent shape: a primitive value, followed by 0 or more functions that are applied to it.

packages/driver/src/cy/commands/misc.ts
BlueWinds: No, but I have https://github.com/cypress-io/cypress/pull/23734 open separately to address that.

Basically, their use is completely unsupported, and in 11.x we'll throw an error rather than give them weird behavior.

packages/driver/src/cypress/error_messages.ts
BlueWinds: I'm mirroring the name of `invalid_new_command` here - I'd be fine changing them both to be `add`, but I think they should stay the same (since they're mirror error messages).

packages/driver/src/cypress/error_messages.ts
BlueWinds: Done; Also added a pair of tests around the errors thrown by adding queries.

packages/driver/src/cypress/error_messages.ts
BlueWinds: Since ensureNull  / ensureNotNull aren't actually used in this PR - and in fact I haven't found a good use for them yet - I went ahead and removed them entirely, along with the associated error messages. Leftover stuff from earlier iterations on this PR that I hadn't cleaned up.

packages/driver/src/cypress/error_messages.ts
BlueWinds: As above - this is actually unused, and just some leftover code I forgot to clean up.

packages/driver/src/cypress/utils.ts
BlueWinds: I don't think we do. Imagine if you do something like this:

```
cy.get('button').as('loginButton').click()

...stuff happens, the page navigates, more commands are run...

cy.get('@loginButton').should('not.exist')
```

The command log would look like:

```
- get
- click

...other stuff...

- get
  - should
```
We wouldn't want to update the first log message with the now non-existent button while executing `.get('@loginButton')` - we'd want to leave it intact, as it was when the command resolved, even if we later invoke get getFn() (to get the subject for the second .get()).

packages/driver/cypress/e2e/commands/aliasing.cy.js
BlueWinds: For the second part - yes, yes we should assert that the second entry is `.get()`. Added an assertion.

packages/driver/src/cypress/cy.ts
BlueWinds: Removed these two entirely.

packages/driver/src/cypress/cy.ts
BlueWinds: I added them here to follow the existing pattern where all methods on `cy` are bound to `this`.

I don't think it makes any sense to add tests here - binding a method to ensure `this` is always consistent isn't something that requires cleaning up between tests.

packages/driver/src/cypress/command.ts
```diff
@@ -12,7 +14,7 @@ export class $Command {
       // the id prefix needs to be unique per origin, so there are not
       // collisions when commands created in a secondary origin are passed
       // to the primary origin for the command log, etc.
-      attrs.id = _.uniqueId(`cmd-${window.location.origin}-`)
+      attrs.id = `${attrs.chainerId}-cmd-${idCounter++}`
```
BlueWinds, line 17: Discussing chainer and command IDs with @emilyrohrbough in slack, we came to the conclusion that it would be very nice for debugging if it were clear which commands belong to which chainers; thus this change (and the associated one above in `packages/driver/src/cypress/chainer.ts`).

This results in command and chainer IDs that look like this, where it's easy to tell at a glance what belongs to what. It also eliminates the confusion that chainerIds weren't sequential - `_.uniqueId()` shares only a single counter, so you'd get chainer-2, cmd-3, chainer-4, etc.

This is only to make IDs easier to reason about and debug, no user-visible impact.

image

BlueWinds:
I've pushed some additional changes to this branch since the last review, merging in develop and fixing some subsequent merge conflicts and test failures.

Re-requesting review from those who've approved, and thank you for your patience while I got this into shape again!

packages/driver/src/cy/commands/querying/root.ts
```diff
@@ -1,40 +1,25 @@
-import _ from 'lodash'
-import type { Log } from '../../../cypress/log'
-
-interface InternalRootOptions extends Partial<Cypress.Loggable & Cypress.Timeoutable> {
-  _log?: Log
-}
-
 export default (Commands, Cypress, cy, state) => {
```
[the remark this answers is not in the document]
lmiller1990: Good candidate for a post 11.x blog post

packages/driver/src/cypress/command_queue.ts
```diff
@@ -163,7 +198,21 @@ export class CommandQueue extends Queue<$Command> {
       let ret
       let enqueuedCmd
 
+      // Queries can invoke other queries - they are synchronous, and get added to the subject chain without
+      // issue. But they cannot contain commands, which are async.
+      // This callback watches to ensure users don't try and invoke any commands while inside a query.
```
lmiller1990, line 203: Will be interested to review the documentation PR associated with queries (we will need to make sure how they work and requirements/limitations are clear to users, so they can author their own).

packages/driver/src/cypress/cy.ts
```diff
@@ -831,68 +829,76 @@ export class $Cy extends EventEmitter2 implements ITimeouts, IStability, IAssert
...
+      // this is the first call on cypress
+      // so create a new chainer instance
+      const chainer = new $Chainer(cy.specWindow)
+
+      if (cy.state('chainerId')) {
+        cy.linkSubject(chainer.chainerId, cy.state('chainerId'))
+      }
 
-      const chainerId = this.state('chainerId')
+      const userInvocationStack = $stackUtils.captureUserInvocationStack(cy.specWindow.Error)
```
lmiller1990, line 884: I am not fully understanding what's going on here - why does `_addQuery` need to grab the userInvocationStack? Isn't this for error handling? Do we expect that here when they initially call `useQuery` (or is this to handle the cases where they used `useQuery` incorrectly?

lmiller1990 (approved):
I ran the tests and poked around, getting a better understanding of all this. I'll approve since the code works and we've been over it a few times, but I do have one last question. It's more for my understanding.

packages/driver/src/cypress/command_queue.ts
BlueWinds: After discussing this with Brian, we're holding off making the API public, which is why the PR switched to `_addQuery`, indicating an internal method. Documentation will wait until the end of the project, when we're ready to remove the underscore and have worked out all the kinks.

packages/driver/src/cypress/cy.ts
BlueWinds: Good question, it took me a while to figure out how stack traces work in Cypress.

This function is literally cy.get(): A few lines above, we have `cy[name] = function (...args) { ...this function... }`, so the stack trace here is the transition point between 'user test code' and 'internal Cypress code'. We capture it so that later, when the command actually executes and possibly throws an error, we can point out to users where in *their* code an error occurred.

Here are a couple of slightly simplified stack traces to show why this is the best place. With the simple command `cy.get('body')`, the stack trace at this point looks like:

```
    name                       cy.ts:797  <-- This anonymous function, which is literally `cy.get`
    js                         querying.cy.js:16   <-- The code in the test file calling cy.get('body')
    __stackReplacementMarker   cy.ts:85
    ...lots of internal Cy stuff...
```

Here's what a stack trace looks like when an error is thrown inside cy.get():

image

Any error in the command is thrown much later, as part of command queue execution - the location where cy.get() was called *in the user's code* is long lost, and this stack trace contains no useful information for the user trying to figure out how they wrote their test wrong / how their application is failing.

Thus, for queries and commands both, we capture the *user* stack trace when the command is added to the queue, so that we can display this if the command throws an error later while executing.

(none of this is query specific; commands work the same way with regards to userStackTraces)

BlueWinds:
Currently failing tests are flake; cypress-schematic will be fixed in develop soon (thanks Ryan!), and the driver-integration-tests-electron-experimentalSessionAndOrigin is known flake from cy.origin + cy.go().

packages/driver/src/cypress/command_queue.ts
lmiller1990: Sorry I should clarify; I'll be interested in seeing it when it's available (not saying we should have it done now - definitely can come later).

packages/driver/src/cypress/cy.ts
lmiller1990: Ok, that explains how this works (had a vague idea, but this clarified). Last question would be why was this added here specifically in this PR? Did we move the logic here from somewhere else? Or did the `useQuery` addition require this to be added here?

BlueWinds: It's added here because `_addQuery` is a mirror to the existing `addCommand` - you can see the exact same line,

`const userInvocationStack = $stackUtils.captureUserInvocationStack(cy.specWindow.Error)`

up earlier in this file, line 791 inside addCommand.

`_addQuery` and `addCommand` are fairly similar in a lot of ways, but the details differ enough that I didn't want to mesh in a bunch of `if (isQuery) {...}` code branches into addCommand, and decided that some code duplication was the lesser of two evils.

PR #23689, opened 2022-09-06
fix: Add polyfill for Chrome V65+

### LINKED ISSUE #23509: cypress 10.4+ in chrome 70 breaks with globalThis
### Current behavior

Cypress run on chrome 70 cannot start with error: "Uncaught ReferenceError: globalThis is not defined"

### Desired behavior

cypress run with chrome 70

### Test code to reproduce

Any test with cypress 10.4+ run on chrome 70 should reproduce.

### Cypress Version

10.4+

### Node version

v16.5.0

### Operating System

linux

### Debug Logs

```shell
root@me-cd-test-k8s-3137-446pp-dqxkt-wgg0b:/home/jenkins/agent/workspace/ME-CD/test-k8s# DEBUG=cypress:launcher:* node test.js cypress/prepare/role.cy.js /tmp/test-result/3137/admin
✔  Verified Cypress! /root/.cache/Cypress/10.4.0/Cypress
[16160:0823/155424.906310:ERROR:zygote_host_impl_linux.cc(263)] Failed to adjust OOM score of renderer with pid 16344: Permission denied (13)
[16344:0823/155425.028386:ERROR:gpu_memory_buffer_support_x11.cc(44)] dri3 extension not supported.
2022-08-23T07:54:28.962Z cypress:launcher:detect detecting if the following browsers are present [ { name: 'chrome', family: 'chromium', channel: 'stable', displayName: 'Chrome', versionRegex: /Google Chrome (\S+)/m, binary: [ 'google-chrome', 'chrome', 'google-chrome-stable' ], minSupportedVersion: 64 }, { name: 'chromium', family: 'chromium', channel: 'stable', displayName: 'Chromium', versionRegex: /Chromium (\S+)/m, binary: [ 'chromium-browser', 'chromium' ], minSupportedVersion: 64 }, { name: 'chrome', family: 'chromium', channel: 'beta', displayName: 'Chrome Beta', versionRegex: /Google Chrome (\S+) beta/m, binary: 'google-chrome-beta', minSupportedVersion: 64 }, { name: 'chrome', family: 'chromium', channel: 'canary', displayName: 'Canary', versionRegex: /Google Chrome Canary (\S+)/m, binary: 'google-chrome-canary', minSupportedVersion: 64 }, { name: 'firefox', family: 'firefox', channel: 'stable', displayName: 'Firefox', versionRegex: /^Mozilla Firefox ([^\sab]+)$/m, binary: 'firefox', minSupportedVersion: 86 }, { name: 'firefox', family: 'firefox', channel: 'dev', displayName: 'Firefox Developer Edition', versionRegex: /^Mozilla Firefox (\S+b\S*)$/m, binary: [ 'firefox-developer-edition', 'firefox' ], minSupportedVersion: 86 }, { name: 'firefox', family: 'firefox', channel: 'nightly', displayName: 'Firefox Nightly', versionRegex: /^Mozilla Firefox (\S+a\S*)$/m, binary: [ 'firefox-nightly', 'firefox-trunk' ], minSupportedVersion: 86 }, { name: 'edge', family: 'chromium', channel: 'stable', displayName: 'Edge', versionRegex: /Microsoft Edge (\S+)/m, binary: [ 'edge', 'microsoft-edge' ], minSupportedVersion: 79 }, { name: 'edge', family: 'chromium', channel: 'canary', displayName: 'Edge Canary', versionRegex: /Microsoft Edge Canary (\S+)/m, binary: 'edge-canary', minSupportedVersion: 79 }, { name: 'edge', family: 'chromium', channel: 'beta', displayName: 'Edge Beta', versionRegex: /Microsoft Edge Beta (\S+)/m, binary: 'edge-beta', minSupportedVersion: 79 }, { name: 'edge', family: 'chromium', channel: 'dev', displayName: 'Edge Dev', versionRegex: /Microsoft Edge Dev (\S+)/m, binary: [ 'edge-dev', 'microsoft-edge-dev' ], minSupportedVersion: 79 } ]
Missing baseUrl in compilerOptions. tsconfig-paths will be skipped
2022-08-23T07:54:43.180Z cypress:launcher:detect detecting if the following browsers are present [ { name: 'chrome', family: 'chromium', channel: 'stable', displayName: 'Chrome', versionRegex: /Google Chrome (\S+)/m, binary: [ 'google-chrome', 'chrome', 'google-chrome-stable' ], minSupportedVersion: 64 }, { name: 'chromium', family: 'chromium', channel: 'stable', displayName: 'Chromium', versionRegex: /Chromium (\S+)/m, binary: [ 'chromium-browser', 'chromium' ], minSupportedVersion: 64 }, { name: 'chrome', family: 'chromium', channel: 'beta', displayName: 'Chrome Beta', versionRegex: /Google Chrome (\S+) beta/m, binary: 'google-chrome-beta', minSupportedVersion: 64 }, { name: 'chrome', family: 'chromium', channel: 'canary', displayName: 'Canary', versionRegex: /Google Chrome Canary (\S+)/m, binary: 'google-chrome-canary', minSupportedVersion: 64 }, { name: 'firefox', family: 'firefox', channel: 'stable', displayName: 'Firefox', versionRegex: /^Mozilla Firefox ([^\sab]+)$/m, binary: 'firefox', minSupportedVersion: 86 }, { name: 'firefox', family: 'firefox', channel: 'dev', displayName: 'Firefox Developer Edition', versionRegex: /^Mozilla Firefox (\S+b\S*)$/m, binary: [ 'firefox-developer-edition', 'firefox' ], minSupportedVersion: 86 }, { name: 'firefox', family: 'firefox', channel: 'nightly', displayName: 'Firefox Nightly', versionRegex: /^Mozilla Firefox (\S+a\S*)$/m, binary: [ 'firefox-nightly', 'firefox-trunk' ], minSupportedVersion: 86 }, { name: 'edge', family: 'chromium', channel: 'stable', displayName: 'Edge', versionRegex: /Microsoft Edge (\S+)/m, binary: [ 'edge', 'microsoft-edge' ], minSupportedVersion: 79 }, { name: 'edge', family: 'chromium', channel: 'canary', displayName: 'Edge Canary', versionRegex: /Microsoft Edge Canary (\S+)/m, binary: 'edge-canary', minSupportedVersion: 79 }, { name: 'edge', family: 'chromium', channel: 'beta', displayName: 'Edge Beta', versionRegex: /Microsoft Edge Beta (\S+)/m, binary: 'edge-beta', minSupportedVersion: 79 }, { name: 'edge', family: 'chromium', channel: 'dev', displayName: 'Edge Dev', versionRegex: /Microsoft Edge Dev (\S+)/m, binary: [ 'edge-dev', 'microsoft-edge-dev' ], minSupportedVersion: 79 } ]

====================================================================================================

  (Run Starting)

  │ Cypress:        10.4.0                                                                         │
  │ Browser:        Chrome 70 (headless)                                                           │
  │ Node Version:   v16.5.0 (/usr/local/bin/node)                                                  │
  │ Specs:          1 found (role.cy.js)                                                           │
  │ Searched:       cypress/prepare/role.cy.js                                                     │


  Running:  role.cy.js                                                                      (1 of 1)
[cypress-log-to-output] Attempting to connect to Chrome Debugging Protocol
2022-08-23T07:54:44.983Z cypress:launcher:browsers launching browser { browser: { name: 'chrome', family: 'chromium', channel: 'stable', displayName: 'Chrome', version: '70.0.3538.77', path: 'google-chrome', minSupportedVersion: 64, majorVersion: 70, isHeadless: true, isHeaded: false }, url: 'about:blank' }
2022-08-23T07:54:44.985Z cypress:launcher:browsers spawning browser with args { args: [ 'about:blank', '--test-type', '--ignore-certificate-errors', '--start-maximized', '--silent-debugger-extension-api', '--no-default-browser-check', '--no-first-run', '--noerrdialogs', '--enable-fixed-layout', '--disable-popup-blocking', '--disable-password-generation', '--disable-single-click-autofill', '--disable-prompt-on-repos', '--disable-background-timer-throttling', '--disable-renderer-backgrounding', '--disable-renderer-throttling', '--disable-backgrounding-occluded-windows', '--disable-restore-session-state', '--disable-new-profile-management', '--disable-new-avatar-menu', '--allow-insecure-localhost', '--reduce-security-for-testing', '--enable-automation', '--disable-print-preview', '--disable-device-discovery-notifications', '--autoplay-policy=no-user-gesture-required', '--disable-site-isolation-trials', '--metrics-recording-only', '--disable-prompt-on-repost', '--disable-hang-monitor', '--disable-sync', '--disable-web-resources', '--safebrowsing-disable-download-protection', '--disable-client-side-phishing-detection', '--disable-component-update', "--simulate-outdated-no-au='Tue, 31 Dec 2099 23:59:59 GMT'", '--disable-default-apps', '--use-fake-ui-for-media-stream', '--use-fake-device-for-media-stream', '--disable-ipc-flooding-protection', '--disable-backgrounding-occluded-window', '--disable-breakpad', '--password-store=basic', '--use-mock-keychain', '--disable-dev-shm-usage', '--disable-gpu', '--no-sandbox', '--proxy-server=http://localhost:39355', '--headless', '--window-size=1280,720', '--force-device-scale-factor=1', '--remote-debugging-port=33382', '--remote-debugging-address=127.0.0.1', '--user-data-dir=/root/.config/Cypress/cy/production/browsers/chrome-stable/run-16160', '--disk-cache-dir=/root/.config/Cypress/cy/production/browsers/chrome-stable/run-16160/CypressCache' ] }
2022-08-23T07:54:45.198Z cypress:launcher:browsers chrome stderr: [0823/155445.197221:ERROR:gpu_process_transport_factory.cc(980)] Lost UI shared context.
2022-08-23T07:54:45.200Z cypress:launcher:browsers chrome stderr: DevTools listening on ws://127.0.0.1:33382/devtools/browser/13cdca87-00d4-4c07-903c-5064f790ef5a
2022-08-23T07:54:45.208Z cypress:launcher:browsers chrome stderr: [0823/155445.206777:ERROR:zygote_host_impl_linux.cc(259)] Failed to adjust OOM score of renderer with pid 16474: Permission denied (13)
[cypress-log-to-output] Connected to Chrome Debugging Protocol
2022-08-23T07:54:46.896Z cypress:launcher:browsers chrome stderr: [0823/155446.895003:INFO:CONSOLE(39990)] "Uncaught ReferenceError: globalThis is not defined", source: http://me-e2e-cypress10.me.magic.com/__/assets/index.23f3e8ba.js (39990)
```

### Other

globalThis is not supported in chrome 70, 10.4+ somehow use it directly in bundled code.

rockindahizzy:
- Closes #23509

### User facing changelog

- Fix regression causing incompatibility with Chrome 70 and older

### Steps to test

Try to run e2e tests (I used the kitchen sink app) using Chrome 70.
Prior to this change, nothing will render in the UI and tests will timeout.
After the fix, tests will run normally.
I can provide the chrome binary (macOS) if needed.

### PR Tasks

- na Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- na Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?

### DIFF
--- a/packages/app/vite.config.mjs
+++ b/packages/app/vite.config.mjs
@@ -32,7 +32,7 @@ export default makeConfig({
     }),
     Legacy({
       targets: ['Chrome > 64'],
-      modernPolyfills: true
+      modernPolyfills: ['es.global-this'],
     })
   ],
 })
--- a/packages/app/vite.config.mjs
+++ b/packages/app/vite.config.mjs
@@ -31,7 +31,7 @@ export default makeConfig({
       }],
     }),
     Legacy({
-      targets: ['chrome 65'],
+      targets: ['Chrome > 64'],
       modernPolyfills: true
     })
   ],
--- a/packages/app/vite.config.mjs
+++ b/packages/app/vite.config.mjs
@@ -33,6 +33,7 @@ export default makeConfig({
     Legacy({
       targets: ['Chrome >= 64', 'Firefox >= 86', 'Edge >= 79'],
       modernPolyfills: true,
+      renderLegacyChunks: false,
     })
   ],
 })
--- a/packages/app/vite.config.mjs
+++ b/packages/app/vite.config.mjs
@@ -2,6 +2,7 @@ import { makeConfig } from '../frontend-shared/vite.config.mjs'
 import Layouts from 'vite-plugin-vue-layouts'
 import Pages from 'vite-plugin-pages'
 import Copy from 'rollup-plugin-copy'
+import Legacy from '@vitejs/plugin-legacy'
 import { resolve } from 'path'
 
 export default makeConfig({
@@ -29,5 +30,9 @@ export default makeConfig({
         dest: 'dist',
       }],
     }),
+    Legacy({
+      targets: ['chrome 65'],
+      modernPolyfills: true
+    })
   ],
 })
--- a/packages/app/vite.config.mjs
+++ b/packages/app/vite.config.mjs
@@ -31,8 +31,8 @@ export default makeConfig({
       }],
     }),
     Legacy({
-      targets: ['Chrome > 64'],
-      modernPolyfills: ['es.global-this'],
+      targets: ['Chrome >= 64', 'Firefox >= 86', 'Edge >= 79'],
+      modernPolyfills: true,
     })
   ],
 })
--- a/browser-versions.json
+++ b/browser-versions.json
@@ -1,4 +1,5 @@
 {
   "chrome:beta": "105.0.5195.28",
-  "chrome:stable": "104.0.5112.101"
+  "chrome:stable": "104.0.5112.101",
+  "chrome:minimum": "64.0.3282.0"
 }
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -32,6 +32,7 @@
     "@types/faker": "5.5.8",
     "@urql/core": "2.4.4",
     "@urql/vue": "0.6.2",
+    "@vitejs/plugin-legacy": "^2.1.0",
     "@vitejs/plugin-vue": "2.2.4",
     "@vitejs/plugin-vue-jsx": "1.3.8",
     "@vueuse/core": "7.2.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2175,6 +2175,11 @@
   dependencies:
     regenerator-runtime "^0.13.4"
 
+"@babel/standalone@^7.18.13":
+  version "7.19.0"
+  resolved "https://registry.yarnpkg.com/@babel/standalone/-/standalone-7.19.0.tgz#23cbe261557a72dcc58ba56c3a2125944a251c4f"
+  integrity sha512-P8lYynBl10MYpgkhGEV1fOsGDE/6lQSYp+e2YvqUPMWHT4JaQrOkM49FizjswrRzJ6KZTgG6ZY90/1GYvAR3/Q==
+
 "@babel/template@^7.0.0", "@babel/template@^7.16.7", "@babel/template@^7.4.0", "@babel/template@^7.5.4", "@babel/template@^7.6.0", "@babel/template@^7.8.6":
   version "7.16.7"
   resolved "https://registry.yarnpkg.com/@babel/template/-/template-7.16.7.tgz#8d126c8701fde4d66b264b3eba3d96f07666d155"
@@ -7433,6 +7438,17 @@
   dependencies:
     "@urql/core" "^2.3.6"
     wonka "^4.0.14"
+
+"@vitejs/plugin-legacy@^2.1.0":
+  version "2.1.0"
+  resolved "https://registry.yarnpkg.com/@vitejs/plugin-legacy/-/plugin-legacy-2.1.0.tgz#0170c582f4cccc1a6b0391c258e6cee209542fd9"
+  integrity sha512-en3h0L7okBonSYKJx81bU8AVFPjSCiUSz8xUDAW8J0CxskfxSt/VJKbZO6G9yCVgZLywGoO8PNAfOQWVLUWZ2A==
+  dependencies:
+    "@babel/standalone" "^7.18.13"
+    core-js "^3.25.0"
+    magic-string "^0.26.2"
+    regenerator-runtime "^0.13.9"
+    systemjs "^6.12.4"
 
 "@vitejs/plugin-react@1.3.1":
   version "1.3.1"
@@ -12972,10 +12988,10 @@ core-js@^2.4.0, core-js@^2.4.1, core-js@^2.5.0:
   resolved "https://registry.yarnpkg.com/core-js/-/core-js-2.6.12.tgz#d9333dfa7b065e347cc5682219d6f690859cc2ec"
   integrity sha512-Kb2wC0fvsWfQrgk8HU5lW6U/Lcs8+9aaYcy4ZFc6DDlo4nZ7n70dEgE5rtR0oG6ufKDUnrwfWL1mXR5ljDatrQ==
 
-core-js@^3.5.0:
-  version "3.22.0"
-  resolved "https://registry.yarnpkg.com/core-js/-/core-js-3.22.0.tgz#b52007870c5e091517352e833b77f0b2d2b259f3"
-  integrity sha512-8h9jBweRjMiY+ORO7bdWSeWfHhLPO7whobj7Z2Bl0IDo00C228EdGgH7FE4jGumbEjzcFfkfW8bXgdkEDhnwHQ==
+core-js@^3.25.0, core-js@^3.5.0:
+  version "3.25.0"
+  resolved "https://registry.yarnpkg.com/core-js/-/core-js-3.25.0.tgz#be71d9e0dd648ffd70c44a7ec2319d039357eceb"
+  integrity sha512-CVU1xvJEfJGhyCpBrzzzU1kjCfgsGUxhEvwUV2e/cOedYWHdmluamx+knDnmhqALddMG16fZvIqvs9aijsHHaA==
 
 core-util-is@1.0.2, core-util-is@~1.0.0:
   version "1.0.2"
@@ -22892,7 +22908,7 @@ macos-release@^2.2.0:
   resolved "https://registry.yarnpkg.com/macos-release/-/macos-release-2.4.1.tgz#64033d0ec6a5e6375155a74b1a1eba8e509820ac"
   integrity sha512-H/QHeBIN1fIGJX517pvK8IEK53yQOW7YcEI55oYtgjDdoCQQz7eJS94qt5kNrscReEyuD/JcdFCm2XBEcGOITg==
 
-magic-string@0.26.2, magic-string@^0.26.1:
+magic-string@0.26.2:
   version "0.26.2"
   resolved "https://registry.yarnpkg.com/magic-string/-/magic-string-0.26.2.tgz#5331700e4158cd6befda738bb6b0c7b93c0d4432"
   integrity sha512-NzzlXpclt5zAbmo6h6jNc8zl2gNRGHvmsZW4IvZhTC4W7k4OlLP+S5YLussa/r3ixNT66KOQfNORlXHSOy/X4A==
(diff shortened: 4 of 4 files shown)


rockindahizzy:
I'm not sure if there is a way/need to test this fix

lmiller1990 (approved):
Nice, looks good to me. We should probably have Kitchensink running on our lowest supported Chrome for every branch.

For anyone interested, this is how the plugin works:

```html
<script type="module">try{import.meta.url;import("_").catch(()=>1);}catch(e){}window.__vite_is_modern_browser=true;</script>
  <script type="module">!function(){if(window.__vite_is_modern_browser)return;console.warn("vite: loading legacy build because dynamic import or import.meta.url is unsupported, syntax error above should be ignored");var e=document.getElementById("vite-legacy-polyfill"),n=document.createElement("script");n.src=e.src,n.onload=function(){System.import(document.getElementById('vite-legacy-entry').getAttribute('data-src'))},document.body.appendChild(n)}();</script>
```

It makes two entries, and only serves the legacy if the first one fails. "Legacy" is `try{import.meta.url;import("_").catch(()=>1);}` -  either `import.meta.url` doesn't exist or `import()` isn't defined.

This means we ship two bundlers - legacy and non legacy - so the binary is a couple hundred kb extra, not ideal, but probably the only way to support older browsers :shrug:.

mike-plummer (changes requested):
I think a bit more work is needed here. there's more than just the `globalThis` issue, that's just the first one that got hit. I get the following error and the SpecsList fails to render (sidebar does though) in Chromium v70 - `replaceAll` is Chrome 85+

Screen Shot 2022-09-07 at 9 00 08 AM

lmiller1990:
Right @mike-plummer good point - I was just reviewing for the globalThis polyfill.

Since we specify `esnext` here, TS won't warn us. I don't think we should use `esnext` (at all) since it assumes our target environment supports the latest JS features, like `replaceAll`, which it doesn't.

Line: https://github.com/cypress-io/cypress/blob/develop/packages/frontend-shared/tsconfig.json#L6 

The confusing thing is our minimal Node.js version has different features to our minimum Chrome version. Ideally, we'd have two `tsconfig.json` we extend from: one for frontend code and one for backend code. Those would be configured in a way that ensures we cannot write code that isn't supported.

For now, I think we should change the main `tsconfig.json` for frontend code (`frontend-shared/tsconfig.json`) to not target `esnext`, but something lower (`es6` aka `es2015` is probably fine). Then TS should give us compilation errors, for example telling us that `replaceAll` isn't okay. 

Edit: I just tried this in a simple project and it works, but **not** in the monorepo - probably since there's a bunch of `tsconfig.json` files and it's not clear how/which one is getting consumed during compilation? So some manual work or tinkering is probably required.

lmiller1990 (commented):
Left some thoughts about how we might modify/change tsconfig to ensure we don't write invalid code in the future - might require some additional work.

rockindahizzy:
@lmiller1990, if we set it to es2015, will we even need to polyfill?

lmiller1990:
> @lmiller1990, if we set it to es2015, will we even need to polyfill?

Yes we do - here's why. `lib: es2015` only impacts language *features* (think syntax) - not additional/new methods. Eg:

```json
{
  "compilerOptions": {
    "target": "ES5",
    "lib": ["ES2021"]
  },
  "include": ["./index.ts"]
}
```

```ts
// index.ts
"AAA".replaceAll("A", "B")

export function blah () {
  return () => 'AAA'
}
```

`yarn tsc -p .` gives:

```js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.blah = void 0;
"AAA".replaceAll("A", "B");
function blah() {
    return function () { return 'AAA'; };
}
exports.blah = blah;
```

Notice it will transform the `() => 'AAA'` to `function` (no arrow functions in `ES5`) but `replaceAll` is compiled as is. 

One argument for this behavior is because it's possible for a user to do `String.prototype.replaceAll = ...`, and TS would have no way to know. So, the compiler only compiles features.

If we want to polyfill old methods, we'd need to use babel or to include a polyfill.

lmiller1990 (commented):
Commented!

tgriesser:
> Nice, looks good to me. We should probably have Kitchensink running on our lowest supported Chrome for every branch.

@lmiller1990 is this defined anywhere? I could be wrong but I believe Chrome 65 was EOL in 2018:

https://endoflife.software/applications/browsers/google-chrome

lmiller1990:
> > Nice, looks good to me. We should probably have Kitchensink running on our lowest supported Chrome for every branch.
> 
> @lmiller1990 is this defined anywhere? I could be wrong but I believe Chrome 70 was EOL in 2018:
> 
> https://endoflife.software/applications/browsers/google-chrome

Hm, the issue says we should add the polyfill https://github.com/cypress-io/cypress/issues/23509. Which leads me to assume we support Chrome <70?

I think we probably need to actually define what minimum browser versions we support, and make sure that is running on Kitchensink (eg - if not Chrome 70, than what version). See [thread](https://cypressio.slack.com/archives/C011BAPC614/p1662666994632629) - I'll also tag you.

rockindahizzy:
We support the following browser versions [according to the docs](https://docs.cypress.io/guides/guides/launching-browsers#Browser-versions-supported)

- Chrome 64 and above.
- Edge 79 and above.
- Firefox 86 and above.

rockindahizzy:
> I think a bit more work is needed here. there's more than just the `globalThis` issue, that's just the first one that got hit. I get the following error and the SpecsList fails to render (sidebar does though) in Chromium v70 - `replaceAll` is Chrome 85+
> 
> Screen Shot 2022-09-07 at 9 00 08 AM

@mike-plummer, I am now polyfilling all modules. `replaceAll` should be working now

packages/app/vite.config.mjs
```diff
@@ -29,5 +30,9 @@ export default makeConfig({
         dest: 'dist',
       }],
     }),
+    Legacy({
+      targets: ['Chrome >= 64', 'Firefox >= 86', 'Edge >= 79'],
+      modernPolyfills: true,
```
mike-plummer, line 35: Why enable this option? The docs seem to recommend against it. I think all we care about is the `polyfills` option which is `true` by default and auto-generated off the `targets` entries

rockindahizzy: It didn't work for me without this option set 😕

mike-plummer: Huh, interesting. I must be misunderstanding the intent of that field.

mike-plummer (approved):
LGTM, tested on Chromium v70

astone123 (approved):
Worked for me 👍🏻

ZachJW34 (approved):
Paired with @rockindahizzy, works great!

rockindahizzy:
@lmiller1990, I looked into the tsconfig changes a little more and am thinking we break that change into another issue to avoid blowing up this issue and not getting the fix into this release.

Also, I'm still having a hard time understanding what changing the tsconfig really adds since we are relying on vite to polyfill for us.

lmiller1990:
@rockindahizzy I don't think we need to change tsconfig anymore - if we are shipping the entire polyfill, it should be fine. I was more thinking "can we avoid shipping a huge polyfill" but I don't think this is an option until we officially drop support for older browsers.

Issue #23083, opened 2022-08-03
[ACI] Cohort assignment and persistence to support A/B testing capability for Smart Notifications


mike-plummer:
# What is Needed

Capability to assign a user to selected "cohort" for a given "feature".  This cohort can be used by the app to vary the experience for the given users within a "feature".  The cohort functionality will encapsulate the logic for selecting a cohort for the user and persisting it in global state by "feature" so that once a cohort is selected, the user has a consistent experience each time it uses that "feature".  

See [tech brief](https://github.com/cypress-io/tech-briefs/tree/main/briefs/a-b-testing-app) for suggested implementation details.  Main goal for this issue would be to introduce reusable functionality for user cohorts that is not dependent on any part of the banner framework itself.

## Deliverable

* Reusable code that encapsulates the cohort functionality (proposed solution in tech brief calls for Vue composable)
  * See **Goals** section of [tech brief](https://github.com/cypress-io/tech-briefs/pull/45) for important implementation details

# Why is it Needed

Product desires a way to evaluate how different versions of copy might affect uptake/action on certain alert banners. To do this they want the ability to display different versions of copy and measure the rate at which users act on those versions. They can use this insight to improve phrasing in the future to ensure we're using language that resonates with our users.

See: #23118 

# Other

Should be able to persist the cohorts selected using the existing preferences framework. The "measure" aspect is covered by a separate analytics issue. Need to consider long-term cleanup - can we remove cohort information for banners we don't have anymore?

baus:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/23083?planning-poker) with ZenHub @amehta265 @astone123 @lmiller1990 @marktnoonan @mike-plummer @rockindahizzy @warrensplayer @ZachJW34

PR #23691, opened 2022-09-06
feat: Banner Analytics

### LINKED ISSUE #23084: [ACI] Banner analytics
# What is Needed

A capability to communicate to the dashboard that a particular smart notification has been shown to the user and if the user completes the call to action (CTA) for that notification.  With the introduction of A/B testing for this notifications ( #23118 ), the cohort ID that represents the copy shown should also be communicated.  

The Cloud API is introducing a new endpoint to capture the "viewed" event for each of the four smart banner notifications.  For tracking the CTA completion, two methods will be used.  For the Login banner and the Create Organization banner, the UTM parameters currently being passed in the URL will be updated to include cohort information.  For the Create Project banner, new fields in theGraphQL mutation will be introduced to pass this information. 

## Deliverables
* Ability to call new Cloud API endpoint when a banner notification is shown for the first time to a user (see Cloud issue and PR below)
* Update URLS used from the banners for login (i.e. authentication) to the dashboard as well as opening the "Create Organization" page to pass new UTM parameters See [spreadsheet](https://docs.google.com/spreadsheets/d/1Cs933TLXcS63PmMlXA4wKLJtQSFH90dProvzI9oyEB4/edit#gid=0) lines 17-18
* Update GraphQL mutation for creating a project to accept new parameters defined in the cloud for tracking 

# Why is it Needed

Product desires a way to know how many times a given banner was displayed and how many times actions from a given banner were taken. By knowing how many times each cohort was shown vs acted on a banner Product can measure effectiveness of different versions of copy.

# Other

* Cloud Issue: https://cypress-io.atlassian.net/browse/CLOUD-772
* Cloud PR: https://github.com/cypress-io/cypress-services/pull/4709

mike-plummer:
- Closes #23084 

### User facing changelog
No user impact

### Additional details
Adding aggregate recording of when "smart notification" banners are displayed as part of A/B testing implementation for these banners. Also adding new UTM parameters to Login and Create Project flows from those banners.

Hooks for "cohort" identifier are currently stubbed pending A/B cohort generation work being done under #23083 
Endpoint is currently hardcoded to use Staging so we do not unnecessarily pollute Prod data during development - we can filter this out if needed but no sense recording to Prod before we have complete data

### Steps to test
**Banner display event recording**
1. Identify a project that you have opened previously (at least 4 days ago) - banners will not display unless project is > 4 days old. If no old projects are available locally you can mock this out by manually editing the project's state.json file. Additionally, banners cannot have all been previously-dismissed - if they have been a manual edit to the state file will be needed.
2. Open Cypress with `DEBUG=cypress:data-context:sources:EventCollectorDataSource` .
3. Once launchpad is open ensure you are logged out, then proceed to the Specs List
4. The smart Login notification should display, check your terminal and validate that a message similar to this was displayed once:
```
[cy:open:dev:70043]:   cypress:data-context:sources:EventCollectorDataSource Recorded event: { campaign: 'aci_082022_login', messageId: '', medium: 'dev' } +0ms
```
**Note:** This will only be recorded the **_first_** time the banner is displayed, so you may need to clear the `banners` key from your project's `state.json` file if you've been shown the banners previously in this project
5. Do not proceed with Login, you can re-use this project for the next test phase

**UTM Parameters for Login**
At this time the new UTM parameters for Login are not fully hooked up (pending cohort). To validate the functionality you can manually stub out a utm_content value prop of "A" on the `LoginBanner` use in `LoginBanner.vue` and complete a login cycle from the banner using `DEBUG=cypress:*`. This should output a message to your terminal resembling this (note `utm_content` entry):

```
[cy:open:dev:71774]:   cypress:server:auth Received GET to /redirect-to-auth, redirecting: { fullLoginUrl: 'https://dashboard-staging.cypress.io/test-runner-login?utm_source=Binary%3A%20App&utm_medium=Specs%20Login%20Banner&utm_campaign=Log%20In&utm_content=A&.....&cypressVersion=10.7.0&platform=darwin' } +36ms
```

**UTM Parameters for Project Creation**
At this time the new UTM parameters for Cloud Project Creation are not fully hooked up (pending "cohort"). To validate the functionality you can manually stub out a cohort value in `SelectCloudProjectModal.vue` and run cypress while enabling "GraphQL Over Fetch" from the launchpad developer menu. Open a project without a project ID, open your browser devtools, and use the smart notification banner to create a *new* project via the modal. Find an entry in your devtools Network tab with operation ID `SelectCloudProjectModal_CreateCloudProject` and validate that it has a payload similar to the following:

```
...
campaign: "Create project"
cohort: "some value"
medium: "Specs Create Project Banner"
source: "Binary: App"
```

### PR Tasks


packages/app/src/specs/banners/TrackedBanner.vue
```diff
@@ -43,15 +46,33 @@ mutation TrackedBanner_SetProjectState($value: String!) {
...
 const props = withDefaults(defineProps<TrackedBannerComponentProps>(), {})
 
 const emit = defineEmits<TrackedBannerComponentEmits>()
 
 const stateQuery = useQuery({ query: TrackedBanner_ProjectStateDocument })
 const setStateMutation = useMutation(TrackedBanner_SetProjectStateDocument)
+const reportSeenMutation = useMutation(TrackedBanner_RecordBannerSeenDocument)
+const bannerInstanceId = ref(nanoid())
+
+const recordBannerShown = useDebounceFn(() => {
```
warrensplayer, line 64: Why the need for the debounce here?  Does it take time for the `props.modelValue` to settle or something?

warrensplayer (changes requested):
@mike-plummer Overall, looks great!  Since it looks like this issue already starts to implement making the API call in the banners, are you able to go ahead and finish the linked issue in this same branch and update this PR?

packages/app/src/specs/SpecsListBanners.vue
```diff
@@ -254,6 +263,11 @@ watch(
     showConnectBanner.value = !hasBannerBeenDismissed(BannerIds.ACI_082022_CONNECT_PROJECT) && isLoggedIn && isMemberOfOrganization && !isProjectConnected && hasFourDaysOfCypressUse
     showCreateOrganizationBanner.value = !hasBannerBeenDismissed(BannerIds.ACI_082022_CREATE_ORG) && isLoggedIn && isOrganizationLoaded && !isMemberOfOrganization && hasFourDaysOfCypressUse
     showLoginBanner.value = !hasBannerBeenDismissed(BannerIds.ACI_082022_LOGIN) && !isLoggedIn && hasFourDaysOfCypressUse
+
```
lmiller1990, line 266: Not saying we do it in this PR, but this watcher is so huge and complex - I think if we get any more complex we should move this into it's own function somewhere else, and potentially test it in isolation (it could just be boring old JS function, not coupled to the UI, since 90% of it is just "if X then Y else Z". 

Interested in your thoughts here - fat components vs thin components kind of thing... more of a philosophical discussion a real comment or change request.

packages/data-context/src/actions/EventCollectorActions.ts
```diff
@@ -0,0 +1,34 @@
...
+  medium: string
+}
+
+export class EventCollectorActions {
+  constructor (private ctx: DataContext) {}
+
+  async recordEvent (event: CollectableEvent): Promise<boolean> {
+    try {
+      // TODO: Stubbed with Staging for development
+      const dashboardUrl = 'https://dashboard-staging.cypress.io' //this.ctx.cloud.getDashboardUrl()
```
lmiller1990, line 18: Should we have some kind of process.env check here, or will this change in a subsequent PR?

packages/data-context/src/actions/EventCollectorActions.ts
```diff
@@ -0,0 +1,34 @@
...
+
+      await this.ctx.util.fetch(
+        `${dashboardUrl}/anon-collect`,
+        { method: 'POST', body: JSON.stringify({ ...event }) },
```
lmiller1990, line 22:
```suggestion
        { method: 'POST', body: JSON.stringify(event) },
```

packages/data-context/src/actions/EventCollectorActions.ts
```diff
@@ -0,0 +1,34 @@
...
+      )
+
+      debug(`Recorded event: %o`, event)
+
+      return true
+    } catch (err) {
+      debug(`Failed to record event %o due to error %o`, event, err)
```
lmiller1990, line 29: Nice... always good to have more `debug`.

packages/server/lib/gui/auth.ts
```diff
@@ -24,7 +24,7 @@ const buildLoginRedirectUrl = (server) => {
   return `http://127.0.0.1:${port}/redirect-to-auth`
 }
 
-const buildFullLoginUrl = (baseLoginUrl, server, utmSource, utmMedium) => {
+const buildFullLoginUrl = (baseLoginUrl, server, utmSource, utmMedium, utmContent) => {
```
lmiller1990, line 27: Nit: last three args should be an object of options, like `utmData`. Otherwise, it's easy to mix them up since they are all strings.

packages/data-context/src/actions/EventCollectorActions.ts
lmiller1990: Edit I saw 

> Endpoint is currently hardcoded to use Staging so we do not unnecessarily pollute Prod data during development - we can filter this out if needed but no sense recording to Prod before we have complete data


Disregard this

lmiller1990 (approved):
Some misc comments, no blocker.

lmiller1990:
Posted some nits, not really blockers, I also tried it out, seems to work as expected! Eg I see banner and correct log, etc.

image

packages/frontend-shared/src/gql-components/Auth.vue
```diff
@@ -194,7 +195,7 @@ const handleLoginOrContinue = async () => {
 
   loginInitiated.value = true
 
-  login.executeMutation({ utmMedium: props.utmMedium, utmSource: getUtmSource() })
+  login.executeMutation({ utmMedium: props.utmMedium, utmContent: props.utmContent || null, utmSource: getUtmSource() })
```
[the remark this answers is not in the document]
mike-plummer: I struggled with this a bit - the tricky thing is that GQL treats an arg as `string | null`, but an optional argument for a JS function would be `string | undefined`. It felt cleaner from a function contract perspective to leave it as optional and deal with the misalignment this way, but I'm open to suggestions

packages/frontend-shared/src/gql-components/Auth.vue
```diff
@@ -204,7 +205,7 @@ const handleLogout = () => {
 const handleTryAgain = async () => {
   await reset.executeMutation({})
 
-  login.executeMutation({ utmMedium: props.utmMedium, utmSource: getUtmSource() })
+  login.executeMutation({ utmMedium: props.utmMedium, utmContent: props.utmContent || null, utmSource: getUtmSource() })
```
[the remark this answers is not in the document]
mike-plummer: Same as above

packages/app/src/specs/banners/TrackedBanner.vue
mike-plummer: The `watchEffect` was firing twice, not sure exactly which of its dependencies was updating to cause the second firing. The only solutions I could think of were to do the debounce or move this to a lifecycle hook (`onMounted`). The lifecycle hook wouldn't catch post-mount changes to props (like a delayed resolution of `hasBannerBeenShown`) so this seemed like the best solution.

The `messageId` would allow us to filter out the duplicates on the backend, but it's better to avoid sending duplicates in the first place if there's an easy way not to

packages/app/src/specs/SpecsListBanners.vue
mike-plummer: Agreed, this is a monster. I know that the work Mark is doing will move a lot of this into a reusable wrapper component so I think that will help clean this up (or at least split the concerns into multiple smaller chunks). If that weren't the case I would lean towards building a couple composables to make some of the common checks this is doing a bit more reusable (`useIsLoggedIn`, `useIsOrgMember`, etc), and I'm sure most of the composable logic could be structured to test as pure boring JS

packages/frontend-shared/src/gql-components/Auth.vue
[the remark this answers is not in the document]
astone123: Yeah I've run into this issue with GraphQL in the past too. The type is `string | null` rather than `string | undefined` 😕

warrensplayer:
@mike-plummer Looked over the latest changes and they look great.  I just need to pull the code to run through testing it one more time.

packages/app/src/specs/banners/CreateOrganizationBanner.vue
```diff
@@ -8,6 +8,12 @@
     class="mb-16px"
     :icon="OrganizationIcon"
     dismissible
+    :has-banner-been-shown="hasBannerBeenShown"
+    :event-data="{
+      campaign: 'Set up your organization',
+      medium: 'Specs Create Organization Banner',
+      cohort: '' // TODO Connect cohort
```
marktnoonan, line 15: What's the trigger for this to get completed? Will it be done as part of the existing cohort work after this is merged? It would be good to add some context/link to the issue. Though I think your are coordinating with @warrensplayer already.

packages/data-context/src/actions/EventCollectorActions.ts
astone123: One thing to note here - the runner is now gathering data from production cloud by default when in development. So `this.ctx.cloud.getDashboardUrl()` will resolve to the production cloud URL because `DEFAULT_INTERNAL_CLOUD_ENV` will be `production`.

We probably don't want to send these events to production when in dev mode, so we might want to use `CYPRESS_INTERNAL_ENV` instead, and only use the production URL when that variable is `production`.

packages/app/src/specs/banners/CreateOrganizationBanner.vue
mike-plummer: Correct, @warrensplayer is working on the Cohort assignment & persistence logic under #23083 and will hook these up as part of that work

marktnoonan:
For the second test scenario:

> utm_content value prop of "A" on the `LoginBanner` use in `LoginBanner.vue`

- this should be "`LoginModal` use", right? And when cohorts are set up, we'll communicate them via utm_content. Works great.

All the manual testing checks out for me.

packages/server/lib/gui/auth.ts
marktnoonan: I like objects in general when there's multiple params though typescript helps a bunch. I'd agree not essential to change this, just a preference.

packages/app/src/specs/banners/TrackedBanner.vue
```diff
@@ -10,16 +10,26 @@
...
 <script setup lang="ts">
 import Alert from '@packages/frontend-shared/src/components/Alert.vue'
-import { watchEffect } from 'vue'
+import { ref, watchEffect } from 'vue'
 import { gql, useMutation, useQuery } from '@urql/vue'
-import { TrackedBanner_ProjectStateDocument, TrackedBanner_SetProjectStateDocument } from '../../generated/graphql'
+import { TrackedBanner_ProjectStateDocument, TrackedBanner_RecordBannerSeenDocument, TrackedBanner_SetProjectStateDocument } from '../../generated/graphql'
 import { set } from 'lodash'
+import { useDebounceFn } from '@vueuse/core'
+import { nanoid } from 'nanoid'
```
ZachJW34, line 18: I don't see `nanoid` in any of our `package.json`s, can we add it to the app? We shouldn't rely on a transitive dep.

ZachJW34 (approved):
Went through the flow, thanks for the great testing steps! One small comment

packages/app/src/specs/banners/CreateOrganizationBanner.vue
astone123: I think it's #23083

packages/data-context/src/actions/EventCollectorActions.ts
mike-plummer: @astone123 Good point. I'd convinced myself they should be the same but you're probably right, we don't want to write events to production when running locally even though we're reading data from prod

packages/app/src/specs/banners/TrackedBanner.vue
```diff
@@ -43,15 +53,37 @@ mutation TrackedBanner_SetProjectState($value: String!) {
...
+    campaign,
+    messageId: bannerInstanceId.value,
+    medium,
+    cohort: cohort || null,
+  })
+}, 1000)
 
 watchEffect(() => {
   if (props.modelValue) {
+    if (!props.hasBannerBeenShown && props.eventData) {
```
marktnoonan, line 82: Non-blocking: It seems like the debounce here could be avoided if we directly watch only the `modelValue`. The purpose of the `watchEffect` is that you _want_ to run an effect for every reactive value referenced inside that changes, so the `watchEffect` into a debounced function is a little unintuitive to me.

Here's how it would look without:

```js
const recordBannerShown = ({ campaign, medium, cohort }: EventData) => {
  reportSeenMutation.executeMutation({
    campaign,
    messageId: bannerInstanceId.value,
    medium,
    cohort: cohort || null,
  })
}

watch(() => props.modelValue, (newVal) => {
  if (newVal) {
    if (!props.hasBannerBeenShown && props.eventData) {
      // We only want to record the banner being shown once per user, so only record if this is the *first* time the banner has been shown
      recordBannerShown(props.eventData)
    }

    updateBannerState('lastShown')
  }
}, {
  immediate: true,
})
```

I do like the debounce test though, since it is quite important to not repeatedly send these. Maybe I'm missing something else about the watchEffect though.

astone123 (approved):

marktnoonan (approved):

warrensplayer (approved):
Everything is working for me.  I would just echo Zach's comment that the `nanoid` dependency needs to be added to the package.json so it is not a transitive dependency

mike-plummer:
@marktnoonan I refactored the watcher to remove the need for the debounce
@ZachJW34 & @warrensplayer Added the missing `nanoid` dependency
@astone123 Added new env var to control the cloud environment that should be used for event collection (`CYPRESS_INTERNAL_ENV` is set to `development` vite when running locally, so creating new var in `gulpConstants` before vite gets involved)

scripts/gulp/gulpConstants.ts
```diff
@@ -10,24 +10,29 @@ declare global {
 
 export const DEFAULT_INTERNAL_CLOUD_ENV = process.env.CYPRESS_INTERNAL_ENV || 'production'
 
+export const DEFAULT_INTERNAL_EVENT_COLLECTOR_ENV = process.env.CYPRESS_INTERNAL_ENV || 'staging'
```
astone123, line 13: good call 👍🏻

Issue #23690, opened 2022-09-06
Snapshot generation fails when a dependency moves from healthy to deferred/norewrite


ryanthemanuel:
### What would you like?

Provide a better developer experience for when a dependency moves from healthy to deferred/norewrite during the snapshot generation process

PR #23692, opened 2022-09-06
fix: iterative snapshot generation with a new intermediate deferred or norewrite dependency 

### LINKED ISSUE #22989: Adding a New Require in Cypress Code Can Cause Generating Prod Snapshots to Fail
### What would you like?

Currently, it's broken to iteratively generate snapshots when you introduce a new require or import that is both not a leaf and either a `deferred` or `norewrite` dependency.

### Why is this needed?

While it would be possible to regenerate the snapshots from scratch, this takes a long time and it would be preferable to be able to iteratively generate the snapshots.

ryanthemanuel:
- Closes #22989

### User facing changelog

n/a

### Additional details

Currently, iterative snapshot generation fails under the following scenario:

1. Generate a snapshot
2. Add a require for a new deferred or norewrite file that is not a leaf.
3. Regenerate the snapshot using the snapshot metadata from step 1 as a starting point

The problem is in the logic that determines the next phase of snapshot doctoring. If all leaves have been accounted for but there is an unaccounted for intermediate dependency a next doctoring phase will not be found. I am refactoring the code to ensure that this edge case also accounts for finding the appropriate next phase of doctoring.

Note: There is still a scenario that will fail snapshot generation. In this case snapshot generation will fail if a healthy dependency changes to a deferred or norewrite dependency. I am leaving a skipped test in the PR for that scenario, but I did some analysis and there were no instances of this happening since we released 10.0 so it likely will not occur that frequently. I logged https://github.com/cypress-io/cypress/issues/23690 to address this later. Until that is addressed, snapshots will have to be generated from scratch if that happens. I will add instructions for how to do that on my subsequent PR that incorporates snapshot generation into the monorepo.

### Steps to test

I added a test case as part of this PR, you can verify this works by seeing the test case fail without my changes and then ensure that it succeeds with my changes.

n/a

### PR Tasks


tooling/v8-snapshot/test/fixtures/iterative/templates/intermediate-norewrite.js
```diff
@@ -0,0 +1,250 @@
+require ('./norewrite')
+
+// Generated by CoffeeScript 1.9.3
+(function() {
```
mjhenkes, line 4: Is this all test code? What part of it makes it 'norewrite', should we reduce this to just that?

mjhenkes (approved):

tooling/v8-snapshot/test/fixtures/iterative/templates/intermediate-norewrite.js
ryanthemanuel: Updated: [`47fd989` (#23692)](https://github.com/cypress-io/cypress/pull/23692/commits/47fd9891fd645ccbe7085433d66f1997d2ee75f2)

tooling/v8-snapshot/test/utils/bundle.ts
```diff
@@ -12,7 +12,7 @@ export function readBundleResult (cacheDir: string) {
 
 export function readSnapshotResult (cacheDir: string) {
   const metaFile = path.join(cacheDir, 'snapshot-meta.json')
-  const meta = require(metaFile)
+  const meta = fs.readJSONSync(metaFile)
```
mschile, line 15: Does the "no synchronize file operations" rule apply here for the v8 tooling?

https://github.com/cypress-io/cypress/blob/develop/CONTRIBUTING.md#enfile-or-emfile

ryanthemanuel: I believe we're only concerned with this at runtime (so tests and scripts don't worry about it):

https://github.com/cypress-io/cypress/blob/develop/.eslintrc.js#L83
https://github.com/cypress-io/cypress/blob/develop/.eslintrc.js#L41-L47

mschile (approved):

PR #23693, opened 2022-09-07
chore: fix test and types in server

lmiller1990:
Follow up https://github.com/cypress-io/cypress/pull/23579/files#r963726642

### DIFF
--- a/packages/launcher/lib/browsers.ts
+++ b/packages/launcher/lib/browsers.ts
@@ -166,7 +166,7 @@ export function launch (
   debuggingPort: number,
   args: string[] = [],
   defaultBrowserEnv = {},
-): LaunchedBrowser {
+) {
   debug('launching browser %o', { browser, url })
 
   if (!browser.path) {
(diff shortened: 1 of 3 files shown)


packages/launcher/lib/browsers.ts
```diff
@@ -166,7 +166,7 @@ export function launch (
   debuggingPort: number,
   args: string[] = [],
   defaultBrowserEnv = {},
-): LaunchedBrowser {
```
lmiller1990, line 169: Return value is implicit

image

packages/server/lib/browsers/firefox.ts
```diff
@@ -561,7 +561,7 @@ export async function open (browser: Browser, url: string, options: BrowserLaunc
 
       debug('closing firefox')
 
-      originalBrowserKill.apply(browserInstance, args)
+      return originalBrowserKill.apply(browserInstance, args)
```
lmiller1990, line 564: Needed to return this so the overridden `kill` has the correct signature (needs a return value of `boolean`)

packages/server/lib/browsers/firefox.ts
```diff
@@ -531,12 +531,12 @@ export async function open (browser: Browser, url: string, options: BrowserLaunc
 
   debug('launch in firefox', { url, args: launchOptions.args })
 
-  const browserInstance = await launch(browser, 'about:blank', remotePort, launchOptions.args, {
```
lmiller1990, line 534: it's not async

BlueWinds (approved):

PR #23696, opened 2022-09-07
fix: Use tsconfig from build if exists (closes #23673)

### LINKED ISSUE #23673: TSConfig seemingly not used when passed in via angular's build options (see #23161)
### Current behavior

In https://github.com/cypress-io/cypress/issues/23161 we were allowed to pass in an Angular config object, which also includes a tsconfig-path. However, that tsconfig path seems to be fully ignored; if I pass in a random string, everything builds/tests as it should. I dug a bit through the code and looking at e.g https://github.com/cypress-io/cypress/blob/a9d97114073f0ddc2983d775bfef60145df99902/npm/webpack-dev-server/src/helpers/angularHandler.ts#L135, it always uses <projectroot>/tsconfig.json. 

### Desired behavior

In my Angular / NX apps I have:
- tsconfig.json (baseline)
- tsconfig.app.json
- tsconfig.jest.json
- tsconfig.editor.json

I would also like to have a tsconfig.cypress.json to ensure the correct paths etc are used without having to interfere with the rest of my tsconfig-setup.

### Test code to reproduce

```typescript
// cypress.config.ts
export default defineConfig({
  component: {
    video: true,
    videosFolder: root + 'dist/cypress/components/videos',
    screenshotsFolder: root + 'dist/cypress/components/screenshots',
    devServer: {
      framework: 'angular',
      bundler: 'webpack',
      options: { // temporary, this will be done automatically when we get it directly from nx
        projectConfig: {
          root: projectRoot,
          sourceRoot: projectRoot + 'src',
          buildOptions: {
            tsConfig: _projectRoot + 'tsconfig.app.json',
            ... more config (that works yay)
          },
        }
      }
    },
    specPattern: 'cypress/specs/**/*.cy.ts',
  },
});
```

### Cypress Version

10.7.0

### Node version

v18.4.0 (also tried with 16.2.0)

### Operating System

macOS 12.5

### Debug Logs

_No response_

### Other

_No response_

yusijs:
- Closes #23673

### User facing changelog
Change that allows passing in a default tsConfig flag in buildOptions to angularHandler

### Additional details
Allow users to pass in a specific tsconfig.json (e.g to avoid collisions with jest and similar)

### Steps to test
N/A

N/A

### PR Tasks

As a quick note - I was wondering if I should add any tests for this, but seemed like such a small change it should not warrant one. However if it seems necessary, I can look into adding a test for it.

### DIFF
--- a/npm/webpack-dev-server/src/helpers/angularHandler.ts
+++ b/npm/webpack-dev-server/src/helpers/angularHandler.ts
@@ -133,7 +133,7 @@ export async function generateTsConfig (devServerConfig: AngularWebpackDevServer
   includePaths.push(cypressTypes)
 
   const tsConfigContent = JSON.stringify({
-    extends: getProjectFilePath('tsconfig.json'),
+    extends: getProjectFilePath(buildOptions.tsConfig ?? 'tsconfig.json'),
     compilerOptions: {
       outDir: getProjectFilePath('out-tsc/cy'),
       allowSyntheticDefaultImports: true,
--- a/npm/webpack-dev-server/test/handlers/angularHandler.spec.ts
+++ b/npm/webpack-dev-server/test/handlers/angularHandler.spec.ts
@@ -195,7 +195,8 @@ const expectGeneratesTsConfig = async (devServerConfig: AngularWebpackDevServerC
   let tsConfig = JSON.parse(await fs.readFile(tsConfigPath, 'utf8'))
 
   expect(tsConfig).to.deep.eq({
-    extends: toPosix(path.join(projectRoot, 'tsconfig.json')),
+    // verifies the default `tsconfig.app.json` is extended
+    extends: toPosix(path.join(projectRoot, 'tsconfig.app.json')),
     compilerOptions: {
       outDir: toPosix(path.join(projectRoot, 'out-tsc/cy')),
       allowSyntheticDefaultImports: true,
@@ -211,6 +212,7 @@ const expectGeneratesTsConfig = async (devServerConfig: AngularWebpackDevServerC
   const modifiedBuildOptions = cloneDeep(buildOptions)
 
   delete modifiedBuildOptions.polyfills
+  modifiedBuildOptions.tsConfig = 'tsconfig.cy.json'
 
   const modifiedDevServerConfig = cloneDeep(devServerConfig)
   const supportFile = path.join(projectRoot, 'cypress', 'support', 'component.ts')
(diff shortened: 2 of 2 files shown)


lmiller1990 (commented):
Generally we wants tests for anything - especially a regression like this.

You could modify one of the projects in `system-tests/angular*` to regress, than ensure it passes. Eg this one: https://github.com/cypress-io/cypress/tree/develop/system-tests/projects/angular-14. 

To run that test, you would add `systemTests.it.only` (otherwise it'll take a long time) then do:

```sh
cd system-tests
yarn test component_testing_spec
```

It should run the tests.

If you could add this, it'd be great - this could be in the next release. Otherwise, it'll still get added by someone internally, but might take a little longer.

Let me know if you need any more help to get the test working.

yusijs:
I did amend an existing test to ensure it uses the correct tsconfig.json. Can add an additional test to ensure it falls back to `tsconfig.json` if it is left blank.

lmiller1990:
I could be missing something (don't know this part of the code base really well) but it looks like the test would pass either way, since if `options.tsconfig` is null, it'll fall back to `tsconfig.json` (if it exists, which it does).

Eg, if I remove your commit, will something fail?

Here's what I was thinking (just thinking out loud here):

We could copy+paste [`system-tests/projects/angular-14`](https://github.com/cypress-io/cypress/blob/develop/system-tests/projects/angular-14/package.json#L30) into a new directory (maybe angular-14-build-options). We would then add an empty `tsconfig.json` that won't work. Then, in `buildOptions` in `angular-14-build-options/cypress.config.ts`, we'd provide one that *does* work. 

This way, it'll use the valid one from `buildOptions` - then, if we did revert your commit, that test would fail (since it'd default to the empty/invalid `tsconfig.json`.

Let me know if 1) this makes sense or 2) I'm missing something and we don't actually need this extra test to cover the regression.

yusijs:
> 1) this makes sense (Cypress test infrastructure is pretty big and complex)
It kind of makes sense, but as you say the test infra is pretty complex. If my commit is removed, nothing will stop working and tests will still pass (due to falling back to old behaviour w/ tsconfig.json to avoid regressions). It seems pretty straight forward to write a test like you mentioned though, so I can give that a spin :) 

> 2) I'm missing something and we don't actually need this extra test to cover the regression.

So the reason I decided not to write a test originally was simply that I didn't see a value to the test. Previously there were no test to verify that the user-defined tsConfig was used, but after my changes to angularHandler.spec.ts it now verifies that it uses whichever file is passed in, but still falls back to the previous behaviour (i.e using the root tsconfig.json). 

That doesn't necessarily mean we dont need a specific test for this, I just didn't see any value from it (and if Cypress has taught me anything, we should write valuable tests :) )

ZachJW34:
Changes look good!

Would be good to modify `npm/webpack-dev-server/test/handlers/angularHandler.spec.ts -> expectGeneratesTsConfig` to verify that the generated tsConfig has the `"extends": "<project-root>/<buildOptions.tsConfig>"

lmiller1990:
I tried to add at test but no luck - are we sure this is working as expected? It still generated a `tsconfig.json`, but it didn't have my additional fields.

I think in additional to adding a test, we need to document what this does. Maybe in `npm/angular/README.md`? Does it override the existing tsconfig, or merge? Also, should I pass an object of options or a path to a tsconfig.json?

yusijs:
Eeh, it works as it should, yes. It doesnt change the generated tsconfig aside from the filename it references (since it extends it). I began writing a test myself but I never got the test to actually make any sense, since it already tests for the exact behaviour expected. 

@jordanpowell88 do you have any input here (I believe you did the buildOptions themselves)?

yusijs:
> Changes look good!
> 
> Would be good to modify `npm/webpack-dev-server/test/handlers/angularHandler.spec.ts -> expectGeneratesTsConfig` to verify that the generated tsConfig has the `"extends": "<project-root>/<buildOptions.tsConfig>"`

It does, I think?

https://github.com/cypress-io/cypress/pull/23696/files#diff-59bd76d1ab6c0d9fb482c07fccbb0662ecb8cb09994a0d611f7f3b5383b36a81R198

lmiller1990:
Nice, let's add that test and merge this up - it can go into the next release (Tue 13th Sep).

yusijs:
Added it in :) I was a bit worried originally about changing the existing tests to much, but makes sense like this :)

ZachJW34 (approved):

lmiller1990 (approved):

lmiller1990:
Nice job!

PR #23575, opened 2022-08-26
fix: issue-7715

rachelruderman:
- Closes #7715

### User facing changelog
In the terminal, users will now see the source map line and column number in the error stack trace. Fixes #7715

### Additional details

Previously, we were finding the stack trace with the source map -- but we weren't doing anything with it 😱 it was appended to the error as `error.sourceMappedStack` and left to linger, perhaps for debugging purposes.

This PR reassigns the `stack` property to the value of the source mapped stack and removes the lone property `sourceMappedStack`. 

The purists among us may say,
> Hey! Should we really be manipulating the `stack` property? 

To which I will answer, 
> We were already manipulating the `stack` property by pointing it to the user's invocation stack via [preferredStackAndCodeFrameIndex](https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cypress/error_utils.ts#L434). This change just ensures the right line number/column is printed in the already-manipulated stack.

### Steps to test
1. Create a dummy test that will fail, i.e. in `packages/driver/cypress/e2e/commands/agents.cy.js`
```
Cypress.env('retries', 0)

it.only('should fail', () => {
  expect(false).to.be.true
    // OR
    // cy.get('invalid')
    // 	.should('be.visible')
    // OR
    // cy.origin('http://foobar.com:3500')
})
```
2. From the project root, start the runner in one tab with `cd packages/runner && yarn watch`
3. From the project root, run your test in another tab with `yarn cypress:run --headed --project packages/driver --spec packages/driver/cypress/e2e/commands/agents.cy.js --browser chrome --no-exit`
4. Confirm that the error stack line number/column in the terminal matches the value in VS Code
5. Confirm that the error stack line number/column in the GUI matches the line number/column in VS Code

**For example, all three of these show line 8, column 25: `cypress/e2e/commands/agents.cy.js:8:25`**
- VS Code:
Screen Shot 2022-08-26 at 1 02 31 PM

- Terminal:

Screen Shot 2022-08-26 at 1 02 15 PM

- GUI:

Screen Shot 2022-08-26 at 1 07 39 PM

### How has the user experience changed?
[...]

PR #23730, opened 2022-09-08
chore(webkit): update error stack parsing and related system tests

tbiethman:
- Related to #6422

### User facing changelog

N/A

### Additional details

WebKit error stacks currently have a number of documented issues, resulting in stack traces that aren't very helpful:

[https://bugs.webkit.org/show_bug.cgi?id=86493](https://bugs.webkit.org/show_bug.cgi?id=86493)
[https://bugs.webkit.org/show_bug.cgi?id=243668](https://bugs.webkit.org/show_bug.cgi?id=243668)
[https://bugs.webkit.org/show_bug.cgi?id=174380](https://bugs.webkit.org/show_bug.cgi?id=174380)

Screen Shot 2022-09-08 at 11 12 28 AM

This PR updates WebKit stack lines with some placeholder function/location values, which makes them more readable (if not more functional) and allows us to better parse them for presentation within the reporter and snapshot normalization. We are using `<unknown>` in the existing parsing logic when function names/locations could not be determined and I continued that here, but we could use something more verbose if it would be helpful.

It's important to note that the lack of call locations in these stacks makes a lot of our existing strategies for stack inspection/replacement much harder (if not impossible) to implement. A [number of changes](https://bugs.webkit.org/show_bug.cgi?id=243668#c3) have recently landed in WebKit that should improve these stacks, at least so far as removing a lot of these internal stack items that this PR is working with. But the full impact of these changes (and the timeline for their inclusion in a new playwright-webkit release) is yet to be determined.

### Steps to test

* Open project of choice in WebKit
* Update spec to fail (throw error in hook, `expect(true).to.eq(false)`, throw uncaught error in fixture, etc.)
* View stack trace of error, you shouldn't see any lonesome "@" symbols or stack entries missing a function name or location entirely

Before:

Screen Shot 2022-09-07 at 11 58 14 AM

After:

Screen Shot 2022-09-08 at 10 26 11 AM

### PR Tasks


packages/driver/src/cypress/stack_utils.ts
```diff
@@ -390,7 +396,35 @@ const normalizedStack = (err) => {
...
+    // https://bugs.webkit.org/show_bug.cgi?id=174380
+    //
+    // We update these stack entries with placeholder names/locations to more closely align
+    // the output with other browsers, minimizing the visual impact to the stack traces we render
+    // within the command log and console and ensuring that the stacks can be identified within
+    // and parsed out of test snapshots that include them.
+    errStack = errStack.replaceAll(webkitStackLineRegex, (match, ...parts: string[]) => {
+      // We patch WebKit's Error within the AUT as CyWebKitError, causing it to
+      // be presented within the stack. If we detect it within the stack, we remove it.
+      if (parts[0] === '__CyWebKitError') {
```
flotwig, line 416: Is this because it's showing up in the stack trace of every `new Error`? Should we just not name the function then?

flotwig (approved):

packages/driver/src/cypress/stack_utils.ts
tbiethman: I went back and forth on this. If we don't name it, it'll still render here, just as yet another anonymous '<unknown>' function. Given the frequency with which it'll be presented (for any uncaught exception thrown by the AUT), I figured its presence in the stacks might be confusing, named or not. But I could understand the argument that removing it is an improper obfuscation of what is actually being executed?

flotwig: hmm, yeah... maybe we just oughtta name it `Error`, then. This seems fine too. Not too big an issue for experimental.

rachelruderman:
Ahhhhhh I also have an open PR https://github.com/cypress-io/cypress/pull/23575 that messes with stack_utils, MAY THE GIT MERGE BEGIN!!  Since my stuff is annoying and a lil complicated and I know how to test it, let's merge yours first and I'll figure out the conflict 🥲 ⚔️

rachelruderman (approved):

chrisbreiding (approved):

flotwig (approved):

Issue #14656, opened 2021-01-20
Test gets into runaway loop if cy.wrap is inside cy.should


bahmutov:
Cypress v6.3.0
```js
it('gets into a loop', () => {
  cy.wrap('bar')
    .should(() => {
      cy.wrap('foo')
    })
})
```

Runs away, never finishes

https://user-images.githubusercontent.com/2212006/105213850-39648400-5b1d-11eb-92fe-0d04751496e6.mp4

Example in https://github.com/cypress-io/cypress-test-tiny/tree/runaway

sainthkh:
Problem solved. I'll upload the PR with why it happened tomorrow.

sainthkh:
The problem is related with the command stack. I explained why in #15386.

sainthkh:
I tried 2 approaches and learned that allowing cy commands inside `should` callback is impossible because of `Cypress detected that you returned a promise in a test, but also invoked one or more cy commands inside of that promise.` error. I tried to solve this problem by removing `Promise` inside `shouldFnWithCallback`, but it also failed. It might be because `should` command is called inside a `Promise`.

If you want to check that cy commands don't work inside `should` callback, test the code below:

```js
it('gets into a loop', () => {
  cy.wrap('bar')
    .should(() => {
      cy.wrap('foo')
      cy.wrap('test')
    })
})
```

All you can see is `wrap foo`. `wrap test` is never shown. Because the error is thrown. 

If you want to check the thrown error, wrap `fn` call with `try...catch` below.

https://github.com/cypress-io/cypress/blob/cb281d17e4e4cce3eff22f367dbb86837b299439/packages/driver/src/cy/commands/asserting.js#L11-L24

----

I also thought what's the best solution for this. Showing "cy commands are not allowed inside `should` command" might be a good idea.

What do you think, @bahmutov?

jennifer-shehane:
Oof, I feel like I've seen this behavior before. Ideally I would like this to just work as expected, but probably throwing an error would be better than an infinite loop. I"m not sure what the rest of the team thinks though.

bahmutov:
Just randomly hit this again, ughh, and aside from this I do see runaway tests sometimes, but larger ones, seems there might be edge cases in the internal logic

```js
it('never finishes', () => {
  cy.wrap(1).should(() => cy.log('here'))
})
```

https://user-images.githubusercontent.com/2212006/140448613-37acd0d6-3106-4977-8559-f9f2dc4ddbe9.mp4

[...]

Issue #5963, opened 2019-12-15
get() inside should() with callback applies default assertion


gabbersepp:
### Current behavior:

Using `get().should(not.exist)` inside should with callback applies default `exist` assertion.

![image](https://user-images.githubusercontent.com/3234900/70860669-c7d30300-1f24-11ea-884a-1ba8e99d6133.png)

### Desired behavior:

`should` will skip the default 

### Steps to reproduce: (app code and test code)

Use this example:

```js
describe("ad", () => {
    it("test", () => {
        cy.visit("https://www.cypress.io/")
        cy.get("*[aria-label='pricing']")
            .invoke('text').should(someValue => { 
                cy.get("asdad", {timeout: 5000}).should("not.exist");
            })
    })
})
```

The element "asdad" does not exist  but even though cypress fails while applying the default asertion.

### Versions

Cypress 3.8.0

jennifer-shehane:
Confirmed this behavior.

## `.should()` chaining

```js
it('test', () => {
  cy.visit('https://www.cypress.io/')
  cy.get('h1').should(() => {
    cy.get('asdad').should('not.exist') // fails incorrectly
  })
})
```

<img width="448" alt="Screen Shot 2019-12-16 at 11 37 22 AM" src="https://user-images.githubusercontent.com/1271364/70880738-75fbad00-1ff8-11ea-93af-bee9d3b03e46.png">


## `.then()` chaining

```js
it('test', () => {
  cy.visit('https://www.cypress.io/')
  cy.get('h1').then(() => {
    cy.get('asdad').should('not.exist') // passes correctly
  })
})
```

<img width="449" alt="Screen Shot 2019-12-16 at 11 36 59 AM" src="https://user-images.githubusercontent.com/1271364/70880719-654b3700-1ff8-11ea-9a11-08dbeb59db52.png">

gabbersepp:
Hey,
T tried to dive deeper into the cypress internals to fix this bug. I found the line `isDefaultAssertionErr = cmds.length is 0` in `assertions.coffee` that seems to control wether the default assertions are applied. But I found no place in the `driver` package that makes use of this information. Maybe you can give me a little hint where I can find the correct place?

emilyrohrbough:
https://github.com/cypress-io/cypress/pull/23734 made a breaking change to now throw an error if Cypress commands are invoked from inside a `.should()` callback since this resulted in unusual and undefined behavior.

Closing this issue since this is no longer relevant.

PR #23734, opened 2022-09-08
feat: Error if command used inside .should() callback

### LINKED ISSUE #22587: Callback form of should chained off cy.window receives null when retried

BlueWinds:
- Closes https://github.com/cypress-io/cypress/issues/22587
- Closes https://github.com/cypress-io/cypress/issues/5963
- Closes https://github.com/cypress-io/cypress/issues/14656

### User facing changelog
Cypress now throws an error if commands are invoked from inside a `.should()` callback. This previously resulted in unusual and undefined behavior; it is now explicitly an error.

### Additional details
Callbacks from .should() are invoked repeatedly as part of `verifyUpcomingAssertions`; Cypress commands used inside of them are therefore added to the queue repeatedly, causing all sorts of unusual and unexpected behavior, including corrupting the Cypress subject, command log, and command queue in general.

It's better to throw a helpful error than it is let Cypress end up in an undefined state and have the user guess what went wrong.

### Steps to test
The [linked issue](https://github.com/cypress-io/cypress/issues/22587) has one example of what can go wrong if we don't error out early.

image

### PR Tasks
- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4755

### DIFF
--- a/packages/driver/cypress/e2e/commands/assertions.cy.js
+++ b/packages/driver/cypress/e2e/commands/assertions.cy.js
@@ -305,7 +305,7 @@ describe('src/cy/commands/assertions', () => {
       // https://github.com/cypress-io/cypress/issues/22587
       it('does not allow cypress commands inside the callback', (done) => {
         cy.on('fail', (err) => {
-          expect(err.message).to.eq('`cy.should()` failed because you invoked a command inside the callback. Use `cy.then()` instead of `cy.should()`, or move any commands outside the callback function.\n\nThe command invoked was:\n\n  > `cy.log()`')
+          expect(err.message).to.eq('`cy.should()` failed because you invoked a command inside the callback. `cy.should()` retries the inner function, but commands are already retried. Use `cy.then()` instead of `cy.should()`, or move any commands outside the callback function.\n\nThe command invoked was:\n\n  > `cy.log()`')
 
           done()
         })
--- a/packages/driver/cypress/e2e/commands/assertions.cy.js
+++ b/packages/driver/cypress/e2e/commands/assertions.cy.js
@@ -302,6 +302,19 @@ describe('src/cy/commands/assertions', () => {
         })
       })
 
+      // https://github.com/cypress-io/cypress/issues/22587
+      it('does not allow cypress commands inside the callback', (done) => {
+        cy.on('fail', (err) => {
+          expect(err.message).to.eq('`cy.should()` failed because you invoked a command inside the callback. Use `cy.then()` instead of `cy.should()`, or move any commands outside the callback function.\n\nThe command invoked was:\n\n  > `cy.log()`')
+
+          done()
+        })
+
+        cy.window().should((win) => {
+          cy.log(win)
+        })
+      })
+
       context('remote jQuery instances', () => {
         beforeEach(function () {
           this.remoteWindow = cy.state('window')
--- a/packages/driver/cypress/e2e/commands/assertions.cy.js
+++ b/packages/driver/cypress/e2e/commands/assertions.cy.js
@@ -305,7 +305,7 @@ describe('src/cy/commands/assertions', () => {
       // https://github.com/cypress-io/cypress/issues/22587
       it('does not allow cypress commands inside the callback', (done) => {
         cy.on('fail', (err) => {
-          expect(err.message).to.eq('`cy.should()` failed because you invoked a command inside the callback. `cy.should()` retries the inner function, but commands are already retried. Use `cy.then()` instead of `cy.should()`, or move any commands outside the callback function.\n\nThe command invoked was:\n\n  > `cy.log()`')
+          expect(err.message).to.eq('`cy.should()` failed because you invoked a command inside the callback. `cy.should()` retries the inner function, which would result in commands being added to the queue multiple times. Use `cy.then()` instead of `cy.should()`, or move any commands outside the callback function.\n\nThe command invoked was:\n\n  > `cy.log()`')
 
           done()
         })
--- a/packages/driver/src/cy/commands/asserting.ts
+++ b/packages/driver/src/cy/commands/asserting.ts
@@ -34,11 +34,24 @@ export default function (Commands, Cypress, cy, state) {
   const shouldFnWithCallback = function (subject, fn) {
     state('current')?.set('followedByShouldCallback', true)
 
+    const commandEnqueued = (obj) => {
+      $errUtils.throwErrByPath(
+        'should.command_inside_should', {
+          args: { action: obj.name },
+        },
+      )
+    }
+
     return Promise
     .try(() => {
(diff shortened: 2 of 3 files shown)


packages/driver/cypress/e2e/commands/assertions.cy.js
```diff
@@ -302,6 +302,19 @@ describe('src/cy/commands/assertions', () => {
         })
       })
 
+      // https://github.com/cypress-io/cypress/issues/22587
+      it('does not allow cypress commands inside the callback', (done) => {
+        cy.on('fail', (err) => {
+          expect(err.message).to.eq('`cy.should()` failed because you invoked a command inside the callback. Use `cy.then()` instead of `cy.should()`, or move any commands outside the callback function.\n\nThe command invoked was:\n\n  > `cy.log()`')
```
flotwig, line 308: Maybe we can educate the user as to why?

```suggestion
          expect(err.message).to.eq('`cy.should()` failed because you invoked a command inside the callback. `cy.should()` retries the inner function, but commands are already retried. Use `cy.then()` instead of `cy.should()`, or move any commands outside the callback function.\n\nThe command invoked was:\n\n  > `cy.log()`')
```

packages/driver/src/cy/commands/asserting.ts
```diff
@@ -34,12 +34,25 @@ export default function (Commands, Cypress, cy, state) {
...
+          args: { action: obj.name },
+        },
+      )
+    }
+
     return Promise
     .try(() => {
       const remoteSubject = cy.getRemotejQueryInstance(subject)
 
+      Cypress.once('command:enqueued', commandEnqueued)
```
flotwig, line 49: Users would have no issues using `cy` commands inside of `should` as long as the command doesn't `verifyUpcomingAssertions` right? So lots of commands like `cy.intercept()`, `cy.setCookies`, etc. would all work in `.should`... It does seem like this has the potential to break a lot of tests that are innocently using these commands in `.should`. I'd propose merging this to `11.0.0` as a breaking change.

BlueWinds: Commands that don't `verifyUpcomingAssertions` still create chainers and can potentially modify the subject; The undefined behavior is more about the command queue (adding commands to it multiple times when .should() retries) and the command log (logging multiple times, each time the command is re-added to the queue) and the subject (when commands resolve as part of another command, the new subject replaces the old one - this causes issues from inside assertions).

"retries inside of retries" isn't the issue, it's "calling cypress commands repeatedly" that's the problem.

packages/driver/cypress/e2e/commands/assertions.cy.js
BlueWinds: Whoops, this only updated the test, not the error_messages. Added commit to address there as well (and also reworded slightly).

packages/driver/src/cy/commands/asserting.ts
flotwig: I think I'm not totally getting it, but in your opinion, are there testing patterns users use that are sane-looking that would break after this change?

BlueWinds: In this particular case, I care more about munging the subject. This is desired behavior:

```
cy.then(() => {
  cy.wrap('foo')
}).should('equal', 'foo')
```

But this is weird behavior: 

```
cy.window().should(subject => {
  expect(subject).to.equal('foo')
  cy.wrap('foo')
})
```

They are both logical consequences of "chainers created inside of other commands become the new subject", but the second one is just *weird*. We should provide guard rails and guide our users towards tests that make sense.

BlueWinds: I don't think there are any sane-looking tests that will break. But also, users do all sorts of things we don't really expect?

I'm inclined to agree with marking this breaking and putting it out as part of 11.x - that was my initial intention, but `break:'` apparently isn't a semantic PR title.

flotwig (approved):
Looks good for 11.0.0.

packages/driver/cypress/e2e/commands/assertions.cy.js
```diff
@@ -302,6 +302,19 @@ describe('src/cy/commands/assertions', () => {
         })
       })
 
+      // https://github.com/cypress-io/cypress/issues/22587
+      it('does not allow cypress commands inside the callback', (done) => {
```
rachelruderman, line 306: This may be a silly question, but if we're being more prescriptive about the use of `should`, should we allow _any_ callbacks with `should`? 

Note: There's probably/definitely use cases I'm not aware of, but I thought `should` was only meant to be used with assertions like `should('exist')` rather than with callbacks

rachelruderman: Ah I see the [docs do show you can use callbacks](https://docs.cypress.io/api/commands/should#Chainers), but would it be better not to? 🤷‍♀️

rachelruderman (approved):

packages/driver/cypress/e2e/commands/assertions.cy.js
BlueWinds: https://docs.cypress.io/api/commands/should#Function has a lot of examples of using a function, and I think "accepting a user-supplied callback function" has fairly wide applicability.

Now that I look at the docs (which I probably should have started with!), it looks like we already have guidance for users around the contents of their functions.

> Be sure not to include any code that has side effects in your callback function. The callback function will be retried over and over again until no assertions within it throw.

So this PR is just being a bit stricter about it, saying "all cypress commands have side effects (on the cypress command queue and the command log, if nothing else)."

rachelruderman: I like your change, helping our users not get stuck in these scenarios in the first place

lmiller1990 (approved):
Nice 💯

emilyrohrbough:
@BlueWinds Does documentation / our examples need updated to explain this change?

PR #23735, opened 2022-09-08
feat: add cohorts for A/B testing in Specs List Banners

### LINKED ISSUE #23083: [ACI] Cohort assignment and persistence to support A/B testing capability for Smart Notifications

### LINKED ISSUE #23118: [ACI] Implement A/B testing versions of copy in Smart Notification banners
# Update smart notification banners to use A/B versions of copy

Issue ( #23083 ) introduces an ability to generate cohorts for users to be used in different use cases within the app.  Use the new cohort feature to select a cohort per banner to use to control which version of copy as shown [here](https://docs.google.com/document/d/1PFN79d3deDyqDVFmVIBe4cVMzWiEll0J6XCpesBZrSg/edit)

See [tech brief](https://github.com/cypress-io/tech-briefs/pull/45)

Smart Notification banners 
| Banner | Section that changes |
| ------- | --------------------- |
| [Login](https://github.com/cypress-io/cypress/issues/22933) | body |
| [Create Organization](https://github.com/cypress-io/cypress/issues/22934) |  header |
| [Connect Project](https://github.com/cypress-io/cypress/issues/22798) |  body |
| [Record](https://github.com/cypress-io/cypress/issues/22935) | action |

warrensplayer:
- Closes https://github.com/cypress-io/cypress/issues/23083
- Closes https://github.com/cypress-io/cypress/issues/23118

### User facing changelog

n/a

### Additional details

In order to be able to A/B test different concepts in the app, an ability to assign a user to a particular cohort is needed.  A cohort can be assigned per feature or functionality that needs to be A/B tested.  

In this first use case, A/B testing is limited to the ACI Smart Prompt banners being shown on the Specs list page of the app.  See the details in https://github.com/cypress-io/cypress/issues/23118 that describes the specific changes.

The cohort functionality is encapsulated in a composable called `useCohorts` in the `frontend-shared` package.  

#### Prerequisites

The cohort functionality is used within the Smart Prompt banners used on the Specs list page of the app.  In order to fully test the different banners, the project used for testing will need to be in a certain state.

- The project should have been opened more than 4 days ago.  This can be confirmed by looking at the `state.json` file under the "App Data" directory and then navigating to "projects" and then the directory with the project name and hash.  The `firstOpened` value should be a timestamp older than 4 days ago
- Remove the `banners` entry from the same file. This will reset the apps tracking of when banners were viewed or dismissed.
- Make sure the project is not connected to a dashboard project id by removed the "lastProjectId" from the same file and the `cypress.config.js` file for the project as appropriate

#### Testing cohorts

* After first starting the app with the launchpad loaded, select the "Developer tools > View App Data"  menu item
* Within the directory that opens, open the `cache` file in a text editor
* Confirm that there is not a "COHORTS" entry in the json data
* Make sure you are logged out from the dashboard

* Login Banner
	* Open the project in either e2e or component testing mode
	* Confirm that the Login Banner shows
	* Reopen the `cache` file from above and confirm there is now a COHORTS entry with a key of `aci_082022_login` that matches the bannerId in `LoginBanner.vue`.  The key should have an object with the name matching the bannerId and a cohort value of either 'A' or 'B'.  Confirm the body copy of the Login Banner matches the appropriate copy for either A or B shown in the document [here](https://docs.google.com/document/d/1PFN79d3deDyqDVFmVIBe4cVMzWiEll0J6XCpesBZrSg/edit#)
	* Navigate away from the Specs list page by going to the Runs page and then back to the Specs list.  The same copy should be showing in the Login Banner from before and the "COHORTS" value in the `cache` file should not have changed
	* Close the browser for the app either manually or from the launch pad.
	* Edit the `cache` file to remove the key for `aci_082022_login`
	* Open the app again in either mode and confirm that the `cache` file has a new entry for `aci_082022_login` and that the appropriate copy is showing
	* If the algorithm selected the same cohort as before, you can test the other option in two ways:
	  * Follow the steps above for a couple of more times to edit the `cache` file and remove the same entry until the algorithm chooses the other cohort
	  * Edit the `cache` file to manually edit the cohort to the other value (`A` or `B`)
*  Remaining banners
  * Test the other Smart Prompt banners as described here: https://github.com/cypress-io/cypress/pull/23256
  * For each banner, confirm the same "cohort" behavior and compare the results with the expected A/B testing values [here](https://docs.google.com/document/d/1PFN79d3deDyqDVFmVIBe4cVMzWiEll0J6XCpesBZrSg/edit#)
  * The keys in the "COHORTS" section should match the bannerIds for each banner
    * `aci_082022_createOrganization`
    * `aci_082022_connectProject`
    * `aci_082022_record`
* UTM Parameters
  * The calculated cohort values should also be passed with the resulting call to actions from each banner.  Run through the tests for this [PR](https://github.com/cypress-io/cypress/pull/23691) to confirm the cohort values are passed in the external links for the Login and Create Organization and in the mutation for creating a project.  The Record Runs banner does not track the cohort when running the command

Once fully implemented, the only user experience change will be that different copy may be used in the banners depending on which cohort the user is assigned to.

### PR Tasks


packages/app/src/specs/banners/TrackedBanner.vue
```diff
@@ -8,9 +8,36 @@
...
+  value: string
+}
+
+export const getOptionForCohort = (name: string, options: CohortOption[]) => {
+  const cohortConfig: CohortConfig = {
+    name,
+    cohorts: options.map((option) => option.cohort),
+  }
+
+  const cohortSelected = useCohorts(cohortConfig)
```
mike-plummer, line 25: This feels error-prone - using a composable within a plain JS function will blow up if somebody happens to try use it outside a component (in a unit test, for example). Worth extracting this into the `useCohorts` composable so all the cohort logic is in one place?

packages/app/src/specs/banners/TrackedBanner.vue
```diff
@@ -8,9 +8,36 @@
...
+
+  return computed(() => {
+    if (cohortSelected.value) {
+      return options.filter((option) => option.cohort === cohortSelected.value)[0]
```
mike-plummer, line 29: `options.find((option) => option.cohort === cohortSelected.value)`

packages/app/src/specs/banners/TrackedBanner.vue
```diff
@@ -67,7 +94,7 @@ const setStateMutation = useMutation(TrackedBanner_SetProjectStateDocument)
 const reportSeenMutation = useMutation(TrackedBanner_RecordBannerSeenDocument)
 const bannerInstanceId = ref(nanoid())
 
-watchEffect(() => {
+watch(() => props.eventData?.cohort, () => {
```
mike-plummer, line 97: This feels wrong. What about a situation where the banner is initially not visible (`modelValue === false`) then later made visible - this watcher won't fire again for that change, and we'll never record the event.

packages/frontend-shared/src/composables/examples/UseCohortsExample.vue
```diff
@@ -0,0 +1,42 @@
...
+  name: 'login',
+  cohorts: props.copyOptions.map((option) => option.cohort),
+  algorithm: props.algorithm,
+}
+
+const cohortSelected = useCohorts(cohortConfig)
+
+const cohortChoice = computed(() => {
+  if (cohortSelected.value) {
+    return props.copyOptions.filter((option) => option.cohort === cohortSelected.value)[0]
```
mike-plummer, line 34: `props.copyOptions.find((option) => option.cohort === cohortSelected.value)`

packages/frontend-shared/src/composables/examples/UseCohortsExample.vue
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  return { value: '' }
+})
+
+//http://localhost:5173/__/#/specs/runner?file=src/composables/examples/UseCohortsExample.cy.tsx
```
mike-plummer, line 40: Remove?

packages/frontend-shared/src/utils/weightedChoice.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Randomly choose an index from an array based on weights
+ *
+ * Based on algorithm found here: https://dev.to/trekhleb/weighted-random-algorithm-in-javascript-1pdc
+ *
+ * @param weights array of numbered weights that correspond to the indexed values
+ * @param values array of values to choose from
+ */
+const weightedChoice = (weights: number[], values: any[]) => {
```
mike-plummer, line 15: Do we need to validate the lengths of the arrays match?

packages/graphql/src/schemaTypes/objectTypes/gql-Cohorts.ts
```diff
@@ -0,0 +1,10 @@
+import { inputObjectType, objectType } from 'nexus'
+
+export const Cohort = objectType({
+  name: 'Cohort',
+  description: 'used to distinguish one group of users from another',
+  definition (t) {
+    t.string('name', { description: 'name used to identify the cohort topic (e.g. "LoginBanner" ) ' })
```
mike-plummer, line 7: I already commented on this, but I keep getting tripped up thinking this is the name of an individual cohort rather than the name for a grouping of cohorts

packages/graphql/src/schemaTypes/objectTypes/gql-Cohorts.ts
```diff
@@ -0,0 +1,10 @@
...
+    t.string('cohort', { description: 'value used to identify the indicate the cohort (e.g. "A" or "B")' })
```
mike-plummer, line 8: grammar: "identify the indicate"

mike-plummer (changes requested):

packages/app/src/specs/banners/ConnectProjectBanner.vue
```diff
@@ -83,4 +91,14 @@ function handleModalClose () {
...
 
+const optionSelected = getOptionForCohort(bannerId, props.bodyCopyOptions)
+
+const bodyCopy = computed(() => {
+  if (optionSelected.value.value) {
+    return t(optionSelected.value.value)
+  }
+
+  return ''
+})
```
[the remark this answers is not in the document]
warrensplayer: I like the streamline idea.  There is one banner that does not do the internationalization because it results in a different `run` command.

packages/app/src/specs/banners/ConnectProjectBanner.vue
```diff
@@ -55,19 +55,27 @@ query ConnectProjectBanner {
...
+  bodyCopyOptions?: CohortOption[]
 }>(), {
   modelValue: false,
   hasBannerBeenShown: true,
+  bodyCopyOptions: () => {
+    return [
+      { cohort: 'A', value: 'specPage.banners.connectProject.contentA' },
+      { cohort: 'B', value: 'specPage.banners.connectProject.contentB' },
+    ]
+  },
```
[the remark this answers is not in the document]
warrensplayer: Yes, the prop was just for testablity because of the randomness.  I will try out injecting the algorithm instead.  I think I see where you are going with that.

packages/app/src/specs/banners/TrackedBanner.vue
warrensplayer: Good catch there.  I had to change this from the watchEffect to the watch because I did not want it to fire before the cohort was set.  I will test out the scenario of when that modalValue changes and see if it records the event.  The logic is definitely not as clear cut as I would like, but I think it will still work.  Will the `props.modalValue` ever be false here?  I believe the `v-else-if` on the `LoginBanner` in the `SpecsListBanners` will prevent this code from running unless the `props.modalValue` is already true, right?

packages/app/src/specs/banners/TrackedBanner.vue
```diff
@@ -8,9 +8,36 @@
...
 
+<script lang="ts">
+import { CohortConfig, useCohorts } from '@packages/frontend-shared/src/composables/useCohorts'
+
+export type CohortOption = {
+  cohort: string
+  value: string
+}
+
+export const getOptionForCohort = (name: string, options: CohortOption[]) => {
```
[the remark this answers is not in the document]
warrensplayer: Which param are you talking about?  I do like the "experiement" term though.  Definitely open to more clear parameter names.

packages/app/src/specs/banners/TrackedBanner.vue
warrensplayer: I knew there was something that would get the first value out of the array but `find` was not coming to mind. Thanks!

packages/frontend-shared/src/composables/examples/UseCohortsExample.vue
warrensplayer: Yes!

packages/frontend-shared/src/composables/examples/UseCohortsExample.vue
warrensplayer: Thanks agin

packages/frontend-shared/src/utils/weightedChoice.ts
warrensplayer: Yes, that would be a good idea.  I will add that check.

packages/app/src/specs/banners/RecordBanner.vue
```diff
@@ -58,29 +58,44 @@ query RecordBanner {
...
+const props = withDefaults(defineProps<{
   modelValue: boolean
   hasBannerBeenShown: boolean
+  commandOptions?: CohortOption[]
 }>(), {
   modelValue: false,
   hasBannerBeenShown: true,
+  commandOptions: () => {
+    return [
+      { cohort: 'A', value: 'cypress' },
```
lmiller1990, line 70: Is running `cypress` without `npx` really advisable? Most people won't have it installed globally. Also, I did not see this A/B test specified in the docs of Figma - is it there somewhere?

packages/app/src/specs/banners/TrackedBanner.cy.tsx
```diff
@@ -25,7 +25,7 @@ describe('<TrackedBanner />', () => {
 
     // Initially mount as visible
     // @ts-ignore
-    cy.mount({ render: () => <TrackedBanner data-cy="banner" bannerId="test-banner" v-model={shown.value} hasBannerBeenShown={false} /> })
+    cy.mount({ render: () => <TrackedBanner data-cy="banner" bannerId="test-banner" v-model={shown.value} hasBannerBeenShown={false} eventData={{} as any}/> })
```
lmiller1990, line 28: Should we just pass in some real data? The interface is pretty simple:

```js
type EventData = {
  campaign: string
  medium: string
  cohort?: string
}
```

packages/app/src/specs/banners/TrackedBanner.vue
lmiller1990: I wonder if there's a race condition here? Since `useCohorts` makes an API call, it's actually async - so this could potentially be undefined, then the API call finishes, then this would re-compute.

As an aside, surprisingly, this would not blow up in a unit test - Vue's reactivity is decoupled from the renderer, so you can actually use it outside the context of Vue - it's really "just javascript" (except it's not really clear what you'd want to use it outside of Vue for, I guess, except for writing a unit test - at that point, I'd argue you extract the logic out of the composable, and unit test that).

packages/app/src/specs/banners/TrackedBanner.vue
```diff
@@ -8,9 +8,38 @@
...
+    name,
+    cohorts: options.map((option) => option.cohort),
+  }
+
+  const cohortSelected = useCohorts(cohortConfig)
+
+  const emptyOption = { cohort: '', value: '' }
+
+  return computed(() => {
+    if (cohortSelected.value) {
```
lmiller1990, line 30: Since `cohortSelected` is set async after a GraphQL mutation runs, if, for whatever reason, is there any problem if `cohortSelected` is never defined (dashboard down, slow network, some random unexpected error)? Or everything just continues along with no noticeable impact on the user?

packages/app/src/specs/banners/TrackedBanner.vue
```diff
@@ -8,9 +8,38 @@
   </Alert>
 </template>
 
+<script lang="ts">
```
lmiller1990, line 11: Should we just put this in a separate file? I know this is valid but I find it strange to have multiple `<script>` tags in a single Vue file.

packages/frontend-shared/src/composables/examples/UseCohortsExample.cy.tsx
```diff
@@ -0,0 +1,21 @@
...
+    { cohort: 'B', value: 'Notification Title B' },
+  ]
+
+  it('should show value for one cohort with default algorithm', () => {
+    cy.mount(() => <UseCohortsExample copyOptions={copyOptions}/>)
+    cy.findByTestId('result').contains('Notification Title')
+  })
+
+  it('should show value for one cohort with supplied algorithm', () => {
+    const weighted25_75 = WEIGHTED([25, 75])
```
lmiller1990, line 16: I can't see where we actually choose the cohort - is this in another PR somewhere? Eg - what is the ratio between A and B, and how often am I in A vs B?

I wasn't as deeply involved in the finer details of the ACI project, I may have missed something obvious that has already been discussed somewhere.

packages/graphql/src/schemaTypes/objectTypes/gql-Cohorts.ts
lmiller1990: Any suggestions for a better name? My initial instinct was this would be something like `currentUserCohort`.

We also have. the field `cohort`, so when you query off this it is

```gql
cohort {
  cohort
  name
}
```

I wonder if the top level should be called something different. 

If we do want to include info about the current user cohort, I suppose it'd be on the `cachedUser` field

```gql
cachedUser {
  cohort {
    name
    # ...
  }
}
```

I don't think the naming is too bad, but I have not worked on ACI as much, so take my opinion with a grain of salt.

lmiller1990 (commented):
Code and coverage looks thorough! I left some questions, it's definitely tricky to manage  everything with all the async fetching + UI layer and finding the best way to manage it all.

lmiller1990:
Another thought: we have a lot of complicated manual testing for most ACI PRs. I don't see any reason we cannot invest a bit of time into some better testing infra, just in the runner - we can add some endpoints to update the cache or app data if needed, and some to force the app to refetch the data. I don't see much other than "re-open the app" here we can't automate. Just an idea, it's possible there is technical blockers I'm not aware of.

packages/frontend-shared/src/composables/examples/UseCohortsExample.cy.tsx
warrensplayer: The WeightedAlgorithm is what actually selects the cohort in `useCohorts`.  The `fetchCohort` function in `useCohorts` is where the logic occurs.  The logic checks the `cache` file to see if a cohort has already been selected for a given "name" (this is the term that Mike commented is confusing, which I agree with).  If the cohort for that name has already been selected, then it just returns that value (the "else" case), if it has not been selected, then it uses the configured weighted algorithm to choose one (`algorithm.pick`) and then store it in the cache and return that choice.  

The default algorithm is the `WEIGHTED_EVEN` one which is being used by the Spec list banners to provide even weight to either copy choice being tested in the banners.  I provided a way to customize the weighting after hearing about a potential future case where the A/B testing would be run on a 25/75 weighting.

packages/app/src/specs/banners/RecordBanner.vue
warrensplayer: The requirements for the A/B testing of the command was defined in the document describing the other A/B copy options here: https://docs.google.com/document/d/1PFN79d3deDyqDVFmVIBe4cVMzWiEll0J6XCpesBZrSg/edit#

The documentation here (https://docs.cypress.io/guides/guides/command-line) mentions `npx` along side the various other ways to run cypress.  Currently, the `<RunsEmpty />` component just shows the command without the `npx` option on it. 

Do you think we should be encouraging one use over another?

packages/app/src/specs/banners/TrackedBanner.vue
warrensplayer: I felt weird about doing it this way as well.  Looking at pulling this logic into the `useCohort` composable, although this logic is more implementation specific (even though it is the only implementation at the moment).

packages/app/src/specs/banners/ConnectProjectBanner.vue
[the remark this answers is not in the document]
warrensplayer: Left this alone for now, but can look at a refactor if this gets touched again.

packages/app/src/specs/banners/TrackedBanner.cy.tsx
warrensplayer: Just following existing pattern where the values of this eventData do not matter for this particular test.

packages/app/src/specs/banners/TrackedBanner.vue
warrensplayer: Moved this logic into the composable

warrensplayer: Moved this logic into the composable

packages/app/src/specs/banners/TrackedBanner.vue
warrensplayer: Moved to useCohorts composable

packages/graphql/src/schemaTypes/objectTypes/gql-Cohorts.ts
warrensplayer: I added documentation comments to useCohorts. Hopefully that helps with the names

warrensplayer:
> Another thought: we have a lot of complicated manual testing for most ACI PRs. I don't see any reason we cannot invest a bit of time into some better testing infra, just in the runner - we can add some endpoints to update the cache or app data if needed, and some to force the app to refetch the data. I don't see much other than "re-open the app" here we can't automate. Just an idea, it's possible there is technical blockers I'm not aware of.

@lmiller1990 Agreed about the complexity f the ACI testing.  I have it on my TODO to look at better testing for this.  I will reach out to discuss more and get your input.

packages/app/src/specs/banners/RecordBanner.vue
pstakoun: @warrensplayer there's no way to track the performance of the A/B test of this record run step, right? If so, I think we should just not do it and show `npx` for all cases.

packages/app/src/specs/banners/TrackedBanner.vue
mike-plummer: Given how things are currently structured, yes, `v-model` will always *happen* to be true when initially rendered, but there's no guarantee we won't add a new banner next month someplace else that doesn't rely on a `v-if` and uses `v-model` dynamically. There's no way at this layer for the component to tell the difference between "I have no cohort" vs "I haven't yet received a value for cohort" - that would need to be communicated by refactoring this (another prop to communicate `hasCohortResolved`, or a composable that uses `undefined` vs `null` to communicate "none" vs "not yet".

In any event, this watcher now has a couple gaps that are relying on race conditions. What happens if cohort resolves before the banner displays? The watcher will never fire because it doesn't have `immediate: true` - that's part of the reason it was a `watchEffect` before

packages/app/src/specs/banners/RecordBanner.vue
warrensplayer: @pstakoun That is correct.  We cannot track the performance of users running these command line options at this time.  I will remove this for now.

packages/data-context/src/actions/CohortsActions.ts
```diff
@@ -0,0 +1,20 @@
...
+export class CohortsActions {
+  constructor (private ctx: DataContext) {}
+
+  async getCohort (name: string) {
+    return this.ctx._apis.cohortsApi.getCohort(name)
+  }
+
+  async insertCohort (cohort: Cohort) {
+    return this.ctx._apis.cohortsApi.insertCohort(cohort)
+  }
```
mike-plummer, line 19: Some `debug` logging in here would be a nice to have, not a blocker

packages/app/src/specs/banners/RecordBanner.vue
warrensplayer: @pstakoun On second thought, I will discuss with you shortly during our meeting on options.

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
```diff
@@ -684,6 +684,25 @@ export const mutation = mutationType({
...
+      },
+      resolve: async (source, args, ctx) => {
+        const cohort = {
+          name: args.name,
+          cohort: args.cohort,
+        }
+
+        await ctx.actions.cohorts.insertCohort(cohort)
+
+        return true
```
mike-plummer, line 702: Is the expectation to just throw in the event of an error? Or should this catch and resolve `false`?

packages/frontend-shared/src/utils/weightedChoice.ts
mike-plummer: Non-blocker: Validate lengths > 0

packages/frontend-shared/src/composables/useCohorts.ts
```diff
@@ -0,0 +1,102 @@
...
+    const cohortFromCache = await getCohortFromCache(config.name)
+
+    let cohortSelected: string
+
+    if (!cohortFromCache || !_.includes(cohortIds, cohortFromCache)) {
+      const algorithm = config.algorithm || WEIGHTED_EVEN(cohortIds)
+
+      const pickedCohort = algorithm.pick(cohortIds)
+
+      setCohortInCache(config.name, pickedCohort)
```
mike-plummer, line 90: Need to await here?

packages/app/src/specs/banners/TrackedBanner.vue
warrensplayer: So, the current issue which breaks the recording of the event is that the `hasBannerBeenShown` is first false but then set to true before this watchEffect gets triggered.  Rethinking how to order these operations to prevent this race condition.

packages/frontend-shared/src/composables/useCohorts.ts
warrensplayer: Yes!

packages/app/src/specs/banners/ConnectProjectBanner.vue
```diff
@@ -55,19 +56,27 @@ query ConnectProjectBanner {
...
 
-withDefaults(defineProps<{
+const props = withDefaults(defineProps<{
   modelValue: boolean
   hasBannerBeenShown: boolean
+  bodyCopyOptions?: CohortOption[]
 }>(), {
   modelValue: false,
   hasBannerBeenShown: true,
+  bodyCopyOptions: () => {
```
ZachJW34, line 66: Is this prop only used for testing purposes? Adding a comment would be useful here if so, I see this pattern on a few components (I assume it's to remove the random choice between A or B cohort)

packages/frontend-shared/src/composables/examples/UseCohortsExample.cy.tsx
```diff
@@ -0,0 +1,21 @@
+import { WEIGHTED } from '../../utils/weightedChoice'
```
ZachJW34, line 1: What does this test prove? Since it comes down to randomness, all you can assert on is that `Notification Title` is present which doesn't add much value, especially since you've hooked up `useCohort` in other CT tests.

packages/frontend-shared/src/composables/useCohorts.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  const fetchCohort = async () => {
+    const cohortFromCache = await getCohortFromCache(config.name)
+
+    let cohortSelected: string
+
+    if (!cohortFromCache || !_.includes(cohortIds, cohortFromCache)) {
+      const algorithm = config.algorithm || WEIGHTED_EVEN(cohortIds)
+
+      const pickedCohort = algorithm.pick(cohortIds)
```
ZachJW34, line 88: Was there a reason the `pick` isn't done on the server? We have to make two network requests, one to get the existing cohort and another to set if it doesn't exist. Couldn't we make a request and on the server generate a cohort if it doesn't exist? Would simplify this code a bit

ZachJW34:
Tested the various scenarios and all worked well! I agree with the sentiment about testing these PRs, it takes a lot of time to walk through all the steps

packages/app/src/specs/banners/RecordBanner.vue
warrensplayer: After talking with Peter, we are going to remove the A/B test and always show the `npx cypress` option.

astone123:
Tested the various banners and everything looks good to me. One question about the `utmSource` parameter. I wonder if there's a way for us to differentiate between dev traffic and production traffic based on the `CYPRESS_INTERNAL_ENV` variable.

For instance, I just logged in and saw that `utmSource: "Binary: App"` was sent to the production cloud environment. Since my environment is dev, it would be nice if we had a way to not muddy the waters with the data being sent to production. Just a thought, might be able to handle it in an issue down the road.

mike-plummer:
@astone123 I think the only way around that would be to put in a "kill switch" to avoid appending UTM params to the URLs when running locally. We're inherently "crossing the streams" by loading data from Production when running locally, and in order to load production data you have to log in to Prod. We could probably partially-solve this by filtering out users who are members of the Cypress org, but that wouldn't cover every internal user

astone123:
@mike-plummer yeah this does get complex because we want to show data **from** production, but we ideally don't want to send any data **to** production (...except for login, because that's how we get the prod data 🤔). 

I'm probably not understanding what `utmSource` is supposed to be used for, but it seems like we could add some information to say "hey, this action was performed in the context of `CYPRESS_INTERNAL_ENV === 'dev'`", which would allow us to filter out any actions that were taken in that context when we gather metrics from prod.

astone123:
For anyone reading the exchange above - @mike-plummer created an issue for this https://github.com/cypress-io/cypress/issues/23790

packages/frontend-shared/src/composables/useCohorts.ts
warrensplayer: Moving to server

packages/data-context/src/actions/CohortsActions.ts
warrensplayer: Adding debug info

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
warrensplayer: Updated to return cohort

packages/app/src/specs/banners/ConnectProjectBanner.vue
warrensplayer: Removing this prop

packages/frontend-shared/src/composables/examples/UseCohortsExample.cy.tsx
warrensplayer: I felt like it would be valuable to at least test that the composable was returning one of the options it was given.

packages/app/src/specs/banners/TrackedBanner.vue
warrensplayer: Updated v-else-if condition on when to show the banner to only show once the cohort has been set

warrensplayer:
**Reactor updates to fix race condition**
I have completed the refactor to address the race condition between selecting and storing a cohort for a given banner and sending the "viewed" event for the first view.  This refactor pulled the decision point for selecting the cohort out of `TrackedBanner` and brought it up to `SpecsListBanners`.  The individual banners are now passed the selected cohort as a prop along with the value to use that is being A/B tested. The banner component is responsible for putting the value in the appropriate location and passing the cohort value to the `TrackedBanner` component that sends the event.  The cohort is also used as either a utm parameter for an external URL or in the mutation to create a project.  

The refactor also pushed the logic to choose the cohort to the data-context layer.  This is encapsulated in the `CohortsActions`.

packages/app/src/specs/SpecsListBanners.vue
```diff
@@ -272,6 +276,46 @@ watch(
...
+const getCohortForBanner = (bannerId: string) => {
+  const cohortConfig: CohortConfig = {
+    name: bannerId,
+    options: bannerCohortOptions[bannerId],
+  }
+
+  return useCohorts(cohortConfig)
+}
+
+const cohorts = reactive({} as Record<'login' | 'connectProject' | 'organization', any>)
```
lmiller1990: I have some questions about this (sorry I only just noticed as part of the re-review...)

This type and usage felt a little off. I found out why - we have a kind of "double reactivity" here. We are using `reactive` to hold properties which all point to `Ref` (which is also part of the reactivity system).

It's similar to doing something like

```js
const [foo1, setFoo1] = useState()
const [foo1, setFoo1] = useState()

const [foos, setFoos] = useState()
setFoos({ setFoo1, setFoo2 })
```

I'm not sure if this really makes sense. I tried a few things, and landed on https://github.com/cypress-io/cypress/compare/stokes/23083_aci_cohorts...lmiller/stokes/23083_aci_cohorts-v1?expand=1.

```ts
type BannerType = 'login' | 'connectProject' | 'organization'
const cohorts: Partial<Record<BannerType, ReturnType<typeof getCohortForBanner>>> = {}
```

Alternatively, just have 3 separate variables:

```ts
const cohortLogin = ref<ReturnType<typeof getCohortForBanner> | undefined>()
// ...
```

This might be more simple - avoid thinking about the reactivity problem all together.
\
I don't know if `cohorts` actually needs to be `reactive`. Either way, this change revealed a bug - `getCohortForBanner` which returns `useCohorts` which returns `ref<CohortOption | undefined>()`. I got a TS error, which I fixed by doing an additional null check (see above link). 

The properties (`cohorts.login` etc) are nullable - but we are passing them to `<ConnectProjectBanner />` etc, which has a non-nullable prop type. I think we need this null check.

We missed this because of the `as Record<any>` cast. Can a user be in *no* cohort? If they are, I don't know if we'd want to display the modals.

Not sure if this actual manifested in a functional bug, but definitely worth clarifying + improving the types (ideally avoiding `any`).

packages/app/src/specs/banners/TrackedBanner.vue
```diff
@@ -68,7 +68,7 @@ const reportSeenMutation = useMutation(TrackedBanner_RecordBannerSeenDocument)
 const bannerInstanceId = ref(nanoid())
 
 watchEffect(() => {
-  if (props.modelValue && !props.hasBannerBeenShown && props.eventData) {
+  if (props.modelValue && !props.hasBannerBeenShown && props.eventData && props.eventData?.cohort) {
```
lmiller1990:
```suggestion
  if (props.modelValue && !props.hasBannerBeenShown && props.eventData?.cohort) {
```

lmiller1990: I don't think we need the additional `&&`?

packages/data-context/src/DataActions.ts
```diff
@@ -83,4 +84,9 @@ export class DataActions {
   get eventCollector () {
     return new EventCollectorActions(this.ctx)
   }
+
+  @cached
+  get cohorts () {
+    return new CohortsActions(this.ctx)
```
lmiller1990, line 90: Nice, I think this makes sense

lmiller1990 (commented):
I've got one question, there's still something a little off about how we manage the different banners/cohorts - I left a reply, we might want to sync on this one?

packages/app/src/specs/SpecsListBanners.vue
```diff
@@ -228,10 +232,10 @@ const showConnectBanner = ref(false)
 const showCreateOrganizationBanner = ref(false)
 const showLoginBanner = ref(false)
 
-const hasRecordBannerBeenShown = ref(true)
-const hasConnectBannerBeenShown = ref(true)
-const hasCreateOrganizationBannerBeenShown = ref(true)
-const hasLoginBannerBeenShown = ref(true)
+const hasRecordBannerBeenShown = ref(false)
```
lmiller1990, line 235: I am looking forward to moving these into the store. It will massively simplify things... eg instead of the huge v-if/else, and we emit an event to update these, the components themselves could be responsible for updating this. Inside the `<RecordBanner />` we could just do `store.hasRecordBannerBeenShown = true`, right now the `<SpecsListBanner />` IS the store.

warrensplayer: Agreed!

packages/app/src/specs/banners/TrackedBanner.vue
warrensplayer: You are correct and I thought I had done that.  I am going crazy with this!

packages/data-context/src/util/weightedChoice.ts
```diff
@@ -0,0 +1,55 @@
...
+  const choice = _.transform(cumulativeWeights, (result, value, index) => {
+    if (value >= randomNumber) {
+      result.chosenIndex = index
+    }
+
+    return result.chosenIndex === -1
+  }, { chosenIndex: -1 })
+
+  return values[choice.chosenIndex]
+}
```
warrensplayer: Any ideas for not needing the two `@ts-ignores` I added for referencing the last value in an array?

packages/app/src/specs/SpecsListBanners.vue
```diff
@@ -272,6 +276,46 @@ watch(
...
+  ],
+}
+
+const getCohortForBanner = (bannerId: string) => {
+  const cohortConfig: CohortConfig = {
+    name: bannerId,
+    options: bannerCohortOptions[bannerId],
+  }
+
+  return useCohorts(cohortConfig)
```
ZachJW34: After I login, I'm getting this error that freezes the app:
Screen Shot 2022-09-15 at 9 43 26 AM

Looks like it's due to the recent refactor. Does this `useCohorts` have to be a composable or can it be converted to be a simple utility function?

packages/app/src/specs/SpecsListBanners.vue
warrensplayer: I do not believe there is a double reactivity problem because Vue unwraps refs in side `reactive`. See: https://vuejs.org/guide/essentials/reactivity-fundamentals.html#ref-unwrapping-in-reactive-objects

Also, I do still need the cohorts to be reactive because of the mutation it uses is async.  I was following this pattern for my composable: https://www.vuemastery.com/blog/coding-better-composables-5-of-5/

The problem with using `reactive` and trying to strongly type the value part of the `Record` is the unwrapping the `reactive` does.  You can assign either a `ref` or the unwrapped value any place inside a `reactive` object, but would always access it as unwrapped.  

I liked using `reactive` because of the unwrapping so you would not have to do all the `.value` calls in the template.  I thought about the separate variables as well, but they would have to be double nested `ref`s or just be defined with `let` and set after the `getCohortForBanner` returns.  

I am checking in an update that looks pretty ugly, but does seem to work that defines the `cohorts` variable like:
```
type BannerType = 'login' | 'connectProject' | 'organization'

const cohorts: Partial<Record<BannerType, CohortOption | Ref<CohortOption | undefined>>> = reactive({})
```

Let me know what you think.

packages/app/src/specs/SpecsListBanners.vue
warrensplayer: Argh.  I will take a look

warrensplayer: Fix committed and pushed.  Working on a test now to validate the fix.

packages/app/src/specs/SpecsListBanners.vue
warrensplayer: Nevermind.... Went back to the `any` for the moment since that did introduce another type check issue.

packages/data-context/src/actions/CohortsActions.ts
```diff
@@ -1,7 +1,11 @@
 import type { Cohort } from '@packages/types'
 import type { DataContext } from '..'
+import { WEIGHTED, WEIGHTED_EVEN } from '../util/weightedChoice'
+const debug = require('debug')('cypress:data-context:cohorts')
```
mike-plummer: Zach B schooled me on this a few days ago - we have [rules for how to namespace debug loggers](https://github.com/cypress-io/cypress/blob/develop/guides/debug-logs.md#choosing-a-namespace). This should probably be `cypress:data-context:actions:CohortActions`

packages/data-context/src/actions/CohortsActions.ts
```diff
@@ -1,7 +1,11 @@
...
 
 export interface CohortsApiShape {
+  getCohorts(): Promise<Record<string, Cohort>>
+
   getCohort(name: string): Promise<Cohort>
```
mike-plummer: I think there's a hole in the typings - couldn't this be nil if you try to get a cohort that hasn't been inserted yet? Not sure whether it makes sense to tighten up the typings in `server/lib/cohorts.ts` (everything coming out of there is defined as `any`), or better to just ensure the `CohortsApiShape` is properly reflecting the expected types

packages/data-context/src/util/weightedChoice.ts
```diff
@@ -1,7 +1,7 @@
 import _ from 'lodash'
 
 export type WeightedAlgorithm = {
-  pick: Function
+  pick: (values: any[]) => string
```
mike-plummer: This is picking an element out of the `values` array and returning it, right? If so the types are incorrect here. Either should be `(values: string[]) => string` if it guarantees `string` or `<T>(values: T[]) => T` to allow returning whatever the type of the input array is. This would then flow down into your `weightedChoice` function below where there are a couple strict TS issues being disguised by the implicit `any` in play from this.

packages/data-context/src/util/weightedChoice.ts
```diff
@@ -13,20 +13,22 @@ export type WeightedAlgorithm = {
  * @param values array of values to choose from
  */
 const weightedChoice = (weights: number[], values: any[]) => {
-  if (weights.length !== values.length) {
-    throw new Error('The length of the weights and values must be the same')
+  if (weights.length > 0 && values.length > 0 && weights.length !== values.length) {
+    throw new Error('The length of the weights and values must be the same and greater than zero')
   }
```
mike-plummer: This doesn't validate that the arrays are non-empty

```suggestion
  if (weights.length === 0 || values.length === 0 || weights.length !== values.length) {
    throw new Error('The `weights` and `values` parameters must be the same non-zero length')
  }
```

packages/app/src/specs/SpecsListBanners.vue
```diff
@@ -272,6 +276,50 @@ watch(
...
+const cohortBuilder = useCohorts()
+
+const getCohortForBanner = (bannerId: string) => {
+  const cohortConfig: CohortConfig = {
+    name: bannerId,
+    options: bannerCohortOptions[bannerId],
+  }
+
+  return cohortBuilder.getCohort(cohortConfig)
+}
```
mike-plummer, line 303: nitpick: Just to avoid cluttering up this component the actual list of options and the function to select based on `bannerId` could get pulled out to a utility under `/specs/banners`, or even a composable

mike-plummer (changes requested):
Getting real close here, just a few TS and organization things

packages/data-context/src/util/weightedChoice.ts
warrensplayer: Yes, values should be a `string[]`

packages/data-context/src/util/weightedChoice.ts
warrensplayer: Not sure where my brain was for this.  Will fix and add tests.

packages/data-context/src/actions/CohortsActions.ts
warrensplayer: Thansk for pointing this out.  Will update

packages/data-context/src/actions/CohortsActions.ts
warrensplayer: Lachlan explained that part of the history of the code base was that the server was not well typed and was hard to introduce types, so part of the what the data-context package is doing is putting a typed layer on top of the server.  I will update the `CohortsApiShape` types to reflect what is expected.

packages/app/src/specs/SpecsListBanners.vue
warrensplayer: I will look at better organizing this with the refactoring work.  I wish I did not have any cohort logic in this component but needed it here to control showing the banners at the right time.

packages/app/src/specs/SpecsListBanners.vue
lmiller1990: Right I see why `cohorts` needs to be `reactive` now - the refs are initially `undefined` and (possibly) assigned later on.

I think the original problem still stands doesn't it? By using `any` are we missing out on some value typescript static checking?


I won't hold this one up too much more - I could be missing something, I thought this would be straight forward to type, but maybe that's not the case.

warrensplayer: Ended up going with your answer but just left off the last part of the null check since the CohortOption, if present, will always have a cohort

packages/frontend-shared/src/composables/useCohorts.ts
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Composable that encapsulates the logic for choosing a cohort from a list of configured options.
+ *
+ * @remarks
+ * The logic for this composable will first check the cache file to determine if a cohort has already been saved for the given cohort `name`. If found, that cohort will be returned.  If not found or the option found does not match an existing option, a weighted algorithm will be used to pick from the list of CohortOptions. The picked value will be stored in the cache and returned.
+ *
+ * @returns object with getCohort function for returning the cohort
+ */
+export const useCohorts = () => {
```
ZachJW34, line 47: After the refactoring, I feel like the better pattern here would have been to use `pinia`, where the `cohortOptionSelected` would have been stored as `state` and the `getCohort` would be an `action`. It seems strange that a `composable` is returning a function and not reactive state.

ZachJW34 (approved):
Gave it one last test and it works great. I've got one comment about the `useCohort` composable, but not going to block on it. Once CI and some TS issues are fixed, LGTM!

lmiller1990 (approved):

packages/frontend-shared/src/composables/useCohorts.ts
warrensplayer: Can a `store` make a GraphQL mutation?  That is what I was trying to encapsulate in `useCohorts`?

mike-plummer (approved):

packages/frontend-shared/src/composables/useCohorts.ts
ZachJW34: You should be able to use `useMutation` inside an action, see https://pinia.vuejs.org/core-concepts/actions.html#accessing-other-stores-actions where they utilize composables inside pinia actions.

packages/data-context/src/util/weightedChoice.ts
lmiller1990: Hmm maybe this:

Only line I am unsure about

```js
const randomNumber = Math.random() * (cumulativeWeights[cumulativeWeights.length - 1] ?? 1)
```

Should we error if `cumulativeWeights[cumulativeWeights.length - 1]` is null? Should be impossible?

```js
import _ from 'lodash'

export type WeightedAlgorithm = {
  pick: (values: string[]) => string
}

/**
 * Randomly choose an index from an array based on weights
 *
 * Based on algorithm found here: https://dev.to/trekhleb/weighted-random-algorithm-in-javascript-1pdc
 *
 * @param weights array of numbered weights that correspond to the indexed values
 * @param values array of values to choose from
 */
const weightedChoice = (weights: number[], values: any[]) => {
  if (weights.length === 0 || values.length === 0 || weights.length !== values.length) {
    throw new Error('The length of the weights and values must be the same and greater than zero')
  }

  const cumulativeWeights = weights.reduce<number[]>((acc, curr) => {
    if (acc.length === 0) {
      return [curr]
    } 

    const next = acc[acc.length - 1]

    if (!next) {
      return acc
    }

    return [...acc, next]
  }, [])

  const randomNumber = Math.random() * (cumulativeWeights[cumulativeWeights.length - 1] ?? 1)

  const choice = _.transform(cumulativeWeights, (result, value, index) => {
    if (value >= randomNumber) {
      result.chosenIndex = index
    }

    return result.chosenIndex === -1
  }, { chosenIndex: -1 })

  return values[choice.chosenIndex]
}

export const WEIGHTED = (weights: number[]): WeightedAlgorithm => {
  return {
    pick: (values: any[]): string => {
      return weightedChoice(weights, values)
    },
  }
}

export const WEIGHTED_EVEN = (values: any[]): WeightedAlgorithm => {
  return WEIGHTED(_.fill(Array(values.length), 1))
}
```

PR #23758, opened 2022-09-09
perf: store tests & corresponding logs once

emilyrohrbough:
Apart of #23391
Apart of #21135

While profile Cypress to determine the memory impact we have, I noticed we are collecting all of the tests & corresponding logs x2.

The intent was to use `numOfTestsInMemory` to clear out the log attributes (mostly for the snapshot attributes) to reduce the number of logs stored in memory, however currently with all test attributes being stored on the browser, we need to maintain all log attributes to correctly repopulate the reporter with logs when a navigation mid-spec causes top to be reset when a new origin is visited. 

Screen Shot 2022-09-09 at 11 37 26 AM

For now, I have removed the duplicate storage & have essentially made the `numOfTestsInMemory` configuration obsolete for the time being since it wasn't truly reducing the number of tests in memory. I see two options for handling this:

1. reduce log attributes on stored browser tests and maintain a shallow list of all tests to run. This will allow restarting correctly and we can update the reporter to show "no logs due to number of tests in memory" to describe why there are no log details to show.

2. we push tracking the tests to the browser side by hooking into the existing mocha events we only leverage in run mode. We can extend the reporter in the server to track the current run state for a mid-spec navigation.

I'm in favor of Option 2.

### PR Tasks


packages/driver/src/cypress/log.ts
```diff
@@ -568,7 +568,8 @@ class LogManager {
       return
     }
 
-    // bail if we've reset the logs due to a Cypress.abort
+    // bail - when navigating to a new origin mid-test (i.e. change in top)
```
emilyrohrbough, line 571:
```suggestion
    // bail - when navigating to a new origin mid-spec (i.e. change in top)
```

packages/driver/src/cypress/runner.ts
```diff
@@ -505,26 +502,18 @@ const normalizeAll = (suite, initialTests = {}, setTestsById, setTests, onLogsBy
...
+  const normalizedSuite = normalize(suite, tests, previouslyRanTests, onLogsById, getRunnableId, getHookId, getOnlyTestId, getOnlySuiteId, createEmptyOnlyTest)
 
-  if (setTests) {
-    let i = 0
+  let i = 0
 
-    const testsArr = _.map(tests, (test) => {
-      test.order = i += 1
+  const testsArr = _.map(tests, (test) => {
+    test.order = i += 1
```
AtofStryker, line 510: is `tests` an array in this case? I think the second argument of the callback is the index and likely gets rid of `i`.
```suggestion
  const testsArr = _.map(tests, (test, index) => {
    test.order = index + 1
```

packages/driver/src/cypress/runner.ts
```diff
@@ -1071,8 +1071,8 @@ export default {
         foundTest = findTestInSuite(suite, isNotAlreadyRunTest)
       }
 
-      // if test has retried, we getTestById will give us the last attempt
-      foundTest = foundTest && getTestById(foundTest.id)
+      // if test has retried, we getTest will give us the last attempt
```
AtofStryker, line 1074:
```suggestion
      // if test has retried, getTest will give us the last attempt
```

packages/driver/src/cypress/runner.ts
```diff
@@ -1723,28 +1733,14 @@ export default {
...
-            return cleanup(queue)
-          }
-        }
-
-        return cleanup(_testsQueue)
+      cleanupQueue (_numTestsKeptInMemory) {
+        // this wasn't helpful since we store all tests in memory twice....
+        // must offload test state to the server to correctly leverage /
+        // store the defined number of tests in memory.
+        // TODO: EMILY - finishing making this work
```
AtofStryker, line 1740: Do we have an issue open for this that we can link here and assign to you?

packages/driver/src/cypress/log.ts
```diff
@@ -568,7 +568,8 @@ class LogManager {
       return
     }
 
-    // bail if we've reset the logs due to a Cypress.abort
+    // bail - when navigating to a new origin mid-spec (i.e. change in top)
```
BlueWinds, line 571:
```suggestion
    // bail when navigating to a new origin mid-spec (i.e. change in top)
```

packages/driver/src/cypress/runner.ts
```diff
@@ -736,7 +724,7 @@ const normalize = (runnable, tests, initialTests, onLogsById, getRunnableId, get
       suite.suites = []
 
       normalizedSuite.suites = _.compact(_.map(suiteSuites, (childSuite) => {
-        const normalizedChildSuite = normalize(childSuite, tests, initialTests, onLogsById, getRunnableId, getHookId, getOnlyTestId, getOnlySuiteId, createEmptyOnlyTest)
+        const normalizedChildSuite = normalize(childSuite, tests, previouslyRanTests, onLogsById, getRunnableId, getHookId, getOnlyTestId, getOnlySuiteId, createEmptyOnlyTest)
```
BlueWinds, line 727: Thanks for renaming this - previouslyRanTests is much clearer about what's going on and why tests would be in there.

packages/driver/src/cypress/runner.ts
chrisbreiding: This is definitely helpful and important for reducing memory usage.

I think there might be some understandable confusion since this config property is named `numTestKeptInMemory`. A more accurate name would be `numTestsToKeepSnapshotsFor`. It's not really about the tests kept in memory, but the snapshots those tests hold onto. Without offloading the tests to the server (which would be a good improvement), we need to keep all tests and their metadata in memory. That in itself shouldn't take up that much memory. They're fairly small objects. What _do_ take up a lot of memory are DOM snapshots. `LogUtils.reduceMemory(attrs)` is what removes the snapshots from memory by nulling them out from the logs within the test, allowing them to be garbage collected.

This change is keeping snapshots from being removed, so this will actually make memory consumption worse.

I tested this out with the following:

```js
const { _ } = Cypress

const getSnapshotsLength = () => {
  const tests = Cypress.runner.getTests()
  const snapshots = _.flatMap(tests, 'commands[0].snapshots')
  .filter((snapshot) => snapshot != null) // filter out snapshots that have been nulled out

  return snapshots.length
}

_.times(100).forEach((i) => {
  const num = i + 1

  it(`test #${num}`, () => {
    cy.log(`log #${num}`).then(() => {
      console.log(`# of snapshots after ${num} tests:`, getSnapshotsLength())
    })
  })
})
```

I temporarily added the `Cypress.runner.getTests()` method:

Screen Shot 2022-09-12 at 11 26 07 AM

On `develop`, the number of snapshots stops at 50 (it logs 51 because the test code runs before the snapshots gets cleared by the driver, but it does end up being 50).

Screen Shot 2022-09-12 at 11 21 42 AM

On this branch, the number of snapshots keeps increasing with every test and every command.

Screen Shot 2022-09-12 at 11 27 08 AM

`LogUtils.reduceMemory(attrs)` needs to remain until we implement offloading tests to the server or it's going to make memory usage worse in the interim.

chrisbreiding (changes requested):
Do you have measurements of how much this reduces memory usage? 

Since the 2 arrays store the same test references, the only extra overhead is the extra array allocation. The tests themselves shouldn't be duplicated in memory. 

It might still make sense to not be 2 arrays. I'm not sure why there are 2 to begin with, but unfortunately I don't think it's going to reduce memory consumption much to use 1 array instead of 2.

packages/driver/src/cypress/runner.ts
emilyrohrbough: @chrisbreiding interesting. Thanks for the insightful repo steps. This wasn't completely obvious, I pushed up some changes to maintain this behavior. Curious on your thoughts.

packages/driver/src/cypress/runner.ts
```diff
@@ -1124,21 +1122,6 @@ export default {
     specWindow.addEventListener('error', onSpecError('error'))
     specWindow.addEventListener('unhandledrejection', onSpecError('unhandledrejection'))
 
-    // hold onto the _runnables for faster lookup later
-    let _test: any = null
-    let _tests: any[] = []
-    let _testsById: Record<string, any> = {}
```
chrisbreiding, line 1130: Getting rid of the `_testsById` index will hurt performance when there are a lot of tests since looking up a test will be O(n) instead of O(1). Unless this greatly reduces memory usage, I don't think removing it is worth it.

chrisbreiding (changes requested):

tusharbmc:
Could someone please attend this PR, aw snap error occurrence is increasing.

PR #23760, opened 2022-09-09
fix: Webpack 5 automatic publicPath error

### LINKED ISSUE #18435: webpack v5 error when module contains dynamic import: "Automatic publicPath is not supported in this browser"
### Current behavior

When using webpack v5 and a test contains a dynamic import (or imports a module that contains a dynamic import) the test will fail with the following error: `Automatic publicPath is not supported in this browser`.

```
  Running:  example.test.js                                                                 (1 of 1)


  1) An uncaught error was detected outside of a test

  0 passing (214ms)
  1 failing

  1) An uncaught error was detected outside of a test:
     Error: The following error originated from your test code, not from Cypress.

  > Automatic publicPath is not supported in this browser

When Cypress detects uncaught errors originating from your test code it will automatically fail the current test.

Cypress could not associate this error to any specific test.

We dynamically generated a new test to display this failure.
      at eval (http://localhost:61013/__cypress/tests?p=cypress/integration/example.test.js:155:34)
      at eval (http://localhost:61013/__cypress/tests?p=cypress/integration/example.test.js:158:13)
      at eval (http://localhost:61013/__cypress/tests?p=cypress/integration/example.test.js:261:12)
      at eval (<anonymous>)
```

### Desired behavior

No error.

### Test code to reproduce

Full reduced test case: https://github.com/OliverJAsh/cypress-webpack-5-public-path-error

This might seem like a contrived example because I have reduced it down from a much larger use case.

`cypress/integration/lazy.js`:
```js
export const a = 1;

```
`cypress/integration/example.test.js`:
```js
() => import('./lazy');

describe("foo", () => {
    it("works", () => {});
});

```

`cypress/plugins/index.js`:
```js
const webpackPreprocessor = require("@cypress/webpack-preprocessor");

    on(
        "file:preprocessor",
        webpackPreprocessor({
            webpackOptions: {},
        })
    );
};

```
`cypress.json`:
```json
{}
```
`package.json`:
```json
{
  "dependencies": {
    "@cypress/webpack-preprocessor": "^5.9.1",
    "cypress": "^8.5.0",
    "webpack": "^5.58.1"
  }
}

```

### Cypress Version

8.5.0

### Other

I was able to workaround the error by setting `publicPath: ''`  inside of here:

https://github.com/cypress-io/cypress/blob/2c6cd9e09923c853ad96df3efc3496e27a7c1b3e/npm/webpack-preprocessor/index.ts#L204

I'm not really sure I understand why this issue happens nor how this fixes the issue, but FWIW this fix was recommended here: https://stackoverflow.com/questions/64294706/webpack5-automatic-publicpath-is-not-supported-in-this-browser?rq=1.

The issue does not reproduce when using webpack v4.

caseyjhol:
- Closes #18435

### User facing changelog
Fixed an issue with Webpack 5 and modules that contain dynamic imports. Fixes #18435.

### Additional details
This PR implements the fix suggested in the relevant issue. The default publicPath value in Webpack < 5 is `''`. In Webpack 5, it's `'auto'`. So, this is a backwards compatible change that will have no effect in prior versions of Webpack.

### PR Tasks


emilyrohrbough:
@ZachJW34 I know you are familiar with this package (at least more-so than me) and can better evaluate the impact of adding this default value. Mind taking a look? Wondering if this should really be in the batteries-included package?

ZachJW34:
@emilyrohrbough not too familiar but the change makes sense to me. With the change being in `@cypress/webpack-preprocessor`, the `@cypress/webpack-preprocessor-batteries-included` package will inherit the changes made to it and we want the change in both.

Do you know if we document the usage of our webpack preprocessors with Webpack 5? I feel like a system-test would make sense here, even if it's not officially supported since it seems like a lot of users are doing this.

lmiller1990:
We do document webpack 5

> This version is only compatible with webpack 4.x+ and Babel 7.x+.

There have been a few PRs fixing the preprocessor for webpack 5 over the last little while. I agree a system test makes sense, lucky there's a minimal reproduction here we could use: https://github.com/cypress-io/cypress/issues/18435

ZachJW34:
@caseyjhol have you tried this change out for a project that contains a dynamic import? I wrote a system test [here](https://github.com/cypress-io/cypress/compare/zachw/wp-webpack-5-test) and had to make some additional changes to disable webpack chunking, otherwise the webpack will create an additional chunk for the dynamic import and then fail to fetch the chunk.

caseyjhol:
@ZachJW34 Ah, good catch. This change resolved an issue I was having when an import contained a function that called a dynamic import, but that function wasn't actually being called in my spec. I tested calling the function from my spec and ran into the error you described.

I think disabling Webpack chunking makes sense for now, but it might be good to get to the root of the problem - _why_ isn't code splitting working? I could imagine a scenario where someone writes a test to explicitly check that code splitting is working as expected. However, I'm sure in most cases having code splitting disabled for tests would be more desirable.

ZachJW34:
@caseyjhol I don't think there is a valid use case for testing that something was actually code-split as that is an implementation detail of the bundler being used. The bundling of the spec and the bundling of the application under test are separate, so I don't see any value in testing that your specs are being code-split.

The `webpack-preprocessor` compiles a spec when a request for the spec is made and the assumption is that all the code for that spec is bundled in a single file that we can execute once fetched. In most scenarios, most people shouldn't be dynamically importing code in their spec files as there is no benefit over a top-level `import/require`. That said, it should still work (especially when it comes to importing a library that might be using a dynamic import, the syntax should be valid but we don't actually care about the code-splitting). To actually support code-splitting, we would have to have an endpoint to proxy the request to where we disted the result of the compilation, or run a dev-server locally that we could proxy to. Not sure there is any value in this so I'm good with limiting the number of chunks to 1.

As for the PR, do you want me to submit a PR to your fork (given that I have permissions)? Or you could copy/paste the code from my branch and push up a commit. Up to you!

caseyjhol:
@ZachJW34 Agreed - that makes sense! ~~That would be great if you could update my PR if you have the chance. Otherwise I'll try to get to it ASAP.~~ Merged in your changes.

ZachJW34:
~~@caseyjhol can you give me permissions to push to your fork?~~

Edit: Looks like you got it! Cool will review soon

ZachJW34:
@caseyjhol looks like the `webpack-preprocessor` tests are failing (due to my recommended changes 😅). Can you add:
```js
const LimitChunkCountPluginStub = sinon.stub()

webpack.optimize = {
  LimitChunkCountPlugin: LimitChunkCountPluginStub,
}
```

to the `npm/webpack-preprocessor/test/unit/index.spec.js` before the mock is registered? This would be right after the [const webpack = sinon.stub()](https://github.com/caseyjhol/cypress/blob/077d1ef43d393e3caa564d2378724d020bbb20ad/npm/webpack-preprocessor/test/unit/index.spec.js#L12) on line 12. This should fix up the test.

system-tests/projects/webpack-preprocessor-webpack-5/cypress.config.js
```diff
@@ -0,0 +1,50 @@
+// Since we are linking the `@cypress/webpack-preprocessor` package,
+// `require('webpack')` will resolve the version found in our repo (v4).
+// We need to override the webpack require to point to the local node_modules
+const Module = require('module')
+
+const originalModuleLoad = Module._load
+
+Module._load = function (request, parent, isMain) {
+  if (request === 'webpack' || request.startsWith('webpack/')) {
+    const resolvePath = require.resolve(request, {
```
lmiller1990, line 10: Do we need to try/catch this?

lmiller1990 (commented):
One q, this is looking pretty good

system-tests/projects/webpack-preprocessor-webpack-5/cypress.config.js
ZachJW34: Wouldn't hurt adding it for safety, where it falls back to the original load if it throws

ZachJW34 (approved):

lmiller1990 (approved):

system-tests/projects/webpack-preprocessor-webpack-5/cypress.config.js
lmiller1990: Ok, I'll leave it to youif you want to add it... dropped a ✅

lmiller1990:
Nice job @caseyjhol this will be in the next release :tada:

OliverJAsh:
I'm not sure this fixed https://github.com/cypress-io/cypress/issues/18435. I can still reproduce this issue. See https://github.com/cypress-io/cypress/issues/24034.

PR #23762, opened 2022-09-09
feat: implement new login and "connect project" logic

### LINKED ISSUE #23379: feat(launchpad): show "connect project" after login if project is not connected

### LINKED ISSUE #23766: [ACI post-project connection] Connect LoginConnectModals to new logic
Drive the LoginConnectModal flow fully with the logic created in #23765.

**References**
Whimsical: https://whimsical.com/get-to-first-run-TgixRWciqE1chPGDCHk5b2
Figma: https://www.figma.com/file/V0lCq4uBplf9J3GZRiD02m/ACI-Fast-Follows?node-id=19661%3A172759
Clickup: https://app.clickup.com/t/18033298/PM-2556

### LINKED ISSUE #23767: [ACI post-project connection] Connect smart banners to new logic
Drive the SpecsListBanners behavior fully with the logic created in #23765.

**Acceptance Criteria**

1. Log In, Create Organization, and Connect Project variants should all open the Connect Dashboard Modal Flow, which should change state accordingly
2. Record Run variant should have a minimum 1 day delay from when any other Smart Prompt Banner variant is shown
3. Record Run variant should have a minimum 1 day delay from when the Connect Dashboard Modal Flow is shown
4. Record Run variant should show when logged_in && organizations > 0 && project_connected && runs === 0 && specs_written_not_matching_examples > 0 
5. All Smart Prompt Banners should have a minimum 1 day delay from when the Auto-Open Docs Menu is shown (confirm, is likely already implemented)

**References**
Whimsical: https://whimsical.com/get-to-first-run-TgixRWciqE1chPGDCHk5b2
Figma: https://www.figma.com/file/V0lCq4uBplf9J3GZRiD02m/ACI-Fast-Follows?node-id=19661%3A172759
Clickup: https://app.clickup.com/t/18033298/PM-2556

marktnoonan:
- Closes #23379
- Closes #23766 
- Closes #23767

Co-authored with @warrensplayer 

There are lots of changes in this PR, though many of the are minor changes that come from using the new Login Connect store, moving files around, and simplifying some component APIs. 

This PR:
- displays the "connect project" button after login in the Launchpad
- displays a "record run" modal prompt if the user is ready to record their first run, when the "continue" button is pressed after login (if a testing type is chosen and the project is connected)
- renders the login-connect modals flow as a single component at the root of each app
- wires up the Smart Banners with their new rules via the login-connect-store and prompt manger
- wires up the new Login Connect modal itself to be driven by the new store
- add as a CloudViewerAndProject component to consolidate much of the GQL-querying needed to support the new Login Connect store
- fixes some flake, adds some missing e2e coverage for login from launchpad
- adds a couple of missing/incorrect UTM params

Allowing users to connect projects from the Launchpad login flow required moving many components and queries into frontend-shared, which had various implications for tests.

### Developer facing improvements

Through incremental changes introduced since Cypress 10.0.0, the implementation for ways to trigger modals and track state related to logging in and connecting projects has become unwieldy, with similar ideas implemented in multiple places, in slightly different ways. The complexity of the current implementation exceeds the complexity of the underlying ideas, and this PR attempts to bring them closer together by using the logic merged in #24101 to drive the UI.

- Previously we had 4 LoginModal components and 5 CloudConnectModals being rendered throughout the app, with each parent component managing opening and closed states & user progress. Now all those parent components trigger the single LoginConnectModal component for that app, which opens in the correct state based on the store and looks after itself once open.
- This cleans up lots of events and prop drilling how in the Login, CloudConnect, and Auth components work together
- Got rid of a lot of `modelValue` bindings where we were not leveraging `v-model` behavior.
- Driving more things from the store makes component tests a little easier to manage as it's all clientside.
- The `setPreferences` action in data-context will now do a lodash `merge` instead of a simple loop resetting properties that were passed to it, this avoids some reading/merging of deep preferences that we were having to do on the client now that we have nested values in there like `banners`, which was causing a problem in this PR as we'd step on freshly saved data with some other new sibling data.

#### Further improvements

We could always go further with refactoring, but had to draw the line somewhere to merge this. Things this PR doesn't address:

- quite a bit of repetition in GQL queries, I think we can use fragments more to avoid this?
- GQL queries requesting things (like `firstOrganization` in this PR) that aren't used directly by the component but need to stay up to date, and if left out can cause problems - this feels like a bit of a bad pattern, it's easy to remove something that's not obviously being used, only to find out that it can throw off some other component.
- There are more things we could hook up to the new LoginConnect store,
- Thought about pulling out the `setUpStatus` function from `isAllowedFeature.cy.ts`, if all of our tests use this, it will simplify them and they won't need to know details of what values in the store lead to each state.

#### Features 
- Users can now connect a project to the Dashboard when logging in from the Launchpad, if a testing type has been selected.
- In both the App and the Launchpad, users whose projects are connect to the dashboard, and who are ready to record their first, run can see the appropriate prompt after login

#### Minor Bugfixes
- Within the Cloud Connect modal, user can no longer click the button to create or connect a project unless a project name has been chosen or entered.
- Within the Cloud Connect modal, background polling of cloud data will no longer cause a chosen project to become unselected
- The Create Organization and Select Project modals will now only render external links when the URL is present

#### Launchpad - Login/Connect/Record flow

[Loom video of expectations for these steps in the launchpad](https://www.loom.com/share/776253e15ae9434b9c2114a819da5b65) 

- Global Mode 
  - [ ] Log in via the top nav
  - [ ] CTA should be "Continue" after login
  - [ ] Login modal should close after clicking continue
   
- With any project open
  - [ ] on "Choose testing type" screen, log in via the top nav
  - [ ] CTA should be "Continue" after login
  - [ ] Login modal should close after clicking continue

- With a project that has no project id 
  - [ ] on "Choose a Browser" screen, log in via the top nav
  - [ ] CTA should be "Connect Project" after login
  - [ ] Connect/Create Project modal should open after clicking "Connect Project"
  - [ ] After connecting project, config should reload, closing modal and causing a spinner

- With a project that has a project id, but no recorded runs in the dashboard
    - [ ] on "Choose a Browser" screen, log in via the top nav
    - [ ] CTA should be "Continue" after login
    - [ ] "Record Your First Run" modal should open after clicking "Continue"
    - [ ] Record command should be copyable
    - [ ] Modal should be dismissable
 
#### App - Login/Connect/Record flow

- With an **unconnected project** open (no project ID in config) ([loom video](https://www.loom.com/share/2452eb2bf279416f9915648626f7f62c))
  - [ ] setup - remove `promptsShown.loginModalRecord` from state.json for this project if it is present
  - [ ] on the specs list page, log in via the top nav
  - [ ] CTA should be "Connect Project" after login
  - [ ] Connect/Create Project modal should open after clicking "Connect Project"
  - [ ] Choose a project that has **no runs recorded** in the dashboard (you might need to make an empty project in your personal org for this)
  - [ ] After connecting project, config should reload, closing modal and causing a spinner
  - [ ] "Record Your First Run" smart banner should be displayed

- With a **connected project** that has **no** recorded runs ([Loom video](https://www.loom.com/share/38693cb6f40b47a2a103e449abfa3387)]
  - [ ] setup - remove `promptsShown.loginModalRecord` from state.json for this project if it is present
  - [ ] on the specs list page, log in via the top nav
  - [ ] CTA should be "Continue" after login
  - [ ] After clicking "Continue", the "record first run" modal should appear
  - [ ] Record command should be copyable
  - [ ] Modal should be dismissable
  - [ ] After dismissing modal, "Record Your First Run" smart banner should **not** be displayed at the top of the specs list
  - [ ] `promptsShown.loginModalRecord` should exist in state.json for this project

- With a **connected project** that **has** recorded runs 
  - [ ] on the specs list page, log in via the top nav
  - [ ] CTA should be "Continue" after login
  - [ ] After clicking "Continue", the modal should simply dismiss
 
####

### PR Tasks


packages/app/cypress/e2e/runs.cy.ts
```diff
@@ -714,7 +717,7 @@ describe('App: Runs', { viewportWidth: 1200 }, () => {
       // cy.percySnapshot() // TODO: restore when Percy CSS is fixed. See https://github.com/cypress-io/cypress/issues/23435
 
       cy.remoteGraphQLIntercept((obj) => {
-        if (obj.operationName === 'Runs_currentProject_cloudProject_cloudProjectBySlug') {
+        if (obj.operationName?.includes('cloudProject_cloudProjectBySlug')) {
```
marktnoonan, line 729: Then changes in these intercepts let them behave correctly now that more than just the `Runs` component is using them - don't need a strict match on full operation name

packages/app/src/runs/CloudConnectButton.cy.tsx
```diff
@@ -1,80 +1,37 @@
 import CloudConnectButton from './CloudConnectButton.vue'
-import { CloudConnectButtonFragmentDoc } from '../generated/graphql-test'
-import { CloudUserStubs } from '@packages/graphql/test/stubCloudTypes'
+import { useLoginConnectStore } from '@packages/frontend-shared/src/store/login-connect-store'
```
marktnoonan, line 2: This button no longer depends on gql, we test it via values in the store instead.

packages/app/src/runs/CloudConnectButton.vue
```diff
@@ -1,79 +1,29 @@
...
-    :prefix-icon="isLoggedIn ? ChainIcon : UserIcon"
+    :prefix-icon="user.isLoggedIn ? ChainIcon : UserIcon"
     prefix-icon-class="icon-dark-white icon-light-transparent"
-    @click="openConnection"
+    @click="openLoginConnectModal({ utmMedium: props.utmMedium })"
   >
-    {{ isLoggedIn ? t('runs.connect.buttonProject') : t('runs.connect.buttonUser') }}
+    {{ user.isLoggedIn ? t('runs.connect.buttonProject') : t('runs.connect.buttonUser') }}
   </Button>
-  <LoginModal
```
marktnoonan, line 10: No longer need these here, this button will open the LoginConnectModal at the root.

packages/app/cypress/e2e/specs_list_latest_runs.cy.ts
```diff
@@ -160,7 +160,7 @@ describe('App/Cloud Integration - Latest runs and Average duration', { viewportW
   })
 
   context('when no runs are recorded', () => {
-    beforeEach(() => {
+    it('shows placeholders for all visible specs', { defaultCommandTimeout: 6000 }, () => {
```
marktnoonan, line 163: This test was flaky for me, some minor changes in this file to address that.

graphql-codegen.yml
```diff
@@ -126,5 +126,7 @@ generates:
     <<: *vueTesting
 
   './packages/frontend-shared/src/generated/graphql-test.ts':
-    documents: './packages/frontend-shared/src/gql-components/**/*.{vue,ts,tsx,js,jsx}'
+    documents: 
+      - './packages/frontend-shared/src/gql-components/**/*.{vue,ts,tsx,js,jsx}'
```
ZachJW34, line 130: We had this distinction where gql-connected components lived in `gql-components`. Now, composables are now gql-connected. Would it make sense to have a composables folder inside of `gql-components` which are gql aware?

The codegen is a gotcha, as soon as someone tries to create a pinia action that is store away they will run into this as well. Might be good to have a single folder that we can point gql-codegen at that will contain anything gql connected.

packages/app/src/App.vue
```diff
@@ -4,13 +4,31 @@
...
+    <CloudViewerAndProject />
+    <LoginConnectModals />
+  </template>
 </template>
 
+<script setup lang="ts">
+import LoginConnectModals from '@cy/gql-components/LoginConnectModals.vue'
+
+import { useRoute } from 'vue-router'
+import CloudViewerAndProject from '../../frontend-shared/src/gql-components/CloudViewerAndProject.vue'
```
ZachJW34: Standardize on `frontend-shared` import? We wither use `@packages/frontend-shared` or relative import or `@cy/`

packages/app/src/specs/banners/ConnectProjectBanner.vue
```diff
@@ -14,75 +13,43 @@
...
     }"
-    @update:model-value="value => emit('update:modelValue', value)"
   >
     <p class="mb-24px">
-      {{ cohortOption.value }}
+      <!--
+        unref allows plain object to be used in testing
+        while a ref is provided in actual usage
+      -->
+      {{ unref(cohortOption)?.value }}
```
ZachJW34: This is a bit of a code-smell, I would expect the test to be more verbose (to accomodate ref vs plain object) rather than leak into app code

packages/frontend-shared/src/gql-components/CloudViewerAndProject.vue
```diff
@@ -0,0 +1,133 @@
+<script setup lang="ts">
```
ZachJW34, line 9: Should this be a composable? This component just triggers side-effects and has no template.

ZachJW34 (commented):
A lot to review here! Tested and it all works great, going to give it another review in the morning!

packages/app/src/specs/banners/ConnectProjectBanner.vue
marktnoonan: yeah I was thinking of undoing this as well, will give it another look 👍

marktnoonan: Updated, I see where I wrapped this already-reactive value in a computed, causing the `.value.value` weirdness that `unref` is for.

packages/frontend-shared/src/gql-components/CloudViewerAndProject.vue
marktnoonan: We have a pattern of only doing queries in .vue files, so that's the main reason I stuck with that. I don't mind a component that doesn't render anything, if you (or somebody else) have a strong preference to make this a composable, I'd be fine with that.

packages/app/src/App.vue
marktnoonan: Yeah I can standardize the ones in this PR at least, we are pretty inconsistent.

graphql-codegen.yml
marktnoonan: Agreed, for now I've moved the new composable into `gql-components/composables`.

packages/app/src/specs/SpecsListBanners.cy.tsx
```diff
@@ -73,58 +68,97 @@ describe('<SpecsListBanners />', () => {
...
+        bannerTrueProjectConditions[bannerTestId].forEach((status: keyof LoginConnectState['project']) => {
+          loginConnectStore.setProjectFlag(status, true)
+        })
+
         cy.get(`[data-cy="${bannerTestId}"]`).should('be.visible')
       })
 
       it('should be preempted by spec not found banner', () => {
-        mountWithState(gql, stateWithFirstOpenedDaysAgo(4), { isSpecNotFound: true })
+        mountWithState(gql, undefined, { isSpecNotFound: true })
```
marktnoonan: Sorta don't love passing this `undefined`, but also don't want to touch more code in this PR to refactor this function.

packages/app/src/specs/banners/CreateOrganizationBanner.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <TrackedBanner
+    v-if="cohortOption"
```
marktnoonan, line 3: Without this check it's possible we render the banner and then pieces of text content pop in once the cohort is determined.

packages/app/src/specs/banners/TrackedBanner.vue
```diff
@@ -60,40 +58,35 @@ mutation TrackedBanner_recordBannerSeen($campaign: String!, $messageId: String!,
...
 
 function updateBannerState (field: 'lastShown' | 'dismissed') {
-  const savedState = stateQuery.data.value?.currentProject?.savedState ?? {}
+  const savedBannerState = stateQuery.data.value?.currentProject?.savedState?.banners ?? {}
 
-  set(savedState, ['banners', props.bannerId, field], Date.now())
+  set(savedBannerState, [props.bannerId, field], Date.now())
 
-  setStateMutation.executeMutation({ value: JSON.stringify(savedState) })
+  setStateMutation.executeMutation({ value: JSON.stringify({ banners: savedBannerState }) })
```
marktnoonan, line 89: Scoping this to just banners avoids resetting stale `promptsShown` state.

packages/data-context/src/actions/LocalSettingsActions.ts
```diff
@@ -17,10 +18,8 @@ export class LocalSettingsActions {
     const toJson = JSON.parse(stringifiedJson) as AllowedState
 
     // update local data
-    for (const [key, value] of Object.entries(toJson)) {
-      this.ctx.coreData.localSettings.preferences[key as keyof AllowedState] = value as any
-    }
 
+    _.merge(this.ctx.coreData.localSettings.preferences, toJson)
```
marktnoonan, line 22: With nested objects in state now, fully replacing each key doesn't always make sense, so we merge instead to preserve any existing nested values not explicitly overridden by the client.

packages/frontend-shared/src/gql-components/modals/CreateCloudOrgModal.vue
```diff
@@ -122,7 +123,7 @@ onBeforeUnmount(() => {
   window.clearTimeout(timer)
 })
 
-const createOrgUrl = computed(() => props.gql.createCloudOrganizationUrl || '#')
```
marktnoonan, line 125: We never want a dead link, better to hide the link if it is somehow not provided in gql.

packages/frontend-shared/src/gql-components/modals/LoginModal.cy.tsx
```diff
@@ -85,48 +99,6 @@ describe('<LoginModal />', { viewportWidth: 1000, viewportHeight: 750 }, () => {
       cy.contains(text.login.bodySuccess.replace('{0}', cloudViewer.fullName)).should('be.visible')
       cy.contains('a', cloudViewer.fullName).should('have.attr', 'href', 'https://on.cypress.io/dashboard/profile')
     })
-
-    it('shows "connect project" after login if required by prop, and emits expected events', () => {
```
marktnoonan, line 89: moved to LoginConnectModal spec

packages/app/cypress/e2e/runs.cy.ts
```diff
@@ -116,6 +116,10 @@ describe('App: Runs', { viewportWidth: 1200 }, () => {
           obj.result.data.cloudViewer.organizations.nodes = []
         }
 
+        if (obj.result.data?.cloudViewer?.firstOrganization?.nodes) {
+          obj.result.data.cloudViewer.firstOrganization.nodes = []
```
warrensplayer, line 120: I know that the need for `firstOrganization` was introduced in previous work, but I hope we can take some more time soon to find out why we have to rely on this trick while we also have the `organizations` being returned at the same time.

packages/app/cypress/e2e/runs.cy.ts
```diff
@@ -181,7 +191,7 @@ describe('App: Runs', { viewportWidth: 1200 }, () => {
       cy.startAppServer('component')
 
       cy.remoteGraphQLIntercept(async (obj, testState) => {
-        if (obj.operationName === 'CloudConnectModals_RefreshCloudViewer_refreshCloudViewer_cloudViewer') {
+        if (obj.operationName === 'LoginConnectModals_LoginConnectModalsQuery_cloudViewer') {
```
warrensplayer, line 194: This change was just updating existing code, but looking at the rest of the code base, this spec file is the only place that there is conditional logic based on `operationName`.  This is a fragile pattern to have look at this string.  This is definitely a code smell to me but I have not looked at what the alternative would be yet.

packages/app/src/specs/SpecsListBanners.vue
```diff
@@ -249,32 +222,33 @@ watch(
...
-    // Not having this check can cause a brief flicker of the 'Create Org' banner while org data is loading
-    const isOrganizationLoaded = !!cloudViewer?.firstOrganization
-    const isMemberOfOrganization = (cloudViewer?.firstOrganization?.nodes?.length ?? 0) > 0
-    const isProjectConnected = !!currentProject?.projectId && currentProject.cloudProject?.__typename === 'CloudProject'
-    const hasNoRecordedRuns = currentProject?.cloudProject?.__typename === 'CloudProject' && (currentProject.cloudProject?.runs?.nodes?.length ?? 0) === 0
-    const hasFourDaysOfCypressUse = (Date.now() - currentProject?.savedState?.firstOpened) > interval('4 days')
+  return {
+    component: componentsByStatus[loginConnectStore.userStatus] ?? null,
+    cohortOption: getCohortForBanner(bannerIds[loginConnectStore.userStatus]),
+  }
```
warrensplayer:
```suggestion
  return reactive({
    component: componentsByStatus[loginConnectStore.userStatus] ?? null,
    cohortOption: getCohortForBanner(bannerIds[loginConnectStore.userStatus]),
  })
```
Wrapping the return in `reactive` should allow for the un-nesting of the `refs` but still allow them to be reactive.  That way, all the nested `.value` calls can be removed.  I have not fully tested this yet, but I believe this should work.

packages/frontend-shared/src/gql-components/CloudViewerAndProject.vue
ZachJW34: I don't really see a difference between queries and mutations, and mutations are allowed outside of components. Maybe the pattern I'm more familiar with is:

```html
<CloudViewerAndProject>
  <App />
</CloudViewerAndProject>
```

The example above, it's more apparent that `CloudViewerAndProject` is having an effect on the App whereas our use case it seems like you could remove it (though that would cause everything to break).

I also stumbled upon [setup-stores](https://pinia.vuejs.org/core-concepts/#setup-stores) which could be interesting.

Not going to block on this, but a detailed comment about what this component is doing would be useful since figuring out why these queries are being executed is indirect.

packages/app/src/specs/SpecsListBanners.vue
```diff
@@ -294,6 +268,10 @@ const bannerCohortOptions = {
 const cohortBuilder = useCohorts()
 
 const getCohortForBanner = (bannerId: string) => {
+  if (!bannerCohortOptions[bannerIds[loginConnectStore.userStatus]]) {
```
warrensplayer:
```suggestion
  if (!bannerCohortOptions[bannerId]) {
```

warrensplayer (commented):
@marktnoonan Still going through items, but wanted to post this since you were still working in this area. I hope my suggestions with using `reactive` in `SpecsListBanners` simplifies things.

packages/frontend-shared/src/gql-components/modals/SelectCloudProjectModal.vue
```diff
@@ -254,7 +255,26 @@ mutation SelectCloudProjectModal_SetProjectId( $projectId: String! ) {
           runs(first: 10) {
             nodes {
               id
-              ...RunCard
```
ZachJW34, line 257: Why was this change made?

packages/frontend-shared/src/gql-components/LoginConnectModalsContent.vue
```diff
@@ -0,0 +1,89 @@
...
+  } else {
+    keepLoginOpen.value = false
+  }
+}
+
+// in tests, the gql prop can sometimes be `null` for a split second, causing
+// "Detached from DOM" problems, so we are debouncing gql before it gets passed
+// into the login/connect/record modals. This seems like a good idea anyway,
+// it will also prevent flashes of very temporary states
+const gqlRef = ref<LoginConnectModalsContentFragment | null>(null)
```
ZachJW34, line 81: Can this component just be gated to not allow gql to be null? 

```ts
<LoginConnectModalsContent
    v-if="loginConnectStore.isLoginConnectOpen && query.data.value"
    :gql="query.data.value"
  />
```

packages/frontend-shared/src/gql-components/modals/SelectCloudProjectModal.vue
```diff
@@ -317,12 +337,20 @@ const projectOptions = computed(() => {
 const newProject = ref(projectOptions.value.length === 0)
 const pickedProject = ref<typeof projectOptions.value[number]>()
 
-watchEffect(() => {
-  if (projectOptions.value.length === 1) {
+watch(projectOptions, (newVal, oldVal) => {
+  // avoid unselecting currently chosen project
+  // can happen when gql updates due to polling
+  if (isEqual(newVal, oldVal)) {
```
ZachJW34, line 342: Rather than check for equality of `projectOptions`, can you check if `pickedProject.value` is defined?

packages/frontend-shared/src/gql-components/LoginConnectModals.vue
```diff
@@ -0,0 +1,57 @@
...
+  if (newVal) {
+    executeQuery()
+  }
+}, {
+  immediate: true,
+})
+
+const executeQuery = async () => {
+  await query.executeQuery()
+}
```
ZachJW34:
```suggestion
const executeQuery = async () => {
  await query.executeQuery()
}

whenever(() => loginConnectStore.isLoginConnectOpen, executeQuery)
```

Short and sweet? Didn't test this though

packages/data-context/src/codegen/code-generator.ts
```diff
@@ -159,3 +162,22 @@ async function fileExists (absolute: string) {
     return false
   }
 }
+
+export async function hasNonExampleSpec (testTemplateDir: string, specs: string[]): Promise<boolean> {
```
ZachJW34, line 166: This code feels like it doesn't belong here and would make more sense in `ProjectActions`

ZachJW34 (commented):
Some questions and nits, everything works great! Will approve once the questions are responded too, nothing was blocking.

packages/app/src/specs/SpecsListBanners.vue
marktnoonan: nice, I will check that out, I had drafted a comment about this but your review beat me to it 🙏

packages/app/cypress/e2e/runs.cy.ts
marktnoonan: These mutation names do get into a lot of places, and since their naming convention depends on our component architecture, we end up expressing that architecture in our tests, which yeah does feel a bit off, but also not a _huge_ burden in practice imo, since we rarely change the architecture.

packages/frontend-shared/src/gql-components/LoginConnectModalsContent.vue
marktnoonan: That's how it started, the problem is that a brief `null` causes an unmount/mount which leads to the detached-from-dom failures in testing. This happens _after_ the query has been resolved initially, so it's not just going from off to on, it goes off-on-off-on. I suspect this is something that could be addresses in the gql testing setup, I never saw it happening in the actual application, but didn't see any obvious thing I could change on the gql side.

packages/data-context/src/codegen/code-generator.ts
marktnoonan: I think @warrensplayer had a reason for keeping it here & calling it from `ProjectActions`, maybe he can refresh my memory?

packages/frontend-shared/src/gql-components/CloudViewerAndProject.vue
marktnoonan: Nice, I will add a comment

packages/frontend-shared/src/gql-components/modals/SelectCloudProjectModal.vue
marktnoonan: The `RunCard` fragment is defined in packages/app, I got an "unknown fragment" error from the gql codegen after moving this component into frontend-shared. Maybe there's a better way to define this fragment to make it reusable in both places.

packages/data-context/src/codegen/code-generator.ts
warrensplayer: @ZachJW34 @marktnoonan I felt a little weird about putting it there, but kept it in this file because I was using internal methods (`allFilesInDir` and `fileExists`) that I did not want to refactor at the time I created that method.

packages/app/src/specs/SpecsListBanners.cy.tsx
warrensplayer: Can the `undefined` be an empty object instead?  `{}`

packages/frontend-shared/src/gql-components/modals/SelectCloudProjectModal.vue
marktnoonan: In some cases we need to replace or unset the value of `pickedProject` - if you have value there and change organizations in the dropdown above, we use this watcher to clear out the now stale choice. 

Unless both orgs have the exact same project list, in which case this watcher does nothing (but the stale works in both lists, so it's ok).

packages/data-context/src/actions/LocalSettingsActions.ts
warrensplayer, line 22: I see that `LocalSettingsActions` does not have any unit tests. Can you see if it would be easy to add a unit test for the method this is in?

packages/frontend-shared/src/gql-components/LoginConnectModals.vue
marktnoonan: I like this, updated.

warrensplayer: Nice!  I did not know about `whenever`.  I need to spend more time looking through the `vueUse` library!

packages/app/src/specs/SpecsListBanners.vue
marktnoonan: ha, yes 🤦

packages/frontend-shared/src/gql-components/LoginConnectModalsContent.vue
ZachJW34: Oh I see, would be good to have a comment saying this only happens in testing and a TODO to figure out why.

packages/frontend-shared/src/gql-components/modals/SelectCloudProjectModal.vue
ZachJW34: Gotcha, was it Launchpad specific? I could imagine it working when building app but not launchpad. Regardless, change makes sense.

warrensplayer (commented):
Overall, the code is great.  The part I love most is all the deletion of code!  That was a great update to SpecsListBanners to consolidate the creation of the cohort with the determination of the banner to show.  

Only one strange thing I am seeing during testing and not sure if it is an existing issue or not.  Logging in with the menu item in the header seems fine.  However, using the tooltip or the LoginBanner, sometimes causes things to get "stuck".  I have attached a screen recording to show one example. In this example, one item has the "loading" skeleton stuck showing.  Also, when logging back out, the LoginBanner does not reappear.  Not sure what is happening here yet.

https://user-images.githubusercontent.com/1702361/197272915-e073e786-8110-4c7e-86ac-287d7cecbe43.mov

packages/frontend-shared/src/gql-components/LoginConnectModalsContent.vue
marktnoonan: That's what I was shooting for in this existing comment - I'll at the TODO and see if I can make the comment more clear.

packages/data-context/src/codegen/code-generator.ts
ZachJW34: I'm fine with where it is then, could do with a comment saying a refactor/move would be good.

packages/app/src/specs/SpecsListBanners.vue
```diff
@@ -230,10 +231,7 @@ const bannerToShow = computed(() => {
     needsRecordedRun: RecordBanner,
   }
 
-  return {
-    component: componentsByStatus[loginConnectStore.userStatus] ?? null,
-    cohortOption: getCohortForBanner(bannerIds[loginConnectStore.userStatus]),
-  }
+  return componentsByStatus[loginConnectStore.userStatus] ?? null
```
warrensplayer, line 234: Any particular reason you split the `cohortOption` out of this method again? I liked that the `bannerToShow` encapsulated both the banner and the props it needed, but I do not think it is worth refactoring again. It works this way as well.

warrensplayer (approved):
Looking great to me!  Great job @marktnoonan

packages/app/src/specs/SpecsListBanners.vue
marktnoonan: I liked it too, but it turned out Vue was logging warnings about wrapping the components in a `reactive` object. I don't recall the exact warning, but splitting them up seemed the best way to get what we wanted, at least for now.

packages/data-context/src/actions/LocalSettingsActions.ts
marktnoonan: Just commenting for the record that these have been added!

marktnoonan:
@ZachJW34 I think I responded to your comments, let me know if there's anything else that sticks out.

ZachJW34 (approved):
LGTM!

PR #23771, opened 2022-09-09
fix(svelte): default mount log to true

ZachJW34:
### User facing changelog
Logs the Svelte mount command by default

### Additional details
The mount log should default to true, but we missed the fallback. Found when pairing with @rockindahizzy (thanks Rocky!)

### Steps to test
Throw a `.only` on [svelte component testing](https://github.com/cypress-io/cypress/blob/b902297935b13ad11e763cef6bc02ab4c649e7c4/system-tests/test/component_testing_spec.ts#L136) and run `yarn workspace @tooling/system-tests test component_testing_spec.ts --no-exit --headed`, or test locally against a Svelte app. Check the reporter for a `mount <MyComponent ...>` log.

Before:
Screen Shot 2022-09-09 at 4 06 10 PM

After:
Screen Shot 2022-09-09 at 4 05 10 PM

### PR Tasks


npm/svelte/src/mount.ts
```diff
@@ -76,7 +76,7 @@ export function mount<T extends SvelteComponent> (
     // by waiting, we are delaying test execution for the next tick of event loop
     // and letting hooks and component lifecycle methods to execute mount
     return cy.wait(0, { log: false }).then(() => {
-      if (options.log) {
+      if (options.log !== false) {
```
lmiller1990, line 79: Nice, all our adapters do this - this could be part of GA, the standard API that adapters must comply with. Part of that could be logging the component name, and we could enforce it via an interface or something.

lmiller1990 (approved):

system-tests/project-fixtures/svelte/src/mount.cy.js
```diff
@@ -53,4 +53,34 @@ describe("Svelte mount", () => {
...
+
+      cy.wrap(Cypress.$(window.top.document.body)).within(() =>
+        cy
+          .contains("displays component name in mount log")
+          .closest(".collapsible")
+          .click()
+          .within(() =>
+            cy
+              .get(".command-name-mount")
+              .should("contain", "mount<Counter ... />")
```
rockindahizzy, line 69: Maybe it would be helpful for future us to add a comment detailing the fact there is no space here purposefully, but I'm not too concerned with it.

rockindahizzy (approved):

PR #23776, opened 2022-09-12
fix: force forward slash to backslash on Windows in spec search input

### LINKED ISSUE #23426: Updated specs list search in 10.6 is not working with forward slash
### Current behavior

In 10.6 specs list search was updated to work file path. While it is working now if path contains back slash, it does NOT work with forward slashes in the path. For example if I copy file path from IntelliJ (on Windows) it copies it with forward slashes, so I can't just copy-paste file path but need to modify it with forward slashes

### Desired behavior

Specs list search works with both forward or back slashes

### Test code to reproduce

Just try search for files with path containing forward slash
image
image

### Cypress Version

10.6.0

### Node version

16.7.0

### Operating System

Windows 10

### Debug Logs

_No response_

### Other

_No response_

astone123:
- Closes #23426

### User facing changelog

When users on Windows use the spec list search, if their search term contains any instances of `/`, they will be replaced with `\` to match the Windows path. 

### Additional details

Users on Windows couldn't copy/paste paths from anywhere that used `/` for directories into the specs search because Windows uses `\` for paths. This forces the text input to use `\` on Windows so that search terms will match directories on Windows.

### Steps to test

**Windows**
1. Open a project that has at least one spec in a directory that is not in the project root
2. Search for the spec by its directory (for example, if your spec is `src\specs\MySpec.cy.tsx`, search for `src/specs/`)
3. Verify that the directory names are highlighted in the specs list and your spec appears
4. Repeat these steps for the inline spec list (the spec list that shows in the side bar when a test is running

**Unix**
1. Follow the steps for Windows and verify that nothing has changed with the specs search

Windows users will see their `/` change to `\` when they use the spec search

### PR Tasks


mike-plummer (commented):
Works well for me, both main and inline specs search work as described on Linux & Windows. There's a tiny gap where a windows filepath pasted into Cypress running on unix won't resolve, but that's outside the scope of the original defect. I do have two nitpicks:

1. I would like to see a test for this. We have a few OS-specific tests floating around using a pattern like this:
```
if (os.platform() === 'win32') {
  ...
}
```
2. In Windows when typing into the search field, if I enter a single forward slash `/` it is changed to a backslash (expected) and the cursor jumps to the end of the line (unexpected). Dynamically altering a model value often causes this sort of behavior, so I wonder if it would be better to move the path conversion logic into the search algo rather than trying to mutate it within the UI element

packages/app/src/specs/InlineSpecListHeader.vue
```diff
@@ -110,7 +111,7 @@ const input = ref<HTMLInputElement>()
 const onInput = (e: Event) => {
   const value = (e.target as HTMLInputElement).value
 
-  emit('update:specFilterModel', value)
+  emit('update:specFilterModel', getPathForPlatform(value))
```
ZachJW34, line 114: We could move the `getPathForPlatform` into the `useSpecFilter` function, that way we could consolidate the platform handling in one spot.

ZachJW34 (commented):
Let's get a test for this new functionality

packages/app/cypress/e2e/specs_list_e2e.cy.ts
```diff
@@ -237,6 +237,14 @@ describe('App: Spec List (E2E)', () => {
         cy.get('button').contains('23 Matches')
       })
 
+      it('normalizes directory path separators for Windows', function () {
```
astone123, line 240: This will fail on `develop` for Windows right now because `e2e/accounts` won't yield any results on Windows. With my changes it passes

mike-plummer (approved):

ZachJW34 (approved):
Perfect!

PR #23780, opened 2022-09-12
fix: Update name of debug logger

mike-plummer:
- Closes 

### User facing changelog
N/A

### Additional details
Fixing name of debug logger, no functional impact

### PR Tasks

### DIFF
--- a/packages/data-context/src/actions/EventCollectorActions.ts
+++ b/packages/data-context/src/actions/EventCollectorActions.ts
@@ -1,7 +1,7 @@
 import type { DataContext } from '..'
 import Debug from 'debug'
 
-const debug = Debug('cypress:data-context:sources:EventCollectorDataSource')
+const debug = Debug('cypress:data-context:sources:EventCollector')
 
 interface CollectableEvent {
   campaign: string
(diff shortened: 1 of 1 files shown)


warrensplayer (approved):
Makes sense to me.

packages/data-context/src/actions/EventCollectorActions.ts
```diff
@@ -1,7 +1,7 @@
 import type { DataContext } from '..'
 import Debug from 'debug'
 
-const debug = Debug('cypress:data-context:sources:EventCollectorDataSource')
+const debug = Debug('cypress:data-context:EventCollector')
```
flotwig, line 4: According to https://github.com/cypress-io/cypress/blob/develop/guides/debug-logs.md#choosing-a-namespace this would be

```suggestion
const debug = Debug('cypress:data-context:actions:EventCollectorActions')
```

which is verbose, but more predictable...

mike-plummer: Thanks @flotwig !

flotwig (approved):

marktnoonan (approved):

PR #23789, opened 2022-09-12
fix: typo in viewport dropdown

### LINKED ISSUE #23632: Typo in viewport menu
### Current behavior

Dropdown menu for firefox browser dimensions has missing text or missing link.  

### Current text:

**Additionally, you can override this value 
in your or via the command.**

Screenshot from 2022-08-30 18-04-16

### Expected text:

**Additionally, you can override this value 
in your ____ or via the command ____.**

**Additionally, you can override this value 
in your `cypress.config` or via the command `npx cypress something`.**

### Test code to reproduce

- `$ yarn add cypress`
- `$ yarn run cypress open`
- Select E2E testing
- Select Firefox browser
- Run example test `spec.cy.ts`
- Open then browser viewport dimensions dropdown menu

### Cypress Version

^10.7.0

### Node version

16.15.0

### Operating System

Ubuntu 20.04.5 LTS

### Debug Logs

_No response_

### Other

- DEVICE: acer a515 laptop
- MEMORY: 5.7gib
- CPU: AMD® Ryzen 3 3200u with radeon vega mobile gfx × 4 
- GFX: AMD® Radeon(tm) vega 3 graphics
- OS NAME: Ubuntu 20.04.5 LTS
- OS TYPE: 64-bit
- GNOME: 3.36.8
- WINDOW: X11
- PKG MANAGER: yarn
- TERMINAL: vscode built-in

FireNdIce3:
- Closes #23632 

### PR Tasks


packages/frontend-shared/src/locales/en-US.json
```diff
@@ -810,7 +810,7 @@
       "buttonText": "Review the guide on viewports",
       "buttonHref": "https://on.cypress.io/viewport",
       "infoText": "The viewport determines the width and height of your application under test. By default the viewport will be {0} by {1} for {2} testing.",
-      "configText": "Additionally, you can override this value in your {2} or via the {4} command."
+      "configText": "Additionally, you can override this value in your cypress.config or via the command npx cypress something."
```
emilyrohrbough, line 813: @FireNdIce3 Thank you for opening a PR to fix this issue! Looking at where this translation is used, it looks liek it should be populating with the user's configuraiton file name or with cy.viewport(): https://github.com/cypress-io/cypress/blob/develop/packages/app/src/runner/SpecRunnerHeaderOpenMode.vue#L109

Could you see if updating to the following correctly populates this translation?

```suggestion
      "configText": "Additionally, you can override this value in your {0} or via the {1} command."
```

If it does, can we ensure this test is passing? https://github.com/cypress-io/cypress/blob/develop/packages/app/src/runner/SpecRunnerHeaderOpenMode.cy.tsx#L184

FireNdIce3: I have made the changes. I guess it should work fine, here is my commit Id 4b09dc3

chrisbreiding (changes requested):
Looks good!

Screen Shot 2022-09-16 at 10 02 02 AM

Could you add a test for this? I think adding assertions to [this test](https://github.com/cypress-io/cypress/blob/b9fec60604d359129a6036d777f98d898e117007/packages/app/cypress/e2e/cypress-in-cypress-e2e.cy.ts#L36) and [this test](https://github.com/cypress-io/cypress/blob/b9fec60604d359129a6036d777f98d898e117007/packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts#L29) should suffice.

pratiqdev:
** Ignore this comment! **
commit https://github.com/cypress-io/cypress/commit/4b09dc3b781cf5bc0baef85ab3b4c4cef1872444 solves the issue.

---

I was unable to resolve the issue locally so I have not created a PR. I took some notes while testing / exploring and I will include them here.

## Files

1. `packages/frontend-shared/src/locales/en-US.json`
2. `packages/app/src/runner/SpecRunnerHeaderOpenMode.vue` 
3. `packages/app/src/runner/SpecRunnerHeaderOpenMode.cy.tsx` 
4. `packages/app/cypress/e2e/cypress-in-cypress-e2e.cy.ts`

---

## Origin 

The issue seems to originate within the function responsible for 
parsing and replacing in-string placeholder values with real values.
This text has string-vars like: `in string var like {this} replaced`.

The original (unreplaced) text segment exists in a locale source (file-1)
at `runner.viewportTooltip.configText`

The adjacent entry of `runner.viewportTooltip.infoText` has its 
placeholder values correctly replaced at `{0}`, `{1}` and `{2}`

---

The locale object values are passed to an `<i18n-t>` component as the 
`keypath` prop in (file-2)

```ts
<i18n-t
  tag="p"
  keypath="runner.viewportTooltip.configText"
  class="mb-24px"
>
```
The test file (file-3) contains a test at line 196 that checks for the 
existence and visibility of the correctly replaced string:


```ts
cy.contains('Additionally, you can override this value in your cypress.config.js or via the cy.viewport() command.')
                                                               ^^^^^^^^^^^^^^^^^            ^^^^^^^^^^^^^
.should('be.visible')
```

There is also an example cypress test in (file-4) that tests for the 
adjacent config text `runner.viewportTooltip.infoText` but, does not 
test for `runner.viewportTooltip.configText`

chrisbreiding:
@FireNdIce3 Do you think you'll be able to add a test for this?

emilyrohrbough:
@chrisbreiding Interestingly our Component tests for this are now failing with this fixed behavior which was testing it originally https://app.circleci.com/pipelines/github/cypress-io/cypress/43400/workflows/3eb83d3f-5f15-496c-a78c-d9e62e697373/jobs/1811601

- https://github.com/cypress-io/cypress/blob/develop/packages/app/src/runner/SpecRunnerHeaderOpenMode.cy.tsx#L184
- https://github.com/cypress-io/cypress/blob/develop/packages/app/src/runner/SpecRunnerHeaderOpenMode.cy.tsx#L207

Wonder why these are inconsistent 🤔

chrisbreiding:
> Interestingly our Component tests for this are now failing with this fixed behavior which was testing it originally [app.circleci.com/pipelines/github/cypress-io/cypress/43400/workflows/3eb83d3f-5f15-496c-a78c-d9e62e697373/jobs/1811601](https://app.circleci.com/pipelines/github/cypress-io/cypress/43400/workflows/3eb83d3f-5f15-496c-a78c-d9e62e697373/jobs/1811601)
> 
>     [develop/packages/app/src/runner/SpecRunnerHeaderOpenMode.cy.tsx#L184](https://github.com/cypress-io/cypress/blob/develop/packages/app/src/runner/SpecRunnerHeaderOpenMode.cy.tsx?rgh-link-date=2022-09-20T18%3A31%3A16Z#L184)
>     [develop/packages/app/src/runner/SpecRunnerHeaderOpenMode.cy.tsx#L207](https://github.com/cypress-io/cypress/blob/develop/packages/app/src/runner/SpecRunnerHeaderOpenMode.cy.tsx?rgh-link-date=2022-09-20T18%3A31%3A16Z#L207)
> 
> Wonder why these are inconsistent 🤔

@emilyrohrbough Took a little digging, but I figured this out and figured out the root cause of the original bug. 

The [`<i18n-t>` component](https://vue-i18n.intlify.dev/guide/advanced/component.html) can do string interpolation in two ways. The way we were using it (and still using it in other places) relied on list interpolation, where `{0}`, `{1}`, etc rely on the order and placement of the children nested within the `<i18n-t>` component.

The original code looks like this (with children indexes noted):

```
    <i18n-t ...>
0:    <!-- disable rule to prevent trailing space from being added to <InlineCodeFragment/> -->
1:    <!-- eslint-disable-next-line vue/singleline-html-element-content-newline -->
2:    <InlineCodeFragment class="font-medium text-xs leading-5">{{ props.gql.configFile }}</InlineCodeFragment>
3:    <!-- eslint-disable-next-line vue/singleline-html-element-content-newline -->
4:    <InlineCodeFragment class="font-medium text-xs leading-5">cy.viewport()</InlineCodeFragment>
    </i18n-t>
```

So we used the interpolated values `{2}` and `{4}`. This works locally, in development, and in CI.

_However_, I'm guessing that when we build for production and produce the binary, the comments are stripped out before it does the i18n replacement, so the Vue markup ends up looking like this:

```
    <i18n-t ...>
0:    <InlineCodeFragment class="font-medium text-xs leading-5">{{ props.gql.configFile }}</InlineCodeFragment>
1:    <InlineCodeFragment class="font-medium text-xs leading-5">cy.viewport()</InlineCodeFragment>
    </i18n-t>
```

So then the interpolated values should be `{0}` and `{1}`. `{2}` and `{4}` don't even exist, so it ends up putting in empty strings (it would be nice if the build failed in that case, eh?). That's why the component tests used to pass but failed when using `{0}` and `{1}`. But `{0}` and `{1}` would have worked in the production app.

To resolve this, I changed the markup to use the other string interpolation method, [slots syntax](https://vue-i18n.intlify.dev/guide/advanced/component.html#slots-syntax-usage). It ensures the same values are interpolated regardless of child order or the presence/lack-of-presence of comments.

emilyrohrbough:
@chrisbreiding whoa....nice find! I bet you are correct that `prod` mode was stripping these out.

emilyrohrbough (approved):

ZachJW34 (approved):
Learned something new! I do like the template connections. LGTM!

chrisbreiding (approved):

Issue #6526, opened 2020-02-20
cy.ng fails with misleading error that does not show the real issue in the callback


bahmutov:
Cypress v4.0.2

In project https://github.com/cypress-io/cypress-example-phonecat/tree/fail the failure is misleading

To run:
- clone
- `npm install`
- `npm run dev` and open the test

The test is as follows

```js
it('fails', () => {
  cy.visit('/index.html#!/phones')
  cy.ng('model', '$ctrl.query').as('q')
  function getNames($names) {
    return $names.map(function (index, el) {
      return Cypress.$(el).text()
    }).get()
  }

  cy.get('@q').type('tablet')
  cy.ng('binding', 'phone.name')
    .should(function ($names) {
      expect(getNames($names), 'default order of names').to.deep.eq [
        'Motorola XOOM™ with Wi-Fi',
        'MOTOROLA XOOM™'
      ]
    })
})
```

Notice the real problem - missing braces around the array, should be `expect(getNames($names), 'default order of names').to.deep.eq([ ...`

But the error shows the following, implying that the `cy.ng` has failed to find the names.

<img width="1279" alt="Screen Shot 2020-02-20 at 3 23 23 PM" src="https://user-images.githubusercontent.com/2212006/74975590-49e41a00-53f5-11ea-954d-147624208699.png">

(you can add console.log or debugger statements into the `should` callback to see that the names are actually there)

chrisbreiding:
This must be the culprit: https://github.com/cypress-io/cypress/blob/e0736b1358ef95c571122dd1dd9c8601f6361b94/packages/driver/src/cy/commands/angular.coffee#L38-L38

emilyrohrbough:
We have removed support for `cy.ng()` in https://github.com/cypress-io/cypress/pull/23791. For support testing Angular in Cypress, see https://docs.cypress.io/guides/component-testing/angular/overview.

PR #23791, opened 2022-09-12
feat: move .contains() and .shadow() to be queries; remove cy.ng()

### LINKED ISSUE #7306: Re-query elements that are found 'detached' from the DOM
### Current status
Update 12/06/2022: This work as been complete and will be released in Cypress v12 today.

This is currently in progress. See https://github.com/cypress-io/cypress/issues/7306#issuecomment-1182059194, https://github.com/cypress-io/cypress/issues/7306#issuecomment-1182427284 and https://github.com/cypress-io/cypress/issues/7306#issuecomment-1188245931 for a brief description of the plan and updates on current status.

### Current behavior:

Currently when Cypress queries an element that was passed from a parent element (say in `.get()`, `.click()` or `.find()`), if the element at some point becomes detached from the DOM, we throw an error:

```
CypressError: cy.click() failed because this element is detached from the DOM.

<element>

Cypress requires elements be attached in the DOM to interact with them.

The previous command that ran was:

  cy.get()

This DOM element likely became detached somewhere between the previous and current command.

Common situations why this happens:
  - Your JS framework re-rendered asynchronously
  - Your app code reacted to an event firing and removed the element

You typically need to re-query for the element or add 'guards' which delay Cypress from running new commands.

https://on.cypress.io/element-has-detached-from-dom
```

Screen Shot 2020-11-02 at 10 27 54 AM

### Desired behavior:

Really users just want to re-query the DOM for the previously found element and click on the new element. Because it's more often than not that the framework underneath has created a completely new DOM element and inserted it into the DOM. 

The proposal is to re-query the previously found element and continue the command with any newfound elements. Perhaps this could be an option passed to the command to allow this or not. 

### Reproducible Example

Provided in https://github.com/cypress-io/cypress/issues/9032

`index.html`

```html
<form>
    <select onchange="document.forms[0].submit()">
        <option value="1">First</option>
    </select>
    <input />
</form>
```

`spec.js`

```js
it('this will fail but should not', () => {
    cy.visit('index.html')
    cy.get("select").select("First")
    // adding any wait time will make it pass
    // cy.wait(0)
    cy.get("input").type("Hallo")
})
```

### Related issues

- https://github.com/cypress-io/cypress/issues/6215
- https://github.com/cypress-io/cypress/issues/5743
- https://github.com/cypress-io/cypress/issues/4946
- https://github.com/cypress-io/cypress/issues/4414
- https://github.com/cypress-io/cypress/issues/3670
- https://github.com/cypress-io/cypress/issues/9032

BlueWinds:
- Closes https://github.com/cypress-io/cypress/issues/23804

### User facing changelog
cy.ng() was an outdated, undocumented Cypress command; it has been removed.

`.contains()` and `.shadow()` have been heavily refactored, as part of the ongoing effort to resolve #7306. `.contains()` in particular has improved performance.

### Additional details
Queries-that-use-other-queries and command-that-use-queries are now supported to a much improved extent; see the new tests in cy.cy.ts for some minimal examples, or the new version of .contains() which relies on .get() for a more practical use.

Contains() and Shadow() are now queries, using the new _addQuery API introduced in https://github.com/cypress-io/cypress/pull/23665. This means they append to the subject chain, rather than replace it.

Removing .ng() happened as part of this effort because it also relied on the old command-based version of .get() - rather than update it alongside the other commands, we decided to remove it instead. See https://cypressio.slack.com/archives/C01TQQCQLJY/p1663015490536119 for internal discussion around the subject.

As part of rewriting .contains(), I had to go into `dom/elements/find.ts`. The existing implementation involved overwriting jquery functionality, and creating and throwing away a selector function every time .contains() retried; The new version properly registers `:cy-contains()` with jquery, with no need to override the built-in `:contains()`. This should result in less memory-churn when retrying `.contains()` / `.should('contain.text')`.

### Steps to test
This PR brings us to the first example of a fixed detached DOM error:

```
it.only('test', () => {
  cy.$$('body').append('<button id="btn">foobar</button>')

  setTimeout(() => {
    cy.$$('#btn').remove()
    cy.$$('body').append('<button id="btn" class="active">foobaz</button>')
  }, 1000)

  cy.get('#btn').contains('foo').should('have.class', 'active')
})
```

In develop, this results in a Detached DOM error - .get() resolves to foobar, but then the timeout removes it and adds foobaz (with the same ID). .contains() then retries forever with foobar as its subject, eventually failing.

But in this branch, .contains() now uses the logic in `retryQuery` in command_queue.ts, which properly re-fetches the subject each attempt - no Detached DOM error!

Future PRs will expand on this, moving more commands into queries, and also updating commands so that they can receive the same benefits.

### PR Tasks


packages/driver/cypress/e2e/commands/querying/querying.cy.js
```diff
@@ -1293,6 +1291,7 @@ space
 
       it('is case sensitive when matchCase is undefined', () => {
         cy.get('#test-button').contains('Test')
+        cy.contains('test').should('not.exist')
```
BlueWinds, line 1294: Adding this assertion because I had problems show up in Launchpad component tests due to incorrect handling of case in .contains(). I thought I had the casing correct because we had tests covering it, but turns out they only asserted the positive case and not the negative.

packages/driver/src/cy/commands/index.ts
```diff
@@ -4,8 +4,6 @@ import * as Agents from './agents'
 
 import * as Aliasing from './aliasing'
 
-import * as Angular from './angular'
```
BlueWinds, line 7: `cy.ng()` is undocumented and old, and consulting with Chris and Brian, we've decided to remove it. This is technically a breaking change, but I haven't been able to find anyone in the wild using it.

I'm sure  there's someone, but... :woman_shrugging:

packages/driver/src/cypress/error_messages.ts
```diff
@@ -298,7 +298,7 @@ export default {
       docsUrl: 'https://on.cypress.io/contains',
     },
     length_option: {
-      message: `${cmd('contains')} cannot be passed a \`length\` option because it will only ever return 1 element.`,
```
BlueWinds, line 301: This error message was incorrect - you can in fact pass a length option. It just has to be 1. :woman_shrugging: Not a change in `.contains()`s behavior - it's always been this way.

packages/driver/src/cy/commands/querying/querying.ts
```diff
@@ -148,375 +150,8 @@ function getAlias (selector, log, cy) {
...
-  onRetry?: Function
-  verify?: boolean
-}
-
-interface InternalContainsOptions extends Partial<Cypress.Loggable & Cypress.Timeoutable & Cypress.CaseMatchable & Cypress.Shadow> {
-  _log?: Log
-}
-
 export default (Commands, Cypress, cy, state) => {
-  /*
```
BlueWinds, line 164: In this PR, I've resolved the temporary and confusing state of having two implementations of `.get()`. It was ugly, but now it's gone!

packages/driver/src/dom/elements/find.ts
```diff
@@ -224,55 +223,91 @@ export const getElements = ($el) => {
   return els
 }
 
-export const getContainsSelector = (text, filter = '', options: {
-  matchCase?: boolean
-} = {}) => {
-  const $expr = $.expr[':']
+/*
```
BlueWinds, line 226: I hadn't intended to do anything here in this PR / initiative, but I kept running into bugs where the old logic would break with multiple .contains() statements + aliases in the same test. This was happening because I was only calling `getContainsSelector` once, rather than each time a retry occurred, and not cleaning it up properly; the method I thought was just returning a string was also modifying the global jquery Object!

The old `getContainsSelector` had side effects, making it hard to use safely. The new one is purely functional - it takes arguments and returns a string. It also performs better, since we register these three handlers once, and allow jquery to compile / cache them, rather than adding the `:cy-contains()` expression with internal state on each retry loop.

packages/driver/cypress/e2e/commands/querying/querying.cy.js
```diff
@@ -1744,54 +1735,14 @@ space
...
+          expect(err.message).to.eq('`cy.contains()` only ever returns one element, so you cannot assert on a `length` greater than one.')
           expect(err.docsUrl).to.eq('https://on.cypress.io/contains')
 
           done()
         })
 
         cy.contains('Nested Find').should('have.length', 2)
       })
-
-      it('restores contains even when cy.get fails', (done) => {
```
BlueWinds, line 1756: The "restores contains" tests are no longer relevant, because we're not overriding jQuery's builtin `:contains` in the first place. Nothing to clean up! :)

lmiller1990:
Quick smoke test before review. It works as expected.

# Cypress 10.9

image

# issue-7306-addQuery-custom-commands

image

packages/driver/src/cy/commands/index.ts
lmiller1990: Should be fine - if someone is using it, we can just point them at Angular Component Testing.

packages/driver/src/dom/elements/find.ts
lmiller1990: > just returning a string was also modifying the global jquery Object!

:cry: 

Glad to see we are moving away from mutating things (especially using functions that you wouldn't expect to mutate something).

packages/driver/src/dom/elements/find.ts
```diff
@@ -224,55 +223,91 @@ export const getElements = ($el) => {
...
+export const getContainsSelector = (text, filter = '', options: {
+  matchCase?: boolean
+} = {}) => {
+  if (_.isRegExp(text) && options.matchCase === false && !text.flags.includes('i')) {
+    text = new RegExp(text.source, text.flags + 'i') // eslint-disable-line prefer-template
   }
 
-  // we set the `cy-contains` jquery selector which will only be used
-  // in the context of cy.contains(...) command and selector playground.
-  $expr['cy-contains'] = cyContainsSelector
```
lmiller1990, line 275: Is this the mutation you were referring to?

packages/driver/src/cy/commands/querying/root.ts
```diff
@@ -4,7 +4,7 @@ export default (Commands, Cypress, cy, state) => {
       timeout: options.timeout,
     })
 
-    cy.state('current').set('timeout', options.timeout)
+    this.set('timeout', options.timeout)
```
lmiller1990, line 7: Just to clarify, are there equivalent? My understanding is `cy.state('current')` is just the current Cypress instance. `this.set` points to the same thing - an instance of the class defined in `driver/src/cypress.ts`?

lmiller1990 (approved):
Few questions, but looks good - this was an easy one to review. Refreshing to see a PR with more deletions than additions... I feel like the a lot of technical debt is starting to get addressed.

packages/driver/src/dom/elements/find.ts
```diff
@@ -224,55 +223,91 @@ export const getElements = ($el) => {
...
-  } else if (_.isString(text)) {
-    cyContainsSelector = function (elem) {
-      let testText = normalizeWhitespaces(elem)
+    return testText.includes(text)
+  }
+})
+
+function isSubmit (elem: Element): elem is HTMLInputElement {
+  return elem.tagName === 'INPUT' && (elem as HTMLInputElement).type === 'submit'
+}
```
lmiller1990, line 276: Why do we have special logic for `isSubmit`?

Also won't any button in a form trigger a submit? eg

```html
<form>
  <button>submit</button>
</form>
```

In this case the button would behave like a submit. 

It is only used once:

```js
// taken from jquery's normal contains method
  return function (elem) {
    if (isSubmit(elem)) {
      return regex.test(elem.value)
    }
```

It's not really clear why we do this here, do you happen to know why? Not sure you needed to dive into that logic while you were here, but I'm curious either way :thinking:

packages/driver/cypress/e2e/commands/querying/querying.cy.js
```diff
@@ -776,8 +776,6 @@ describe('src/cy/commands/querying', () => {
         cy.get('#missing-el').should('have.prop', 'foo')
       })
 
-      it('throws when using an alias that does not exist')
```
emilyrohrbough, line 779: Should there be a test case added for this instead of deleted this placeholder test?

packages/driver/cypress/e2e/commands/querying/querying.cy.js
```diff
@@ -919,6 +917,12 @@ describe('src/cy/commands/querying', () => {
   })
 
   context('#contains', () => {
+    it('should keep multiple contains() separate', () => {
```
emilyrohrbough: What does this test mean?

packages/driver/src/cypress/cy.ts
```diff
@@ -896,6 +899,10 @@ export class $Cy extends EventEmitter2 implements ITimeouts, IStability, IAssert
   }
 
   now (name, ...args) {
+    if (this.queryFns[name]) {
+      return this.queryFns[name].apply(this.state('current'), args)
```
emilyrohrbough, line 901: Shouldn't the first argument be `this` so the context is shared for query commands? And the second subject would be `this.state('current')`?

packages/driver/src/cy/commands/querying/querying.ts
```diff
@@ -591,216 +229,172 @@ export default (Commands, Cypress, cy, state) => {
...
 
-  Commands.add('shadow', { prevSubject: 'element' }, (subject, options) => {
-    const userOptions = options || {}
-
-    options = _.defaults({}, userOptions, { log: true })
-
-    const consoleProps: Record<string, any> = {
-      'Applied To': $dom.getElements(subject),
-    }
+  Commands._addQuery('shadow', function contains (userOptions: ShadowOptions = {}) {
```
emilyrohrbough, line 366: The way `_addQuery` is currently implemented, this changes .shadow from being a child command which requires an element subject to a dual command which an optional previous subject. This is new behavior that I don't anticipate is intentinal.

emilyrohrbough (changes requested):
I have only gotten partially through the review, but look like the changes to `.shadow()` is unintentionally introducing new behaviors to the command.

packages/driver/src/cy/commands/index.ts
chrisbreiding: Should we keep a stubbed version around that throws an error explaining it's been removed and they should use component testing?

packages/driver/src/cy/commands/querying/querying.ts
chrisbreiding: Seems like this is handled below by `cy.ensureSubjectByType(subject, 'element')` instead of being part of the `_addQuery` functionality. Is that right?

packages/driver/src/cy/commands/querying/root.ts
chrisbreiding: `cy.state('current')` is the currently running command, not the Cypress instance. Inside queries, looks like `this` is applied as [a command created to wrap the query function](https://github.com/cypress-io/cypress/blob/1cf3c0bffd001a23eaa272220134a892a9c5a814/packages/driver/src/cypress/cy.ts#L854-L866).

packages/driver/src/cy/commands/querying/querying.ts
```diff
@@ -591,216 +229,172 @@ export default (Commands, Cypress, cy, state) => {
...
-        options._log = Cypress.log({
-          message: _.compact([filter, text]),
-          type: subject ? 'child' : 'parent',
-          timeout: options.timeout,
-          consoleProps: () => {
-            return consoleProps
-          },
-        })
+      if (!subject || (!$dom.isElement(subject) && !$elements.isShadowRoot(subject[0]))) {
+        subject = cy.$$('body')
```
chrisbreiding: I think this overwriting of `subject` is what's causing [this test failure](https://app.circleci.com/pipelines/github/cypress-io/cypress/44204/workflows/166b09d4-ef92-4dea-8fc3-ba106454609c/jobs/1851712) ([this assertion](https://github.com/cypress-io/cypress/blob/1cf3c0bffd001a23eaa272220134a892a9c5a814/packages/driver/cypress/e2e/e2e/origin/commands/querying.cy.ts#L149) in particular).

I guess I could go either way on whether the test should be updated or the `Applied To` should be changed back to undefined if there's no subject.

chrisbreiding (changes requested):
This looks really good to me. Just some test failures that need to be addressed.

packages/driver/src/cy/commands/index.ts
BlueWinds: Personally I don't think so. `.ng()` doesn't show up in our docs anywhere, not even a passing mention in the changelog. I don't think it's worth keeping any form of backwards compatibility (even an error message) beyond mentioning this change as breaking in Cy 11.

packages/driver/src/dom/elements/find.ts
BlueWinds: Yep. In develop this is overridden each time `.contains()` retries, with cyContainsSelector holding internal state (rather than letting jquery parse the args and pass it in).

packages/driver/src/cy/commands/querying/root.ts
BlueWinds: Yep. This is a DX decision I made around queries - they have direct access to their own `$Command` instance (via `this`), and properties on the command instance are where you store any information that needs to 'make it out' of the query closure (such as timeout).

packages/driver/src/dom/elements/find.ts
BlueWinds: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit

You can have `<input type="submit" value="Text displayed to user" />` - and the `value` property shows up on the screen. So `.contains()` wants to look at it as if it were text content on the page.

packages/driver/src/cy/commands/querying/querying.ts
BlueWinds: Aye. _addQuery does not have builtin functionality to assert on how queries can be used - each query that wants to has to explicitly run its own `ensure()`s to validate its position and incoming subject.

The reasons for this are a bit multifaceted, but it boils down to "parent / child / dual don't map cleanly to how Cypress implements them." You can execute parent commands after child commands, as long as the parent command returns null. And for commands we already have code to *derive* parent/child/dual from the `prevSubject` anyway. `prevSubject` is the "natural" concept, and parent/child/dual is a computed property.

So for Queries, I skipped that whole set of logic. The goal is to make it clearer when and how this code executes, to someone reading or writing queries. Each query states, explicitly with an `ensures` statement (rather than via an argument to _addQuery), exactly what subjects are allowed.

emilyrohrbough: @BlueWinds Then why does this default to `type: 'dual'` and `previousSubject: 'optional'`? https://github.com/cypress-io/cypress/blob/issue-7306-addQuery-custom-commands/packages/driver/src/cypress/cy.ts#L85

emilyrohrbough: Shouldn't it default to `parent` and `none`  and the query commands that change this behavior override this?

packages/driver/src/cy/commands/querying/querying.ts
BlueWinds: Thanks for pinpointing a failure!

I'm leaning towards updating the assertion - I think it makes sense to highlight the `body` tag we queried against. Especially looking ahead a bit towards iframe support, when a command's context could be just part of the page, "this is the section of the DOM we looked at" could possibly be useful.

packages/driver/src/cy/commands/querying/querying.ts
BlueWinds: It defaults to dual/optional because I want to bypass the existing logic entirely. This goes pretty deep into the weeds of how `prevSubject` is implemented.

`prevSubject` is checked *before* invoking the command / query. But for queries, we instead want to throw an error *inside the retry loop* - the subject might (is expected to!) change between retries, and we want to run the validation each time.

So we use 'optional' to signal that Cypress shouldn't try running any assertions before invoking the command - any incoming subject is acceptable. The query will validate the incoming subject later, in the inner queryFn.

emilyrohrbough: Can you link through where this is? I'd like to understand where this bypasses..I have been pretty deep in this code recently, and my understanding of addCommand, is that we are actually [wrapping the commands callback ](https://github.com/cypress-io/cypress/blob/issue-7306-addQuery-custom-commands/packages/driver/src/cypress/cy.ts#L744)(which we aren't doing with _addQuery()) which enforces the previous subject check for us.

packages/driver/cypress/e2e/commands/querying/querying.cy.js
BlueWinds: Good thought. Was writing an assertion to cover this when it occurred to me to check on the error being thrown - dug through and found that this case is already covered here: https://github.com/cypress-io/cypress/blob/issue-7306-addQuery-custom-commands/packages/driver/cypress/e2e/commands/aliasing.cy.js#L485

It's a bit confusing that alias tests are split between `aliasing.cy.js` and `querying.cy.js`, but I don't think I want to move them as part of this PR. Another bit of cleanup to be done later.

packages/driver/cypress/e2e/commands/querying/querying.cy.js
BlueWinds: I've changed the name of this test to `does not interfere with other aliased .contains()` and moved it further down in the file, but not sure how to succinctly capture what it's guarding against in the test title.

Added a longer explanation as a code comment; let me know if that's still not clear and I'll see what I can do.

packages/driver/src/cy/commands/querying/querying.ts
BlueWinds: Ok, went through the code and looks like my explanation is out of date - what I said is why I *originally* had optional / dual, but it no longer applies. Thank you for pressing on the matter!

`optional` and `dual` aren't needed at all, so I've removed them. Queries just plain old don't interact with the concept of `prevSubject` / `type` anymore - they follow their own codepath which bypasses them entirely, and my above explanation was badly out of date. Just had to tweak .contains() slightly to make sure it properly showed up as a child log message.

packages/driver/cypress/e2e/commands/querying/querying.cy.js
```diff
@@ -1550,19 +1550,13 @@ space
         })
       })
 
-      it('sets type to parent when subject isnt element', () => {
```
BlueWinds, line 1554: This is a behavioral change: .contains() now logs as a child whenever it's used as a child command. Previously, it logged as a child if used as a child command *unless the previous subject was body/window*.

This exception is not a useful one. Simple, easy to understand rules are preferable. If the previous command returns a subject (not null), .contains() now shows as a child of that command.

chrisbreiding:
I guess this PR can close [this issue](https://github.com/cypress-io/cypress/issues/6526) since `cy.ng()` won't be around anymore.

BlueWinds:
Ok, took longer than I'd hoped but have all tests passing (except one which I commented out with a TODO to re-enable as part of a follow-up - since this is into a feature branch, I'd like to put that off for now and start work related to `.within()` in a fresh PR).

chrisbreiding (changes requested):
Looks like there are a lot of unrelated commits and changes in the diff. Might be caused by merging `develop` directly into this branch instead of merging it into the `issue-7306` branch and then merging that branch into this one.

As it is, it's difficult to review the changes.

BlueWinds:
@chrisbreiding - Merged develop into issue-7306, then rebased this PR off that branch. I'll be more careful about that in the future; I thought that if I merged develop into both branches, it would work itself out, but apparently not!

chrisbreiding (approved):

PR #23792, opened 2022-09-13
chore: migrate runner/dom.js to app part 1

sainthkh:
- Part 1 of #23187

### User facing changelog

### Additional details
- Why was this change necessary? => To remove legacy `pkg/runner`. #23187 is too big. I'll break it into 3 PRs.
- Any implementation details to explain? => Check the comments below:
- 
### Steps to test

Run Selector Playground and check if Highlights are shown correctly. 

N/A

### PR Tasks


packages/app/src/runner/aut-iframe.ts
```diff
@@ -307,22 +311,22 @@ export class AutIframe {
       // bail if our el no longer exists in the parent body
       if (!this.$.contains(body, element)) return
 
-      // switch to using outerWidth + outerHeight
+      // switch to using offsetWidth + offsetHeight
       // because we want to highlight our element even
       // if it only has margin and zero content height / width
-      const dimensions = this.dom.getOuterSize($_el)
+      const dimensions = this._getOffsetSize($_el.get(0))
```
sainthkh, line 317: jquery `outerWidth` and `outerHeight` are equal to `offsetWidth` and `offsetHeight` when [only first argument is given to the function](https://github.com/cypress-io/cypress/blob/566a1a208cb51d193fd1838a4c35ab872a729fe7/packages/runner/src/dom/dimensions.js#L79-L84).

Compare the images.

* JQuery `outerWidth`



[doc link](https://api.jquery.com/outerwidth/)

* Vailla js `offsetWidth`



[doc link](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetWidth)

sainthkh:
I tested with my local Ubuntu 20.04. And all the failure are flaky.

packages/runner/src/selector-playground/css.js
```diff
@@ -0,0 +1,166 @@
+export const css = `
```
chrisbreiding, line 1: What's the reason this can't remain `scss`?

sainthkh: I first wrote the code in this way to understand which code can be reused. 

This code will be completely removed in the final version. I think both `scss` and `css` versions are OK for now.

chrisbreiding: I see. Maybe this PR should also remove [selector-playground.scss](https://github.com/cypress-io/cypress/blob/823ffd0ca920c82170705d9abfa60592b8f3979a/packages/runner/src/selector-playground/selector-playground.scss) and [selector-playground.scss.d.ts](https://github.com/cypress-io/cypress/blob/823ffd0ca920c82170705d9abfa60592b8f3979a/packages/runner/src/selector-playground/selector-playground.scss.d.ts) since it's replacing them?

sainthkh: I thought about removing them with the next PR (migrating React `Highlight` component to Vue `Highlight` component). I agree it makes sense to do it now.

chrisbreiding (approved):

packages/app/src/runner/aut-iframe.ts
```diff
@@ -502,4 +506,292 @@ export class AutIframe {
...
+      obj.height === previousLayer.offsetHeight
+  }
+
+  private _getDimensionsFor (dimensions, attr, dimension) {
+    return dimensions[`${dimension}With${attr}`]
+  }
+
+  private listeners: any[] = []
+
+  private _addOrUpdateSelectorPlaygroundHighlight ($el, $body, selector?, showTooltip?, onClick?) {
```
ZachJW34: Any reason you changed this function to be a list of parameters rather than a single options object? I'm a fan of the latter as it makes reading whoever calls this function a lot easier to parse.
```suggestion
  private _addOrUpdateSelectorPlaygroundHighlight ({ $el, $body, selector, showTooltip, onClick }: any) {
```

ZachJW34 (approved):
Gave it a test, works great. One small nit but not blocking.

packages/app/src/runner/aut-iframe.ts
sainthkh: I cannot remember clearly why I changed this. I think I did this because the order of arguments are not consistent and it was hard to copy and paste. 

I rolled it back.

PR #23800, opened 2022-09-13
chore: fix flake for experimentalStudio config error test

### LINKED ISSUE #23743: Fix flake test: "experimentalStudio is not a valid config for component testing"
### Current behavior

Fails sometimes on CI, sometimes linux, usually windows. I think it is a detached DOM related issue (maybe).

### Desired behavior

Does not fail.

### Test code to reproduce

This is an example: https://app.circleci.com/pipelines/github/cypress-io/cypress/43088/workflows/d2b0ef06-de0a-4040-bd99-b40c726ff83e/jobs/1796860

Just run it a few times locally - it should fail eventually, even on linux/Mac.

### Cypress Version

10.7

### Node version

16.16

### Operating System

linux, mac,windows

### Debug Logs

_No response_

### Other

_No response_

astone123:
- Closes #23743

### User facing changelog

n/a

### Additional details

This test was failing seemingly because there was a race condition between when `writeFileInProject` was finished writing the new config file and when the config was loaded for the project. This resulted in `experimentalStudio` not being set for `component`, which resulted in the error not showing in the launchpad. This caused the test to fail intermittently.

I fixed this by creating a separate configuration file in the `experimentalSingleTabRunMode` project and then passed the file (`cypress-invalid-studio-experiment.config.js`) in as the `--config-file` option to `cy.openProject` during the test setup. This seemed to work more reliably. My guess is that's because we're not waiting on a file to be written to the project; it's already there. 

We should probably do some more investigation into why this race condition is happening. There are a bunch of other tests that I see using this pattern, so I wonder why this isn't an issue for those tests, as it seems like this was only an issue for this specific test. Something to think about.

### Steps to test

Run `config-warning.cy.ts` in `launchpad` on your machine multiple times (throw a `.only` on line 114 to speed things up) and make sure that the test doesn't fail. I ran this on my machine about 15-20 times and didn't see a failure with my change, whereas before it would fail about every 3-4 runs. It's not great to rely on this type of testing but I'm not sure of a better way to check if it's still flaky.

n/a

### PR Tasks

### DIFF
--- a/packages/launchpad/cypress/e2e/config-warning.cy.ts
+++ b/packages/launchpad/cypress/e2e/config-warning.cy.ts
@@ -111,29 +111,9 @@ describe('experimentalSingleTabRunMode', () => {
 })
 
 describe('experimentalStudio', () => {
-  // TODO: fix this flaky test. https://github.com/cypress-io/cypress/issues/23743
-  it.skip('is not a valid config for component testing', { defaultCommandTimeout: THIRTY_SECONDS }, () => {
+  it('is not a valid config for component testing', () => {
     cy.scaffoldProject('experimentalSingleTabRunMode')
-    cy.openProject('experimentalSingleTabRunMode')
-    cy.withCtx(async (ctx) => {
-      await ctx.actions.file.writeFileInProject('cypress.config.js', `
-        const { defineConfig } = require('cypress')
-
-        module.exports = defineConfig({
-          component: {
-            experimentalStudio: true,
-            devServer () {
-              // This test doesn't need to actually run any component tests
-              // so we create a fake dev server to make it run faster and
-              // avoid flake on CI.
-              return {
-                port: 1234,
-                close: () => {},
-              }
-            },
-          },
-        })`)
-    })
+    cy.openProject('experimentalSingleTabRunMode', ['--config-file', 'cypress-invalid-studio-experiment.config.js'])
 
     cy.visitLaunchpad()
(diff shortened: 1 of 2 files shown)


marktnoonan:
@astone123 I seem to hit this failure sometimes, is this the same as the flake?

Screen Shot 2022-09-13 at 12 02 55 PM

I'm doing 

```js
Cypress._.times(100, () => {
it.only('is not a valid config for component testing', () => {
...
	}
}
```

And it failed 17 times - gonna test with a longer `defaultCommandTimeout`.

marktnoonan:
Increasing timeout didn't seem to help - when it hangs, it hangs. I also tried a 2s wait at the start of each repetition in case this was file system related and would benefit from cool down. Doing `_.times(100)` for other tests in that spec has the same problem, a certain % hang on the spinner never get to the testing type picker. 

This feels like a completely separate race condition in starting up/launching projects.

astone123:
@marktnoonan yeah that looks different than the flake we were seeing before. Probably because I removed the custom timeout option. We might need to put that back

astone123:
@marktnoonan oof - yeah, that's not great. seems like an issue related to these types of tests in general 😕

ZachJW34:
FWIW I ran the test 100+ times and didn't see failure for the unskipped test

marktnoonan (approved):

ZachJW34 (approved):

PR #23811, opened 2022-09-13
chore: use a file for cache versioning and not env

flotwig:
### User facing changelog

n/a

### Additional details

- This changes how we change the cache version. Previously, it was set in CircleCI env. This requires users to have access to edit Circle env, which we are trying to restrict. Now it's set in cache-version.txt which doesn't require escalated privs to change.
- Another benefit is that changing cache-version to try and chase a bug on one branch won't cause everyone else's branches to suddenly take 4x as long to build/in-progress pipelines on other branches to fail

### PR Tasks

	- The circle-env stuff used CACHE_VERSION as a sanity check that our env isn't included in a job, so I switched it to use another env var that still exists (COPY_CIRCLE_ARTIFACTS)


MuazOthman (approved):

rachelruderman (commented):
Any special steps to test?

rachelruderman (commented):
Also, lame duck question... any chance these changes are related to the Percy failures in CI?

flotwig:
> Any special steps to test?

ahh you got me. Check the cache keys we are using, for example in `node_modules_install`, you'll see it starting with `v{checksum}...`

> Also, lame duck question... any chance these changes are related to the Percy failures in CI?

Seems unlikely, I have not touched anything for `develop` yet.

rachelruderman:
> Also, lame duck question... any chance these changes are related to the Percy failures in CI?

I should provide more context to this question! Basically, if I recall correctly, Percy uses the `CIRCLE_WORKFLOW_ID` environment variable for finalizing builds. If the same `CIRCLE_WORKFLOW_ID` gets reused, it'll be all like "ThIs BuiLd YoU'rE tRyInG To RuN WaS AlReaDy FiNaLizEd". I realize I'm completely butchering this explanation and my memory on this stuff is super fuzzy. Extremely fuzzy. The fuzziest. @marktnoonan plz correct the heck out of this explanation 🥲 🙏 

Can you try pushing new commit(s) to this branch within the cached timeframe and seeing if the [Percy CI failure](https://app.circleci.com/pipelines/github/cypress-io/cypress/43421/workflows/dc1ba710-2230-4360-8ca1-efde4932b3bf/jobs/1812678) is consistent? If so, then I think we have an issue on our hands. If not, I bet it was a one-off and I'll be glad to throw an approval your way

Screen Shot 2022-09-14 at 1 18 36 PM

rachelruderman (approved):
I merged in develop and see the Percy failure didn't reappear! I think I've just been traumatized by Percy and have it on the 🧠 , soz for the distraction

PR #23829, opened 2022-09-14
fix: Fix missing `it.skip` function in Angular tests

### LINKED ISSUE #23409: it.skip not working in Angular component test
### Current behavior

```
import { StepperComponent } from './stepper.component'

describe('StepperComponent', () => {
  it.skip('mounts', () => {
    cy.mount(StepperComponent)
    cy.contains('[data-cy=stepper]', 'works')
  })

  it('mounts using a template', () => {
    cy.mount(`<app-stepper data-cy=mounted [count]="101"></app-stepper>`, {
      declarations: [StepperComponent],
    }).then(console.log)
  })
})
```
Screen Shot 2022-08-17 at 17 10 47

Workaround: use `xit` instead of `it.skip` and it skips it

### Desired behavior

`it.skip` should work :)

### Test code to reproduce

Following https://docs.cypress.io/guides/component-testing/quickstart-angular in Cypress v10.6.0
package.json file

```json
{
  "name": "test-angular-component",
  "version": "0.0.0",
  "scripts": {
    "ng": "ng",
    "start": "ng serve",
    "build": "ng build",
    "watch": "ng build --watch --configuration development",
    "test": "ng test"
  },
  "private": true,
  "dependencies": {
    "@angular/animations": "^14.1.0",
    "@angular/common": "^14.1.0",
    "@angular/compiler": "^14.1.0",
    "@angular/core": "^14.1.0",
    "@angular/forms": "^14.1.0",
    "@angular/platform-browser": "^14.1.0",
    "@angular/platform-browser-dynamic": "^14.1.0",
    "@angular/router": "^14.1.0",
    "rxjs": "~7.5.0",
    "tslib": "^2.3.0",
    "zone.js": "~0.11.4"
  },
  "devDependencies": {
    "@angular-devkit/build-angular": "^14.1.3",
    "@angular/cli": "~14.1.3",
    "@angular/compiler-cli": "^14.1.0",
    "@types/jasmine": "~4.0.0",
    "cypress": "^10.6.0",
    "jasmine-core": "~4.2.0",
    "karma": "~6.4.0",
    "karma-chrome-launcher": "~3.1.0",
    "karma-coverage": "~2.2.0",
    "karma-jasmine": "~5.1.0",
    "karma-jasmine-html-reporter": "~2.0.0",
    "prettier": "^2.7.1",
    "typescript": "~4.7.2"
  }
}
```

### Cypress Version

10.6.0

### Node version

16

### Operating System

mac

### Debug Logs

_No response_

### Other

_No response_

mike-plummer:
- Closes #23409 

### User facing changelog
* Fixes an issue where `it.skip` was not available in Angular component tests

### Additional details
`zone.js/testing` library patches Mocha to be "zone-aware", but the patch does not include `it.skip`. This is written up in that repo under [this issue](https://github.com/angular/angular/issues/46297), but it's not seeing any movement. Since this is a relatively trivial fix on our end and has annoying impacts on our users it seemed better to patch on our side pending any potential fix from that library.

While debugging this issue I wrote up tests that validate `skip` and `only` for all major Mocha syntax blocks - I've added them to this PR even though `only` isn't really related since they seemed like useful checks to have in case Zone breaks these in the future

### Steps to test
System Tests should validate

Manual validation would involve creating an NG project, installing Cypress and patching in updated `angular` package and validating that a test with `it.skip` does not fail with a parsing error

Users should be able to use `it.skip` in Angular tests

### PR Tasks

### DIFF
--- a/system-tests/test/component_testing_spec.ts
+++ b/system-tests/test/component_testing_spec.ts
@@ -122,6 +122,14 @@ describe(`Angular CLI major versions`, () => {
       browser: 'chrome',
       expectedExitCode: 0,
     })
+
+    systemTests.it(`v${majorVersion} Zone.JS does not break Mocha`, {
+      project: `angular-${majorVersion}`,
+      spec: 'src/app/zonejs-mocha-skip.cy.ts,src/app/zonejs-mocha-only.cy.ts',
+      testingType: 'component',
+      browser: 'chrome',
+      expectedExitCode: 0,
+    })
   }
 
   systemTests.it('angular 14 custom config', {
--- a/system-tests/test/component_testing_spec.ts
+++ b/system-tests/test/component_testing_spec.ts
@@ -113,19 +113,15 @@ describe(`Angular CLI major versions`, () => {
   systemTests.setup()
 
   for (const majorVersion of ANGULAR_MAJOR_VERSIONS) {
-    const spec = `${majorVersion === '14' ? 'src/app/components/standalone.component.cy.ts,src/app/mount.cy.ts' : 'src/app/mount.cy.ts'}`
+    let spec = 'src/**/*.cy.ts'
+
+    if (majorVersion === '13') {
+      spec = `${spec},!src/app/components/standalone.component.cy.ts`
+    }
 
     systemTests.it(`v${majorVersion} with mount tests`, {
       project: `angular-${majorVersion}`,
       spec,
-      testingType: 'component',
-      browser: 'chrome',
-      expectedExitCode: 0,
-    })
-
-    systemTests.it(`v${majorVersion} Zone.JS does not break Mocha`, {
-      project: `angular-${majorVersion}`,
-      spec: 'src/app/zonejs-mocha-skip.cy.ts,src/app/zonejs-mocha-only.cy.ts',
       testingType: 'component',
       browser: 'chrome',
       expectedExitCode: 0,
--- a/cli/package.json
+++ b/cli/package.json
@@ -161,7 +161,9 @@
       "import": "./svelte/dist/cypress-svelte.esm-bundler.js",
       "require": "./svelte/dist/cypress-svelte.cjs.js",
       "types": "./svelte/dist/index.d.ts"
-    }
+    },
+    "./angular/restore-mocha": "./angular/restore-mocha.js",
+    "./angular/save-mocha": "./angular/save-mocha.js"
   },
   "workspaces": {
     "nohoist": [
--- a/cli/package.json
+++ b/cli/package.json
@@ -161,9 +161,7 @@
       "import": "./svelte/dist/cypress-svelte.esm-bundler.js",
       "require": "./svelte/dist/cypress-svelte.cjs.js",
       "types": "./svelte/dist/index.d.ts"
-    },
-    "./angular/restore-mocha": "./angular/restore-mocha.js",
-    "./angular/save-mocha": "./angular/save-mocha.js"
+    }
   },
   "workspaces": {
     "nohoist": [
--- a/npm/angular/package.json
+++ b/npm/angular/package.json
@@ -26,9 +26,7 @@
     "zone.js": ">=0.11.0"
   },
   "files": [
-    "dist",
-    "save-mocha.js",
-    "restore-mocha.js"
+    "dist"
   ],
   "types": "dist/index.d.ts",
   "license": "MIT",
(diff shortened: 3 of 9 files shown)


system-tests/test/component_testing_spec.ts
```diff
@@ -122,6 +122,14 @@ describe(`Angular CLI major versions`, () => {
       browser: 'chrome',
       expectedExitCode: 0,
     })
+
+    systemTests.it(`v${majorVersion} Zone.JS does not break Mocha`, {
```
ZachJW34: Rather than run these scaffolding twice, I say we edit the spec for the test above to include these tests so:
```ts
for (const majorVersion of ANGULAR_MAJOR_VERSIONS) {
  let spec = 'src/**/*.cy.ts'

  if (majorVersion === '13') {
    spec = `${spec},!src/app/components/standalone.component.cy.ts`
  }

  systemTests.it(`v${majorVersion} with mount tests`, {
    project: `angular-${majorVersion}`,
    spec,
    testingType: 'component',
    browser: 'chrome',
    expectedExitCode: 0,
  })
}
```

These tests are expensive so if we can include them in another test I think it's a win.

ZachJW34 (approved):
Works great! One small comment on system tests

system-tests/test/component_testing_spec.ts
```diff
@@ -122,6 +122,14 @@ describe(`Angular CLI major versions`, () => {
...
+      project: `angular-${majorVersion}`,
+      spec: 'src/app/zonejs-mocha-skip.cy.ts,src/app/zonejs-mocha-only.cy.ts',
+      testingType: 'component',
```
lmiller1990: I actually think this is the wrong place for this test. Also, we are not actually verifying that `it.skip` actually does what's it supposed to, only that it is **available**.

We could move this to a Cypress in Cypress test in `app/cypress/e2e`, and we could then assert against the command log that the correct number of tests are executed and/or skipped? That would verify not only that it works, but `skip` is the correct one (since we will see it visually marked as skipped in the Commanad Log, as well as the number of executed tests).

lmiller1990: You would add it here: https://github.com/cypress-io/cypress/blob/develop/packages/app/cypress/e2e/cypress-in-cypress-component.cy.ts

If you haven't seen the Cypress in Cypress yet it's really cool - you can still use this project (in system tests) but you can now assert things are actually happening in the UI, too.

What do you think?

lmiller1990 (commented):
Nice job fixing it, I think we can do a better test w/ Cypress in Cypress, lmk what you think!!

system-tests/test/component_testing_spec.ts
mike-plummer: @lmiller1990 It looks like `cypress-in-cypress` is configured to use React for CT, so were you thinking we would stand up a new `cypress-in-cypress` permutation using Angular?

ZachJW34: We don't need to use cypress-in-cypress to test against the command log, we have an existing pattern where we point Cypress at the command log inside of system-tests [here](https://github.com/cypress-io/cypress/blob/2d47cefdea3241e4453cb6f709d76e1eeb486423/system-tests/project-fixtures/svelte/src/mount.cy.js#L61). I could go either way, but I'm a fan of testing as much as possible in system-tests and only using cy-in-cy if absolutely required since the system-tests will always be less costly.

mike-plummer: Took the approach Zach proposed to write some tests to validate the behavior of `skip` and `only`. Definitely not as elegant as `cypress-in-cypress`, but very performant and seems to provide the validation that was requested

lmiller1990: Oh I hadn't seen that @ZachJW34, neat.

The reason I suggested cypress-in-cypress is we have these really cool helpers: https://github.com/cypress-io/cypress/blob/6ee305ba411c646298018a965d249a418ac68671/packages/app/cypress/e2e/support/execute-spec.ts#L14-L18.

You can do

```js
cy.waitForSpecToFinish({
  passCount: 4
  failCount: 0
  pendingCount: 1
})
```

Which is exactly what we want here - it's way more readable, at least to me. Maybe we should find a way to share more of these commands between cypress-in-cypress and system-tests in the future.

Anyway, as long as we've got *something* preventing this regression, I'm happy.

lmiller1990 (approved):

PR #23833, opened 2022-09-15
fix: relativeFile is now relative to repo root

### LINKED ISSUE #22971: Incorrect spec path relayed to dashboard
### Current behavior

An issue was logged to the dashboard about an [incorrect GitHub link for specs](https://cypress-io.atlassian.net/browse/CLOUD-599), which upon further investigation, seems to suggest the test runner is at fault for the incorrect link. Speaking with @RobBickel about the issue, he shared some details of his investigation, noting the following:

> cypress-services derives it from the database. It is constructed from the codeframe. The codeframe is generated in the test runner. The test runner uses the error message to generate a stacktrace, and uses the file uri from where the occured in the error, applies some regex, and adds the relative path to the codeframe. Most of the time the regex and join work just fine, but sometimes we get an bad url which causes that to boil up to the customer clicking and getting 404 from the dash.

Looking at an example generated link that fails compared to what it SHOULD BE, it is possible that monorepo's mess with the current logic.

Bad Link:
[https://github.com/cypress-io/cypress-services/blob/5915dd88fd957431fbb1ee0642bf7e69292c62c1/cypress/integration/organizations/ProjectsList_spec.ts#L480](https://github.com/cypress-io/cypress-services/blob/5915dd88fd957431fbb1ee0642bf7e69292c62c1/cypress/integration/organizations/ProjectsList_spec.ts#L480)

Good Link:
[https://github.com/cypress-io/cypress-services/blob/5915dd88fd957431fbb1ee0642bf7e69292c62c1/frontend/packages/dashboard/cypress/integration/organizations/ProjectsList_spec.ts#L480](https://github.com/cypress-io/cypress-services/blob/5915dd88fd957431fbb1ee0642bf7e69292c62c1/frontend/packages/dashboard/cypress/integration/organizations/ProjectsList_spec.ts#L480)

For more information regarding internal conversations about this issue, see the Cloud issue referenced above.

### Desired behavior

When the dashboard displays a link for a failed spec, the URL is correct.

### Test code to reproduce

* Record a failing test to the dashboard (Note: this *may* have to be part of a monorepo)
* Within Latest Runs, find the run including your failed test and click on it
* Find and click on the failed test
* Within the overlay that displays to the right, find and click on the link to the failing spec file
* Note the 404 message

### Cypress Version

10.3.1

### Other

Here's a screenshot showing the specific link that is sometimes incorrect.
image

rachelruderman:
- Closes https://github.com/cypress-io/cypress/issues/22971

### User facing changelog
The `relativeFile` property on the error `codeFrame` object is now relative to the repository root. Previously, it was relative to the Cypress config file root. 

### Additional details
- Adds internal config property `repoRoot` that uses the `.git` directory, if it exists, to identify the repository root. 
- Updates `relativeFile` in the error codeframe object to be relative to the repoRoot, if the repoRoot is set.

Why do we need this, you ask? Well, in the dashboard, we had an edge case for monorepos where we were building the incorrect URL to link to the user's source code in GitHub, BitBucket, etc. The root of the issue (see what I did there? 😛) was that the file path provided to the dashboard was relative to the cypress config. When the cypress config was located at the repo root, everything worked! But when it wasn't, as is the case for monorepos, the URL was incorrect. 

### Steps to test
1. In the `cypress` monorepo, make a failing test in `packages/app/cypress/e2e/reporter_header.cy.ts`
1.  From `cypress/packages/app`, record a run using `CYPRESS_INTERNAL_ENV=staging yarn cypress:run:e2e --headless --key ENTER_PROJECT_KEY --spec **/e2e/reporter_header.cy.ts --record --browser chrome`
1. Go to the [Dashboard staging site](https://dashboard-staging.cypress.io/). Find the test run, open the error, and click the GitHub link in the side panel. Confirm the URL works 🎉 [Here's one I made](https://dashboard-staging.cypress.io/projects/ypt4pf/runs/2648/test-results/fc170ff4-7a01-41ad-8124-89376fd5d36c)

No change to user experience for standard repos. For monorepos, the `relativeFile` value is now relative to the monorepo root. See example:
**Before**: `cypress/e2e/runs.cy.ts`
**After**: `packages/app/cypress/e2e/runs.cy.ts`

### PR Tasks


packages/driver/cypress/e2e/cypress/stack_utils.cy.js
```diff
@@ -93,6 +126,15 @@ describe('driver/src/cypress/stack_utils', () => {
...
       expect($stackUtils.getCodeFrame(originalErr)).to.be.undefined
     })
+
+    it('returns relativePath from root when `absoluteFile` starts with `repoRoot`', () => {
+      // question - how do i mock a value in the cypress config?
+      cy.stub($sourceMapUtils, 'getSourceContents').returns(sourceCode)
+      cy.spy($stackUtils, 'getCodeFrameFromSource')
+      $stackUtils.getCodeFrame(originalErr)
+      // question - i know this is being called b/c my console log is showing up
+      expect($stackUtils.getCodeFrameFromSource).to.have.been.called
```
rachelruderman: Here I spy on `getCodeFrameFromSource`. I know it's being called during the test because if I throw a console log inside `getCodeFrameFromSource`, it shows up. However, my assertion on line 136 fails 🥲 ? por que?

packages/driver/cypress/e2e/cypress/stack_utils.cy.js
```diff
@@ -93,6 +126,15 @@ describe('driver/src/cypress/stack_utils', () => {
 
       expect($stackUtils.getCodeFrame(originalErr)).to.be.undefined
     })
+
+    it('returns relativePath from root when `absoluteFile` starts with `repoRoot`', () => {
+      // question - how do i mock a value in the cypress config?
```
rachelruderman: Here I'd like to set the value of `Cypress.config('repoRoot')`. 

I've tried... 
1. Setting it directly via `Cypress.config('repoRoot, 'dev')`. This didn't work -- I think b/c Tim and I made this a non-configurable value, similar to `projectRoot`. 
2. Stubbing it via `cy.stub(Cypress, 'config').toFakeFn(() => ({ ...Cypress.config(), repoRoot: 'dev' }))`. This ended up crashing Cypress with an infinite loop of uncaught errors from lodash 👹

packages/driver/cypress/e2e/cypress/stack_utils.cy.js
```diff
@@ -23,6 +23,39 @@ describe('driver/src/cypress/stack_utils', () => {
...
+
+    it('returns relativeFile if absoluteFile does not start with `repoRoot`', () => {
+      Cypress.config('repoRoot', 'User/ruby/test-repo')
+      const result = $stackUtils.getRelativePathFromRoot(relativeFile, absoluteFile)
+
+      expect(result).to.equal(relativeFile)
+    })
+
+    it('returns the relative path from root if the absoluteFile starts with `repoRoot`', () => {
+      Cypress.config('repoRoot', repoRoot)
```
rachelruderman, line 52: This one isn't working see comment below

packages/driver/cypress/e2e/cypress/stack_utils.cy.js
```diff
@@ -23,6 +23,39 @@ describe('driver/src/cypress/stack_utils', () => {
...
+    })
+
+    it('returns relativeFile if `repoRoot` is not set in the config', () => {
+      const result = $stackUtils.getRelativePathFromRoot(relativeFile, absoluteFile)
+
+      expect(result).to.equal(relativeFile)
+    })
+
+    it('returns relativeFile if absoluteFile does not start with `repoRoot`', () => {
+      Cypress.config('repoRoot', 'User/ruby/test-repo')
```
rachelruderman, line 54: This one isn't working see comment below

cli/types/cypress.d.ts
```diff
@@ -2993,6 +2993,7 @@ declare namespace Cypress {
     // Internal or Unlisted at server/lib/config_options
     namespace: string
     projectRoot: string
+    repoRoot: string
     devServerPublicPathRoute: string
```
rachelruderman: This PR adds `repoRoot` to the Cypress config. We detect it by assuming that the directory with the `.git` folder is the repo root. 

Previously, `relativeFile` in the error codeframe was relative to the Cypress config file. Now, it's relative to the repo root. This fixes an edge case with monorepos where we had an incomplete path for `relativeFile`, resulting in GitHub links to the user's source code being busted in the dashboard.

Question: Our assumption about `.git` being at the repo root might not always be correct -- for example, Tim mentioned that some users may not have git when running in CI, i.e. maybe the test files are hanging out in an S3 bucket. Def an edge case though. 

**Should we make `repoRoot` a configurable config option? Right now it's internal/unlisted, like `projectRoot`. If so, would we have to get sign-off from product?**

lmiller1990: Hmm can you clarify `projectRoot`? It's great to have a format definition for `repoRoot` - might be good to know what exactly the difference it between that and `projectRoot`.

I'd expect `repoRoot` to be nullable - representing a test run where a project doesn't have an Git. Should this be `string | null`? Better yet, `Readonly<string | null>`?

I wonder if it should actually be `gitRepoRoot`? We don't support other source control software (which is why we have `GitDataSource`, not `SourceControlDataSource`, for example).

Is there a scenario a user would want to configure `repoRoot`? I can't think of one.

packages/data-context/src/DataContext.ts
```diff
@@ -97,6 +97,10 @@ export class DataContext {
     this.lifecycleManager = new ProjectLifecycleManager(this)
   }
 
+  get git () {
+    return this._coreData.currentProjectGitInfo
```
lmiller1990:
```suggestion
    return this.coreData.currentProjectGitInfo
```

I am not sure what, if anything, is different between `_coreData` and `coreData`, but I think we generally want to use the non `_` version.

packages/driver/cypress/e2e/cypress/stack_utils.cy.js
lmiller1990: I don't know how to fix it just by looking at it, but I can pull it down and find out why

packages/driver/cypress/e2e/cypress/stack_utils.cy.js
lmiller1990: Could be a red herring, but I think the problem here is related to our compilation process (webpack, babel, typescript etc...)

I put a debugger. We hit this point twice; once during the original call in your test and once during the actual failed assertion. It looks like the first time it points to the correct stub, but the second time it's slightly different, you can see it's got `__WEBPACK__`...

image

image

lmiller1990: I think what's happening is you are stubbing the function, but when the code actually runs it's calling the webpack module cache version, which is possibly immutable (or just the cached version w/o the `cy.spy` wrapper).

packages/driver/cypress/e2e/cypress/stack_utils.cy.js
lmiller1990: This one is passing for me 

image

lmiller1990: Actually, it passes with `it.only`, but not in the group. Code smell... something is mutating something.

lmiller1990: If I do `context.only('getRelativePathFromRoot')` this is passes, but then if you also add `  context.only('.getCodeFrame', () => {` it fails. These two tests have some kind impact on each other.

lmiller1990: I think https://github.com/cypress-io/cypress/pull/23833/files#diff-430e30e905812f3055196475c7f8d091ec640ccf95dec5e52dae9799204674daR130 is the problematic one, you've mistakenly called out this one since they both have the same test description. Line 130 is the failing spec.

packages/driver/cypress/e2e/cypress/stack_utils.cy.js
lmiller1990: I think this is actually getting set how'd you think, I tested like this:

```js
Cypress.config('repoRoot', 'dev')
console.log(Cypress.config('repoRoot')) // logs 'dev'
```

It won't update on the back-end in data context - but I don't think we want do that, this is just for testing, yeah?

packages/driver/cypress/e2e/cypress/stack_utils.cy.js
lmiller1990: Either way, I think we don't even need this - how about doing this? https://github.com/cypress-io/cypress/compare/update-git-info...lachlan/update-git-info?expand=1

packages/driver/cypress/e2e/cypress/stack_utils.cy.js
lmiller1990: Also I think this should be `Cypress.config('repoRoot, '/dev')`, right (note the leading `/`). Your test can't pass otherwise, since `absoluteFile` always starts wtih `/dev`.

lmiller1990 (commented):
Weird problem with `cy.spy`, not sure what's going on there.

[Either way, I think we can write the test in a different fashion](https://github.com/cypress-io/cypress/compare/update-git-info...lachlan/update-git-info?expand=1). Let me know what you think!

packages/driver/cypress/e2e/cypress/stack_utils.cy.js
rachelruderman: You figured it out!!!! I was missing the leading `/`. Thank you 😻

cli/types/cypress.d.ts
rachelruderman: 1. Updated to `Readonly<string | null>` 🎉 
2. Re: why not `gitRepoRoot`: it actually started off as `gitRepoRoot` 😄 , but ultimately decided that `git` is just an implementation detail. In the edge case Tim provided where some users won't have git in CI, there are other methods we could use to detect the root, i.e. looking for the root-most `package.json`. However since it's very much an edge case I think it's not the kind of thing we should invest in unless a problem presents itself. Keeping `repoRoot` generic gives us space to expand on the implementation should the need arise. 
3. Re: scenarios where we would want to configure `repoRoot`: in case our auto-detection of repoRoot for whatever reason doesn't work for the user's use case. That's an edge case though, thinking about it now I think it's best to keep things simple -- non-configurable -- unless the need presents itself

packages/data-context/src/DataContext.ts
rachelruderman: @tgriesser oooh do you know the answer to this question?

lmiller1990:
> Run the https://github.com/cypress-io/cypress-services/pull/4806 locally using yarn setup -> yarn start. We'll need this to test since there are changes in that repo to

I don't have services running locally - any chance to deploy your patch to staging? I think this is pretty easy (maybe Alejandro can help) and it would make QAing much quicker.

For now I will just review the changes, if it's too much trouble to get staging deployed I can look into configuring services locally.

cli/types/cypress.d.ts
```diff
@@ -2993,6 +2993,7 @@ declare namespace Cypress {
     // Internal or Unlisted at server/lib/config_options
     namespace: string
     projectRoot: string
+    repoRoot: Readonly<string | null>
```
lmiller1990: Maybe this doesn't make sense after all - I suppose if we *really* want it to be readonly, we will need something at runtime too.

Since this is undocumented, maybe it's fine to not mark as `Readonly`. We have no commitment to maintaining an undocumented property.

Could go either way on this, up to you, just thinking out loud - not a blocking change request.

packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -461,6 +461,15 @@ export class ProjectConfigManager {
     )
   }
 
+  private getRepoRoot = () => {
+    /*
+      repoRoot is used to detect the full file path for the relativeFile in stack_utils
+      it's needed to show the correct link to files in repo mgmt tools like GitHub in the dashboard
+      right now we assume it's same as `.git` dir, if need be can update to look for most root-level package.json
+    */
```
lmiller1990:
```suggestion
    /*
      Used to detect the correct file path when a test fails. 
      It is derived and assigned in the packages/driver in stack_utils.
      It's needed to show the correct link to files in repo mgmt tools like GitHub in the dashboard.
      Right now we assume the repoRoot is where the `.git` dir is located.
    */
```

lmiller1990: Just took a stab, I think we want to be very specific with this comment since the relationship between this getter, `data-context` and `driver` is not very clear.

packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -461,6 +461,15 @@ export class ProjectConfigManager {
     )
   }
 
+  private getRepoRoot = () => {
```
lmiller1990:
```suggestion
  get getRepoRoot () {
```

Can be a getter, no side effects here

packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -479,6 +488,7 @@ export class ProjectConfigManager {
       cliConfig: options.config ?? {},
       projectName: path.basename(this.options.projectRoot),
       projectRoot: this.options.projectRoot,
+      repoRoot: this.getRepoRoot(),
```
lmiller1990: Looks like this method is only called once - do we even need to define a `getRepoRoot`? Should we just do:

```js
repoRoot: this.options.ctx.git?.gitBaseDir
```

Then add the comment about `repoRoot` in `GitDataSource` or something?

packages/driver/src/cypress/stack_utils.ts
```diff
@@ -142,13 +142,25 @@ const getCodeFrameFromSource = (sourceCode, { line, column: originalColumn, rela
...
 }
 
+const getRelativePathFromRoot = (relativeFile, absoluteFile) => {
+  // at this point relativeFile is relative to the cypress config
+  // we need it to be relative to the repo root, which is different for monorepos
+  const repoRoot = Cypress.config('repoRoot')
+
+  if (absoluteFile?.startsWith(`${repoRoot}/`)) {
+    return absoluteFile.replace(`${repoRoot}/`, '')
+  }
```
lmiller1990, line 162: Just as a sanity check, is this working on windows? I am not sure if stack traces are always using posix (so `/`). I am pretty sure they are, but good to check, since we aren't running these tests on windows CI until they merge. I'm thinking we either

- manually run these tests on windows
- enable windows CI

Not sure which is faster/easier, up to you.

lmiller1990 (commented):
Couple of nits, main concern is around windows, see final comment.

This looks like it was quite a bit of work, nice job diving into `driver`... it's definitely the most complex part of Cypress.

packages/data-context/src/data/ProjectConfigManager.ts
```diff
@@ -461,6 +461,15 @@ export class ProjectConfigManager {
     )
   }
 
+  private getRepoRoot = () => {
+    /*
+      repoRoot is used to detect the full file path for the relativeFile in stack_utils
+      it's needed to show the correct link to files in repo mgmt tools like GitHub in the dashboard
+      right now we assume it's same as `.git` dir, if need be can update to look for most root-level package.json
+    */
+    return this.options.ctx.git?.gitBaseDir
```
emilyrohrbough, line 471: Will this work for projects that use git submodules?

packages/data-context/src/data/ProjectConfigManager.ts
rachelruderman: Ah yes, I originally had it as a one-liner, but felt for others navigating this code in the future it might be easy to miss. I added `getRepoRoot` to make it more explicit and have the long-form comment. Personally I think it's the best spot for it but if you or others feel strongly happy to move 😸

packages/driver/src/cypress/stack_utils.ts
rachelruderman: Great callout! Just enabled Windows CI, fingers crossed 🤞

packages/data-context/src/data/ProjectConfigManager.ts
rachelruderman: Oh dear, I hadn't heard of git submodules. From what I can see, the previous implementation did not support git submodules and these changes do not either. Additionally, the top results on Google regarding git submodules are from ~2015/2016, and a 2016 blog post from GitHub says:
> Before you add a repository as a submodule, first check to see if you have a better alternative available. Git submodules work well enough for simple cases, but these days there are often better tools available for managing dependencies than what Git submodules can offer.

I think we might be able to get some hard numbers from the data team on their usage by paid users, but the old results + recommendation from GitHub to use alternatives makes me think the use of git submodules is likely not very common.

Given the existing lack of git submodule support and likeliness of this being a rare edge case, I would consider adding support a separate issue and would recommend fixing it if/when we run into it. What do you think?

packages/driver/src/cypress/stack_utils.ts
rachelruderman: Yay it's passing on Windows! 🪟

packages/data-context/src/DataContext.ts
tgriesser: Yeah I feel like what @lmiller1990 is suggesting sounds right here

tgriesser (changes requested):
Let's go with @lmiller1990's suggestion on the `coreData`, otherwise 👍

packages/data-context/src/data/ProjectConfigManager.ts
rachelruderman: I switched from `private` to `get` and then it was failing as it no longer was a property on `this` 🥲 I'm not currently familiar with using get/private in JavaScript, but it seems fine to leave as `private` since at this time it is only intended to be used in this file

packages/driver/src/cypress/stack_utils.ts
rachelruderman: Ahhhh @tbiethman pointed out that the windows executors do not run driver tests. I'll test this manually on a Windows VM later today!

packages/data-context/src/data/ProjectConfigManager.ts
rockhold: @rachelruderman, for the incorrect links where you were seeing an ```@``` included in the URL, did you ever determine the cause? Submodules are something that I had briefly considered being related to that. The [original Jira](https://cypress-io.atlassian.net/browse/CLOUD-599) logged on the Cloud side has a specific link for a client with an ```@``` in the URL that may still be valid. It may require that you impersonate someone with access to that project in the dashboard to see it, just fyi.

A [quick search](https://sourcegraph.com/search?q=context:global+file:.gitmodules&patternType=standard) finds some big projects that are using submodules, so we may not want to completely rule it out as something to address. However, I would definitely support the creation of a separate issue to track that work. Before picking it up, we could sync with a data scientist to help us understand how impactful a fix there may be and then use that info to prioritize the change.

rachelruderman: Ahhh I forgot about the `@` -- it was called out in the [original JIRA](https://cypress-io.atlassian.net/browse/CLOUD-599), but the "Good" / "Bad" examples in the [GitHub issue](https://github.com/cypress-io/cypress/issues/22971) were just for monorepos 🤦‍♀️ 

AFAIK this PR would not address the `@` issue. Does it make sense to create a separate GitHub issue for that? Or would you like me to continue with this ticket in the upcoming sprint and incorporate it?

rockhold: My thoughts:
* Let's figure out what the situation with the ```@``` is all about (is that for submodules or something else?)
* Then let's verify whether this fix resolves that situation too
* Once we understand that, we can determine if we want to split it out to another story or continue working on it from this one

Most likely, we'll want to split it out, given where we're at on this sprint. Unless the situation with the ```@``` is a pretty common one (and it impacts a lot of people), the priority of the remaining work may be lower than what you have here. And in that case, pushing these current changes out will at least resolve the issue for users with monorepos.

packages/driver/src/cypress/stack_utils.ts
rachelruderman: @tbiethman is a national treasure. This was broken on windows and has now been fixed. THANK YOU!

packages/driver/src/cypress/stack_utils.ts
```diff
@@ -142,13 +142,28 @@ const getCodeFrameFromSource = (sourceCode, { line, column: originalColumn, rela
...
     language: getLanguageFromExtension(relativeFile),
   }
 }
 
+const getRelativePathFromRoot = (relativeFile, absoluteFile) => {
+  // at this point relativeFile is relative to the cypress config
+  // we need it to be relative to the repo root, which is different for monorepos
+  const repoRoot = Cypress.config('repoRoot')
+  const posixAbsoluteFile = (Cypress.config('platform') === 'win32')
+    ? absoluteFile?.replaceAll('\\', '/')
```
rachelruderman, line 157: I wanted to use `path.posix` here per the [Writing cross-platform JavaScript](https://github.com/cypress-io/cypress/blob/develop/guides/writing-cross-platform-javascript.md) guide but it was undefined on path :(

lmiller1990 (approved):

packages/driver/src/cypress/stack_utils.ts
lmiller1990: `path` is only available in Node.js so it won't show up in front-end code :+1:  

Looks good!

rachelruderman: @lmiller1990 we do have access to path here -- `import path from 'path'` on line 3, used in a few places -- it just didn't have `posix` as a property 👀

tgriesser (approved):

packages/data-context/src/data/ProjectConfigManager.ts
rachelruderman: Re: git submodules, i created a [ticket for the data team](https://cypress.clickup.com/t/18033298/DATA-904) and they got back to us. It seems they might not have enough information on their end to detect git submodules

tbiethman (approved):

packages/driver/src/cypress/stack_utils.ts
lmiller1990: @rachelruderman Can sync up if you want to know more about this kind of thing (bundling, how it works etc). 

What I think is happening is this: webpack will automatically polyfill (basically stubs) node core modules in an attempt to let you bundle some code originally written for Node.js in the browser. 

Webpack 5 dropped this (since the idea of an auto polyfill is not good; developers should not be bundling Node.js code for the browser, doesn't really make sense). We are currently on webpack 4, though, so it auto polyfills the core modules. 

Ref: https://gist.github.com/ef4/d2cf5672a93cf241fd47c020b9b3066a

rachelruderman: Ohhhhhhhhh mega interesting! Thank you for the explanation! 💟

lmiller1990: I think webpack is going away soon!! https://github.com/cypress-io/cypress/pull/23928

PR #23844, opened 2022-09-15
fix: force Webpack to emit assets on error

### LINKED ISSUE #23219: Projects for some frameworks don't surface compilation errors/warnings
### Current behavior

After an initial clean build (no compilation errors), compilation errors are no longer surfaced in the runner. Editing a spec file to throw a compilation error causes the app to reload but with no visual indicator that the compilation failed (and the tests will even rerun). You can see the errors in the console and the terminal, but users aren't always paying attention to these.

https://user-images.githubusercontent.com/25158820/183713613-860e9594-eff7-4544-a70a-d1e29a47cd94.mov

### Desired behavior

The user should be made aware of the compilation error in the runner, otherwise they may be led to believe that their test is passing or their changes are being reflected.

### Test code to reproduce

I was only able to reproduce this with certain Webpack 5 projects, such as a Next.js application and our in progress Angular work. CreateReactApp v5 doesn't exhibit this behavior

1. `npx create-next-app my-next-app && cd my-next-app`
2. `npm i -D cypress`
3. `yarn cypress open`
4. Walk through CT setup and launch a browser
5. Create `pages/home.cy.js` with contents:
```js
import Home from "./index";

it("should render", () => {
  cy.mount(<Home />);
});
```
6. Run test in Cypress
7. Modify `pages/home.cy.js` so that it will throw a compilation error, notice the app refreshes and test runs even though spec is invalid.

### Cypress Version

10.4.0

### Other

The issue seems to be related to Webpack's caching feature, and more complicated frameworks (such as Next.js) have more sophisticated caching mechanisms. This behavior is inline with normal dev experience, the only difference is that most frameworks utilize an overlay to display the error. We axed the overlay in #21599, but it would prove useful in this scenario. There are some downsides of enabling the overlay, since it will display over the AUT for any compilation errors/warnings, even if the error isn't related to the current test.

A more robust fix would be to disable the dev-server from serving files that have compilation errors. My thoughts are that since there is a cached version of the module, it is getting served in the broken modules place.

astone123:
- Closes #23219

### User facing changelog

Compile errors will now be surfaced in the command log during testing for Angular and Next projects

### Additional details

Some frameworks (Angular and Next) configure Webpack to not emit assets that have errored. So, if there's a compile error during development in these projects, instead of serving the module that has the compilation error and throwing it, Webpack serves the successful asset from the previous build. The result is that the compilation error only appears in the console, and our command log and AUT look fine - as if nothing has happened.

This behavior is undesirable because if there is a compilation error, we want to fail the test and make the user aware. This PR adds `emitOnErrors: true` (`noEmitOnErrors: false` for Webpack 4) to our default Webpack configuration to override these values and force Webpack to emit modules even when there is an error regardless of the framework configuration.

### Steps to test

1. Revert my changes in `makeDefaultWebpackConfig.ts`
2. Run the tests for `angular.cy.ts` and `next.cy.ts` in `npm/webpack-dev-server`
3. Verify that the tests fail because errors do not appear in the command log when a compilation error is written to the component file
4. Put the changes to `makeDefaultWebpackConfig.ts` back and repeat step 2
5. Verify that the tests pass now and the errors appear in the command log as expected

### PR Tasks


npm/webpack-dev-server/src/makeDefaultWebpackConfig.ts
```diff
@@ -24,9 +24,17 @@ export function makeDefaultWebpackConfig (
...
+  // Webpack <= 4 uses `noEmitOnErrors: false`
+  const emitOnErrorsProperty = {
+    [config.sourceWebpackModulesResult.webpack.majorVersion === 5 ? 'emitOnErrors' : 'noEmitOnErrors']:
+    config.sourceWebpackModulesResult.webpack.majorVersion === 5,
+  }
+
   const finalConfig = {
     mode: 'development',
     optimization: {
+      ...emitOnErrorsProperty,
```
ZachJW34:
```suggestion
  const optimization = <Record<string, any>>{}

  if (config.sourceWebpackModulesResult.webpack.majorVersion === 5) {
    optimization.emitOnErrors = true
  } else {
    optimization.noEmitOnErrors = false
  }

  const finalConfig = {
    mode: 'development',
    optimization: {
      ...optimization,
```

I found this a bit hard to parse, being more explicit reads a bit nicer.

npm/webpack-dev-server/cypress/e2e/nuxt.cy.ts
```diff
@@ -49,6 +49,28 @@ for (const project of PROJECTS) {
...
+
+        await ctx.actions.file.writeFileInProject(
+          tutorialVuePath,
+          (await ctx.file.readFileInProject(tutorialVuePath)).replace('export', 'expart'),
+        )
+      })
+
+      // The test should fail and the stack trace should appear in the command log
+      cy.get('.failed > .num', { timeout: 10000 }).should('contain', 1)
+      cy.contains('The following error originated from your test code, not from Cypress.').should('exist')
```
ZachJW34, line 69: Is there something common in the stack trace we can assert against? I like this message, is it not present for all the apps?

npm/webpack-dev-server/__snapshots__/makeWebpackConfig.spec.ts.js
```diff
@@ -29,10 +30,14 @@ exports['makeWebpackConfig ignores userland webpack `output.publicPath` and `dev
   },
   "devServer": {
     "progress": true,
-    "overlay": false
+    "overlay": true,
+    "client": {
```
ZachJW34: Snapshot from previous runs? Were not tweaking the overlay so I wouldn't expect this snapshot to change execpt for the `emitOnErrors`

npm/webpack-dev-server/cypress/e2e/create-react-app.cy.ts
```diff
@@ -51,6 +51,25 @@ for (const project of WEBPACK_REACT) {
...
+
+      cy.withCtx(async (ctx) => {
+        await ctx.actions.file.writeFileInProject(
+          ctx.path.join('src', 'App.js'),
+          (await ctx.file.readFileInProject(ctx.path.join('src', 'App.js'))).replace('export', 'expart'),
+        )
+      })
+
+      // The test should fail and the stack trace should appear in the command log
+      cy.get('.failed > .num', { timeout: 10000 }).should('contain', 1)
```
lmiller1990: I wonder if we can consider sharing or just copy pasting this helper: 

https://github.com/cypress-io/cypress/blob/develop/packages/app/cypress/e2e/support/execute-spec.ts#L26-L42

It's way more readable. Thoughts?

lmiller1990 (commented):
Some suggestions, nice job!

npm/webpack-dev-server/src/makeDefaultWebpackConfig.ts
```diff
@@ -24,9 +24,17 @@ export function makeDefaultWebpackConfig (
 
   debug(`Using HtmlWebpackPlugin version ${version} from ${importPath}`)
 
+  // Webpack 5 uses `emitOnErrors: true`
+  // Webpack <= 4 uses `noEmitOnErrors: false`
+  const emitOnErrorsProperty = {
+    [config.sourceWebpackModulesResult.webpack.majorVersion === 5 ? 'emitOnErrors' : 'noEmitOnErrors']:
```
lmiller1990: Do we have coverage for both Webpack 4 & 5? Not immediately obvious

npm/webpack-dev-server/cypress/e2e/nuxt.cy.ts
astone123: It is - I wanted to ensure that the expected stack trace was shown in the command log, but I think we could just use this to be consistent across all of the tests. I'll update for consistency. 

The reason I used this phrase here was because I was having trouble selecting the message in the stack trace (I think it is due to there being `>` characters in there? Not sure).

npm/webpack-dev-server/cypress/e2e/create-react-app.cy.ts
astone123: Good call, I didn't see that

npm/webpack-dev-server/src/makeDefaultWebpackConfig.ts
astone123: Yeah we test multiple versions for most frameworks, some that use Webpack 4 and some that use 5

npm/webpack-dev-server/__snapshots__/makeWebpackConfig.spec.ts.js
astone123: Yeah not sure what that was about... should be worked out now 👍🏻

npm/webpack-dev-server/cypress/e2e/angular.cy.ts
```diff
@@ -22,8 +22,8 @@ for (const project of WEBPACK_REACT) {
     it('should mount a passing test', () => {
       cy.visitApp()
       cy.contains('app.component.cy.ts').click()
-      cy.waitForSpecToFinish()
-      cy.get('.passed > .num').should('contain', 1)
+      cy.waitForSpecToFinish({ passCount: 1 })
```
warrensplayer, line 25: I am getting typescript errors from this call due to the version of the command in the `webpack-dev-server` not having any arguments for `waitForSpectToFinish`.  It is defined in `npm/webpack-dev-server/cypress/support/commands.ts`. The command that does have it is in the `app` package defined here: `packages/app/cypress/e2e/support/execute-spec.ts`

npm/webpack-dev-server/cypress/e2e/angular.cy.ts
```diff
@@ -52,7 +51,26 @@ for (const project of WEBPACK_REACT) {
...
+      cy.contains('app.component.cy.ts').click()
+      cy.waitForSpecToFinish({ passCount: 1 })
+
+      // Create compilation error
+      cy.withCtx(async (ctx) => {
+        await ctx.actions.file.writeFileInProject(
+          ctx.path.join('src', 'app', 'app.component.ts'),
+          (await ctx.file.readFileInProject(ctx.path.join('src', 'app', 'app.component.ts'))).replace('class', 'classaaaaa'),
+        )
+      })
```
warrensplayer:
```suggestion
      cy.withCtx(async (ctx) => {
        const componentFilePath = ctx.path.join('src', 'app', 'app.component.ts')
        await ctx.actions.file.writeFileInProject(
          componentFilePath,
          (await ctx.file.readFileInProject(componentFilePath)).replace('class', 'classaaaaa'),
        )
      })
```
Just a nit-pick about not duplicating the creation of the file path

warrensplayer (commented):
Overall looks good.  Just need to update the version of the `watiForSpecToFinish` command

npm/webpack-dev-server/cypress/e2e/angular.cy.ts
astone123: Huh - for some reason I don't get these TS errors 🤔 maybe something is messed up with my IDE configuration. Good catch

npm/webpack-dev-server/test/makeWebpackConfig.spec.ts
```diff
@@ -64,13 +69,16 @@ describe('makeWebpackConfig', () => {
...
+          emitOnErrors: false, // This will be overridden by makeWebpackConfig.ts
+        },
       },
       devServerEvents: new EventEmitter(),
     }
     const actual = await makeWebpackConfig({
       devServerConfig,
       sourceWebpackModulesResult: createModuleMatrixResult({
-        webpack: 4,
+        webpack: 5,
```
astone123, line 81: This should be 5, not 4, since `webpack-dev-server` v4 uses Webpack 5. Confused yet?

npm/webpack-dev-server/test/makeWebpackConfig.spec.ts
```diff
@@ -24,12 +23,18 @@ describe('makeWebpackConfig', () => {
...
+        optimization: {
+          noEmitOnErrors: true, // This will be overridden by makeWebpackConfig.ts
+        },
       },
       devServerEvents: new EventEmitter(),
     }
     const actual = await makeWebpackConfig({
       devServerConfig,
-      sourceWebpackModulesResult: sourceDefaultWebpackDependencies(devServerConfig),
+      sourceWebpackModulesResult: createModuleMatrixResult({
```
astone123, line 34: Updated this to use `createModuleMatrixResult`, because we claim to be testing `webpack-dev-server` v3 in this test, but the version of Webpack was getting resolved to v5, when it should be v4.

ZachJW34 (approved):

warrensplayer (approved):
Looks good!

rockindahizzy (approved):

PR #23915, opened 2022-09-21
feat: output video file path on one row during run mode

### LINKED ISSUE #23913: Make video file path clickable
### What would you like?

When a video recording of a test finishes processing, it currently behaves as follows:

| - Finished processing: | /This/Is/A/Video/File/Path/   | (1 second)  |
|                                       | To/A/Video/Recording.mp4 |                      |

This results in the video file path not being clickable, or able to copy/paste without also including the `(1 second)` cell. Altering this table so the path isn't broken would make troubleshooting a bit faster.

### Why is this needed?

Allows videos to be opened directly from the terminal, resulting in faster troubleshooting.

### Other

_No response_

zachz-indeed:
- Closes #23913 

### User facing changelog
Moves video output link to it's own line in `run` mode to make the video output link more easily clickable in the terminal.

### Additional details
Improves the user experience to allow users to directly click the recorded link inside the terminal or copy/paste from a CI terminal more easily

### Steps to test
Easiest way to see changes locally is to run a single test in `cypress run` mode. For omitting values under `system-test`s in the core repo, please run the system tests and compare the snapshot.

#### Before
https://user-images.githubusercontent.com/3980464/213530519-059c7d55-17b9-4b3b-b50c-b192253fd3bb.mp4

#### After
https://user-images.githubusercontent.com/3980464/213529911-2e685005-9b43-44a0-ba31-15d1b9db424b.mp4

### PR Tasks


mjhenkes:
@zachz-indeed, This seems like a good change. Could you provide a before and after screenshot for us to compare?

zachz-indeed:
(Ignore font color differences)

Before: 
Screen Shot 2022-10-10 at 12 31 09 PM

After:
Screen Shot 2022-10-10 at 12 29 02 PM

zachz-indeed:
@mjhenkes - I added a before/after screenshot of the behavior.

I'm also getting an 'Unauthorized' status in CircleCI on the build step as of this morning. Do you have any ideas on what may have caused it/how to resolve? Thanks!

nagash77:
Hello @zachz-indeed !  Thank you for opening a PR to Cypress.  We have been reviewing your PR internally and have a few changes we would like before this gets merged:

1.  Please add two spaces and a dash like all the previous lines to the lines you have added.
2. Change to `Finished processing: X seconds` (no aligned-right seconds)
3. Ensure it works with progress lines between ‘Started…’ and ‘Finished…’ lines and please attach screenshots of the larger context in which your changes are displayed.

zachz-indeed:
Hi @nagash77 - thank you for the feedback! I have made the recommended changes. Here is a screenshot of what that looks like now:
Screen Shot 2022-11-04 at 12 56 56 PM

I will work on getting the snapshots updated as well. Please let me know if there's anything else I should update or provide. This is my first open source contribution, so learning a lot as I go :) Thanks again!

lmiller1990:
Hi @zachz-indeed - I can help you get this merged. It looks like some of the snapshots are not updated or conflicted.

If you want to merge in develop and resolve the conflicts, I can handle updating the snapshots - you just need to run the system test with `SNAPSHOT_UPDATE=1`, but it takes a really long time, since this PR is basically updating *every* snapshot (hundreds).

It's probably quicker if someone who works in the repo takes care of that, so if you just want to resolve the conflicts, I can help with the snapshot updates.

AtofStryker:
@lmiller1990 @zachz-indeed I should be able to take a look at this today and get the snapshots updated. Might take a while though 😅

packages/server/lib/util/print-run.ts
```diff
@@ -508,13 +508,16 @@ export function displayVideoProcessingProgress (opts: { videoName: string, video
...
           gray('-'),
           gray('Finished processing:'),
-            `${formatPath(opts.videoName, getWidth(table, 2), 'cyan')}`,
-            gray(dur),
+          gray(dur),
         ])
 
         console.log(table.toString())
 
         console.log('')
```
emilyrohrbough, line 538: Not sure we need a new line between the video details and the video output name.

AtofStryker:
one of the issues I am running into is that the processing time may/may not vary, which leads to some flakiness within our snapshots. Though I agree that I think the process time is more useful, the lift might be larger here to actually implement within our snapshots. I think for this reason we will likely need to go back to `(x second)`, or possibly put this behind a flag where we can toggle the actual time being printed vs omitting it. This way our snapshots do not flake, and end users get the benefit of seeing the time. How does this sound @emilyrohrbough @lmiller1990 ?

emilyrohrbough:
@AtofStryker The processing time should be overridden in the snapshots for `(x seconds)` like you are saying. It looks like these are captured correctly in the snapshot changes. Where are you seeing this at?

NEVERMIND - I see it's in the test failures. Yeah the regex likely needs updated to correctly account for this if this isn't being process correctly.

astone123:
Snapshots are being correctly processed now. We just need to update the snapshots for all of the system tests

AtofStryker:
I'm updating the snapshots now and should have this pushed up today

AtofStryker:
OK I got the snapshots updated 😅 . Took a while to figure it out but everything should now be green in CI

packages/driver/cypress/e2e/e2e/security.cy.js
```diff
@@ -3,4 +3,21 @@ describe('security', () => {
     cy.visit('/fixtures/security.html')
     cy.get('div').should('not.exist')
   })
+
```
lmiller1990: Was the intentionally merged here?

I think something went wrong during the rebase, it looks like some extra files in `proxy` and this one got added.

AtofStryker: something is off as this branch is missing 478407ec38b305cb5a55add1f089fbaf1920ceec. What I might do to make this change set simpler is to create a new branch with a single commit of the changes based off `develop`. @zachz-indeed you would likely need to recreate the branch locally when this happens

AtofStryker:
@lmiller1990 @emilyrohrbough @zachz-indeed I got the necessary change set based on [0843b0a](https://github.com/cypress-io/cypress/pull/23915/commits/0843b0a9198bfb8dd6ddde741b3641cb20d5e34c). This should only contain the changes to stdout, printing the run, and any snapshot updates needed

packages/driver/cypress/e2e/e2e/security.cy.js
lmiller1990: I think we need someone to own this, I think we should just make a new branch and pick the files we need. I can do it if no-one else has bandwidth, but probably not until the first week of Feb.

AtofStryker: I took care of it. [0843b0a](https://github.com/cypress-io/cypress/pull/23915/commits/0843b0a9198bfb8dd6ddde741b3641cb20d5e34c) is the commit with files we need with snapshots updated with the CI run in [f6aee7c](https://github.com/cypress-io/cypress/pull/23915/commits/f6aee7cc33b02b0a1184441b76f560f2370458f3). This should be ready for review, which I don't feel super comfortable reviewing simply because of how many changes I committed.

system-tests/__snapshots__/browser_crash_handling_spec.js
```diff
@@ -217,28 +220,33 @@ exports['Browser Crash Handling / when the browser process crashes in chrome / f
...
 
 ────────────────────────────────────────────────────────────────────────────────────────────────────
                                                                                                     
-  Running:  chrome_process_crash.cy.js                                                      (1 of 2)
+  Running:  chrome_tab_crash.cy.js                                                          (1 of 2)
 
 
 
-We detected that the Chrome process just crashed with code 'null' and signal 'SIGTRAP'.
+We detected that the Chromium Renderer process just crashed.
```
lmiller1990: I wonder why this changed :thinking:

lmiller1990 (commented):
One of the snapshots has some unexpected changes (maybe). The rest looks good. I think we should find out about that one snapshot, and see if that was correct.

system-tests/__snapshots__/browser_crash_handling_spec.js
lmiller1990: Oh I see, I think the snapshots in the file are just in a different order now. It's just confusing now GH shows this as a diff - I guess the order than @AtofStryker ran the updates locally was different to what the previous person who updated the snapshots ran, so the order they were wrote to the file is changed.

packages/driver/cypress/e2e/e2e/security.cy.js
lmiller1990: I took a look, I think it looks fine. Any particular change you are concerned with? It's a lot of files, but all the changes are the same and what I'd expect.

lmiller1990 (approved):

system-tests/__snapshots__/browser_crash_handling_spec.js
AtofStryker: Getting these snapshots to update correctly was a bit of a challenge 😅 . Sometimes the snapshots do update in a different order but should be the same relative to their export.

I noticed we have a conflict in `browser_crash_handling_spec` from the [improved memory work](https://github.com/cypress-io/cypress/commit/d1dc1258eafa6ff047abc913ebb930984e02a38d#diff-4c495b06189c0abcf578534fcb0e4f030df90cc9c4b3930c5923b61dfbb7025cR29). I am rebuilding that snapshot now to fix the conflict and will update @zachz-indeed 's branch shortly

BlueWinds (approved):

mjhenkes (approved):

packages/driver/cypress/e2e/e2e/security.cy.js
AtofStryker: nothing in particular. I think the two src file updates make sense and the snapshots are G2G and match what is expected

PR #23916, opened 2022-09-21
feat: CT stack traces

### LINKED ISSUE #21720: CT Stack Traces
## **Summary**
In E2E we show a stack trace if there is an error in the test. It is (5) here:

!image-2022-02-09-20-16-55-107.png|thumbnail! 

We want the same thing for CT. Right now, webpack dev server does not supply the correct source map - in fact, it doesn't have one at all. Also, even if it did, the seam between the dev-server and CT runner is not set up for stack traces.

For CT we go here https://github.com/cypress-io/cypress/blob/10.0-release/packages/driver/src/cypress/script_utils.ts#L49 as opposed to E2E which goes on and calls this https://github.com/cypress-io/cypress/blob/10.0-release/packages/driver/src/cypress/script_utils.ts#L36 which gets the source map.

## **Acceptance Criteria**
* Should return the source map from webpack dev server
* Should extract and display it for component testing

### **Resources**
Brian telling Lachlan about source maps for an hour or so:

Meeting Recording:
https://cypress-io.zoom.us/rec/share/hiYJyb9yjp7W7UKPbow0gG3gum1Az7kXA7Yq8XtQGU0AcNUEpdSTo43JL-Zs59uJ.-2OgUXVt2mTAcgZp
Access Passcode: **message @lmiller1990**

### **Open Pull Requests**
Any PRs needed for review

┆Issue is synchronized with this [Jira Task](https://cypress-io.atlassian.net/browse/UNIFY-1070) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/11242">image-2022-02-09-20-16-55-107.png</a>
┆author: Lachlan Miller
┆epic: CT Post Unification Release
┆friendlyId: UNIFY-1070
┆priority: Low
┆taskType: Task

mike-plummer:
- Closes #21720 

### User facing changelog
* Component tests that fail now display a code frame of the location of the error within the test runner. This includes failing Cypress commands and assertions in your tests, and can often show the location of errors within component code as well.

### Additional details
As much as possible we've tried to integrate CT stacktraces into the existing architecture of E2E sourcemap/codeframe logic. The majority of the changes in this PR are porting & adding tests to validate error-handling in CT mimics E2E whenever possible.

**Remaining Issues as Follow-Ons**
1. Svelte - sourcemaps are generated but do not line up correctly (yield `null` values from our lookups), so Svelte projects will not have codeframes initially (#23918)
2. Vite - many Vite sourcemaps display "off by one/two columns" issues. Code frames are available but point to a slightly incorrect position (#23919)
3. Uncaught errors in CT are caught/reported in the command log multiple times - existing issue in 10.x but noticeable now that we have more tests around error-handling in CT (#23920)

### Steps to test
This has pretty robust system tests, but it is fairly easily to test this out locally by building a React/Vue/Angular project and inserting test steps that intentionally fail.

**Before**
before

**After**
after-1
after-2

### PR Tasks


packages/app/cypress/e2e/runner/ct-framework-errors.cy.ts
```diff
@@ -0,0 +1,271 @@
+import type { ProjectFixtureDir } from '@tooling/system-tests/'
```
mike-plummer, line 1: This file is a quick smoke test for each of our major frameworks (and major versions thereof) to validate basic error conditions work as expected: uncaught error on mount, sync error, async error, and command failure.

Svelte is currently skipped due to issue with sourcemaps, will be a follow-on task

packages/app/cypress/e2e/runner/reporter-ct.errors.cy.ts
```diff
@@ -0,0 +1,704 @@
+import * as specLoader from './support/spec-loader'
```
mike-plummer, line 1: This file is mostly duplicated from `reporter.errors.cy.ts` - intent is to execute the same error validation in a CT context that is currently being done in E2E. Major changes:

1. Removed `cy.visit` tests (not valid for CT)
2. Remove `cy.route` and `cy.server` tests (long-deprecated, probably not worth spending time on for CT)
3. Replaced the `uncaught exceptions` block content to use a set of component-based errors

mike-plummer: I wasn't able to find a good way to avoid the duplication of this file - there are quite a few unique cases that would seriously complicate a shared version of this test, and kept running into issues with the "mututally exclusive e2e & ct `specPattern`" behavior that prevented using the same file for both. Good news is that the underlying tests are shared which is a win

packages/app/cypress/e2e/runner/support/spec-loader.ts
```diff
@@ -3,11 +3,11 @@ export const shouldHaveTestResults = ({ passCount, failCount, pendingCount }) =>
   failCount = failCount || '--'
```
mike-plummer, line 3: Had to increase some timeouts here to work around potentially-long bootstrap times for these tests when run under CT due to bundler overhead

packages/app/cypress/e2e/runner/support/verify-failures.ts
```diff
@@ -151,7 +152,7 @@ const verifyFailure = (options) => {
     if (uncaught) {
       cy.log('uncaught error has an associated log for the original error')
       cy.get('.command-name-uncaught-exception')
-      .should('have.length', 1)
+      .should(mode === 'component' ? 'have.length.gte' : 'have.length', 1)
```
mike-plummer, line 156: Uncaught errors in CT are reported multiple times, follow-on task to clean up that behavior

packages/app/cypress/e2e/runner/support/verify-failures.ts
```diff
@@ -206,10 +208,5 @@ export const verifyInternalFailure = (props) => {
 
     cy.get('.runnable-err-stack-trace')
     .should('include.text', stackMethod || method)
-
```
mike-plummer, line 209: Due to CT bundling behaviors these errors can be attributed to application/test code rather than internal Cypress logic, in which case a code frame is available. Edge case for certain Vite setups

system-tests/projects/runner-e2e-specs/cypress/e2e/errors/events.cy.js
```diff
@@ -1,18 +1,18 @@
 describe('event handlers', { defaultCommandTimeout: 0 }, () => {
```
mike-plummer, line 1: Refactoring the error conditions here to use features available in both E2E and CT tests so it can be used in both

packages/app/cypress/e2e/runner/reporter-ct-generator.ts
```diff
@@ -0,0 +1,706 @@
+import * as specLoader from './support/spec-loader'
```
mike-plummer, line 1: This file is mostly duplicated from `reporter.errors.cy.ts` - intent is to execute the same error validation in a CT context that is currently being done in E2E. Major changes:

1. Removed `cy.visit` tests (not valid for CT)
2. Remove `cy.route` and `cy.server` tests (long-deprecated, probably not worth spending time on for CT)
3. Replaced the `uncaught exceptions` block content to use a set of component-based errors

I wasn't able to find a good way to avoid the duplication of these tests - there are quite a few unique cases that would seriously complicate a shared version with e2e, and kept running into issues with the "mututally exclusive e2e & ct `specPattern`" behavior that prevented using the same file for both. Good news is that the underlying tests in the system test are shared which is a win

npm/vite-dev-server/src/plugins/cypress.ts
```diff
@@ -105,6 +105,23 @@ export const Cypress = (
...
+      try {
+        if (/\.js$/.test(id) && !/\/\/# sourceMappingURL=/i.test(code)) {
+          // The Vite dev server serves esnext, meaning many JS files aren't transpiled thus no sourcemaps are sent along. However,
+          // a sourcemap is generated by Vite which is in effect an "identity" sourcemap mapping 1-to-1 to the output file. We can
+          // grab this and pass it along as a sourcemap we want Vite to embed into the output, giving Cypress a sourcemap to use
+          // for codeFrame lookups.
+
+          return {
+            code,
+            map: this.getCombinedSourcemap(),
```
lmiller1990, line 122: Right, took me a bit to realize this is from the underyling Rollup API https://rollupjs.org/guide/en/#thisgetcombinedsourcemap

According to the Vite docs `map: null` should do the same thing? https://vitejs.dev/guide/api-plugin.html#transforming-custom-file-types

npm/webpack-dev-server/src/loader.ts
```diff
@@ -59,22 +61,33 @@ export default function loader (this: unknown) {
...
       }
       return accSpecLoaders
-  }, [loadSupportFile])
+  }, [])
+
+  if (${!!supportFile}) {
+    var supportFile = {
+      absolute: "${supportFileAbsolutePath}",
+      relative: "${supportFileRelativePath}",
+      relativeUrl: "/__cypress/src/cypress-support-file.js",
```
lmiller1990, line 83: Why do we hard code this here?

npm/webpack-dev-server/src/loader.ts
```diff
@@ -59,22 +61,33 @@ export default function loader (this: unknown) {
   ctx.cacheable(false)
   const { files, projectRoot, supportFile } = ctx._cypress
 
-  const supportFileAbsolutePath = supportFile ? JSON.stringify(path.resolve(projectRoot, supportFile)) : undefined
-
-  return `
-  var loadSupportFile = ${supportFile ? `() => import(${supportFileAbsolutePath})` : `() => Promise.resolve()`}
+  const supportFileAbsolutePath = supportFile ? path.resolve(projectRoot, supportFile) : undefined
```
lmiller1990: Hmmm I thought we needed `JSON.stringify` here to handle things like paths with spaces :thinking: I am surprised nothing broke when you changed this.

I wonder if we lost some test coverage at some point for the case of support file with spaces?

npm/webpack-dev-server/src/makeDefaultWebpackConfig.ts
```diff
@@ -51,6 +51,7 @@ export function makeDefaultWebpackConfig (
         ...(config.devServerConfig.framework === 'angular' ? { scriptLoading: 'module' } : {}),
       }),
     ],
+    devtool: 'inline-source-map',
```
lmiller1990, line 54: Does this take precedence over the user's wepbackConfig `devtool` setting? I wonder if we have a test for use case - where the user chose something different to `inline-source-map` in the webpack config?

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -162,6 +162,8 @@ export async function makeWebpackConfig (
     dynamicWebpackConfig,
   )
 
+  delete mergedConfig.output?.chunkFilename
```
lmiller1990, line 172: Why do we need this?

Should we just delete the entire `output` key?

packages/app/cypress.config.ts
```diff
@@ -52,6 +52,7 @@ export default defineConfig({
       // Delete this as we only want to honor it on parent Cypress when doing E2E Cypress in Cypress testing
       delete process.env.HTTP_PROXY_TARGET_FOR_ORIGIN_REQUESTS
       process.env.CYPRESS_INTERNAL_E2E_TESTING_SELF = 'true'
+      process.env.CYPRESS_INTERNAL_VITE_OPEN_MODE_TESTING = 'true'
```
lmiller1990, line 55: Is this required to make the e2e tests pass? I think this is needed specifically for Cypress in Cypress? If that is correct this could simply be `process.env.CYPRESS_INTERNAL_VITE_OPEN_MODE_TESTING = process.env.CYPRESS_INTERNAL_E2E_TESTING_SELF === 'true` - I think this line and the previous one (line 54) kind of need to be in lock step, they seem to go together.

packages/app/cypress/e2e/runner/ct-framework-errors.cy.ts
```diff
@@ -0,0 +1,271 @@
...
+[17, 18].forEach((reactVersion) => {
+  describe(`React ${reactVersion}`, {
+    viewportHeight: 768,
+    viewportWidth: 1024,
+    // Limiting tests kept in memory due to large memory cost
+    // of nested spec snapshots
+    numTestsKeptInMemory: 1,
+  }, () => {
+    before(() => {
+      cy.scaffoldProject(`react${reactVersion}` as ProjectFixtureDir)
```
lmiller1990: Here is a neat trick to avoid the type cast... the problem is `react${reactVersion}` is seen as `react${<any string or number>}` which is outside the constraints of `ProjectFixtureDir`. If the version array is immutable and constant, TS can infer the value, so:

```ts
const reactVersions = [17, 18] as const
                               ^^^^^^^^  this is the trick

reactVersions.forEach((reactVersion) => {
  describe(`React ${reactVersion}`, {
    viewportHeight: 768,
    viewportWidth: 1024,
    // Limiting tests kept in memory due to large memory cost
    // of nested spec snapshots
    numTestsKeptInMemory: 1,
  }, () => {
    before(() => {
      cy.scaffoldProject(`react${reactVersion}`)
    })
```

Now `react${reactVersion}` can be inferred as `react${17|18}` which fits the constraints. :100:

packages/app/cypress/e2e/runner/ct-framework-errors.cy.ts
```diff
@@ -0,0 +1,271 @@
...
+        'element-that-does-not-exist',
+      ],
+      codeFrameRegex: /Errors\.cy\.js:26/,
+      stackRegex: /Errors\.cy\.js:26/,
+    })
+  })
+})
+
+// TODO: Svelte sourcemaps are generated but are not working properly on Webpack or Vite
+describe.skip('Svelte', {
```
lmiller1990, line 297: I think we should log an issue that explains the problem and points to this file with some useful info, so someone can pick it up in the future without the need to be onboarded to stack traces in a 1:1 sync fashion, thoughts?

packages/app/cypress/e2e/runner/reporter-ct-generator.ts
```diff
@@ -0,0 +1,706 @@
...
+
+      verify('with expect().<foo>', {
+        column: [25, 26],
+        message: `expected 'actual' to equal 'expected'`,
+        codeFrameText: 'with expect().<foo>',
+      })
+    })
+
+    // Vite does not provide the `require` syntax used by this test
+    if (server !== 'Vite') {
```
lmiller1990: In this case would we want to say `server === 'Webpack'`? I think that's more clear - "When it's not Vite" vs "When it's webpack".

packages/app/cypress/e2e/runner/support/spec-loader.ts
```diff
@@ -20,6 +20,9 @@ export type LoadSpecOptions = {
   pendingCount?: number | string
   hasPreferredIde?: boolean
   projectName?: 'runner-e2e-specs' | 'session-and-origin-e2e-specs'
+  mode?: 'e2e' | 'component'
```
lmiller1990, line 23:
```suggestion
  mode: 'e2e' | 'component'
```

This is not optional, but fully understand if making it non-nullable massively bloats this PR.

packages/app/cypress/e2e/runner/support/verify-failures.ts
lmiller1990: Can we log a follow up issue and link it to the GA epic?

system-tests/projects/runner-e2e-specs/cypress/e2e/errors/events.cy.js
```diff
@@ -1,18 +1,18 @@
...
   it('event assertion failure', () => {
-    cy.on('window:load', () => {
+    cy.on('command:end', () => {
```
lmiller1990, line 3: The docs don't specify this as CT only, we should update them at some point https://docs.cypress.io/api/events/catalog-of-events#App-Events

lmiller1990 (commented):
Some questions/comments, I'm going to spend some time testing locally now.

npm/vite-dev-server/src/plugins/cypress.ts
```diff
@@ -105,6 +105,23 @@ export const Cypress = (
         return res.end(transformedIndexHtml)
       })
     },
+    transform (code, id, options?) {
+      try {
+        if (/\.js$/.test(id) && !/\/\/# sourceMappingURL=/i.test(code)) {
+          // The Vite dev server serves esnext, meaning many JS files aren't transpiled thus no sourcemaps are sent along. However,
+          // a sourcemap is generated by Vite which is in effect an "identity" sourcemap mapping 1-to-1 to the output file. We can
+          // grab this and pass it along as a sourcemap we want Vite to embed into the output, giving Cypress a sourcemap to use
+          // for codeFrame lookups.
```
lmiller1990:
```suggestion
          // The Vite dev server does not perform transpilation. Source code is served as it is written, thus no sourcemaps are generated by default. However,
          // Vite uses Rollup internally to support code transformation, which is in effect an "identity" sourcemap mapping 1-to-1 to the output file. Vite gives us access to the underyling Rollup API, including `getCombinedSourcemap`. 
          // We can grab this and pass it along as a sourcemap we want Vite to embed into the output, giving Cypress a sourcemap to use
          // for codeFrame lookups.
          // @see https://rollupjs.org/guide/en/#thisgetcombinedsourcemap
```

lmiller1990: Not suggesting using this as-is, but I think we might want to be quite descriptive about what's actually happening here It's quite complex and unless you know exactly how Vite works, it would be quite difficult to make the connection to Vite -> Rollup -> `getCombinedSourcemap`.

npm/vite-dev-server/src/plugins/cypress.ts
lmiller1990: Edit, apparently not - as per your comment above, we *do* need to call this function, but I don't quite know why. I still think I'm missing some understanding here. If you write something like:

```ts
interface Blah {
  foo: string
}

const blah = (): Blah => { 
  throw Error('argh')
  return { foo: "bar" }
}

blah()
```

And serve via Vite - it strips out TS, but this would require a sourcemap to get correct error location - if Vite doesn't generate sourcemaps, how does this error manifest in a regular Vite dev env?

I'll try it out. Hm.

lmiller1990: Seems Vite is generating a sourcemap normally. I wonder why we need to do this - why can't we do `map: null` and get it like the docs specify? 

image

npm/webpack-dev-server/src/helpers/angularHandler.ts
```diff
@@ -253,8 +254,16 @@ async function getAngularCliWebpackConfig (devServerConfig: AngularWebpackDevSer
   return config
 }
 
+function removeSourceMapPlugin (config: Configuration) {
+  config.plugins = config.plugins?.filter((plugin) => {
+    return (plugin as any).constructor.name !== 'SourceMapDevToolPlugin'
```
lmiller1990: I wonder if we can get types, we seem to have them here when we do something similar:

https://github.com/cypress-io/cypress/blob/develop/npm/webpack-dev-server/src/makeWebpackConfig.ts#L53

Alternatively, if we aren't sure of the type, maybe we should be defensive

```suggestion
    return (plugin as any)?.constructor?.name !== 'SourceMapDevToolPlugin'
```

lmiller1990 (commented):
Great work! I asked a bunch of questions.

I tried it out a bit locally and ran the tests, it seems to working as expected.

packages/app/cypress/e2e/runner/support/verify-failures.ts
lmiller1990: Looks like you already did!

packages/app/cypress/e2e/runner/ct-framework-errors.cy.ts
lmiller1990: Edit: you did that already, I found it.

npm/webpack-dev-server/src/makeDefaultWebpackConfig.ts
astone123: I think we can test this in `makeWebpackConfig.spec.ts` by updating one of the configs to have `devtool: 'eval'` (or something similar) and then the snapshot should still result in `devtool: 'inline-source-map'`

npm/vite-dev-server/src/plugins/cypress.ts
mike-plummer: @lmiller1990 I think it comes down to the difference between serving a "pure" JS file vs a typescript or JSX sort of file. The `getCombinedSourcemap()` workaround was necessary for Vite to provide a `sourceMappingURL` entry within pure JS files it serves up (for example, a basic Cypress test without any crazy imports, typescript, JSX, etc).

Vite has to transpile TS and JSX files so a sourcemap is appended automatically, but for JS files that don't require transpilation (just served up as-is) the sourcemap was not getting appended unless I added that explicit call.

npm/webpack-dev-server/src/makeWebpackConfig.ts
mike-plummer: This is a condition that @ZachJW34 uncovered during his initial spike - I believe that if there was a custom `chunkFilename` it was causing issues with the support file. If he can confirm (or correct me) I'll add a comment to document for the future

packages/app/cypress/e2e/runner/ct-framework-errors.cy.ts
mike-plummer: I'll add some issue links to my TODOs for future reference

packages/app/cypress/e2e/runner/support/spec-loader.ts
mike-plummer: Absolutely agree that's better, but I did this intentionally to avoid bloating the PR. Ends up touching quite a few places that felt like a unnecessary complication

system-tests/projects/runner-e2e-specs/cypress/e2e/errors/events.cy.js
mike-plummer: The event itself works fine in CT, but the test was doing a `cy.visit` (which doesn't work in CT) and using this paired event listener. I had to replace the `cy.visit` with a different test condition, thus the change in events

packages/app/cypress.config.ts
mike-plummer: Yes, without this I was getting "cross origin" errors when running some Vite-related system tests. As for your suggestion, I don't see how that would be any different - since we set `CYPRESS_INTERNAL_E2E_TESTING_SELF` to `true` right above it would be functionally equivalent, wouldn't it?

npm/webpack-dev-server/src/helpers/angularHandler.ts
mike-plummer: Seems like this cast isn't necessary after all, removing

npm/webpack-dev-server/src/loader.ts
mike-plummer: Providing the absolute & relative fields here was a change Zach proposed to align how files were queued up for loading in E2E and CT. Since much of this logic is really occurring outside the safety of TS it seemed like a good idea to align the contracts even though we're just using the `load` fn at the moment

npm/webpack-dev-server/src/loader.ts
mike-plummer: Good question - this came out of the spike so I may ask @ZachJW34 to chime in - happen to remember why this change was made?

npm/webpack-dev-server/src/makeDefaultWebpackConfig.ts
mike-plummer: @lmiller1990 The output of this fn is overlaid atop the user/framework config on line 161 of `makeWebpackConfig` so this should take precedence over any user configs. I believe there are existing tests validating that behavior but I can look at expanding those to specifically check the `devtool` field if we feel that would be valuable

npm/vite-dev-server/src/plugins/cypress.ts
mike-plummer: @lmiller1990 I've updated the comment a bit to provide more context, let me know if you would like to see additional changes

npm/webpack-dev-server/src/makeDefaultWebpackConfig.ts
astone123: Added more coverage for this in https://github.com/cypress-io/cypress/pull/23916/commits/dac21273aee95dcd4bbdde10f0223203e494fde0 just for good measure

astone123 (approved):
There are some places in the dev server tests where the stack trace doesn't show in the command log, but after talking to Mike yesterday, these specific cases makes sense. I updated all of the e2e tests in the dev server packages to "document" this behavior.

Looks good 🎉

npm/webpack-dev-server/src/makeWebpackConfig.ts
ZachJW34: Some frameworks (like Next.js) change this value. Next.js specifically sets it to `'static/chunks/[name].js'` which means the network request to grab a spec becomes `http://localhost:xxxx/<dev-server-public-path>/static/chunks/spec-<x>.js`. Deleting this normalizes all specs to `http://localhost:xxxx/<dev-server-public-path>/spec-<x>.js` which is required since we need to make a fetch request to extract the source map. There might be other ways but this seemed easiest.

I wouldn't delete the `output` key outright, there are a bunch of options that could affect the compilation (like `module`).

npm/webpack-dev-server/src/loader.ts
ZachJW34: I thought the `JSON.Stringify` simply wrapped the string in quotes, so rather than do `load: () => import("${thing}")` you can do `load: () => import(${JSON.stringify(thing)})`. I'd be fine reverting back to using `JSON.stringify`, just in case it's doing more than what I thought.

npm/webpack-dev-server/src/loader.ts
ZachJW34: We do use the `relativeUrl` when making the fetch request (to get the source map), `absolute` and `relative` was just my attempt at making it symmetrical with e2e. When I was spiking into this I was running into an issue getting the paths and to work so I moved it from the reduce and hardcoded it. Definitely could be cleaned up a bit.

Also, this can be hardcoded due to the magic comment `webpackChunkName` below

packages/app/cypress/e2e/runner/ct-framework-errors.cy.ts
ZachJW34: Is there a reason we don't have `Next` or `Nuxt` in this test?

packages/driver/src/cypress/stack_utils.ts
```diff
@@ -12,6 +12,8 @@ import { getStackLines, replacedStack, stackWithoutMessage, splitStack, unsplitS
 
 const whitespaceRegex = /^(\s*)*/
 const customProtocolRegex = /^[^:\/]+:\/{1,3}/
+// string containing `webpack:///./`
+const webpackDevtoolNamespaceRegex = /webpack:\/{3}\.\//
```
ZachJW34: I had to tweak this a bit in the original spike due to custom-namespaces. For example, this is Next's: 
Screen Shot 2022-09-22 at 12 52 27 PM

You can see the namespace show up in the relative path e.g. `N_E_/pages/index.cy.js`. The original regex was `/webpack:\/\/(.*)?\.\//` to account for this, did this not work out?

Maybe we can use [output.devtoolModuleFilenameTemplate](https://webpack.js.org/configuration/output/#outputdevtoolmodulefilenametemplate) to enforce consistency across frameworks.

mike-plummer: @ZachJW34 Ah, that makes more sense now. The issue I ran into was that the regex you crafted broke relative paths if a spec was opened from outside the project root like `../e2e/specName.cy.ts`. I see what the regex intent was now, I'll take a look to see if there's a way to craft one that scrubs the custom namespace without killing relative pathing

mike-plummer: @ZachJW34 I've adjusted the regex and it seems to work as expected for Next.js apps (as well as for the system test that failed with the original pattern)

npm/vite-dev-server/src/plugins/cypress.ts
lmiller1990: Oh right, so you are saying it will automatically append the sourceMap for anything other than plain JS... this is only required for the case of plain JS (eg, does it "just work" for TSX, for example, even if I remove this?) This is my understanding based on your comment. I will try this to find out if my assumption is correct.

system-tests/projects/runner-e2e-specs/cypress-ct-webpack.config.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  numTestsKeptInMemory: 0,
+  video: false,
+  e2e: {
+    specPattern: 'no-matches/**.cy.*',
+  },
+  component: {
+    specPattern: 'cypress/e2e/**/*.cy.{js,jsx,ts,tsx}',
```
lmiller1990: Just to check; what's the value running **all** of these? Are they related to stack traces? It looks like some might be, like the error ones. I'm not sure if they are all useful (I need to look more closely).

npm/vite-dev-server/client/initCypressTests.js
```diff
@@ -24,12 +24,22 @@ if (supportFile) {
```
lmiller1990, line 24: Not related to this file but @brian-mann pointed out we will also want to run 

- https://github.com/cypress-io/cypress/blob/develop/system-tests/test/error_ui_spec.ts
- https://github.com/cypress-io/cypress/blob/develop/system-tests/projects/webpack-preprocessor/cypress/e2e/failing.cy.js
- https://github.com/cypress-io/cypress/blob/develop/system-tests/projects/webpack-preprocessor-awesome-typescript-loader/cypress/e2e/failing.cy.ts

As well, but with CT.

These are the ones that verify the typescript compiler is correctly monkey patched to generate the correct source maps in E2E when using webpack-preprocessor. 

The monkey patch is here: https://github.com/cypress-io/cypress/blob/develop/npm/webpack-preprocessor/lib/typescript-overrides.ts#L39-L45

It's possible this is actually not required. if {webpack,vite}-dev-server already do this (or something like this). We can find out by running these tests in CT, and see if they pass. If they don't, we might need something similar.


Ref: https://cypressio.slack.com/archives/C011BAPC614/p1662579814502909

npm/webpack-dev-server/src/loader.ts
lmiller1990: image

Seems to add backslashes, at least in the console - I'm pretty sure this was here for a specific reason, might be good to revert it until we know it's safe to remove.

npm/webpack-dev-server/src/loader.ts
lmiller1990: Right, I see. Definitely agree on aligning with E2E where possible.

packages/app/cypress.config.ts
lmiller1990: Yeah, my suggestions seems weird now I'm re-reading it - line 52 makes this more than clear enough.

npm/webpack-dev-server/src/makeWebpackConfig.ts
lmiller1990: Right, gotcha. I think the comment has since been added which clarifies a lot, thanks!

packages/app/cypress/e2e/runner/support/spec-loader.ts
lmiller1990: :+1:

system-tests/projects/runner-e2e-specs/cypress-ct-webpack.config.js
mike-plummer: They're not all being run, just specifying a pattern to allow those specs to be found and run individually in the system test. Stacktraces is specifically only running the set of specs under `cypress/e2e/errors/*` so I could make the pattern more specific but it shouldn't cause any harm to match more specs than needed since only a subset is actually being run

npm/vite-dev-server/src/plugins/cypress.ts
mike-plummer: @lmiller1990 Correct - this is here to handle the case where *source JS* files would not get sourcemaps. Pretty much everything else did without this change

npm/vite-dev-server/client/initCypressTests.js
mike-plummer: I am working this [in another branch](https://github.com/cypress-io/cypress/compare/mikep/21720-ct-stack-traces...mikep/21720-stacktrace-ts-patch) to avoid complicating this PR any further. Will open as a follow-on PR in case there are complications so we don't block the main stacktraces featureset from release - seems to be working as expected once I ported the patch logic over except for a special case for Angular

rachelruderman:
Has this been tested on Windows as well? If not, would be good to also test manually, either on a Windows VM or real machine, since the driver tests do not run on the Windows CI workflow

rockindahizzy (approved):
This is monstrous. Nice work!

packages/driver/src/cypress/script_utils.ts
```diff
@@ -25,11 +24,15 @@ const extractSourceMap = ([script, contents]) => {
...
+  return Bluebird.each(scripts, (_script: any) => {
+    const [script, contents] = _script
+
+    if (script.load) {
+      return script.load()
+    }
 
-  return null
+    return specWindow.eval(`${contents}\n//# sourceURL=${script.fullyQualifiedUrl}`)
+  })
```
ZachJW34, line 35: Small cleanup we can do below in `runScripts`, the `if (scripts.length && typeof scripts[0] === 'function') {` is never valid now since we are sending up the spec payload now for CT, so it can be simplified to:

```ts
export default {
  runScripts: runScriptsFromUrls
}
```

ZachJW34 (approved):
L
G
T
M

packages/app/cypress/e2e/runner/ct-framework-errors.cy.ts
ZachJW34: bump

mike-plummer: @ZachJW34 No particular reason; I think I got distracted making sure we had base framework coverage. Unfortunately it looks like the existing Next & Nuxt system-test projects don't extend the base React and Vue fixtures, so I would have to duplicate the `Error` components and test suites I created out of those. Happen to know of a good way to avoid that? Not complex, just annoying to duplicate that sort of code

npm/vite-dev-server/src/plugins/cypress.ts
lmiller1990: Makes sense, since normally you don't need source maps for JS - nothing to map :tada: This worked out surprisingly well, glad Vite exposes the Rollup API.

npm/vite-dev-server/src/plugins/cypress.ts
lmiller1990: Looks good!

npm/vite-dev-server/client/initCypressTests.js
lmiller1990: Ok, sounds good

lmiller1990:
This looks good. I will hold off on :heavy_check_mark: since we need to test windows, the final frontier.

lmiller1990 (dismissed):
Just need to verify windows :)

I pushed a commit to run on windows https://github.com/cypress-io/cypress/pull/23916/commits/537c927ad62d8d4a5dcd07658b1c6c48bb37b612

Most of the windows fails are the same, example - first test in `Errors.cy.jsx` in react17:

## Linux (correct, stack trace + points to file)
image

## Windows (not correct, no stack trace, points to dev server)

image

packages/app/cypress/e2e/runner/ct-framework-errors.cy.ts
lmiller1990: Seems like a smoke test for Next.js might be nice... it's our most popular framework. It also looks like there is a `project-fixtures/next`, maybe we can piggy back off that.

lmiller1990:
I found the root cause, it's here: https://github.com/cypress-io/cypress/blob/238ead3c53399c022f7c99236ccc709768395a61/packages/driver/src/cypress/source_map_utils.ts#L72. I think the source map consumer map will need to be posix, even on windows.

## Linux

image

## Windows

image

Ok, fixed! https://github.com/cypress-io/cypress/pull/23916/commits/f214dacead2b15fa232e902f811e1f432fd7a249

I've also merged in https://github.com/cypress-io/cypress/pull/23975 so we can actually run on windows CI and see if anything breaks.

system-tests/projects/runner-e2e-specs/package.json
```diff
@@ -0,0 +1,20 @@
+{
```
lmiller1990, line 1: I think we should consider against adding this here - this will mean `yarn install` runs on every project using `runner-e2e-specs`, which is a LOT - often when it's not needed (in the case of E2E). It looks like this might slow CI down quite a bit.

Also this is runner-e2e-specs - maybe we want `runner-ct-specs`?

lmiller1990: Also every project using runner-e2e-specs seems to be broken on windows CI - finding out why...

PR #23928, opened 2022-09-22
chore: Build driver with vite

sainthkh:
- Closes 

### PR Tasks

### DIFF
--- a/patches/@cypress+unique-selector+0.4.4.patch
+++ b/patches/@cypress+unique-selector+0.4.4.patch
@@ -1,0 +1,293 @@
+diff --git a/node_modules/@cypress/unique-selector/lib/index.esm.js b/node_modules/@cypress/unique-selector/lib/index.esm.js
+new file mode 100644
+index 0000000..60a443e
+--- /dev/null
++++ b/node_modules/@cypress/unique-selector/lib/index.esm.js
+@@ -0,0 +1,265 @@
++'use strict';
++
++Object.defineProperty(exports, "__esModule", {
++  value: true
++});
++
++var _getID = require('./getID');
++
++var _getClasses = require('./getClasses');
++
++var _getCombinations = require('./getCombinations');
++
++var _getAttributes = require('./getAttributes');
++
++var _getNthChild = require('./getNthChild');
++
++var _getTag = require('./getTag');
++
++var _isUnique = require('./isUnique');
++
++var _getParents = require('./getParents');
++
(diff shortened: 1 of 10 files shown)


patches/@cypress+unique-selector+0.4.4.patch
```diff
@@ -0,0 +1,293 @@
+diff --git a/node_modules/@cypress/unique-selector/lib/index.esm.js b/node_modules/@cypress/unique-selector/lib/index.esm.js
+new file mode 100644
+index 0000000..60a443e
+--- /dev/null
```
lmiller1990, line 4: Can we just fix an release a new cypress unique selector instead of patch package?

lmiller1990 (commented):
Nice to see this!! I'm not sure how much is left but excited to see this, I'm guessing there's still a bit to do - I thought we had some kind of circular dependency problem in driver? 

I wonder why we need to polyfill `path` and `stream` and `util` - what's using that in driver?

patches/@cypress+unique-selector+0.4.4.patch
sainthkh: I just wanted to test if everything works with this change. I'll open a new PR at `unique-selector` later.

sainthkh:
I checked the circular dependency problem with [`dpdm`](https://github.com/acrazing/dpdm). The result is that our circular dependency only exists for TypeScript types. So, we could simply move on.

I polyfill them because they were in `vite.config.mjs` of `driver`. I'll remove them and see what's going on.

patches/@cypress+unique-selector+0.4.4.patch
sainthkh: As you can see below, errors are flooding.

lmiller1990:
I think it's fine to polyfill for now, just curious - it would be super nice to eventually move everything to ESM eventually and have a simple `vite.config` - our `webpack.config` is immensely complex right now, one of the nice things about Vite is you generally need less config.

chrisbreiding:
@sainthkh Have you been able to make any progress on this PR?

sainthkh:
@chrisbreiding 

I learned that Webpack generates some polyfills for global variables while Vite does not. And many packages have their own way of handling them, so it's a bit hard for me to finish this PR. Maybe researching the polyfill code of Webpack polyfill for `global`, `window`, `process` might help.

I'm sorry. I think I should close this for now.

PR #23948, opened 2022-09-22
chore: remove rawJson from Cypress.state()

### LINKED ISSUE #23945: Remove `rawJson` from stored Cypress state
### What would you like?

https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cypress.ts#L222

Check the config value received when Cypress starts and you'll see we are storing the `rawJson` with the value & resolved.from values in Cypress.state().

We collect this date for presenting the config resolution in the App, but the driver doesn't need/use this info.

emilyrohrbough:
- Closes #23945

[develop/packages/driver/src/cypress.ts#L222](https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cypress.ts?rgh-link-date=2022-09-22T18%3A42%3A39Z#L222)

Check the config value received when Cypress starts and you'll see we are storing the rawJson with the value & resolved.from values in Cypress.state().

We collect this date for presenting the config resolution in the App, but the driver doesn't need/use this info.


lmiller1990:
Since this is not part of the public API (undocumented) I don't know if we need to mark it as BREAKING, do we? BREAKING is usually for public API changes (I understand this might technically break someone, though, but it's equally weird to see a deprecation in a major that was never an official feature to begin with).

mjhenkes (approved):

emilyrohrbough:
yeah, I can go either way, but would rather we just push it out in 11.0 just in case.

ryanthemanuel (approved):

PR #23953, opened 2022-09-22
feat: add RecordRunModal component

### LINKED ISSUE #23764: [ACI post-project connection] Create Cloud Connection Modal “Record a Run” state
**Context**

After logging in, users who have not connected their project are currently invited to do so and can complete connecting their project through what feels like the same modal that was present after login. When the project is connected, the modal is closed and the user is back in the app at whatever point they happened to be on.

**Requested Change**
Create a new state in the LoginConnect modal flow that shows the command needed to record a run and allows the user to copy the command to their clipboard. It should match this design, available in this [Figma board]()https://www.figma.com/file/V0lCq4uBplf9J3GZRiD02m/ACI-Fast-Follows?node-id=19661%3A172759)

Screen Shot 2022-09-09 at 5 15 20 PM

For this ticket, just add the component but do not hook it up. That has been split out into a separate issue (#23766). The exact logic for when this should be shown will be exposed in #23765, but for the sake of doing things in parallel, there's no reason to wait on actually making the component.

**Acceptance Criteria**

- This new "Record Run" variant of the modal is displayed in a component test and has a percy snapshot
- It matches the design in Figma
- It has a Learn More Link: https://on.cypress.io/cypress-run-record-key
- The Learn Morn link uses the UTM params for where the flow was initially triggered from; same as used for login and other links in the Connect Dashboard Modal Flow; for example, if the flow was triggered from the nav, use medium="Nav" and campaign="Log In"
- Esc, click away, and the X in the top-right corner close the modal
- Component is not displayed in the app at all, and can be merged with no disruption to existing functionality.

**Suggested Implementation** (as a guide for estimates)

- Extract TerminalPrompt + recordCommand logic from RecordBanner.vue to get a reusable RecordPrompt component - this already matches the Figma design, no need to re-implement.
- Use the new RecordPrompt in RecordBanner 
- Use the new RecordPrompt in a new "RecordRunModal" that would be rendered by of CloudConnectModals.vue or LoginConnectModals.vue
- Add component tests convering all items in the AC

astone123:
- Closes #23764

### User facing changelog

n/a

### Additional details

This PR adds the `RecordRunModal` UI component and refactors some logic from `RecordBanner` to be reusable between these two components. The work to integrate this component into the UI will be done in #23766

### Steps to test

The component isn't actually used anywhere in the UI yet, so take a look at the test

1. Run `RecordRunModal.cy.tsx` in `frontend-shared`
2. Look at the component and ensure that it meets all of the criteria laid out in the issue
3. Verify that `RecordBanner` still works as before

n/a

### PR Tasks


packages/frontend-shared/src/gql-components/RecordPrompt.vue
```diff
@@ -0,0 +1,53 @@
...
+import { gql, useQuery } from '@urql/vue'
+import { useI18n } from '@cy/i18n'
+import TerminalPrompt from '@cy/components/TerminalPrompt.vue'
+import { computed } from 'vue'
+import { RecordPromptDocument } from '../generated/graphql'
+
+const { t } = useI18n()
+
+gql`
+query RecordPrompt {
```
lmiller1990: I *think* it makes sense for modals to have their own `useQuery` (as opposed to fragment + props).

I read https://relay.dev/docs/principles-and-architecture/thinking-in-relay/ in more delay and it seems FB, at least, are heavily using the "big prop object + single query" approach... I see the value in this, but also interested in exploring the "`useQuery` + fetch from the cache" too, and seeing what works best for our use case.

packages/frontend-shared/src/gql-components/RecordPrompt.vue
```diff
@@ -0,0 +1,53 @@
...
+  }
+}
+`
+
+const firstRecordKey = computed(() => {
+  return (query.data?.value?.currentProject?.cloudProject?.__typename === 'CloudProject' && query.data?.value?.currentProject.cloudProject.recordKeys?.[0]?.key) ?? '<record-key>'
+})
+
+const recordCommand = computed(() => {
+  const componentFlagOrSpace = query.data?.value?.currentProject?.currentTestingType === 'component' ? ' --component ' : ' '
```
lmiller1990:
```suggestion
  const componentFlagOrSpace = query.data.value?.currentProject?.currentTestingType === 'component' ? ' --component ' : ' '
```

Nit... I think `data` is always defined and doesn't need `?.`.

packages/frontend-shared/src/gql-components/RecordRunModal.cy.tsx
```diff
@@ -0,0 +1,34 @@
...
+  it('renders open', () => {
+    cy.mount(<RecordRunModalVue isModalOpen={true} close={cy.stub()} utmMedium="Nav" />)
+
+    cy.contains('Record your first run').should('be.visible')
+    cy.findByTestId('copy-button').should('be.visible')
+
+    cy.percySnapshot()
+  })
+
+  it('calls close when X is clicked', () => {
```
lmiller1990, line 22: Not sure if it's covered somewhere else but do we have coverage for "closes standard modal when ESC is pressed"? I see that as a requirement, not sure if we have the coverage -- if we don''t might be a good time to add it.

packages/frontend-shared/src/gql-components/RecordRunModal.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+  <StandardModal
+    class="transition transition-all duration-200"
+    variant="bare"
+    :title="t('specPage.banners.record.title')"
+    :model-value="isModalOpen"
+    :help-link="helpLink"
+    :no-help="!helpLink"
+    data-cy="record-run-modal"
+    @update:model-value="close"
```
lmiller1990, line 10: I don't think we are doing `v-model` correctly (anywhere, lol, we should make this change everywhere). [Docs on v-model](https://vuejs.org/guide/components/events.html#usage-with-v-model).

```suggestion
    :help-link="helpLink"
    :no-help="!helpLink"
    data-cy="record-run-modal"
    v-model="isModalOpen"
```

`v-model` is basically a shorthand for:

```html
<CustomInput
  :modelValue="searchText"
  @update:modelValue="newValue => searchText = newValue"
/>

<!-- can be simplifed -->
<CustomInput v-model="searchText" />
```

packages/frontend-shared/src/gql-components/RecordRunModal.vue
```diff
@@ -0,0 +1,36 @@
...
+
+const props = defineProps<{
+  isModalOpen: boolean
+  close: () => void
+  utmMedium: string
+  utmContent?: string
+}>()
+
+const helpLink =
+`https://on.cypress.io/cypress-run-record-key?utm_medium="${props.utmMedium}",utm_source="${getUtmSource()}"${props.utmContent ? `,utm_content="${props.utmContent}"` : ''}`
```
lmiller1990: Will any of these props values ever change? If so, we'd want this to `computed`.

packages/app/src/specs/banners/RecordBanner.vue
```diff
@@ -16,47 +16,16 @@
...
-        recordKeys {
-          id
-          key
-        }
-      }
-    }
-  }
-}
-`
+import RecordPrompt from '@packages/frontend-shared/src/gql-components/RecordPrompt.vue'
```
lmiller1990: Should this be in `packages/app`? I don't think we will ever use the `RecordPrompt` in Launchpad.

packages/frontend-shared/src/gql-components/RecordPrompt.vue
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+}
+`
+
+const firstRecordKey = computed(() => {
+  return (query.data?.value?.currentProject?.cloudProject?.__typename === 'CloudProject' && query.data?.value?.currentProject.cloudProject.recordKeys?.[0]?.key) ?? '<record-key>'
+})
+
+const recordCommand = computed(() => {
```
lmiller1990, line 21: Seems we are missing coverage around this, or are we happy to rely on Percy for testing this?

lmiller1990 (commented):
Some minor comments here and there :+1:

packages/frontend-shared/src/gql-components/RecordRunModal.cy.tsx
astone123: We do have coverage for this in `StandardModal.cy.tsx`
```
    it('closes with Esc key', () => {
      cy.get('body').type('{esc}')
      cy.get('@updateSpy').should('have.been.calledOnceWith', false)
    })
```

packages/frontend-shared/src/gql-components/RecordRunModal.vue
astone123: When I try this I get an ESLint error where I try to pass `v-model="isModalOpen"`
```
Unexpected mutation of "isModalOpen" prop.
```
and my test fails because the `close` prop function is never called when I click X 🤔

packages/frontend-shared/src/gql-components/RecordRunModal.vue
astone123: I'm not sure what the answer is yet... I think we can handle that during the implementation

packages/frontend-shared/src/gql-components/RecordPrompt.vue
astone123: We cover that in `RecordBanner.cy.tsx`, so I think just covering it with Percy here is fine

packages/app/src/specs/banners/RecordBanner.vue
astone123: I put it in `frontend-shared` because that's where `Auth` is, which I think is where `RecordRunModal` will be used 🤔 not sure where to put it

mike-plummer: If I understand the linked issue correctly the intent is to eventually give users a push to record as part of the login flow (login modal => connect modal => record modal), so it likely will need to be a shared component since the login modal is used in launchpad & app

packages/frontend-shared/src/gql-components/RecordRunModal.vue
mike-plummer: Are we sure about the formatting of this? Should the query params really be separated by `,` instead of `&`? Why not use the existing `getUrlWithParams` utility function to build this?

packages/frontend-shared/src/gql-components/RecordPrompt.vue
mike-plummer: I'm a bit torn on it in this case. I'm in favor of using a separate `useQuery` when we aren't 100% sure we'll need the data or if the data is expensive (e.g, cloud project runs) or may need to be re-fetched, but in this case the data is technically cloud data but it's pretty lightweight and other parts of the app already fetch it. Seems like encapsulating the query here really just makes this difficult to test.

Personally I would wrap this with a data-managing component with the sole responsibility of doing the query and mapping the result to props, then this component would just accept props and be very simple to test the actual input => output behaviors of without wrangling GQL mocks

packages/frontend-shared/src/gql-components/RecordRunModal.vue
astone123: Oh wow, yeah having `,` as the separator was a mistake. I didn't see that there was a utility function, will definitely use that 👍🏻

packages/frontend-shared/src/gql-components/RecordPrompt.vue
lmiller1990: >  I'm in favor of using a separate useQuery when we aren't 100% sure we'll need the data or if the data is expensive (e.g, cloud project runs) or may need to be re-fetched

This was what I was thinking. I can see an argument for "keep doing it how we do it", though. I'm not really sold on changing it **just** for the purpose of testing - we could easily just use `cy.intercept` to test the modal.

Could go either way - I've seen code bases making use of `useQuery` everywhere (mainly Apollo based one) and others doing the "one big prop" pattern (mainly Relay). I guess it's just competing philosophies... in which case, I wonder if we should go with "one big prop" consistently?

Definitely good to standardize how we'd like to handle this, document it, and even considering linting for it.

packages/frontend-shared/src/gql-components/RecordRunModal.vue
lmiller1990: Hm ok... weird. I think `<StandardModal>` needs a bit of a refactor. I think

```html
<StandardModal 
  :open="open"
  @close="close"
/>
```

Seems more readable than `@update:model-value="close"`. Either way, happy to go with this for now, seems this is the pattern we are using everywhere else.

packages/frontend-shared/src/gql-components/RecordPrompt.vue
lmiller1990: @astone123 thoughts? Maybe we do want to stick with the single query/prop pattern for now?

astone123: I ended up separating the gql logic out into an "adapter" component and then passing the data down into `RecordPrompt`. I also wrote separate component tests for `RecordPrompt`

packages/app/src/specs/banners/RecordBanner.vue
```diff
@@ -16,47 +16,16 @@
     }"
     @update:model-value="value => emit('update:modelValue', value)"
   >
-    <p class="mb-24px">
```
warrensplayer, line 19: This copy should remain part of the banner. The RecordPrompt should just encapsulate the logic for creating the prompt itself.  That way, different places in the app that use it can have different copy surrounding the prompt.

packages/frontend-shared/src/gql-components/RecordPrompt.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+  <p class="mb-24px">
```
warrensplayer: Leave the copy block out of the RecordPrompt component so that the surrounding copy can be customized where it is used.

packages/frontend-shared/src/gql-components/RecordRunModal.vue
```diff
@@ -0,0 +1,43 @@
...
+    class="transition transition-all duration-200"
+    variant="bare"
+    :title="t('specPage.banners.record.title')"
+    :model-value="isModalOpen"
+    :help-link="helpLink"
+    :no-help="!helpLink"
+    data-cy="record-run-modal"
+    @update:model-value="close"
+  >
+    <div class="max-w-175 py-7 px-6 text-gray-600">
```
warrensplayer, line 12: Add a copy block here that would have the same content as the banner for this implementation.  This allows the `RecordPrompt` to just encapsulate the logic around the TerminalPrompt and the logic to create it for recording a run.

warrensplayer (changes requested):
The `RunsEmpty.vue` component has the same logic to construct a TerminalPrompt for recording runs.  It should also be refactored to use the new RecordPrompt component.  To accomplish that, it would require a change to the RecordPrompt component to pull out the copy part above the prompt.  I have noted that in other comments within the code.

warrensplayer (changes requested):
The `RunsContainer.vue` is still spreading the `RunsEmpty` fragment on line 82 that no longer exists.

packages/app/src/runs/RunsEmpty.cy.tsx
```diff
@@ -1,14 +1,25 @@
...
+        id: 'cloud_id',
+        recordKeys: [{
+          __typename: 'CloudRecordKey',
+          id: 'recordKey1',
+          key: 'abcd-efg-1234',
+        }],
+      } as any,
+    } as any
+
+    cy.contains('Record your first run').should('be.visible')
```
mike-plummer: nitpick: I think elsewhere we're trying to avoid hardcoding strings and instead use the message bundle:

```
...
cy.contains(defaultMessages.specPage.banners.record.title).should('be.visible')
```

packages/frontend-shared/src/gql-components/RecordRunModal.cy.tsx
```diff
@@ -0,0 +1,40 @@
...
+    cy.mount(<RecordRunModalVue isModalOpen={true} close={cy.stub()} utmMedium="Nav" utmContent="content" />)
+
+    cy.contains('Need help').should('have.attr', 'href', 'https://on.cypress.io/cypress-run-record-key?utm_medium=Nav&utm_source=Binary%3A+Launchpad&utm_content=content')
+  })
+
+  it('sends UTM parameters with help link without UTM content prop', () => {
+    cy.mount(<RecordRunModalVue isModalOpen={true} close={cy.stub()} utmMedium="Nav" />)
+
+    cy.contains('Need help').should('have.attr', 'href', 'https://on.cypress.io/cypress-run-record-key?utm_medium=Nav&utm_source=Binary%3A+Launchpad&utm_content=')
+  })
```
mike-plummer, line 42: Can we add a test case for if the gql query fails? Seems like we may end up with an empty/invalid record command if we fail to retrieve record keys for whatever reason

mike-plummer (changes requested):

packages/app/src/runs/RunsEmpty.cy.tsx
lmiller1990: It looks like a mixed bag - some tests hard-code it.

I find the hard-coded strings version of the test *much* more readable and easier to search for. I guess we should probably standardize this, though.

lmiller1990 (approved):
Seems good to me, just need to address a few nits. I'll approve, assuming those are going to get addressed :heavy_check_mark:

packages/frontend-shared/src/gql-components/RecordRunModal.cy.tsx
astone123: I do have a test in `RecordPrompt.cy.tsx` that I think covers this case. If we fail to retrieve a record key then we just put `<record-key>` in the command
```
  it('renders with no props', () => {
    cy.mount(<RecordPromptVue />)

    cy.findByText('npx cypress run --record --key <record-key>').should('be.visible')
  })
```

packages/app/src/runs/RunsEmpty.cy.tsx
astone123: @lmiller1990 same, I wonder if there's a way that we could use the i18n Ally plugin to show the actual messages in the test code like it does for Vue components 🤔

mike-plummer (approved):

PR #23890, opened 2022-09-20
fix(npm/vue): update types

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/23653

### User facing changelog

Update type definition for `mount` function exported from `cypress/vue`.

### Steps to test

I could not reproduce this type error in the main Cypress repo. It might be possible via system test, but this seems really overkill. I tested by following the reproduction in the issue.

## Prep

1. checkout this branch
2. yarn install
3. in npm/vue run yarn build

## Reproduction

Now grab the user's reproduction: 

- clone user's repo https://github.com/JoostKersjes/cypress-test-tiny
- check out his branch: vue-3.2.38-type-error
- install
- `npm run check` - observe error

## Verify Fix

- manually removing `rm -rf node_modules/cypress/vue` 
- replace with the new build - eg `cp -r ~/path/to/your/monorepo/cli/vue ~/path/to/reproduction/node_modules/cypress/vue`
- run `npm run check` again - no error

I'd really like to figure out how better to test these types, but I don't know a good solution right now. Vue types are very painful, there's way too many ways to define a component.

### How has the user experience changed?

Better and more correct type inference. 

### PR Tasks

- [ ] Have tests been added/updated?
- [x] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [ ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [ ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?

PR #23954, opened 2022-09-23
fix: fix regression in npm/vue

lmiller1990:
I introduced a regression in npm/vue in this PR https://github.com/cypress-io/cypress/pull/23890. I removed the default value, `{}` for `options`: https://github.com/cypress-io/cypress/pull/23890/files#diff-509a4e791955082b305c1c45c80f14ebc70cd98ad4c3eaec78b5576e6179c9c0L241.

Amazingly we had no tests that did `cy.mount(...)` without passing any mounting options, so we didn't catch this regression. I fixed the regression and added a test.


packages/graphql/schemas/schema.graphql
```diff
@@ -722,6 +722,7 @@ enum ErrorTypeEnum {
   BAD_POLICY_WARNING_TOOLTIP
   BROWSER_NOT_FOUND_BY_NAME
   BROWSER_NOT_FOUND_BY_PATH
+  BROWSER_UNSUPPORTED_LAUNCH_OPTION
```
lmiller1990, line 725: This error got added in a different PR, but the schema wasn't committed, so I just added it here. We should probably not commit generated files anyway, I think we can probably `gitignore` this file in a future PR.

mike-plummer: Committing the schema I think allows us to work offline or if the dashboard happens to be down which are nice guards to have in place. I wonder if there's a bot that could check and auto-PR schema changes once a day or whatever?

mike-plummer (approved):

rockindahizzy (approved):

PR #23956, opened 2022-09-23
fix: prevent window from being rewritten in set/getLocalStorage

### LINKED ISSUE #22986: `window` is Being Rewritten By V8 Snapshot in `routes-e2e` for getLocalStorage

mschile:
- Closes #22986

### User facing changelog

n/a

### Additional details

`window` was being incorrectly rewritten in `getLocalStorage` and `setLocalStorage` in `routes-e2e.ts` since the javascript was inline with the html. In order to resolve the issue, I moved the inline html to their own files and returned those from the routes instead. No functional changes were made to the `html` or `js`, it was just moved from `routes-e2e.ts` to its own file.

### Steps to test

n/a

n/a

### PR Tasks


packages/server/lib/html/get-local-storage.html
```diff
@@ -0,0 +1,27 @@
+<html>
```
emilyrohrbough, line 1: Any way we can pull the server changes into develop now?

ryanthemanuel (approved):

packages/server/lib/html/get-local-storage.html
mschile: We could, though I'm not sure we gain much since this isn't actually fixing an issue in develop and there would be a slight disconnect from the issue since it deals with snapshots. 🤷

emilyrohrbough: It would be less to review when v8-snapshots is merged into develop

mschile: If the PR is changed to develop, we wouldn't be able to verify the change is actually correct since snapshots aren't used in develop.

emilyrohrbough: wouldn't you be able to verify once the latest develop changes are merged in?

mschile: Sure, but that is after the fact. You would be trusting me that this fix is correct prior to actually being able to verify the change. Maybe that is okay since we know html files aren't snapshotted but I don't think there's much to gain going to develop first.

emilyrohrbough (approved):

PR #23957, opened 2022-09-23
fix: do not double handle errors in component testing

### LINKED ISSUE #23920: CT Stacktraces: Duplicate reporting of uncaught errors
Component tests improperly report the same uncaught error multiple times - likely caused by differences in architecture between e2e and ct which allows multiple error listeners to process the uncaught error. Logic changes were made to some tests to account for this during CT Stacktrace work (see `.should(mode === 'component' ? 'have.length.gte' : 'have.length', 1)` line of `verify-failures.ts`)

To reproduce, create a component that throws an error from a click listener then invoke that listener via a component test.

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/23920

### User facing changelog

Fix a bug where unhandled exceptions and errors would be logged twice in Component Testing.

### Additional details

Unlike End To End Testing which has two iframes:
- Spec Frame, for the spec.
- AUT Frame, for the user's application.

Component Testing only has one iframe:
- AUT Frame

AUT Frame is _also_ the Spec Frame, since we serve the specs from a dev server - they are bundled as a single file.
We don't want to bind two error handlers, or we end up logging errors twice. For this reason, we conditionally add these event listeners here - Component Testing errors are captured and logged in `contentWindowListeners#bindToListeners` in `cypress/cy.ts`.

### Steps to test

See steps to reproduce in https://github.com/cypress-io/cypress/issues/23920. Make a component that errors when you clic a button, you can see that before we'd log the same error multiple times - now we do not.

Also, you could copy+paste my regression test (added in driver) into develop and watch it fail.

### PR Tasks


system-tests/__snapshots__/vite_dev_server_fresh_spec.ts.js
```diff
@@ -391,11 +395,15 @@ We dynamically generated a new test to display this failure.
   1 failing
 
   1) is missing React:
-     ReferenceError: The following error originated from your test code, not from Cypress.
+     ReferenceError: The following error originated from your application code, not from Cypress.
```
lmiller1990, line 588: Weird one - I think application is more accurate here. In this example the user wrote a JSX file but didn't do `import React from 'react'` - I'd say that's their application code.

For CT, application/spec is really a blurred line - it's one bundle and one iframe, so it's kind of the same thing.

ZachJW34: Yeah a bit weird, I added a `setTimeout(() => throw new Error('wtf))` inside an `it` and the error is the same (type=application). Not sure of a way to differentiate between app and test code when it comes to CT unless you looked at the error stack.

I would be in favor of it always saying `test` code for CT

packages/driver/cypress.config.ts
```diff
@@ -20,4 +21,14 @@ export default defineConfig({
     },
     baseUrl: 'http://localhost:3500',
   },
+  component: {
+    specPattern: 'cypress/component/**/*.cy.js',
+    supportFile: false,
+    devServer: (devServerOptions) => {
```
ZachJW34, line 27: Any reason to use the function over the object API?

mike-plummer (approved):
LGTM 🚀

ZachJW34 (approved):

system-tests/__snapshots__/vite_dev_server_fresh_spec.ts.js
lmiller1990: We could do this, we'd need to move this conditional error listener. It's hard to say if it's your component code or your test code, since there is only one frame in CT - in E2E, it's much more clear, since there's the AUT and Spec Frames.

One option would be to make more generic for CT - something like "The following error originated from your code"?

packages/driver/cypress.config.ts
lmiller1990: This package (driver) doesn't have any component tests (there aren't any components to be tested). 

If you use the object API, you need to specify `framework` - and we don't have one here.

Unless it's possible to just use `bundler` without `framework`? I don't think we support this, do we?

system-tests/__snapshots__/vite_dev_server_fresh_spec.ts.js
lmiller1990: Actually, I thought about it more - this is probably worth spending a bit of time on, it's quite a foundational part of Cypress. I am not sure if there is a good way to know if it's spec or app error, though. I'll take a look.

lmiller1990: Ok, something like

1. Get stack trace
2. Get origin of error
3. If spec file -> spec error
4. else -> App error

This means we need to see if the file in the stack traces matches the specPattern glob. This is tricky.

lmiller1990:
Moving this back to in progress - there's a few more things we need to do, eg - verify the right error message ("error is from your test code" vs "error is from your application code", etc).

system-tests/__snapshots__/vite_dev_server_fresh_spec.ts.js
ZachJW34: What happens if the error occurs on the support file? Like you mentioned earlier, I'm not sure it makes sense to make the distinction between app and test code for CT. You could have a cypress utils file that would be considered "App" code, or a Component Harness only used for CT (but since the code doesn't come from a `cy.{js,ts}` file it is considered "App" code).

The distinction between test code and app code for e2e is perfectly defined, not so much with CT so I'm not sure it's worth spending too much time on this if the distinction can't be made.

packages/driver/cypress.config.ts
ZachJW34: The types might not be happy but you should be able to do:
```js
devServer: {
  bundler: 'webpack',
  framework: undefined
} as any
```

system-tests/__snapshots__/vite_dev_server_fresh_spec.ts.js
lmiller1990: Yeah I agree, maybe I'll spend 1 more hour or so. Basic heuristic could be

- spec file OR support file: test code
- anything else: application code

If this is too much work, I'll just pick one and go with it - either way, the stack trace shows, which is the main thing.

Either way, while working on this, I found a few more edge case, going to add some more tests.

packages/driver/src/cypress/error_utils.ts
```diff
@@ -384,16 +385,28 @@ const errByPath = (msgPath, args?) => {
   })
 }
 
-const createUncaughtException = ({ frameType, handlerType, state, err }) => {
+type FrameType = 'spec' | 'app'
+
+interface UncaughtException {
```
lmiller1990: Slowly typing things!

system-tests/__snapshots__/vite_dev_server_fresh_spec.ts.js
lmiller1990: @ZachJW34 I thought about this more. I compiled a list of every error and how the UI looks after this change: https://www.notion.so/cypressdx/Errors-adf3361e7c174c388404142148a865a5

WDYT? "Application Error" seems more accurate that "Spec Error" for the majority of cases.

lmiller1990:
@mike-plummer @ZachJW34 back in review. I dismissed your reviews - please take another look (code is mostly the same, just added some tests and cleaned some things up).

I added more tests to ensure we don't double log errors, building on top of @mike-plummer stack trace tests. 

```js
verify('sync error', {
      fileName: 'Errors.vue',
      line: 24,
      column: 16,
      uncaught: true,
      uncaughtMessage: 'sync error',
      message: [
        'The following error originated from your application code',
        'sync error',
      ],
      codeFrameText: 'Errors.vue',
    }).then(() => {
        verifyErrorOnlyCapturedOnce('Error: sync error') // <==== this is new 
    })
})
```

Created a doc showing all the different cases: https://www.notion.so/cypressdx/Errors-adf3361e7c174c388404142148a865a5

There is also one outstanding bug: https://github.com/cypress-io/cypress/pull/23957/files#diff-0848e3c3d5abbc06e692c0d0ace10635a59161668729e1915c4042d227a65a27R94

ReactDOM seems to double-throw errors for some reason - only React component that error still double-register. I think we need to look into ReactDOM - it seems to do something different other libraries. I created a follow up issue: https://github.com/cypress-io/cypress/issues/24011

ZachJW34 (approved):
Thanks for putting together those error conditions @lmiller1990, and the additional tests are great! I'm good with the behavior

mike-plummer (approved):
LGTM except for one small cleanup task - are we able to remove this special case now that the duplicates aren't being logged? https://github.com/cypress-io/cypress/pull/23957/files#diff-29381ef9f5aca1b836d8ff0879ad99a5659a320230b2d74ec5665227de062b22L156 (`verify-failures.ts`, line 155-156)

I will approve assuming you respond/address the above comment

lmiller1990:
@mike-plummer done, will merge once CI is green. ✅

PR #23968, opened 2022-09-23
fix: Let cy.type accept number as well

simenbrekken-visma:
The types for `cy.type` only allow `text` to be of type `string` but it is always cast to a string anyway: 

https://github.dev/cypress-io/cypress/blob/238ead3c53399c022f7c99236ccc709768395a61/packages/driver/src/cy/commands/actions/type.ts#L165

### User facing changelog
Update types for `cy.type` to allow for `number` as well as `string`

### PR Tasks


mjhenkes (approved):

astone123 (approved):

emilyrohrbough (changes requested):
@simenbrekken-visma We do however check on the input that throws if the [input is not a string or is a number](https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cy/commands/actions/type.ts#L127). See this test: https://github.com/cypress-io/cypress/blob/develop/packages/driver/cypress/e2e/commands/actions/type_errors.cy.js#L325-L334

astone123:
@emilyrohrbough yeah it looks like we do some validation when the input is of type `date` to make sure that a `YYYY-MM-DD` string is passed in, not a number or an invalid string.

I'm starting to think that we don't want to make this change.

astone123:
@simenbrekken-visma After seeing what @emilyrohrbough brought up, I'm going to close this pull request. If people want to pass numbers into `type`, they'll have to cast the number to a string beforehand. Thanks!

simenbrekken-visma:
Fair enough, as a solution for anyone that wants to make this easier (e.g for QAs writing tests), another solution is to overwrite `cy.type`.

PR #23973, opened 2022-09-26
chore: Migrate react Highlight component to Vue

sainthkh:
- Part 2 of #23187

### User facing changelog

N/A

### Additional details
- Why was this change necessary? => To remove `pkg/runner`.

* I rewrote Highlight react component in Vue.
* As `floating-vue` uses `floating-ui` 0.1.0, it doesn't support shadow DOM properly. So, I used `floating-ui` directly to implement the tooltip. -> With @lmiller1990's review, we don't need `floating-ui`. I removed it.
* I used Vue custom elements to use CSS values directly. But the problem is that it doesn't render CSS to the live code. Maybe we need to remove them. -> I removed component tests and moved the placement test to e2e.

### Steps to test

Run Selector Playground and see Highlights working correctly.

### PR Tasks


packages/app/src/runner/selector-playground/Highlight.ce.vue
```diff
@@ -0,0 +1,80 @@
...
+
+defineProps <{
+  selector: string
+  style: StyleValue
+}>()
+
+const highlightRef: Ref<HTMLElement | null> = ref(null)
+const tooltipRef: Ref<HTMLElement | null> = ref(null)
+
+onMounted(() => {
```
lmiller1990: Both the React and Vue version of this  manipulate the DOM so heavily... we need `ref` and `nextTick` a lot, seems like a code smell.

Do you think we should just do this without a UI lib and just use regular JS/TS?

lmiller1990: WDYT @sainthkh

sainthkh: I tried it and did it. It became much simpler!

sainthkh: I'm sorry. I wasn't sure if it was possible. So, I tested it first and learned that it is doable. As our tooltip is simple, we don't need a heavy tool like `floating-ui`. 

I'm not sure we also don't need it for `studio` UIs. But it's not necessary for `Highlight`.

lmiller1990: Neat - I had a feeling the UI lib (react, vue) were both making this more complex than necessary.

packages/app/src/runner/selector-playground/Highlight.ce.vue
```diff
@@ -0,0 +1,81 @@
+<template>
```
lmiller1990, line 1: What is the naming convention here? Why is this `.ce.vue`?

packages/app/src/runner/selector-playground/Highlight.ce.vue
```diff
@@ -0,0 +1,81 @@
...
+      transform: 'rotate(0deg)',
+    }
+    : {
+      left: `8px`,
+      top: `24px`,
+      transform: 'rotate(180deg)',
+    }
+</script>
+
+<style>
```
lmiller1990, line 57: I don't think this is actually needed here - it looks like you grab it via `Highlight.styles` elsewhere. I tried deleting this entire block and just doing `css: `.......` [here](https://github.com/cypress-io/cypress/pull/23973/files#diff-533ffcf1ce12c3f364ebb004a571fce7c2e34670aa16e6b7ac36a1a5b2cd45b9R764) and it still works - thoughts? How about move this out of there into that file (just injecting a string).

lmiller1990 (commented):
Quick few comments.

I'd almost consider just writing plain JS for this... it feels a little weird to remount a new Vue app every time we move the mouse just to do some highlighting, probably not the most performant thing to do, especially considering how simple the code is. but this is probably fine too, since it's only for open mode.

lmiller1990:
Couple of questions

packages/app/src/runner/selector-playground/Highlight.ce.vue
sainthkh: ce stands for custom elements. 

Without this we cannot get the css string with `import` syntax because they're injected to <style> tag. 


It's for shadow dom and  web components.

https://vuejs.org/guide/extras/web-components.html#sfc-as-custom-element

packages/app/src/runner/selector-playground/Highlight.ce.vue
sainthkh: I placed it here because we usually write related code in one SFC file. 

They're the styles for the highlight and its related elements, I think ce.vue files are better place for them.

packages/app/src/runner/selector-playground/Highlight.ce.vue
lmiller1990: Interesting, did not know about this.

lmiller1990 (approved):
This looks good to me. I gave it a test and it's working as expected.

I did not know about the `.ce.vue` feature. Neat.

astone123 (approved):

PR #23981, opened 2022-09-26
chore: Use the same dashboard project ID for all Cypress monorepo packages

### LINKED ISSUE #23964: Use 'cypress' project for all monorepo packages on production dashboard
We have two separate projects for the Cypress monorepo in the Cypress Dashboard:

- `cypress-launchpad` - the `app`, `frontend-shared`, and `launchpad` packages report to this project
- `cypress` - all of the other packages report to this project

We should use one project for all of the packages in the Cypress monorepo to keep things consistent and avoid confusion. 

**Suggested Implementation**:
Modify the Cypress configurations of the `app`, `frontend-shared`, and `launchpad` packages to use `ypt4pf` for the `projectId` field. This is the project ID for the `cypress` project in the production dashboard environment.

Note that we will lose some historical data for the tests that run in these packages since they will stop recording in `cypress-launchpad` and start in `cypress`. This should only be a problem for a month or two once we've built up flakiness data and other run history.

astone123:
- Closes https://github.com/cypress-io/cypress/issues/23964

### User facing changelog

n/a

### Additional details

We have two separate projects for the Cypress monorepo in the Cypress Dashboard:

`cypress-launchpad` - the `app`, `frontend-shared`, and `launchpad` packages record to this project
`cypress` - all of the other packages record to this project

This PR makes it so that all of the packages in the monorepo record runs to the same project in the dashboard, namely the `cypress` project. This means that the `cypress-launchpad` project will not be updated with anymore recorded runs and will serve only as a historical project.

It will take a little while for us to build up run history in the `cypress` project for the packages that have been moved over, so for the first few days, viewing the specs list for these packages might show an empty state for "latest runs" or flakiness.

### Steps to test

1. Check out this branch
2. Open CT or e2e for the following packages: `app`, `frontend-shared`, `launchpad`
3. Verify that there is at least one "latest run" for each test in each of the packages (this tells us that CI is successfully recording runs to the new project and app is ingesting them correctly)

n/a

### PR Tasks

### DIFF
--- a/packages/app/cypress.config.ts
+++ b/packages/app/cypress.config.ts
@@ -2,11 +2,8 @@ import { defineConfig } from 'cypress'
 import getenv from 'getenv'
 import { initGitRepoForTestProject, resetGitRepoForTestProject } from './cypress/tasks/git'
 
-const CYPRESS_INTERNAL_CLOUD_ENV = getenv('CYPRESS_INTERNAL_CLOUD_ENV', process.env.CYPRESS_INTERNAL_ENV || 'development')
-const CYPRESS_INTERNAL_DEV_PROJECT_ID = getenv('CYPRESS_INTERNAL_DEV_PROJECT_ID', process.env.CYPRESS_INTERNAL_DEV_PROJECT_ID || 'sehy69')
-
 export default defineConfig({
-  projectId: CYPRESS_INTERNAL_CLOUD_ENV === 'staging' ? 'ypt4pf' : CYPRESS_INTERNAL_DEV_PROJECT_ID,
+  projectId: getenv('CYPRESS_INTERNAL_DEV_PROJECT_ID', process.env.CYPRESS_INTERNAL_DEV_PROJECT_ID || 'ypt4pf'),
   retries: {
     runMode: 2,
     openMode: 0,
--- a/packages/app/cypress.config.ts
+++ b/packages/app/cypress.config.ts
@@ -1,9 +1,8 @@
 import { defineConfig } from 'cypress'
-import getenv from 'getenv'
 import { initGitRepoForTestProject, resetGitRepoForTestProject } from './cypress/tasks/git'
 
 export default defineConfig({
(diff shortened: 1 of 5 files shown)


astone123:
If anyone has admin access to the project, we can delete the `TEST_LAUNCHPAD_RECORD_KEY` variable once this goes in

chrisbreiding (approved):

packages/app/cypress.config.ts
```diff
@@ -2,11 +2,8 @@ import { defineConfig } from 'cypress'
 import getenv from 'getenv'
 import { initGitRepoForTestProject, resetGitRepoForTestProject } from './cypress/tasks/git'
 
-const CYPRESS_INTERNAL_CLOUD_ENV = getenv('CYPRESS_INTERNAL_CLOUD_ENV', process.env.CYPRESS_INTERNAL_ENV || 'development')
-const CYPRESS_INTERNAL_DEV_PROJECT_ID = getenv('CYPRESS_INTERNAL_DEV_PROJECT_ID', process.env.CYPRESS_INTERNAL_DEV_PROJECT_ID || 'sehy69')
-
 export default defineConfig({
-  projectId: CYPRESS_INTERNAL_CLOUD_ENV === 'staging' ? 'ypt4pf' : CYPRESS_INTERNAL_DEV_PROJECT_ID,
+  projectId: getenv('CYPRESS_INTERNAL_DEV_PROJECT_ID', process.env.CYPRESS_INTERNAL_DEV_PROJECT_ID || 'ypt4pf'),
```
ZachJW34: Do we still need the `getenv`? Why do most of our packages just have `projectId: 'ypt4pf'` but our frontend packages use all of the env variables.

flotwig (approved):

packages/app/cypress.config.ts
astone123: Not sure why... I left these in because I figure we should be able to override the project IDs with an environment variable if we need to, but yeah it looks like only these packages use this pattern

I think the original reason was because these packages used a different project ID depending on whether we were in staging or production. Now it will always be the same, so I think I'll remove these

astone123: @ZachJW34 I removed `getenv` and just hard-coded the values for projectId like we do in the other packages

ZachJW34 (approved):

PR #23984, opened 2022-09-26
fix: correctly fail session when setup func fails

### LINKED ISSUE #21399: [sessions] Collapse Behavior in Command Log
### What would you like?

- When should it collapse? What should collapse? Etc.
- Validate and/or correct all collapse behavior
- Behavior and verbiage for all of: beginning operation, running command within operation, final state once it’s done

### Why is this needed?

_No response_

### Other

_No response_

emilyrohrbough:
- Closes #21399

### User facing changelog
Correctly fail `cy.session()` command when the setup function has a failing command or assertion. Fixes [#21399](https://github.com/cypress-io/cypress/issues/21399).

Before | After
--|--
Screen Shot 2022-09-26 at 11 56 22 AM | Screen Shot 2022-09-26 at 11 55 50 AM

### PR Tasks


system-tests/projects/session-and-origin-e2e-specs/cypress/e2e/session/errors.cy.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Used in cy-in-cy tests in @packages/app.
+ */
+it('setup has failing command', () => {
+  cy.session('session_1', () => {
+    cy.get('does_not_exist', { timeout: 500 })
```
chrisbreiding:
```suggestion
    cy.get('does_not_exist', { timeout: 1 })
```

Any reason not to have it fail faster?

emilyrohrbough: good idea

chrisbreiding (approved):

packages/driver/src/cy/commands/sessions/index.ts
```diff
@@ -170,11 +170,23 @@ export default function (Commands, Cypress, cy) {
           displayName: recreateSession ? 'Recreate session' : 'Create new session',
           message: '',
           type: 'system',
-        }, () => {
+        }, (setupLogGroup) => {
           return cy.then(async () => {
+            // Catch when a cypress command fails in the setup function to correctly update log status
+            // before failing command and ending command queue.
+            cy.state('onCommandFailed', (err) => {
```
mjhenkes, line 177: does 'onCommandFailed' have to live in state? It would be great if we didn't have to rely on global (to cy) state for this.

emilyrohrbough: I spent (quite a bit of) time looking into what alternative options we could implement here. Unfortunately, with the current command queue structure, I think this is the quickest win. To work around this would require some reworking on how the command queue is setup for nested-commands. (also sessions is just the weirdo command that is essentially a "workflow" what requires a series of commands before the workflow command is successful).

mjhenkes (approved):

PR #23985, opened 2022-09-26
fix: update way that cohorts are cached to remove duplicate key

warrensplayer:
- Closes 

### User facing changelog

n/a

### Additional details

Updating the way that the cohorts are stored in the `cache` file to remove a duplicate value being stored.  The original implementation had the "COHORTS" section of the cache file that would look like:
```
"COHORTS": {
    "aci_082022_login": {
      "cohort": "A",
      "name": "aci_082022_login"
    }
  }
```
This duplicated the key in the main COHORTS object into the "name" field of the value for that key.  The updated code will store the value like:
```
"COHORTS": {
    "aci_082022_login": {
      "cohort": "A"
    }
  }
```
The API for the cohorts section of the cache does not change.  The `cache.js` file encapsulates modifying the structure on the way to and from the file to keep the expected values the same.

### Steps to test

The unit test  `cache_spec.js` should verify that the internal change did not change the overall output.  The original code was implemented in this PR: https://github.com/cypress-io/cypress/pull/23735

Follow the steps in the linked PR and check that the values stored in the `cache` file match the expected structure above.

n/a

### PR Tasks

### DIFF
--- a/packages/server/lib/cache.js
+++ b/packages/server/lib/cache.js
@@ -186,7 +186,13 @@ module.exports = {
   },
 
   getCohorts () {
-    return fileUtil.get('COHORTS', {})
+    return fileUtil.get('COHORTS', {}).then((cohorts) => {
+      Object.keys(cohorts).forEach((key) => {
+        cohorts[key].name = key
+      })
+
+      return cohorts
+    })
   },
 
   insertCohort (cohort) {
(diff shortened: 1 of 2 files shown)


astone123 (approved):
Looks good 👍🏻

ZachJW34 (approved):

PR #23988, opened 2022-09-26
chore: refactor app.yml config

### LINKED ISSUE #22994: Ensure that the Data in `packages/server/config/app.yml` is Embedded in the V8 Snapshot

mschile:
- Closes #22994 

### User facing changelog

n/a

### Additional details

The current `config/app.yml` needed to be refactored due to containing many obsolete values and it wasn't included in the snapshot since yaml files do not get snapshotted.
* Converted `app.yml` to `app.json`
  * Removed obsolete fields:
    * `desktop_url`
    * `desktop_manifest_url`
    * `chromium_url`
    * `chromium_manifest_url`
    * `on_url`
  * Moved `cdn_url` field directly to `upload.js` since the value doesn't change between environments.
* Removed `konfig` module as it seemed like overkill for loading a single config value.

### Steps to test

* create a prod v8 snapshot (`DEBUG=cypress:snap*(info|debug|error) yarn build-v8-snapshot-prod`)
* Verify the `app.json` content in the snapshot
* Verify Cypress can still interact with the dashboard

n/a

### PR Tasks


tooling/v8-snapshot/cache/prod-darwin/snapshot-meta.cache.json
```diff
@@ -3652,10 +3594,12 @@
     "./packages/scaffold-config/src/dependencies.ts",
     "./packages/scaffold-config/src/frameworks.ts",
     "./packages/scaffold-config/src/supportFile.ts",
+    "./packages/server/config/app.json",
```
ryanthemanuel, line 3623: 🎉

package.json
```diff
@@ -179,7 +179,6 @@
     "js-codemod": "cpojer/js-codemod",
     "jscodemods": "https://github.com/cypress-io/jscodemods.git#01b546e",
     "jscodeshift": "0.7.0",
-    "konfig": "0.2.1",
```
emilyrohrbough, line 182: 🎉

packages/server/lib/util/routes.js
```diff
@@ -1,8 +1,8 @@
 const _ = require('lodash')
 const UrlParse = require('url-parse')
-const konfig = require('../konfig')
 
-const apiUrl = konfig('api_url')
+const app_config = require('../../config/app.json')
+const apiUrl = app_config[process.env.CYPRESS_KONFIG_ENV || process.env.CYPRESS_INTERNAL_ENV || 'development'].api_url
```
ryanthemanuel: Is CYPRESS_KONFIG_ENV exposed at all? Wondering if we should rename/get rid of it if we're no longer using konfig. Seems like it might cause more confusion than anything.

mschile: I don't see it documented anywhere for external use but we do use it to control sending our results to the dashboard when we run our tests. Thus, I am just going to rename it to `CYPRESS_CONFIG_ENV`.

mschile: renamed [ed83366](https://github.com/cypress-io/cypress/pull/23988/commits/ed833667d8ffad4e78934a4c523eb60951c4f6ec)

ryanthemanuel (approved):

emilyrohrbough (approved):
Still think these changes would be beneficial to pull into develop now. Less divergence / easier review when this is ready.

PR #24021, opened 2022-09-28
fix: Assertions properly remain pending when paired with some commands

BlueWinds:
### User facing changelog
Fixes a regression where assertions would show as "failed" while they were still retrying

### Additional details
Commands that pass `onFail` callbacks to `verifyingUpcomingAssertions` - such as cypress-testing-library's `findByTestId()` and our own `readFile()` - were showing assertions as "failed" while they were in fact still retrying. See the attached video for an example.

### Steps to test
```
cd packages/app
yarn dev
```

Open e2e testing then run `cypress/e2e/runs.cy.ts`.

`when a project is created, injects new projectId into the config file, and sends expected UTM params` is the test featured in the video, and is a good example of the behavior. `it.only()` that one.

Regression behavior:

https://user-images.githubusercontent.com/3003404/192853596-b97004e7-6c4a-47ac-9ac9-e1d6f8d62dde.mp4

With this PR, there is no longer a flash of red on each assertion - it shows grey (pending) until it resolves with either success or failure.

### PR Tasks

### DIFF
--- a/packages/driver/cypress/e2e/commands/assertions.cy.js
+++ b/packages/driver/cypress/e2e/commands/assertions.cy.js
@@ -635,7 +635,7 @@ describe('src/cy/commands/assertions', () => {
         cy.get('button:first').should('have.class', 'does-not-have-class')
       })
 
-      it('has a pending state while retrying', (done) => {
+      it('has a pending state while retrying queries', (done) => {
         cy.on('command:retry', (command) => {
           const [getLog, shouldLog] = cy.state('current').get('logs')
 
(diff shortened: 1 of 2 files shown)


packages/driver/cypress/e2e/commands/assertions.cy.js
```diff
@@ -648,6 +648,19 @@ describe('src/cy/commands/assertions', () => {
         cy.get('button:first', { timeout: 100 }).should('have.class', 'does-not-have-class')
       })
 
+      it('has a pending state while retrying for commands with onFail', (done) => {
+        cy.on('command:retry', (command) => {
+          const [readFileLog, shouldLog] = cy.state('current').get('logs')
```
emilyrohrbough, line 653: Seems like this has a chance to be flakey since this state will be updated as the command resolves and eventually change to 'pass'. Did you notice this as all when running locally?

BlueWinds: Shouldn't be - because we we call `done()` inside this retry, Cypress cancels all active commands and doesn't continue retrying commands. `cy.on('command:retry', () => { ...; done() }` is a pretty common pattern in the driver tests.

emilyrohrbough (approved):

lmiller1990 (approved):

ryanthemanuel (approved):

PR #24063, opened 2022-09-30
fix: Address Vite sourcemap edge cases

### LINKED ISSUE #23919: CT Stacktraces: Investigate Vite sourcemap mismatches
Sourcemaps generated by Vite are often shifted by 1-3 columns, resulting in misaligned code frames (errors point to incorrect part of line that generated the error). This is worked around in current system tests by providing multiple column mappings to accept (see `reporter-ct-vite.errors.cy.ts`), and at runtime users will sometimes see codeframes that are incorrectly positioned.

Need to investigate whether the sourcemaps themselves are incorrect, or whether we have logic in place that is breaking them due to Webpack-specific behaviors. This typically presents in JS-only files that are not transpiled by Vite (not JSX, TS, etc), so this may indicate an issue with the workaround we've added to `vite-dev-server` to always serve sourcemaps (see `transform` function of `cypress.ts` plugin under `vite-dev-server`)

mike-plummer:
- Closes #23919 

### User facing changelog
* Fixes issues where JS files served by Vite that generate errors/failures could present code frames in the runner that point to incorrect lines of source code

### Additional details
**Context**
Vite does not transpile most _*.js_ files, thus does not serve them up with sourcemaps - this prevents our codeFrame logic from presenting applicable source in the event of a failure or error out of those files. A workaround was added in #21720 to force Vite to serve sourcemaps for those files by sending the current sourcemap along as the `map` key in the return value in our custom plugin.

**Problem**
In some scenarios (such as JS files that are processed by the Vite React plugin which uses Babel instead of esbuild) the sourcemaps didn't line up correctly and we would present incorrect lines of code. It's unclear exactly why this is happening but may be due to Vite using esbuild and React plugin using Babel and having to port sourcemap content over during that translation.

**Solution**
This PR updates our sourcemap workaround to instead inline the sourcemap content into the served code and tell Vite there isn't a map by providing a "dummy" map with no mappings. This effectively resets the sourcemap context - the map content is there in the output source if the file is served without further mutation, and can be parsed out and adjusted if needed by other plugins like React.

The sourcemap logic is moved into a separate plugin - this helps organize it a bit better and also lets us force this plugin to be a "post" phase execution so it occurs as close to the end of processing as possible.

### Steps to test
Should be covered by system tests

To manually test:
1. set up a React project with Vite (ensure `@vitejs/plugin-react` is applied).
2. Create a pure JS spec file and add tests that are guaranteed to fail:
```
describe('spec file with import', () => {
   it('should generate uncaught error', () => {
    throw new Error('uncaught')
  })
      
  it('should generate failed command', () => {
    cy.get('#does-not-exist', { timeout: 100 })
  })
})
```
4. Run the spec in the runner, verify that an accurate code frame is displayed for each test. Note that the code frame will likely be off by a column but should point to the correct line.
5. Add the following import to the spec file (will cause it to be processed by the react plugin): `import React from 'react'`
6. Re-run the spec in the runner, verify the code frame still points to the correct line
    * Without this PR the code frames would be off by several lines

### PR Tasks


ZachJW34 (approved):
Just to verify, the column is still off by one in my testing (pure js). Seems like that is expected just wanted to call it out. The line # is what is most important and verified that everything is working properly so great work!

mike-plummer:
> Just to verify, the column is still off by one in my testing (pure js). Seems like that is expected just wanted to call it out. The line # is what is most important and verified that everything is working properly so great work!

@ZachJW34 Correct, this just resolves some cases where we resolved incorrect lines. My suspicion is that the "off by one" is related to [this logic](https://github.com/cypress-io/cypress/blob/develop/packages/driver/src/cypress/stack_utils.ts#L138) but I haven't been able to determine why that adjustment is necessary for webpack/babel sourcemaps but not rollup/esbuild

astone123 (approved):
nice! 🎉

Issue #24062, opened 2022-09-30
Angular 15 polyfills API change causes component testing to fail


Coly010:
### Current behavior

Component testing will fail with Angular 15 workspaces due to a change in Angular's build and test options API for polyfills, which has changed from `string` to `Polyfills` which is a type alias for `string[] | string`.

The current error is:
```
Your configFile threw an error from: cypress.config.ts

We stopped running your tests because your config file crashed.

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received an instance of Array
```

You can see this in our Nx Cloud run for Angular 15 support branch: https://staging.nx.app/runs/2PyEc1jfqo6/task/e2e-angular-extensions:e2e

If you do a Find in Page and search for "fail" you'll see the error output.

### Desired behavior

Component testing should work correctly for Angular 15 workspaces.

### Test code to reproduce

Create an "Angular 15"* Workspace.

`npx @angular/cli@next new myapp`

Add component testing.

Try to run component tests.

> * Angular 15 is still in development so it's only next releases currently.

### Cypress Version

latest

### Node version

v16

### Operating System

All

ZachJW34:
@Coly010 are you interested in contributing a fix?

Coly010:
Yes!

astone123:
Closed with #24064

PR #24064, opened 2022-09-30
fix(webpack-dev-server): handle polyfills array for upcoming angular 15 change

Coly010:
- Closes  #24062 

### User facing changelog

n/a

### Additional details

Polyfills for Angular can now be an array of strings, and some of those can be node_modules. 
Therefore, this PR does two things to accommodate this change.

It checks if it is an array, then checks if the path starts with the source root of the project. If it does, it performs the additional path logic, if not, it leaves it as is.

It then adds the include paths of all elements of the polyfills array.

Added additional project fixture for angular 15 as it requires an ES target of 2022 for Angular 15

### Steps to test

Standard happy path for component testing but using Angular@next 

### PR Tasks


ZachJW34:
@Coly010 I pushed up a change reusing the `angular` project-fixture to reduce the number of files added, and added the new version to some other places where system-tests are run. System-tests seem happy with Angular v15!

jordanpowell88 (approved):
This looks good to me! Thanks for being proactive on this @Coly010

Coly010:
Awesome! Looks better without the duplicated fixture!

ZachJW34 (approved):
LGTM!

PR #24097, opened 2022-10-03
feat: Disable file watching in component tests in run mode

### LINKED ISSUE #22959: CT DevServer should not watch files in Run Mode
### Current behavior

The webpack and vite dev-servers, by default, watch all compilation files. This makes sense in open mode, but there is no need to do this in run-mode and can be taxing, especially when it comes to CI machines. We disable all config/spec watching in run-mode, but this philosophy doesn't extend to the CT dev-servers.

### Desired behavior

Disable file watching for CT dev-servers when using run mode.

### Test code to reproduce

1. Check out the branch my branch on [cypress-component-testing-apps](https://github.com/cypress-io/cypress-component-testing-apps/tree/zachw/watch-in-run-mode)
2. `cd ./react-cra5-ts`
3. `npm i`
4. `npx cypress run --component`
5. After you see the webpack compile stats in the terminal, edit `src/App.tsx`. You will see another webpack recompilation. I've added a long wait in the spec to allow time to tweak while the test is running.

Screen Shot 2022-07-27 at 2 42 18 PM

### Cypress Version

10.3.1

### Other

I've only tested webpack but I'm assuming vite is the same. There are some other options we should consider disabling, such as `liveReload` and `hmr` if applicable.

mike-plummer:
- Closes #22959 

### User facing changelog
* Disable file watching when executing component tests in "run" mode

### Additional details
Should have a positive impact on performance by not attempting to watch/reload any dev files in the project when running Cypress tests in `run` mode

### Steps to test
Reproduction instructions are available in #22959. When running with this branch you should not see reload/recompile events if the `App.tsx` file is touched when executing test in `run` mode, but they should still occur in `open` mode

No user-facing impact, minor performance gain

### PR Tasks


ZachJW34 (commented):
Looks good, any way to e2e test this change? I know cy-in-cy is open mode, could we use the devServer function to manipulate the config and verify? Something like:

```js
module.exports = {
  component: {
    devServer(devServerOptions) {
      devServerOptions.cypressConfig.isTextTerminal = true
      
      return cypressWebpackDevServer({ ...devServerOptions, framework: 'react' })
    }
  }
}
```

mike-plummer:
> Looks good, any way to e2e test this change? I know cy-in-cy is open mode, could we use the devServer function to manipulate the config and verify? Something like:
> 
@ZachJW34 Will look into this. Would be nice to find a way to just isolate as a `*-dev-server` test. There's an existing test in `devServer-e2e.spec.ts` that validates recompile when the list of specs changes, maybe I can adapt that setup to validate behaviors when source files change 🤔

npm/webpack-dev-server/src/createWebpackDevServer.ts
```diff
@@ -1,5 +1,6 @@
 import debugLib from 'debug'
-import type { Configuration } from 'webpack-dev-server-3'
+import type { Configuration as WebpackDevServer3Configuration } from 'webpack-dev-server-3'
+import type { Configuration as WebpackDevServer4Configuration } from 'webpack-dev-server'
```
astone123, line 3: Thanks for adding types to these config objects 💯

astone123 (approved):
Nice, have we gathered any metrics about how this might affect performance in run mode? If it's significant enough, maybe we could brag about it in a blog post or something 🎉

ZachJW34 (approved):

mike-plummer:
@astone123 No metrics to speak of at this point. I view this as more of a housekeeping task to guard against potentially mutating tests mid-run (kicking off a run and making changes in the background while waiting) or poorly-constructed tests that modify themselves or their dependencies via filesystem ops. Once it merges to `develop` I'll take a look at the CI stats, see if there's any noticeable improvement in our runs times since we're doing a lot of manual ops that most users wouldn't be attempting

PR #24101, opened 2022-10-03
feat: add logic for "cypress-triggered events"

### LINKED ISSUE #23765: [ACI post-project connection] Implement consolidated logic for “Cypress-triggered events” like displaying the CI prompt, smart banners, and post-login messages.
**Context**
This [Whimsical diagram](https://whimsical.com/get-to-first-run-TgixRWciqE1chPGDCHk5b2)(only visible to internal Cy employees) gathers together the logic that has evolved to manage smart banners, the old CI prompt, LoginConnect modal state, as well as the state of various buttons in the Top Nav, Specs List Page, Settings Page that allow users to log in and/or connect a project to he Dashboard.

The diagram represents User-Triggered Actions with solid arrows between states, and Cypress-Triggered Actions with dotted arrows. Cypress-triggered actions like displaying a notification are intended to only happen when certain specific rules have been met, and also shouldn't happen close together, so as to avoid becoming annoying, and increase the chance that they are useful.

Currently the logic around these things has been implemented similarly in more than one place, and it's time to consolidate all of it so that we can easily add new features that depend on this logic.

**Requested Change**

A solution is needed that gathers all of the logic from that Whimsical into a single system whose overall purpose is to

1. determine if Cypress has permission to do something, and 
2. provide a reactive single source of truth to define which of the small number of possible states a user is currently in. 

For example:

SpecsListBanners.vue currently uses a watcher to set various flags as the user logs in and a project gets setup, so it can determine which banner variant should be shown. It includes logic for "whether the user state is correct" (has a project? etc) to show a certain banner, and "whether Cypress is allowed" to show the banner (has more than 4 days of Cypress use?). Using the new solution, when SpecsListBanner mounts it would check which of the N possible states a user is in (which is provided from a store or similar) and if that state maps to a variant of the banner, it would call a function that checks if it has permission to open. After opening and after being dismissed, it would make the appropriate calls to this helper to record the updated date and keep everything in sync.

The "what state is the user in" question is partially solved by the `login-connect-store` introduced in #23762 and the related `CloudViewerAndProject` component that provides state via a scoped slot. But these are essentially stepping stones to a more holistic system. That branch has not been merged. It could be a starting point for this work, or just a reference point to start fresh.

The "whether Cypress is allowed" to do something question is not centrally managed at the moment. It is about to get a little more complex as now there is a 1-day delay being introduced.

**Rationale for doing this**
One reason to gather this logic centrally is to have a smoother developer experience using these kinds of values and figuring out if we can display something to the user in an increasingly complicated set of circumstances. It should be clear how to quickly "add a button" anywhere and tell it how to behave in the context of the LoginConnect state, without having to recreate the logic to derive that state. This should make us more nimble with future requests.

Another developer reason to extract this stuff is that it has nothing to do with Vue, this business logic can and should be unit tested on its own and independent of a UI framework

There's a third reason to pause and do this now: to get aligned with Product in terms of how we are naming the states users can be in, and how we define the rules for Cypress-triggered actions. The less translation we need when we talk about these things and implementing them, the better. Chris France is already using state charts to ferret out all the allowed states of these flows for figma and prototype reasons. While I'm not suggesting we get a full state machine pattern going here, we should be thinking in that spirit, and especially we should go out of our way to reject impossible states, which might lead to showing the banner and the CI prompt at the same time, or something like that. I'd expect long term this kind of thing might evolve towards a state machine.

For now a clear and human readable expression of the rules, plus some validation, is likely good enough (see suggested implementation below).

**Acceptance Criteria**

- user state like `const isProjectConnected = !!currentProject?.projectId && currentProject.cloudProject?.__typename === 'CloudProject'` is computed in one place and accessible to components that need it
- cypress-triggered actions check if they have permission to happen, and record when they have happened
- business logic related to the states and transitions in Whimsical is encapsulated in plain TS helpers that are unit tested, and the unit tests describe all of the expected scenarios
- logic from the Whimsical (like a 1-day delay) is implemented in the helper code and tested, but not wired up in the app yet, let's merge the helper, then do the wiring in #23768, #23767, and #23766.

**Suggested implementation** (for estimation purposes)

This could be done in many ways, I don't think this ticket needs to be too prescriptive as to the implementation. Though it could be broken down further into multiple steps. 

I do have an pseudocode example of how the code that manages this in a common language with Product might look, which would let us easily change rules or add new rules. This of course implies a set of functions and validators that aren't shown that would actually parse objects like this and apply the rules.

<details>
<summary>
Psuedocode for how we might manage the rules in a human readable way
</summary>

```js
const userStates = {  
// this names the 4 states defined in whimsical and expresses the rules in a human readable way
// a validator function should be used, and throw if 2 states have the same set of conditions
// we can only be in one of these at a time, ever
  loggedOut: {
    matches: ['notLoggedIn']
  },
  needsOrg: {
    matches: ['isLoggedIn', 'hasNoOrgs']
   }, 
  needsConnectedProject: {
    matches: [
      'isLoggedIn', 
      'hasAtLeastOneOrg', 
      'hasNoConnectedProject'
    ]
   }, 
  needsRecordedRun : {
    matches: [
      'isLoggedIn', 
      'hasAtLeastOneOrg',
      'hasConnectedProject', 
      'hasNoRecordedRuns', 
      minSpecsInCurrentTestingType(4)
    ]
   }  
}

const features = {
  smartBanner: { // feature - where we are in the app
    cypressTriggerRules: [ // conditions that must be true for Cypress to trigger this feature to open
      minDaysSince('firstOpen', 4), 
      minDaysSince('docsAutoOpen', 1),
      minDaysSince('connectProject', 1),
      minDaysSince('recordRunShown', 1),
      isInState(['loggedOut', 'needsOrg', 'needsConnectedProject', 'needsRecordedRun'])
      // needs a validator that the rules do not conflict
      // or can nest rules in paths to prevent impossible states
    ],
    content: { // content - customized per cohort - can live in existing i18n json
      loggedOut: { // user state - derived from gql
          a: 'Log in',
          b: 'Log in, please',
          default: 'Log in' // no cohort
      },
      needsOrg: {
          a: 'Do an org now',
          b: 'Make an org pls',
          default: 'Make org'
      }
    }
  },
  loginConnectModal: {
    // ...
  }
}

```
</details>

marktnoonan:
- Closes #23765

### User facing changelog

No user facing changes.

### Additional details

This PR adds a utility function and store to support it, which between them gather together the logic needed to drive features related to the App and Cloud Integration. Much of the work was originally done on the branch for #23762, which will merge _after_ this. We've split this out into a PR that closes #23765 to make it easier to review. That other draft PR is a good place to observe the impact of using this function, as it implements the Smart Banner and Login flows.

 A good place to look at the improvement in DX is the watcher in SpecsListBanners.vue on that branch, that manages the open/closed banner state. Not included in this PR, but as a reference: Before:

  ```ts
 watch(
  cloudData,
  ([cloudViewer, cachedUser, currentProject]) => {
    // Cached user covers state where we're authenticated but data isn't loaded yet
    const isLoggedIn = !!cachedUser?.id || !!cloudViewer?.id
    // Need to be able to tell whether the lack of `firstOrganization` means they don't have an org or whether it just hasn't loaded yet
    // Not having this check can cause a brief flicker of the 'Create Org' banner while org data is loading
    const isOrganizationLoaded = !!cloudViewer?.firstOrganization
    const isMemberOfOrganization = (cloudViewer?.firstOrganization?.nodes?.length ?? 0) > 0
    const isProjectConnected = !!currentProject?.projectId && currentProject.cloudProject?.__typename === 'CloudProject'
    const hasNoRecordedRuns = currentProject?.cloudProject?.__typename === 'CloudProject' && (currentProject.cloudProject?.runs?.nodes?.length ?? 0) === 0
    const hasFourDaysOfCypressUse = (Date.now() - currentProject?.savedState?.firstOpened) > interval('4 days')
    
    showRecordBanner.value = !hasBannerBeenDismissed(BannerIds.ACI_082022_RECORD) && isLoggedIn && isProjectConnected && isMemberOfOrganization && isProjectConnected && hasNoRecordedRuns && hasFourDaysOfCypressUse
    showConnectBanner.value = !hasBannerBeenDismissed(BannerIds.ACI_082022_CONNECT_PROJECT) && isLoggedIn && isMemberOfOrganization && !isProjectConnected && hasFourDaysOfCypressUse
    showCreateOrganizationBanner.value = !hasBannerBeenDismissed(BannerIds.ACI_082022_CREATE_ORG) && isLoggedIn && isOrganizationLoaded && !isMemberOfOrganization && hasFourDaysOfCypressUse
    showLoginBanner.value = !hasBannerBeenDismissed(BannerIds.ACI_082022_LOGIN) && !isLoggedIn && hasFourDaysOfCypressUse
    
    hasRecordBannerBeenShown.value = hasBannerBeenShown(BannerIds.ACI_082022_RECORD)
    hasConnectBannerBeenShown.value = hasBannerBeenShown(BannerIds.ACI_082022_CONNECT_PROJECT)
    hasCreateOrganizationBannerBeenShown.value = hasBannerBeenShown(BannerIds.ACI_082022_CREATE_ORG)
    hasLoginBannerBeenShown.value = hasBannerBeenShown(BannerIds.ACI_082022_LOGIN)
  },
  { immediate: true },)
```

After: 

```ts
watch(cloudData, () => {
  if (!isAllowedFeature('specsListBanner')) {
    return
  }

  showRecordBanner.value = userStatusMatches('needsRecordedRun')
  showConnectBanner.value = userStatusMatches('needsProjectConnect')
  showCreateOrganizationBanner.value = userStatusMatches('needsOrgConnect')
  showLoginBanner.value = userStatusMatches('isLoggedOut')

  hasRecordBannerBeenShown.value = hasBannerBeenShown(BannerIds.ACI_082022_RECORD)
  hasConnectBannerBeenShown.value = hasBannerBeenShown(BannerIds.ACI_082022_CONNECT_PROJECT)
  hasCreateOrganizationBannerBeenShown.value = hasBannerBeenShown(BannerIds.ACI_082022_CREATE_ORG)
  hasLoginBannerBeenShown.value = hasBannerBeenShown(BannerIds.ACI_082022_LOGIN)
},
{ immediate: true })
```

The end result is we can now ask a helper function "Is the X allowed?" and have that helper provide the answer based on the user (and project's) progression through the "login -> connect -> record" journey (defined in the store's `userStatus` getter) and the other various constraints around timing that are defined in the rules for that feature as part of `isAllowedFeature.ts`.

On that implementation branch (#23762), the store is populated through a component that watches graphql for updates to the required field. The reason to forward those values into a client side store is to make testing a little more straightforward and allow components throughout both App and Launchpad to access these values as needed without adding fields to graphql. The store in this PR also includes some values like `isLoginConnectOpen` that support new modal behavior not specifically addressed in this PR that that are used in that branch.

There is also a little extra abstraction going on in the `rules` object of `isAllowedFeature`. The goal is to be able to clearly see in that section what are the rules under which Cypress will, say, show a banner, for a given user status. Between these rules and the userStatus getter in the store, we have two places where all of the lines and notes in this whimsical are managed: https://whimsical.com/get-to-first-run-TgixRWciqE1chPGDCHk5b2 - meaning understanding and maintaining these rules, and driving different features off of this core set of data, should require less work in future. 

### Steps to test

Open the unit test for the helper: `packages/frontend-shared/src/utils/isAllowedFeature.cy.ts`. Run the spec and see that all combinations are passing.

In `isAllowedFeature.ts`, change some rules. Setting something like `minTimeSinceEvent(events.cypressFirstOpened, '40 days')` on a feature should cause some related tests to fail based on the starting value we set of "5 days before now" for `firstOpened`. 

### Feedback sought

I'm especially curious if this is considered readable by others, and any other feedback is welcome too of course.

### PR Tasks


packages/frontend-shared/src/utils/isAllowedFeature.cy.ts
```diff
@@ -0,0 +1,219 @@
...
+
+    // set a default valid number of days since first open & nav prompt shown
+    // individual tests may override
+    setFirstOpened(Date.now() - interval('5 days'))
+    setPromptShown('ci1', Date.now() - interval('5 days'))
+
+    switch (status) {
+      case 'isLoggedOut':
+        expect(store.userStatus).to.eq('isLoggedOut')
+        break
```
warrensplayer: Explicitly set `isLoggeIn` to false here. The store defaults  `isLoggedIn` to false, but if someone changes a test to call `setUpStatus` to an option that sets it to true first and then calls `setUpStatus` and passes in `isLoggedOut` then it will not have the correct values set in the store.

packages/frontend-shared/src/utils/isAllowedFeature.cy.ts
```diff
@@ -0,0 +1,219 @@
...
+          }
+        })
+      })
+
+      context('banners HAVE been dismissed', () => {
+        userStatuses.forEach((status) => {
+          if (status === 'allTasksCompleted') {
+            // no banner matches this state, so nothing can be dismissed, skip it
+            return
+          }
```
warrensplayer: Nit-pick: This is the only time that a status is skipped in any of the tests.  It caught my attention when looking at the results in Cypress where I noticed that status was not present in this context.  I thought at first that a test case was missing.  When I looked at the test code here, I saw that it was being skipped on purpose.  In real world usage, `isAllowedFeature` will always return `false` for  for a status of `allTasksCompleted` for the `specsListBanner` feature.  Should this be expressed here some how?

packages/frontend-shared/src/utils/isAllowedFeature.ts
```diff
@@ -0,0 +1,104 @@
...
+  if (isNaN(waitTimestamp)) {
+    throw new Error(`incorrect format for waitTime provided, value must be \`n days\`, \`n minutes\` etc. Value received was ${waitTime}`)
+  }
+
+  return (Date.now() - eventTime) > interval(waitTime)
+}
+
+export const isAllowedFeature = (
+  featureName: 'specsListBanner' | 'docsCiPrompt',
+  loginConnectStore: LoginConnectStore,
```
warrensplayer, line 34: What advantage does passing the store in have over the encapsulation of just referencing the store inside this function?  For the tests purposes, it can simply use the store to set the values like the implementation will do and then this function can pull in the store to read from it.

packages/frontend-shared/src/store/login-connect-store.ts
```diff
@@ -0,0 +1,134 @@
...
+export type LoginConnectBooleanField = 'isLoggedIn' | 'isProjectConnected' | 'isConfigLoaded' | 'isOrganizationLoaded' | 'isMemberOfOrganization' | 'hasRecordedRuns' | 'hasNoRecordedRuns' | 'hasNonExampleSpec'
+
+export const useLoginConnectStore = defineStore({
+  id: 'loginConnect',
+
+  state (): LoginConnectState {
+    return {
+      isLoginConnectOpen: false,
+      utmMedium: '',
+      isLoggedIn: false,
```
warrensplayer: Should we pull this login modal state out into its own store and rename this store to just `connect-store`?   Looking at it more, it seems like we are combining two different needs into this one store.

packages/frontend-shared/src/store/login-connect-store.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+  },
+  actions: {
+    openLoginConnectModal ({ utmMedium }: { utmMedium: string }) {
+      this.isLoginConnectOpen = true
+      this.utmMedium = utmMedium
+    },
+    closeLoginConnectModal () {
+      this.isLoginConnectOpen = false
+      this.utmMedium = ''
```
warrensplayer, line 80: If we do pull the login modal stuff out, then these actions would go with it along with the `setLoginError` below.

warrensplayer (changes requested):
This was a great idea to pull this code out into its own PR.  It definitely helped me focus more on the logic here.  After looking through it another time, I have a few comments/suggestions/questions.

packages/frontend-shared/src/utils/isAllowedFeature.cy.ts
marktnoonan: Hmm. I think it would be more clear if named this test `returns false for banner X`, then it would be more clear that we are intending to cycle through banners, not statuses as such. I could make it clearer in the test code as well - iterate over the banner IDs, then we'd avoid a `skip`

packages/frontend-shared/src/utils/isAllowedFeature.ts
marktnoonan: I'm not sure this is all that much, but it feels slightly less coupled to the idea of "a store" this way. It's convenient for these tests to pass the real store since there's code in there we want to test, but all that's needed is really an object with the correct properties. I think for practical purposes it's just a preference.

packages/frontend-shared/src/store/login-connect-store.ts
marktnoonan: I think that makes sense but going to give it a little more thought. We can remove it from this PR though, since it's not directly used here, and look at the ergonomics of a different store in the branch that implements the login flow. It's probably fine to separate them and have a small "LoginConnectModal" store.

packages/frontend-shared/src/store/login-connect-store.ts
```diff
@@ -0,0 +1,134 @@
...
+    },
+    setBannersState (banners: BannersState) {
+      this.bannersState = banners
+    },
+    setLatestBannerShownTime (timestamp: number) {
+      this.latestBannerShownTime = timestamp
+    },
+  },
+  getters: {
+    userStatus (state) {
```
amehta265, line 106: I was just wondering what the parameter 'state' refers to? Is this the state of the current cycle that user is in when using the dashboard?

amehta265 (commented):
This refactor definitely makes attributes for the ACI more understandable making the entire flow (described on the Whimsical) more structured where everything is in one place. I can also see how this could be scalable as it would be pretty easy to add new states and update the getters and actions!

packages/frontend-shared/src/store/login-connect-store.ts
```diff
@@ -0,0 +1,134 @@
...
+      isLoginConnectOpen: false,
+      utmMedium: '',
+      isLoggedIn: false,
+      isProjectConnected: false,
+      isConfigLoaded: false,
+      isOrganizationLoaded: false,
+      isMemberOfOrganization: false,
+      hasNoRecordedRuns: false,
+      error: false,
+      hasRecordedRuns: false,
```
ZachJW34: Any point in having a `hasRecordedRuns` and a `hasNoRecordedRuns`?

packages/frontend-shared/src/utils/isAllowedFeature.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  // converting to lowercase since `interval` will error on "Day" vs "day"
+  const waitTimestamp = interval(waitTime.toLocaleLowerCase())
+
+  if (isNaN(waitTimestamp)) {
+    throw new Error(`incorrect format for waitTime provided, value must be \`n days\`, \`n minutes\` etc. Value received was ${waitTime}`)
+  }
+
+  return (Date.now() - eventTime) > interval(waitTime)
```
ZachJW34:
```suggestion
  return (Date.now() - eventTime) > waitTimestamp
```

Already converted, why not use it?

packages/frontend-shared/src/utils/isAllowedFeature.ts
```diff
@@ -0,0 +1,104 @@
...
+      needsRecordedRun: [],
+      needsOrgConnect: [],
+      needsProjectConnect: [],
+      isLoggedOut: [],
+    },
+  }
+
+  let rulesToCheck = [...rules[featureName].base]
+
+  if (userStatus !== 'allTasksCompleted') {
```
ZachJW34: Is this check necessary? You have `userStatusIsNot('allTasksCompleted')` in base which would cause the `rulesToCheck.every` to fail.

packages/frontend-shared/src/store/login-connect-store.ts
ZachJW34:
```suggestion
    userStatus (state): UserStatus {
```

Good to enforce this constraint so that when you go to add a new one it's accounted for here.

packages/frontend-shared/src/store/login-connect-store.ts
```diff
@@ -0,0 +1,134 @@
...
+  'isLoggedOut',
+  'needsOrgConnect',
+  'needsProjectConnect',
+  'needsRecordedRun',
+  'allTasksCompleted',
+] as const
+
+export type UserStatus = typeof userStatuses[number]
+
+export type LoginConnectBooleanField = 'isLoggedIn' | 'isProjectConnected' | 'isConfigLoaded' | 'isOrganizationLoaded' | 'isMemberOfOrganization' | 'hasRecordedRuns' | 'hasNoRecordedRuns' | 'hasNonExampleSpec'
```
ZachJW34: I'd expect the states here to be contextualized in the store. It would be easier to understand if I saw state structured like so:

```ts
state (): LoginConnectState {
    return {
      states: {
        isLoggedIn,
        isProjectConnected,
	...
      },
      bannerState: {},
      ...
  }
)
```

packages/frontend-shared/src/utils/isAllowedFeature.ts
```diff
@@ -0,0 +1,104 @@
...
+  const waitTimestamp = interval(waitTime.toLocaleLowerCase())
+
+  if (isNaN(waitTimestamp)) {
+    throw new Error(`incorrect format for waitTime provided, value must be \`n days\`, \`n minutes\` etc. Value received was ${waitTime}`)
+  }
+
+  return (Date.now() - eventTime) > interval(waitTime)
+}
+
+export const isAllowedFeature = (
```
ZachJW34, line 32: I was expecting this function to do more... This function tells you if the feature is allowed to show, and we are deriving the user-status but the example you've shown:

```ts
  showRecordBanner.value = userStatusMatches('needsRecordedRun')
  showConnectBanner.value = userStatusMatches('needsProjectConnect')
  showCreateOrganizationBanner.value = userStatusMatches('needsOrgConnect')
  showLoginBanner.value = userStatusMatches('isLoggedOut')
  ...
```

has the developer deriving the state again. We have all the state, could this function return the specific feature to show rather than just a boolean? We know we can only show one banner at a time so all we would be asking is: What banner to show given this state/user-status.

Did we consider something like:
```ts
type Banner = null | RecordBanner | ConnectBanner | CreateOrganizationBanner | LoginBanner
const banner = allowedFeature('specsListBanner', state)
```

Then in the dom you could conditionally render based on the returned banner status.

packages/frontend-shared/src/store/login-connect-store.ts
marktnoonan: whoops, `hasRecordedRuns` ends up not being used anywhere on the usage branch. My goal is for all the rules to be phrased in a truthy way, so I wouldn't mind having both in principle, but we can safely drop this one

packages/frontend-shared/src/utils/isAllowedFeature.ts
marktnoonan: There's a particular reason the code looks that way that I probably should have called out, it's for an easier diff between the old and new logic. In reality this should end up driven by a single computed value for the correct banner as you've suggested. It ends up changing a lot of lines in that file in an already-big PR, so I'd rather do it as a different step. Maybe not needed since the code's been shown here, reviewers won't be seeing it "cold" in among many other changes.

packages/frontend-shared/src/utils/isAllowedFeature.ts
marktnoonan: `base` is per feature, the docs CI prompt currently shows after 4 days regardless of the user's connection status. Which makes some sense, you can record a run locally, so "all tasks are completed", but we don't know if you have connected CI. So that's the where we fall through and only want to apply the base rules.

I think `allTasksCompleted` should be treated more like the rest, it's weird that I put it in the base rule. It kind of evolved that way. Will update.

packages/frontend-shared/src/store/index.ts
```diff
@@ -0,0 +1,7 @@
+import { createPinia as _createPinia } from 'pinia'
+
+// Reusable installation function, used as an entry point for tests that
+// require an identical setup to main.ts
+export const createPinia = () => {
+  return _createPinia()
+}
```
mike-plummer, line 9: I'm not sure I understand the purpose of this wrapper function. Why not just directly call the library-supplied function from the `beforeEach`?

packages/frontend-shared/src/utils/isAllowedFeature.ts
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @param eventTime - timestamp of an event
+ * @param waitTime - time to compare with, such as `1 day`, `20 minutes`, etc, to be parsed by `human-interval` package
+ */
+const minTimeSinceEvent = (eventTime: number | undefined, waitTime: string) => {
+  if (!eventTime) {
+    // if there is no event time, the event has never been recorded
+    // so the "time since" is infinite, return true here
+    return true
+  }
```
mike-plummer: This feels backwards to me - I would think this function would enforce a period of time since an event occurred, thus if the event hasn't happened then we don't satisfy the condition

packages/frontend-shared/src/store/index.ts
marktnoonan: I copied this pattern from our existing pinia testing setup - it's small and it works so I didn't mess with it, but I do see your point 🤔

packages/frontend-shared/src/utils/isAllowedFeature.ts
marktnoonan: Maybe this is a problem with the name of the function (or the comment). We don't require that one event happened before the other, we only care that, if the first did happen, Cypress doesn't fire something else without a cooldown period. So this is the logic we want.

But I get why it would seem like a generic function with this name would work the other way. I sort of think, in context, reading the rules, this is OK with the current name, but the comment should be updated. I'll think on a better name though (and would love a suggestion if you have one).

packages/frontend-shared/src/store/index.ts
marktnoonan: ah, ok, it's just needed for the `export * from './login-connect-store'` to not throw an error,  but that line wasn't there yet. For the sake of consistency, I've added that line to re-export the store like we do in the other packages, and have the slightly shorter import paths.

packages/frontend-shared/src/utils/isAllowedFeature.cy.ts
marktnoonan: updated

packages/frontend-shared/src/store/login-connect-store.ts
marktnoonan: I've ended up deciding that we should leave this login modal stuff in here. Given how closely some of these things are related - the UTM params, the tooltips that close when the login modal opens, maybe some other things, where I think it will be easier to keep track in one place. 

Based on Zach's comment I structured the states a bit better, into `user` and `project`, with just a few top-level states remaining. This feels less cluttered and I think it's fine for now.

packages/frontend-shared/src/utils/isAllowedFeature.ts
marktnoonan: I restructured this a little and was able to remove the `userStatusIsNot` helper and this check. Let me know what you think of new version, I tried a couple ideas and settled on what I have now as a reasonable balance.

packages/frontend-shared/src/store/login-connect-store.ts
marktnoonan: Updated this to roughly what I shared with you on slack.

warrensplayer (approved):
I really like the organization of the state in the store.  Much easier to comprehend!

ZachJW34 (approved):
Looks like some ts-errors but going to approve. Looking forward to seeing this used, cleans up a lot and I think I'll get a better understanding once I see it in action.

PR #24117, opened 2022-10-04
fix: CSS import in CT Support file is not working

### LINKED ISSUE #23025: CSS import in component is not respected
### Current behavior

Styled has been ignored

Screen Shot 2022-07-30 at 13 23 01

### Desired behavior

Same code in storybook

Screen Shot 2022-07-30 at 13 23 40

### Test code to reproduce

[Reproduce repo](https://github.com/davidNHK/cypress-reproduce)

### Cypress Version

10.3.1

### Other

Integrate with WebPack 5 and tailwindCSS

amehta265:
- Closes #23025 

### User facing changelog

Webpack >=4 adds a new feature: giving the user the ability to add sideEffects to their package.json files. If a user adds the CT Support file as a part of their side effect, all global application styles that may have been imported in the support file (e.g. import ./example.css) are not applied. 
- The change implemented ensures that all css files are flagged as containing side Effects and thus are not tree-shaken. This will ensure that there will be no dead-code removal of css files by Webpack and so all styles imported to the CT Support file will be enforced in the user's component irrespective of the sideEffects set by the user.

### Additional details

- Why the problem exists: Webpack 4 introduced several optimizations including dead code removal through tree shaking. Webpack follows all your import and export statements keeping track of the libraries and functions you use and don’t use. So, when it sees something is being imported but not being used, it considers that as ‘dead-code’ and will tree-shake it. But dead code is not always clear though. In our case, the CT support file imports *.css files but never utilizes them. Webpack believes this is dead code and therefore removes them. But this is inherently wrong. Some imports, simply by being included in the bundle have an important impact on the application e.g. global stylesheets and global configurations (in our case *.css and *.command.js are examples of this).
- Other considered solutions:
1. Scaffolding the user's package.json file and directly editing the sideEffects flag from there. This could be messy and could lead to a lot of architecture changes with how we interact and consume the package.json
2. Setting sideEffects: true globally in the module.rules. This tells Webpack that all files contain sideEffects preventing any file from being tree shaken increasing bundle size and slowing down the application.
3. Add optimization.sideEffects: false. This means that any sideEffects flag set in the package.json will be ignored. Issue is that if the user needs some files to be protected from tree shaking while running cypress, setting this to false could lead to a situation where the user’s sideEffects flag in their package.json is ignored and therefore their intended use case is not met.
- Implemented solution: Set sideEffects: true for css modules through module.rules. In our case we want all css files to have a flag of sideEffects: true. This solution is better than setting global sideEffects: true as only css files will be prevented from tree shaking as compared to all the files in the latter option.

### Steps to test

1. Change directory to webpack-dev-server
2. Run yarn cypress:open
3. Run the react.cy.ts testing file. All the projects tested within this file have been configured to include sideEffects in their respective package.json. I have included an assertion to ensure that the background color is applied to the component after the global css files are imported into the CT Support file.

Before:
Screen Shot 2022-10-04 at 12 22 07 PM

After:
Screen Shot 2022-10-04 at 12 23 02 PM

### PR Tasks


npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -69,6 +69,22 @@ function modifyWebpackConfigForCypress (webpackConfig: Partial<Configuration>) {
     }
   }
 
+  if (webpackConfig?.module?.rules) {
+    let newRules: (RuleSetRule | '...')[] = []
+
+    webpackConfig.module.rules.forEach((ele: any) => {
+      const regex = new RegExp(ele.test)
+
+      if (regex.test('test.css')) {
```
ZachJW34: Does this work if they are using `.scss` or another flavor os css like stylus or less?

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -69,6 +69,22 @@ function modifyWebpackConfigForCypress (webpackConfig: Partial<Configuration>) {
     }
   }
 
+  if (webpackConfig?.module?.rules) {
+    let newRules: (RuleSetRule | '...')[] = []
+
+    webpackConfig.module.rules.forEach((ele: any) => {
```
astone123: Could we type this better and maybe use a different variable name like `ruleSet` or something similar?

It looks like it's possible for `ele` to equal `...` in some cases, but we're typing it as `any`, so we could potentially run into issues reading properties off of it. It's also just confusing not to have a type

npm/webpack-dev-server/src/makeWebpackConfig.ts
astone123: I think there might be some pretty common cases that we won't be able to account for using this method. I don't think we can guarantee that `test.css` will match the regex for the rule that we want to set the `sideEffects` property on.

Are we able to set this property globally in the configuration, like you mention in option 2 in the PR description? The downside would be that this could lead to large bundle sizes depending on how the app is configured 🤔

astone123 (changes requested):

npm/webpack-dev-server/src/makeWebpackConfig.ts
amehta265: You are right. This does not work for other css file extensions like .less/.scss. So the more I think about this I now realize that there could be an inordinate amount of regex expressions that a user can specify in his config, potentially, one that we haven't accounted for in our conditions.

amehta265: You are absolutely correct. This solution definitely overlooks several possible regex configs.
As suggested pivoting to another solution would ensure more consistency.

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -69,6 +69,10 @@ function modifyWebpackConfigForCypress (webpackConfig: Partial<Configuration>) {
     }
   }
 
+  if (webpackConfig?.module?.rules) {
```
astone123: Maybe we should leave a comment here about why we're doing this? I also want to make sure we're all in agreement that this is the right way to go considering that it could have a potentially large effect on bundle sizes. @ZachJW34 what are your thoughts?

amehta265: I have added a comment on my latest commit. Also:
Based on conversations I've had with @astone123 and @ZachJW34  there seems to be two possible ways to tackling this as summarized below.
- The first is to introduce a new rule in the user's `webpack config` targeting all major styling files e.g. `.css, .scss, .sass, .styl, .less` and setting the `sideEffect: true` for them. The benefit of this solution is that we guarantee that sideEffects are set for the above files as we specifically target them in the new rule we add .
- The second direction would be to modify the existing rules in `webpack config` by setting `sideEffects: true` for any existing rules that are loading styling files. The benefit is that this could be considered 'safer' as we do not add any new rules. However, there is a higher chance of missing styling files especially when the `rules.test` set by the user are complicated and don't directly refer to the file extensions.

I have gone with the first option but I'm definitely interested in everyone's thoughts

system-tests/project-fixtures/react/src/App.cy.jsx
```diff
@@ -4,5 +4,5 @@ import { App } from './App'
 it('renders hello world', () => {
   cy.mount(<App />)
   // Click on the header here to ensure that the AUT is interactable. This ensures that the dev server overlay is not displaying
-  cy.get('h1').contains('Hello World').click()
+  cy.get('h1').contains('Hello World').should('have.css', 'background-color', 'rgb(255, 0, 0)').click()
```
astone123: Nitpick, but can we add this as a separate test case here? The test case is called "renders hello world" but I think we should be explicit about what we're looking for with a separate test case. Otherwise it's not clear why the assertion is checking for that CSS, imo

astone123 (approved):
I left a minor comment. Looks good to me! Nice work on this 🎉 definitely want to get some input from others about this solution just to make sure we're doing the right thing here

system-tests/projects/webpack4_wds3-react/package.json
```diff
@@ -4,10 +4,15 @@
...
     "babel-loader": "^8.2.4",
+    "css-loader": "4",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
+    "style-loader": "2",
     "webpack": "^4",
     "webpack-dev-server": "^3"
   },
+  "sideEffects": [
+    "./cypress/support/*.js"
```
ZachJW34: Can this just be `"sideEffects": false`?

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -69,6 +69,12 @@ function modifyWebpackConfigForCypress (webpackConfig: Partial<Configuration>) {
     }
   }
 
+  // To prevent styling files to be tree shaken by webpack, we set sideEffects: true ensuring that
+  // they are not unintentionally dropped during testing in production mode.
+  if (webpackConfig?.module?.rules) {
+    webpackConfig?.module?.rules.push({ exclude: /node_modules/, test: /\.(s[ac]ss$|less$|styl$|css$)/i, sideEffects: true })
```
ZachJW34: Any reason for the `exclude`? Does this have an affect on css files that are imported from `node_modules` in the support file?

ZachJW34 (commented):
Just a few comments. I realized that Cypress + sideEffects is not ideal. Makes me almost want to disable them entirely (maybe with [this](https://webpack.js.org/configuration/optimization/#optimizationsideeffects)). The reason is that it doesn't just affect css files but any code that's imported like `import 'xxx'`. The `import './commands.js'` would be tree-shaken away as well, so it's not just css. Still, I think this is an improvement.

npm/webpack-dev-server/src/makeWebpackConfig.ts
amehta265: No particular reason excluding the node-modules. I just wanted to ensure that webpack is not bundling node_modules and hence I added this. This will have an affect on css files imported from `node_modules` and should be removed. An oversight on my end.

astone123:
@ZachJW34 I didn't see that `optimization.sideEffects` configuration value. That would be cool if we could always just disable side effects completely 💡

system-tests/projects/webpack4_wds3-react/package.json
amehta265: Yes. Setting sideEffects to false here would yield the same result.

ZachJW34 (approved):

rockindahizzy (approved):
Nice and simple

npm/webpack-dev-server/src/makeDefaultWebpackConfig.ts
```diff
@@ -33,6 +33,11 @@ export function makeDefaultWebpackConfig (
     optimization.noEmitOnErrors = false
   }
 
+  // To prevent files from being tree shaken by webpack, we set optimization.sideEffects: false ensuring that
+  // webpack does not recognize the sideEffects flag in the package.json and thus files are not unintentionally
+  // dropped during testing in production mode.
+  optimization.sideEffects = false
```
lmiller1990, line 39: Neat, did not know about this option!

amehta265: Yup! Had to do some digging around to find out what was going wrong but eventually found this!

PR #24122, opened 2022-10-04
fix: angular and nuxt ct tests now fail on uncaught exceptions

ZachJW34:
### User facing changelog
Fixes issue where Angular and Nuxt CT uncaught exceptions would allow test to pass. CT Tests that have uncaught exceptions will now cause the test to fail.

### Additional details
The fix for Nuxt was to remove some outdated code from the `vue2` package and force the error outside of Nuxt's default error handling. For Angular, a custom ErrorHandler was added that just re-throws the error.

### Steps to test
I tested by checking out my branch and making sure `npm/angular` and `npm/vue2` were built with the latest changes (normally by just running `yarn` at root). Then, I copied and pasted the contents of `dist` for each package into `node_modules/cypress/<framework>/dist` of an external project. For Angular, you have to remove the `.angular` directory otherwise the module will be cached.

The system-tests added can also be run as they scaffold an entire Nuxt/Angular application

Before:

https://user-images.githubusercontent.com/25158820/193910636-b105fa29-57a1-46c2-9602-33524d0fdf68.mov

After:

https://user-images.githubusercontent.com/25158820/193910694-800722f8-983b-4a77-850a-88775b63999c.mov

### PR Tasks

### Other

Keeping in draft for now, this would be a breaking change but these frameworks are listed as Alpha so we might move ahead. Will bring up in CT product sync.

### DIFF
--- a/packages/app/cypress/e2e/runner/ct-framework-errors.cy.ts
+++ b/packages/app/cypress/e2e/runner/ct-framework-errors.cy.ts
@@ -40,9 +40,11 @@ function loadErrorSpec (options: Options): VerifyFunc {
   cy.get('.runnable-header', { log: false }).should('be.visible')
   // Extended timeout needed due to lengthy Angular bootstrap on Windows
   cy.contains('Your tests are loading...', { timeout: 60000, log: false }).should('not.exist')
+  // Then ensure the tests have finished
+  cy.get('[aria-label="Rerun all tests"]', { timeout: 30000 })
   cy.findByLabelText('Stats').within(() => {
-    cy.get('.passed .num', { timeout: 30000 }).should('have.text', `${passCount}`)
-    cy.get('.failed .num', { timeout: 30000 }).should('have.text', `${failCount}`)
+    cy.get('.passed .num').should('have.text', `${passCount}`)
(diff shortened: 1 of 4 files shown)


packages/app/cypress/e2e/runner/ct-framework-errors.cy.ts
```diff
@@ -40,9 +40,11 @@ function loadErrorSpec (options: Options): VerifyFunc {
   cy.get('.runnable-header', { log: false }).should('be.visible')
   // Extended timeout needed due to lengthy Angular bootstrap on Windows
   cy.contains('Your tests are loading...', { timeout: 60000, log: false }).should('not.exist')
+  // Then ensure the tests have finished
+  cy.get('[aria-label="Rerun all tests"]', { timeout: 30000 })
```
ZachJW34, line 44: Need to let the test finish before asserting against the pass/fail states

rockindahizzy (approved):
👍 🔥

mike-plummer (approved):
💯

jordanpowell88 (approved):
LGTM! Been wanting to get this fixed. Great job with this @ZachJW34

PR #24125, opened 2022-10-05
fix: The record key is obstructed in v10

### LINKED ISSUE #22091: The record key is obstructed in v10
### Current behavior

Setting up the Dashboard recording using v10.0.2 and cannot select just the key, need to select the entire command using the button and then trim it to use just the key 

Screen Shot 2022-06-03 at 09 26 44

### Desired behavior

Be able to select just the record key

### Test code to reproduce

add Dashboard to any project

### Cypress Version

10.0.2

### Other

_No response_

gitstart:
- closes https://github.com/cypress-io/cypress/issues/22091

### User facing changelog
Fixed a bug where the user can select just the key, need to select the entire command using the button, and then trim it to use just the key

### Additional details
Setting up the Dashboard recording using v10.0.2 and cannot select just the key, need to select the entire command using the button and then trim it to use just the key

### Steps to test
- Add dashboard to any project.
- See the window, prompting you to run the provided command in your local or CI environment.
- Verify that you can select/highlight a specific portion of the provided command.
- Verify that you can pan around (scroll) horizontally in case of command overflow.

BEFORE

https://user-images.githubusercontent.com/45232708/194007590-65570315-4969-4182-a3af-00987707f9d4.mov

AFTER

https://user-images.githubusercontent.com/45232708/194009673-e3e07f79-246c-4c5f-942d-9a46ae9ec46b.mov

### PR Tasks


gitstart:
Hi @lmiller1990, Could you please take a look at this PR, and help trigger CI if all good?. Thanks.

packages/app/src/runs/RunsEmpty.cy.tsx
```diff
@@ -21,6 +21,6 @@ describe('RunsEmpty', () => {
     } as any
 
     cy.contains(defaultMessages.specPage.banners.record.title).should('be.visible')
-    cy.contains('npx cypress run --component --record --key abcd-efg-1234').should('be.visible')
+    cy.get('[data-cy="terminal-prompt-input"]').should('have.value', 'npx cypress run --component --record --key abcd-efg-1234')
```
astone123: Maybe this is a personal preference, but I like to use `contains` rather than test IDs because we are trying to act like the user in our tests. Users don't know about test IDs, they just scan the page for what they are trying to find, kind of like contains.

This also makes the test more brittle. If a test ID is updated, the test will fail, but why should the test fail? Nothing broke and the user experience is still there, so the test should pass.

Consider reverting your changes to these tests

gitstart: Thanks @astone123.
These changes were made based on some limitations noticed when using `.contain`.
As a result of updating command text, from DOM element to an input value as shown in the image below in `TerminalPrompt`

Screenshot 2022-10-07 at 7 16 55 AM

`cy.contain(<command text>)` fails, since it's not able to match command text anymore.
Alternatively, to bypass the usage of testId here, I could update this to `findByDisplayValue`.
WDYT? or perhaps there's an alternative way to keep the `.contain()` with current state of `TerminalPrompt`?

astone123: Oh yeah I missed that. In this case, I think using `findByDisplayValue` would be the best approach. I can make that update and then we can move this through. Thanks!

astone123:
This change is causing an issue in launchpad with the way that the inputs show up. Looking into this now
Screen Shot 2022-10-07 at 11 51 28 AM

gitstart:
Thanks @astone123 👍, taking a look as well.

gitstart:
Hi @astone123,
Kindly help take a look at the resolution of UI regression pushed in the last commits. Thanks.

packages/app/src/runs/RunsEmpty.vue
```diff
@@ -10,7 +10,10 @@
     <p class="h-48px mb-8px text-gray-600">
       {{ t("runs.empty.description") }}
     </p>
-    <RecordPromptAdapter />
+
+    <span class="w-full">
+      <RecordPromptAdapter />
+    </span>
```
gitstart: Wrapper with `w-full` className was added here to maintain existing UI, with fully spanned TerminalPrompt.

warrensplayer:
```suggestion
    <RecordPromptAdapter class="w-full"/>
```
I see why you had to add the additional `w-full` class here since this component is inside a `flex` container with the direction set to column and the align-items set to center.  `span` would not be appropriate here semantically, however you do not need it at all.  It should work to add a class prop to the `RecordPromptAdapter` and Vue will pass it down to add it to the list of classes on the top level element that is rendered.  https://vuejs.org/guide/essentials/class-and-style.html#binding-html-classes.

After this update, I think this PR will be good!

AtofStryker:
@astone123 is this ready for review?

astone123:
@AtofStryker yep, just one outstanding comment from Stokes

packages/app/src/runs/RunsEmpty.vue
gitstart: Thanks @warrensplayer that works.

warrensplayer (approved):
Looks good to me.  I merged in develop and waiting for green on CI and one more approval. @astone123 or @AtofStryker , can one of you take another look?

AtofStryker (approved):
Code changes look good as does CI. I also ran locally and got expected behavior. Great work @gitstart!

gitstart:
This PR was pushed through [Gitstart](https://www.gitstart.com/), with contributions from @chidiebereojingwa, @raph941, @phunguyenmurcul.

PR #24158, opened 2022-10-06
fix: include cypress version in header to anon-collect endpoint 

### LINKED ISSUE #24157: fix: include cypress version in header to anon-collect endpoint
### Current behavior

Calling the `anon-collect` endpoint in `EventCollectorActions.ts` does not include a 'x-cypress-version' header like other cloud endpoints include. 

### Desired behavior

Include a 'x-cypress-version' header when calling the `anon-collect` endpoint in `EventCollectorActions.ts` like other cloud endpoints. 

### Test code to reproduce

The original issue was released here: https://github.com/cypress-io/cypress/pull/23691

Testing this update will require coordinating with the cloud team to verify the header is being captured.

### Cypress Version

10.9.0

### Node version

n/a

### Operating System

n/a

### Debug Logs

_No response_

### Other

_No response_

warrensplayer:
- Closes https://github.com/cypress-io/cypress/issues/24157

### User facing changelog

n/a

### Additional details

Including a version header was part of the original requirements for the call to the `/anon-collect` endpoint but was not included

### Steps to test

Follow instructions on the PR that implemented this feature.  https://github.com/cypress-io/cypress/pull/23691

n/a

### PR Tasks

### DIFF
--- a/packages/data-context/src/actions/EventCollectorActions.ts
+++ b/packages/data-context/src/actions/EventCollectorActions.ts
@@ -1,6 +1,8 @@
 import type { DataContext } from '..'
 import Debug from 'debug'
 
+const pkg = require('@packages/root')
+
 const debug = Debug('cypress:data-context:actions:EventCollectorActions')
 
 interface CollectableEvent {
@@ -29,7 +31,10 @@ export class EventCollectorActions {
         `${dashboardUrl}/anon-collect`,
         {
           method: 'POST',
-          headers: { 'Content-Type': 'application/json' },
+          headers: {
+            'Content-Type': 'application/json',
+            'x-cypress-version': pkg.version,
+          },
           body: JSON.stringify(event),
         },
       )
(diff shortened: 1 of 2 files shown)


estrada9166 (approved):
Looks good to me

packages/data-context/src/actions/EventCollectorActions.ts
```diff
@@ -1,6 +1,8 @@
 import type { DataContext } from '..'
 import Debug from 'debug'
 
+const pkg = require('@packages/root')
```
mike-plummer, line 4: I think this is already maintained in the context, so this may be duplicative. Could you do something like:

`'x-cypress-version': (await this.ctx.versions.versionData()).current.version`?

warrensplayer: I was just following the pattern from `VersionsDataSounce`.  It gets the version this way

ZachJW34 (approved):

PR #24161, opened 2022-10-06
fix: add lib/cli to package.json exports

VinceOPS:
### User facing changelog

n/a

### Additional details

Bazel `rules_nodejs` repository is providing [a Bazel example for Cypress 7+](https://github.com/bazelbuild/rules_nodejs/tree/stable/examples/cypress). It demonstrates how `@bazel/cypress` can be used to run Cypress tests using the Bazel build system. However it relies on a file, `run-cypress.js`, which imports [`cypress/lib/cli`](https://github.com/bazelbuild/rules_nodejs/blob/stable/packages/cypress/internal/run-cypress.js#L1). 

With Cypress 10+, this causes an error `ERR_PACKAGE_PATH_NOT_EXPORTED`.

> Package subpath './lib/cli' is not defined by "exports" in /private/var/tmp/_bazel_[.....]/yarn/node_modules/cypress/package.json

**This is not surprising, as [Cypress 7 was not using `exports` entries](https://github.com/cypress-io/cypress/blob/v7.7.0/cli/package.json).**

Adding the proper entry to `exports` in `cli/package.json` solves the problem. And the Cypress tests are properly executed.

### Steps to test

- Clone the `https://github.com/bazelbuild/rules_nodejs`
- `cd` to examples/cypress
- `npm i -g @bazel/bazelisk` (in case you don't have bazel)
- `yarn test`

See it's working.

- Now use `yarn` to upgrade to Cypress 10+.
- Repeat the operations
- See that it fails with `ERR_PACKAGE_PATH_NOT_EXPORTED`

Getting into the execroot of Bazel and changing the package.json in `node_modules/cypress` to add this new entry to `exports` fix the error.

n/a

### PR Tasks


warrensplayer:
@VinceOPS The "exports" section of the CLI package.json file is there to provide the public facing API for that module.  Have you taken a look at the Cypress [Module API](https://docs.cypress.io/guides/guides/module-api) that is already exposed in the "exports" section?  It should be your entry point into driving Cypress via code. Any reason you could not go that route for supporting Cypress 10+?

VinceOPS:
> @VinceOPS The "exports" section of the CLI package.json file is there to provide the public facing API for that module. Have you taken a look at the Cypress [Module API](https://docs.cypress.io/guides/guides/module-api) that is already exposed in the "exports" section? It should be your entry point into driving Cypress via code. Any reason you could not go that route for supporting Cypress 10+?

Good morning @warrensplayer! Indeed, that's a much better approach. I focused too much on the "error" itself, and did not search for an appropriate solution for the current Cypress version. Thanks for your answer! So the work will definitely be on the `@bazel/cypress` side. Will eventually propose a custom entry point for them, which use Module API instead.

Have a good day!

closed~

PR #24170, opened 2022-10-07
feat: utilize electron fuses to ensure that the v8 snapshot is not included in the renderer process

### LINKED ISSUE #22985: Tests Run in Electron when Using V8 Snapshot Will Sometimes Crash
### What would you like?

When using v8 snapshots, electron should not crash when given a long running spec

### Why is this needed?

Each v8 context in the app uses a separate copy of the snapshot. That includes electron browsers and iframes within them. We also have a leaking iframe problem. So multiply a non trivial snapshot size times all of the iframes in a long running spec. And we run out of memory.

### Solution

* ~~Work with electron to provide a way to provide separate snapshots for main and renderer processes: https://github.com/electron/electron/issues/35170~~
* Upgrade to v21 once that issue is fixed and incorporate the changes

ryanthemanuel:
- Closes #22985

### User facing changelog

n/a

### Additional details

Due to an existing iframe memory leak and the fact that our V8 snapshot ended up getting instantiated inside of every V8 context (including every memory leak), we would crash out of memory when Cypress tests were run in the electron browser. We [worked with electron](https://github.com/electron/electron/issues/35170) to ensure that V8 snapshots would only be included in the main process. This PR makes the changes needed to utilize the functionality added in the linked electron PR. Specifically we need to use an electron fuse to ensure that we are reading our custom v8 snapshot on the main process (while keeping the default snapshot on other processes)

### Steps to test

Mostly this is tested via the v8 snapshot smoke test during binary creation. To test this manually, you can launch cypress in open mode and ensure that `window.snapshotResult` is undefined

n/a

### PR Tasks


scripts/binary/build.ts
```diff
@@ -173,6 +173,10 @@ export async function buildCypressApp (options: BuildCypressAppOpts) {
   }, { spaces: 2 })
 
   fs.writeFileSync(meta.distDir('index.js'), `\
+${!['1', 'true'].includes(process.env.DISABLE_SNAPSHOT_REQUIRE) ?
+`if (!global.snapshotResult && process.versions && process.versions.electron) {
```
mschile:
```suggestion
`if (!global.snapshotResult && process.versions?.electron) {
```

mschile (approved):

.circleci/config.yml
```diff
@@ -62,11 +65,12 @@ windowsWorkflowFilters: &windows-workflow-filters
     or:
     - equal: [ develop, << pipeline.git.branch >> ]
     - equal: [ 'feature/v8-snapshots', << pipeline.git.branch >> ]
+    - equal: [ 'ryanm/feat/upgrade-electron-with-v8', << pipeline.git.branch >> ]
     - matches:
           pattern: "-release$"
           value: << pipeline.git.branch >>
 
-executors:
+executors:
```
ryanthemanuel:
```suggestion
executors:
```

mjhenkes (approved):

Issue #23902, opened 2022-09-20
Update remaining queries to use _addQuery()


BlueWinds:
### What would you like?

Following on from https://github.com/cypress-io/cypress/issues/23550 and https://github.com/cypress-io/cypress/issues/23804, we need to migrate the remaining commands that will be queries to use the new API. The remaining list is:

    .children()
    .closest()
    .debug()
    .document()
    .eq()
    .filter()
    .first()
    .focused()
    .getCookie()
    .getCookies()
    .hash()
    .invoke()
    .its()
    .last()
    .location()
    .next()
    .nextAll()
    .nextUntil()
    .not()
    .parent()
    .parents()
    .parentsUntil()
    .pause()
    .prev()
    .prevAll()
    .prevUntil()
    .siblings()
    .title()
    .url()
    .window()

While this looks like quite a huge list, the process is quite quick for most command -> query migrations, and in particular many of these are simply aliases - .last(), .first(), .next() etc are all defined in the same file and use the same logic underneath the hood.

BlueWinds:
Completed in https://github.com/cypress-io/cypress/pull/24203

PR #24203, opened 2022-10-11
feat: addQuery Remaining Queries

BlueWinds:
- Closes https://github.com/cypress-io/cypress/issues/23902

## User facing changelog
- If `cy.getCookie()` fails to find a cookie, you can no longer traverse the non-existent cookie's properties (eg, with `.its('value)`). If you expect a cookie not to exist, assert that directly, eg: `cy.getCookie('foo1').should('not.exist')`
- `.invoke()` no longer waits for promises to resolve. If you wish to await a promise, use `cy.then()` instead.

This PR updates the following commands to be queries:
```
.debug()
.document()
.end()
.focused()
.hash()
.its()
.invoke()
.location()
.log()
.pause()
.url()
.title()
.window()
```

All traversals are also updated (traversals are grouped together because they share code):
```
.children()
.closest()
.eq()
.filter()
.find()
.first()
.last()
.next()
.nextAll()
.nextUntil()
.not()
.parent()
.parents()
.parentsUntil()
.prev()
.prevAll()
.prevUntil()
.siblings()
```

Error messages when selectors fail, and when commands receive invalid subjects, have also been improved; see below for screenshots.

## Steps to test
```
it('passes', () => {
  cy.$$('body').prepend('<div id="foo"><button>foo</button></div>')

  setTimeout(() => {
    cy.$$('#foo').replaceWith('<span id="foo"><a class="active">bar</a></span>')
  }, 1000)

  cy.get('#foo').children().should('have.class', 'active')
})
```

This test passes on this branch, and fails on `develop` or `issue-7306`.

### Traversal failure
When queries initially succeed but later fail (because the DOM has updated), the error message now shows the whole query that led up to the failure, rather than just the element it failed on.

Old:
> Timed out retrying after 500ms: Expected to find element: `children`, but never found it. Queried from element: `<span#foo>`

New:
> Timed out retrying after 500ms: Expected to find element: `children`, but never found it. Queried from:
>
> > `cy.get(body).find(#foo)`

image

### Subject validation
In addition to fixing some Detached DOM errors (but not all of them!), several error messages have been updated or added.

When a command receives an invalid subject, it now lists all valid subject types, rather than only the first plus the mysterious phrase "all X subject validations failed".

Old:
> cy.scrollTo() failed because it requires a DOM element.
> ...
> All 2 subject validations failed on this subject.

New:
>  cy.scrollTo() failed because it requires a DOM element or window.
> ...

image

### Accessing properties

When using .its() or .invoke() to access nested properties, error messages now contain the full path the user requested. This more closely matches the error message to what the user wrote in their test.

Old:
> Timed out retrying after 100ms: `cy.invoke()` errored because the property: `bar` returned a `string` value instead of a function. cy.invoke() can only be used on properties that return callable functions.
>
> cy.invoke() waited for the specified property `bar` to return a function, but it never did.

New:
> Timed out retrying after 100ms: `cy.invoke()` errored because the property: `foo.bar` returned a `string` value instead of a function. cy.invoke() can only be used on properties that return callable functions.
>
> cy.invoke() waited for the specified property `foo.bar` to return a function, but it never did.

image

### Failed assertions following traversals

When a failed assertion follows a traversal, we now use "subject" as a placeholder to indicate that we never actually reached the assertion.

Old:
> [find] `#wrapper`
> [contains] Newer
> [assert] expected `:cy-contains('Newer'), [type='submit'][value~='Newer']` to have class `updated`

> Timed out retrying after 4000ms: Expected to find content: 'Newer' within the element: <div#wrapper> but never did.

New:
> [find] `#wrapper`
> [contains] Newer
> [assert] expected `subject` to have class `updated`

> Timed out retrying after 4000ms: Expected to find content: 'Newer' within the element: <div#wrapper> but never did.

AssertionError
Timed out retrying after 4000ms: Expected to find content: 'Newer' within the element: <div#wrapper> but never did.
image

## PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4785


packages/driver/cypress/e2e/commands/assertions.cy.js
```diff
@@ -134,21 +134,6 @@ describe('src/cy/commands/assertions', () => {
       cy.noop(obj).its('requestJSON').should('have.property', 'teamIds').should('deep.eq', [2])
     })
 
-    // TODO: make cy.then retry
```
BlueWinds, line 137: In the intervening years since this TODO was left, we've refined and better defined exactly what retries and what doesn't.

`.then()` will never retry, and should not, so removing the commented out tests.

packages/driver/cypress/e2e/commands/connectors.cy.js
```diff
@@ -729,16 +729,15 @@ describe('src/cy/commands/connectors', () => {
...
               return 'foo'
-            } }).invoke(() => {
-              return {}
-            })
+            } })
+            .invoke(() => {})
           })
 
-          it('throws when first parameter is neither of type object nor of type string nor of type number', function (done) {
+          it('throws when we can\'t determine both a valid options and path', function (done) {
```
BlueWinds, line 736: .invoke is fundamentally too flexible. Consider the following invoke commands, all valid.

```
// Valid
.invoke('method', 'arg1')
.invoke({log: false}, 'method', 'arg1')

// Invalid
.invoke(false, 'method', 'arg1')
.invoke(false, 'arg1')
```

The invalid cases *both have the signature [bool, string].* Is this an invalid options argument, or an invalid path argument? The function signature of .invoke() is too versatile / vague to know for sure.

The new implementation throws a slightly different error, because it's simpler / has fewer conditionals and switch statements than the original. I think it's just as valid of an error.

packages/driver/cypress/e2e/commands/connectors.cy.js
```diff
@@ -903,30 +902,19 @@ describe('src/cy/commands/connectors', () => {
               Command: 'invoke',
               Function: '.bar()',
               Subject: this.obj,
+              'With Arguments': [],
```
BlueWinds, line 905: .invoke() now always includes `With Arguments`, even if it's an empty array. I strongly prefer consistency in my values over special cases (eg, omit when it's empty).

packages/driver/cypress/e2e/commands/connectors.cy.js
```diff
@@ -903,30 +902,19 @@ describe('src/cy/commands/connectors', () => {
...
-        })
-
-        it('#consoleProps as a function reduced property with args', function () {
+        it('#consoleProps as a deep function property with args', function () {
           cy.noop(this.obj).invoke('math.sum', 1, 2, 3).then(function () {
             expect(this.lastLog.invoke('consoleProps')).to.deep.eq({
               Command: 'invoke',
               Function: '.math.sum(1, 2, 3)',
               'With Arguments': [1, 2, 3],
-              Subject: this.obj['math'],
```
BlueWinds, line 929: Behavior change: .invoke() now always puts the actual subject as the `Subject` in its console logs. The subject was this.obj, the property retrieved is `.math.sum`. IMO this is more consistent and intuitive.

packages/driver/cypress/e2e/commands/connectors.cy.js
```diff
@@ -1058,7 +1047,7 @@ describe('src/cy/commands/connectors', () => {
           cy.on('fail', (err) => {
             const { lastLog } = this
 
-            expect(err.message).to.include('Timed out retrying after 100ms: `cy.invoke()` errored because the property: `baz` does not exist on your subject.')
+            expect(err.message).to.include('Timed out retrying after 100ms: `cy.invoke()` errored because the property: `foo.bar.baz.fizz` does not exist on your subject.')
```
BlueWinds, line 1050: Behavior change: .invoke() now logs the full path when we tried to fetch it and failed. This makes it easier to debug and more consistent with the rest of Cypress' API design.

I think the fact it didn't used to work this way is more a quirk of implementation than any sort of thought-out-feature.

packages/driver/cypress/e2e/commands/misc.cy.js
```diff
@@ -229,9 +229,8 @@ describe('src/cy/commands/misc', () => {
 
       it('throws when wrapping an array of windows', (done) => {
         cy.on('fail', (err) => {
-          expect(err.message).to.include('`cy.scrollTo()` failed because it requires a DOM element.')
+          expect(err.message).to.include('`cy.scrollTo()` failed because it requires a DOM element or window.')
```
BlueWinds, line 232: Behavior change (screenshot in PR description): Rather than the mysterious 'All X subject validations failed' message, we now just list out what the valid subjects are.

packages/driver/cypress/e2e/commands/querying/focused.cy.js
```diff
@@ -95,7 +95,7 @@ describe('src/cy/commands/querying', () => {
         cy.get('body').focused().then(function () {
           const { lastLog } = this
 
-          expect(lastLog.get('type')).to.eq('parent')
+          expect(lastLog.get('type')).not.to.eq('child')
```
BlueWinds, line 98: Logs are parents unless they're made children - the value is now `undefined` rather than `'parent'` - no practical difference.

packages/driver/cypress/e2e/cypress/cy.cy.js
```diff
@@ -416,10 +415,10 @@ describe('driver/src/cypress/cy', () => {
         return orig(`foo${arg1}`)
       })
 
-      Cypress.Commands.overwrite('first', (orig, subject) => {
-        subject = $([1, 2])
+      Cypress.Commands.overwrite('each', (orig, subject, cb) => {
```
BlueWinds, line 418: .first() is now a query, and you can't overwrite queries with commands. You can:
1. Overwrite commands with commands or queries
2. Overwrite queries with other queries

This is basically a limitation of the .overwrite() works: it expects to receive the old command function as an argument. Well, queries don't have command functions. They aren't commands. So there are a couple of options here:

1. Leave things as they stand in this PR (eg, queries cannot be overwritten by commands).
2. Add caveats to .overwrite(), saying that if you overwrite a query with a command, you'll get a query function rather than a command function as an argument, and that these behave differently.

It's something I want to spend more time thinking about before release, but I don't necessarily want to block this PR into the feature branch on making a decision one way or another.

packages/driver/cypress/e2e/e2e/origin/commands/connectors.cy.ts
```diff
@@ -111,8 +111,8 @@ context('cy.origin connectors', () => {
         expect(consoleProps.Function).to.equal('.text()')
         expect(consoleProps.Yielded).to.equal('button')
 
-        expect(consoleProps.Subject).to.have.property('tagName').that.equals('BUTTON')
-        expect(consoleProps.Subject).to.have.property('id').that.equals('button')
+        expect(consoleProps.Subject[0]).to.have.property('tagName').that.equals('BUTTON')
+        expect(consoleProps.Subject[0]).to.have.property('id').that.equals('button')
```
BlueWinds, line 115: .invoke() is now consistent with .its() (see the test directly above this in the same file), and the Subject is `jQuery(Button)` rather than a raw `Button`.

packages/driver/cypress/e2e/e2e/origin/cookie_behavior.cy.ts
```diff
@@ -305,7 +305,7 @@ describe('Cookie Behavior with experimentalSessionAndOrigin=true', () => {
                   // firefox actually sets the cookie correctly
                   cy.getCookie('foo1').its('value').should('equal', 'bar1')
                 } else {
-                  cy.getCookie('foo1').its('value').should('equal', null)
+                  cy.getCookie('foo1').should('equal', null)
```
BlueWinds, line 308: Discussed these test changes with Bill - this only ever passed because of an implementation quirk in the interaction between .getCookie() and .its(). 

Basically, .its() is now more correct, and this test - relying on a bug - needed to be rewritten to assert what it actually meant to assert.

packages/driver/cypress/e2e/e2e/origin/logging.cy.ts
```diff
@@ -30,8 +30,7 @@ describe('cy.origin logging', () => {
     })
   })
 
-  // TODO: fix flaky test https://github.com/cypress-io/cypress/issues/21300
-  it.skip('logs cy.origin as group when failing with validation failure', () => {
+  it('logs cy.origin as group when failing with validation failure', () => {
```
BlueWinds, line 33: Linked issue is resolved, uncommenting out these tests (haven't seen any flake in them in this branch).

packages/driver/src/cy/commands/connectors.ts
```diff
@@ -142,303 +140,158 @@ export default function (Commands, Cypress, cy, state) {
...
-    return invokeBaseFn(userOptions || { log: true }, subject, str, ...args)
-  }
+  // to allow the falsy value 0 to be used
+  const isPath = (str) => (!!str || str === 0)
 
-  const invokeFn = (subject, userOptionsOrStr, ...args) => {
-    const userOptionsPassed = _.isObject(userOptionsOrStr) && !_.isFunction(userOptionsOrStr)
-    let userOptions: Record<string, any> | null = null
-    let str = null
+  Commands._addQuery('its', function its (path, options: Partial<Cypress.Loggable & Cypress.Timeoutable> = {}, ...args) {
```
BlueWinds, line 146: I don't particularly like the way .its() and .invoke() are extremely intertwined. I spent a little bit of time pulling them apart, but didn't do it all the way - you can see here that .its() still has logic for handling .invoke() error messages. :/

packages/driver/src/cy/commands/connectors.ts
```diff
@@ -142,303 +140,158 @@ export default function (Commands, Cypress, cy, state) {
...
-      })
-    }
-
-    if (isCmdIts && args && args.length > 0) {
-      $errUtils.throwErrByPath('invoke_its.invalid_num_of_args', {
-        onFail: options._log,
-        args: { cmd: name },
-      })
-    }
+    const log = this.get('_log') || (options.log !== false && Cypress.log({
```
BlueWinds, line 173: This is one of the ways .invoke / .its are intertwined: .invoke() creates a log with a different message, then calls .its(). .its() then pulls out that log message and keeps it up to date.

Ugly.

packages/driver/src/cy/commands/querying/querying.ts
```diff
@@ -327,11 +329,13 @@ export default (Commands, Cypress, cy, state) => {
       }
     })
 
+    const withinSubject = cy.state('withinSubject')
+
     return (subject) => {
-      cy.ensureSubjectByType(subject, ['optional', 'window', 'document', 'element'], this)
+      cy.ensureSubjectByType(subject, ['optional', 'element', 'window', 'document'], this)
```
BlueWinds, line 335: Moved 'element' to the front, so that the new error messages say "expects a DOM element, window or document" - putting the most common / likely value first.

packages/driver/src/cy/commands/window.ts
```diff
@@ -102,108 +88,50 @@ export default (Commands, Cypress, cy, state) => {
...
-    window (userOptions: Partial<Cypress.Loggable & Cypress.Timeoutable> = {}) {
-      const options: InternalWindowOptions = _.defaults({}, userOptions, { log: true })
-
-      if (options.log) {
-        options._log = Cypress.log({ timeout: options.timeout })
-      }
-
-      const getWindow = () => {
-        const window = state('window')
+  Commands._addQuery('title', function title (options: Partial<Cypress.Loggable & Cypress.Timeoutable> = {}) {
```
BlueWinds, line 91: Nice example of how a query can have the same functionality, but be a lot shorter because we don't have to re-write the retry logic inside each one. Was very pleased with how the changes to this file panned out.

system-tests/projects/remote-debugging-disconnect/cypress/e2e/spec.cy.ts
```diff
@@ -9,15 +15,13 @@ describe('e2e remote debugging disconnect', () => {
...
+    cy.then(() => {
+      const onRetry = () => {
+        return callAutomation().catch(() => cy.retry(onRetry, {}))
+      }
 
-    // this will attempt to run a CDP command, realize the socket is dead, enqueue it,
-    // and start the reconnection process
-    cy.wrap(Cypress)
-    // @ts-ignore
-    .invoke('automation', 'remote:debugger:protocol', {
```
BlueWinds, line 17: This test is a victim of the breaking change to .invoke(): It no longer accepts asynchronous functions (like Cypress.automation).

system-tests/projects/remote-debugging-disconnect/plugins.js
```diff
@@ -12,6 +12,8 @@ let server
 // this is a transparent TCP proxy for Chrome's debugging port
 // it can kill all existing connections or shut the port down independently of Chrome or Cypress
 const startTcpProxy = () => {
+  console.error('starting tcp proxy ', { realPort, fakePort })
```
BlueWinds, line 15: Just added an extra logging line that was helpful to me understanding how this system test worked.

packages/driver/cypress/e2e/commands/connectors.cy.js
```diff
@@ -1659,19 +1569,19 @@ describe('src/cy/commands/connectors', () => {
           cy.wrap(obj).its('foo.bar.baz.fizz')
         });
 
-        [null, undefined].forEach((val) => {
+        [null/*, undefined*/].forEach((val) => {
```
flotwig, line 1572: If this is intentionally commented out, there should be a comment explaining why.

packages/driver/cypress/e2e/e2e/origin/cookie_behavior.cy.ts
flotwig: Is this `.its` bug recorded anywhere? Even if there's no issue for it, I think it's worth mentioning in the changelog. We're probably not the only folks with tests relying on this quirk.

packages/driver/cypress/e2e/commands/querying/focused.cy.js
flotwig: In that case, do you think it'd be cleaner to change `type` to a boolean flag? `log.isParent`?

flotwig (commented):
The "behavior changes" you've mentioned make sense to me :+1:

chrisbreiding (approved):
Looks great! It's awesome all these commands will now be resilient to detached dom elements. I really like the error message improvements too.

Should this have an accompanying docs PR for the invoke changes?

BlueWinds:
Added a new test string change, and matching test: **Failed assertions following traversals**. See that section of the PR description for an image of the change.

Going through PR feedback now.

packages/driver/cypress/e2e/commands/connectors.cy.js
BlueWinds: Not intentional; Fixed in latest commit.

packages/driver/cypress/e2e/commands/querying/focused.cy.js
BlueWinds: Possibly? I'd have to spend more time looking at logs specifically - there's a lot of difficulties and intricacies with them that I think could benefit from a top-level re-examination. I'm not sure parent / child should be a property of the log at all - it's more a property of the log's relationship to other log messages. "Nesting level" might be a more appropriate description than "type".

With that said, I'm not super excited about doing it as part of detached dom - I went down that rabbit hole a little bit earlier in the project and noped out when the scope started expanding on me.

packages/driver/cypress/e2e/e2e/origin/cookie_behavior.cy.ts
BlueWinds: Hm. I struggled a bit with how to express this. Adding this to the changelog:

> If `cy.getCookie()` fails to find a cookie, you can no longer traverse the non-existent cookie's properties (eg, with `.its('value)`). If you expect a cookie not to exist, assert that directly, eg: `cy.getCookie('foo1').should('not.exist')`

BlueWinds:
> Looks great! It's awesome all these commands will now be resilient to detached dom elements. I really like the error message improvements too.
> 
> Should this have an accompanying docs PR for the invoke changes?

There definitely should be an accompanying docs PR; I've just been putting updating the docs off for a bit until we finish going through the product review phase for queries, so I can go through the docs page by page and update everything that needs updating all at once.

With that said, the invoke changes specifically are straightforward, so I'll start a detached DOM docs branch today.

BlueWinds:
Matching docs PR: https://github.com/cypress-io/cypress-documentation/pull/4785

packages/driver/src/cy/commands/screenshot.ts
```diff
@@ -528,9 +528,9 @@ export default function (Commands, Cypress, cy, state, config) {
       // we are not limited to "within" subject
       // https://github.com/cypress-io/cypress/issues/14253
       if (userOptions.capture !== 'runner') {
-        const withinSubject = state('withinSubject')
+        const withinSubject = cy.getSubjectFromChain(cy.state('withinSubjectChain') || [])
```
emilyrohrbough, line 531: I'd be nice if getSubjectFromChain would handle checking for undefined.

BlueWinds: Yeah, I've been thinking the same thing.

I'm going to do it in the next PR rather than here - I already have some changes queued to the subject API, so rolling it in there makes sense.

flotwig (approved):

PR #24231, opened 2022-10-12
chore: try bumping timeout for Angular webpack-dev-server tests

astone123:
These Angular test projects take forever to compile, so these tests frequently fail timing out on the tests loading screen. I bumped the timeout to 60 seconds for each project. This should immediately help, but in the future we should figure out ways to slim down our Angular system test projects so that they don't take as long to compile.

### DIFF
--- a/npm/webpack-dev-server/cypress/e2e/angular.cy.ts
+++ b/npm/webpack-dev-server/cypress/e2e/angular.cy.ts
@@ -22,7 +22,7 @@ for (const project of WEBPACK_REACT) {
     it('should mount a passing test', () => {
       cy.visitApp()
       cy.contains('app.component.cy.ts').click()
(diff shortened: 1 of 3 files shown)


marktnoonan (approved):

BlueWinds (approved):
Do you have any ideas on how to speed up / slim down compilation here? 60s timeouts feel quite long.

(that's not a blocker on a bandaid solution, just a question from someone not terribly familiar with Angular)

astone123:
@BlueWinds I'm not sure, but a couple things that come to mind are removing unnecessary dependencies or maybe tweaking some things with the development server. It probably doesn't need to be 60 seconds, it probably averages compiling around 35-40 seconds but I wasn't sure how to test since it's only an issue in CI 😞

PR #24262, opened 2022-10-14
feat: cleanup binary with respect to v8 snapshot dependencies

### LINKED ISSUE #23123: Optimize the Binary with V8 Snapshots by Removing Unneeded Dependencies

ryanthemanuel:
- Closes #23123

### User facing changelog

n/a

### Additional details

This PR's main goal is to decrease the binary size by removing node module files that are no longer needed since they are included in the V8 snaphshot. Care needs to be taken as not everything in Cypress is executed in a context that can utilize the V8 snapshot. Specifically the places that can't are defined by these entry points:

* The initial code that is used to boot strap the snapshot
* The child process code that is used to load consumer config
* The rewriter worker thread

We determine the full tree using the entry points above using `esbuild`. Care needs to be taken to flag any `require.resolve` dependencies as new entry points since they do not get pulled into the tree by `esbuild`. We flag these using the warnings provided by `esbuild` and then continue to run `esbuild` until we have no more entry points to add. In this way, we can determine all of the files that absolutely need to remain.

So our general thought is to use the files in the snapshot cache as a starting point of what can be removed, filter out the dependencies determined using the above process, and then remove the resulting set of files.

This PR:

* Implements the above algorithm
* Turns back on minify in production for the snapshot JS file. It was originally determined that this doesn't affect performance all that much. While this is true in terms of startup, it *does* affect the size of the snapshot binary file itself, thus reducing the size of the overall binary.
* Perform some additional cleanup to the binary (e.g. intermediary build files, ESM distributions, docs, tests, etc.). The goal here was to get the size of the binary on disk to be close to where it was in previous releases. The zip file size will decrease.
* Clean up any empty directories

### Steps to test

You can use the binaries attached to the last commit to just play around and test this to ensure that nothing is missing that shouldn't be.

n/a

### PR Tasks


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Cypress is a next generation front end testing tool built for the modern web",
   "private": true,
   "scripts": {
-    "binary-build": "node ./scripts/binary.js build",
+    "binary-build": "node --max-old-space-size=8192 ./scripts/binary.js build",
```
ryanthemanuel: These are necessary since we are now running terser as part of the v8 snapshot process

packages/server/lib/util/performance_benchmark.js
```diff
@@ -15,7 +15,7 @@ const debugElapsedTime = (event) => {
   const now = performance.now()
   const delta = now - global.cypressServerStartTime
 
-  debug(`elapsed time at ${event}: ${threeDecimals(delta)}ms`)
+  console.log(`elapsed time at ${event}: ${threeDecimals(delta)}ms`)
```
ryanthemanuel:
```suggestion
  debug(`elapsed time at ${event}: ${threeDecimals(delta)}ms`)
```

packages/server/test/integration/http_requests_spec.js
```diff
@@ -1668,7 +1668,7 @@ describe('Routes', () => {
           'set-cookie': [
             'foo=bar; Path=/',
             '___utmvmXluIZsM=fidJKOsDSdm; path=/; Max-Age=900',
-            '___utmvbXluIZsM=bZM\n    XtQOGalF: VtR; path=/; Max-Age=900',
+            '___utmvbXluIZsM=bZM    XtQOGalF: VtR; path=/; Max-Age=900',
```
ryanthemanuel:
```suggestion
            '___utmvbXluIZsM=bZM\n    XtQOGalF: VtR; path=/; Max-Age=900',
```

packages/server/test/integration/http_requests_spec.js
```diff
@@ -1678,7 +1678,7 @@ describe('Routes', () => {
           expect(res.headers['set-cookie']).to.deep.eq([
             'foo=bar; Path=/',
             '___utmvmXluIZsM=fidJKOsDSdm; path=/; Max-Age=900',
-            '___utmvbXluIZsM=bZM\n    XtQOGalF: VtR; path=/; Max-Age=900',
+            '___utmvbXluIZsM=bZM    XtQOGalF: VtR; path=/; Max-Age=900',
```
ryanthemanuel:
```suggestion
            '___utmvbXluIZsM=bZM\n    XtQOGalF: VtR; path=/; Max-Age=900',
```

package.json
ryanthemanuel:
```suggestion
    "binary-build": "node ./scripts/binary.js build",
```

tooling/v8-snapshot/src/doctor/process-script.async.ts
```diff
@@ -24,8 +25,13 @@ export class AsyncScriptProcessor {
   constructor () {
     logInfo('Initializing async script processor')
 
-    const maxWorkers = os.cpus().length
-    const minWorkers = maxWorkers / 2
+    if (process.env.CI) {
+      logInfo('Running in CI', os.cpus(), fs.existsSync('/sys/fs/cgroup/cpuset/cpuset.cpus') ? fs.readFileSync('/sys/fs/cgroup/cpuset/cpuset.cpus', 'utf8') : 'no cpuset.cpus')
+    }
+
```
ryanthemanuel:
```suggestion
```

tooling/v8-snapshot/src/doctor/process-script.async.ts
```diff
@@ -3,6 +3,7 @@ import { strict as assert } from 'assert'
 import os from 'os'
 import WorkerNodes from 'worker-nodes'
 import type { ProcessScriptOpts, ProcessScriptResult } from '../types'
+import fs from 'fs'
```
ryanthemanuel:
```suggestion
```

package.json
```diff
@@ -15,8 +15,8 @@
     "build": "yarn build-npm-modules && lerna run build --stream --no-bail --ignore create-cypress-tests --ignore cypress --ignore \"'@packages/{runner}'\" --ignore \"'@cypress/{angular,react,react18,vue,vue2,mount-utils,svelte}'\" && node ./cli/scripts/post-build.js && lerna run build --stream --scope create-cypress-tests",
     "build-npm-modules": "lerna run build --scope cypress --scope @cypress/mount-utils && lerna run build --scope \"'@cypress/{angular,react,react18,vue,vue2,svelte}'\"",
     "build-prod": "lerna run build-prod-ui --stream && lerna run build-prod --stream --ignore create-cypress-tests && node ./cli/scripts/post-build.js && lerna run build-prod --stream --scope create-cypress-tests --scope",
-    "build-v8-snapshot-dev": "node tooling/v8-snapshot/scripts/setup-v8-snapshot-in-cypress.js --env=dev",
-    "build-v8-snapshot-prod": "node tooling/v8-snapshot/scripts/setup-v8-snapshot-in-cypress.js",
+    "build-v8-snapshot-dev": "node --max-old-space-size=8192 tooling/v8-snapshot/scripts/setup-v8-snapshot-in-cypress.js --env=dev",
```
ryanthemanuel, line 18: This is necessary since we are now using `terser` to minify the snapshot

scripts/binary/binary-cleanup.js
```diff
@@ -0,0 +1,190 @@
...
+  }
+}
+
+const getDependencyPathsToKeep = async () => {
+  let entryPoints = new Set([
+    // This is the entry point for the server bundle. It will not have access to the snapshot yet. It needs to be kept in the binary
+    require.resolve('@packages/server/index.js'),
+    // This is a dynamic import that is used to load the snapshot require logic. It will not have access to the snapshot yet. They needs to be kept in the binary
+    require.resolve('@packages/server/hook-require.js'),
+    // These dependencies are started in a new process or thread and will not have access to the snapshot. It needs to be kept in the binary
```
ryanthemanuel:
```suggestion
    // These dependencies are started in a new process or thread and will not have access to the snapshot. They need to be kept in the binary
```

scripts/binary/binary-cleanup.js
```diff
@@ -0,0 +1,190 @@
...
+    // These dependencies are started in a new process or thread and will not have access to the snapshot. They need to be kept in the binary
+    require.resolve('@packages/server/lib/plugins/child/require_async_child.js'),
+    require.resolve('@packages/server/lib/plugins/child/register_ts_node.js'),
+    require.resolve('@packages/rewriter/lib/threads/worker.ts'),
+    // These dependencies use the `require.resolve(<dependency>, { paths: [<path>] })` pattern where <path> is a path within the cypress monorepo. These will not be
+    // pulled in by esbuild but still need to be kept in the binary.
+    require.resolve('webpack'),
+    require.resolve('webpack-dev-server', { paths: [path.join(__dirname, '..', '..', 'npm', 'webpack-dev-server')] }),
+    require.resolve('html-webpack-plugin-4', { paths: [path.join(__dirname, '..', '..', 'npm', 'webpack-dev-server')] }),
+    require.resolve('html-webpack-plugin-5', { paths: [path.join(__dirname, '..', '..', 'npm', 'webpack-dev-server')] }),
```
mjhenkes, line 54: we use both 4 and 5?

scripts/binary/binary-cleanup.js
```diff
@@ -0,0 +1,190 @@
...
+  if (fileNames.length === 0) {
+    await fs.rmdir(directory)
+  }
+}
+
+const getDependencyPathsToKeep = async () => {
+  let entryPoints = new Set([
+    // This is the entry point for the server bundle. It will not have access to the snapshot yet. It needs to be kept in the binary
+    require.resolve('@packages/server/index.js'),
+    // This is a dynamic import that is used to load the snapshot require logic. It will not have access to the snapshot yet. They needs to be kept in the binary
```
mschile:
```suggestion
    // This is a dynamic import that is used to load the snapshot require logic. It will not have access to the snapshot yet. It needs to be kept in the binary
```

scripts/binary/binary-cleanup.js
```diff
@@ -0,0 +1,190 @@
...
+      'freebsd',
+      'linux',
+      'openbsd',
+      'sunos',
+      'win32'].map((platform) => require.resolve(`default-gateway/${platform}`)),
+  ])
+  let esbuildResult
+  let newEntryPointsFound = true
+
+  // The general idea here is to run esbuild with on entry points that are used outside of the snapshot. If, during the process,
```
mschile:
```suggestion
  // The general idea here is to run esbuild on entry points that are used outside of the snapshot. If, during the process,
```

scripts/binary/binary-cleanup.js
```diff
@@ -0,0 +1,190 @@
...
+  // 1. Retrieve all dependencies that still need to be kept in the binary. In theory, we could use the bundles generated here as single files within the binary,
+  // but for now, we just track on the dependencies that get pulled in
+  const keptDependencies = await getDependencyPathsToKeep()
+
+  // 2. Gather the dependencies that could potentially be removed from the binary due to being in the snapshot
+  const potentiallyRemovedDependencies = [...snapshotMetadata.healthy, ...snapshotMetadata.deferred, ...snapshotMetadata.norewrite]
+
+  // 3. Remove all dependencies that are in the snapshot but not in the list of kept dependencies from the binary
+  await Promise.all(potentiallyRemovedDependencies.map(async (dependency) => {
+    if (![...keptDependencies, 'package.json', 'packages/server/server-entry.js'].includes(dependency.slice(2)) && !dependency.includes('marionette-client')) {
```
mschile: What's special about the `marionette-client`?

mschile: Maybe add the extra dependencies above when declaring `keptDependencies` instead of here.

scripts/binary/binary-cleanup.js
ryanthemanuel: Yes. We try and match the webpack version being used by the consuming project

mjhenkes (approved):

scripts/binary/binary-cleanup.js
ryanthemanuel: Good catch. Added a comment here: [`6680189` (#24262)](https://github.com/cypress-io/cypress/pull/24262/commits/66801890ef5989448eb17cbcef56b99f2ef328ab)

`marionette-client` is very weird and dynamic in how they require dependencies (https://github.com/mozilla-b2g/gaia/blob/master/tests/jsmarionette/client/marionette-client/lib/marionette/index.js#L7-L13)

ryanthemanuel: Fixed here: [`6680189` (#24262)](https://github.com/cypress-io/cypress/pull/24262/commits/66801890ef5989448eb17cbcef56b99f2ef328ab)

mschile (approved):

Issue #21398, opened 2022-05-09
[sessions] Improve Validation Error UI in Recreated Session


emilyrohrbough:
### What would you like?

The `cy.session` command is re-using the test error component to share details when a session failed validation and the `cy.session` command catches the error to then recreate the session. This is a bit unexpected in the UI and we'd like to update the UI to display these details differently.

Current UI (10.0 visuals): 

<img width="1206" alt="Screen Shot 2022-05-09 at 9 20 22 AM" src="https://user-images.githubusercontent.com/14099737/167430508-ea5bdf0c-9b25-49ca-b00e-289de1b4c33f.png">

### Designs
https://www.figma.com/file/E86dG4e1dcuyRe33B2j3zU/app-reporter%3A-sessions?node-id=17733%3A151907

PR #24266, opened 2022-10-14
feat: enhance test error UI

emilyrohrbough:
- Apart of #21398 - This portion updates the UI for better error handling in `cy.session()`. Next step will be to update how `cy.session()` captures and reports these errors. It was getting too big when doing the work all-in-one.

### User facing changelog
I wouldn't necessarily call these note-worthy changes for the changelog. Addressed in [#24266](https://github.com/cypress-io/cypress/pull/24266).

- Enhancements were made to improve the UI around reporter test errors in the Reporter.

### User Facing Changing
Review Percy screenshots to see the full impact of changes.

(edit: latest push tweaked these a bit, but these are the general look. See Percy!)

**Test Error:**
Screen Shot 2022-10-14 at 2 41 47 PM

**Recovered In-Test Error (currently only observed when using `cy.session()`)**
Screen Shot 2022-10-14 at 2 42 54 PM

### PR Tasks
- [] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 

### DIFF
--- a/packages/reporter/src/commands/command-model.ts
+++ b/packages/reporter/src/commands/command-model.ts
@@ -35,7 +35,6 @@ export interface CommandProps extends InstrumentProps {
   wallClockStartedAt?: string
   hookId: string
   isStudio?: boolean
-  showError?: boolean
   group?: number
   groupLevel?: number
   hasSnapshot?: boolean
@@ -45,7 +44,7 @@ export interface CommandProps extends InstrumentProps {
 export default class Command extends Instrument {
   @observable.struct renderProps: RenderProps = {}
   @observable.struct sessionInfo?: SessionProps['sessionInfo']
-  @observable err = new Err({})
+  @observable err?: Err
   @observable event?: boolean = false
   @observable isLongRunning = false
   @observable number?: number
@@ -56,7 +55,6 @@ export default class Command extends Instrument {
   @observable children: Array<Command> = []
   @observable hookId: string
   @observable isStudio: boolean
-  @observable showError?: boolean = false
   @observable group?: number
   @observable groupLevel?: number
   @observable hasSnapshot?: boolean
@@ -92,6 +90,7 @@ export default class Command extends Instrument {
 
     return this._isOpen || (this._isOpen === null
       && (
+        this.err?.isRecovered ||
         // command has nested commands
         (this.name !== 'session' && this.hasChildren && !this.event && this.type !== 'system') ||
         // command has nested commands with children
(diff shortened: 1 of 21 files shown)


packages/reporter/src/commands/command.tsx
```diff
@@ -258,6 +259,80 @@ interface Props {
...
 }
 
+const CommandDetails = observer(({ model, groupId, aliasesWithDuplicates }) => (
+  <span className={cs('command-info')}>
+    <span className='command-method'>
+      <span>
+        {model.event && model.type !== 'system' ? `(${displayName(model)})` : displayName(model)}
+      </span>
+    </span>
+    {!!groupId && model.type === 'system' && model.state === 'failed' && <StateIcon aria-hidden className="failed-indicator" state={model.state}/>}
```
emilyrohrbough: Note for Monday -- this is where the fail-indicator class came from -- verify you don't need to re-add `vertical-align: middle`

AtofStryker (commented):
Is the goal of `isRecovered` to eventually be used outside of `cy.session`? Does that mean session validation fails but the rest of the test continues?

emilyrohrbough:
> Is the goal of isRecovered to eventually be used outside of `cy.session`?

Not necessarily, this is just cleaning up how this is currently implemented. I could see this potentially being used in the future by other "workflow-like" commands but there are no plans ATM that I am away of to create others commands of this style. Think of this as a try-catch-finally or try-catch-try_something_else.

> Does that mean session validation fails but the rest of the test continues?

Yes, it does. In the context of `cy.session()` this only apply in this flow:
```restore session -> validate session -> FAIL validation -> recreate session -> validate session -> if pass continue to next command OR if fail, fail test```

**Review by AtofStryker — APPROVED** [review]
wonder if we want to name the `isRecovered` to something session specific until it is used more generically later? Either way changes look good on my end!

File: packages/reporter/src/commands/command-model.ts [hunk 2]
```diff
@@ -148,7 +149,14 @@ export default class Command extends Instrument {
...
 
-    this.err.update(props.err)
+    if (props.err) {
+      if (!this.err) {
+        this.err = new Err(props.err)
+      } else {
+        this.err.update(props.err)
+      }
+    }
+
```
mjhenkes, line 159: Would it make sense to share this logic with the attempt-model?

mjhenkes (approved):

PR #24286, opened 2022-10-18
chore: Migrate AssertionsOptions menu to Vue.

sainthkh:
- Closes #23187

### User facing changelog

### Additional details
- Why was this change necessary? => To remove legacy runner.
- Any implementation details to explain? => I used `floating-ui/dom` to create menu.

### Steps to test

1. expermentalStudio on
2. right click elements to open the menu.

N/A

### PR Tasks


packages/app/src/runner/studio/AssertionType.ce.vue
```diff
@@ -0,0 +1,119 @@
...
+  e.stopPropagation()
+}
+
+const _close = (e) => {
+  if (popperElement.value && popperElement.value.contains(e.relatedTarget)) {
+    return
+  }
+
+  isOpen.value = false
+  e.stopPropagation()
```
AtofStryker: might be a good opportunity to leverage [event modifiers](https://vuejs.org/guide/essentials/event-handling.html#event-modifiers) to bake into the template over the handler and would eliminate the need to pass in the event

packages/app/src/runner/studio/AssertionType.ce.vue
```diff
@@ -0,0 +1,119 @@
...
+const isOpen = ref(false)
+const hasOptions = props.options && !!props.options.length
+const popperElement: Ref<HTMLElement | null> = ref(null)
+
+const _open = (e) => {
+  isOpen.value = true
+  e.stopPropagation()
+}
+
+const _close = (e) => {
```
AtofStryker:
```suggestion
const onClose = (e) => {
```

packages/app/src/runner/studio/AssertionType.ce.vue
```diff
@@ -0,0 +1,119 @@
...
+  type: string
+  addAssertion: any
+  options: any
+}>()
+
+const isOpen = ref(false)
+const hasOptions = props.options && !!props.options.length
+const popperElement: Ref<HTMLElement | null> = ref(null)
+
+const _open = (e) => {
```
AtofStryker:
```suggestion
const onOpen = (e) => {
```

packages/app/src/runner/studio/AssertionType.ce.vue
```diff
@@ -0,0 +1,119 @@
...
+const _close = (e) => {
+  if (popperElement.value && popperElement.value.contains(e.relatedTarget)) {
+    return
+  }
+
+  isOpen.value = false
+  e.stopPropagation()
+}
+
+const _click = (e) => {
```
AtofStryker:
```suggestion
const onClick = (e) => {
```

packages/app/src/runner/studio/AssertionsMenu.ce.vue
```diff
@@ -0,0 +1,130 @@
...
+
+const props = defineProps <{
+  jqueryElement: any
+  possibleAssertions: any[]
+  addAssertion: any
+  closeMenu: () => void
+  highlightStyle: StyleValue
+}>()
+
+const _addAssertion = (...args) => {
```
AtofStryker: Do we usually prefix the method names with an underscore?

AtofStryker (changes requested):
not to worried about the stylistic things, but I think the code might be able to be cleaned up a bit with the event modifiers suggestion.

packages/app/src/runner/studio/AssertionType.ce.vue
sainthkh: Done.

packages/app/src/runner/studio/AssertionType.ce.vue
sainthkh: Done.

packages/app/src/runner/studio/AssertionType.ce.vue
sainthkh: Done.

packages/app/src/runner/studio/AssertionsMenu.ce.vue
sainthkh: I tried not to change much from the original React code. That's why there are so many leading underscores in the names. I think it's not our convention in Vue components. 

I removed them.

packages/app/src/runner/studio/AssertionType.ce.vue
sainthkh: Done.

packages/app/src/runner/studio/AssertionOptions.ce.vue
```diff
@@ -0,0 +1,98 @@
...
+  <div
+    ref="popper"
+    class="assertion-options"
+    :style="{ position: 'absolute' }"
+  >
+    <div
+      v-for="{ name, value } in options"
+      :key="`${name}${value}`"
+      class="assertion-option"
+      @click.stop="() => onClick(name, value)"
```
AtofStryker, line 10: I think you can just pass the function in like this
```suggestion
      @click.stop="onClick(name, value)"
```

packages/app/src/runner/studio/AssertionsMenu.ce.vue
```diff
@@ -0,0 +1,130 @@
...
+    class="assertions-menu"
+  >
+    <div class="header">
+      <div class="title">
+        <span>Add Assertion</span>
+      </div>
+      <div class="close-wrapper">
+        <a
+          class="close"
+          @click="onClose"
```
AtofStryker:
```suggestion
          @click.stop="onClose"
```

packages/app/src/runner/studio/AssertionsMenu.ce.vue
```diff
@@ -0,0 +1,130 @@
...
+  highlightStyle: StyleValue
+}>()
+
+const addAssertion = (...args) => {
+  args = _.compact(args)
+  props.addAssertion(props.jqueryElement, ...args)
+}
+
+const onClose = (event) => {
+  event.stopPropagation()
```
AtofStryker:
```suggestion
const onClose = () => {
```

AtofStryker (changes requested):
I think there is a bug in the mounter where the menu isn't actually displayed. Ideally we would have a cypress in cypress test to verify this works, but can understand the absence due to experimental

#### 10.10.0 release
10 10 0 assertions with studio


#### This PR
refactor branch

sainthkh:
I reproduced the bug and am now trying to fix it.

packages/app/src/runner/studio/AssertionType.ce.vue
```diff
@@ -0,0 +1,115 @@
...
+            fillRule="evenodd"
+            d="M4.646 1.646a.5.5 0 0 1 .708 0l6 6a.5.5 0 0 1 0 .708l-6 6a.5.5 0 0 1-.708-.708L10.293 8 4.646 2.354a.5.5 0 0 1 0-.708z"
+          />
+        </svg>
+      </span>
+    </div>
+    <AssertionOptions
+      v-if="hasOptions && isOpen"
+      :type="type"
+      :add-assertion="addAssertion"
```
lmiller1990: Rather that passing a function, I think it's more idiomatic to emit and event in Vue.

https://vuejs.org/guide/components/events.html

So

```html
<AssertionOptions
  @addAssertion="addAssertion"
/>
```

In `AssertionOptions.vue`

```
<script setup>
const emit = defineEmits<{
  (event: 'addAssertion', type: string): void
}>()

//

emit('addAssertion', "<type here>")
```

packages/app/src/runner/studio/AssertionType.ce.vue
```diff
@@ -0,0 +1,115 @@
...
+
+const isOpen = ref(false)
+const hasOptions = props.options && !!props.options.length
+const popperElement: Ref<HTMLElement | null> = ref(null)
+
+const onOpen = () => {
+  isOpen.value = true
+}
+
+const onClose = (e) => {
```
lmiller1990: `e: Event`.

Should give you better types.

packages/app/src/runner/studio/mounter.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+const unmountAssertionsMenu = () => {
+  if (app) {
+    app.unmount()
+    app = null
+  }
+}
+
+// TODO: remove these.
```
lmiller1990, line 71: Hmm, I wonder what's happening? I didn't look yet, but would be nice to figure this out.

Are you still coding on this or should I pull & test?

sainthkh:
1. Is Webkit failure flaky or valid?
2. I decided to use `popperjs` instead of `floating-ui`, it seems that it's a bit buggy.
3. I added test for showing the menu and submenu. As we now have `real` click events, we can test them.

AtofStryker:
> 1. Is Webkit failure flaky or valid?

There is some flake that we are working on cleaning up, so it could be related? I gave it a rerun to see if the outcome changes. My guess is flake.

> 2. I decided to use `popperjs` instead of `floating-ui`, it seems that it's a bit buggy.

I don't have any objections here. Seems like a good choice thats lightweight, well used, and gets the job done 💪🏻 .

> 3. I added test for showing the menu and submenu. As we now have `real` click events, we can test them.

Great! Going to take another look fairly soon 😃.

patches/@popperjs+core+2.9.2.dev.patch
```diff
@@ -1,91 +0,0 @@
...
--  if (checkMainAxis || checkAltAxis) {
--    var mainSide = mainAxis === 'y' ? top : left;
--    var altSide = mainAxis === 'y' ? bottom : right;
--    var len = mainAxis === 'y' ? 'height' : 'width';
--    var offset = popperOffsets[mainAxis];
--    var min = popperOffsets[mainAxis] + overflow[mainSide];
--    var max = popperOffsets[mainAxis] - overflow[altSide];
-+  function setOffset(axis) {
-+    var isMainAxis = axis === mainAxis;
-+    var mainSide = axis === 'y' ? top : left;
```
AtofStryker, line 18: Why did we need to edit the patch here?

packages/runner/package.json
```diff
@@ -13,29 +13,9 @@
...
-    "@packages/icons": "0.0.0-development",
-    "@packages/reporter": "0.0.0-development",
-    "@packages/web-config": "0.0.0-development",
-    "@popperjs/core": "2.9.2",
     "babel-plugin-prismjs": "1.0.2",
     "bluebird": "3.5.3",
-    "classnames": "2.3.1",
     "cross-env": "6.0.3",
-    "jquery": "3.1.1",
-    "lodash": "^4.17.21",
```
AtofStryker, line 31: my guess is these packages are no longer used and this is just clean up?

AtofStryker (commented):
I can get the sub menus to show up but when I click on the options nothing seems to happen? I didn't spend a whole lot of time debugging but It looks like there might be issues with the onClick in AssertionOptions? 
click_assertion

patches/@popperjs+core+2.9.2.dev.patch
sainthkh: This popper js is the bundled one from `react-popper`. As we don't need it any more, I removed it and added new popper 2.11.6. So, we don't need the patch for 2.9.2 version.

packages/runner/package.json
sainthkh: Right. We migrated most of the js and ts files to other packages and they're not used any more. 

I think it will save 1 or 2 seconds when `yarn install`.

sainthkh:
Peek 2022-11-02 09-25

I tested it on MDN site, but it works correctly like above. Can you please test it again?

packages/app/src/runner/studio/AssertionOptions.ce.vue
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  return str
+}
+
+const popper: Ref<HTMLElement | null> = ref(null)
+
+onMounted(() => {
+  nextTick(() => {
+    const popperEl = popper.value as HTMLElement
```
lmiller1990, line 55: Is this cast necessary? It shouldn't be?

If it's the null check you are concerned with, I thin we should be throwing an error here. If it's null, we end up doing `(popperEl.parentElement`) which will be `undefined.parentElement`, which is far harder to debug than a more explicit "XXX element was not found. This should be impossible".

packages/app/src/runner/studio/AssertionOptions.ce.vue
```diff
@@ -0,0 +1,92 @@
...
+</template>
+
+<script lang="ts" setup>
+import { createPopper } from '@popperjs/core'
+import { onMounted, ref, nextTick, Ref } from 'vue'
+
+const props = defineProps<{
+  type: string
+  addAssertion: any
+  options: any
```
lmiller1990: Is there any chance we can go find the correct types for this? If we are making a refactor with the goal of improving code quality, I really feel like we want to be as strict and tight on the type definitions are possible. We know what `options` can be, it should be `have.text.` etc? Those are all defined here https://github.com/cypress-io/cypress/blob/develop/packages/app/src/store/studio-store.ts#L61-L94

lmiller1990 (commented):
I reviewed this a few weeks ago but I just realized I did not hit "Submit Review" so all my comments were pending, sorry!

I submitted my review. Some of the comments might be a bit stale, since they were made a few weeks ago.

packages/app/src/runner/studio/AssertionType.ce.vue
sainthkh: Done. I also changed `setPopperElement` to event.

packages/app/src/runner/studio/AssertionType.ce.vue
sainthkh: I re-read the code again and tried to make them as type-safe as possible.

packages/app/src/runner/studio/AssertionOptions.ce.vue
sainthkh: `ref`s are set to `null` in the setup because we need an initial value. But when they're set to the elements in the `template`, they'll never be changed. So I used `cast`. I think it's really unlikely for them to be changed back to `null`.

packages/app/src/runner/studio/AssertionOptions.ce.vue
sainthkh: I re-read the code and defined types.

packages/app/src/runner/studio/mounter.ts
sainthkh: This weird block of code is the best solution I found. 

AFAIK, Vue or Cypress doesn't have a reason not to send events to DOM elements under the root element in shadow DOM.

I doubt it's a default browser behavior, but it's also weird for browsers to work in this way.

packages/app/src/runner/studio/AssertionOptions.ce.vue
```diff
@@ -0,0 +1,96 @@
...
+</template>
+
+<script lang="ts" setup>
+import { createPopper } from '@popperjs/core'
+import { onMounted, ref, nextTick, Ref } from 'vue'
+import type { AssertionOption } from './types'
+
+const props = defineProps<{
+  type: string
+  options: AssertionOption[]
```
lmiller1990, line 35: Nice

packages/app/cypress/e2e/studio/studio.cy.ts
```diff
@@ -255,4 +255,25 @@ it('visits a basic html page', () => {
     // Cypress in Cypress, it redirects us the the spec page, which is not what normally
     // would happen in production.
   })
+
+  it('shows menu and submenu correctly', () => {
```
lmiller1990, line 259: We've already got a test that opens the sub menu, don't we?

lmiller1990 (approved):

lmiller1990:
Gave it a test, seems like everything works as expected. I think we can remove the extra test -- there's another one in the same folder that asserts against the sub menu, isn't there?

packages/app/src/runner/studio/AssertionOptions.ce.vue
sainthkh: :+1:

packages/app/cypress/e2e/studio/studio.cy.ts
sainthkh: The focus of the test is a bit different. 

As we're relocating the UI div with popperjs, it is possible that the UI div is created correctly and can be clicked with our commands but they're not visible on the screen. 

It's the bug @AtofStryker wrote [above](https://github.com/cypress-io/cypress/pull/24286#pullrequestreview-1147926810) and this test tests not to reproduce that problem again.

lmiller1990:
Looks good - just need one more reviewer. Maybe @AtofStryker can take one last final pass (he was previously assigned).

mike-plummer (approved):
Looks good on my local. Awesome work! 💯

PR #24346, opened 2022-10-21
fix: Hovering over mount in command log does not show component in AUT

### LINKED ISSUE #24138: Hovering over mount in command log does not show the mounted component in AUT
## Current Behavior
When hovering over `mount` in the command log of a component test, the AUT shows a blank white view instead of the mounted component.

image

## Desired Behavior

The mount command shows the mounted component in the AUT

amehta265:
- Closes #24138 

### User facing changelog

Previously when hovering over `mount` in the reporter of a component test, the AUT shows a blank white view instead of the mounted component. This is not ideal if the user wants to see the component that has been mounted by the `mount` function.
This PR ensures that the snapshot of the mounted component is taken at the current time thereby guaranteeing component being displayed in the AUT

### Additional details

- Upon further investigation this issue only persists for the `vue2` and `react18` frameworks. For all other frameworks the intended behavior was happening i.e. The mount command shows the mounted component in the AUT.
- `React18` uses the `makeMountFn` from `react` under the hood therefore changes had to be made to the `mount` function inside `npm/react/createMount` for this framework.
- The proposed solution involves the use of the `.wait` and a `mutationObserver`. Using the `.wait(0, ...)` allows cypress to wait until the next tick in the event-loop to take a snapshot of the mounted component. This ensures that all the synchronous code (including the actual mounting of the component) has occurred before we attempt to take a snapshot. The `mutationObserver` listens to the target node (in our case the `data-cy-root`) to check if any element is added to it. If all the elements are added we can be sure that the component exists on the AUT and can thereby take the screenshot.

### Steps to test

The spec file `reporter-ct-mount-hover.cy.ts` iterates over several projects from `system-tests`, scaffolds, opens, and runs a spec file within each project while hovering on the mount function in the command log to ensure the intended functionality.

1. Go to the `packages/app`
2. Run `yarn cypress:open`
3. Select e2e testing on your preferred browser and navigate to the `reporter-ct-mount-hover.cy.ts` spec file.
4. This test iterates over all major frameworks testing hovering over the mount in the command-log ensuring that the component is visible in the AUT.

Before: 
Screen Shot 2022-10-21 at 3 34 50 PM
After:
Screen Shot 2022-10-21 at 3 29 19 PM

### PR Tasks


astone123 (approved):
I tested this with React 18 and Vue 2 and I can see snapshots as expected. I wonder why this was only an issue with React 18 and Vue 2. Any idea? 

Glad to see consistency between the different frameworks 🎉

marktnoonan:
@amehta265 I'm testing this out and don't see the update on this Vue 2 / Nuxt project, wonder if it renders content later than initial mount? Haven't dug in yet. But that might be expected for Nuxt.

nuxt mount

nuxt first command

astone123:
@marktnoonan how are you getting the changes to `npm/vue2` into your project? It was kind of tricky for me. I ended up building `npm/vue2` and then copying the `npm/vue2/dist` folder from my cypress repo into `node_modules/vue2` in my project.

marktnoonan:
@astone123 I was using the e2e test that Ankit set up. I'll try as you suggested!

marktnoonan:
OK that worked. And just building the npm/vue package fixed the example in the test. I'm going to take another look at the tests, as it probably shouldn't have been passing before I built npm/vue?

Screen Shot 2022-10-24 at 4 19 51 PM

packages/app/cypress/e2e/runner/reporter-ct-mount-hover.cy.ts
```diff
@@ -0,0 +1,46 @@
...
+    }),
+    it(`While hovering on Mount(), shows component on AUT for ${projectName}`, () => {
+      if (`${projectName}` === 'react18') {
+        cy.openProject(projectName, ['--config-file', 'cypress-vite.config.ts'])
+        cy.startAppServer('component')
+        cy.visitApp()
+        cy.contains(`${test}`).click()
+        cy.waitForSpecToFinish()
+        cy.get('.collapsible-header-inner:first').click().get('.command.command-name-mount > .command-wrapper').realHover().then(() => {
+          cy.get('iframe.aut-iframe').its('0.contentDocument.body').then(cy.wrap).within(() => {
```
marktnoonan, line 30: @amehta265 I think we need to add some assertions in here about what is found (or not) inside the iframe. Inside this `within`, we should make sure the cy-root element has some content.

marktnoonan (changes requested):
Requesting some additional test coverage before we merge to make sure we would catch a regression

packages/app/cypress/e2e/runner/reporter-ct-mount-hover.cy.ts
amehta265: I just added two more assertions. The first one ensures that the content body contains at least 2 children i.e. `[data-cy-root]` and the component itself. The second assertions ensures that we can find`[data-cy-root]` within the content body.

npm/vue2/src/index.ts
```diff
@@ -410,6 +412,17 @@ export const mount = (
 
     Cypress.vue = VTUWrapper.vm
     Cypress.vueWrapper = VTUWrapper
+  }).wait(0, { log: false }).then(() => {
+    if (optionsOrProps.log !== false) {
+      const observer = new MutationObserver(() => callbackFunc)
```
ZachJW34: Can you explain the benefit of using `MutationObserver` over something like `Vue.nextTick` here? The observer seems complex, and to add it I think we would need more safety around its use. For example, what happens if a component throws an error before rendering? Is the observer ever disconnected?

Same comment for React, I was a bit hesitant but we could use [act](https://reactjs.org/docs/test-utils.html) to let the component stabilize.

packages/app/cypress/e2e/runner/reporter-ct-mount-hover.cy.ts
marktnoonan: Hmm, these tests all still pass for me against the old code. There could be a few things going on.

- the component gets mounted into the `data-cy-root` element, so I think we need to check its children, siblings inserted beside it in `0.contentDocument.body` are likely elements that do indicate the component mounted, but they aren't the component.

- I'm not sure about the realHover, it seems I get different results against the old code when I use realHover vs just `click` to pin the snapshot ... Real hover passes on the old code, even when I update the test to check for children of `data-cy-root`. Pinning the snapshot with `Click` fails as expected on the old code and passes on the new code. 

Is it possible realHover is not having the intended effect and we are capturing the end state of the test, not the snapshot from hovering the command? If it's more reliable, even though the ticket was for hover, I think it's safe to say that if the pinned snapshot has the correct DOM content, the hover is fine, so you might be able to switch to click.

To switch between versions of the mount command, I am running `yarn build-prod` on packages/vue2 in develop, and then switching to this branch and running the new e2e test against that old version of the code, and I've removed all but `{ projectName: 'nuxtjs-vue2-configured', test: 'Tutorial.cy' }` from `PROJECTS`.

marktnoonan: Here's my variation that fails on the old code and passes on the new: 

```js
 cy.get('.command.command-name-mount > .command-wrapper').click().then(() => {
          cy.get('iframe.aut-iframe').its('0.contentDocument.body').then(cy.wrap).within(() => {
            cy.get('[data-cy-root]').children().should('have.length.at.least', 1)
          })
        })
```

It's probably possible to debug the hover some more - maybe the assertion about children runs before Cy has had a chance to react to the hover and update the DOM & we just need to wait for something.

The main thing is to get these tests to fail against the old code for the commands you fixed.

npm/vue2/src/index.ts
amehta265: I believe the implementation using the `MutationObserver` here was to generalize the solution across different frameworks. As soon as we get the children for the `data-cy-root`, the observer is disconnected as you will see in the `callbackFunc`.
I can see why we can implement framework specific solutions because only `vue2` and `react` are impacted so I'll definitely look into the methods mentioned.

packages/app/cypress/e2e/runner/reporter-ct-mount-hover.cy.ts
amehta265: As discussed, I will use `click()` instead of `realHover()` or `trigger('mouseover')` as they are not testing the intended functionality. As you mentioned, the tests seems to be failing for the old code when using `click()` which should be the intended outcome as the fix is not implemented in the old code.
As far as asserting the children of `data-cy-root`, this will not always work because some frameworks such as `angular `do not mount the component under test as a child of `data-cy-root`. Therefore, I will use your recommended approach and tweak the test for `angular` frameworks.

npm/react/src/createMount.ts
```diff
@@ -54,6 +54,26 @@ export const makeMountFn = (
...
+      $el: (el.children.item(0) as unknown) as JQuery<HTMLElement>,
+      consoleProps: () => {
+        return {
+          // @ts-ignore protect the use of jsx functional components use ReactNode
+          props: jsx.props,
+          description: type === 'mount' ? 'Mounts React component' : 'Rerenders mounted React component',
+          home: 'https://github.com/cypress-io/cypress',
+        }
+      },
+    }).snapshot('mounted').end()
```
emilyrohrbough: Did you try removing `.snapshot('mounted').end()`? When the command ends, the driver will snapshot automatically (unless the log's attributes are `snapshot: false`) and the command queue will end the log.

amehta265: Unfortunately this did not work. I don't have a good understanding about the driver snapshot and command queue so I don't understand why your proposed solution wouldn't work. The problem I believe is when we hover on the `mount()` command, the `data-cy-root` has not loaded completely and therefore the component does not show up mounted on the AUT.

packages/app/cypress/e2e/runner/reporter-ct-mount-hover.cy.ts
amehta265: I have updated the tests. All `vue2` and `react18` tests fail when using the old code while all tests pass with the updated `mount` in the current branch.

marktnoonan (approved):

npm/vue2/src/index.ts
amehta265: As discussed, I have used `Vue.nextTick()` in `npm/vue2` to ensure the hover on mount command functionality for vue2 projects and have used `wait(fn, 0)` in `npm/react`

packages/app/cypress/e2e/runner/reporter-ct-mount-hover.cy.ts
```diff
@@ -0,0 +1,51 @@
+import type { fixtureDirs } from '@tooling/system-tests'
+
+type ProjectDirs = typeof fixtureDirs
+
+const PROJECTS: {projectName: ProjectDirs[number], test: string}[] = [
+  { projectName: 'angular-15', test: 'app.component' },
```
ZachJW34: Let's change this to `angular-14`, Angular 15 is in beta so it shouldn't be the version we test our stable versions against

npm/vue2/src/index.ts
```diff
@@ -410,7 +412,7 @@ export const mount = (
 
     Cypress.vue = VTUWrapper.vm
     Cypress.vueWrapper = VTUWrapper
-  })
+  }).wait(0, { log: false })
```
ZachJW34: Do we need the wait here?

amehta265: No we do not. An oversight on my end from the previous implementation. This has been removed in the latest commit.

packages/app/cypress/e2e/runner/reporter-ct-mount-hover.cy.ts
amehta265: This has been updated in the latest commit

npm/vue2/src/index.ts
```diff
@@ -363,10 +363,12 @@ export const mount = (
   })
   .then((win) => {
     if (optionsOrProps.log !== false) {
-      Cypress.log({
-        name: 'mount',
-        message: [message],
-      }).snapshot('mounted').end()
+      Vue.nextTick(() => {
```
ZachJW34: I'd expect the `nextTick` to be wrapped within the chain so as to ensure the tick and log is captured before the next Cypress command is fired. You could convert this function to be async and `await Vue.nextTick()` as the end.

lmiller1990: I am not sure we need async/await - could just do it like this, right?

```js
.then((win) => Vue.nextTick())
.then(() => {
  /* stuff */
})
```
____

If anyone was curious, under-the-hood `nextTick` is basically just `Promise.resolve` (with various fallbacks). It actually uses `MutationObserver` as a fallback, interestingly enough.

packages/app/cypress/e2e/runner/reporter-ct-mount-hover.cy.ts
```diff
@@ -0,0 +1,51 @@
...
+  { projectName: 'angular-14', test: 'app.component' },
+  { projectName: 'vueclivue2-configured', test: 'HelloWorld.cy' },
+  { projectName: 'react-vite-ts-configured', test: 'App.cy' },
+  { projectName: 'react18', test: 'App.cy' },
+  { projectName: 'create-react-app-configured', test: 'App.cy' },
+  { projectName: 'vueclivue3-configured', test: 'HelloWorld.cy' },
+  { projectName: 'nuxtjs-vue2-configured', test: 'Tutorial.cy' },
+]
+
+for (const { projectName, test } of PROJECTS) {
```
lmiller1990, line 16: Tests are very thorough, and I know this is already :heavy_check_mark: , but I wonder if we want to consider adding these to an existing test in the future? Scaffolding all these projects is pretty expensive (in terms of CI time) for something that could be exercised in one of the other tests we've got for the various frameworks.

npm/vue2/src/index.ts
ZachJW34: As long as the `.then(() => Vue.nextTick())` is done after the component creation I'm cool with it!

amehta265: This sounds good. I've made the necessary changes on the latest commit

lmiller1990 (approved):

lmiller1990:
Looks good, but CI has some fails - might want to verify if those are real fails (and fix it) or flake (if so - probably want to make sure we aren't merging flaky tests).

PR #24364, opened 2022-10-24
fix: Update testIsolation types definition

mikstime:
### User facing changelog
Fixed documentation typo

### PR Tasks

### DIFF
--- a/cli/types/cypress.d.ts
+++ b/cli/types/cypress.d.ts
@@ -2808,7 +2808,7 @@ declare namespace Cypress {
      */
     supportFile: string | false
     /**
-     * The test isolation level applied to ensure a clean slate between tests.
+     * The test isolation level applied to ensure a clean state between tests.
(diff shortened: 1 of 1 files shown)


cli/types/cypress.d.ts
```diff
@@ -2808,7 +2808,7 @@ declare namespace Cypress {
      */
     supportFile: string | false
     /**
-     * The test isolation level applied to ensure a clean slate between tests.
+     * The test isolation level applied to ensure a clean state between tests.
```
emilyrohrbough, line 2811: @mikstime Thanks for the PR. Technically this wasn't a typo, I intentionally used the word slate here, but don't have a strong preference on the verbiage of slate or state. Both convey the same meaning.

emilyrohrbough (approved):

AtofStryker (approved):

cli/types/cypress.d.ts
mikstime: Well in that case i'll drop it

PR #24367, opened 2022-10-24
fix: Fix sourcemaps when using cy.origin() dependencies

### LINKED ISSUE #23365: Fix source maps / code frame for cy.origin with dependencies
### What would you like?

Currently, if there's a test failure inside the `cy.origin()` callback and it requires dependencies, the error code frame shows the object the callback is replaced with and not the original source code. We need to connect the extracted callback with the source code through source maps and ensure the code frame shows the original source.

The following code:

```js
  Cypress.require('./does-not-exist')
})
```

yields the following error/code frame:

Image

### Why is this needed?

It's important for debugging failures to see the original source code. It's confusing to see this artifact the source callback is replaced with instead of the callback itself.

chrisbreiding:
- Closes #23365

### User facing changelog

- When utilizing dependencies within the `cy.origin()` callback, errors now show the correct code frame

### Additional details

Ensures a new source map is generated when the webpack-preprocessor changes the test code, fixing what the code frame displays.

### Steps to test

Run [this test](https://github.com/cypress-io/cypress/blob/4e667e5383a4df482756f3b1b3d572c3a97ac7df/packages/driver/cypress/e2e/e2e/origin/dependencies.cy.ts#L7-L13
), but change the assertion to fail:

```js
  it('works with require()', () => {
    cy.origin('http://www.foobar.com:3500/', () => {
      const lodash = require('lodash')

      expect(lodash.get({ foo: 'foo' }, 'foo')).to.equal('bar')
    })
  })
```

The code frame should point to `cy.origin` and show the original source code.

Before

Image

After

Screen Shot 2022-10-24 at 11 15 20 AM

### PR Tasks


npm/webpack-preprocessor/patches/merge-source-map+1.1.0.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/node_modules/merge-source-map/index.js b/node_modules/merge-source-map/index.js
```
chrisbreiding, line 1: This dependency works perfectly well once this small change is made.

system-tests/__snapshots__/cy_origin_error_spec.ts.js
```diff
@@ -1,75 +0,0 @@
-exports['e2e cy.origin errors / captures the stack trace correctly for errors in cross origins to point users to their "cy.origin" callback'] = `
```
chrisbreiding, line 1: Removed this snapshot since we already test the stdout specifically for what we're looking for and it doesn't provide any extra value.

system-tests/projects/integration-outside-project-root/e2e/failing.cy.js
```diff
@@ -13,11 +13,11 @@ context('validation errors', function () {
     Cypress.config('isInteractive', true)
```
chrisbreiding, line 13: This and the next few files just needed to update to the new fail/verify API.

npm/webpack-preprocessor/package.json
```diff
@@ -26,6 +27,8 @@
     "loader-utils": "^2.0.0",
     "lodash": "^4.17.20",
     "md5": "2.3.0",
+    "merge-source-map": "^1.1.0",
```
emilyrohrbough: Why did you decide to use `merge-source-map` vs a more recently updated dependency? Looks like this isn't being maintained.

Also, are we still only support dependency in cy.origin when using webpack? I thought recent changes removed the need to use the custom API / webpack logic?

yarn.lock
```diff
@@ -26252,7 +26259,7 @@ pascalcase@^0.1.1:
   resolved "https://registry.yarnpkg.com/pascalcase/-/pascalcase-0.1.1.tgz#b363e55e8006ca6fe21784d2db22bd15d7917f14"
   integrity sha1-s2PlXoAGym/iF4TS2yK9FdeRfxQ=
 
-patch-package@6.4.7:
+patch-package@6.4.7, patch-package@^6.4.7:
```
emilyrohrbough: Can we unify the patch-package version?

npm/webpack-preprocessor/package.json
chrisbreiding: > Why did you decide to use merge-source-map vs a more recently updated dependency? Looks like this isn't being maintained.

Do you have any suggestions? This is the only I could find that works.

> Also, are we still only support dependency in cy.origin when using webpack? I thought recent changes removed the need to use the custom API / webpack logic?

Yes, it still requires the webpack preprocessor. The recent changes only removed the need for `Cypress.require()`.

yarn.lock
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/24367/commits/5575c8a4d87ab7d6229ba2216fd41931d3814174

npm/webpack-preprocessor/lib/cross-origin-callback-loader.ts
```diff
@@ -155,22 +157,20 @@ export default function (source: string, map, meta, store = crossOriginCallbackS
...
   }
 
-  debug('callback with original source')
-  // if no requires/imports were found, callback with the original source/map
-  this.callback(null, source, map)
+  // if we found requires/imports, re-generate the code from the AST
+  const result = generate(ast, { sourceMaps: true }, {
+    [resourcePath]: source,
+  })
+  const newMap = mergeSourceMaps(map, result.map)
```
mjhenkes, line 176: What happens if you just return the new source map?

chrisbreiding: The new source map is based off the code at that point in the webpack process. It's an intermediary version of the source. So if the original is TypeScript, it might through ts-loader, then babel, then get to this loader as some es5 version of the code without TypeScript. Since it's not the original source, it results in stack traces and code frames being incorrect. The original source map needs to be merged with the one generated by the changes made in this loader for it the resulting one to be correct.

npm/webpack-preprocessor/package.json
mjhenkes: Since we're patching this, why don't we lock it down and don't pull it in optimistically.

chrisbreiding: > Since we're patching this, why don't we lock it down and don't pull it in optimistically.

Addressed in https://github.com/cypress-io/cypress/pull/24367/commits/844950ab68737202abbb5491cc248bd705ae06f1

npm/webpack-preprocessor/lib/cross-origin-callback-loader.ts
emilyrohrbough: @chrisbreiding minding adding some of this insight in as a comment for future devs? Sources maps are always tricky to finally get the hang of.

chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/24367/commits/78b97468b902f1687b72326924448d8bd1b7bc8a

emilyrohrbough:
These changes overall seem good. I haven't approved this because I'm hesitant about adding `merge-source-map` with a known issue we are trying to patch in a `postinstall`. Have we done postinstall patches in npm modules we've released before? Have we had any issues with dependency resolution? Is there risk this post install should manipulate the wrong package if multiple versions of the @cypress/webpack-preprocess are in the dependency tree? It  doesn't seem like the logic here is overwhelming complex. Should we just implement our own? Curious what others think.

chrisbreiding:
> Have we done postinstall patches in npm modules we've released before? 

Yes, the `cli` has patches.

> Have we had any issues with dependency resolution? Is there risk this post install should manipulate the wrong package if multiple versions of the @cypress/webpack-preprocess are in the dependency tree? 

No, the `postinstall` will only run for the package it's defined in.

> Should we just implement our own? 

I don't see any value in doing that. It's more straightforward to keep it as a dependency than to copy the code, attribute it, and maintain it ourselves. We'll realistically never touch the code again. While it's very unlikely `merge-source-map` will get any updates, it's more likely than us updating it, and it would be easier to upgrade in the rare case it does get updated.

mjhenkes (approved):

emilyrohrbough (approved):

PR #24369, opened 2022-10-24
fix: normalize vite config resolution

### LINKED ISSUE #24133: Normalize DevServer Config Resolution
### Current behavior

The config resolution for `@cypress/webpack-dev-server` and `@cypress/vite-dev-server` differ.

`@cypress/vite-dev-server` will always source a `<project-root>/vite.config.js` and merge the properties with the `devServer.viteConfig` value if supplied.

`@cypress/webpack-dev-server` will only source a `<project-root>/webpack.config.js` if and only if `devServer.webpackConfig` is undefined.

### Desired behavior

The config resolution should be equivalent across dev-servers. I'd argue the `@cypress/webpack-dev-server` is the proper config resolution, as it allows the user to opt out of Cypress trying to merge in their project root config (which might be a config they don't want to use for CT testing).

### Test code to reproduce

Run `DEBUG=cypress:vite-dev-server:resolve-config npx cypress open` on a Vite based project that has a `<project-root>/vite.config.js`. Look in the terminal after clicking on CT testing in the launchpad and the resolved config will be logged. Close Cypress and edit the `cypress.config.js -> component.devServer.viteConfig` property (add a plugin, change port). The changes made will be merged with the properties defined in `<project-root>/vite.config.js`.

Run `DEBUG=cypress:webpack-dev-server:makeWebpackConfig npx cypress open` on a Webpack based project that has a `<project-root>/webpack.config.js`. Follow similar steps described above. Notice that after adding a `cypress.config.js -> component.devServer.webpackConfig`, the contents of `<project-root>/webpack.config.js` are no longer merged.

https://user-images.githubusercontent.com/25158820/194119332-f72cc2dd-2f4a-4a3e-bc82-b689b3999d3c.mov

__NOTE__: This would be a breaking change

ZachJW34:
- Closes #24133 

### User facing changelog
`vite.config.{js,ts}` will no longer be sourced when `component.devServer.viteConfig` is provided in the project's Cypress config

### Additional details
This change brings the vite-dev-server config resolution inline with webpack-dev-server which was deemed to be the proper resolution of the config. With this change, we will no longer source the project's Vite config if they have provided a `viteConfig` property in their Cypress `component.devServer` config.

The previous behavior was confusing when dealing with edge-cases. If a user happened to `require` their Vite Config in there Cypress config and returned the full config, we would still source their vite config which could potentially end up double merging and duplicating plugins.

### Steps to test
See the `Test code to reproduce` in the original issue.

https://user-images.githubusercontent.com/25158820/197627366-a42b6123-9fc2-4f43-bce6-79aa95fe90c0.mov

NOTE: Before merging, during squash add `BREAKING CHANGE` to footer message

### PR Tasks

TODO: Figure out what branch to target, this will be a breaking change but `v11` branch already exists and we don't want those changes until v12.


packages/app/cypress.config.ts
```diff
@@ -21,19 +21,6 @@ export default defineConfig({
     devServer: {
       bundler: 'vite',
       framework: 'vue',
-      viteConfig: {
-        optimizeDeps: {
```
ZachJW34, line 25: [Vite 3.0](https://vitejs.dev/blog/announcing-vite3.html#dev-improvements) made many improvements for handling unoptimized deps. The problem this was fixing was reloads that would occur mid-test when Vite was scanning deps. From my testing, this no longer happens and we can safely remove this list.

npm/vite-dev-server/src/resolveConfig.ts
```diff
@@ -70,49 +102,27 @@ export const createViteDevServerConfig = async (config: ViteDevServerConfig, vit
...
       },
       host: '127.0.0.1',
+      // Disable file watching and HMR when executing tests in `run` mode
+      ...(isTextTerminal
+        ? { watch: { ignored: '**/*' }, hmr: false }
+        : {}),
     },
     plugins: [
       Cypress(config, vite),
-      CypressInspect(config),
```
ZachJW34, line 89: A user can push this plugin in themselves, doesn't really make sense for us to support this, especially since it's use is not documented.

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -49,16 +45,7 @@ export const CYPRESS_WEBPACK_ENTRYPOINT = path.resolve(__dirname, 'browser.js')
  */
 function modifyWebpackConfigForCypress (webpackConfig: Partial<Configuration>) {
   if (webpackConfig?.plugins) {
-    webpackConfig.plugins = webpackConfig.plugins.filter((plugin) => {
-      if (removeList.includes(plugin.constructor.name)) {
-        /* eslint-disable no-console */
-        console.warn(`[@cypress/webpack-dev-server]: removing ${plugin.constructor.name} from configuration.`)
```
ZachJW34, line 55: We modify the webpack config pretty heavily but don't console log for every tweak we make. Removing this log so as to clean up the terminal output.

npm/webpack-dev-server/src/makeWebpackConfig.ts
```diff
@@ -183,12 +146,5 @@ export async function makeWebpackConfig (
 
   debug('Merged webpack config %o', mergedConfig)
 
-  if (process.env.WEBPACK_PERF_MEASURE) {
```
ZachJW34, line 186: We don't have this documented, and it doesn't work for many Webpack 5 apps so it's getting removed.

npm/vite-dev-server/src/resolveConfig.ts
```diff
@@ -4,52 +4,84 @@
...
     }
+
+    debug('Resolved config file at', configFile, 'using root', projectRoot)
+
+    resolvedOverrides = { configFile }
   }
 
+  const finalConfig = vite.mergeConfig(
+    resolvedOverrides,
+    makeCypressViteConfig(config, vite),
```
ZachJW34, line 56: This used to be swapped, but we don't want users overriding our base config. This makes it so our vite config takes precedence (same as how we handle Webpack config in webpack-dev-server).

astone123 (approved):
I tested this and can confirm that if I provide a `viteConfig` property in my Cypress config that my `vite.config.ts` file is not merged into the resolved config 👍🏻

rockindahizzy (approved):
👍

PR #24374, opened 2022-10-25
feat: Add Angular CT Schematic

### LINKED ISSUE #22976: Add Angular CT Schematic
## Part 1
It would be great for users of Angular Component Testing to be able to use the angular CLI to generate component tests for them. We do something similar with the Cypress Angular Schematic for `e2e`:

```bash
ng generate @cypress/schematic:e2e
```

I would like to propose the same functionality but for component testing:

```bash
ng generate @cypress/schematic:ct button
```

This would essentially generate a file named `button.component.cy.ts` that looks like the following:

```ts
import { ButtonComponent } from './button.component'

describe('ButtonComponent', () => {
  it('can mount', () => {
    cy.mount(ButtonComponent)
  })
})
```

## Part 2
By having this, we could then also wrap the current Angular CLI component generator to also call `@cypress/schematic:ct` as well. This way users who use the angular CLI to generate a component would get the following files

```bash
ng generate component button
```

### Scaffolded Files:

- `button.component.ts`
- `button.component.css`
- `button.component.cy.ts`

This would also require a schematic that would also setup your Angular CLI project to use this schematic. Something that would look like the following:

```bash
ng add cypress-angular-ct
```

To increase adoption of Angular CT support we could also throw this as a prompt at the end of the Cypress Launchpad to ask users if they want to add this schematic or something like that

### Why is this needed?

This will improve the developer experience for users by making the generating of CT specs super simple and ultimately an after thought if we wrap around the `ng generate component` schematic. By implementing this we will increase users adoption of Angular CT and improve over DX by doing all the boilerplate work for  developers. 

### Other

_No response_

w0wka91:
- Closes https://github.com/cypress-io/cypress/issues/22976

### Additional details
Why was this change necessary? What is affected by this change?

This change enables the user to generate a new component including a component spec file. By implementing this we increase users adoption of Angular CT and improve over DX by doing all the boilerplate work for developers.
This change also adds @cypress/schematic to the default schematics. This enables the user to use all of the commands without prefixing them with the package name.

### Steps to test
Generate a new Angular Application and add @cypress/schematic with `ng add <path/to/cypress/schematic>`

Before the user had to execute the following commands to generate a component and a component spec file:
```
ng generate component button
ng generate @cypress/schematic:spec button --component
```
Now the user can just execute this command:
`ng generate component button`
Also, the user can now use any @cypress/schematic generate command without prefixing it with the package name:
So this `ng generate @cypress/schematic:spec {name}`
becomes this:  `ng generate spec {name}`

### PR Tasks

- [ X] Have tests been added/updated?
- [ X] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [ X] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [ X] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?


jordanpowell88 (changes requested):
This PR looks great! I would love to see a cypress tests that validates all of this though [here](https://github.com/cypress-io/cypress/tree/develop/npm/cypress-schematic/src).

w0wka91:
Just added a cypress test to the PR, not sure if this is enough tho. It generates a new component and executes the component test which was generated so I think it covers all of the implemented behaviour:
1. Wrap the original `ng generate component` and generate a new component including a component spec file
2. Execute `@cypress/schematic` without prefixing.

jordanpowell88 (approved):

jordanpowell88:
Thank you @w0wka91. Looks good to me. I'm excited to have this feature enabled.

npm/cypress-schematic/src/schematics/ng-add/index.spec.ts
```diff
@@ -26,6 +28,10 @@ describe('@cypress/schematic: ng-add', () => {
     skipPackageJson: false,
   }
 
+  const readAngularJson = (tree: UnitTestTree) => {
+    return tree.readJson('/angular.json') as JsonObject
```
lmiller1990, line 32: Quick q, it is a reasonable assumption that `angular.json` exists in this location? I made a new project with Nx using Angular and they put Cypress tests in a separate project directory to the main App, meaning `angular.json` might be somewhere else.

cc @jordanpowell88, I think we talked about something similar to this before

lmiller1990 (commented):
One quick question!

npm/cypress-schematic/src/schematics/ng-add/index.spec.ts
jordanpowell88: @lmiller1990 This generator is for angular CLI projects (which all have an `angular.json` file). In your Nx example they already have this functionality

npm/cypress-schematic/src/schematics/ng-generate/component/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export default function (options: any): Rule {
+  return (_: Tree, _context: SchematicContext) => {
+    return chain([
+      externalSchematic('@schematics/angular', 'component', {
+        ...options,
+        skipTests: true,
+      }),
+      (tree: Tree, _context: SchematicContext) => {
+        const componentPath = tree.actions.filter((a) => a.path.includes(`${options.name}.component.ts`))
```
ZachJW34: This logic doesn't work when trying to generate a nested component e.g. `ng g c components/hello`. The component is generated, but a cypress test is not. This is because
```js
a = 'components/hello/hello.component.ts'
a.path.includes('components/hello.component.ts') === false
```

Is there a way to get the list of pending changes from the tree? Then we would just have to find a file ending in `.component.ts`

ZachJW34 (changes requested):

npm/cypress-schematic/src/schematics/ng-generate/component/index.ts
ZachJW34: I pushed up [dd2cf7a](https://github.com/cypress-io/cypress/pull/24374/commits/dd2cf7a87e4ce277bf2ab4a6190a0b457282c3bc) which fixes `ng g c foo/bar` example. Take a look and lmk if the change makes sense and then we can get this merged!

ZachJW34 (approved):

npm/cypress-schematic/src/schematics/ng-generate/component/index.ts
lmiller1990: Seems like this is failing on CI, taking a look...

lmiller1990: Is this supposed to generate a `cy.ts` file? It does not seem to work - tried locally. It does make a `spec` file, but I don't think that's the goal here, it is?

lmiller1990 (changes requested):
The latest changes seem to be breaking CI, I tried to fix it but I'm actually not clear on what's supposed to be generated - my understanding is a `cy.ts` file with `cy.mount`, but this is not happening.

I tried to debug this, it looks like it points to `ng-component/generators/index.ts` that has:

```ts
 externalSchematic('@schematics/angular', 'component', {
        ...options,
        skipTests: true,
      }),
```

I have rebased and resolved the conflicts w/ develop.

Also tested, not working for me?

```
$ ng generate component baz
CREATE src/app/baz/baz.component.scss (0 bytes)
CREATE src/app/baz/baz.component.html (18 bytes)
CREATE src/app/baz/baz.component.spec.ts (578 bytes)
CREATE src/app/baz/baz.component.ts (264 bytes)
UPDATE src/app/app.module.ts (524 bytes)
```
No `cy.ts` is generated.

npm/cypress-schematic/src/schematics/ng-add/index.ts
```diff
@@ -306,6 +307,33 @@ function modifyAngularJson (options: any): Rule {
...
+  return (tree: Tree, _: SchematicContext) => {
+    if (tree.exists('./angular.json')) {
+      const angularJsonVal = getAngularJsonValue(tree)
+      const angularSchematic = '@schematics/angular'
+      const cli = {
+        ...angularJsonVal.cli,
+        schematicCollections: ['@cypress/schematic', ...angularJsonVal?.cli?.schematicCollections ?? []],
+      }
+
+      if (cli.schematicCollections.indexOf('@schematics/angular') === -1) {
```
ZachJW34, line 320: `schematicCollections` is only for Angular-14 apps. Previously, the value was `defaultCollection` but it isn't an array. Should we override the default value instead for Angular 13 apps?

npm/cypress-schematic/src/schematics/ng-generate/component/index.ts
ZachJW34: It's passing on Angular-14 but not on Angular-13. I believe it is due to a problem with module resolution, similar to what we've seen with system-tests where it is resolving Angular deps from within the monorepo and not the scaffolded project.

lmiller1990:
Looks like we've got a real CI failure.

ZachJW34:
@w0wka91 I think it makes sense to not change the `defaultCollection` as a user might have a value here and changing it could break their schematic invocations. Pushing to the `schematicCollections` is fine since it supports multiple values. WDYT?

w0wka91:
@ZachJW34 Makes sense. So I would suggest we append the defaultCollection (if it it not null) to the schematicCollections array. What do you think?

ZachJW34:
@w0wka91 I think we drop support for Angular 13 for this version. I don't think schematics should support multiple ranges. I tried implementing what I suggested and any schematics invoked via `ng g @cypress/schematic` would not work since this package has dependencies on Angular 14. When executing the schematic, the package would resolve `v14` of angular inside the context of an Angular 13 app.

@jordanpowell88 is it cool to drop support for v13 for this schematic? We could support it but we'll have to manage our deps differently.

ZachJW34:
@w0wka91 I chatted with @jordanpowell88 a bit and pushed up a commit that removes support for v13. The component schematic was not working with v13 due to misconfigured dependencies, and furthermore the component schematic adheres to the v14 of the angular component schematic (standalone) so it wouldn't work regardless of the deps issue. Hoping for a green build!

ZachJW34 (approved):

ZachJW34:
@lmiller1990 Can you re-review? If it looks good and we merge, we should include a `BREAKING CHANGE` in the footer of the commit message since we are dropping support for Angular v13 for this schematic package.

lmiller1990:
Sure thing, main question: it looks like Angular 13 still has a couple hundred k installs / week: https://www.npmjs.com/package/@angular/core?activeTab=versions. It is two majors behind now, though, and my understanding is Angular culture is to upgrade to the latest versions pretty quickly.

Are we okay with breaking the schematic in a Cypress minor? I am not sure if we've got an official cadence around this. Or is there some way users on Angular 13 could still use this older version?

ZachJW34:
Ideally I think the `@cypress/schematic` should shadow the release of Angular. Users could still install older versions of the schematic that would correspond to their Angular version. This schematic package isn't included in the binary/cli so we can have breaking changes to the package without affecting the cypress app itself.

With the addition of the `ng g @cypress/schematic:component` schematic, the schema (options) for the schematic rely on the Angular component schematic since the cypress component schematic calls through to the Angular schematic. When the contract changes (which is the case for the v13 to v14 upgrade introducing the standalone option), this package needs to update to accommodate. @jordanpowell88 thoughts?

The best user experience is to have the cypress component schematic match 1:1 with the Angular schematic which we can only do by dropping support for older versions if the contract changes. We could support multiple versions but we would lose this experience, specifically when running `ng g @cypress/schematic:component --help` as it would show only a subset of the available options due to use tailoring the inputs to conform across multiple interfaces.

lmiller1990 (approved):

lmiller1990:
Sounds good. Let's merge this up. A release went out just now, sow this will be in the next release.

PR #24396, opened 2022-10-25
feat: add support for next 13

ZachJW34:
### User facing changelog
Support Next.js v13

### Additional details
Bring our current support for Next.js 11 + 12 to v13. No support for `turbopack` or React SC (yet).

### Steps to test
- Run `npx create-next-app@latest my-app`
- cd into app and run `npm i -D cypress`
- Check out this branch, run `yarn && yarn cypress:open`, choose `my-app`
- Walk through setup
- Add a test, I added `pages/index.cy.tsx` and run it

```ts
import Home from './index'

it('should render', () => {
  cy.mount(<Home />)
})
```

https://user-images.githubusercontent.com/25158820/200433321-c23a3164-93d5-4c18-83a2-223bb74022cb.mov

### PR Tasks


theoludwig:
Any way we can help? Is this can be prioritized?
Thanks! :smile:

ZachJW34:
The check for React 18 has been removed [recently](https://github.com/vercel/next.js/pull/42141), need to verify if our internal check needs to stick around. Might need it to support 13.0.0 (no longer needed in 13.0.1)

theoludwig:
> The check for React 18 has been removed [recently](https://github.com/vercel/next.js/pull/42141), need to verify if our internal check needs to stick around. Might need it to support 13.0.0 (no longer needed in 13.0.1)

Awesome! I just tested, and indeed, I can confirm that updating Next.js (v13.0.2), works!

astone123:
@ZachJW34 looks like everything was working fine with Next 13 except for the dependency version warning. So this PR is mainly just adding 13 in the versions list and adding a system test, right?

ZachJW34:
@astone123 It wasn't working for Next.js v13.0.0, but v13.0.1 + didn't need any source code changes (other than the deps warning). Good to be safe, and the `hasReactRoot` property is something we missed for v12 so makes sense to keep in.

astone123 (approved):

npm/webpack-dev-server/cypress/e2e/next.cy.ts
```diff
@@ -1,7 +1,7 @@
 /// <reference path="../support/e2e.ts" />
 import type { ProjectFixtureDir } from '@tooling/system-tests/lib/fixtureDirs'
 
-const WEBPACK_REACT: ProjectFixtureDir[] = ['next-11', 'next-12', 'next-11-webpack-4', 'next-12.1.6']
+const WEBPACK_REACT: ProjectFixtureDir[] = ['next-11', 'next-12', 'next-11-webpack-4', 'next-12.1.6', 'next-13']
```
lmiller1990: While we are here, are we able to drop next-12? Just to keep things nice and fast.We've got next-12.1.6, should cover us. Even if next-12.0.0 isn't working, we could just ask people to update to 12.1.6.

Alternatively, we could drop next-11 (two majors behind now). I am not sure if we actually landed somewhere on EOLing official support for majors, though.

Understand that the webpack resolution is different between next-12 and next-12.1.6, so if you think it's not worth the risk, that's fine.

lmiller1990 (approved):
Code seems fine, one comment/question - I think we need to get CI under control, it's getting slower and slower, any chance to reduce the amount of projects?

If not, all good, let's keep it in mind for other PRs.

npm/webpack-dev-server/cypress/e2e/next.cy.ts
ZachJW34: I'm cool with dropping the `next-11`, at least from our system-tests.

PR #24415, opened 2022-10-26
fix: remove dependence on @cypress/<dep> types

### LINKED ISSUE #24330: CT mount types relying on `@cypress/` don't work
### Current behavior

`cypress/react18` has types defined in `node_modules/cypress/react18/dist/index.d.ts` that look like:

```ts
/// <reference types="cypress" />
/// <reference types="cypress" />
import React from 'react';
import type { MountOptions, UnmountArgs } from '@cypress/react'; // NOT GOOD
export declare function mount(jsx: React.ReactNode, options?: MountOptions, rerenderKey?: string): Cypress.Chainable<import("@cypress/react").MountReturn>;
export declare function unmount(options?: UnmountArgs): Cypress.Chainable<undefined>;
```

Notice the `@cypress/react` import (`@` prefix). The types won't be found unless the user happens to have installed the mount package from npm which is unlikely. The lack of the types leads to `any` typing of the options.
Screen Shot 2022-10-20 at 5 20 51 PM

### Desired behavior

The types for `cypress/react18` (using mount from cypress npm module) should be equivalent to the types for `@cypress/react18`. This problem also exists for our other mount packages as they have a reliance on `@cypress/mount-utils`, so they should be cleaned up as well. In summary, the types should work regardless of where the mount is coming from.

### Test code to reproduce

Create a react18 project outside of our repo with `npm init vite`. Choose TS and create a test.

### Cypress Version

10.10.0

### Node version

v16.17.1

### Operating System

macOS 12.4

### Debug Logs

_No response_

### Other

_No response_

ZachJW34:
- Closes #24330

### User facing changelog
Fix types associated with `cy.mount` that we incorrectly `any` typed

### Additional details
When importing from `cypress/<framework>`, the types for the mount options were incorrectly typed as `any` if the option type was dependent on another cypress package such as `@cypress/react18 => @cypress/react` or `@cypress/vue => @cypress/mount-utils`. The bundled code rolls up these dependencies so that they are all in one file and yet the types weren't being rolled. This resulted in `index.d.ts` containing `import {Something} from '@cypress/<something>'` which would be converted to `any` since `@cypress/<something>` doesn't exist once we've moved the packages to the cli package. It also doesn't exist when the package is downloaded from npm since we don't list dependencies that are entirely rolled up inside packageJson.dependencies.

I've replaced the type generation with `rollup-plugin-dts` which provides consistent behavior between rolling up source code and types. If all of the source code for `@cypress/react` gets rolled up when bundling `@cypress/react18`, all of it's types should be rolled up as well.

### Steps to test
In an external repo that has CT setup, checkout the types for `cy.mount(comp, options)`, you should notice the options typed as any for react and vue. Checkout this branch, run `yarn` and then copy `cli/vue` into your external projects `node_modules/cypress` directory, replacing the existing folder and checkout the types again.

https://user-images.githubusercontent.com/25158820/198145921-35c875ef-0c48-4b4d-a815-14d966dc505e.mov

### PR Tasks


npm/vue2/patches/@vue+test-utils+1.3.0.patch
```diff
@@ -0,0 +1,24 @@
+diff --git a/node_modules/@vue/test-utils/types/index.d.ts b/node_modules/@vue/test-utils/types/index.d.ts
```
ZachJW34: Vue test utils was importing some types that weren't exported, and not exporting some types that we were importing. Since we bundle the types, we can get away with patching them.

scripts/sync-exported-npm-with-cli.js
```diff
@@ -29,6 +29,8 @@ packlist({ path: currentPackageDir })
   const exportName = currentPackageConfig.name.replace('@cypress/', '')
   const outDir = path.join(cliPath, exportName)
 
+  fs.removeSync(outDir)
```
warrensplayer, line 33: Comment for this line just to keep the logic clear.

```
// Remove output directory to clean up old files before building
```

warrensplayer (commented):
@ZachJW34 I am not a typescript export, but I think I follow what this PR is trying to accomplish.

Questions 
1. The PR only describes how to test the `vue` package.  This PR affects all bundled component libraries, correct?  Do we need to manually test those as well? 
2. Without the patch to `@vue/test-utils` in the `vue2` package, was the new rollup plugin erroring?  How should a tester of this PR confirm that this patch is correct?
3. Just curious, but why does the `vue2` include `@vue/test-utils` as a dependency and `vue` just has it has a devDependency?  Is it because `vue2` relies on an older version?

ZachJW34:
@warrensplayer

1. Yes this PR affects all packages. A good repo to use is https://github.com/cypress-io/cypress-component-testing-apps, the process is the same for testing those (make sure to have built all of the mount packages, though it is accomplished by running `yarn`)
2. Yes the rollup plugin was erroring. You could either remove the patch and reinstall, or you can go into `npm/vue2/node_modules/@vue/test-utils` and manually revert the changes the patch is making then build vue2 again to see the errors.
3. I'm not sure why it is marked as a dependency. The entire package gets rolled up so it shouldn't need to be installed, and won't be when importing from `cypress/vue2` vs `@cypress/vue2`. I'll move it to be a `devDependency` to be more consistent.

scripts/sync-exported-npm-with-cli.js
ZachJW34: Done in [3618f14](https://github.com/cypress-io/cypress/pull/24415/commits/3618f146a42fb057b5168c21acf538579c3b9fac)

rockindahizzy (approved):
Everything worked for me! 

Random question: Is there anyway to test this stuff? Probably not a ton of value in it, but worth a question at least.

ZachJW34:
@rockindahizzy I don't know a straightforward way. The reason this flew under the radar is because our workspace installs our packages into `node_modules` which allows the types to resolve properly. Even system-tests get the `cli` package symlinked, which will end up resolving node_modules within the cypress project.

npm/mount-utils/create-rollup-entry.mjs
```diff
@@ -67,5 +68,10 @@ export function createEntries (options) {
     console.log(`Building ${format}: ${finalConfig.output.file}`)
 
     return finalConfig
+  }).concat({
+    input,
+    output: [{ file: 'dist/index.d.ts', format: 'es' }],
+    plugins: [dts({ respectExternal: true })],
```
lmiller1990, line 74: Woah, did not know about rollup-plugin-dts. Neat.

npm/mount-utils/package.json
```diff
@@ -15,6 +15,7 @@
     "@rollup/plugin-commonjs": "^17.1.0",
     "@rollup/plugin-node-resolve": "^11.1.1",
     "rollup": "^2.38.5",
+    "rollup-plugin-dts": "^4.2.3",
     "rollup-plugin-typescript2": "^0.29.0",
     "typescript": "^4.7.4"
   },
```
lmiller1990, line 21: Below script (inline-types) was how I solved this hackily because I didn't know about rollup-inline-types. The solution here is far better.

npm/vue2/patches/@vue+test-utils+1.3.0.patch
lmiller1990: We can just patch test utils if you want?

These types are handwritten and probably not complete - test utils v1 was (and still is) written in JS. It's basically in maintenance mode - I'm not sure many people are using TS with Vue 2, anyway.

npm/vue2/patches/@vue+test-utils+1.3.0.patch
```diff
@@ -0,0 +1,24 @@
...
+index cc15565..9f10c1b 100644
+--- a/node_modules/@vue/test-utils/types/index.d.ts
++++ b/node_modules/@vue/test-utils/types/index.d.ts
+@@ -1,7 +1,9 @@
+ import Vue, { VNodeData, ComponentOptions, FunctionalComponentOptions, Component, RenderContext } from 'vue'
+-import { DefaultProps, PropsDefinition } from 'vue/types/options'
```
lmiller1990: Since Test Utils is a devDep should we make this a dev patch? https://github.com/ds300/patch-package#dev-only-patches

lmiller1990 (commented):
This is a great solution, nice!!

Few comments, nothing big.

npm/vue2/patches/@vue+test-utils+1.3.0.patch
lmiller1990: Edit... I published a new Test Utils version (v1.3.1) that has this patch. https://github.com/vuejs/vue-test-utils/releases/tag/1.3.1

What about using that?

warrensplayer (commented):
It makes sense to try out the patch to Vue test utils that Lachlan published and revert that patch file. Other than that I think this is good to go.

npm/vue2/patches/@vue+test-utils+1.3.0.patch
ZachJW34: Thanks @lmiller1990, will update and remove the patch!

ZachJW34: Done in [01be0bc](https://github.com/cypress-io/cypress/pull/24415/commits/01be0bce155a89bae2845ca0d4154de955a8c199)

ZachJW34:
@warrensplayer Patch removed in [01be0bc](https://github.com/cypress-io/cypress/pull/24415/commits/01be0bce155a89bae2845ca0d4154de955a8c199)!

lmiller1990 (approved):

Issue #23901, opened 2022-09-20
Detached DOM: Update $actionability to support cy.currentSubject()


BlueWinds:
### What would you like?

Queries solve a major problem: how can Cypress safely requery the DOM when the subject becomes detached? The Cypress command queue also handles this cleanly and internally for queries and their associated assertions.

However, commands - such as .click() - still receive a static subject, and this subject can "go stale" as the DOM mutates. Because there is no generic handling of subjects - each command contains its own logic - fixing this will involved updating each command one at a time.

We have to start somewhere, to understand exactly what this solution looks like and establish patterns for updating future commands. This will also help scope out how much work remains on other commands, whether it is simple or complex to update each one.

BlueWinds:
Merged into feature branch; closing.

PR #24417, opened 2022-10-26
feat: Fix detached DOM errors for all Cypress commands

BlueWinds:
- Closes https://github.com/cypress-io/cypress/issues/23901

### User facing changelog
Fixed the majority of Detached DOM errors, and updates the error messages when they do occur pointing out the correct way to solve them.

#### Breaking Change
Cypress *always* requeries aliases when they are referenced. This can result in certain tests that use to pass failing. For example:

```
cy.findByTestId('popover')
  .findByRole('button', { expanded: true })
  .as('button')
  .click()

cy.get('@button')
  .should('have.attr', 'aria-expanded', 'false')
```

previously passed, because the initial button was collapsed when first queried, and then later expanded. However, in Cypress 12, this test fails because the alias is always requeried from the DOM, effectively resulting in the following execution:

```
cy.findByTestId('popover')
  .findByRole('button', { expanded: true })
  .click()

cy.findByTestId('popover')
  .findByRole('button', { expanded: true }) // A button which matches here (is expanded)...
  .should('have.attr', 'aria-expanded', 'false') // ...will never pass this assertion.
```

You can rewrite tests like this to be more specific; in our case, we changed the alias to be the *first* button rather than the *unexpanded button.*

```
cy.findByTestId('popover')
  .findAllByRole('button')
  .first()
  .as('button')
```

#### Breaking Change
If `.invoke()` is followed by additional commands or assertions, it will call the named function multiple times. This has the benefit that following assertions can more reliably use the functions return value.

If this behavior is undesirable - you want the function invoked only once - move the following commands / assertions to their own chain command. For example, rewrite

```
cy.get('input').invoke('val', 'text').type('newText')
```

to

```
cy.get('input').invoke('val', 'text')
cy.get('input').type('newText')
```

**Breaking Change:** `.invoke()` now throws an error if the function returns a promise. If you wish to call a method that returns a promise and wait for it to resolve, us `.then()` instead of `.invoke()` For example, rewrite

```
cy.wrap(myAPI)
  .invoke('makeARequest', 'http://example.com')
  .then(res => { ...handle response... })
```

to

```
cy.wrap(myAPI)
  .then(api => api.makeARequest('http://example.com'))
  .then(res => { ...handle response... })
```

### Additional details
The breaking changes above actually happened in https://github.com/cypress-io/cypress/issues/23902, but they're far more heavily exposed here, and I've finally managed to articulate the breaking change in a user-digestable way.

In addition to the breaking changes mentioned above, queries are no longer displayed as "pending" in the command log while following commands are retrying; they resolve the first time they pass. New behavior:

new-pending-behavior

Old behavior:

old-pending-behavior

The error messages around detached DOM errors have updated significatly. Users will encounter them less frequently; the most common cases won't occur at all anymore, but when users do encounter errors, we now make specific recommendations on how to resolve the error.

There are now two detached dom error messages, fairly similar but with different verbage based on the situation in which they occur. New error messages:

detached-new-error-2
detached-new-error

Old error message:

detached-old-error

### PR Tasks
- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4835 - Just starting review. Would like to not block this PR on the docs.


packages/driver/cypress/e2e/commands/actions/check.cy.js
```diff
@@ -803,58 +824,6 @@ describe('src/cy/commands/actions/check', () => {
   })
 
   context('#uncheck', () => {
-    it('does not change the subject', () => {
```
BlueWinds: These test removals are not necessary for detached DOM; but looking through the file, the tests for `.check()` and `.uncheck()` are extremely redundant, considering they share an implementation.

```
Commands.addAll({ prevSubject: 'element' }, {
    check (subject, values, options) {
      return checkOrUncheck.call(this, Cypress, cy, 'check', subject, values, options)
    },

    uncheck (subject, values, options) {
      return checkOrUncheck.call(this, Cypress, cy, 'uncheck', subject, values, options)
    },
  })
```

That is, they invoke the same function under the hood, sharing 90% of the logic and varying only in very specific ways, eg:

```
  const isNoop = ($el) => {
    return type === 'check'
      ? $el.prop('checked')
      : !$el.prop('checked')
  }
```

But we test the same features of both commands, eg:

```
  context('#check', () => {
    it('does not change the subject', () => {...})
    it('changes the subject if specific value passed to check', () => {...})

    it('can check a checkbox or radio', () => {...})
  })

  context('#uncheck', () => {
    it('does not change the subject', () => {...})
    it('changes the subject if specific value passed to uncheck', () => {...})

    it('can uncheck a checkbox', () => {...})
    it('throws specifically on a radio', () => {...})
  })
```

So we're basically running the same tests twice. There are certainly some tests that are different - eg, you can check a checkbox or radio, but you cannot uncheck a radio. But I don't see much value in testing the same code multiple times for the parts they do share, and spent a few minutes removing some of the duplicates in this file.

The same can be said of several other commands, but I didn't want to go through too many files and do the same thing unless other people agree with the approach.

I tried to ask questions about this on slack, but didn't get much engagement, so here it is in PR form.

packages/driver/cypress/e2e/commands/actions/click.cy.js
```diff
@@ -813,22 +834,20 @@ describe('src/cy/commands/actions/click', () => {
     })
 
     it('places cursor at the end of [contenteditable]', () => {
-      cy.get('[contenteditable]:first')
-      .invoke('html', '<div><br></div>').click()
-      .then(expectCaret(0))
+      cy.get('[contenteditable]:first').as('edit')
```
BlueWinds, line 837: Another test fallen victim to the breaking change around cy.invoke; because it's now a query, it can (will) be called multiple times while executing other commands; this resulted in the cursor moving around and the test failing.

Fortunately, the solution is simple; move .click() to its own chain, so .invoke() isn't part of its subject. I'll include this recommendation as part of the patch notes / upgrade guide when Detached DOM goes out.

packages/driver/cypress/e2e/commands/actions/selectFile.cy.js
```diff
@@ -628,23 +628,30 @@ is being covered by another element:
         })
       })
 
-      it('waits until input stops animating', {
-        defaultCommandTimeout: 1000,
-      }, () => {
-        let retries = 0
+      it('retries until label is not disabled', () => {
```
BlueWinds: There's no need to test animation specifically - .disabled is part of `$actionability.verify()` as well. Enabling and disabling an element basically tests the same thing, and in general I prefer less-mocked tests over more-mocked tests.

packages/driver/cypress/e2e/cypress/cy.cy.js
```diff
@@ -553,6 +536,21 @@ describe('driver/src/cypress/cy', () => {
       })
     })
 
+    it('closes each log as the query completes', (done) => {
```
BlueWinds, line 539: This is a behavior change, a result of Brian and my discussions around command log completion. See PR description for gifs of the behavior under verification here.

packages/driver/src/cy/actionability.ts
```diff
@@ -453,6 +455,19 @@ const verify = function (cy, $el, config, options, callbacks: VerifyCallbacks) {
     // the checks and firing the event!
     const retryActionability = () => {
       try {
+        if (callbacks.subjectFn) {
```
BlueWinds, line 458: This is the business end of the PR: `$actionability.verify()` now accepts a subjectFn() callback, to requery the subject between retries. Each command is responsible for passing in a function that resolves the subject; usually a thunk wrapped around `cy.getSubjectFromChain()`, but some commands do additional work on top of that; see each updated command for details.

packages/driver/src/cy/commands/actions/click.ts
```diff
@@ -194,17 +198,21 @@ export default (Commands, Cypress, cy: $Cy, state, config) => {
       // once we establish the coordinates and the element
       // passes all of the internal checks
       return $actionability.verify(cy, $el, config, individualOptions, {
+        subjectFn: options.subjectFn || (() => cy.getSubjectFromChain(subjectChain).eq(index)),
```
BlueWinds, line 201: `perform()` is called one element at a time, and each checks actionability individually, so the subjectFn() needs to pull out the element matching the current index after resolving the subjectChain into a set of elements.

This algorithm has some implications. While it fixes the case where the DOM rerenders between clicks, it fails on a case like so:

```
<tr><td><button>Delete Row</button></td></tr>
<tr><td><button>Delete Row</button></td></tr>
<tr><td><button>Delete Row</button></td></tr>

cy.get('button').click({ multiple: true })
```

The DOM changes - not just rerenders, but changes meaningfully - between each invocation of `perform()`. This will result in errors (resolving .get('button') returns fewer elements with each click, so .get('button').eg(2) will end up not receiving an element).

This case was explicitly laid out in the tech brief as "will not fix", but wanted to call it out now that we're here in practice.

packages/driver/src/cy/commands/actions/scroll.ts
```diff
@@ -345,23 +324,53 @@ export default (Commands, Cypress, cy, state) => {
...
-        if (!options.ensureScrollable) {
-          return
-        }
+      const subjectChain = cy.subjectChain()
 
+      const ensureScrollability = () => {
         try {
-          // make sure our container can even be scrolled
-          return cy.ensureScrollability($container, 'scrollTo')
+          subject = cy.getSubjectFromChain(subjectChain)
```
BlueWinds, line 334: Moved a bunch of these checks asserting on the state of the subject inside the retry loop, so that .scrollTo() can wait for them to pass if they initially fail.

packages/driver/src/cy/commands/actions/select.ts
```diff
@@ -104,13 +103,16 @@ export default (Commands, Cypress, cy) => {
...
+        options.$el = cy.getSubjectFromChain(subjectChain)
         let notAllUniqueValues
 
         // throw if <select> is disabled
-        if (!options.force && options.$el.prop('disabled')) {
-          node = $dom.stringify(options.$el)
-          $errUtils.throwErrByPath('select.disabled', { args: { node } })
+        if (!options.force) {
+          cy.ensureElement(options.$el, 'select', options._log)
+          cy.ensureNotDisabled(options.$el, options._log)
```
BlueWinds, line 116: Standardizing error messages - .select() now uses the same `ensures` functions as other commands, since its original error message wasn't meaningfully different (had less information than the default).

It was also missing an 'attached to DOM' check entirely.

packages/driver/src/cy/commands/actions/selectFile.ts
```diff
@@ -259,37 +261,48 @@ export default (Commands, Cypress, cy, state, config) => {
         })
       }
 
-      let eventTarget = subject
+      const getEventTargetFromSubject = (subject) => {
```
BlueWinds, line 264: selectFile did not properly retry its assertions about the subject; eg, it did not wait for the input to become enabled, but threw an error immediately.

I've moved these verifications inside the `subjectFn` we pass to $actionability.verify(), so they'll now properly retry in all situations (not just detached DOM situations).

packages/driver/src/cy/commands/actions/trigger.ts
```diff
@@ -117,12 +117,16 @@ export default (Commands, Cypress, cy, state, config) => {
         subject = options.$el.first()
       }
 
+      const subjectChain = cy.subjectChain()
```
BlueWinds, line 120: This is an example of the simplest possible case - `.trigger()` requires no additional logic. It reads the current subject chain and passes a function resolving it in to `$actionability.verify` as `subjectFn`.

packages/driver/src/cy/commands/aliasing.ts
```diff
@@ -36,7 +35,7 @@ export default function (Commands, Cypress, cy) {
       const log: any = _.last(cy.queue.logs({
         instrument: 'command',
         event: false,
-        chainerId,
+        chainerId: this.get('chainerId'),
```
BlueWinds, line 38: No functional difference to this change, I just liked the way it looked better down here. :woman_shrugging:

packages/driver/src/cypress/command.ts
```diff
@@ -53,21 +45,17 @@ export class $Command {
...
+      }
+
       if (log.get('_error')) {
         log.error(log.get('_error'))
       } else {
         log.set('snapshot', false)
         log.finish()
       }
     })
-
```
BlueWinds, line 63: Because all commands now end immediately on resolution, we don't need to traverse back up the chain looking for unended log messages.

packages/driver/src/cypress/error_messages.ts
```diff
@@ -143,28 +143,6 @@ export default {
     },
   },
 
-  selectFile: {
```
BlueWinds, line 146: selectFile wasn't properly alphabetized. No changes other than moving it lower in the file.

packages/driver/src/cypress/utils.ts
```diff
@@ -403,28 +403,4 @@ export default {
   isPromiseLike (ret) {
     return ret && _.isFunction(ret.then)
   },
-
```
BlueWinds, line 406: Moved this function onto `cy` directly, because we want to expose it to libraries / users that want to write custom commands.

.circleci/config.yml
```diff
@@ -27,8 +27,7 @@ mainBuildFilters: &mainBuildFilters
     branches:
       only:
         - develop
-        - fix-ci-deps
-        - issue-23843_electron_21_upgrade
+        - issue-7306
```
BlueWinds, line 30: My intent here is to run the full suite of tests - and build the binary - in the feature branch, and then PR into there to fix any issues discovered.

packages/app/cypress/e2e/top-nav.cy.ts
```diff
@@ -195,7 +195,7 @@ describe('App Top Nav Workflows', () => {
       })
 
       it('hides dropdown when version in header is clicked', () => {
-        cy.findByTestId('cypress-update-popover').findByRole('button', { expanded: false }).as('topNavVersionButton').click()
+        cy.findByTestId('cypress-update-popover').findAllByRole('button').first().as('topNavVersionButton').click()
```
BlueWinds, line 198: These are an example of a test broken by the changes - see PR description for details.

packages/app/package.json
```diff
@@ -28,7 +28,7 @@
     "@intlify/vite-plugin-vue-i18n": "2.4.0",
     "@packages/frontend-shared": "0.0.0-development",
     "@percy/cypress": "^3.1.0",
-    "@testing-library/cypress": "8.0.0",
+    "@testing-library/cypress": "BlueWinds/cypress-testing-library#119054b5963b0d2e064b13c5cc6fc9db32c8b7b5",
```
BlueWinds, line 31: This is a fork of cypress-testing-library which uses queries. My current plan is to wait for this PR to merge into `issue-7306`, which will build the cypress binary; I can then update my fork of `cypress-testing-library` to point to that branch for its own tests, and open a PR upstream.

I was running into issues while pointing only to the fork / branch, where the `node_modules` cache wouldn't update when I updated my fork - pointing to a specific commit here in the monorepo helped.

packages/launchpad/cypress/e2e/top-nav-launchpad.cy.ts
```diff
@@ -504,7 +504,7 @@ describe('Launchpad Top Nav Workflows', () => {
             cy.findByRole('button', { name: 'Log in' }).click()
           })
 
-          cy.findByRole('dialog', { name: 'Log in to Cypress' }).within(() => {
+          cy.findByRole('dialog').within(() => {
```
BlueWinds, line 507: After the first button click, the dialog no longer has the name "Log in to Cypress" - it's turned into an error popup. Because we now re-query the DOM though, .findByRole() failed to find any dialog with a matching name, and the test failed.

We're now within *a dialog*, rather than within *a dialog named Log In*. This PR forced us as test-writers to be clearer about exactly what we want - IMO a good side effect, but unfortunately a breaking change.

cli/types/cypress.d.ts
```diff
@@ -486,6 +489,9 @@ declare namespace Cypress {
       ): void
       overwrite<T extends keyof Chainable>(name: T, fn: CommandFnWithOriginalFn<T>): void
       overwrite<T extends keyof Chainable, S extends PrevSubject>(name: T, fn: CommandFnWithOriginalFnAndSubject<T, PrevSubjectMap[S]>): void
+
```
flotwig, line 548: Huh, kinda surprised that there aren't type comments for any of these.

packages/app/package.json
flotwig: :+1: Specific commit is always the way to go, IMO. Using a branch causes the caching issue/bug with `yarn`, plus it makes builds unpredictable since the branch's head commit can change at any time.

packages/driver/cypress/e2e/commands/actions/check.cy.js
flotwig: This is definitely the struggle with having all of driver tests be integration tests, we're testing the full stack each time. I get why you want to remove these. But besides CI cost, what's the issue if we keep the tests? Even though they're highly redundant, I don't know that the downside to having them is enough to warrant removing them.

packages/driver/cypress/e2e/commands/actions/click.cy.js
flotwig: Seeing the test changes from making `.invoke` a query command, I'm wondering if it's the right thing to do. I can see this causing confusion when someone just wants to run a function and chooses to use `.invoke` based off the name of the command alone. I feel like I knew the answer once but I forgot, I checked the brief and #24203 but couldn't find it, why is `.invoke` being changed to a query command again?

packages/driver/cypress/e2e/commands/assertions.cy.js
```diff
@@ -643,29 +643,7 @@ describe('src/cy/commands/assertions', () => {
 
         cy.on('fail', () => {})
 
-        cy.readFile('does-not-exist.json').should('exist')
-      })
-
-      it('throws when the subject isnt in the DOM', function (done) {
```
flotwig, line 649: Why remove this test? I can't tell why it would conflict with the other changes.

system-tests/projects/vite-ct-function-api/package.json
```diff
@@ -5,7 +5,7 @@
   "dependencies": {},
   "devDependencies": {
     "@cypress/vite-dev-server": "file:../../../npm/vite-dev-server",
-    "@testing-library/cypress": "8.0.2",
+    "@testing-library/cypress": "BlueWinds/cypress-testing-library",
```
flotwig:
```suggestion
    "@testing-library/cypress": "BlueWinds/cypress-testing-library#119054b5963b0d2e064b13c5cc6fc9db32c8b7b5",
```

system-tests/projects/vite-ct-object-api/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.0-test",
   "private": true,
   "dependencies": {
-    "@testing-library/cypress": "8.0.2",
+    "@testing-library/cypress": "BlueWinds/cypress-testing-library",
```
flotwig:
```suggestion
    "@testing-library/cypress": "BlueWinds/cypress-testing-library#119054b5963b0d2e064b13c5cc6fc9db32c8b7b5",
```

packages/driver/src/cypress/error_messages.ts
```diff
@@ -1880,26 +1876,49 @@ export default {
...
 
-          The previous command that ran was:
+          > \`cy.get('button').click().click()\`
 
-            > ${cmd(obj.prev)}
+          to
 
-          This DOM element likely became detached somewhere between the previous and current command.
+          > \`cy.get('button').click()\`
+          > \`cy.get('button').click()\`
```
flotwig: This feels like a weird pattern to encourage, which the [tech brief notes](https://github.com/cypress-io/tech-briefs/tree/main/briefs/detached-dom#aliases). Users already use Cy aliases as a way to avoid re-writing long selectors. With detached DOM, you could do something like...

```typescript
          > \`cy.get('button').as('button').click()\`
          > \`cy.get('@button').click()\`
```

...and it would also still work, right? What do you think about having this error message encourage the use of aliases to re-query the DOM, instead of reusing the selector itself?

packages/driver/src/cy/commands/actions/click.ts
```diff
@@ -194,17 +198,21 @@ export default (Commands, Cypress, cy: $Cy, state, config) => {
...
+
         onScroll ($el, type) {
           return Cypress.action('cy:scrolled', $el, type)
         },
 
-        onReady ($elToClick, coords) {
+        onReady ($elToClick, coords, $el) {
```
lmiller1990, line 207: Note for myself: I have reviewed until this file, I will continue tomorrow.

emilyrohrbough (changes requested):
>  previously passed, because the initial button was collapsed when first queried, and then later expanded. However, in Cypress 12, this test fails because the alias is always requeried from the DOM, effectively resulting in the following execution:

> cy.findByTestId('popover')
  .findByRole('button', { expanded: true })
  .click()

> cy.findByTestId('popover')
  .findByRole('button', { expanded: true }) // A button which matches here (is expanded)...
  .should('have.attr', 'aria-expanded', 'false') // ...will never pass this assertion.
  
  
A few weeks (months?) ago when we had talked, we had discussed that re-running queries would omit assertions when. Why would aliases keep reference to assertions? This seems unexpected that a set of commands are aliases instead of the query. Would it be more appropriate if you can only alias query commands or intercepts instead of anything?

cli/types/cypress.d.ts
emilyrohrbough: Links to docs at least would be nice!

packages/launchpad/package.json
```diff
@@ -28,7 +28,7 @@
     "@packages/frontend-shared": "0.0.0-development",
     "@percy/cypress": "^3.1.0",
     "@purge-icons/generated": "0.8.1",
-    "@testing-library/cypress": "8.0.0",
+    "@testing-library/cypress": "BlueWinds/cypress-testing-library#119054b5963b0d2e064b13c5cc6fc9db32c8b7b5",
```
emilyrohrbough, line 31: Do we have to reference this version of the testing library for these changes to be successful?  If we don't upgrade, what is the results/errors? What can users expect if they don't upgrade?

packages/driver/types/internal-types-lite.d.ts
```diff
@@ -33,6 +33,7 @@ declare namespace Cypress {
     isStable: IStability['isStable']
     fail: (err: Error, options:{ async?: boolean }) => Error
     getRemoteLocation: ILocation['getRemoteLocation']
+    subjectChain: (chainerId?: string) => any[]
```
emilyrohrbough: I assume we can't be more specific on the return types because of `cy.wrap()`?

packages/driver/src/cypress/log.ts
```diff
@@ -168,7 +168,7 @@ const defaults = function (state: StateFunc, config, obj) {
     if (_.isFunction(obj.type)) {
       const chainerId = current && current.get('chainerId')
 
-      obj.type = obj.type(current, (state('subjects') || {})[chainerId])
+      obj.type = obj.type(current, cy.subjectChain(chainerId))
```
emilyrohrbough, line 171: Any reason assertions doesn't handle this directly?

packages/driver/src/cypress/error_messages.ts
```diff
@@ -1880,26 +1876,49 @@ export default {
...
+
+            `,
+        docsUrl: 'https://on.cypress.io/element-has-detached-from-dom',
+      }
+    },
+    detached_after_command (obj) {
+      return {
+        message: stripIndent`\
+          ${cmd(obj.name)} failed because the page updated as a result of this command, but you tried to continue the command chain. The subject is no longer attached to the DOM, and Cypress cannot requery the page after commands such as ${cmd(obj.name)}.
```
emilyrohrbough, line 1908: This wording standalone is a little confusing. Wouldn't the error be on the continued command chain when the dom is no longer on the page?
```suggestion
          ${cmd(obj.name)} failed because the page mutated after running the previous command. The subject is no longer attached to the DOM, and Cypress cannot requery the page to run ${cmd(obj.name)}.
```

packages/driver/src/cypress/error_messages.ts
```diff
@@ -1880,26 +1876,49 @@ export default {
...
+
+          We initially found matching element(s), but while waiting for them to become actionable, they disappeared from the page. Common situations why this happens:
+            - Your JS framework re-rendered asynchronously
+            - Your app code reacted to an event firing and removed the element
 
-          Cypress requires elements be attached in the DOM to interact with them.
+          You can typically solve this by breaking up a chain. For example, rewrite:
 
-          The previous command that ran was:
+          > \`cy.get('button').click().click()\`
```
emilyrohrbough, line 1894: I think this example is another area I would have expected to work under the hood like:
```
1. get button
2. click button
3. Try click button // DETACHED!! // get button // click button (if exits PASS ...or if still not exists FAIL)
```
I would have expected the action to be omitted from the subject chain & the query itself be re-executed.

It seems like my understanding on how this part would be implemented is a bit off.

packages/driver/src/cypress/error_messages.ts
```diff
@@ -1656,6 +1630,28 @@ export default {
     },
   },
 
+  selectFile: {
+    docsUrl: 'https://on.cypress.io/selectfile',
+    invalid_action: {
+      message: `${cmd('selectFile')} \`action\` can only be \`select\` or \`drag-drop\`. You passed: \`{{action}}\`.`,
+    },
+    invalid_array_file_reference: {
+      message: `${cmd('selectFile')} must be passed an array of Buffers or objects with non-null \`contents\`. At files[{{index}}] you passed: \`{{file}}\`.`,
```
emilyrohrbough, line 1639: What is `files[{{index}}]` referencing?

packages/driver/src/cypress/error_messages.ts
```diff
@@ -1656,6 +1630,28 @@ export default {
...
+    },
+    invalid_single_file_reference: {
+      message: `${cmd('selectFile')} must be passed a Buffer or an object with a non-null \`contents\` property as its 1st argument. You passed: \`{{file}}\`.`,
```
emilyrohrbough, line 1642: Below you say it can also attach strings. Missing array of strings or no?

packages/driver/src/cypress/error_messages.ts
```diff
@@ -1656,6 +1630,28 @@ export default {
...
+    },
+    multiple_elements: {
+      message: `${cmd('selectFile')} can only be called on a single element. Your subject contained {{num}} elements.`,
+    },
+    not_file_input: {
+      message: `${cmd('selectFile')} can only be called on an \`<input type="file">\` or a \`<label for="fileInput">\` pointing to or containing one. Your subject is: \`{{node}}\`.`,
```
emilyrohrbough: The reference of subject is an implementation detail IMO. 
```suggestion
      message: `${cmd('selectFile')} can only be called on an \`<input type="file">\` or a \`<label for="fileInput">\`, but received the element:
      
      \`{{node}}\`.`,
```

packages/driver/src/cypress/cy.ts
```diff
@@ -1269,25 +1285,27 @@ export class $Cy extends EventEmitter2 implements ITimeouts, IStability, IAssert
...
   }
 
   /* Given a chain of functions, return the actual subject. `subjectChain` might look like any of:
+   * []
    * [<input>]
    * ['foobar', f()]
    * [undefined, f(), f()]
    */
-  getSubjectFromChain (subjectChain: SubjectChain) {
+  getSubjectFromChain (subjectChain: SubjectChain | null | undefined) {
```
emilyrohrbough, line 1304: Why doesn't getSubjectFromChain() call this.subjectChain directly?

packages/driver/src/cy/commands/actions/select.ts
```diff
@@ -104,13 +104,16 @@ export default (Commands, Cypress, cy) => {
         $errUtils.throwErrByPath('select.invalid_multiple')
       }
 
+      const subjectChain = cy.subjectChain()
+
       const getOptions = () => {
+        options.$el = cy.getSubjectFromChain(subjectChain)
```
emilyrohrbough, line 110: If I'm understanding correctly, we are requerying each time a chained command executed, not when it's retried?

packages/driver/src/cypress/command.ts
```diff
@@ -53,21 +45,17 @@ export class $Command {
         log.snapshot()
       }
 
+      if (!log.get('snapshots')) {
```
emilyrohrbough: nit: `if (log.get('next') || !log.get('snapshots'))`

If I'm understanding correctly, we are requerying each time a chained command executed? So 
```
cy.get().should('...').should('...').click()

```

BlueWinds:
> > previously passed, because the initial button was collapsed when first queried, and then later expanded. However, in Cypress 12, this test fails because the alias is always requeried from the DOM, effectively resulting in the following execution:
> 
> > cy.findByTestId('popover')
> > .findByRole('button', { expanded: true })
> > .click()
> 
> > cy.findByTestId('popover')
> > .findByRole('button', { expanded: true }) // A button which matches here (is expanded)...
> > .should('have.attr', 'aria-expanded', 'false') // ...will never pass this assertion.
> 
> A few weeks (months?) ago when we had talked, we had discussed that re-running queries would omit assertions when. Why would aliases keep reference to assertions? This seems unexpected that a set of commands are aliases instead of the query. Would it be more appropriate if you can only alias query commands or intercepts instead of anything?

`{expanded: true}` is not an assertion - it's an option on `.findByRole()`, which resolves to a query eqivelent to  `.get('[aria-role="button"][aria-expanded="true"]')`. Rewritten that way, I think it's fairly clear why this is a problem:

```
cy.findByTestId('popover')
  .get([aria-role="button"][aria-expanded="true"])
  .as('button')
  .click()

cy.get('@button').should('have.attr', 'aria-expanded', 'false')

---> 

cy.findByTestId('popover')
  .get([aria-role="button"][aria-expanded="true"])
  .should('have.attr', 'aria-expanded', 'false')
```

We're not keeping a reference to an assertion 'the button should be expanded' - we're re-checking the DOM for 'a button which is expanded'.

cli/types/cypress.d.ts
BlueWinds: Good call - some comments here can definitely help make sense of the mess of TS definitions.

I'm adding links to a couple of anchors that don't exist yet:
- #Queries
- #Overwrite-Existing-Queries

I'll be creating these in the docs PR, which isn't ready yet.

packages/driver/cypress/e2e/commands/actions/check.cy.js
BlueWinds: The costs are:
- Every time I update the error message, I had to update it not just across multiple files (each file needs to test the functionality), but multiple times within each file. It's not obvious in the PR here, but I touched these tests many times.
- Every time I want to verify my changes to the command, I have to run the tests locally. The longer the test file, the slower the feedback loop.
- Every time I push my changes, CI runs - the longer our tests, the slower the feedback cycle.

Basically, all code has a cost (including tests). Working on these commands for several months, I've become *acutely* aware of the costs associated with duplicate tests. This change isn't for the future, it's for the past - specifically, past me, and all the time wasted watching driver tests run and tracking down duplicates of tests within the same file.

packages/driver/cypress/e2e/commands/actions/click.cy.js
BlueWinds: So imagine we leave `.invoke()` as a command.

A test written like this is then vulnerable to Detached DOM errors - if in response to the element's html being changed, React decided to rerender the page (perhaps updating a "preview" area based on the content?), `.click()` would fail - the original contenteditable is detached from the DOM.

The solution there would be to requery the DOM between commands - exactly what we've done here. Basically, making .invoke() into a query *revealed that the test was poorly written to begin with* - our case was simple enough that we don't run into detached dom errors, but it followed a bad pattern.

----

So that's why "this test change isn't alarming", but why is .invoke() better as a selector than as a command?

Because you can use a query like a command, by making it a terminating element on a chain. But you can never use a command like a query - there lie dragons.

```
cy.$$('input').first().val('foo')

setTimeout(() => {
  cy.$$('input').first().val('bar')
}, 1000)

cy.get('input').first().invoke('val').should('eql', 'bar')
```

As a command, .invoke() runs *once* - you get out the value, and then retry assertions on the returned value - 'foo' - until it times out. There only workaround is to not use invoke() at all.

----

This change is actually fairly straightforward to explain to users. In the upcoming docs PR, I'm adding a warning near the top of .invoke():

> .invoke() is a query, not a command. As explained in retryability (link to guide), this means *it will be called more than once* if anything follows it in the same command chain. If you want the function to be called only once, make sure it's at the end!

> For example, `cy.get('button').invoke('click').should('have.focus')` will result in the button being clicked at least twice. Rewrite this to `cy.get('button').invoke('click'); cy.get('button').should('have.focus')`

Exact wording TBD.

packages/driver/cypress/e2e/commands/assertions.cy.js
BlueWinds: This case is covered below by 'throws when the subject eventually isn't in the DOM`.

It still passes, but it's adding no value compared to later tests in the file that test the same functionality.

packages/driver/src/cypress/command.ts
BlueWinds: Combined the conditionals. :+1: 

----

Did this comment get cut off? Feels like there's some missing text here, and maybe another point I'm not getting?

packages/driver/src/cypress/cy.ts
BlueWinds: Basically, subject chains are used in more places than just `state('subjects')` - notably `.as()` and `.within()`. They're also often kept in memory for use later.

```
cy.get('div').within(() => {
  cy.get('button').as('button')
})

cy.get('@button') 
```

getSubjectFromChain is used a couple of times here:
- `cy.get('@button')` used getSubjectFromChain here on the alias subject chain `[undefined, get(button)]`.
- `cy.get('button')` cached the withinSubjectChain `[undefined, get(div)]` when it first executed, so that when it was rerun - after `.within()` has resolved - it could resolve the withinSubject to a specific element.

packages/driver/src/cypress/error_messages.ts
BlueWinds:
```
cy.get('input').selectFile([{ contents: 'foo' }, 33])

-->

`cy.selectFile()` must be passed an array of Buffers or objects with non-null `contents`. At files[1] you passed: `33`.

```

33 is not a valid file.

packages/driver/src/cypress/error_messages.ts
BlueWinds: Honestly I don't remember. it's been too long since I worked on selectFile. I do recall the variety of possible subject combinations being quite a hassle to keep track of, and I wish I'd been stricter on exactly what can be passed in.

I admit not to being super interested in revisiting the selectFile error messages in this PR, since all I did was alphabetize them.

packages/driver/src/cypress/error_messages.ts
BlueWinds: Done.

packages/driver/src/cypress/error_messages.ts
BlueWinds: The way you're describing was my initial idea; after discussing with Brian for a couple weeks, we settled on the current implementation. All commands *reset the subject chain*.

https://github.com/cypress-io/tech-briefs/tree/main/briefs/detached-dom#why-do-we-reset-the-subject-chain-after-a-command lays out the reasoning.

> ## Why do we reset the subject chain after a command?

> The initial draft of this tech brief proposed that we could treat commands as no-op queries, and continue passing the subject chain through them. This solves the above case - the user would not need to update their test. But attempting to guess the user's intention has other dangers, as in the following example.

> `cy.get('checkbox').click().should('have.class', 'checked')`

> Consider the following sequence of events:

>    1. Cypress locates a "I agree to receive emails" checkbox on the page and clicks it.
>    2. Checking this causes the user's application to realize the user is not logged in, and redirects them to the login page.
>    3. Cypress sees the "receive emails" checkbox is no longer attached to the DOM, and requeries the subject.
>    4. It now locates a "Keep me logged in" on the login page, and sees that it has the "checked" class.
>    5. The test passes, despite the application exhibiting unexpected behavior.

> We therefore treat all commands as barriers, ending the current subject chain and beginning a new one. If this results in a detached DOM error, we expect the user to resolve this confusion rather than attempting to guess their intention.

packages/driver/src/cypress/error_messages.ts
BlueWinds: That definitely works (and is how I redid some of our own tests). I put the simplest possible version here just to keep the error message short; I'll update to suggest aliases instead, since yeah, that matches a more realistic scenario better.

packages/driver/src/cypress/error_messages.ts
BlueWinds: I'm going to disagree here: I think the existing wording more accurately reflects the actual problem.

I click a link, the page navigates.

"The page updated as a result of this command" is clearer than "The page mutated after running this command."

packages/driver/src/cypress/log.ts
BlueWinds: My suspicion is that it relates to the extremely roundabout way assertions generate exceptions, but really have no idea; this is just existing functionality that I've kept working as the concept of 'subject' changed.

I'd prefer not to push any more log changes into this PR unless necessary.

packages/driver/src/cy/commands/querying/focused.ts
```diff
@@ -1,13 +1,13 @@
...
 export default (Commands, Cypress, cy, state) => {
-  Commands._addQuery('focused', function focused (options: Partial<Cypress.Loggable & Cypress.Timeoutable> = {}) {
+  Commands.addQuery('focused', function focused (options: Partial<Cypress.Loggable & Cypress.Timeoutable> = {}) {
     const log = options.log !== false && Cypress.log({ timeout: options.timeout })
 
     return () => {
       let $el = cy.getFocused()
 
-      log && log.set({
+      log && cy.state('current') === this && log.set({
```
chrisbreiding, line 12: Under what circumstance would `cy.state('current')` not equal `this` when this is invoked?

BlueWinds (commented):
Will address the rest of the comments soon, just wanted to address I've gotten to so far.

packages/driver/src/cypress/utils.ts
lmiller1990: Did we decided on the name we will use for the public API? I know we've got the concept of a `Chainable` in the public API, but would we be naming this `getSubjectFromChain` or something more concise like `cy.getSubject()`?

lmiller1990 (commented):
I got through the files. I'll give it a proper test tomorrow and report back. I think this is looking good!

packages/driver/src/cypress/command.ts
emilyrohrbough: Yeah, looks like it did. I has started to ask about some of the details but then started to double-check myself and I'm wondering if my original thoughts are valid. It looks like certain command are always re-executing the subject chain functions when the commands starts. Is that correct or is the requerying only happening on a retry if the subject is detached?

BlueWinds: That's correct; the subject is determined fresh at the start of every command or query. There are two reasons for this, principled and practical.

Consider the following test.

```
cy.$$('<button>Greetings.</button>').prependTo('body')

cy.get('button:first').as('btn')

setTimeout(() => {
  cy.$$('<button>Hello!</button>').prependTo('body')
}, 1000)

cy.wait(2000)

cy.get('@btn')
```

Should I get "Greetings" or "Hello!"? "Greetings" didn't detach from the DOM, so should we return it? But if it *had* detached - if we'd replaced it with a copy of itself before adding in "Hello", we'd get a different result!

So we always re-query for consistency's sake.

The more practical reason is so that I don't have to thread "the current subject" though the `command_queue` - we only need to track the subjectChain, not the subject *and* the subjectChain used to generate it.

packages/driver/src/cy/commands/querying/focused.ts
BlueWinds: When re-running the query as part of a subjectChain later on. It solves a case like this:

```
cy.get('button').click()

setTimeout(() => {
  cy.$$('button').addClass('active').blur()
}, 1000)

cy.focused().eq(0).should('have.class', 'active')
```

At the same time the button gets the active class, it blurs, so the query+assertion ('the first focused element should have the active class') never pass at the same time. However, we want `.focused()` to show one element resolved in the command log: There *was* a focused element when it first passed, when we took the DOM snapshot, before moving onto .eq(0); the fact that we later retried the query shouldn't erase this fact.

(this is a contrived example, but equivalent stuff shows up 'in the wild' quite regularly. I hadn't noticed the case until Brian pointed it out during one of our demos)

packages/driver/src/cypress/utils.ts
BlueWinds: So in this version there are a couple of methods on `cy`:

- `cy.subject()` - Resolves "what is the current subject" (optionally, for a chainerId other than the current one, but by default, just 'what's the subject right now').
- `cy.subjectChain()` - Returns the current subject chain (optionally, for a chainerId other than the current one).
- `cy.getSubjectFromChain()` - Takes a subject chain and resolves it to an actual value. No default behavior; you need a subjectChain to call this.

https://github.com/cypress-io/cypress/pull/24417/files#r1012215289 talks a little bit about why we need all three - but I think these names are both descriptive and short.

Open to suggestions if people have thoughts, but all three of them are needed, and I think each should be named with the other two in mind.

packages/driver/src/cypress/command.ts
emilyrohrbough: Is there impact on the overall test times by rerunning these every time?

packages/driver/types/internal-types-lite.d.ts
BlueWinds: Fixed the type here to properly use `SubjectChain`. Also added proper types for `subjects` and `subjectLinks` to state.ts, as I was poking around trying to understand our types.

I don't understand internal-types-lite.d.ts; some files seem perfectly happy to reference the $Cy class definition, while other files need the methods to be defined here as well. I added the types here so that `check-ts` would pass, but I do not understand why I did that. :woman_shrugging:

packages/launchpad/package.json
BlueWinds: The current version of cypress-testing-library continues to work - it doesn't break or throw any errors.

However, our tests were failing with that version because of this branch's more diligent checking for DOM updates. Specifically, we run into the following situation with our app tests:

1. `cy.get(someElement).scrollTo(somePlace)`.
2. Cypress starts scrolling, and Vue begins to re-render a virtualized list as the page scrolls.
3. `cy.getElementByTestId(somethingOrOther).click()` begins to execute. The *page isn't stable yet*. Our virtualized list has a nice fade-in effect, looks super cool. But this means that *the DOM continues to re-render even after `getElementByTestId` has resolved.
4. .click() waits for the page to stabilize, and the link to be actionable. But it never will be; the subject is detached from the DOM, and we can't retry the **command** `getElementByTestId`.

Whoops, detached DOM error - exactly what this branch was trying to solve. ^^;;

If users don't upgrade, they'll run into the same sort of issues, where some tests using cypress-testing-library that previously passed now fail - probably mostly those that run commands on animating elements.

There are two solutions. The traditional workarounds for Detached DOM errors - adding otherwise unnecessary .should() or .wait() commands can be used, but the better solution is to move cypress-testing-library to use queries, in which case 'it all just works', more reliably and less detached-DOMy than ever before, which is what I've done in my fork.

I could add the workarounds to get our tests passing rather than point to the updated cypress-testing-library, it wouldn't be a huge lift. But I wasn't fond of the idea of adding workarounds for the very bug I aim to solve. ^^;;

BlueWinds: Related: the fact that we find cypress-testing-library useful enough for our own tests to use it heavily indicates to me that we should make it first-party; at least maintain it ourselves, and at best integrate it directly into Cypress (not even as a plugin, just straight into the base Cy experience).

That's a different conversation, just something I've been thinking about recently. testing-library is a standard for a reason, and cypress-testing-library is a very thin wrapper around it; there wouldn't be a huge code burden, now that queries have made it extremely simple to implement.

packages/driver/src/cypress/error_messages.ts
emilyrohrbough: I'm confused, Is this failure associated with a command retry or a chained child command?

packages/driver/src/cypress/command.ts
BlueWinds: Good idea to measure, I last ran benchmarks a couple months ago, so it was fairly out of date.

New benchmark shows there does appear to be a cost.

```
it.only('is fast?', () => {

  const now = Date.now()

  for (let i = 0; i < 10000; i++) {
    cy.get('button', { log: false }).first({ log: false }).focus({ log: false })
  }

  cy.then(() => {
    cy.log('end: ' + (Date.now() - now) + 'ms')
  })
})
```

This executes 20,000 queries and 10,000 commands. `{log: false}` is used, because otherwise the snapshotting / command log completely swamps any possible change; that's by far the slowest part of any command.

I ran this test three times against each branch. In `develop`, this takes an average of 10.037s, while in `issue-7306-addQuery-actionability`, it takes an average of 10.615s. 

About 5% slower for executing the queries - but keep in mind how extremely focused the test is. My first attempt, with only 1000 iterations, didn't show any difference at all - the slowdown was below the threshold of detection.

It's a 0.02ms (not a typo) overhead per command, which is only measurable with tens of thousands of executions.

cli/types/cypress.d.ts
```diff
@@ -462,30 +465,98 @@ declare namespace Cypress {
...
       add<T extends keyof Chainable>(name: T, options: CommandOptions & {prevSubject: false}, fn: CommandFn<T>): void
+
+      /**
+       * Add a custom child command
+       * @see https://on.cypress.io/api/commands#Child-Commands
+       */
       add<T extends keyof Chainable, S = any>(name: T, options: CommandOptions & {prevSubject: true}, fn: CommandFnWithSubject<T, S>): void
+
+      /**
+       * Add a custom command that validates the prevSubject
```
emilyrohrbough: I thought this added a custom command that could be used as a parent or a chained child command?

packages/app/cypress/e2e/specs_list_e2e.cy.ts
```diff
@@ -245,8 +245,7 @@ describe('App: Spec List (E2E)', () => {
         cy.findByText('No specs matched your search:').should('not.be.visible')
       })
 
-      // TODO: fix flaky test https://github.com/cypress-io/cypress/issues/23305
-      it.skip('saves the filter when navigating to a spec and back', function () {
+      it('saves the filter when navigating to a spec and back', function () {
```
emilyrohrbough, line 248: 🎉

packages/driver/cypress/e2e/commands/actions/click.cy.js
```diff
@@ -2120,41 +2149,21 @@ describe('src/cy/commands/actions/click', () => {
         cy.get('.badge-multi').click()
       })
 
-      it('throws when subject is not in the document', (done) => {
```
emilyrohrbough, line 2123: Why was this test removed? I'd have expected the new error message to be verified here to suggest .click().click() should be broken up.

packages/driver/cypress/e2e/commands/actions/click.cy.js
```diff
@@ -3269,26 +3278,6 @@ describe('src/cy/commands/actions/click', () => {
         cy.dblclick()
       })
 
-      it('throws when subject is not in the document', (done) => {
```
emilyrohrbough, line 3272: same question, why remove?

packages/driver/cypress/e2e/commands/actions/click.cy.js
```diff
@@ -3708,26 +3697,6 @@ describe('src/cy/commands/actions/click', () => {
         cy.rightclick()
       })
 
-      it('throws when subject is not in the document', (done) => {
```
emilyrohrbough, line 3711: same question, why remove?

packages/driver/cypress/e2e/commands/actions/selectFile.cy.js
```diff
@@ -628,25 +654,6 @@ is being covered by another element:
         })
       })
 
-      it('waits until input stops animating', {
```
emilyrohrbough, line 631: Why did we remove this test?

cli/types/cypress.d.ts
BlueWinds: Hm. Looking at the docs more closely, it looks like this can be used to add *either* a child or a dual command - in both cases you pass in prevSubject. It's dual if it includes `optional`, and a child if it does not. Updating comment. :+1:

packages/driver/cypress/e2e/commands/commands.cy.js
```diff
@@ -90,13 +90,13 @@ describe('src/cy/commands/commands', () => {
...
       cy.on('fail', (err) => {
-        expect(err.message).to.eq('`Cypress.Commands._addQuery()` is used to create new queries, but `get` is an existing Cypress command or query, or is reserved internally by Cypress.\n\n If you want to override an existing command or query, use `Cypress.Commands.overrideQuery()` instead.')
+        expect(err.message).to.eq('`Cypress.Commands.addQuery()` is used to create new queries, but `get` is an existing Cypress command or query, or is reserved internally by Cypress.\n\n If you want to override an existing command or query, use `Cypress.Commands.overrideQuery()` instead.')
         expect(err.docsUrl).to.eq('https://on.cypress.io/custom-commands')
 
         done()
       })
 
-      Cypress.Commands._addQuery('get', () => {
+      Cypress.Commands.addQuery('get', () => {
```
emilyrohrbough, line 99: I thought we were holding off on exposing this API?

packages/driver/src/cypress/cy.ts
```diff
@@ -245,15 +245,17 @@ export class $Cy extends EventEmitter2 implements ITimeouts, IStability, IAssert
...
     this.addCommand = this.addCommand.bind(this)
-    this._addQuery = this._addQuery.bind(this)
+    this.addQuery = this.addQuery.bind(this)
     this.now = this.now.bind(this)
     this.onBeforeAppWindowLoad = this.onBeforeAppWindowLoad.bind(this)
     this.onUncaughtException = this.onUncaughtException.bind(this)
     this.setRunnable = this.setRunnable.bind(this)
     this.cleanup = this.cleanup.bind(this)
-    this.setSubjectForChainer = this.setSubjectForChainer.bind(this)
-    this.currentSubject = this.currentSubject.bind(this)
```
emilyrohrbough, line 255: +1 to change.

packages/driver/src/cypress/error_messages.ts
```diff
@@ -1936,7 +1957,7 @@ export default {
           > ${subjectChainToString(obj.subjectChain)}`
     },
     state_subject_deprecated: {
-      message: `${cmd('state', '\'subject\'')} has been deprecated and will be removed in a future release. Consider migrating to ${cmd('currentSubject')} instead.`,
+      message: `${cmd('state', '\'subject\'')} has been deprecated and will be removed in a future release. Consider migrating to ${cmd('subject')} instead.`,
```
emilyrohrbough, line 1960: Changing from cy.currentSubejct to cy.subject is breaking. Do we need to make a err message for those using who may have already updated to use cy.currentSubject to say it's been removed & they need to use cy.subject instead?

packages/driver/src/cypress/error_messages.ts
emilyrohrbough: Yeah I reviewed this from the bottom up & missed this was being moved to be in alphabetical order.

cli/types/cypress.d.ts
BlueWinds: The more I look at it, the more I think this type is actually just malformed.  This case covers the following:

```
.add('foo', { prevSubject: ['optional'] }, ...) // A dual command
.add('foo', { prevSubject: 'window' }, ...) // A child command that only allows one type of subject
```

vs. the one below it, which covers:

```
.add('foo', { prevSubject: ['optional'] }, ...) // A dual command - the same case as above.
.add('foo', { prevSubject: ['element', 'document'] }, ...) // A child command
.add('foo', { prevSubject: ['element', 'optional'] }, ...) // A dual command that can be either a parent or a child taking an element.
```

The `| ['optional']` in this line isn't necessary; the case below already covers it.

Updating both the type and the comment.

packages/driver/src/cypress/error_messages.ts
BlueWinds: This is associated with a failure during verifyUpcomingAssertions.

You do something like:

```
cy.get('button').click().should('have.class', 'active')
```

and the element gets detached *after clicking*, but while verifying `.should()`.

packages/launchpad/package.json
emilyrohrbough: Do you have a PR open to cypress-testing-library for these changes? Do we need an issue to update these dependencies once this is released & that version of cypress-testing-library is released?

packages/launchpad/cypress/e2e/top-nav-launchpad.cy.ts
emilyrohrbough: Will the be an obvious change to the cy.findByRole behavior for users upgrading? I wouldn't have expected this to be broken, even after reviewing these changes 😬

packages/driver/src/cypress/error_messages.ts
BlueWinds: I'd prefer to mention it in the changelog as breaking and not leave behind any cruft.

As far as I know, the only consumer of `state('subject')` (other than us) is cypress-testing-library, and they haven't updated yet.

packages/driver/cypress/e2e/commands/actions/click.cy.js
BlueWinds: We verify this case multiple times in several other files; this test was pure duplication.

packages/driver/cypress/e2e/commands/actions/click.cy.js
BlueWinds: Our driver codebase has a *ton* of duplicate code. Each command is tested as if it were entirely independent - but they're not, they share huge portions of code. Detached dom errors are part of $actionability and part of verifyUpcomingAssertions. We don't need to test it separately for every command that uses these two bits.

packages/driver/cypress/e2e/commands/actions/click.cy.js
BlueWinds: Same reason as above.

packages/driver/cypress/e2e/commands/actions/selectFile.cy.js
BlueWinds: This test was badly written (I'm not throwing shade, I wrote it when I was new at the company :D ); it mocks things unnecessarily, and duplicates  the tests around it. "Are we using actionability? Yes? Then we don't need to assert of every quirk of actionability in every command that uses it.

packages/driver/cypress/e2e/commands/commands.cy.js
BlueWinds: We were holding off doing it in 10.x - didn't want to release the API until we'd completed the migration and discovered any pitfalls. That was a concern while I was working in 'develop`, but I moved to a feature branch for unrelated reasons shortly after we made the decision.

My understanding is that we'll make it fully public in 12.x.

packages/launchpad/package.json
BlueWinds: My current plan is to get the docs PR into review and merge this one into `issue-7306`, then open a PR upstream.

The docs PR will be nice to help explain these changes - what and why, and merging into `issue-7306` will be nice because it'll build a binary I can point them to (so their automated tests can pass without manual massaging to point to a branch).

BlueWinds: Opened #24529 and #24530 to track that work.

packages/launchpad/cypress/e2e/top-nav-launchpad.cy.ts
BlueWinds: Same - after debugging I could see why it was failing, but it's not something I *expected* either. :grimacing: 

I don't think it's terribly obvious, but I also don't think it's avoidable. This is just the cost of fixing detached dom errors - if your DOM changes, Cypress will notice the change more reliably now.

cli/types/cypress.d.ts
BlueWinds: I reverted the change to the types; it started getting a bit larger than I expected. But still updated the comment. :+1:

packages/driver/src/cypress/utils.ts
lmiller1990: These names seem reasonable. Considering the only real consumers of these APIs are mostly us, and "power users", I think it makes sense they use the standard Cypress nomenclature.

packages/launchpad/package.json
lmiller1990: I think we should consider shipping the relevant parts of Testing Library as part of Cypress. It's pretty popular, will make upgrading less confusing, and avoid any more of these situations where we need an third party to update, and then for users to potentially update two dependencies (Cypress and Testing Library) in lockstep.

While a much needed improvement, the detached DOM changes are already going to be somewhat disruptive; anything we can do to minimize that is ideal.

I think this could also be useful for Cypress Studio, which uses the same selectors as [Cypress.SelectorPlayground](https://docs.cypress.io/api/cypress-api/selector-playground-api#Syntax). I think the `findByRole` and `findByLabel` etc from TL can be better than our `data-cy` and `data-testid` selectors (in some cases, anyway).

We should talk about this more internally, but I'm on board with this.

lmiller1990 (commented):
I left some comments (mostly questions) which were answered. I tried it out with some of the examples in https://github.com/cypress-io/cypress/issues/7306 and it seems to work great.

I don't have much else to add, but I don't think I'm deep enough in this PR for my approve to carry as much weight as the other reviewers, so I'm going to hold off ✅ for now.

Anything else I can do to move this forward @BlueWinds? Maybe kicking off the testing library discussion internally? Do you think we have bandwidth before 12.0 to do this?

BlueWinds:
> Anything else I can do to move this forward @BlueWinds? Maybe kicking off the testing library discussion internally? Do you think we have bandwidth before 12.0 to do this?

I should have the associated docs PR ready for review by end of day today, and have started talking with the community team about other plugins and how to handle working with the community over these changes. I hope to open a PR for `cypress-testing-library` by end of week.

With that said, I'd like to get some approvals on this and merge it in, so that we can see the results of the binary build in `issue-7306` and I can work on any new tests failing there. That's my next step - approvals and merging, while working on the less technical side of the release plan.

cli/types/cypress.d.ts
```diff
@@ -462,30 +465,98 @@ declare namespace Cypress {
...
+       * Add a custom query
+       * @see https://on.cypress.io/api/commands#Queries
+       */
+      addQuery<T extends keyof Chainable>(name: T, fn: QueryFn<T>): void
+
+      /**
+       * Overwrite an existing Cypress command or query with a new query
+       * @see https://on.cypress.io/api/commands#Overwrite-Existing-Queries
+       */
+      overwriteQuery<T extends keyof Chainable>(name: T, fn: QueryFn<T>): void
```
chrisbreiding:
```suggestion
```

This has been removed, right?

chrisbreiding (changes requested):

cli/types/cypress.d.ts
BlueWinds: Good catch. I'd only searched for `overwriteQuery` within the  `/packages/` folder, missed this one living outside it.

chrisbreiding (approved):

packages/driver/cypress/e2e/commands/actions/check.cy.js
flotwig: Yeah, but these tests aren't duplicates. They cover 95% of the same stuff but 5% is different, so I don't think it's a good idea to remove them completely. It doesn't look like the changes were needed for this PR, so can you revert them for now until we have a less duplicative way of testing `.uncheck`?

BlueWinds: They aren't 5% different, though. They're 5% different from any *one test* - but there's no 5% that they cover that nothing else does.

With that said, yes, I'll revert this if you feel strongly because I don't want to block this PR on unrelated changes.

flotwig: Yeah, please do. I also dislike the fact that this wastes time the way it is but we can't remove the canonical tests for a command without guarantees that it's still being tested (code coverage when?)

BlueWinds: I'd like to ask a bit more pointedly though: what specific test deleted here covers a specific thing that we'd lose coverage over? I didn't blindly delete all the `unchecked` tests - there are still quite a number of them, and the ones left cover that 5% you're talking about.

BlueWinds: Restored tests in commit 22b10f409934c43228ab96e0addde287245364be.

BlueWinds: It was not "all .uncheck() tests", it was "this test, and this test, and this other test" - I looked at the tests and the code they executed, and compared them one-by-one to the `.check()` tests, which is the same guarantee that we get with any other code change, when combined with PR reviews. :woman_shrugging:

flotwig: I'm not as familiar with these tests as you. But since we do switch some behavior between `check` and `uncheck`,  even though these tests still exist in `context('#check'`, it's not clear to me that we are actually not losing any coverage by removing them, given my limited powers of brain-powered static analysis. Since this PR isn't held back by having these tests, I figured we could just as well keep them for now, to avoid having to review 400 lines of removed tests in an unrelated PR.

I love deleting excess tests, though. Given that you feel confident we aren't losing any coverage here, feel free to request me on a follow-up PR to remove these. I'd be happy to give it a closer look instead of just delivering my immediate reaction. But I don't want to block this PR any longer with this.

flotwig (approved):

Issue #23688, opened 2022-09-06
Upgrade to Lerna 5+ with Nx


jordanpowell88:
### What would you like?

I would like to propose we upgrade the repo to use Lerna 5 and enable Nx. This will enable Nx's caching capabilities as well as bring it up to date with the most recent version of Lerna. 

https://blog.nrwl.io/lerna-used-to-walk-now-it-can-fly-eab7a0fe7700

### Why is this needed?

This will speed up our builds by leveraging Nx caching and update Lerna to be on the most recent version.

PR #24424, opened 2022-10-27
refactor: move linting from root to each lib 

jordanpowell88:
### User facing changelog
n/a

### Additional details
- Currently we lint the entire repo at the root level with one giant `.eslintignore` file. Ideally these are in individual libs to narrow the scope of executing the lint script.
- This also prefaces to be able to leverage caching these lint scripts in the future when we update to using Lerna 6. - [23688](https://github.com/cypress-io/cypress/issues/23688)

### Steps to test
`yarn lint`

n/a

### PR Tasks


npm/vite-dev-server/package.json
```diff
@@ -12,7 +12,8 @@
     "cypress:open": "yarn cypress:run-cypress-in-cypress gulp open --project .",
     "watch": "tsc -w",
     "test": "yarn test-unit",
-    "test-unit": "mocha -r ts-node/register/transpile-only --config ./test/.mocharc.js"
+    "test-unit": "mocha -r ts-node/register/transpile-only --config ./test/.mocharc.js",
+    "lint": "eslint --ext .js,.ts,.json, ."
```
emilyrohrbough, line 16: Most of these `.eslintignore` files are 1:1. Can you do `eslint --ignore-path -- ../../.eslintignore` when these are shared and using custom `.eslintignore` files when needed?

jordanpowell88: Certainly could be ignored from a shared `.eslintignore` file at the root. My first goal was two-fold:
- move to linting to individual the lib level
- avoid having a massive `.eslintignore` file at the root of the workspace

lmiller1990: What is the benefit of having lots of tiny files?

> Ideally these are in individual libs to narrow the scope of executing the lint script

Does this actually mean eslint will have to execute on less files? Is there some perf improvement here?

I don't really see the benefit of having **more** files - I actually prefer one file, so I can just add all my linting rules there. Right now every package has a slightly different set of rules, if anything wouldn't want to move towards 1 eslint file and one eslintignore file for everything?

lmiller1990:
All for making our linting less complex, but I'm not sure I see the benefit here - if anything, spreading things out seems like more complexity to manage? I could be missing something, though, like some perf or dependency benefits.

lmiller1990 (commented):
Quick comment, not sure I fully grok the benefit/improvement here.

npm/vite-dev-server/package.json
jordanpowell88: @emilyrohrbough @lmiller1990 The benefit is that later, when we update Lerna (to utilize Nx) we can utilize caching. This way we only have to lint thing that have been affected by changes. Right now we do this once at the root level and therefore we will have to lint essentially every single time on any change

emilyrohrbough: @jordanpowell88 to use nx we need the additional files or do we just need to lint command in each subpackage/workspace?

lmiller1990: I am not familiar with the intricacies of Nx caching - if this gets us closer to a better monorepo DX, happy to bikeshed my opinion of "all the linting rules in one file" in favor of moving forward.

I wonder if we can have some way to extend a base set of ignored files? Eg a `linting` package or something could expose it.

lmiller1990 (approved):
Going to approve, not the biggest fan of tons of little files but in the spirit of improving the monorepo DX, if this helps with that, I am on board.

npm/vite-dev-server/package.json
emilyrohrbough: Okay, followed up with @jordanpowell88. We don't actually need separate `. eslintignore ` files for each workspace to have it's own linting script & to use nx caching. Sounds like there are performance implications.

> if you change the root estlintignore file it would invalidate the cache and essentially require a re-run of everything to lint (not a huge deal if it doesn't happen often) but if I just change one line for one lib then it still affects performance

emilyrohrbough (approved):

flotwig (approved):

PR #24470, opened 2022-10-31
fix: remove last mounted component upon subsequent mount calls

### LINKED ISSUE #24329: cy.mount should remove the last mounted component if called multiple times in same test
### Current behavior

Scenario:

```ts
it('calls mount twice', () => {
  cy.mount(MyApp)
  cy.mount(MyApp)
})
```

Framework dependent behavior:

- Angular: Throws error upon second mount
- Vue: Mounts multiple components
- Svelte: Mounts multiple components
- React: Removes and remounts
- Vue2: Mounts multiple components

### Desired behavior

The last component mounted should be removed from the DOM/cleaned up.

### Test code to reproduce

```ts
import HelloWorld from "./components/HelloWorld.vue";

it("default", () => {
  cy.mount(HelloWorld, { props: { msg: "Component 1" } });
  cy.mount(HelloWorld, { props: { msg: "Component 2" } });
  cy.mount(HelloWorld, { props: { msg: "Component 3" } });
});
```

Screen Shot 2022-10-20 at 5 11 51 PM

### Cypress Version

10.10.0

### Node version

v16.17.1

### Operating System

macOS 12.4

### Debug Logs

_No response_

### Other

_No response_

ZachJW34:
- Closes #24329

### User facing changelog
Subsequent `mount` calls within the same test will remove the last mounted component from the DOM

### Additional details
Calling `mount` multiple times in a single test had different behavior between frameworks. Some (properly) removed the element, some mounted multiple components and another threw an error.

This PR makes all of the frameworks behave consistently by destroying/removing the last component mounted if `mount` is called again within a test.

### Steps to test
I relied on system-tests to test these changes, specifically the tests in `component_testing_spec.ts`. I've added/tweaked the `mount.cy.{ts,js,tsx,jsx}` tests for `react,vue,svelte,angular` to verify the behavior.

https://user-images.githubusercontent.com/25158820/199060873-d455a14b-7b4f-4da1-ba47-1c8f652cb45c.mov

### PR Tasks


npm/angular/src/mount.ts
```diff
@@ -72,6 +72,15 @@ export interface MountConfig<T> extends TestModuleMetadata {
   componentProperties?: Partial<{ [P in keyof T]: T[P] }>
 }
 
+let activeFixture: ComponentFixture<any> | null = null
+
+function cleanup () {
+  activeFixture = null
+  // Not public, we need to call this to remove the last component from the DOM
+  getTestBed()['tearDownTestingModule']()
```
lmiller1990: Do you think we need to be a bit defensive here? Eg

```js
const teardown = getTestBed()['tearDownTestingModule']
if (!teardown) {
  throw Error('Could not teardown component. The version of Angular you are using may not be officially supported').
  // OR just return early?
  return
}

teardown()
```

lmiller1990: Surprised TS isn't erroring here with "might be undefined" - maybe we should make the TS setting more strict.

system-tests/project-fixtures/svelte/src/mount.cy.js
```diff
@@ -83,4 +83,25 @@ describe("Svelte mount", () => {
...
+
+  context('teardown', () => {
+    beforeEach(() => {
+      // component-index.html has anchor element within [data-cy-root] so base # of elements is 1
+      cy.get('[data-cy-root]').children().should('have.length', 1)
+    })
+
+    it("should mount", () => {
+      cy.mount(Counter);
+    });
```
lmiller1990:
```suggestion
    it('should mount', () => {
      cy.mount(Counter);
    });
```

system-tests/project-fixtures/vue/cypress.config.js
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from 'cypress'
```
lmiller1990, line 1: Do we need an entirely new project for this? Any chance we can add the test to an existing project?

system-tests/project-fixtures/vue/src/style.css
```diff
@@ -0,0 +1,90 @@
+:root {
+  font-family: Inter, Avenir, Helvetica, Arial, sans-serif;
+  font-size: 16px;
+  line-height: 24px;
```
lmiller1990: Seems like a lot of extra code in this project that isn't necessary for the tests?

lmiller1990 (commented):
Few comments, code seems good

system-tests/project-fixtures/vue/cypress.config.js
ZachJW34: I like the pattern of having a `project-fixture` project that can cover multiple versions of a framework we support and a `mount.cy.js` that can test the behavior of the mount for that framework. We currently do not have this for `vue` so writing a test that can be shared is not possible. I def think we have too many of these projects and no clear pattern across them, but I'm bullish on this pattern. WDYT?

system-tests/project-fixtures/vue/src/style.css
ZachJW34: Will cleanup!

npm/angular/src/mount.ts
ZachJW34: I like the idea of throwing an error here. Since this is something that might break in a minor, it's best to fail early and loud. The logic in `tearDownTestingModule` is important to guarantee clean state for the next tests so I'm ok with erroring in this case.

system-tests/project-fixtures/vue/src/style.css
ZachJW34: Addressed in [ce2c587](https://github.com/cypress-io/cypress/pull/24470/commits/ce2c5876323a89d9600bfc6bb21e539d7d9bded3)

system-tests/project-fixtures/svelte/src/mount.cy.js
ZachJW34: Addressed in [ce2c587](https://github.com/cypress-io/cypress/pull/24470/commits/ce2c5876323a89d9600bfc6bb21e539d7d9bded3)

npm/angular/src/mount.ts
ZachJW34: Addressed in [ce2c587](https://github.com/cypress-io/cypress/pull/24470/commits/ce2c5876323a89d9600bfc6bb21e539d7d9bded3), also made the types more strict for `npm/angular`

lmiller1990 (approved):
LGTM

npm/angular/src/mount.ts
```diff
@@ -72,6 +72,20 @@ export interface MountConfig<T> extends TestModuleMetadata {
...
 }
 
+let activeFixture: ComponentFixture<any> | null = null
+
+function cleanup () {
+  // Not public, we need to call this to remove the last component from the DOM
+  try {
+    (getTestBed() as any).tearDownTestingModule()
+  } catch (e) {
+    throw new Error(`Failed to teardown component. The version of Angular you are using may not be officially supported, visit https://docs.cypress.io/guides/component-testing/component-framework-configuration for current Angular version support.`)
```
rockindahizzy: Should this use an `on` link?

ZachJW34: Damn you `on`Links!!!

ZachJW34:
@rockindahizzy added the onlinks here: https://github.com/cypress-io/cypress-services/pull/4937

Error thrown
Screen Shot 2022-11-02 at 4 05 47 PM

npm/react/src/mount.ts
```diff
@@ -33,6 +33,9 @@ export function mount (jsx: React.ReactNode, options: MountOptions = {}, rerende
     Cypress.log({ name: 'warning', message })
   }
 
+  // Remove last mounted component if cy.mount is called more than once in a test
```
marktnoonan, line 36: Is this comment only applicable in React 18 or should it be here as well?

>  // React by default removes the last component when calling render, but we should remove the root
  // to wipe away any state

npm/vue/src/index.ts
```diff
@@ -72,24 +72,12 @@ type MountingOptions<Props, Data = {}> = Omit<VTUMountingOptions<Props, Data>, '
...
-  // https://github.com/cypress-io/cypress/issues/17438
-  if (Cypress.testingType !== 'component') {
-    return
-  }
-
-  Cypress.on('test:before:run', () => {
-    Cypress.vueWrapper?.unmount()
-    const el = getContainerEl()
+const cleanup = () => {
+  Cypress.vueWrapper?.unmount()
```
marktnoonan, line 77: If there's no `Cypress.vueWrapper` that would indicate it's the first mount, so possibly we just return here instead of touching the DOM?

Actually wondered, as part of cleanup do we want to unset `Cypress.vueWrapper` and `Cypress.vue`? Probably pointless since mount will overwrite them & cleanup is always called before mount.

npm/vue2/src/index.ts
```diff
@@ -266,6 +265,10 @@ declare global {
   }
 }
 
+const cleanup = () => {
+  Cypress.vueWrapper?.destroy()
```
marktnoonan, line 269: Any need to do `innerHTML = ''` for the root el here? Curious why Vue 3 does it but Vue 2 does not

marktnoonan (approved):
Did manual testing Vue, works great.

Looking at Percy, we have 3 snapshots that rely on `mount` not being cleaned up, to catch all the variations at once. I actually like this behavior from a billing perspective lol.

Not blocking, I'll make a PR into with suggested updates to these tests into this branch if it is still open at the time, or into the freature branch if we've already merged this.

A few other small observations/questions that don't block merging.

Screen Shot 2022-11-02 at 9 40 17 PM

Screen Shot 2022-11-02 at 9 45 46 PM

Screen Shot 2022-11-02 at 9 46 56 PM

rockindahizzy (approved):

npm/react/src/mount.ts
ZachJW34: Only applicable to 18, react17 doesn't have the concept of `root` but just a rerender into the same element.

npm/vue/src/index.ts
ZachJW34: I'm going to remove the `innerHtml = ''`, more consistent with other frameworks. I'll add the cleanup as well.

npm/vue2/src/index.ts
ZachJW34: See https://github.com/cypress-io/cypress/pull/24470#discussion_r1012960512

PR #24471, opened 2022-10-31
feat: Add domain option to cookie commands

### LINKED ISSUE #24264: Update cookie commands to take domain into account
With the addition of cy.origin and cy.session we should 

1. update cy.cookie apis to allow users to pass in the domain as an option
2. update the cy.cookie apis default to use the AUT domain as default if available

Spike into what options we have when setting cookies and the AUT is undefined.

This should affect getCookies, getCookie, setCookies, clearCookie and clearCookies.

chrisbreiding:
- Closes #24264

### User facing changelog

- `cy.getCookie()`, `cy.getCookies()`, `cy.setCookie()`, `cy.clearCookie()`, and `cy.clearCookies()` now accept a `domain` option

### Additional details

Some of the commands already utilized a `domain` option, but it wasn't tested or a part of the types. I also did a small amount of cleanup and refactoring of each command's implementation.

### PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4824


packages/driver/cypress/e2e/commands/cookies.cy.js
```diff
@@ -454,8 +700,7 @@ describe('src/cy/commands/cookies', () => {
     })
   })
 
-  // TODO: fix flaky test https://github.com/cypress-io/cypress/issues/23444
```
chrisbreiding, line 457: This and subsequent flake has been fixed since https://github.com/cypress-io/cypress/pull/23442, just missed unskipping these tests.

packages/driver/src/cy/commands/cookies.ts
```diff
@@ -79,28 +79,16 @@ function cookieValidatesSecurePrefix (options) {
   return options.secure === false
 }
 
-interface InternalGetCookieOptions extends Partial<Cypress.Loggable & Cypress.Timeoutable> {
```
chrisbreiding, line 82: I refactored out the need for these internal types by using locally scoped variables instead of mutating the `options` objects.

packages/runner/injection/patches/cookies.ts
```diff
@@ -63,6 +63,17 @@ export const patchDocumentCookie = (requestCookies: AutomationCookie[]) => {
     })
   }
 
+  const setCookie = (cookie: ToughCookie | string) => {
```
chrisbreiding, line 66: This is a (somewhat) incidental fix that I discovered while working on `cy.setCookie()`. The `document.cookie` patch should not throw an error when attempting to set a cookie with a different domain, which matches the behavior of the real `document.cookie`.

cli/types/cypress.d.ts
```diff
@@ -1252,14 +1252,14 @@ declare namespace Cypress {
...
-    getCookie(name: string, options?: Partial<Loggable & Timeoutable>): Chainable<Cookie | null>
+    getCookie(name: string, options?: Partial<CookieOptions>): Chainable<Cookie | null>
 
     /**
      * Get all of the browser cookies.
      *
      * @see https://on.cypress.io/getcookies
      */
-    getCookies(options?: Partial<Loggable & Timeoutable>): Chainable<Cookie[]>
+    getCookies(options?: Partial<CookieOptions>): Chainable<Cookie[]>
```
emilyrohrbough: Based on conversations earlier today, do we want to hold off on allowing a domain option for getCookies() such that is only returns the current dommain & we adding a new command `cy.getAllCookies()`? So far, I'm personally in favor of using one command with options vs adding a new command of this approach but see the value in the return structure. 🤷🏻‍♀️

chrisbreiding: We still want `cy.getCookies()` to accept a domain option, because by default it uses the AUT domain. Users need a way to specify a different domain. The domain option also gives it parity with the other cookie commands. 

I don't think `cy.getAllCookies()` will take a domain option, so it wouldn't provide the same functionality. The only overlap is `cy.getCookies({ domain: null })` will essentially be the same as `cy.getAllCookies()`. I think I might just leave `domain: null` undocumented for now, since it will be made obsolete by `cy.getAllCookies()`.

That said, we didn't really finish that discussion about this and I'm still not 100% convinced we need a `cy.getAllCookies()`, so maybe we'll hash that out some more before we start implementing it.

emilyrohrbough: +1 to hashing it out before doing the work.

packages/driver/cypress/e2e/commands/cookies.cy.js
```diff
@@ -2,21 +2,317 @@ const { assertLogLength } = require('../../support/utils')
...
+
+        cy.getCookies({ domain: null }).should('have.length', 2)
+      })
+    })
+  })
+
+  context('#getCookie', () => {
+    it('returns the cookie from the domain matching the AUT by default', () => {
+      cy.visit('http://localhost:3500/fixtures/generic.html')
+      cy.setCookie('foo', 'bar', { domain: 'www.foobar.com' })
```
AtofStryker, line 87: should we also test domain hierarchy here as well, like having the following cookie and asserting it's returned here?

```
cy.setCookie('bar', 'baz', { domain: 'foobar.com' })
```

packages/driver/cypress/e2e/commands/cookies.cy.js
```diff
@@ -2,21 +2,317 @@ const { assertLogLength } = require('../../support/utils')
...
+
+        cy.getCookie('foo', { domain: null })
+        .its('domain').should('equal', 'localhost')
+      })
+    })
+  })
+
+  context('#setCookie', () => {
+    it('sets the cookie on the domain matching the AUT by default', () => {
+      cy.visit('http://localhost:3500/fixtures/generic.html')
```
AtofStryker: I wonder if using a different domain over `localhost` like `www.foobar.com` might make for a better test because we can assert that the current domain should equal `www.foobar.com` and not `foobar.com` to help detect drifts from the cookie spec.

packages/driver/cypress/e2e/commands/cookies.cy.js
```diff
@@ -2,21 +2,317 @@ const { assertLogLength } = require('../../support/utils')
...
+        .its('domain').should('equal', 'localhost')
+      })
+    })
+  })
+
+  context('#clearCookies', () => {
+    it('clears all cookies', () => {
+      cy.setCookie('foo', 'bar')
+      cy.getCookies().should('have.length', 1)
+      cy.clearCookies()
```
AtofStryker, line 180: does `clearCookies` only clear cookies on the current AUT domain, or does it have the effect of `clearAllCookies`? If the later we might want to add other `setCookie` on different domains to further verify the test, or even if only clearing the cookies on the current AUT domain assert that the set cookies on different domains still exist.

packages/driver/cypress/e2e/commands/cookies.cy.js
```diff
@@ -2,21 +2,317 @@ const { assertLogLength } = require('../../support/utils')
...
+        cy.getCookie('foo').should('exist')
+      })
+    })
+
+    it('clears cookies for all domains when domain is null', () => {
+      cy.visit('http://localhost:3500/fixtures/generic.html')
+      cy.setCookie('foo', 'bar')
+      cy.setCookie('foo', 'bar', { domain: 'www.foobar.com' })
+
+      cy.clearCookies({ domain: null })
```
AtofStryker, line 243: I think this answers my question about the `clearCookie` implicit domain 😄

packages/driver/src/cy/commands/cookies.ts
```diff
@@ -79,28 +79,16 @@ function cookieValidatesSecurePrefix (options) {
...
-}
-
-type InternalClearCookieOptions = InternalGetCookieOptions
-
-interface InternalClearCookiesOptions extends Partial<Cypress.Loggable & Cypress.Timeoutable> {
-  _log?: Log
-  cookies?: Cypress.Cookie[]
-  domain?: any
+function validateDomainOption (domain: any, commandName: string, log: Log | undefined) {
+  if (domain != null && !_.isString(domain)) {
```
AtofStryker: might make the comparison clearer at first glance?
```suggestion
  if (domain !== null && domain !== undefined && !_.isString(domain)) {
```

packages/server/lib/browsers/webkit-automation.ts
```diff
@@ -322,12 +298,19 @@ export class WebKitAutomation {
    * Clear all cookies
    * @returns cookies cleared
    */
-  private async clearCookies (): Promise<CyCookie[]> {
-    const allCookies = await this.getCookies()
+  private async clearCookies (cookiesToClear: CyCookie[]): Promise<CyCookie[]> {
+    const allCookies = await this.context.cookies()
+    const persistCookies = allCookies.filter((cookie) => {
```
AtofStryker, line 307: any benefit to adding debug logs here to what cookies are cleared that fit the filter?

packages/driver/cypress/e2e/commands/cookies.cy.js
```diff
@@ -2,21 +2,317 @@ const { assertLogLength } = require('../../support/utils')
...
+
+      cy.getCookie('foo').its('domain').should('equal', 'localhost')
+
+      if (isWebkit || !Cypress.config('experimentalSessionAndOrigin')) return
+
+      cy.origin('http://www.foobar.com:3500', () => {
+        cy.visit('http://www.foobar.com:3500/fixtures/generic.html')
+        cy.setCookie('foo', 'bar')
+
+        cy.getCookie('foo').its('domain').should('equal', '.foobar.com')
```
AtofStryker, line 153: should this be `www.foobar.com` for default domain? I thought we had a story to change the default superDomain but haven't found it. maybe we closed it?

packages/driver/cypress/e2e/commands/cookies.cy.js
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/24471/commits/f740a20961f4f4631f7b45fb98617eeb0e1fc3fc

packages/driver/cypress/e2e/commands/cookies.cy.js
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/24471/commits/bfd2081e03af6e966bd485ab2c1a54de7a6fa2a6

packages/driver/cypress/e2e/commands/cookies.cy.js
chrisbreiding: It clears them on the current superdomain. That's existing behavior, so I didn't change it. 

I didn't add this test, just moved it, so I figured I would leave alone. The tests I added cover different domains.

packages/driver/src/cy/commands/cookies.ts
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/24471/commits/89f51cc73224701fc0cdc190fdadbf5290fd1d56

packages/server/lib/browsers/webkit-automation.ts
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/24471/commits/42ae0a13e96ae4ccfe73b0b8579cf9875fa68321

packages/driver/cypress/e2e/commands/cookies.cy.js
chrisbreiding: The default domain is the superdomain of the AUT, which is existing behavior for all cookie commands, so this is correct. I don't know of a story to change that, but it would be a breaking change. I think it should be a separate PR if we do decide to do that, since this can go straight into a release as-is.

AtofStryker: Agreed. I'm thinking we should do this work to be consistent with cookie behavior, but not into `develop` since it does break the API. Just want to make sure we capture that someplace. I created #24526 and added it into the next sprint to target `v12.0.0`

AtofStryker: @chrisbreiding #24526 should be a relatively simple change right?

packages/driver/cypress/e2e/commands/cookies.cy.js
```diff
@@ -7,13 +7,13 @@ const isWebkit = Cypress.isBrowser('webkit')
...
     it('returns cookies from the domain matching the AUT by default', () => {
-      cy.visit('http://localhost:3500/fixtures/generic.html')
+      cy.visit('http://www.barbaz.com:3500/fixtures/generic.html')
       cy.setCookie('foo', 'bar', { domain: 'www.foobar.com' })
       cy.setCookie('baz', 'qux', { domain: 'foobar.com' })
       cy.setCookie('foo', 'bar')
 
       cy.getCookies().then((cookies) => {
-        expect(cookies[0].domain).to.equal('localhost')
+        expect(cookies[0].domain).to.match(/\.?.barbaz\.com/)
```
AtofStryker: if I did
```
cy.setCookie('qux', 'quuz',  { domain: 'www.barbaz.com'  })
```

and then did `cy.getCookies()` would the `qux=quux` show up? I'm guessing the user needs to be specific with the domain in this case since the default is currently super domain?

packages/driver/cypress/e2e/commands/cookies.cy.js
chrisbreiding: Yeah it's just changing [this one line](https://github.com/cypress-io/cypress/blob/b595636d061d138cccc7d9a1e5f965145e995a6b/packages/driver/src/cy/commands/cookies.ts#L35) plus updating tests as needed.

packages/driver/cypress/e2e/commands/cookies.cy.js
chrisbreiding: Yes, it does get returned because of [we match based on superdomain](https://github.com/cypress-io/cypress/blob/b595636d061d138cccc7d9a1e5f965145e995a6b/packages/server/lib/browsers/cdp_automation.ts#L65). I [updated the getCookies tests](https://github.com/cypress-io/cypress/pull/24471/commits/2277d99fa555f8fadba84269894a5c642252efbb) to check for that.

AtofStryker: OK so that also means cookies from `app.barbaz.com` would get returned, which seems consistent with itself. We can scope that work to #24526

AtofStryker (approved):

packages/driver/src/cy/commands/cookies.ts
```diff
@@ -79,28 +79,16 @@ function cookieValidatesSecurePrefix (options) {
...
-}
-
-type InternalClearCookieOptions = InternalGetCookieOptions
-
-interface InternalClearCookiesOptions extends Partial<Cypress.Loggable & Cypress.Timeoutable> {
-  _log?: Log
-  cookies?: Cypress.Cookie[]
-  domain?: any
+function validateDomainOption (domain: any, commandName: string, log: Log | undefined) {
+  if (domain !== undefined && domain !== null && !_.isString(domain)) {
```
emilyrohrbough, line 83: I am wondering, do we need to add stricter logic to ensure the passes string matches domain patterns? I assume this should not include the schema? or the subdir aspects of a url?

cli/types/cypress.d.ts
```diff
@@ -2633,6 +2633,13 @@ declare namespace Cypress {
     cmdKey: boolean
   }
 
+  interface CookieOptions extends Loggable, Timeoutable {
+    /**
+     * Domain to set cookies on or get cookies from
+     */
+    domain: string
```
emilyrohrbough: Need to be optional to be passive right? 
```suggestion
    domain?: string
```

cli/types/cypress.d.ts
```diff
@@ -2633,6 +2633,13 @@ declare namespace Cypress {
     cmdKey: boolean
   }
 
+  interface CookieOptions extends Loggable, Timeoutable {
+    /**
+     * Domain to set cookies on or get cookies from
```
emilyrohrbough, line 2638:
```suggestion
     * Domain to set cookies on or get cookies from. When domain is not provided, the current domain is used.
```

cli/types/cypress.d.ts
chrisbreiding: All the commands accept `Partial<CookieOptions>`, so it's not necessary to mark it as optional here.

packages/driver/src/cy/commands/cookies.ts
```diff
@@ -192,27 +180,27 @@ export default function (Commands, Cypress, cy, state, config) {
...
+    getCookie (name, userOptions: Partial<Cypress.CookieOptions> = {}) {
+      const options: Partial<Cypress.CookieOptions> = _.defaults({}, userOptions, {
         log: true,
       })
 
       const responseTimeout = options.timeout || config('responseTimeout')
 
       options.timeout = options.timeout || config('defaultCommandTimeout')
 
+      let cookie: Cypress.Cookie
```
emilyrohrbough, line 192: or am I missing where this was pulled out? In the log logic, looks like cookie will never be set atm
```suggestion
      const cookie: Cypress.Cookie = option.cookie
```

emilyrohrbough: NVM...thats the console props 🤦🏻‍♀️ its in the automation request below.

packages/driver/src/cy/commands/cookies.ts
chrisbreiding: I considered that because I remember we originally did that with cy.origin()'s `origin` option, but checked and we now don't do anything more than string validation. I think it's just not worth the effort to validate, since top-level domains, localhost, ip address (ipv4 or ipv6) need to be take into account. Worst case scenario is the user doesn't get the cookies they expect and it should be obvious why if they pass `{ domain: 'non$ens3' }`.

packages/driver/src/cy/commands/cookies.ts
```diff
@@ -297,34 +281,31 @@ export default function (Commands, Cypress, cy, state, config) {
...
       }
 
-      const onFail = options._log
-
       cookie.sameSite = normalizeSameSite(cookie.sameSite)
 
       if (!_.isUndefined(cookie.sameSite) && !VALID_SAMESITE_VALUES.includes(cookie.sameSite)) {
         $errUtils.throwErrByPath('setCookie.invalid_samesite', {
-          onFail,
+          onFail: log,
```
emilyrohrbough, line 308: This continues the existing log, but do we need to pass the log to onFail? Wondering if the commandQueue's fail command logic logs these error correctly for the cookie log.

packages/driver/src/cy/commands/cookies.ts
chrisbreiding: This is set below by the result of `automateCookies('get:cookie')`. There is no `options.cookie`.

packages/driver/src/cy/commands/cookies.ts
```diff
@@ -336,61 +317,61 @@ export default function (Commands, Cypress, cy, state, config) {
...
-        }).catch(handleBackendError('setCookie', 'setting the requested cookie in', onFail))
+        .tap((result) => {
+          resultingCookie = result
+        }).catch(handleBackendError('setCookie', 'setting the requested cookie in', log))
       }, options.timeout)
     },
 
-    clearCookie (name, userOptions: Partial<Cypress.Loggable & Cypress.Timeoutable> = {}) {
-      const options: InternalClearCookieOptions = _.defaults({}, userOptions, {
+    clearCookie (name, userOptions: Partial<Cypress.CookieOptions> = {}) {
```
emilyrohrbough: I think my above suggestion to make domain option would remove the need for `Partial<Cypress.CookieOptions>` throughout since each option should be optional

packages/driver/src/cypress/error_messages.ts
```diff
@@ -325,6 +325,12 @@ export default {
         docsUrl: 'https://on.cypress.io/session',
       }
     },
+    invalid_domain (obj) {
+      return {
+        message: `${cmd('{{cmd}}')} must be passed a valid domain name. You passed: \`{{domain}}\``,
+        docsUrl: `https://on.cypress.io/${_.toLower(obj.cmd)}`,
```
emilyrohrbough, line 331: only a thought -- I know we had talked about documenting domain / super domain / etc at some point with the origin work. Might be worth pointing to that doc some day to give better reference to whats an appropriate domain.

packages/driver/src/cy/commands/cookies.ts
chrisbreiding: This was the existing behavior, I just removed the part where it assigns `log` to the variable `onFail` beforehand, so there's no change in behavior.

packages/runner/injection/patches/cookies.ts
```diff
@@ -63,6 +63,17 @@ export const patchDocumentCookie = (requestCookies: AutomationCookie[]) => {
...
+    try {
+      return cookieJar.setCookie(cookie, url, undefined)
+    } catch (err) {
```
emilyrohrbough, line 69: Should we propagate this error so users are aware it wasn't properly set?

cli/types/cypress.d.ts
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/24471/commits/d4e600dd589e1a6ff6b6bfcf25ca1d193a3c8fc0

packages/server/lib/browsers/webkit-automation.ts
```diff
@@ -322,12 +298,22 @@ export class WebKitAutomation {
...
+      })
+    })
+
+    debug('clear cookies: %o', cookiesToClear)
+    debug('put back cookies: %o', persistCookies)
 
     await this.context.clearCookies()
 
-    return allCookies
+    if (persistCookies.length) await this.context.addCookies(persistCookies)
```
emilyrohrbough, line 318: I think this description needs updated. to Clears all cookies in domain or clears all the provided cookies?

packages/driver/src/cy/commands/cookies.ts
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/24471/commits/d4e600dd589e1a6ff6b6bfcf25ca1d193a3c8fc0

packages/driver/src/cypress/error_messages.ts
chrisbreiding: Yep 😄 https://github.com/cypress-io/cypress-documentation/pull/4824#discussion_r1012160257

cli/types/cypress.d.ts
```diff
@@ -780,15 +780,15 @@ declare namespace Cypress {
      *
      * @see https://on.cypress.io/clearcookie
      */
-    clearCookie(name: string, options?: Partial<Loggable & Timeoutable>): Chainable<null>
+    clearCookie(name: string, options?: Partial<CookieOptions>): Chainable<null>
 
     /**
      * Clear all browser cookies.
      * Cypress automatically clears all cookies before each test to prevent state from being shared across tests. You shouldn't need to use this command unless you're using it to clear a specific cookie inside a single test.
```
```suggestion
     * Cypress automatically clears all cookies before each test to prevent state from being shared across tests. You shouldn't need to use this command unless you're using it to clear a all cookies or a set of cookies inside a single test.
```

packages/runner/injection/patches/cookies.ts
chrisbreiding: No, it should be ignored. That's what the real `document.cookie` would do.

cli/types/cypress.d.ts
emilyrohrbough: When domain is not specified, does this return cookies from all domains?

cli/types/cypress.d.ts
```diff
@@ -780,15 +780,15 @@ declare namespace Cypress {
...
      *
      * @see https://on.cypress.io/clearcookies
      */
-    clearCookies(options?: Partial<Loggable & Timeoutable>): Chainable<null>
+    clearCookies(options?: Partial<CookieOptions>): Chainable<null>
```
emilyrohrbough: When domain is not specified, does this clear cookies in all domains?

packages/server/lib/browsers/webkit-automation.ts
chrisbreiding: It's meant to clear certain cookies, usually for the domain specified. However, webkit only has a "clear all cookies" method, so we have to clear all the cookies and put back the ones we didn't want cleared. I added comments in https://github.com/cypress-io/cypress/pull/24471/commits/ad2f29435695ab144ff40fed3c85eac3ad1f1cdb to clarify this.

cli/types/cypress.d.ts
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/24471/commits/648449c6c507392bcee2a43f5cfba3d3225307c0

cli/types/cypress.d.ts
chrisbreiding: No, current superdomain. Updated the comment in https://github.com/cypress-io/cypress/pull/24471/commits/648449c6c507392bcee2a43f5cfba3d3225307c0.

cli/types/cypress.d.ts
chrisbreiding: No, current superdomain. Updated the comment in https://github.com/cypress-io/cypress/pull/24471/commits/648449c6c507392bcee2a43f5cfba3d3225307c0.

cli/types/cypress.d.ts
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/24471/commits/73bb4c36d60c0418316b06107ef21f2dd3f1804a

packages/driver/src/cypress/error_messages.ts
emilyrohrbough: OHH i was actually thinking more of a "what is domian vs superdomain vs url schema type breakdown" 😅

packages/runner/injection/patches/cookies.ts
emilyrohrbough: Sure, I just seem like if that' be obvious from the commands that tried to set this

emilyrohrbough (approved):

packages/driver/src/cypress/error_messages.ts
chrisbreiding: Exactly. Once Bill writes that doc we can point to it here and in the docs for this option.

PR #24478, opened 2022-10-31
chore: Breakup large spec file to avoid chrome crash

BlueWinds:
### User facing changelog
No user facing changes

### Additional details
Cypress currently face out-of-memory crashes with Chrome in large spec files. Users run into this frequently, and we do as well. This PR doesn't address the root cause, but merely breaks up one of our larger test files to avoid the symptoms.

There are no changes to the tests in question - I just broke up a large file into two parts.

### PR Tasks


rachelruderman (approved):
Nice idea, let's try it!

packages/app/cypress/e2e/runner/reporter.command_errors.cy.ts
```diff
@@ -0,0 +1,512 @@
...
+  // Return scoped verify function with spec options baked in
+  return createVerify({ fileName: Cypress._.last(filePath.split('/')), hasPreferredIde, mode })
+}
+
+describe('errors ui', {
+  viewportHeight: 768,
+  viewportWidth: 1024,
+  // Limiting tests kept in memory due to large memory cost
+  // of nested spec snapshots
+  numTestsKeptInMemory: 1,
```
emilyrohrbough: Might as well as just set this to 0.

BlueWinds: Done.

BlueWinds: The test file was still huge, so I like splitting it up even if we're not crashing chrome with `numTestsKeptInMemory: 0`.

BlueWinds:
A couple of flaky tests in here unrelated to the changes; I don't want to block a flake fix on *other* flake fixes though, so if I get a second approval I will merge without a green build.

chrisbreiding (approved):

PR #24493, opened 2022-11-01
feat: add reject unauthorized to api server calls and standardize CA usage

### LINKED ISSUE #23980: Ensure that calls to `https://api.cypress.io/` will reject if the server certificate does not match
### What would you like?

Currently, we [set](https://github.com/cypress-io/cypress/blob/develop/packages/server/lib/request.js#L21) `NODE_TLS_REJECT_UNAUTHORIZED` to '0' to ensure that the proxying that Cypress does isn't affected by this security. This requirement is not necessary for calls to `https://api.cypress.io/`.

### Why is this needed?

We should ensure that we only lax security requirements when required.

### Other information

We should be able to [set](https://nodejs.org/api/https.html#httpsrequesturl-options-callback) `requestUnauthorized` on the requests to the api and it will override the environment variable.

ryanthemanuel:
- Closes #23980

### User facing changelog

Communication with the Cypress dashboard will now verify CAs and reject on unauthorized calls. If you use a self-signed CA you will need to set `npm_config_ca`, `npm_config_cafile`, or `NODE_EXTRA_CA_CERTS`

### Additional details

To reject the api dashboard calls, we will pass in `rejectUnauthorized` on requests specifically to `https://api.cypress.io/`

By enabling security again for these calls, we potentially run into trouble with self signed corporate certificates. To resolve these issues, we will use the same config options that npm and node use:

- `npm_config_ca` - https://docs.npmjs.com/cli/v8/using-npm/config#ca
- `npm_config_cafile` - https://docs.npmjs.com/cli/v8/using-npm/config#cafile
- `NODE_EXTRA_CA_CERTS` - https://nodejs.org/api/cli.html#node_extra_ca_certsfile

Consumers will need to have these options set for node applications and/or NPM to work in general.

Of note is that we will honor the spirit of the different config options. The npm options will be treated as replacements for CAs and the node option will act as an addition to the root certificates. In both cases we will add the CAs to the CAs provided in the consumers cypress config.

### Steps to test

1. Set up a proxy using something like Charles proxy
2. Set up SSL proxying to include the dashboard (e.g. `api.cypress.io` or `api-staging.cypress.io`)
3. Save the root certificate for the proxy 
4. Run `NODE_EXTRA_CA_CERTS=<path to root certificate> NO_PROXY='<-loopback>' HTTP_PROXY=<proxy url> yarn cypress:open` (add in `CYPRESS_INTERNAL_ENV=staging` if you are hitting the staging dashboard)
5. Try to log in to the dashboard

n/a

### PR Tasks

- [X] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4826


mschile (approved):

packages/network/lib/agent.ts
```diff
@@ -293,7 +371,9 @@ class HttpsAgent extends https.Agent {
       options.port = 443
     }
 
-    super.addRequest(req, options)
+    mergeCAOption(options).then((options) => {
+      super.addRequest(req, options)
+    })
```
flotwig: Originally, we decided to make the agent use callback patterns throughout because callbacks are synchronous, whereas `.then` is asynchronous. So effectively this change adds 1 tick to all outbound HTTPS requests, which can result in a cost of >1ms per request before it's sent outbound, depending on how many tasks are queued for the current or next tick.

What do you think about changing this to still issue requests synchronously where possible? Something like this:

```ts
/** top of file, or on the first HTTPS request **/
let caOptions
const caOptionsPromise = getCaOptions().then(options => {
  // with no extra CA, caOptions is { ca: tls.rootCertificates } or {}
  caOptions = options
  return caOptions
})

/** in addRequest **/
if (caOptions) {
  return super.addRequest(req, merge(options, caOptions))
}

caOptionsPromise.then(caOptions => {
  super.addRequest(req, merge(options, caOptions))
})
```

packages/network/lib/agent.ts
```diff
@@ -8,11 +8,89 @@ import url from 'url'
...
 import { lenientOptions } from './http-utils'
 import { ClientCertificateStore } from './client-certificates'
+import { promises as fs } from 'fs'
+import tls from 'tls'
 
 const debug = debugModule('cypress:network:agent')
 const CRLF = '\r\n'
 const statusCodeRe = /^HTTP\/1.[01] (\d*)/
 
+// Load and cache the contents of process.env.npm_config_cafile so that we don't have to read it
```
flotwig: These new functions feel self-contained enough to move to a `ca.ts` file instead of being at the top of `agent.ts`.

packages/network/test/unit/agent_spec.ts
```diff
@@ -491,6 +513,136 @@ describe('lib/agent', function () {
...
+      {
+        name: 'should use the NODE_EXTRA_CA_CERTS override',
+        NODE_EXTRA_CA_CERTS: path.join(tempDirPath, 'extra_ca_certs.pem'),
+        caContents: [extraCACertsFileContents, ...tls.rootCertificates],
+        option: 'NODE_EXTRA_CA_CERTS',
+      },
+    ].slice().map((testCase) => {
+      context(testCase.name, function () {
+        beforeEach(function () {
+          // PROXY vars should override npm_config vars, so set them to cause failures if they are used
```
flotwig, line 532: Feels like these tests should set NODE_TLS_REJECT_UNAUTHORIZED = '1' in beforeEach, since there should be no TLS errors.

packages/network/lib/agent.ts
flotwig: It feels sorta like a micro-optimization at first, because so many Promises are chained to get to this point, but I think it's worth doing because we're just using the Promise to cache the resolved value here. That can be done synchronously and HTTPS requests can get sent out slightly faster.

ryanthemanuel: I like this idea. Implemented it and it's good to go on the PR now.

packages/network/lib/agent.ts
ryanthemanuel: Good call. I moved them.

packages/network/test/unit/agent_spec.ts
ryanthemanuel: Great idea. Took a little bit of time to get everything talking to each other properly, but this is now working.

packages/network/test/unit/agent_spec.ts
```diff
@@ -491,6 +500,143 @@ describe('lib/agent', function () {
...
+          delete process.env.npm_config_cafile
+          delete process.env.npm_config_ca
+          delete process.env.NODE_EXTRA_CA_CERTS
+          this.debugProxy.stop()
+        })
+
+        it(`CA from ${testCase.option} presented for https request`, function () {
+          return this.request({
+            url: `https://localhost:${HTTPS_PORT}/get`,
+            rejectUnauthorized: true,
```
ryanthemanuel, line 595: Setting `rejectUnauthorized` here rather than using the environment variable

flotwig (approved):
nice!

mschile (approved):

Issue #23768, opened 2022-09-09
[ACI post-project connection


marktnoonan:

baus:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/23768?planning-poker) with Zenhub @amehta265 @astone123 @lmiller1990 @marktnoonan @mike-plummer @rockindahizzy @warrensplayer @ZachJW34

PR #24541, opened 2022-11-04
feat: in-app information about new major versions

### LINKED ISSUE #24380: In-app information about new releases
This will take the following form - a full splash page in the launchpad - with exact content to be confirmed as part of this PR review.

Screen Shot 2022-10-31 at 8 49 27 AM

Other requirements:

- This will be a “single point of entry” in the launchpad
- UI is the same for all users/projects, no mention of migration, if a migration is needed, that is discovered after "Continue"
- After clicking the “Continue” button the next step the user moves to is:
  - Migration Helper, if needed, or
  - Whatever the next step would have been for that user based on what browser and testingType flags they passed to cypress open
- Clicking “Continue” records that the promo for this major version has been seen
- We record this status as a global preference on the users machine, each user should dismiss this UI only once per major version & not see it again when updating other projects etc.

Exact content is TBD - may include an illustration, may be just text.
For v11, we will always show the summary of the v10 version at the bottom, even if you came from v10 - we will not attempt to separate out those who are “leapfrogging” at the moment.

### LINKED ISSUE #24381: Updated Migration UI for Cy 11+
### What would you like?

We need to split out version specific content like "Welcome to Cypress 10!" from the information that users need to perform a migration before they they can launch the latest Cypress.

Designs are WIP but this would be a straight replacement of this screen in v10 with something more generic explaining the need to migrate:

Screen Shot 2022-10-25 at 10 26 19 AM

### Why is this needed?

The reason for this is that migration of `cypress.json` -> `cypress.config.ts` is required from any version below 10 to 10 or above. Other automated migrations might also be possible in future.

The user needs to migrate not because "they are upgrading to X" but rather "their project is in a format that Cypress no longer recognizes".

### Other

_No response_

marktnoonan:
- Closes #24380
- Closes #24381

### User facing changelog

Users will see a one-time Welcome screen the first time they run `cypress open` on a new major version of Cypress.

### Additional details

The most important change is that in Cypress 10 the Welcome screen was only shown when the project needed to go through config migration, in Cypress 11 the Welcome screen will be shown to _all_ users that have not already seen it for that major version.

Clicking the "Continue" button on the Welcome screen will record that the content for that particular Major has been dismissed and the user will not see it again. This is stored as a global user preference, not at the project level.

This Welcome screen appears as the first screen in the launchpad and is blocking until it is dismissed with the "Continue" button. Regardless of whether the user is opening in global mode, project mode, or with `--testingType` and `--browser` CLI flags, this will be the first thing they see.

Clicking "Continue" always leads to the appropriate next step in the launchpad based on how Cypress was opened. These steps can be:

- migration helper 
- global page
- testing type picker 
- browser picker with browser not launched (if only `--testingType` flag was passed)
- browser picker with browser opening (if in project mode and `--testingType` and `--browser` flags were both passed)

For context when reviewing this PR, there are a few decisions to call out about how this system will work (at least, for now):

#### `LocalSettingsActions`
I've updated how `LocalSettingsActions.setPreferences` works to perform a merge with existing settings in stead of fully overwriting the properties. We took a crack at this in #23762 but didn't implement correctly. I've removed the unit test that was added in that PR since, though it passed, it was testing the wrong thing. This merge approach allows setting state in nested objects, which is important as we now have 3 kinds of objects in state.json files - `banners`, `promptsShown`, and `majorVersionWelcomeDismissed`. `promptsShown` has a custom mutation and action just for merging this property, `banners` is merged with existing state on the clientside before being set. I haven't cleaned those up here, but will do so in #23768. We can also probably simplify the overall patterns (deeper in the server) around getting/setting saved state from the file system, as discussed with @warrensplayer, as it can be pretty difficult to debug an issue in this area.

#### How we are managing the new content
This PR introduces some article-like content that's different from most of the existing content in `en-US.json`. `MajorVersionWelcome.vue` is storing that content as plain HTML instead of the i18n system so that we can be flexible with the needs of each "post". Treating this content as "rich text" in this way makes it easier to understand what is being rendered. New static text like button labels and titles have still been added to i18n.

The Welcome content needs to be updated before each major release to highlight new features and breaking changes, so there will always be some custom markup here that will change major-to-major. Most of this content isn't asserted in tests - the interactive elements are, and we have Percy snapshots of the article as a whole. 

I've added a constant `MAJOR_VERSION_FOR_CONTENT` that defines the major version number that needs to be looked up in state to see if the welcome page was dismissed. We check this a few places, including on the server before launching a browser. This constant should only change in a release branch for new major versions & allows us to set up the correct content and behavior before the version actually exists.

The other consideration here is for this content to not rely on and API & work correctly even offline, so everything user-facing is hard coded, even things like release dates that could come from `VersionsDataSource` - we know what they are, so we can put them in the code.

#### Launchpad e2e tests
Since this is a new starting point for the launchpad, all launchpad e2e tests need to deal with it in some way. Following our existing patterns I added a `cy.skipWelcome()` custom command which is called in many places after `cy.visitLaunchpad()`. It can't be fully bound up with `cy.visitLaunchpad()` since there are cases where it will not be there to be dismissed. Some tests visit the launchpad more than once & this is a good thing - on the second visit it should be reading the value stored in global preferences and not show the Welcome page. In some cases I stubbed `localSettingsApi.getPreferences` to return the state with the current major version already dismissed - which seems good but is a lot more verbose. While this is in review I'll take a look at ways to simplify this - maybe add a parameter to `visitLaunchpad`, but the main thing is the tests are functioning.

#### Old video code
Lots of code that only existed to support the v10 launch video was removed

### Steps to test

That's a lot of detail but the test process is pretty quick.

**Basic Test**
- open `projects/__global__/state.json` in your favorite IDE, it should not have a `majorVersionWelcomeDismissed` property
- run `yarn dev` - Cypress should open at the launchpad and you should see the "What's New in Cypress" content
- Click the continue button - you should see the "Choose Testing Type" screen
- notice that an object that looks like this has appeared in your `state.json`: 
```json
"majorVersionWelcomeDismissed": {
    "11": 1667570082797
  }
```
- refresh the launchpad - it should not show the welcome screen

**Detailed test of state.json expectations**
- delete `majorVersionWelcomeDismissed` and refresh the launchpad - it should show the welcome screen again
- click continue as before
- `majorVersionWelcomeDismissed` should be back in the `state.json` now
- add some sibling properties and remove `"11"` from the object.
- refresh the launchpad, it should show the welcome page again
- click continue
- a new `"11"` key should be added in `majorVersionWelcomeDismissed` without blowing away the sibling properties.
- kill and restart Cypress
- you should not see the welcome screen

I suggest doing the basic+detailed test once, and just the basic test for the following situations to verify the user reaches the expected next step:

1. opening with a project and no flags
2. opening with a project and `--testingType` flag
3. opening with a project and `--testingType` and `--browser` flags
4. opening with a project that has a `cypress.json` instead of `cypress.config.ts`
5. opening in global mode

Also tab to the continue button and makes sure you can use it with space and enter keys - keyboard access to dismiss a fully-blocking state is essential.

- 🎥 [Loom video demo here](https://www.loom.com/share/7d121b984c7c40dea8e264a2a60c615f)

New Welcome page looks like this: 
Screen Shot 2022-11-04 at 10 43 10 AM
Screen Shot 2022-11-04 at 10 42 40 AM

Old welcome page:
Screen Shot 2022-11-04 at 10 45 07 AM

The only other change is some new text on the Migration Helper itself at the top:

New: 
Screen Shot 2022-11-04 at 10 48 19 AM

Old:
Screen Shot 2022-11-04 at 10 48 12 AM

### PR Tasks


packages/launchpad/src/migration/MajorVersionWelcome.vue
```diff
@@ -0,0 +1,168 @@
...
+                Released {{ versionReleaseDates['11'] }}
+              </span>
+            </div>
+            <div class="children:mb-16px">
+              <h2 class="font-bold text-18px text-jade-1000">
+                Component Testing Released
+              </h2>
+
+              <p>
+                Component Testing is now generally available for projects using React, Vue, Angular, Next.js, and Nuxt!
```
rockindahizzy: Nuxt won't be GA based on recent discussions

packages/data-context/src/data/ProjectLifecycleManager.ts
```diff
@@ -283,8 +284,12 @@ export class ProjectLifecycleManager {
     if (this.ctx.coreData.cliBrowser) {
       await this.setActiveBrowserByNameOrPath(this.ctx.coreData.cliBrowser)
 
+      const preferences = await this.ctx._apis.localSettingsApi.getPreferences()
+
+      const hasLandingPageBeenDismissed = Boolean(preferences.majorVersionWelcomeDismissed?.[MAJOR_VERSION_FOR_CONTENT])
+
       // only continue if the browser was successfully set - we must have an activeBrowser once this function resolves
```
rockindahizzy, line 291: Can this comment be updated to reflect the new condition as well?

packages/launchpad/src/migration/MajorVersionWelcome.vue
marktnoonan: Oh gotcha, updated

packages/launchpad/src/migration/MajorVersionWelcome.vue
```diff
@@ -0,0 +1,168 @@
...
+                Released {{ versionReleaseDates['11'] }}
+              </span>
+            </div>
+            <div class="children:mb-16px">
+              <h2 class="font-bold text-18px text-jade-1000">
+                Component Testing Released
+              </h2>
+
+              <p>
+                Component Testing is now generally available for projects using React, Vue, Angular, and Next.js!
```
rockindahizzy: This might just be me, but it makes makes my brain feel better to group React and Next.js together since they are related.

```suggestion
                Component Testing is now generally available for projects using React, Next.js, Angular, and Vue!
```

packages/types/src/constants.ts
```diff
@@ -20,3 +20,7 @@ export const MIGRATION_STEPS = ['renameAuto', 'renameManual', 'renameSupport', '
 export type MigrationStep = typeof MIGRATION_STEPS[number]
 
 export const PACKAGE_MANAGERS = ['npm', 'yarn', 'pnpm'] as const
+
+// Note: ONLY change this in code that will be merged into a release branch
+// for a new major version of Cypress
+export const MAJOR_VERSION_FOR_CONTENT = '11'
```
rockindahizzy, line 26: Does it make sense to add a reference to the locales file where there are also hard coded version numbers?

rockindahizzy (commented):
Awesome work! Some super minor things.

packages/data-context/src/actions/LocalSettingsActions.ts
```diff
@@ -14,19 +14,25 @@ export interface LocalSettingsApiShape {
 export class LocalSettingsActions {
   constructor (private ctx: DataContext) {}
 
-  setPreferences (stringifiedJson: string, type: 'global' | 'project') {
+  async setPreferences (stringifiedJson: string, type: 'global' | 'project') {
     const toJson = JSON.parse(stringifiedJson) as AllowedState
 
-    // update local data
-
-    _.merge(this.ctx.coreData.localSettings.preferences, toJson)
```
marktnoonan, line 22: This line was (harmlessly) in the wrong place because `localSettings.preferences` is exclusively dealing with global state, it was missed in an earlier refactor. Because we were accidentally setting everything here, we were able to write a straightforward unit test to check this `localSettings.preferences` in memory, but it was providing false confidence since this is not the same as what gets written to the files.

I do think we should have a unit test here, but it will take a bit of designing to avoid replacing so much of the function that we are only testing our test code.

packages/data-context/src/actions/LocalSettingsActions.ts
```diff
@@ -14,19 +14,25 @@ export interface LocalSettingsApiShape {
...
     if (type === 'global') {
+      // update local data on server
```
warrensplayer, line 21: What about a comment that says that this also happens with the `refreshLocalSettings` function below?

packages/types/src/constants.ts
marktnoonan: Good catch, that's not used any more - it's hanging around from before I moved it to this file. Removed.

packages/launchpad/src/migration/MajorVersionWelcome.vue
marktnoonan: Updated 👍

packages/frontend-shared/cypress/support/e2e.ts
```diff
@@ -356,6 +360,14 @@ function visitLaunchpad () {
   })
 }
 
+function skipWelcome () {
+  // click Continue on Welcome page
+  // and wait for the button to disappear
+  return cy.contains('button', 'Continue')
+  .click()
+  .should('not.exist')
+}
```
warrensplayer: This custom function has no ties to the Welcome screen, it just clicks a "Continue" button.  How about adding a `cy-test` attribute on the welcome screen and looking for the Continue button inside there?

packages/launchpad/cypress/e2e/open-mode.cy.ts
```diff
@@ -102,21 +116,50 @@ describe('Launchpad: Open Mode', () => {
...
       cy.get('button[data-cy=launch-button]').invoke('text').should('include', 'Start E2E Testing in Firefox')
     })
 
-    it('auto-launches the browser when launched with --browser --testingType --project', () => {
+    it('auto-launches the browser when launched with --browser --testingType --project, after Major Version Welcome is dismissed', () => {
       cy.scaffoldProject('launchpad')
       cy.openProject('launchpad', ['--browser', 'firefox', '--e2e'])
       cy.withCtx((ctx, o) => {
         o.sinon.stub(ctx._apis.projectApi, 'launchProject').resolves()
+        // o.sinon.stub(ctx._apis.localSettingsApi, 'getPreferences').callsFake()
```
warrensplayer: Can you remove this commented line?

warrensplayer (commented):
Going to pull the branch and run through the testing steps, but just wanted to send a couple of nit-picks your way.

packages/data-context/src/actions/LocalSettingsActions.ts
marktnoonan: Actually not too sure now that I look at it, what `refreshLocalSettings` does is slightly different since there's no "new" state that we want to put in that cache, it looks like we only use this function to populate it with whatever is on disk when Cypress is opened. A comment here might be confusing (though I'm open if you have a specific comment you think should be added).

packages/launchpad/cypress/e2e/open-mode.cy.ts
marktnoonan: Done 👍

packages/frontend-shared/cypress/support/e2e.ts
marktnoonan: Good call, added a `[data-cy="major-version-welcome-footer"]` to the parent element. Just in case we have other continue buttons in the content sometime.

rockindahizzy (approved):
LGTM!

warrensplayer (approved):
Everything tested out for me just great.

ZachJW34 (approved):

astone123 (approved):
Worked for me 👍🏻 nice 🎉

Issue #19908, opened 2022-01-26
Native support for Apple Silicon (M1 / arm64 architecture)


thomvaill:
### What would you like?

The native support of Apple Silicon by distributing an arm64 macos binary, without having to use Rosetta 2 at all.

**Solution 1:** by distributing two distinct binaries according to the macos architecture (x64 or arm64)
**Solution 2:** by distributing an [Universal macOS Binary](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary)

This is not a duplicate of https://github.com/cypress-io/cypress/issues/4478 because I am focusing only on Apple Silicon binaries, not ARM Linux binaries in general.
This is however a duplicate of https://github.com/cypress-io/cypress/issues/18724 which was closed, but since my request and my proposed solutions are a bit different, I prefer to open a new issue.

### Why is this needed?

@baversjo wrote a very good summary in  https://github.com/cypress-io/cypress/issues/4478:

> Running cypress on an ARM-based M1 mac is pretty slow under rosetta 2. Considering apple no longer seem to selling new intel macs as of q4 2021, I think it's a pretty important feature to add?

On top of that, I think the workaround explained here <https://cypress.io/blog/2021/01/20/running-cypress-on-the-apple-m1-silicon-arm-architecture-using-rosetta-2/> provides a very poor developer experience, having to maintain two different Nodejs installs with Rosetta.

### Proposed solutions

My proposed solution relies on the ability of electron-packager, and thus electron-builder to be able to build macos arm64 binaries on an x64 host, therefore not having to deploy custom CircleCI arm runners, as explained here: https://github.com/electron-userland/electron-builder/issues/5689

#### Solution 1: two distinct binaries: x64 and arm64

`electron-builder.json` becomes:

```json
{
  [...]
  "mac": {
    "target": {
      "target": "zip",
      "arch": [
        "x64",
        "arm64"
      ]
    },
    [...]
  },
  [...]
}
```

I tested: instead of creating one zip file (`build/build/Cypress-9.4.0-mac.zip`), it creates two (`build/build/Cypress-9.4.0-mac.zip` and `build/build/Cypress-9.4.0-arm64-mac.zip`). And as you can see, the binaries are built against the correct architecture:
    
```bash
# Executed on an Intel macbook
➜  cypress git:(develop) ✗ yarn binary-build --platform darwin --version $(node ./scripts/get-next-version.js)
# [...]

[...]

PR #24557, opened 2022-11-06
fix: for ELECTRON_RUN_AS_NODE=1 not working in Cypress 10

### LINKED ISSUE #23636: Unable to start Cypress with ELECTRON_RUN_AS_NODE=1
### Current behavior

If you set env with `ELECTRON_RUN_AS_NODE=1`, with MacOS, you are not able to start the Cypress with
```bash
ELECTRON_RUN_AS_NODE=1 npx cypress run
```
You'll get error:
`Error: Cannot find module '/Users/<username>/Library/Caches/Cypress/10.6.0/Cypress.app/Contents/MacOS/Contents/Resources/app/index.js'`

Checking the code, it seems that here is where it went wrong:
https://github.com/cypress-io/cypress/blob/b615e8ac30b681b4dc8485afe9814ff453da1fbf/cli/lib/exec/spawn.js#L175

The correct path, at least for Mac, should be:
```javascript
path.join(state.getBinaryPkgPath(path.join(executable, '..', '..', '..')), '..', 'index.js');
```

### Desired behavior

Cypress should start correctly

### Test code to reproduce

Try run with
```bash
ELECTRON_RUN_AS_NODE=1 npx cypress run
```

### Cypress Version

10.6.0

### Node version

16.0

### Operating System

macOS

### Debug Logs

_No response_

### Other

Cypress 10.7.0 (latest) also has this problem

XiaolongLyft:
- Closes #23636

### User facing changelog

Fix for user not able to use `ELECTRON_RUN_AS_NODE=1` to [run Cypress without xvfb](https://docs.cypress.io/guides/continuous-integration/introduction#Xvfb).
During the fix, also found a small bug about incorrect version check for `electron` browser.

### Additional details

Need to include package `electron` in one of packages' dependencies. Otherwise, the package will not be included in the binary's packages.
I included it in `server/package.json`, but probably can move `electron` in `devDependencies` to `dependencies`? @lmiller1990  https://github.com/cypress-io/cypress/blob/d52e610c0036a6c37196883b1c270a2e93ec91ff/package.json#L147

### Steps to test

Now running `ELECTRON_RUN_AS_NODE=1 npx cypress run --browser chrome --headless` will work.
But running `ELECTRON_RUN_AS_NODE=1 npx cypress run` with default `electron` browser still won't work with error:
```
electron_1.BrowserWindow is not a constructor
TypeError: electron_1.BrowserWindow is not a constructor
```
But I'm not sure if this is something this PR should fix.

N/A

### PR Tasks

### DIFF
--- a/cli/lib/exec/spawn.js
+++ b/cli/lib/exec/spawn.js
@@ -172,6 +172,12 @@ module.exports = {
 
         if (stdioOptions.env.ELECTRON_RUN_AS_NODE) {
           // Since we are running electron as node, we need to add an entry point file.
+          if (isPlatform('darwin')) {
+            startScriptPath = path.join(state.getBinaryPkgPath(path.join(executable, '..', '..', '..')), '..', 'index.js')
+          } else {
+            startScriptPath = path.join(state.getBinaryPkgPath(path.dirname(executable)), '..', 'index.js')
+          }
+
           startScriptPath = path.join(state.getBinaryPkgPath(path.dirname(executable)), '..', 'index.js')
         } else {
           // Start arguments with "--" so Electron knows these are OUR
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ mainBuildFilters: &mainBuildFilters
     branches:
       only:
         - develop
-        - 'ryanm/fix/improve-binary-cleanup'
+        - 'issue-23636'
 
 # usually we don't build Mac app - it takes a long time
 # but sometimes we want to really confirm we are doing the right thing
@@ -36,7 +36,7 @@ macWorkflowFilters: &darwin-workflow-filters
   when:
     or:
     - equal: [ develop, << pipeline.git.branch >> ]
-    - equal: [ 'ryanm/fix/improve-binary-cleanup', << pipeline.git.branch >> ]
+    - equal: [ 'issue-23636', << pipeline.git.branch >> ]
     - matches:
           pattern: "-release$"
           value: << pipeline.git.branch >>
@@ -45,7 +45,7 @@ linuxArm64WorkflowFilters: &linux-arm64-workflow-filters
   when:
     or:
     - equal: [ develop, << pipeline.git.branch >> ]
-    - equal: [ 'ryanm/fix/improve-binary-cleanup', << pipeline.git.branch >> ]
+    - equal: [ 'ryanm/fix/improve-binary-cleanuissue-23636p', << pipeline.git.branch >> ]
     - matches:
           pattern: "-release$"
           value: << pipeline.git.branch >>
@@ -63,7 +63,7 @@ windowsWorkflowFilters: &windows-workflow-filters
   when:
     or:
     - equal: [ develop, << pipeline.git.branch >> ]
-    - equal: [ 'ryanm/fix/improve-binary-cleanup', << pipeline.git.branch >> ]
+    - equal: [ 'issue-23636', << pipeline.git.branch >> ]
     - matches:
           pattern: "-release$"
           value: << pipeline.git.branch >>
@@ -130,7 +130,7 @@ commands:
       - run:
           name: Check current branch to persist artifacts
           command: |
-            if [[ "$CIRCLE_BRANCH" != "develop" && "$CIRCLE_BRANCH" != "ryanm/fix/improve-binary-cleanup" ]]; then
+            if [[ "$CIRCLE_BRANCH" != "develop" && "$CIRCLE_BRANCH" != "issue-23636" ]]; then
               echo "Not uploading artifacts or posting install comment for this branch."
               circleci-agent step halt
             fi


cli/lib/exec/spawn.js
```diff
@@ -172,7 +172,11 @@ module.exports = {
 
         if (stdioOptions.env.ELECTRON_RUN_AS_NODE) {
           // Since we are running electron as node, we need to add an entry point file.
-          startScriptPath = path.join(state.getBinaryPkgPath(path.dirname(executable)), '..', 'index.js')
+          if (isPlatform('darwin')) {
```
lmiller1990, line 175: Just to clarify if this fixing the issue on darwin only?

Also how did you test this - did you build the project for darwin? Or did you just pick the correct path (no actual testing)?

I've always had problems doing this locally, but I can turn on darwin builds and generate a binary via our CI for testing.

packages/server/lib/browsers/utils.ts
```diff
@@ -249,7 +249,7 @@ const getBrowsers = async () => {
     return browsers
   }
 
-  const version = process.versions.chrome || ''
+  const version = process.versions.electron || ''
```
lmiller1990: Hmm this seems like a pretty obvious bug, surprised it flew under the radar for so long. Will need to check this, good job calling this out

packages/server/package.json
```diff
@@ -56,6 +56,7 @@
     "dayjs": "^1.9.3",
     "debug": "^4.3.2",
     "duplexify": "4.1.1",
+    "electron": "21.0.0",
```
lmiller1990: I think this is intentionally a devDep - we bundle electron in later during the build step, and it's declared in the top level package.json. I *think* adding this here as a regular dep will bloat the binary with an additional electron (need to check this).

Why do we need to add it here?

lmiller1990:
Nice job, main question is around https://github.com/cypress-io/cypress/pull/24557#discussion_r1014991317

I tried running this is my local Cypress dev env...

```
yarn run v1.22.19
$ cross-env TZ=America/New_York node ../../scripts/cypress run --e2e --project . --browser chrome --headless
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module '/Users/lachlan/code/work/cypress/Contents/Resources/app/index.js'
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)
    at node:internal/main/run_main_module:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
$ ls /Users/lachlan/code/work/cypress/Contents/Resour
$ ELECTRON_RUN_AS_NODE=1 yarn cypress:run:e2e --browser chrome --headless
```

How did you test this? Darwin only bugfix? Please bit a bit more specific about the bug/environment, and I can give it a test.

If you need a fully built binary for a specific environment(s) please post here, I can update our CI config and it'll built an installed binary for whatever OS you need with all the dependencies installed. You could then add that to your package.json and we can test the patch against some real projects before we ship it to verify it's working as expected. Thanks!!

cli/lib/exec/spawn.js
XiaolongLyft: >Just to clarify if this fixing the issue on darwin only?

I only tested in darwin - it would be nice if we can have a full covered test via CI with Linux and Windows covered.

How I did the test (in MacOS M1):
Basically, I followed the step here: https://github.com/cypress-io/cypress/issues/19908#issuecomment-1131761214
to build a local binary (using `yarn binary-build --platform darwin`) and moved it into my Libaray/Caches, and tested it in one of my real world projects.

packages/server/package.json
XiaolongLyft: The reason why we're seeing error
```
Cannot find module 'electron'
```
is that in the binary file, in MacOS for example, in `/Library/Caches/Cypress/10.11.0/Cypress.app/Contents/Resources/app/node_modules`, there does not exist the needed `electron` package (you can check in existing released apps), so either we somehow fix it later in the build step to include that package, or we define it in deps.
I do see the CI checks failed because of a larger zip size, so definitely we can look into this more. But either way, the binary size would probably increase.

XiaolongLyft:
Thanks for looking into this really quick @lmiller1990 ! Appreciate that!
>How did you test this? 

I want to mention that this fix is actually in 2 different places, `spawn.js` is in `cli`, and other fixes are suppose to affect the Cypress binary app.
So to effectively test this, I locally built cli and the binary. I replaced the cli section in one of my real projects' in `node_modules`, and I also replaced the binary in `~/Library/Caches/Cypress`. With those 2 parts updated, I can successfully run `ELECTRON_RUN_AS_NODE=1 npx cypress run --browser chrome --headless`.

However, I should point out:
Running `ELECTRON_RUN_AS_NODE=1 npx cypress run` with default `electron` browser still won't work with error:
```
electron_1.BrowserWindow is not a constructor
TypeError: electron_1.BrowserWindow is not a constructor
```
And I think that is not something related to the original issue.

packages/server/lib/browsers/utils.ts
XiaolongLyft: The fix would not work without this. Otherwise, Cypress would not spin up electron browser with `ELECTRON_RUN_AS_NODE=1` because `version` would be an empty string.

cli/lib/exec/spawn.js
lmiller1990: Right I see, good idea, that's an easy way to test it quickly. You wouldn't even need to build it, you could just modify an existing binary.

Anyway, I updated our config to generate a full, installable binary for MacOS: https://github.com/cypress-io/cypress/compare/lmiller-xiaolong/issue-23636?expand=1

It takes a while (1 hour, CI + build). Once it's completed, a bot will post on that link with the install command, and we can give it a test. Please wait a bit, thanks! I'd like to get this tested and merged by the end of the week.

packages/server/lib/browsers/utils.ts
emilyrohrbough: I had asked @flotwig about this before. He said had told me:

> The Chrome key there is the version of chromium bundled into Electron, we use that as the display version because it's more useful to the user since it correlates to Chrome versions. Whereas "Electron 19" to a user is less useful.

cli/lib/exec/spawn.js
lmiller1990: Hmm it seems the binary wasn't generated, I might have done something incorrectly. I can take a look next week (Mon 13th Nov).

packages/server/lib/browsers/utils.ts
lmiller1990: Oh, right, neat trick. We should add a comment.

lmiller1990:
Hey @XiaolongLyft, I am trying to get this to build on MacOS so we can test it out, hopefully this commit will build the binary. https://github.com/cypress-io/cypress/commit/763d1548cd518cbed962c6ad6fd7496e8c030e3f. Doing it on a separate branch since I don't think an external contribute can build the MacOS binaries.

Just letting you know I didn't forget about this, aiming to get it built+tested in the next 24 hours.

lmiller1990:
Ok @XiaolongLyft can you give these a test locally (and CI, if you want) and make sure it's behaving as you expected? *Edit, I think this build is not working, we need to rebase of develop... see below*.

<details><summary>Builds</summary>

[Circle has built](https://circleci.com/gh/cypress-io/cypress/1922776) the darwin arm64 version of the Test Runner.

Learn more about this pre-release platform-specific build at https://on.cypress.io/installing-cypress#Install-pre-release-version.

Run this command to install the pre-release locally:
```
npm install https://cdn.cypress.io/beta/npm/11.1.0/darwin-arm64/lmiller-xiaolong/issue-23636-763d1548cd518cbed962c6ad6fd7496e8c030e3f/cypress.tgz
```

.................

[Circle has built](https://circleci.com/gh/cypress-io/cypress/1922835) the darwin x64 version of the Test Runner.

Learn more about this pre-release platform-specific build at https://on.cypress.io/installing-cypress#Install-pre-release-version.

Run this command to install the pre-release locally:
```
npm install https://cdn.cypress.io/beta/npm/11.1.0/darwin-x64/lmiller-xiaolong/issue-23636-763d1548cd518cbed962c6ad6fd7496e8c030e3f/cypress.tgz
```
______

</details>

Hmm this isn't working for me, I got this in my existing project.

```
$ /Users/lachlan/code/rhythm/node_modules/.bin/cypress run --component
It looks like this is your first time using Cypress: 11.1.0


Cypress failed to start.

This may be due to a missing library or dependency. https://on.cypress.io/required-dependencies

Please refer to the error below for more details.

----------

/Users/lachlan/Library/Caches/Cypress/beta-11.1.0-lmiller-xiaolong/issue-23636-763d1548/Cypress.app/Contents/MacOS/Cypress: bad option: --no-sandbox
/Users/lachlan/Library/Caches/Cypress/beta-11.1.0-lmiller-xiaolong/issue-23636-763d1548/Cypress.app/Contents/MacOS/Cypress: bad option: --smoke-test
/Users/lachlan/Library/Caches/Cypress/beta-11.1.0-lmiller-xiaolong/issue-23636-763d1548/Cypress.app/Contents/MacOS/Cypress: bad option: --ping=998

----------

Platform: darwin-arm64 (21.3.0)
Cypress Version: 11.1.0
```

A new project is even more weird:

```
$ /Users/lachlan/code/dump/cy11/node_modules/.bin/cypress run --browser electron
/Users/lachlan/Library/Caches/Cypress/beta-11.1.0-lmiller-xiaolong/issue-23636-763d1548/Cypress.app/Contents/Resources/app/index.js:2
  throw new Error('global.snapshotResult is not defined. This binary has been built incorrectly.')
  ^

Error: global.snapshotResult is not defined. This binary has been built incorrectly.
    at Object.<anonymous> (/Users/lachlan/Library/Caches/Cypress/beta-11.1.0-lmiller-xiaolong/issue-23636-763d1548/Cypress.app/Contents/Resources/app/index.js:2:9)
    at Module._compile (node:internal/modules/cjs/loader:1120:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1175:10)
    at Module.load (node:internal/modules/cjs/loader:988:32)
    at Module._load (node:internal/modules/cjs/loader:829:12)
    at c._load (node:electron/js2c/asar_bundle:5:13339)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)
    at node:internal/main/run_main_module:17:47
error Command failed with exit code 1.
```

I think I might have broken something when I rebased Cypress 11 into your branch, let me try again.

___

I'm going to try and pick this up and get it over the line... I force pushed to your branch, rebased on the latest develop. https://github.com/cypress-io/cypress/pull/24557/commits/abb607d25f1af0a98795758a1b2692ace94a3c9e

packages/server/lib/browsers/utils.ts
lmiller1990: @XiaolongLyft can you please revert this, as pointed out it's intentional.

packages/server/package.json
lmiller1990: Yep, CI check: https://app.circleci.com/pipelines/github/cypress-io/cypress/45714/workflows/902afa25-43af-4183-bdf4-e3348df80565/jobs/1922692?invite=true#step-113-91

lmiller1990:
Hmm this isn't working for me, I got

```
$ /Users/lachlan/code/rhythm/node_modules/.bin/cypress run --component
It looks like this is your first time using Cypress: 11.1.0


Cypress failed to start.

This may be due to a missing library or dependency. https://on.cypress.io/required-dependencies

Please refer to the error below for more details.

----------

/Users/lachlan/Library/Caches/Cypress/beta-11.1.0-lmiller-xiaolong/issue-23636-763d1548/Cypress.app/Contents/MacOS/Cypress: bad option: --no-sandbox
/Users/lachlan/Library/Caches/Cypress/beta-11.1.0-lmiller-xiaolong/issue-23636-763d1548/Cypress.app/Contents/MacOS/Cypress: bad option: --smoke-test
/Users/lachlan/Library/Caches/Cypress/beta-11.1.0-lmiller-xiaolong/issue-23636-763d1548/Cypress.app/Contents/MacOS/Cypress: bad option: --ping=998

----------

Platform: darwin-arm64 (21.3.0)
Cypress Version: 11.1.0
```

packages/server/lib/browsers/utils.ts
lmiller1990: Edit: I reverted it for you.

packages/server/package.json
lmiller1990: I will revert this and see how we go.

lmiller1990:
Going to create a fresh PR and kick this off here https://github.com/cypress-io/cypress/pull/25207

PR #24558, opened 2022-11-07
chore: minor style updates & fix a windows test

marktnoonan:
This PR fixes a windows Launchpad test that was failing due to config being slow to load. It's only possible to test the "record first run" prompt after a testing type has been chosen, which means the test needs to wait for the "Choose a Browser" screen to appear.

Also tweaks the border of the Major Version Welcome page to match the color in figma, and removes an extra overflow utility class that was causing a double scrollbar on Windows.

### PR Tasks

### DIFF
--- a/packages/app/cypress/e2e/support/execute-spec.ts
+++ b/packages/app/cypress/e2e/support/execute-spec.ts
@@ -22,7 +22,7 @@ declare global {
   }
 }
 
-const waitForSpecToFinish = (expectedResults, timeout?: number) => {
+export const waitForSpecToFinish = (expectedResults, timeout?: number) => {
   // First ensure the test is loaded
   cy.get('.passed > .num').should('exist')
   cy.get('.failed > .num').should('exist')
@@ -39,8 +39,3 @@ const waitForSpecToFinish = (expectedResults, timeout?: number) => {
 }
 
 Cypress.Commands.add('waitForSpecToFinish', waitForSpecToFinish)
-
-// Here we export an empty object to tell the
-// typescript type checker that this definitely is a module.
-// This way, we are allowed to use the global namespace declaration
-export {}
--- a/packages/app/cypress/e2e/support/execute-spec.ts
+++ b/packages/app/cypress/e2e/support/execute-spec.ts
@@ -1,7 +1,6 @@
 import { shouldHaveTestResults } from '../runner/support/spec-loader'
 
-export interface ExpectedResults
-{
+export interface ExpectedResults {
   passCount?: number
   failCount?: number
   pendingCount?: number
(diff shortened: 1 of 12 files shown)


lmiller1990:
Double scroll bar is gone. It says "11.0.0 Released next month", but users will see it this month. Do we even need a byline? How about just "11.0.0 - latest release"?

image

lmiller1990 (commented):
Fix works, left comment about other wording.

warrensplayer:
> Double scroll bar is gone. It says "11.0.0 Released next month", but users will see it this month. Do we even need a byline? How about just "11.0.0 - latest release"?
> 
> image

I fixed the values being passed to Date.UTC that was a parameter to useTimeAgo. The `monthIndex` parameter is `0` based. The display will be a little strange during the day tomorrow because it will report that the release was done at midnight UTC, but then will be correct as time progresses.

After fix:
image

marktnoonan:
Good catch @warrensplayer! We need to update the test for "11.0.0 Released just now"

system-tests/projects/react17/package.json
```diff
@@ -7,7 +7,7 @@
     "@vitejs/plugin-react": "^1.3.2",
     "css-loader": "^4.3.0",
     "style-loader": "^2.0.0",
-    "vite": "2.9.14",
+    "vite": "3.2.2",
```
ZachJW34, line 10: Why was this updated? Should bump `@vitejs/plugin-react` as well

marktnoonan: Some context here in slack: https://cypressio.slack.com/archives/C011BAPC614/p1667810101355419?thread_ts=1667789779.018379&cid=C011BAPC614

packages/app/cypress/e2e/support/execute-spec.ts
```diff
@@ -23,13 +22,10 @@ declare global {
   }
 }
 
-// Here we export the function with no intention to import it
-// This only tells the typescript type checker that this definitely is a module
-// This way, we are allowed to use the global namespace declaration
-export const waitForSpecToFinish = (expectedResults, timeout?: number) => {
+const waitForSpecToFinish = (expectedResults, timeout?: number) => {
   // First ensure the test is loaded
-  cy.get('.passed > .num').should('contain', '--')
```
lmiller1990, line 31: Race condition - sometimes the rest actually executes fast enough, and it's not `--`, but the number of passes. So, better just verify the elements exist.

system-tests/projects/vueclivue2-configured/package.json
```diff
@@ -10,8 +10,6 @@
     "vue": "^2.6.11"
   },
   "devDependencies": {
-    "@cypress/vue2": "file:../../../npm/vue2",
```
lmiller1990, line 13: This doesn't do what you expect on windows apparently. Also, seems like it isn't strictly required.

lmiller1990 (approved):
I messed around a bit to get windows to a reasonable state. There is some flake I could not solve, but it's windows only and the behaviors are working fine and verified via linux tests; I skipped them and created issues in the interest of getting this merged.

- https://github.com/cypress-io/cypress/issues/24575
- https://github.com/cypress-io/cypress/issues/24579

Looks like there is still some flake, but far less than before.

I did `skip ci` for the last commit. I checked CI before merging; everything is ✅ .

PR #24559, opened 2022-11-07
chore: make error more inclusive of future cypress versions

lmiller1990:
I tried installing Cypress 11 in a Cypress 9 project. If you open it in global mode, it says "You are in global mode. Please install Cypress 10 in your project to migrate". 

I tested it out, I was able to migrate from Cypress 9 to Cypress 11 using the migration wizard.

We didn't really discuss if this should work (eg; should we force users to Cypress 10, then Cypress 11?) but it works great as is, I say we just make the minimal change, which is to update the language in the message.


packages/errors/src/errors.ts
```diff
@@ -1559,7 +1559,7 @@ export const AllCypressErrors = {
 
   MIGRATION_CYPRESS_NOT_FOUND: () => {
     return errTemplate`
-      You are running Cypress 10 in global mode and attempting to open or migrate a project where an install of ${fmt.code('cypress')} cannot be found.
+      You are running Cypress 10+ in global mode and attempting to open or migrate a project where an install of ${fmt.code('cypress')} cannot be found.
 
       Ensure that ${fmt.code('cypress@10')} is installed in the project you are attempting to open or migrate.
```
marktnoonan:
```suggestion
      Ensure that ${fmt.code('cypress@10')} or greater is installed in the project you are attempting to open or migrate.
```

marktnoonan (approved):
Good catch, definitely not intending that people upgrade to 10 separately - 9.x -> 11.0 is expected to be fine.

Suggested a change to the line below as well.

warrensplayer (approved):

astone123 (approved):

lmiller1990:
Finally, this is green. This was a bit painful... apparently missing a newline here: https://github.com/cypress-io/cypress/pull/24559/commits/fce1b511b397c9c7179b70a8abaea0bb6ad29cfe was causing CI to fail, without actually making it obvious *why*. Also, it wasn't failing locally.

PR #24565, opened 2022-11-07
fix: Only save session if validation succeeds

### LINKED ISSUE #24208: Only Save Session after Successful Validation
4. Storing the session should happen post validation

chrisbreiding:
- Closes #24208 

### User facing changelog

- Sessions will no longer be stored if validation fails

### PR Tasks


packages/app/cypress/e2e/runner/sessions.ui.cy.ts
```diff
@@ -132,6 +132,29 @@ describe('runner/cypress sessions.ui.spec', {
     // cy.percySnapshot() // TODO: restore when Percy CSS is fixed. See https://github.com/cypress-io/cypress/issues/23435
   })
 
+  // https://github.com/cypress-io/cypress/issues/24208
+  it('does not save session when validation fails', () => {
+    loadSpec({
+      projectName: 'session-and-origin-e2e-specs',
+      filePath: 'session/new_session_and_fails_validation_retries.cy.js',
```
emilyrohrbough, line 139: Can we add and/or update this test to only fail validation 1 time so we can verify the second retry is 'created' a session instead of 'recreated' a session?

packages/app/cypress/e2e/runner/support/spec-loader.ts
```diff
@@ -2,6 +2,7 @@ export const shouldHaveTestResults = ({ passCount, failCount, pendingCount }) =>
   passCount = passCount || '--'
   failCount = failCount || '--'
 
+  cy.get('button.restart').should('be.visible') // ensure tests are finished running
```
emilyrohrbough: Did you need to add this for test retries?

packages/driver/src/cy/commands/sessions/index.ts
```diff
@@ -358,11 +355,15 @@ export default function (Commands, Cypress, cy) {
...
         .then(() => validateSession(existingSession))
-        .then((isValidSession: boolean) => {
+        .then(async (isValidSession: boolean) => {
           if (!isValidSession) {
             return
           }
 
+          sessionsManager.setActiveSession({ [existingSession.id]: existingSession })
+          sessionsManager.registeredSessions.set(existingSession.id, true)
+          await sessions.saveSessionData(existingSession)
```
emilyrohrbough, line 364: It looks like `sessions.saveSessionData()` is also setting `setActiveSession()`, so might be redundant

packages/app/cypress/e2e/runner/sessions.ui.cy.ts
chrisbreiding: Added in https://github.com/cypress-io/cypress/pull/24565/commits/731df94ce6655576392eca54c1833d7e653de079

packages/driver/src/cy/commands/sessions/index.ts
chrisbreiding: Good call. Addressed in https://github.com/cypress-io/cypress/pull/24565/commits/ca450fe58a43c7b419c09de01842131b8cc4d7cb

packages/app/cypress/e2e/runner/support/spec-loader.ts
chrisbreiding: No, it's unrelated. I just noticed that if there's supposed to be more than one failure, but `failCount` is mistakenly set to less than the number of failures, it can result in a false positive because the fail count assertion passes as soon at there's _failCount_ failures, which is mid-test. This ensures it won't try to get the number failures and run the assertion until the tests are finished.

emilyrohrbough (approved):

AtofStryker (approved):

PR #24571, opened 2022-11-07
fix: make component derived info not throw

### LINKED ISSUE #24344: Gracefully handle failures when deriving component name during mount
## Current Behavior
When passing a component to mount in `React` as an object, the mount function cannot derive the component name and throws an error. The error that is surfaced to the user does not point to the cause or solution of the scenario and is confusing for the user. 

## Expected  Behavior
Instead of throwing the error at this stage we should verify that the component is the correct type and throw the error if that is not the case with better messaging for the user

## Additional considerations
Verify that other frameworks handle this situation gracefully and address it in a similar way if this is not the case.

ZachJW34:
- Closes #24344

### User facing changelog
Do not throw error/throw better error when providing improper but still valid arguments to `mount`

### Additional details
We derive information from the component/class passed into `cy.mount` such as the name of the component or whether it is a standalone component (Angular). Our mount would error in certain scenarios:

- React: calling `cy.mount(some_primitive_value)` e.g. `cy.mount('Hello World')`
- Angular: calling mount with an undecorated component e.g. `class MyClass {}; cy.mount(MyClass)`

The goal of this PR isn't to validate what is passed. If `cy.mount(null)` is called, stuff should break and that's expected but if the types allow for certain values like `cy.mount(5)` then it shouldn't throw an error. If it does throw an error, it should be the render throwing the error and not `Cannot read "displayName" of undefined` where we are trying to pull info from the component passed in.

### Steps to test
System-tests help here. If you want to test an external project, you'll first have to build the mount package and then symlink/copy it to your project's `node_modules/cypress/{framework}`

Before:

https://user-images.githubusercontent.com/25158820/200407940-e9588d1f-e240-449f-ae47-b0e2a85d7fc9.mov

After:

https://user-images.githubusercontent.com/25158820/200407996-65bcdb5d-b0d7-45b0-8f8a-6e77c2b5582c.mov

### PR Tasks

### DIFF
--- a/npm/angular/src/mount.ts
+++ b/npm/angular/src/mount.ts
@@ -156,7 +156,7 @@ function bootstrapModule<T> (
   })
 
   // check if the component is a standalone component
-  if ((component as any).ɵcmp.standalone) {
+  if ((component as any).ɵcmp?.standalone) {
     testModuleMetaData.imports.push(component)
   } else {
     testModuleMetaData.declarations.push(component)
--- a/npm/react/src/createMount.ts
+++ b/npm/react/src/createMount.ts
@@ -40,12 +40,6 @@ export const makeMountFn = (
 
   mountCleanup = internalMountOptions.cleanup
 
-  // Get the display name property via the component constructor
-  // @ts-ignore FIXME
-  const componentName = getDisplayName(jsx.type)
-
-  const jsxComponentName = `<${componentName} ... />`
-
   return cy
   .then(() => {
     const reactDomToUse = internalMountOptions.reactDom
@@ -99,6 +93,12 @@ export const makeMountFn = (
       .wait(0, { log: false })
       .then(() => {
         if (options.log !== false) {
+          // Get the display name property via the component constructor
+          // @ts-ignore FIXME
+          const componentName = getDisplayName(jsx)
+
+          const jsxComponentName = `<${componentName} ... />`
+
           Cypress.log({
             name: type,
             type: 'parent',
@@ -108,7 +108,7 @@ export const makeMountFn = (
             consoleProps: () => {
               return {
               // @ts-ignore protect the use of jsx functional components use ReactNode
-                props: jsx.props,
+                props: jsx?.props,
                 description: type === 'mount' ? 'Mounts React component' : 'Rerenders mounted React component',
                 home: 'https://github.com/cypress-io/cypress',
               }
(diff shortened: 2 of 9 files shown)


npm/angular/src/mount.ts
```diff
@@ -156,7 +156,7 @@ function bootstrapModule<T> (
   })
 
   // check if the component is a standalone component
-  if ((component as any).ɵcmp.standalone) {
+  if ((component as any).ɵcmp?.standalone) {
```
lmiller1990, line 159: Forgot Angular uses Greek chars, lol

npm/react/src/createMount.ts
```diff
@@ -108,7 +108,7 @@ export const makeMountFn = (
             consoleProps: () => {
               return {
               // @ts-ignore protect the use of jsx functional components use ReactNode
-                props: jsx.props,
+                props: jsx?.props,
```
lmiller1990, line 111: I think React defaults to an empty object for no props?

system-tests/project-fixtures/react/src/mount.cy.jsx
```diff
@@ -24,4 +24,11 @@ describe('mount', () => {
       cy.get('[data-cy-root]').children().should('have.length', 1)
     })
   });
+
+  it('does not error when rendering primitives', () => {
+    cy.mount('Hello World')
+    cy.mount(5)
+    cy.mount(null)
+    cy.mount(undefined)
```
lmiller1990, line 32: Tests look good, should we write these in npm/react where possible?

lmiller1990 (approved):
No hard blockers but some questions and ideas

npm/react/src/createMount.ts
ZachJW34: Maybe when creating the component but `jsx` here is just the first argument to `mount`, so passing `cy.mount(null)` would throw. I know it doesn't make sense to do that, but the types allow it and React doesn't throw so I thought I'd be safe here.

system-tests/project-fixtures/react/src/mount.cy.jsx
ZachJW34: Would like to but this way allows testing across react17 and react18 which `npm/react` doesn't allow for

npm/angular/src/mount.ts
ZachJW34: I'm pretty sure I had to copy and paste this line cause I don't know how to type `ɵ` on my keyboard 😅

rockindahizzy (approved):
😎

lmiller1990:
Looks good!

PR #24598, opened 2022-11-08
fix: revert dynamic import of webpack config file

ZachJW34:
### User facing changelog
n/a

### Additional details
This PR reverts the change made to `webpack-dev-server` for how we are loading the sourced `webpack.config.{js,ts}`. Previously it was using `loadModule` which is basically a require. We changed it to use a dynamic import but this no longer works when doing `await import('{project-root}/webpack.config.ts')` as ts-node does not process the typescript webpack config file.

No user-facing change as the dynamic import was never published.

### Steps to test
On develop, run `yarn cypress:open` and choose an app that has a `webpack.config.ts` and does not specify it in the `cypress.config.ts`. [svelte-webpack-ts](https://github.com/cypress-io/cypress-component-testing-apps/tree/main/svelte-webpack-ts) is a good example. This will fail.

Checkout this branch and run `yarn`. Run the same steps above and the app will load.

https://user-images.githubusercontent.com/25158820/200634243-bfc59a19-e0ec-426e-bc6b-e1c136fd6347.mov

### PR Tasks


rockindahizzy (approved):
Thanks for fixing this quickly!

astone123 (approved):
This fixed the issue for me 🎉

rockindahizzy (approved):
Worked

Issue #24600, opened 2022-11-08
Flaky test: driver navigation tests in webkit


chrisbreiding:
### Link to dashboard or CircleCI failure

https://app.circleci.com/pipelines/github/cypress-io/cypress/45549/workflows/960f7cf0-f6cf-4a2a-9624-191c43802021/jobs/1913966

### Link to failing test in GitHub

[This test](https://github.com/cypress-io/cypress/blob/18865644090458b850b0cab1fc3c93a86526fe89/packages/driver/cypress/e2e/commands/navigation.cy.js#L756) and others in the same file.

### Analysis

The tests are frequently flaky. They usually all pass or all fail together, so it's probably the same root cause.

### Cypress Version

10.10.0+

### Other

Search for this issue number in the codebase to find the test(s) skipped until this issue is fixed

PR #24601, opened 2022-11-08
chore: Skip flaky navigation tests in webkit

chrisbreiding:
### User facing changelog

N/A

### Additional details

Didn't dig into why these are flaky, but want to keep them from frequently failing for the time being.

Documented in https://github.com/cypress-io/cypress/issues/24600

### PR Tasks

- [N/A] Have tests been added/updated?


mjhenkes (approved):

flotwig (changes requested):
Instead of skipping, can you slap a `{ retries: 15 }` on these, or like `if (Cypress.isBrowser('webkit')) { this.retries(15) }` We don't want to lose this coverage.

flotwig (approved):

Issue #24467, opened 2022-10-31
[Run All Specs] - Spec Explorer - Add "run N specs" button to folder rows


marktnoonan:
### Summary
Currently, users can filter specs and choose a single one to run in open mode. They need the ability to run the currently-filtered list of specs, which replicates the functionality of Run All in 9.x. This implementation adds an extra ability not possible in 9.x: to further select only a specific folder, to run the subset of specs _currently matching the filter_ in that folder (child folders included). This is achieved with a "Run N specs" button present on each folder in the Spec Explorer, instead of a single "Run All"/"Run Filtered" button.

#### Empty state, no visual change
<img width="600" alt="empty state" src="https://user-images.githubusercontent.com/8340719/200563048-64c396b3-0141-42b2-925e-2cc6aca7181c.png">

#### Hover on row but not button
<img width="600" alt="hover on row" src="https://user-images.githubusercontent.com/8340719/200562962-e5bfeb0f-2e6f-479b-94c4-29f27b701f5b.png">

#### Hover on button itself
<img width="600" alt="hover on button" src="https://user-images.githubusercontent.com/8340719/200562661-8eb0ecd8-61c0-4db3-956b-44e08f7d8e38.png">

### Acceptance Criteria

**In e2e mode only**:

1. Buttons should exist on all folders
2. On top-level folder, it should be called "Run N specs" and run all specs that match the current search
3. On all other folders, it should be called "Run N specs" where N is the number of specs in the folder that match the current search and those should be the ones run
4. The button appears when the folder row is hovered
5. Buttons have their own hover state and are activated on click
6. The behavior of clicking the row is still to toggle the open/closed state of the folder
7. When clicked, the button takes the user to the Spec Runner page and runs the selected specs.
8. Button should be reachable by the keyboard and appear when focussed with the tab key

### Resources

Figma prototype: https://www.figma.com/proto/PAOozuCxtfetqP8fUVvg0y/branch/WpNw4mG3LvD1J3zLj80KCr/Cypress-App?node-id=20668%3A150076&starting-point-node-id=20668%3A150076

baus:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/24467?planning-poker) with Zenhub @amehta265 @astone123 @lmiller1990 @marktnoonan @mike-plummer @rockindahizzy @warrensplayer @ZachJW34

[...]

PR #24674, opened 2022-11-14
chore: move non-shared code from frontend-shared to app

lmiller1990:
While working on the Spec List code for https://github.com/cypress-io/cypress/issues/24467 with @amehta265, we found it confusing that a lot of functionality was split across `frontend-shared` and `app`. Turns out a lot of what was in `frontend-shared` isn't shared - it's exclusively consumed in `app`. For this reason, I decided to move the `app` only code into `app`. This makes it much easier to follow the code around.

There are no functionality changes here, I just deleted some unused code and moved some from around.


packages/app/src/specs/spec-utils.ts
```diff
@@ -2,127 +2,7 @@ import fuzzySort from 'fuzzysort'
...
-  id: string
-  name: string
-  children: SpecTreeNode<T>[]
-  isLeaf: boolean
-  parent?: SpecTreeNode<T>
-  data?: T
-  highlightIndexes: number[]
-}
-
-export function buildSpecTree<T extends FoundSpec> (specs: FoundSpec[], root: SpecTreeNode<T> = { name: '', isLeaf: false, children: [], id: '', highlightIndexes: [] }): SpecTreeNode<T> {
```
lmiller1990, line 37: `spec-utils` ended up a dumping ground for random things - many of these functions have one very specific purpose, which is to build the spec tree (not just utils) so I moved them to a more accurately named file.

packages/app/src/specs/tree/useCollapsibleTree.ts
```diff
@@ -0,0 +1,266 @@
+import type { ComputedRef, Ref } from 'vue'
```
lmiller1990, line 1: Moved from frontend-shared, no changes. This is not a shared file, it is only used in App.

packages/app/src/store/specs-store.ts
```diff
@@ -12,16 +12,12 @@ export const useSpecStore = defineStore({
   state (): SpecState {
     return {
       activeSpec: undefined,
-      specFilter: undefined,
```
lmiller1990, line 15: Unused? I think a hold-over from before we used GraphQL and preferences to save the spec filter.

packages/app/src/composables/useSpecFilter.ts
```diff
@@ -25,7 +25,6 @@ export function useSpecFilter (savedFilter?: string) {
 
   function setSpecFilter (specFilter: string) {
     if (specStore.specFilter !== specFilter) {
-      specStore.setSpecFilter(specFilter)
```
lmiller1990, line 28: Unused, this was not doing anything but saving the specFilter in the store for no reason.

packages/app/src/store/specs-store.ts
marktnoonan: Yeah we don't seem to be reading it anywhere 👍

packages/app/src/specs/SpecsList.vue
```diff
@@ -183,12 +183,11 @@ import { gql, useSubscription } from '@urql/vue'
...
 import { useI18n } from '@cy/i18n'
-import { buildSpecTree, fuzzySortSpecs, makeFuzzyFoundSpec, useCachedSpecs } from './spec-utils'
-import type { FuzzyFoundSpec } from './spec-utils'
-import { useCollapsibleTree } from '@packages/frontend-shared/src/composables/useCollapsibleTree'
+import { buildSpecTree, FuzzyFoundSpec, useCollapsibleTree } from './tree/useCollapsibleTree'
+import { fuzzySortSpecs, makeFuzzyFoundSpec, useCachedSpecs } from './spec-utils'
 import RowDirectory from './RowDirectory.vue'
 import SpecItem from './SpecItem.vue'
-import { useVirtualList } from '@packages/frontend-shared/src/composables/useVirtualList'
+import { useVirtualList } from './tree/useVirtualList'
```
marktnoonan, line 190: Since this has been backported to VueUse, do we want to try using the VueUse implementation: https://vueuse.org/core/usevirtuallist/#usevirtuallist

I wonder if we have diverged or if this will just work.

marktnoonan:
Nice to clean some of this up! I have no problems with this move, though I had been thinking in an almost opposite way - that possibly more things should begin their life in `frontend-shared` since extracting things from one of the other packages was relatively painful and time consuming recently around auth/cloud stuff. Had to touch a lot tests and components just to use something we had already built and get it stable again everywhere.

This wouldn't happen if we saw the `app` and `launchpad` packages as sort of thin wrappers, and had `frontend-shared` as the core frontend source of components for both. Then we would always be able to reuse design elements in different contexts.

Not blocking though, just something I meant to bring up.

Is this ready for review? It's still in draft.

packages/app/src/specs/SpecsList.vue
lmiller1990: Nice, we should do this. Not right here, though -- I don't want to introduce any risk or change to code in this PR where possible - this is a quick reshuffling of things.

lmiller1990:
> ... that possibly more things should begin their life in `frontend-shared` since extracting things from one of the other packages was relatively painful ...

Why was it painful? I wasn't involved in this, but it should be as simple as moving and updating some imports. If it's not, our code is coupled to our build infrastructure - which means we should fix that. In this PR, I just did a bunch of dragging around VS Code and a few find and replaces.

> This wouldn't happen if we saw the app and launchpad packages as sort of thin wrappers, and had frontend-shared as the core frontend source of components for both. Then we would always be able to reuse design elements in different contexts.

I do kind of agree with this. I think generic components are fine to include in `frontend-shared`. The main thing I'm moving here is `useVirtualList`, which may even go away once we update to the latest VueUse.

That said, the opposite end of argument would be "don't over abstract; refactor and make generic *if* you need it.

I don't feel too strongly about this either way, the main reason I made this PR is I think it'll make reviewing the upcoming Run All Specs changes a little easier, since we are probably touching that area quite a bit.

> Is this ready for review? It's still in draft.

Moved to review!

rockindahizzy:
Just adding my .02. I'm all for [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), and keeping things as minimal as possible. However, I would suggest that we commit to starting the life of a component in frontend-shared. In addition to Mark's viewpoints, it also makes it much easier for a developer to discover an implementation if something already exists.  We should be careful to not over generalize a component just because it lives here, though. Let the re-usablity be implemented as needed.

ZachJW34 (approved):

marktnoonan (approved):

marktnoonan:
Happy to approve, especially as a broken-out part of other work that makes reviews easier 👏 

> Why was it painful? I wasn't involved in this, but it should be as simple as moving and updating some imports. If it's not, our code is coupled to our build infrastructure

Moving the components was OK, the challenges were mostly around graphql data that the components we moved needed to access. Moving/renaming components meant updating a lot of GQL intercepts, for example, since the naming convent for fragments is tied to the component structure, and a few other things. We can chat more about it some time. Probably ties more into our test infra than the actual location of our files.

@rockindahizzy I'd be fine with "new components are created in frontend-shared by default" and not moving the old stuff right away. We can update the frontend-shared README to guide people to do this, and discuss whether the team agrees in that PR.

PR #24675, opened 2022-11-14
fix: migrating from <10 to current with typescript

### LINKED ISSUE #24643: Migration in V11 not working have to go over 10
### Current behavior

Hello Friends,

today I wanted to upgrade a E2E testing project from version 9 to 11. The migration window "5 minutes" can not finishwork and was there endless. 

Solution was to update from 9 to 10 and then install 11.

### Desired behavior

That a upgrade from V9 to V11 just will work.

### Test code to reproduce

Do not have it public

### Cypress Version

9

### Node version

14.7

### Operating System

Ubuntu 22

### Debug Logs

_No response_

### Other

_No response_

ryanthemanuel:
- Closes #24643 

### User facing changelog

Users can migrate from Cypress versions < 10 to 11 without hanging

### Additional details

The prettier dependency that we use to format code in the 10 migration is not properly being rewritten during the snapshot doctoring process. I am forcing the portions of that dependency that we rely on to be force norewrite so that this no longer happens.

### Steps to test

Migrate a project that uses typescript from 9 -> 11.

The app no longer hangs when migrating from 9 -> 11

### PR Tasks


emilyrohrbough:
Would a system test have caught this error?

tooling/v8-snapshot/src/setup/force-no-rewrite.ts
```diff
@@ -58,4 +58,11 @@ export default [
   'packages/server/lib/open_project.ts',
   'packages/server/lib/project-base.ts',
   'packages/server/lib/socket-ct.ts',
+  'node_modules/prettier/index.js',
+  'node_modules/prettier/parser-babel.js',
+  'node_modules/prettier/parser-espree.js',
+  'node_modules/prettier/parser-flow.js',
+  'node_modules/prettier/parser-meriyah.js',
+  'node_modules/prettier/parser-typescript.js',
+  'node_modules/prettier/third-party.js',
```
emilyrohrbough, line 67: how do we know this won't change under the hood with new versions of `prettier` or if we use new imports from `prettier`?

ryanthemanuel: Unfortunately, we don't have a mechanism to verify this today. cypress in cypress testing does not use snapshots due to how the child cypress server is being launched. This is something I want to investigate a solution to ASAP, but for now I think it's worth getting this fix in.

emilyrohrbough: Does this need to be manually verified at release then for each release?

ryanthemanuel: I think we would only need to verify it if we update the version of prettier that is being used: https://github.com/cypress-io/cypress/blob/develop/packages/data-context/package.json#L45

mjhenkes (approved):
I'll approve this in the spirit of getting the fix out however it would be nice to have tests to insure that we don't regress here in the future.

marktnoonan (approved):

PR #24680, opened 2022-11-14
feat: add cloud recommendation message to CI output

### LINKED ISSUE #24000: Add cloud recommendation message to CI output
[ClickUp](https://app.clickup.com/t/18033298/PM-2552)

Add the following message to the CI output for `cypress run`:
```bash
-----------------------------------------------------------------------------------------

Having trouble debugging your CI failures?
Record your runs to the Cypress Cloud to watch video recordings for each test, debug failing and flaky tests, and integrate with your favorite tools.

>> https://on.cypress.io/cloud-get-started

------------------------------------------------------------------------------------------
```
This message should show after the test summary (in place of the recorded run link) in run mode if the following conditions are met:
- we are in run mode
- the run is **not** being recorded to the dashboard
- we are running in CI
- at least one test failed during the run
- `CYPRESS_NO_COMMERCIAL_RECOMMENDATIONS` environment variable is not set to `true` (users can opt out of seeing this message by setting this true)

astone123:
- Closes https://github.com/cypress-io/cypress/issues/24000

### User facing changelog

n/a

### Additional details

This PR adds a message that prompts the user to go to Cypress Cloud after the test summary table only when **all** of the following conditions are true:
- Cypress is in run mode
- The process is running in CI (using `is-ci` library to determine this)
- `CYPRESS_COMMERCIAL_RECOMMENDATIONS` environment variable is **not** set to `0`
- At least one test in the run failed
- `--record` flag was not passed to the cypress run command (the user is not attempting to record this run to Cypress Cloud)

Services PR for on link: https://github.com/cypress-io/cypress-services/pull/4972
Docs PR: https://github.com/cypress-io/cypress-documentation/pull/4863

### Steps to test

Look at the output from [this run](https://app.circleci.com/pipelines/github/cypress-io/cypress/45747/workflows/fc2c3edc-752b-4adc-887a-0fc9cc3d31e4/jobs/1924189) and verify that the cloud recommendation message appears after the test summary. Note that Circle CI is redacting `cypress.io` from the on link. This is a setting in our Circle CI project and shouldn't be an issue in other's projects.

You can also look at the tests that I added in `cypress_spec.js` and run those

Screen Shot 2022-11-17 at 2 46 33 PM

### PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4863


packages/server/lib/util/print-run.ts
```diff
@@ -22,6 +22,19 @@ type Screenshot = {
   specName: string
 }
 
+export const cloudRecommendationMessage = `
+-----------------------------------------------------------------------------------------
+
+  Having trouble debugging your CI failures?
+  
+  Record your runs to the Cypress Cloud to watch video recordings for each test,
```
lmiller1990: cc @pstakoun, good to get product :eyes:  for any public facing marketing call to action messages.

packages/server/lib/util/print-run.ts
```diff
@@ -274,7 +287,17 @@ export function displaySpecHeader (name: string, curr: number, total: number, es
   }
 }
 
-export function renderSummaryTable (runUrl: string | undefined, results: any) {
+function maybeLogCloudRecommendationMessage (runs: CypressCommandLine.RunResult[], record?: boolean) {
```
lmiller1990: Hmm what do you think either

1. make `record` non nullable
2. or have a default value

Just a personal preference, but I like for booleans to be explicit, so we can communicate the expected default value to future developers.

packages/server/lib/util/print-run.ts
```diff
@@ -274,7 +287,17 @@ export function displaySpecHeader (name: string, curr: number, total: number, es
...
+  if (!process.env.CI || env.get('CYPRESS_NO_COMMERCIAL_RECOMMENDATIONS') || record) {
```
lmiller1990: Shouldn't this be `!record`? Eg, if we are NOT recording, return. Otherwise, we might log the message.

packages/server/lib/util/print-run.ts
```diff
@@ -274,7 +287,17 @@ export function displaySpecHeader (name: string, curr: number, total: number, es
...
+    return
+  }
+
+  if (runs.some((run) => run.stats.failures > 0)) {
+    console.log(cloudRecommendationMessage)
+  }
+}
+
+export function renderSummaryTable (runUrl: string | undefined, results: CypressCommandLine.CypressRunResult, record?: boolean) {
```
lmiller1990: Nit, but same here - can we make this non optional or have a default value?

packages/server/test/integration/cypress_spec.js
```diff
@@ -349,6 +350,48 @@ describe('lib/cypress', () => {
...
+
+        return cypress.start([`--run-project=${this.todosPath}`]).then(() => {
+          expect(console.log).not.to.be.calledWith(cloudRecommendationMessage)
+        })
+      })
+
+      it('does not get logged if not running in CI', function () {
+        delete process.env.CI
+        delete process.env.CYPRESS_NO_COMMERCIAL_RECOMMENDATIONS
+        globalThis.CY_TEST_MOCK.listenForProjectEnd = { stats: { failures: 1 } }
```
lmiller1990, line 397: Nice job finding a way to test this, unit testing `run.js` is often tricky.

lmiller1990 (changes requested):
I *think* there is logic bug - please confirm I'm not going crazy. 

Other than that, code and tests look :100:

packages/server/lib/util/print-run.ts
astone123: No, if the run is being recorded (if `--recorded` is passed, making `record = true`) then we want to exit early and **not** show the message.

We only want to suggest setting up on Cypress Cloud if they are **not already** recording the run. Maybe I could refactor the logic here to make things more clear?

packages/server/lib/util/print-run.ts
```diff
@@ -369,6 +392,8 @@ export function renderSummaryTable (runUrl: string | undefined, results: any) {
       console.log(`  Recorded Run: ${formatPath(runUrl, undefined, 'gray')}`)
       console.log('')
     }
+
+    maybeLogCloudRecommendationMessage(runs, record)
```
ZachJW34: Nit: I feel like this function call doesn't belong in this function but rather right after the `renderSummaryTable`.

packages/server/test/integration/cypress_spec.js
```diff
@@ -349,6 +350,48 @@ describe('lib/cypress', () => {
...
+        delete process.env.CYPRESS_NO_COMMERCIAL_RECOMMENDATIONS
+        globalThis.CY_TEST_MOCK.listenForProjectEnd = { stats: { failures: 0 } }
+
+        return cypress.start([`--run-project=${this.todosPath}`]).then(() => {
+          expect(console.log).not.to.be.calledWith(cloudRecommendationMessage)
+        })
+      })
+
+      it('does not get logged if not running in CI', function () {
+        delete process.env.CI
```
ZachJW34: Does `process.env` get restored after each test? When manipulating global state, I like to have an `after` or `afterEach` that sets the global state back to what it should be such that, if the test fails, it doesn't pollute any other state

ZachJW34 (approved):
Tested it out, works great! Couple small comments

ZachJW34:
@astone123 Does deleting the CircleCI `CYPRESS_NO_COMMERCIAL_RECOMMENDATIONS` have any impact on this being merged?

astone123:
@ZachJW34 yeah, I need that variable in the project in order for the tests to pass, so I need to put it back before I push a follow-up commit, but that will cause everyone else's tests to fail until this branch gets merged... I'll put something in Slack to explain the situation once this is ready to merge

astone123:
Waiting on this for now - going to meet with @pstakoun and Brian tomorrow to get final approval

packages/server/test/integration/cypress_spec.js
astone123: It looks like `process.env` is restored after each test. For instance, if I have `CYPRESS_NO_COMMERCIAL_RECOMMENDATIONS=true` set in the Circle project environment and then I have
```js
delete process.env.CYPRESS_NO_COMMERCIAL_RECOMMENDATIONS
```
in a test, when the next test runs, `process.env` will still have
```js
{
  CYPRESS_NO_COMMERCIAL_RECOMMENDATIONS: true
}
```
even though I deleted it in the test before it

packages/server/test/integration/cypress_spec.js
astone123: yeah, it was tricky. one thing that I found (and I still don't understand why this happens) is that I couldn't get my test case to work like this

```js
it('does something', () => {
  // my test
})
```
and I had to declare the callback function like this

```js
it('does something', function () {
  // my test
})
```

it took me a while to figure out that my test was failing because I was trying to define the callback as an arrow function 🤔 still don't know why

packages/server/lib/util/print-run.ts
astone123: Good call 👍🏻

packages/server/lib/util/print-run.ts
lmiller1990: Right, gotcha, figured I was misreading this (I guess it's not immediately clear, or I just derped out... seems fine).

lmiller1990 (approved):
💯

brian-mann (commented):
.

brian-mann (changes requested):
- Might want to switch to using the `is-ci` npm utility since we use that elsewhere
- Consider where to document this env var in the docs (maybe installation, with another section in Configuration that enumerates env vars that control behavior - with links to other areas)
- The name of the env var sounds good - we may want to align this to the existing `CYPRESS_CRASH_REPORTS` since that doesn't use the "NO" verbiage
- Instead of hard coding the line dividers, go through the same utility methods bc I believe they dynamically adjust their width to the available space in the terminal
- Let's iterate on the colors, formatting, line dividers, to make the message less aggressive
- After iterating just drop in a picture of the formatting and we can do async reviews

astone123:
@brian-mann thoughts on this appearance?

Screen Shot 2022-11-16 at 4 07 28 PM

tooling/v8-snapshot/src/setup/force-no-rewrite.ts
```diff
@@ -69,4 +69,7 @@ export default [
   'node_modules/prettier/parser-meriyah.js',
   'node_modules/prettier/parser-typescript.js',
   'node_modules/prettier/third-party.js',
+  'packages/server/node_modules/is-ci/index.js',
```
ryanthemanuel, line 72: Why do these need to be in the force-no-rewrite category?

astone123: I'm trying to figure out exactly why, but it wouldn't work without adding those (they probably don't _all_ need to be in there).

I think what might be happening is that the exported value from `is-ci` is getting evaluated and snapshotted before `process.env.CI=true` is set, and then it never gets re-evaluated? 

If I don't have these in this file then `is-ci` returns `false` even though we are running in CI and it should be `true`. @ZachJW34 originally suggested trying this when we were working through it

ZachJW34: @astone123 and I paired on this a bit. Not sure we need every one of these but it wasn't working without this addition. I believe it's due to how `is-ci` and `ci-info` are written, where the export is evaluated upon require e.g. `module.exports = !!process.env.CI` and not a function.

Could be totally wrong though!

ryanthemanuel: I think from playing around locally all we need to add here is the first two and I'm good with that for now. Though I still want to figure a little more on why.

brian-mann (approved):

Issue #24548, opened 2022-11-04
[Run All Specs] implement experimentalRunAllSpecs config flag


marktnoonan:
We need an `experimentalRunAllSpecs` Boolean property in the config. Its default value should be `false`. When configured, by users, as `true` it will allow the UI features for running groups of specs to be displayed.

## Acceptance Criteria

`experimentalRunAllSpecs` is a valid top-level boolean config option. The experiment might not support component testing at the outset but it's the kind of option that could apply to both testing types, so no need to nest it under `e2e`.

`experimentalRunAllSpecs` should appear in the Experiments section of the settings page:

<img width="1135" alt="Screen Shot 2022-11-08 at 7 46 36 AM" src="https://user-images.githubusercontent.com/8340719/200571853-e4babb40-189c-44ad-8456-588bd33fe5dd.png">

A PR to the Cypress docs must be made at the same time as this flag is added.

baus:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/24548?planning-poker) with Zenhub @amehta265 @astone123 @lmiller1990 @marktnoonan @mike-plummer @rockindahizzy @warrensplayer @ZachJW34

baus:
Add docs PR

Issue #24468, opened 2022-10-31
[Run All Specs] - Inline Spec Explorer - Add "run N specs" button to folder rows


marktnoonan:
### Summary

Currently, users can filter specs and choose a single one to run in open mode. They need the ability to run the currently-filtered list of specs, which replicates the functionality of Run All in 9.x. This implementation adds an extra ability not possible in 9.x: to further select only a specific folder, to run the subset of specs _currently matching the filter_ in that folder (child folders included). This is achieved with a "play" button present on each folder in the inline spec explorer.

#### Empty state, no visual change
<img width="385" alt="Screen Shot 2022-11-08 at 7 31 03 AM" src="https://user-images.githubusercontent.com/8340719/200564761-985c6347-b485-4812-a271-22fe0ba7e724.png">

#### Hover on row but not button
<img width="364" alt="Screen Shot 2022-11-08 at 7 31 18 AM" src="https://user-images.githubusercontent.com/8340719/200564806-fd247e08-9e61-49a5-a8ba-fba7e7664bb7.png">

#### Hover on button itself
<img width="566" alt="Screen Shot 2022-11-08 at 7 31 44 AM" src="https://user-images.githubusercontent.com/8340719/200564843-de652e06-f741-4253-8228-c0fb2c1d6e0d.png">

### Acceptance Criteria

**In e2e mode only:**

1. "Play" buttons should exist on all folders
2. On top-level folder, have the tooltip "Run N specs" and run all specs that match the current search
3. On all other folders, have the tooltip "Run N specs" where N is the number of specs in the folder that match the current search and those should be the ones run
4. The button appears when the folder row is hovered
5. Buttons have their own hover state and are activated on click
6. The behavior of clicking the row is still to toggle the open/closed state of the folder
7. When clicked, the button takes the user to the Spec Runner page and runs the selected specs.
8. Button should be reachable by the keyboard and appear when focussed with the tab key

### Resources

Figma prototype: https://www.figma.com/proto/PAOozuCxtfetqP8fUVvg0y/branch/WpNw4mG3LvD1J3zLj80KCr/Cypress-App?node-id=20720%3A150686&starting-point-node-id=20668%3A150076

baus:
[...]

Issue #24585, opened 2022-11-08
[Run All Specs] Spike: Handle "Run All Specs" for Component Testing


marktnoonan:
### What would you like?

This was not a feature of Cypress 9 and below. Once https://github.com/cypress-io/cypress/issues/24469 is implemented and working for e2e we should see what the level of effort is to get "Run All" working for component testing.

### Outcome of the spike

Create an issue detailing the changes needed to support "run all" for CT that can be estimated.

ZachJW34:
Spiked into this, got something rough working for both Webpack and Vite


https://user-images.githubusercontent.com/25158820/202501358-442f68f3-0e6f-44e1-9278-a6cf10be21fe.mov



https://github.com/cypress-io/cypress/tree/zachw/run-all-specs-ct

### Summary

The mechanism by which specs are loaded on the frontend are similar between e2e and ct. Whenever a user chooses a spec, an iframe is created with the url containing the path to the spec file to execute. This in turn makes a network request to the backend that resolved an html file that executes:

```js
// e2e 
Cypress.onSpecWindow([
  { requestUrl: 'http://localhost:xxxx/<path-to-support-file>' },
  { requestUrl: 'http://localhost:xxxx/<path-to-spec-file>' },
])

// ct
Cypress.onSpecWindow([
  { load: () => import('<path-to-support-file>') },
  { load: () => import('<path-to-spec-file>') },
])
```

Run all specs relies on changing the payload given to `onSpecWindow` to contain not just `support + spec` but `support + all specs selected`. This is done by changing the iframe url to contain the `__all` keyword allowing the backend to generate an html containing this expanded payload.

Cypress e2e is a bit more straightforward as specs are compiled on demand with Webpack. As you as the `requestUrl` resolves a javascript file, we will attempt to compile and serve it.

Cypress ct differs in that the dev-server is already up and running, so the list of specs needs to be injected into the currently running compilation (either Vite or Webpack). This process is not much different from how the dev-server gets alerted to specs that have been added or deleted, utilizing the dev-server ipc to communicate from the main Cypress process to the config/dev-server process.

### Implementation

#### Webpack

[...]

PR #24683, opened 2022-11-14
feat: re-enable Run All Specs

### LINKED ISSUE #24469: [Run all specs] restore legacy "run all" behavior and add "folder level" filter
### Summary

In the [Run All Specs spike](https://github.com/cypress-io/cypress/issues/23471) @lmiller1990 discussed several options for restoring this behavior. We've chosen Option 1, "Port the exact implementation from Cypress 9." - with one extra wrinkle, the ability to narrow the filtered specs to just those contained underneath a particular folder. So this feature needs to understand "filter + folder" and run a subset of the filtered specs, unless the folder is root of all filtered specs.

To answer some of the question's Lachlan raised in the epic:

> 1. We support Component Testing now. Do we need the same behavior for Component Testing? One point of different is Component Testing works with Vite and Webpack, and uses a dev server. I'm not sure how much complexity this adds, but it's complexity we will need to consider.

The initial implementation does not need to support component testing unless it's very low complexity. It will be fine for the scope of the experiment to expand over time to CT. CT didn't have "run all" in 9.x so there is no issue of parity.

> 2. Where does the "run all specs" button go? Does it go on the first page you see (the "specs" page) or on the "runner" page?
> 3. How can I choose multiple specs? We need checkboxes. Where do those go?

#24467 and #24468 provide the UI

> 4. What if I want to run a different set of specs? One of the main problems with Run All Specs is memory usage. In Cypress 9, when you choose a new combination of specs to run, we re-launch the browser, which releases all the memory. In Cypress 10, you only open the browser once.

For the experimental release, we will match the 9.x behavior and relaunch the browser.

### Acceptance Criteria

1. When all specs in a folder are run, all specs in its child folders that also match the current search are also run
2. When all specs are run, they use "legacy" Run All Specs behavior and simply concat all the specs into a single mega-spec
3. Run All Specs should be behind an experimental flag called `experimentalRunAllSpecs` at the root level of the config

Future note: We will then iterate on this behavior to make it "correct" and deterministic, but for the context of the experiment, the old behavior is correct.

ZachJW34:
- Closes #24469, #24548

### User facing changelog
Brings back the "Run All Specs" feature via the opt in `experimentalRunAllSpecs` flag

### Additional details
Most of the wiring was still there so not a big lift. The UI is temporary and will be done in #24468 and #24467.

It feels much better to not kill the browser when clicking a new set to run. Could we ignore the memory implications since this feature is experimental?

### Steps to test
Check out this branch, run `yarn watch` in one terminal and `yarn cypress:open` in another once the watch is finished. Create a new Cypress project and add `experimentalRunAllSpecs: true` to the config. There will be an ugly button at the top of both spec-lists that will trigger the feature based on the current search filter.

https://user-images.githubusercontent.com/25158820/201992983-e337f85b-50af-456e-ae19-a83c1a77b987.mov

### PR Tasks


ZachJW34:
Docs PR: https://github.com/cypress-io/cypress-documentation/pull/4860

packages/server/lib/controllers/files.js
```diff
@@ -69,54 +68,25 @@ module.exports = {
...
-        })
-        .filter(specFilterFn)
-        .then((specs) => {
-          debug('filtered __all specs %o', specs)
-
-          return specs
-        }).map(convertSpecPath)
+        // In case the user clicked "run all specs" and deleted a spec in the list, we will
+        // only include specs we know to exist
+        const existingSpecs = new Set(ctx.project.specs.map(({ relative }) => relative))
```
emilyrohrbough, line 80: Does this add a watcher for each spec currently running?

emilyrohrbough (commented):
Just curious, how are we ensuring a clean browser context between specs? Are we leveraging the same mechanisms as experimental?

How many specs/tests can run before the Chrome browser crash is observed? https://github.com/cypress-io/cypress/issues/23391

cli/types/cypress.d.ts
```diff
@@ -2995,6 +2995,12 @@ declare namespace Cypress {
     setupNodeEvents: (on: PluginEvents, config: PluginConfigOptions) => Promise<PluginConfigOptions | void> | PluginConfigOptions | void
 
     indexHtmlFile: string
+
+    /**
+     * Enables the "Run All Specs" UI feature, allowing the execution of multiple specs at once (only available for e2e).
```
emilyrohrbough: Just curious, why isn't this available for component testing?

packages/config/src/options.ts
```diff
@@ -204,6 +204,11 @@ const driverConfigOptions: Array<DriverConfigOption> = [
     validation: validate.isBoolean,
     isExperimental: true,
     requireRestartOnChange: 'server',
+  }, {
+    name: 'experimentalRunAllSpecs',
+    defaultValue: false,
+    validation: validate.isBoolean,
+    isExperimental: true,
```
emilyrohrbough, line 211: You'll need to update [@packages/server/lib/experiments](https://github.com/cypress-io/cypress/blob/develop/packages/server/lib/experiments.ts#L53) to include this option & description for the App setting page

cli/types/cypress.d.ts
ZachJW34: Requirements were to just implement the `Run All Specs` as it was in 9.x. I don't think it would be too complicated to get it to work with component but it won't work out of the box. There would need to be some wiring to communicate the selected specs to the dev-server process to make sure the specs are included in the compilation. A bit easier with e2e since the compilation is done on demand

emilyrohrbough: Gottcya, wasn't sure what the expectations / standards were around supporting new features/experiments in both e2e & ct now that CT has gone GA.

packages/config/src/options.ts
ZachJW34: The App settings page was working without this addition, but look like it is used for run-mode terminal printing
https://github.com/cypress-io/cypress/blob/55f6972f1d4a43abccc32b501b2df0b2987f4c17/packages/server/lib/util/print-run.ts#L186

packages/server/lib/controllers/files.js
ZachJW34: No, this just ensures that the saved list of filtered specs doesn't deviate from runnable specs. The watcher is handled in `webpack-preprocessor`

ZachJW34:
@emilyrohrbough I'm looking into the browser state between tests. From what I understand, `Run All Specs` is roughly the equivalent to taking the contents of all specs selected and running them as if they were a single spec. All of the specs get fetched then eval'd, and once that is done the test starts running.

As for memory, I just ran the kitchen-sink example duplicated... alot. I'll see what else I can throw at it
Screen Shot 2022-11-15 at 4 59 22 PM

packages/app/src/composables/useRunAllSpecs.ts
```diff
@@ -0,0 +1,49 @@
...
+
+      // Won't execute unless we are testing since the browser gets killed. In testing,
+      // we can stub `launchProject` to verify the functionality is working
+      router.push({ path: '/specs/runner', query: { file: '__all' } })
+    },
+    isRunAllSpecsAllowed: computed(() => {
+      const isE2E = query.data.value?.currentProject?.currentTestingType === 'e2e'
+
+      const config: ResolvedConfig = query.data.value?.currentProject?.config || []
+      const hasExperiment = config.find(({ field, value }) => field === 'experimentalRunAllSpecs' && value === true)
```
lmiller1990, line 45: I think we can actually assume that if `experimentalRunAllSpecs`, they are in a mode that supports it (E2E only right now). If they attempt to set this for CT, we should raise an error, like we do if they try to use Studio from CT.

packages/app/src/composables/useRunAllSpecs.ts
```diff
@@ -0,0 +1,49 @@
...
+  return {
+    runAllSpecs: async (runAllSpecs: string[]) => {
+      await setRunAllSpecsMutation.executeMutation({ runAllSpecs })
+
+      // Won't execute unless we are testing since the browser gets killed. In testing,
+      // we can stub `launchProject` to verify the functionality is working
+      router.push({ path: '/specs/runner', query: { file: '__all' } })
+    },
+    isRunAllSpecsAllowed: computed(() => {
+      const isE2E = query.data.value?.currentProject?.currentTestingType === 'e2e'
```
lmiller1990, line 42: Since `query.data.value` is a `ref` I don't think this `computed` is necessary?

packages/app/src/runner/unifiedRunner.ts
```diff
@@ -35,6 +35,10 @@ export function useUnifiedRunner () {
           return
         }
 
+        if (queryFile === '__all') {
```
lmiller1990: Should we use the constant 

```suggestion
        if (queryFile === RUN_ALL_SPEC['fileName']) {
```

packages/app/src/runner/unifiedRunner.ts
```diff
@@ -35,6 +35,10 @@ export function useUnifiedRunner () {
...
+          return specStore.setActiveSpec(RUN_ALL_SPEC)
```
lmiller1990: Should it be `RUN_ALL_SPECS` (with an `S` at the end)?

packages/app/src/specs/SpecsList.vue
```diff
@@ -417,6 +420,20 @@ const { refetchFailedCloudData } = useCloudSpecData(
   props.gql.currentProject?.specs as SpecsListFragment[] || [],
 )
 
+const { runAllSpecs, isRunAllSpecsAllowed } = useRunAllSpecs()
```
lmiller1990, line 423: This will launch an additional GraphQL query - was this the intention? Should we follow the pattern of using a `fragment` and spreading it over the top level query, so we just have a single query for the whole page?

I'm not sure we have finalized the pattern here, but this is what we are doing so far in most of the application.

I'm guessing you set this up with the intention of making it re-usable in the runner page, too. Can you speak a bit more to how you see this working across both the spec page and the runner page?

packages/app/src/specs/SpecsList.vue
```diff
@@ -417,6 +420,20 @@ const { refetchFailedCloudData } = useCloudSpecData(
...
+  const filteredSpecs = treeSpecList.value.reduce((acc, node) => {
+    if (node.isLeaf && node.id.startsWith(directory)) {
+      acc.push(node.data?.relative!)
+    }
+
+    return acc
+  }, [] as string[])
+
+  runAllSpecs(filteredSpecs)
+}
```
lmiller1990: I think it's more idiomatic to do the typing as the first type to `reduce`.

```suggestion
function onRunAllSpecs (directory: string) {
  const filteredSpecs = treeSpecList.value.reduce<string[]>((acc, node) => {
    if (node.isLeaf && node.id.startsWith(directory)) {
      acc.push(node.data?.relative!)
    }

    return acc
  }, [])

  runAllSpecs(filteredSpecs)
}
```

Although I'm guessing this will change anyway with the Spec List refactor + work Ankit is doing on the UI.

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -230,6 +231,14 @@ export class ProjectDataSource {
     this._specs = specs
   }
 
+  get runAllSpecs () {
+    return this._runAllSpecs
+  }
+
+  set runAllSpecs (runAllSpecs: string[]) {
```
lmiller1990: Seems this might be more understandable as a method. 

```js
setRunAllSpecs
```

`ctx.setRunAllSpecs(...)` seems more preferrable that `ctx.runAllSpecs = ...`. Right now you have a private var, but you basically make it public with the `set runAllSpecs`.

lmiller1990: Maybe

```js
#runAllSpecs: string[] = []

get runAllSpecs () {
  return this.#runAllSpecs
}

setRunAllSpecs(specs) {
  this.#runAllSpecs = specs
}
```

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
```diff
@@ -737,5 +737,17 @@ export const mutation = mutationType({
         return data
       },
     })
+
+    t.boolean('setRunAllSpecs', {
+      description: 'List of specs to run for the "Run All Specs" Feature',
+      args: {
+        runAllSpecs: nonNull(list(nonNull(stringArg({})))),
```
lmiller1990:
```suggestion
        runAllSpecs: nonNull(list(nonNull(stringArg()))),
```

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
```diff
@@ -737,5 +737,17 @@ export const mutation = mutationType({
...
+      },
+      resolve: async (source, args, ctx) => {
```
lmiller1990:
```suggestion
      resolve: (source, args, ctx) => {
```

packages/server/lib/controllers/files.js
```diff
@@ -69,54 +68,25 @@ module.exports = {
...
+        const filteredSpecs = ctx.project.runAllSpecs.reduce((acc, relSpec) => {
```
lmiller1990, line 81: Took me a moment to figure this out, but got it now

cli/types/cypress.d.ts
lmiller1990: Studio is E2E only too - it raises an error in launchpad if you try to use it with CT. It might make sense to follow that pattern here, too.

packages/app/src/specs/SpecsList.vue
```diff
@@ -111,8 +111,10 @@
               :depth="row.data.depth - 2"
               :style="{ paddingLeft: `${(row.data.depth - 2) * 10}px` }"
               :indexes="row.data.highlightIndexes"
+              :is-run-all-specs-allowed="isRunAllSpecsAllowed"
               :aria-controls="getIdIfDirectory(row)"
               @click.stop="row.data.toggle"
+              @run-all-specs="onRunAllSpecs(row.data.id)"
```
lmiller1990: Nit... can this be `@runAllSpecs`? I think it is much more understandable for the emitted event and the listener to have the same casing. We also generally use camelCase for events in the app, from what I can see.

packages/app/src/specs/SpecsList.vue
```diff
@@ -417,6 +420,20 @@ const { refetchFailedCloudData } = useCloudSpecData(
   props.gql.currentProject?.specs as SpecsListFragment[] || [],
 )
 
+const { runAllSpecs, isRunAllSpecsAllowed } = useRunAllSpecs()
+
+function onRunAllSpecs (directory: string) {
+  const filteredSpecs = treeSpecList.value.reduce((acc, node) => {
+    if (node.isLeaf && node.id.startsWith(directory)) {
```
lmiller1990, line 427: Right, this is how you got the files and count. Maybe we didn't *need* the refactor I suggested! Either way, I think the refactor is valuable - `SpecsList` is really complex right now.

lmiller1990 (commented):
Some more comments/questions, seems to be working as expected.

lmiller1990:
@emilyrohrbough 
> Just curious, how are we ensuring a clean browser context between specs? Are we leveraging the same mechanisms as experimental?

We don't clean the context or do anything special to my knowledge - nor did we in 9.x, which is why this feature was confusing for many users... the specs can bleed into each other.

This is a known limitation we are consciously moving forward with as part of the experiment, but this will need solving before the Run All Specs becomes GA.

cli/types/cypress.d.ts
emilyrohrbough: @lmiller1990 maybe...but studio was also added before CT went GA. It'd be worth bringing the converstation up in our cross-team sync.

packages/app/src/specs/SpecsList.vue
ZachJW34: It doesn't launch an additional graphql query, since both the `config` and `currentTestingType` is already in the cache so it can read it from there (you can also check the network logs). I've seen a shift away from the previous pattern e.g. [CloudViewerAndProject](https://github.com/cypress-io/cypress/blob/21315b0044069f837399b27658737cc368d0c960/packages/frontend-shared/src/gql-components/CloudViewerAndProject.vue#L84) and I personally like having a composable that fully encapsulates all the logic necessary for a certain task.

Even if it did make another network request, with how graphql is set up the cost would be very minimal (websocket request, only resolve a minimal amount of fields). I can refactor to use the previous pattern if you think this pattern doesn't scale, but the fragment spreading is also really annoying :(

I had this plugged up to both spec lists previously, almost exactly the same way. I didn't want to fix the UI up so I pulled it from there. The InlineSpecsList can just import this composable and do similar logic to filter the specs. The spec filtering could be pulled into the composable but I knew the UI work was looking to change how the tree was being built so I didn't want to pull too much in.

packages/app/src/specs/SpecsList.vue
ZachJW34: For sure, I think my IDE auto-formats that. I don't like the camelCase but I assumed it was linting doing a fix-up

Done in [b0ba5c2](https://github.com/cypress-io/cypress/pull/24683/commits/b0ba5c2b0e86e79183a9715b58e6fa22c9aa703a)

packages/data-context/src/sources/ProjectDataSource.ts
ZachJW34: I like it!

Done in [b0ba5c2](https://github.com/cypress-io/cypress/pull/24683/commits/b0ba5c2b0e86e79183a9715b58e6fa22c9aa703a)

packages/app/src/runner/unifiedRunner.ts
ZachJW34: I'll make this change!

Done in [b0ba5c2](https://github.com/cypress-io/cypress/pull/24683/commits/b0ba5c2b0e86e79183a9715b58e6fa22c9aa703a)

packages/app/src/composables/useRunAllSpecs.ts
ZachJW34: What would be the alternative? I'm deriving a reactive value from a reactive source, which I thought is exactly what `computed` is used for.

packages/app/src/composables/useRunAllSpecs.ts
ZachJW34: I'll add the same logic related to `experimentalStudio`, but it is a valid top-level property of the config so it still needs to be guarded here. So if top-level, no error but hide in UI (logic here). If nested under component, error in Launchpad.

Done in [b0ba5c2](https://github.com/cypress-io/cypress/pull/24683/commits/b0ba5c2b0e86e79183a9715b58e6fa22c9aa703a)

packages/app/src/runner/unifiedRunner.ts
ZachJW34: Done in [b0ba5c2](https://github.com/cypress-io/cypress/pull/24683/commits/b0ba5c2b0e86e79183a9715b58e6fa22c9aa703a)

packages/app/src/specs/SpecsList.vue
ZachJW34: Done in [b0ba5c2](https://github.com/cypress-io/cypress/pull/24683/commits/b0ba5c2b0e86e79183a9715b58e6fa22c9aa703a)

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
ZachJW34: Done in [b0ba5c2](https://github.com/cypress-io/cypress/pull/24683/commits/b0ba5c2b0e86e79183a9715b58e6fa22c9aa703a)

packages/app/src/specs/SpecsList.vue
ZachJW34: Yeah I tried using the `list` but then ran into an issue where it was only running a subset of the specs (`list` only contains an array of the _visible_ specs). It's... complex so I think the refactor is valuable. I know the mocks call for the UI to show the number of child elements on hover, which would be expensive to compute on demand which is how I'm doing it here.

ZachJW34:
On link PR for Run All Specs discussion: https://github.com/cypress-io/cypress-services/pull/4982

packages/graphql/src/schemaTypes/objectTypes/gql-Mutation.ts
ZachJW34: Done in [b0ba5c2](https://github.com/cypress-io/cypress/pull/24683/commits/b0ba5c2b0e86e79183a9715b58e6fa22c9aa703a)

cli/types/cypress.d.ts
marktnoonan: Definitely a good thing to bring up in the sync! There's an issue already for adding the CT piece: https://github.com/cypress-io/cypress/issues/24585

> wasn't sure what the expectations / standards were around supporting new features/experiments in both e2e & ct now that CT has gone GA.

We've got some notes on "release stages" from a doc we hoped to finalize around CT release, @rockindahizzy is working on a docs PR that will put those into a form we can discuss in detail. I don't think we have anything in there about minimum expectation for what counts as an "experiment" - seems to be very case-by-case. But we could add this kind of thing there.

In the spirit of getting the feature out so that people who miss "Run All Specs" can get it back, it seemed reasonable to split the CT part up so people can use it sooner. Being "e2e only" is just one of the many problems the 9.x implementation has.

I suspect if we were designing Run All from scratch today, we would not release this implementation even as an experiment, we'd solve more of the problems first. So even if we set particular standards, "Run All Specs" probably is an exception - restoring something we know is broken but that many people found useful even in its broken state.

packages/app/src/specs/SpecsList.vue
lmiller1990: Ah yeah, I figured you were relying on the cache to be populated. I think this is probably fine then - I agree spreading all the fragments is pretty annoying. I wonder if we can have a watcher automatically spread the fragments, that'd be a nice DX.

Anyway, let's move forward with this for now. I think this pattern is actually fine. I hope we can settle on "one way to do things" as we move forward, we really need to get a handle on App, it's getting messy.

packages/app/src/specs/SpecsList.vue
lmiller1990: Yeah, the refactor is going to massively simplify this, I don't think we will even need the cache and/or virtualization (hopefully).

packages/app/src/composables/useRunAllSpecs.ts
lmiller1990: Actually yes, I am not sure what I was thinking. :facepalm:

lmiller1990 (approved):
This is looking pretty good to me. I did a basic test, I'll do more extensive testing when the feature branch is finalized.

astone123 (approved):
This works well for me. Nice work 🎉

PR #24677, opened 2022-11-14
fix: junit reporters crashing on failed asserts

ryanthemanuel:
- Closes #24652 

### User facing changelog

Cypress will no longer crash when using the junit reporter on failed asserts

### Additional details

It appears that we have a delicate ordering for our logic to override mocha resolution. We need to ensure that it is loaded at the appropriate time or things get loaded incorrectly. I tried moving that loading around to various places in the code base, to no avail. The only thing I could get to work was to ensure that the dependencies in this PR don't defer when they load their dependencies. It's not ideal, but we now have a system test for this scenario that should prevent things like this in the future.

### Steps to test

Write a failing test and use the `junit` reporter

### How has the user experience changed?

`cypress run` no longer crashes when an assert fails

### PR Tasks

- [X] Have tests been added/updated?
- [na] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [na] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [na] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?

PR #23732, opened 2022-09-08
feat: incorporate v8 snapshot generation into monorepo build process

ryanthemanuel:
- Closes https://github.com/cypress-io/cypress/issues/22995
- Closes https://github.com/cypress-io/cypress/issues/22982

### User facing changelog

n/a

### Additional details

This PR does several things:

* Adds a script `tooling/v8-snapshot/dist/setup/setup.js` that can be run in `dev` or `prod` mode that generates a snapshot for the server in the monorepo
* Executes the above script:
  * In the post install script in `dev` mode locally
  * As an additional step of the `build` job in `prod` mode in CI
* Add a hook for `v8-snapshot-require` that will be executed on server startup to hook into the v8 snapshot

Incidental things that were fixed include
* Clean up unused code. The v8 snapshot packages were originally created to be fairly customizable and reusable. I removed the features that we don't immediately need for the sake of long term maintenance

Additionally, there are several existing issues that will be fixed in future iterations. The affected tests were commented out with a note on which future issue would be used to fix them

### Steps to test

You can pull the code locally, run `yarn install` and run `yarn cypress <command>` to verify that things work.

### How has the user experience changed?

n/a

### PR Tasks

- [X] Have tests been added/updated?
- [n/a] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [n/a] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [n/a] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?

PR #24684, opened 2022-11-14
fix: issue with loading typescript and v8 snapshot

ryanthemanuel:
### User facing changelog

N/A

### Additional details

No new functional fixes, but the combination of https://github.com/cypress-io/cypress/pull/24677 and https://github.com/cypress-io/cypress/pull/24675 exposed a timing problem. Specifically typescript is no longer detected in a consumer project. `require.resolve` was fixed on [this PR](https://github.com/cypress-io/cypress/pull/23732/files#diff-1b04229a5c2079a49a0dbfc1347476c1ffe708f89610952004c375d4a625a533) however there was a scenario that was missed to ensure that whenever we are calling `require.resolve` the original version is called when `paths` are provided.

### Steps to test

Use the binary from this PR and try and migrate a project that uses typescript from 9 to 11 and ensure that typescript is properly detected

n/a

### PR Tasks

### DIFF
--- a/packages/packherd-require/src/loader.ts
+++ b/packages/packherd-require/src/loader.ts
@@ -898,8 +898,14 @@ export class PackherdModuleLoader {
       parent = this._createModule(fullPath, parent, moduleUri)
     }
 
+    const originalRequireResolve = require.resolve
+
     require.resolve = Object.assign(
-      (moduleUri: string, _options?: { paths?: string[] }) => {
+      (moduleUri: string, options?: { paths?: string[] }) => {
+        if (options && options.paths) {
+          return originalRequireResolve(moduleUri, options)
+        }
+
         return this.tryResolve(moduleUri, parent).fullPath
       },
       {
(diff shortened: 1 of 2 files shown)


packages/packherd-require/src/loader.ts
```diff
@@ -898,8 +898,14 @@ export class PackherdModuleLoader {
       parent = this._createModule(fullPath, parent, moduleUri)
     }
 
+    const originalRequireResolve = require.resolve
+
     require.resolve = Object.assign(
-      (moduleUri: string, _options?: { paths?: string[] }) => {
+      (moduleUri: string, options?: { paths?: string[] }) => {
+        if (options && options.paths) {
```
ryanthemanuel, line 906:
```suggestion
        // Handle the case where options populated. The module is expected to be outside of the cypress snapshot so use the original require.resolve.
        if (options && options.paths) {
```

emilyrohrbough (approved):

mjhenkes (approved):

PR #24697, opened 2022-11-15
feat: Create public `Cypress.ensure` API for use with custom queries

BlueWinds:
### User facing changelog
Creates and adds types to `Cypress.ensure` API, which can be useful for creating custom queries. This new API is based off the previously undocumented `cy.ensure*` methods, which are now removed. 

In addition, adds public types to `cy.now()`, which is again useful for custom queries.

### Additional details
All the methods on `Cypress.ensure` are pre-existing, and were already on `cy`. This PR updates the arguments slightly for consistency, moves many of them to `Cypress`, and and adds publicly-facing types. Those that didn't end up on `Cypress.ensure` are split up around the codebase near where they're used.

These are referenced in the Custom Query documentation, which is "why this, why now." https://github.com/cypress-io/cypress-documentation/pull/4835

### Steps to test
Type tests have been added. All these commands are also used internally.

No change to the user experience for the vast majority of users. Plugin authors will have to do less TS juggling to get things to work.

### PR Tasks
- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4835


cli/types/cypress.d.ts
```diff
@@ -1163,6 +1163,15 @@ declare namespace Cypress {
      */
     end(): Chainable<null>
 
+    ensureElement(subject: any, commandName: string): void
```
BlueWinds: `any => void` is correct here, even for `ensureAttached`, which really only makes sense with DOM elements: you can pass in any subject, and these functions either do nothing and return nothing, or they throw an error.

cli/types/cypress.d.ts
```diff
@@ -1475,6 +1484,13 @@ declare namespace Cypress {
      */
     not(selector: string, options?: Partial<Loggable & Timeoutable>): Chainable<JQuery>
 
+    /**
+     * Invoke a command synchronously, without using the command queue.
+     *
+     * @see https://on.cypress.io/custom-queries
+     */
+    now(name: string, ...args: any[]): Promise<any> | ((subject: any) => any)
```
BlueWinds, line 1542: I tried to be more specific, with `name`, `args` and the return type mapping to `ChainableMethods`, but typescript ended up caught in a recursion, since `now` is one of those `ChainableMethods`.

What we really need to do is go through and separate out "things on `cy` (like now, off, on, once)" from "commands that can be chained (like click, get, scrollTo)", but that would involve a larger refactor of the types than I want to undertake in this PR.

packages/driver/src/cy/commands/misc.ts
```diff
@@ -14,15 +14,15 @@ export default (Commands, Cypress, cy, state) => {
   Commands.add('end', () => null)
   Commands.add('noop', (arg) => arg)
 
-  Commands.addQuery('log', (msg, ...args) => {
+  Commands.add('log', (msg, ...args) => {
```
BlueWinds, line 17: As per discussion in the docs PR, `.log()` should be a command and not a query.

Generally speaking, things that yield `null` don't make sense as queries, I just got overzealous and converted a few too many commands. :P

packages/driver/src/cy/ensures.ts
```diff
@@ -20,9 +20,7 @@ export const create = (state: StateFunc, expect: $Cy['expect']) => {
   // into an array and loop through each and verify
   // each element in the array is valid. as it stands
   // we only validate the first
-  const validateType = (subject, type, cmd = state('current')) => {
-    const name = cmd.get('name')
-
+  const validateType = (subject, type, name) => {
```
BlueWinds: I changed all of these to accept `name`, so that the signatures of the various "ensures" functions match with each other. Name is the only property we read off of `command` anyway. For a while I thought I needed the full command, but that ended up refactored away at some point, and just never got around to reverting the signatures.

The `name` / `command` arguments are new in Cy12, so we can be completely certain no one outside the monorepo is using them.

packages/driver/src/cypress/error_messages.ts
```diff
@@ -853,11 +853,11 @@ export default {
     },
     invalid_new_query: {
       message: '`Cypress.Commands.addQuery()` is used to create new queries, but `{{name}}` is an existing Cypress command or query, or is reserved internally by Cypress.\n\n If you want to override an existing command or query, use `Cypress.Commands.overrideQuery()` instead.',
-      docsUrl: 'https://on.cypress.io/custom-commands',
+      docsUrl: 'https://on.cypress.io/custom-queries',
```
BlueWinds, line 856: This is a new `on` link, for the new page I'm adding in the docs PR. I'll open a services PR to add it shortly.

packages/driver/cypress/e2e/commands/actions/check.cy.js
```diff
@@ -323,24 +323,27 @@ describe('src/cy/commands/actions/check', () => {
     })
 
     it('can set options.waitForAnimations', () => {
-      cy.stub(cy, 'ensureElementIsNotAnimating').throws(new Error('animating!'))
```
BlueWinds, line 326: Because I moved `ensureElementIsNotAnimating` to `$actionability`, which is not public-facing, it's no longer trivial to stub. I've moved these tests to use an actual jQuery animation, and watch the count of command retries - this treats Cypress more as a 'black box', rather than peering inside at exactly how animation is determined.

packages/driver/cypress/e2e/commands/actions/click.cy.js
```diff
@@ -3339,21 +3303,6 @@ describe('src/cy/commands/actions/click', () => {
         })
       })
 
-      // TODO: remove this after 4.0 when {multiple:true} is no longer default
```
BlueWinds, line 3342: TODO from Cypress 4.0, referencing a long-closed issue. TODO done.

Basically, `{multiple: true}` used to be the default, so it was hidden in the console logs; in Cy >=4.0.0, this is no longer the case, so there's no default value to hide.

packages/driver/src/cypress/cy.ts
```diff
@@ -121,6 +120,14 @@ const setTopOnError = function (Cypress, cy: $Cy) {
   top.__alreadySetErrorHandlers__ = true
 }
 
+const ensureRunnable = (cy, cmd) => {
```
BlueWinds, line 123: ensureRunnable is only used in this one file, so I moved it here to be close to where it's used.

packages/driver/src/cy/actionability.ts
```diff
@@ -53,9 +54,127 @@ const getPositionFromArguments = function (positionOrX, y, options) {
   return { options, position, x, y }
 }
 
+const ensureElDoesNotHaveCSS = ($el, cssProperty, cssValue, name: string, onFail) => {
```
BlueWinds, line 57: Several methods from ensures.ts are only used here, and don't really make sense to expose for people to use in queries. I've moved them into this file.

cli/types/cypress.d.ts
```diff
@@ -130,6 +130,53 @@ declare namespace Cypress {
...
+
+    /**
+     * Throws an error if `subject` is not a `document`.
+     */
+    isDocument(subject: any, commandName: string, cy: Chainable): void
+
+    /**
+     * Throws an error if `subject` is not a DOM element attached to the application under test.
+     */
+    isAttached(subject: any, commandName: string, cy: $Cy, onFail?: any): void
```
emilyrohrbough: Should `onFail()` be a function? same question for all of these with onFail parameters

packages/driver/src/cy/assertions.ts
```diff
@@ -276,10 +276,10 @@ export const create = (Cypress: ICypress, cy: $Cy) => {
               return
             }
 
-            return cy.ensureElExistence($el)
+            return Cypress.ensure.elExists($el, cy)
```
emilyrohrbough: In the types, the second parameter was typed as a string. Is `cy` correct?

emilyrohrbough: Actually, looks like this helper is missing from the types

packages/driver/src/cy/assertions.ts
```diff
@@ -276,10 +276,10 @@ export const create = (Cypress: ICypress, cy: $Cy) => {
...
           }
           case 'subject':
-            return cy.ensureExistence(subject)
+            return Cypress.ensure.exists(subject, cy)
```
emilyrohrbough: &  looks like this helper is missing from the types as well

packages/driver/src/cypress/ensure.ts
```diff
@@ -0,0 +1,314 @@
...
+
+      $errUtils.mergeErrProps(err, errProps)
+    }
+
+    throw err
+  }
+}
+
+const isChildCommand = (command, args) => {
+  if (cy.subjectChain(command.get('chainerId')) === undefined) {
```
emilyrohrbough, line 96: The other helpers that leverage `cy` have them passes as an argument. Should we here?

packages/driver/src/cypress/ensure.ts
```diff
@@ -0,0 +1,314 @@
...
+      args: {
+        cmd: command.get('name'),
+        args: _.isString(args[0]) ? `\"${stringifiedArg}\"` : stringifiedArg,
+      },
+    })
+  }
+}
+
+const isNotDisabled = (subject, name: string, onFail) => {
+  if (subject.prop('disabled')) {
```
emilyrohrbough, line 109: with exposing these publicly, any risk someone will pass a non-jquery element to `isNotDisabled`?

packages/driver/src/cy/assertions.ts
emilyrohrbough: This is only used in this file. Is it worth shipping as a helper?

cli/types/cypress.d.ts
```diff
@@ -130,6 +130,53 @@ declare namespace Cypress {
     unsupportedVersion?: boolean
   }
 
+  interface Ensure {
```
emilyrohrbough, line 134: super-nit: could we ship these types in the same order we export?

Missing types for 
- isStrictlyVisible
- exists
- elExists
- isScrollable
- commandCanCommunicateWithAUT

Unless any were meant for internal use

packages/driver/src/cypress/ensure.ts
```diff
@@ -0,0 +1,314 @@
...
+  isVisible,
+  isStrictlyVisible,
+  isNotHiddenByAncestors,
+  exists,
+  elExists,
+  isScrollable,
+  commandCanCommunicateWithAUT,
+
+  // internal functions
+  isType,
```
emilyrohrbough: `isType` is listed as internal, but has public typescript types. do types need removed or can we updat?

packages/driver/src/cy/actionability.ts
```diff
@@ -53,9 +54,127 @@ const getPositionFromArguments = function (positionOrX, y, options) {
...
+    })
+  }
+}
+
+const ensureElIsNotAnimating = ($el, coords = [], threshold, name: string) => {
+  const lastTwo = coords.slice(-2)
+
+  // bail if we dont yet have two points
+  if (lastTwo.length !== 2) {
+    $errUtils.throwErrByPath('dom.animation_check_failed')
```
emilyrohrbough, line 101: nit: future PR, not here -- it might be nice to update these error keys some day to reflect where they actually live.

packages/driver/src/cy/commands/querying/querying.ts
```diff
@@ -375,7 +375,7 @@ export default (Commands, Cypress, cy, state) => {
     })
 
     return (subject) => {
-      cy.ensureSubjectByType(subject, 'element', this)
+      Cypress.ensure.isType(subject, 'element', 'shadow', cy)
```
emilyrohrbough, line 378: honestly love how these read these updates! Nice job.

packages/driver/src/cy/commands/misc.ts
emilyrohrbough: I wasn't even looking for this change here. 👍🏻

cli/types/cypress.d.ts
BlueWinds: Good call, reordered the exports and types to match. While doing so, realized that `isScrollable` has no reason not to be public, and also that it should accept an `onFail` like everything else in its group.

Order is like so:

```
  isType,
  isElement,
  isDocument,
  isWindow,

  isAttached,
  isNotDisabled,
  isNotHiddenByAncestors,
  isNotReadonly,
  isScrollable,
  isVisible,

  // internal functions
  commandCanCommunicateWithAUT,
  exists,
  elExists,
  isChildCommand,
  isStrictlyVisible,
```

packages/driver/src/cy/actionability.ts
BlueWinds: I think an audit of our error messages would be good in general - there are some that are duplicate and can be merged, and I suspect there are also a few that aren't used at all. Something to keep on the backburner. :+1:

packages/driver/src/cy/assertions.ts
BlueWinds: Going to move these two functions into this file. :+1:

packages/driver/src/cy/assertions.ts
BlueWinds: I left `exists` and `elExists` internal because they aren't straightforwardly running checks like some of the other ones - they invoke a lot more code than they look like from their names.

Now that you mention it though, I'm going to move them directly here into this file.

packages/driver/src/cypress/ensure.ts
BlueWinds: Yeah, definitely. Updated.

packages/driver/src/cypress/ensure.ts
BlueWinds: Hm. Interesting thing to think about. The intended usage would be something like

```
Cypress.ensure.isElement(subject, 'myQuery')
Cypress.ensure.isNotDisabled(subject, 'myQuery')
```

If they don't ensure that it is an element first, and it's not an element, as it stands it would probably throw a pretty non-helpful error like "can't read property `prop` of null" or something.

If we added an implicit `isElement()` check to the top of each of these functions, that would provide more helpful error messages - but then we'd be running duplicate code for pretty much every cypress command.

I'm inclined to leave it as-is, but open to other thoughts.

packages/driver/src/cypress/ensure.ts
BlueWinds: Redid / reordered this list, both to specify which ones are internal and to match the order of type definitions. :+1:

cli/types/cypress.d.ts
BlueWinds: Changed the type to `Log`. While it technically accepts a function as well, we don't use it that way internally, and I'd like to leave that aside in this PR, until we decide to more strictly type the internals that this is pointing towards / reconsider how we handle logs and error messages.

packages/driver/src/cypress/ensure.ts
emilyrohrbough: maybe we be a bit more guarded in our check? if the original conditions are not met, ignore it?
```
if (subject.prop && _.isFunction(subject.prop) && subject.prop('disabled') {
```

emilyrohrbough: though ignoring it sort of assumes it's true when it might be invalid? but oh well

cli/types/cypress.d.ts
```diff
@@ -131,6 +131,58 @@ declare namespace Cypress {
...
+     */
+    isAttached(subject: any, commandName: string, cy: Chainable, onFail?: Log): void
+
+    /**
+     * Throws an error if `subject` is a disabled DOM element.
+     */
+    isNotDisabled(subject: any, commandName: string, onFail?: Log): void
+
+    /**
+     * Throws an error if `subject` is hidden by any of its parent elements.
```
```suggestion
     * Throws an error if `subject` is a DOM element hidden by any of its parent elements.
```

chrisbreiding (approved):
In regards to documentation, will `Cypress.ensure` get its own page like [`Cypress.dom`](https://docs.cypress.io/api/cypress-api/dom) or just the mentions on the custom-queries page?

BlueWinds:
> In regards to documentation, will `Cypress.ensure` get its own page like [`Cypress.dom`](https://docs.cypress.io/api/cypress-api/dom) or just the mentions on the custom-queries page?

I was planning to leave it with just the mentions on custom-queries for now. Adding a page documenting it more explicitly can be a post-release task.

AtofStryker (approved):

PR #24699, opened 2022-11-15
chore: re-name dashboard references to Cypress Cloud

### LINKED ISSUE #24386: Rebrand dashboard to cloud
### What would you like?

Change all references in the app from dashboard to cloud.

@mattvCypress has started a branch here: https://github.com/mattvCypress/cypress/tree/change-dashboard-to-cloud

We may want to consider renaming all internal variables as well.

Here are the standards: https://docs.google.com/presentation/d/1DP2cAv0Kb9JyVD3odu4jz0vMWFzuhUEpG_AYhyBZuWk/edit#slide=id.p

### Why is this needed?

Update to corporate branding standards

### Other

_No response_

astone123:
- Closes https://github.com/cypress-io/cypress/issues/24386

### User facing changelog

n/a

### Additional details

We are now referring to dashboard as Cypress Cloud. This PR updates all of the wording around the dashboard to use "Cypress Cloud" instead. I also updated some internal variable names to say "cloud" rather than "dashboard".

### Steps to test

The most important thing to test is that all Cypress Cloud actions still work with these changes. Verify the following:
- You can still log into Cypress Cloud
- The "connect project" modal flow still works
- Runs tab still links to the correct runs
- Latest runs, flaky badges, and average duration still appear on the specs list
- Other cloud integration things that I'm forgetting

Also verify that all visible copy refers to Cypress Cloud (not dashboard) and follows the [brand standards](https://docs.google.com/presentation/d/1DP2cAv0Kb9JyVD3odu4jz0vMWFzuhUEpG_AYhyBZuWk/edit#slide=id.p)

### PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4719


packages/frontend-shared/src/locales/en-US.json
```diff
@@ -322,9 +322,9 @@
           "text": "Configuration guide",
           "link": "https://docs.cypress.io/guides/references/configuration"
         },
-        "dashboardGuide": {
-          "text": "Cypress Dashboard guide",
-          "link": "https://docs.cypress.io/guides/dashboard/introduction"
+        "cloudGuide": {
+          "text": "Cypress Cloud guide",
+          "link": "https://on.cypress.io/features-dashboard"
```
astone123, line 327: I changed this to use an on link because the dashboard link will be invalid once the docs PR is merged

marktnoonan: Good call, I noticed the other day that all links from the app are supposed to be onlinks, including links to docs, so we probably have a little audit on our hands: https://github.com/cypress-io/cypress/blob/develop/CONTRIBUTING.md#adding-links-within-code

packages/app/src/runs/RunsError.spec.tsx
```diff
@@ -9,12 +9,12 @@ describe('<RunsError />', () => {
         return (
```
lmiller1990: Nit and unrelated, but can we rename this `.cy.tsx` instead? I'm not sure why it's using `.spec`, we usually reserve this for tests that run in Node.js.

packages/data-context/src/actions/EventCollectorActions.ts
```diff
@@ -25,10 +25,10 @@ export class EventCollectorActions {
 
   async recordEvent (event: CollectableEvent): Promise<boolean> {
     try {
-      const dashboardUrl = this.ctx.cloud.getDashboardUrl(cloudEnv)
+      const cloudUrl = this.ctx.cloud.getCloudUrl(cloudEnv)
```
lmiller1990, line 28: Nice, glad we are changing the vars, not just the strings, it'll be much clearer.

lmiller1990 (approved):
Looks good - I left one nit, but not really a big deal.

If you did just want to rename without re-running CI, you can commit with a commit message containing `skip ci` and it won't run. https://circleci.com/docs/skip-build/#skipping-a-build

packages/app/src/runs/RunsError.spec.tsx
lmiller1990: I think there's a couple more that do this in `app`, but we could make this change in another PR - I just noticed it here, since this file had modifications.

astone123: Looks like someone just did this on develop 👍🏻

amehta265 (approved):
Everything looks good locally and percy looks good to!

Issue #24441, opened 2022-10-28
[IATR](M0) Debug Page and Tab in left nav


warrensplayer:
Add a new navigation item in the app left nav for "Debug" which will navigate to a new page

**NOTE:** M0 issues will be committed against a feature branch and not released 

### UI Design

* Left nav will have new item below Runs and above Settings for "Debug" that will have a corresponding icon
* The page component will render the [header component](https://github.com/cypress-io/cypress/issues/24442) followed by a list of [failed specs components](https://github.com/cypress-io/cypress/issues/24443)
* Page should have a blank empty state (just text that says "Not logged in") when not logged in
* Page should have a blank empty state (just text that says "Not connected") when logged in but no project connected
* Page should have a blank empty state (just text that says "No runs") when there are no matching runs

<img width="1018" alt="image" src="https://user-images.githubusercontent.com/1702361/198724068-ac8ce7e1-361d-4d2e-afe4-505dfa95d569.png">

_Wireframe from Whimsical. Final design is in progress._

### Inputs

Pull relevant run information from the `login-connect-store` and then fetch failed spec information for that run. See https://github.com/cypress-io/cypress/issues/24440

warrensplayer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/24441?planning-poker) with Zenhub @rockindahizzy @lmiller1990 @ZachJW34 @marktnoonan @amehta265 @astone123

ZachJW34:
This should be internal right? Should the issue be updated to include "Only show new tab in local dev"?

warrensplayer:
> This should be internal right? Should the issue be updated to include "Only show new tab in local dev"?
@ZachJW34 I am adding a note to the IATR M0 issues to indicate that these will be held in a feature branch and not released

Issue #24575, opened 2022-11-08
Flaky test: "App: should re-query for executing runs" (windows mainly)


lmiller1990:
### Link to dashboard or CircleCI failure

https://app.circleci.com/pipelines/github/cypress-io/cypress/45562/workflows/8510305c-0f92-41d6-b8fb-2abaa50db11a/jobs/1914521/tests#failed-test-2

### Link to failing test in GitHub

https://github.com/cypress-io/cypress/blob/963e184446dffe627c7c412c72c428be0da388b0/packages/app/cypress/e2e/runs.cy.ts#L888-L955

These two are flaky on windows.

Edit: First one is fixed  https://github.com/cypress-io/cypress/pull/24833

### Analysis

Not sure if it's race condition, I've seen it flake on linux too, but much less.

These tests override `window.setTimeout` and stub out multiple GraphQL requests - it's quite confusing since there's so much stubbing going on, it's not entirely clear if the flake is in the app code or the test code.

I think we should consider an alternative way to orchestrate these tests that relies less on stubbing, and is more deterministic.

### Cypress Version

10.11

### Other

It's particularly bad on windows, I can reproduce it locally about 90% of the time.

lmiller1990:
Works great in linux! :linux master race:

![image](https://user-images.githubusercontent.com/19196536/200463947-3bc4b91d-a256-4207-895f-aad2b5595211.png)

lmiller1990:
On windows, the stubbed GraphQL response is missing the expected properties. `commitInfo`, `totallDuration` etc are `null`. This is not the case on linux and macOS. I do not know why - I went down to the underlying `cross-fetch` and `node-fetch`, everything is fine and present.

At this point, the data is present - this is the last part before the stubbed response is returned. https://github.com/cypress-io/cypress/blob/e3de5e77b51ee215e4bb9aa5cb161daccf874550/packages/frontend-shared/cypress/e2e/e2ePluginSetup.ts#L242

On the other end, in the browser, it's missing properties. I cannot see any other layer in between the above line and the browser where I can debug this, though.

## MacOS - has all properties, runs are rendered

<img width="1296" alt="image" src="https://user-images.githubusercontent.com/19196536/204957936-c9abde9d-6967-41a3-8e7d-1a8603c8a4bd.png">

## Windows - missing properties, not rendered

![image](https://user-images.githubusercontent.com/19196536/204957442-35a29db8-adf7-4ef5-a36d-1628bed67790.png)

[...]

PR #24722, opened 2022-11-17
feat: IATR-M0 Page Header

### LINKED ISSUE #24442: [IATR](M0) Failures page header component
Create a component to be the header of the new [failures page](https://github.com/cypress-io/cypress/issues/24441) that represents the chosen run and its metadata.

**NOTE:** M0 issues will be committed against a feature branch and not released 

### UI Design
* Will mimic the same header design as used on the Runs page in the app and a Run in the Cloud
* Element that will appear in header
  * Commit message
  * Run number
  * Run Status
  * Git branch name
  * Commit hash (shortened)
  * Committer username
  * Relative time to run
  * Run duration
  * "View in dashboard" link  that will link out to the run in the cloud via the CloudRun.url 

image

_Wireframe from Whimsical. Final design is in progress._

### Inputs
* CloudRun - Should be the same CloudRun type used on the Runs page
  * existing metadata that will be used in the header design
*  FailedSpecs[]: Spec

#### Resources
* [Clickup](https://app.clickup.com/t/18033298/PM-2546)
* [Whimsical](https://whimsical.com/q3-2022-solutions-CB1g7VnZeHsrVdz4gurF2o@AhRp651J2s3PQev2xg2uwbceF) - See V0 section

amehta265:
- Closes #24442 

### User facing changelog

This component is the header of the new [debug page](https://github.com/cypress-io/cypress/issues/24441) that represents the chosen run and its metadata.
The design for this page can be found [here](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/QkCp2F69X0qyLo7qnYQ67e/Cypress-App?node-id=21413%3A202471&t=Enj8Ibq2XyuAaWju-0)

### Additional details

- The `DebugPageHeader` contains metadata related to a user's test. In order to render the status of the test (i.e. number of passed, failed, skipped, pending tests) I have also created another component `DebugResults`.  `DebugResults` utilizes an already existing component `ResultCounts` to render the UI for this section. Changes have been `ResultCounts.vue` to accommodate for a change in the `order` of statuses shown as per design requirements.
- `debugPageHeader` accepts `gql` values as a prop. In case these props are not passed to the component it will render default values.
- Certain functions such as `commitsAhead()` are merely placeholders and will be implemented later on. Furthermore the `gql` query will potentially change in the future

### Steps to test

This issue has 3 component tests. Checkout to this branch to run the tests
Both `DebugPageHeader.vue` and `DebugResults.vue` exist in the `packages/app/` folder. To run these tests:
1. Change directory to `packages/app`
2. Run `yarn dev`
3. Navigate to Component Testing and run `DebugPageHeader.cy.tsx` and `DebugResults.cy.tsx`

To test `ResultsCounts.vue`
1. Go to the cypress root directory
2. Run `yarn cypress:open`
3. Add `frontend-shared` as a project
4. Open frontend-shared in Component Testing and run `ResultCounts.cy.tsx`

Screenshot 2022-11-17 at 12 07 40 PM

### PR Tasks


packages/frontend-shared/src/components/ResultCounts.cy.tsx
```diff
@@ -49,4 +49,24 @@ describe('<ResultCounts />', () => {
 
     cy.percySnapshot()
   })
+
```
warrensplayer, line 52: Add a test that uses the slot

packages/frontend-shared/src/components/ResultCounts.vue
```diff
@@ -36,33 +41,50 @@ const { t } = useI18n()
...
+  },
+}
+
+interface status {
+  value: number|string|null
+  class: string
+  icon: FunctionalComponent<SVGAttributes, {}>
+  name: string
+}
+
```
warrensplayer: You are off to a good start, but the types can be improved on a bit.

Add a union type for the count types:
`type CountType= 'SKIPPED' | 'PENDING' | 'PASSED' | 'FAILED'`

Then you have the type for `status`, but the convention would be to make the first letter capitalized.  i.e. `Status`

Use the union type for the `order` prop and use it to type the ORDER_MAP
`const ORDER_MAP: Record<CountType, Status>`

packages/frontend-shared/src/components/ResultCounts.vue
```diff
@@ -36,33 +41,50 @@ const { t } = useI18n()
...
-      name: t('runs.results.failed'),
-    },
-  ]
+  const acc: status[] = []
+
+  if (props.order) {
+    props.order.forEach((status) => acc.push(ORDER_MAP[status]))
+  } else {
+    acc.push(ORDER_MAP['SKIPPED'], ORDER_MAP['PENDING'], ORDER_MAP['PASSED'], ORDER_MAP['FAILED'])
+  }
```
warrensplayer: I always try to think if there are ways to simplify conditionals in a functional way.  I also look at the different parts the code is trying to accomplish.  Here, we need to figure out what order to use first, then use that order to create an array.  

```suggestion
  const order = props.order || ['SKIPPED', 'PENDING', 'PASSED', 'FAILED']

  return order.map((status) => ORDER_MAP[status])
```

warrensplayer: Also, the Vue `computed` method creates a reactive `ref` that would be updated if any reactive elements inside the `computed` changes.  In order for this to see any changes to the `prop` values for the counts, that map creation needs to move back inside the `computed` method.  Happy to explain more if it would help.

packages/app/src/debug/DebugResults.vue
```diff
@@ -0,0 +1,67 @@
...
+  totalSkipped
+  totalFlakyTests
+}
+`
+
+const props = defineProps<{
+  gql: DebugResultsFragment
+}>()
+
+const defaultResults = {
```
warrensplayer: Pull the `defaultResults` out of the component and just test values with the `gql` prop.

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -0,0 +1,196 @@
...
+}
+
+const defaultValues: DefaultTypes = {
+  id: '3',
+  createdAt: '2022-11-14T14:41:23.078Z',
+  totalDuration: '10m 30s',
+  url: 'cloud.cypress.io',
+  authorEmail: 'hello@cypress.io',
+  commitInfo: defaultscommitInfo,
+}
```
warrensplayer: Remove the default test data and just test with passing in data from the tests

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -0,0 +1,196 @@
+<template>
+  <div
+    data-cy="debug"
+    class="flex flex-col pb-24px gap-16px"
+  >
+    <div
+      :data-cy="`debug-header-${debug.id}`"
+      class="grid px-24px w-full overflow-hidden flex items-center border gap-y-2 py-24px"
```
warrensplayer: Remove `border`. The design does not have a border.

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -0,0 +1,196 @@
...
+          class="font-medium text-lg text-gray-900"
+          data-cy="debug-test-summary"
+        >
+          {{ debug.commitInfo.summary }}
+        </li>
+        <div
+          v-if="defaultscommitInfo.runNumber && commitsAhead"
+          class="border rounded border-gray-100 items-center text-sm h-6"
+          data-cy="debug-runCommit-info"
+        >
```
warrensplayer: `items-center` needs a `flex-row` to get the items to center here.

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -0,0 +1,196 @@
...
+    return props.gql
+  }
+
+  return defaultValues
+})
+
+const relativeCreatedAt = computed(() => dayjs(new Date(debug.value.createdAt!)).fromNow())
+
+// We need a function to calculate the commits ahead property
+const commitsAhead = computed(() => 'You are 2 commits ahead')
```
warrensplayer: Pass this in as a separate prop.  The function to calculate will be external to this component

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -0,0 +1,196 @@
+<template>
+  <div
+    data-cy="debug"
+    class="flex flex-col pb-24px gap-16px"
+  >
+    <div
+      :data-cy="`debug-header-${debug.id}`"
```
warrensplayer: The variable number on the end of this `data-cy` was used on the Runs page because there were multiple Runs being displayed.  That is not needed here.

warrensplayer (changes requested):
Still looking through details, but wanted to get this feedback to you.

packages/frontend-shared/src/components/ResultCounts.vue
```diff
@@ -36,33 +41,50 @@ const { t } = useI18n()
...
-    {
-      value: props.totalFailed,
-      class: 'icon-dark-red-400',
-      icon: FailedIcon,
-      name: t('runs.results.failed'),
-    },
-  ]
+  const acc: status[] = []
+
+  if (props.order) {
```
ZachJW34: Should the order even be configurable? The results this component displays should be consistent across features, otherwise it becomes confusing.

packages/app/src/debug/DebugResults.vue
```diff
@@ -0,0 +1,67 @@
...
+    data-cy="debug-results-holder"
+  >
+    <ResultCounts
+      :total-passed="results.totalPassed"
+      :total-failed="results.totalFailed"
+      :total-skipped="results.totalSkipped"
+      :total-pending="results.totalPending"
+      :order="['PASSED', 'FAILED', 'SKIPPED', 'PENDING']"
+    >
+      <template #prefix>
```
ZachJW34, line 13: For information to show up on this page, does it have to fail? Is the `Failed` text always going to be visible?

packages/app/src/debug/DebugResults.vue
```diff
@@ -0,0 +1,67 @@
...
+        <div
+          class="text-red-400 font-semibold text-sm flex items-center border-r-1px px-2"
```
ZachJW34: This is not a full height border in figma. This is also visible on the first line
Screen Shot 2022-11-17 at 4 47 26 PM

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -0,0 +1,196 @@
...
+          <ExternalLink
+            :data-cy="'debug-header-dashboard-link'"
+            :href="debug.url || '#'"
+            :use-default-hocus="false"
+          >
+            <span class="sr-only">Dashboard Link:</span> View in the dashboard
+          </ExternalLink>
+        </li>
+      </ul>
+      <ul
```
ZachJW34, line 53: This dot separated list is the same in the runs page. Can we pull this into a reusable component?

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -0,0 +1,196 @@
...
+  <div
+    data-cy="debug"
+    class="flex flex-col pb-24px gap-16px"
+  >
+    <div
+      :data-cy="`debug-header-${debug.id}`"
+      class="grid px-24px w-full overflow-hidden flex items-center border gap-y-2 py-24px"
+    >
+      <ul
+        :data-cy="'header-top'"
```
ZachJW34:
```suggestion
        data-cy="header-top"
```

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -0,0 +1,196 @@
...
+          >
+            {{ commitsAhead }}
+          </span>
+        </div>
+        <li class="-mt-8px text-lg text-gray-400">
+          .
+        </li>
+        <li class="font-normal text-sm text-indigo-500">
+          <ExternalLink
+            :data-cy="'debug-header-dashboard-link'"
```
ZachJW34:
```suggestion
            data-cy="debug-header-dashboard-link"
```

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -0,0 +1,196 @@
...
+        >
+          <DebugResults
+            :gql="props.gql"
+          />
+        </li>
+        <li
+          v-if="debug?.commitInfo?.branch"
+          data-cy="debug-header-branch"
+        >
+          <i-cy-tech-branch-h_x16 class="mr-1 icon-dark-gray-300 mr-9px" />
```
ZachJW34: margins/paddings are normally factors of 4 (normally 8), so seeing `mr-9px` makes me think it should be `mr-8px`

packages/frontend-shared/src/components/ResultCounts.vue
amehta265: This is an attempt to make the `ResultCounts.vue` component more generic. In its current implementation, the status can only appear in a certain order i.e. skipped, pending, passed, failed. You can see this generic component in the Runs tab for the app.
Based on the design requirements for this issue, the order of these "statuses" is different. Hence, instead of building a whole new component we can utilize `ResultCounts.vue` and just pass in a different order incase in the future the order of "statuses" is different.

As for the component being consistent, I agree with that. I don't see why the "status" order should differ across pages. However this has been added to meet the design requirements

packages/app/src/debug/DebugResults.vue
amehta265: This is a good point! I just looked at the future milestones and looks like this text could be `Failed`, `Passed`, `Running`, e.t.c. So I will definitely change this to use a variable instead of a fixed value

packages/app/src/debug/DebugPageHeader.cy.tsx
```diff
@@ -0,0 +1,54 @@
+import { DebugPageFragmentDoc } from '../generated/graphql-test'
+import DebugPageHeader from './DebugPageHeader.vue'
+
+const generateTags = (num): any => new Array(num).fill(null).map((_, i) => ({ id: `${i}`, name: `tag${i}`, __typename: 'CloudRunTag' }))
```
lmiller1990:
```suggestion
const generateTags = (num) => new Array(num).fill(null).map((_, i) => ({ id: `${i}`, name: `tag${i}`, __typename: 'CloudRunTag' }))
```

If you remove `any` TS can infer the usage, if you hold you mouse over `generateTags` it should infer the return type as an object w/ the defined properties.

packages/app/src/debug/DebugPageHeader.vue
lmiller1990: Just FYI you will need to use `mountFragment` do this, something like

```js
cy.mountFragment(DebugPageFragment_Doc, {
  onResult: (res) => {
    res.totalDuration = ...
    // etc
  },
  render: gql => <DebugPageHeader gql={gql} />
})
```

The way it works is `res` is mutated, and then you get the new value as the first arg to `render`, and pass it to your component. So you can easily test lots of different prop variations.

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -0,0 +1,196 @@
...
+const debug = computed(() => {
+  if (props.gql) {
+    return props.gql
+  }
+
+  return defaultValues
+})
+
+const relativeCreatedAt = computed(() => dayjs(new Date(debug.value.createdAt!)).fromNow())
+
```
lmiller1990, line 148: Figuring out how to calculate the N commits ahead is going to be fun one (no need to do it here).

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -0,0 +1,196 @@
...
+// We need a function to calculate the commits ahead property
+const commitsAhead = computed(() => 'You are 2 commits ahead')
+
+</script>
+<style scoped>
+ul:nth-child(2) li:not(:first-child)::before {
```
lmiller1990: Woah, this is specific. I think this CSS could break easily.

I'm not sure what it does, but can we use a class or id or something that isn't tied to a specific DOM structure?

packages/app/src/debug/DebugResults.cy.tsx
```diff
@@ -0,0 +1,33 @@
...
+    cy.mount(() => (<DebugResults/>))
+    cy.get('[data-cy=debug-results-holder]').then((items) => {
+      cy.get(`[title=${defaultMessages.runs.results.passed}]`).should('contain.text', 1)
+      cy.get(`[title=${defaultMessages.runs.results.failed}]`).should('contain.text', 7)
+      cy.get(`[title=${defaultMessages.runs.results.skipped}]`).should('contain.text', 6)
+      cy.get(`[title=${defaultMessages.runs.results.pending}]`).should('contain.text', 6)
+    })
+  })
+
+  it('shows the failed icon and the number of passed, skipped, pending, failed tests passed through gql props', () => {
```
lmiller1990, line 6: Definitely recommend writing with `mountFragment`, just do a search in the code base, there are many examples.

packages/app/src/debug/DebugResults.vue
```diff
@@ -0,0 +1,67 @@
...
+        >
+          <component
```
lmiller1990, line 17: Why not just

```vue
  <FailedSolidIcon
    class=...
    data-cy=...
/>
```

`<component>` is good for dynamic content - but this isn't. (Could be missing something).

packages/app/src/debug/DebugResults.vue
```diff
@@ -0,0 +1,67 @@
...
+    </ResultCounts>
+  </div>
+</template>
+
+<script lang="ts" setup>
+import { computed } from 'vue'
+import type { DebugResultsFragment } from '../generated/graphql'
+import { gql } from '@urql/core'
+import FailedSolidIcon from '~icons/cy/status-failed-solid_x24.svg'
+import ResultCounts from '@packages/frontend-shared/src/components/ResultCounts.vue'
```
lmiller1990, line 36: Curious what the team things but this seems like it's not really a good candidate for `frontend-shared`. It doesn't seem like something we will use in other packages (like launchpad).

Thoughts? cc @marktnoonan, we had a brief discussion around this.

lmiller1990 (commented):
Sure looks pretty, left some comments

packages/app/src/debug/DebugResults.vue
amehta265: The "status" text will always be visible in the header component according to the designs

packages/frontend-shared/src/components/ResultCounts.vue
ZachJW34: I would sync back with design regarding this, this result of this list (in my mind) should be consistent regardless of if you're looking at the runs page or the debug page

packages/app/src/debug/DebugPageHeader.vue
amehta265: After more discussion with @warrensplayer  I don't think we want a reusable component here. Despite the similarity between the runs page and this component, there are design parts that are completely distinct. For example: the number of components that show on the separated list and the order in which they show up. Furthermore,  on the runs page, the dot separated list is part of the `listRowHeader` component which also accepts a "Header" section which is not a part of this component. So leaving this as a standalone component would suffice

packages/app/src/debug/DebugPageHeader.cy.tsx
amehta265: I just realized that we do not need to generate tags for this component.

packages/app/src/debug/DebugPageHeader.vue
amehta265: This adds the " . "  after each component in the bottom section of the UI. That way we don't manually have to add  " . " in our template. It is similar to [styling here](https://github.com/cypress-io/cypress/blob/b9d053e46777683a9c502a5e970ef3ae26c563d5/packages/app/src/runs/RunCard.vue#L133)

packages/app/src/debug/DebugResults.vue
amehta265: This will not always be a `FailedSolidIcon`. In future iterations, this could be a `PassedIcon` or hold other statuses hence the `<component>`

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -0,0 +1,155 @@
...
+  runNumber: number
+}>()
+
+const debug = computed(() => props.gql)
+
+const relativeCreatedAt = computed(() => dayjs(new Date(debug.value.createdAt!)).fromNow())
+
+</script>
+<style scoped>
+ul:nth-child(2) li:not(:first-child)::before {
```
warrensplayer: @amehta265 What Lachlan is referring to here is the `ul:nth-child(2)` reference.  This is specifically targeting the second `ul` in the component, but it would break if someone added another `ul` ahead of it.  You could add a class to that `ul` and target that with your CSS selector instead of `ul-nth-child(2)`

packages/app/src/debug/DebugResults.vue
warrensplayer: @lmiller1990 The `ResultsCounts` is an existing component that was just getting reused here with a minor addition.  I do agree that it is only used in `app` so does not need to be in `frontend-shared`, but we can refactor that another time.

warrensplayer (commented):
Just one minor nit-pick

packages/app/src/debug/DebugPageHeader.vue
amehta265: I see, that makes a lot of sense. I will make the necessary updates

warrensplayer (approved):

packages/app/src/debug/DebugPageHeader.cy.tsx
```diff
@@ -0,0 +1,54 @@
...
+          if (result.commitInfo) {
+            result.commitInfo.summary = 'Adding a hover state to the button component'
+            result.commitInfo.branch = 'feature/DESIGN-183'
+            result.commitInfo.authorName = 'cypressDTest'
+          }
+        }
+      },
+      render: (gqlVal) => {
+        return (
+          <DebugPageHeader gql={gqlVal} runNumber={468} commitsAhead={'You are 2 commits ahead'} commitHash={'b5e6fde'}/>
```
lmiller1990:
```suggestion
          <DebugPageHeader gql={gqlVal} runNumber={468} commitsAhead='You are 2 commits ahead' commitHash='b5e6fde' />
```

These are just strings so you do not need to use `{}` to interoperate (that's just for passing JS). But this isn't a big deal.

packages/app/src/debug/DebugPageHeader.cy.tsx
```diff
@@ -0,0 +1,54 @@
...
+    cy.get('[data-cy=debug-runCommit-info]').children().should('have.length', 3)
+    cy.get('[data-cy=debug-runNumber]')
+    .should('have.text', ' Run #468')
+    .should('have.css', 'color', 'rgb(90, 95, 122)')
+
+    cy.get('[data-cy=debug-commitsAhead]')
+    .should('have.text', 'You are 2 commits ahead')
+    .should('have.css', 'color', 'rgb(189, 88, 0)')
+
+    cy.get('[data-cy=debug-results]').should('be.visible')
```
lmiller1990:
```suggestion
    cy.findByTestId('debug-results').should('be.visible')
```

WDYT? We have a custom helper for these.

packages/app/src/debug/DebugResults.vue
lmiller1990: Sounds good to me

lmiller1990 (approved):
Looks good, left one last comment but not a blocker.

lmiller1990:
Updating the feature branch and merging it in here, should fix CI. I'll merge once this is green.

lmiller1990:
Looks like we have a CI failure - re-ran but still failing, want to take a look @amehta265? https://app.circleci.com/pipelines/github/cypress-io/cypress/46041/workflows/a86994bb-0f07-454a-bfaf-60c6048476a0/jobs/1936007/tests#failed-test-0

lmiller1990:
Actually @amehta265 it might be flake, see https://github.com/cypress-io/cypress/issues/24575
If you want to verify it passes locally, I'd recommending pushing a commit to skip this and reference https://github.com/cypress-io/cypress/issues/24575 so we can fix it next sprint.

PR #24738, opened 2022-11-19
feat: Run-all-specs-ui

### LINKED ISSUE #24468: [Run All Specs] - Inline Spec Explorer - Add "run N specs" button to folder rows

amehta265:
- Closes #24468 and #24467: This closes all UI related to run-all-specs

### User facing changelog

Have introduced an run all specs under the experimental flag for e2e testing. This feature allows a user to run all specs under any directory or sub-directory.

When the experimental flag is enabled: hovering on a directory within the specs list or the inline spec list will trigger a action play button and text describing the number of specs within the hovered directory. A user would be able to click on the button to then run all specs within the  chosen directory. This applies for filtered specs as well.

### Additional details

- Have added a `RunAllSpecs.vue` component that is utilized in  `SpecsList.vue` and `InlineSpecsList.vue`. The component will accept a `specNumber` as a prop which represents the number of files within a directory.
- The most important is the `directoryChildren`. This is a reactive variable that is updated every time the "spec list tree" is re-built. It is a mapping `<{[key: string]: string[]}>`where the key is the relative path of a directory and the value is a list of specs within the directory. We use `directoryChildren[directoryId].length` to obtain the number of specs within a specific directory.

### Steps to test

There are two ways of testing the features.
A: Component Tests: `RunAllSpecs.cy.tsx`, `SpecsList.cy.tsx`, `InlineSpecsList.cy.tsx`
1. `cd packages/app`
2. Run `yarn cypress:open`
4. Choose component testing and test each of the files listed above. The contain run all specs related testing

B: Testing a project
1. Open two terminals
2. Run `yarn watch` in one terminal and wait for the app to be built.
3. Run `yarn cypress:open` in the second terminal after the completion of step 2
4.Select a project that contains `e2e` tests for example `cypress-example-kitchensink`. Go the project's `cypress.config` file and under the `e2e:{}` object add: `experimentalRunAllSpecs: true`
5. In the cypress app, select the project for which you have enabled run all specs
6. You should see the `Run N Specs` while hovering on directories in the specsList or see a play button when hovering over directories in the Inline specs list. When pressing these buttons you should be able to RUN ALL SPECS!

Hovering over directory row:
Screenshot 2022-11-18 at 8 35 40 PM
Hovering over button:
Screenshot 2022-11-18 at 8 36 15 PM
Hovering on Inline specs list directory row:
Screenshot 2022-11-18 at 8 30 08 PM
Hovering over button on inline specs list:
Screenshot 2022-11-18 at 8 31 37 PM

### PR Tasks


packages/app/src/specs/SpecsList.vue
```diff
@@ -448,4 +464,9 @@ function onRunAllSpecs (directory: string) {
 .spec-list-container {
   height: calc(100% - 112px)
 }
+
+/** For run all specs group hover to work */
```
lmiller1990: Can use `group` and `group-hover`. https://tailwindcss.com/docs/hover-focus-and-other-states#group-hover

I will push the change and link.

packages/app/src/specs/SpecsListRowItem.vue
```diff
@@ -20,12 +20,14 @@
       >
         <div
           data-cy="specs-list-row-git-info"
+          class="group"
```
lmiller1990, line 23: These do not seem to do anything? `group` is used with `group-*` (like `group-hover`) which I don't see here.

Can we remove these?

packages/app/src/specs/DirectoryItem.vue
```diff
@@ -14,6 +14,7 @@
       class="text-gray-400 group-focus:text-indigo-300"
       highlight-classes="font-bold text-white"
     />
+    <slot name="run-all-specs" />
```
lmiller1990, line 17: I made this more clear by adding a name to the slot. In general, I think we should avoid default slots, unless it's really obvious what it is.

https://github.com/cypress-io/cypress/pull/24738/commits/eb8bc3d6fc3dcd0c0d135eed2946f2f657a0966c

packages/app/src/specs/InlineSpecListTree.vue
```diff
@@ -156,6 +170,28 @@ const resetFocusIfNecessary = (row, index) => {
   }
 }
 
+const directoryChildren = computed(() => {
+  return collapsible.value.tree.reduce<Record<string, string[]>>((acc, node) => {
```
lmiller1990, line 174: Changed to use built-in `Record` type: https://github.com/cypress-io/cypress/pull/24738/commits/6618f603a1f184b2f6663d88c0282a9416114498

lmiller1990:
This looks pretty good. I gave it a test and pushed some minor tweaks. I think the UI needs some tweaking still, but in the interest of getting this ready for release, I'm going to merge it into the feature branch and get CI green. I'll also build a binary so we can test it out.

I'm posting my UI feedback here: https://github.com/cypress-io/cypress/pull/24745#issuecomment-1321291351 as well as anything else I find.

PR #24745, opened 2022-11-21
feat: Re-introduce Run All specs for End to End under experimentalRunAllSpecs flag

### LINKED ISSUE #24168: Run All Specs for E2E
Adds back the ability to run all specs in cypress open mode for End to End testing.

Edit (@lmiller1990) this will be something we incrementally improve over time. The initial implementation will be similar to how it worked in Cypress 9, and focused on supporting End to End testing. Support for Component Testing, and more advanced workflows, will come as the experiment continues and we get feedback.

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/24168

### User facing changelog

Adds a new experimental flag, `experimentalRunAllSpecs`, for End to End Testing. Enable it in `cypress.config`:

```ts
import { defineConfig } from 'cypress'

export default defineConfig({
  e2e: {
    experimentalRunAllSpecs: true
  }
})
```

### Additional details

This PR adds back in support for running multiple tests sequentially in Open Mde for End to End testing. It can be accessed from both the Specs page and the Runner page in the Cypress App:

image

ss

### Steps to test

1. Either check out this branch locally, or install the pre-release binary. [Those are here](https://github.com/cypress-io/cypress/commit/30163cbfb060a21cb7ccc0e424951bcc6113396b#comments).
2. Enable the experiment:

```ts
import { defineConfig } from 'cypress'

export default defineConfig({
  e2e: {
    experimentalRunAllSpecs: true
  }
})
```

Open Cypress in Open Mode. The Specs page and Runner page should be have "Run N Specs" labels that are revealed when you hover a directory. Clicking the button will re-open the browser and execute all the specs sequentially.

You can also use the text search to only execute a sub-set of specs. Saving any of the specs will re-execute all the specs that matched your criteria (search, directory, or both).

Note: It's not fully a11y yet. You cannot do Run All Specs with a keyboard only yet. I filed an issue for a fast follow: https://github.com/cypress-io/cypress/issues/24748. If there's a trivial fix, that'd be great.

See above.

### PR Tasks

- [x Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4860


lmiller1990:
Doing some testing, collecting my feedback here.

Edit: I've fixed these in https://github.com/cypress-io/cypress/pull/24746. Please review.

<details><summary>Feedback</summary>
## Button position on inline spec list is inconsistent
 
The icons appear in different positions. Should it align right?

https://user-images.githubusercontent.com/19196536/202935346-38ae2970-5d1f-4848-9553-f7b6e84b011e.mp4

Another example:

https://user-images.githubusercontent.com/19196536/202934818-54794e7e-3f02-4812-8f40-97c91476dd33.mp4

I think it should always be a fixed width from the right. That's what the designs hint at.

## Experimental flag types

Current I can set `experimentalRunAllSpecs` at the top level. This throws an error, telling me it's E2E only. How about we make the types reflect this, too?

```js
export default = {
  experimentalRunAllSpecs: truer
}
```

IDE says it's valid, but Cypress throws an error. 
</details>

ZachJW34 (approved):

packages/app/src/specs/InlineRunAllSpecs.vue
```diff
@@ -0,0 +1,44 @@
...
+    <button>
+      <IconActionPlaySmall
+        size="16"
+        stroke-color="gray-700"
+        fill-color="transparent"
+        hocus-stroke-color="indigo-500"
+        hocus-fill-color="indigo-100"
+        class="inline-flex align-text-bottom"
+        data-cy="play-button"
+        @click.stop="emits('runAllSpecs')"
```
marktnoonan: I think this click handler should be on the button, not the icon. When I use the keyboard on the button it doesn't rerun specs for me, it just toggles the folder open and closed.

packages/app/src/specs/InlineRunAllSpecs.vue
```diff
@@ -0,0 +1,44 @@
...
+      />
+    </button>
+    <template
+      #popper
+    >
+      <span
+        class="font-normal text-sm inline-flex"
+        data-cy="tooltip-content"
+      >
+        Run {{ specNumber }} specs
```
marktnoonan: We need to handle pluralization here, we could use the i18n for this.

Screen Shot 2022-11-21 at 4 16 03 PM

packages/app/src/specs/InlineRunAllSpecs.vue
marktnoonan: Hmm, moving the handler didn't help, I think because the parent is using `@click.capture` to manage the toggle. That code would need some extra logic to not toggle if the click event came from the button.

packages/app/src/specs/InlineRunAllSpecs.vue
marktnoonan: Oh it's already in the main Spec Explorer, we can do that here.
```suggestion
       {{ t('specPage.runAllSpecs', specNumber) }}
```

packages/app/src/specs/SpecsRunAllSpecs.cy.tsx
```diff
@@ -0,0 +1,39 @@
...
+      })
+    })
+
+    it('renders play icon and expected text', () => {
+      cy.findByTestId('run-all-specs-for-cypress/e2e').as('button').children().should('have.length', 2)
+      cy.get('@button').should('contain.text', 'Run 25 specs')
+      cy.findByTestId('play-button').should('be.visible')
+    })
+
+    it('Renders styles correctly', () => {
```
marktnoonan, line 21: In the future let's use a Percy snapshot for this kind of thing and not assert all these CSS values. It's not important whether or not these exact values are used for these elements, what matters is that visual appearance is as expected.

astone123 (approved):

packages/app/src/specs/InlineRunAllSpecs.vue
lmiller1990: I moved the listener anyway, it should be on the button. There is a follow up for better a11y.

packages/app/src/specs/InlineRunAllSpecs.vue
lmiller1990: done

lmiller1990:
Some known windows linting fails, fixed by https://github.com/cypress-io/cypress/pull/24758

All other pipelines :heavy_check_mark:, merging.

PR #24757, opened 2022-11-21
fix: run-all UI tweaks and keyboard nav

### LINKED ISSUE #24748: Better a11y for experimentalRunAllSpecs
### What would you like?

To be able to use the experimentalRunAllSpecs fully with keyboard only.

As it stands, you need to hover with your mouse to see the button, and you cannot get there without a mouse.

image

This is a bit tricky to implement due to a few things. Primarily, we show the Run All button when the row is hovered. The a11y friendly version is `focus` - when the row is focused, eg by hitting tab, we show the button. The problem is if you hit tab AGAIN to go *to* the Run All button, because the row loses focus, the Run All button is no longer rendered, and cannot be focused.

Ideally, we'd do this entire a11y implementation in CSS, but I'm not sure what the best practice is here.

### Why is this needed?

a11y is important. It was originally slated for the initial experimentalRunAllSpecs release, but due to unexpected complexity, we are going to revisit it separately in the future. This will need to be solve prior to GA, or even Beta.

### Other

_No response_

ZachJW34:
- Closes #24748

### User facing changelog
n/a

### Additional details
Some UI tweaks and adding proper keyboard nav for run-all button.

Things to notice:
- The "play" button was always colored blue when hovering on the row, it should only be blue when hovering over the button
- Keyboard nav did not show the button

### Steps to test
Checkout out the branch, run `yarn watch` and `yarn cypress:open` in separate terminals. In an external project, add `experimentalRunAllSpecs: true` to the `cypress.e2e` config.

Before:

https://user-images.githubusercontent.com/25158820/203130869-c62058bf-60c3-4ca0-80e9-5662a1d05489.mov

After:

https://user-images.githubusercontent.com/25158820/203130660-c654cde4-5472-491b-9ea5-bc025b0613e4.mov

### PR Tasks


amehta265 (approved):
All the UI looks good to me. The tab feature works pretty well on the specs list and all the buttons can be pressed without the use of the mouse which is great!

Just one question: I was wondering if the tabbing functionality should also exist on the `inLineSpecList` because currently it does not

ZachJW34:
It probably should, but it's not a regression from this PR. I also updated the PR to show the button on tru keyboard nav (which is strangely only available on the inline spec list... you can tab the Specs page list and you can keyboard navigate the inline but not both)

packages/app/src/specs/SpecsList.vue
```diff
@@ -464,4 +463,12 @@ function onRunAllSpecs (rowId: string) {
 .spec-list-container {
   height: calc(100% - 112px)
 }
+
+/**
+ * Can't put a group on the parent element as it has downstream effects on the styling of child components
```
lmiller1990, line 468: We should move from Windi to Tailwind - it's got named groups which solves this (and way better docs, and better maintained).

lmiller1990 (approved):

PR #24758, opened 2022-11-21
fix: fix windows-lint CI job

astone123:
The `windows-lint` CI job has been failing as of recent because of different line endings between Windows and Unix operating systems. We have configured Git to replace line endings in JSON files, but not in `.eslintrc` files (which use JSON).

So, when Circle checks out the code on Windows, the line endings are CRLF rather than LF. This causes ESLint to fail when linting `.eslintrc` files (which we only do in `npm/eslint-plugin-dev/test`) because it is expecting different line endings.

This PR adds `**/.eslintrc` files to the `.gitattributes` configuration so that the line endings in those files are always LF, no matter the operating system. I also fixed the other warnings that were being logged in the lint job just to clean up some noise.

[Example failing `windows-lint` job](https://app.circleci.com/pipelines/github/cypress-io/cypress/45961/workflows/3d4288ca-df41-4dd1-afe4-3f640a116451/jobs/1933932)
[Passing `windows-lint` job on most recent commit](https://app.circleci.com/pipelines/github/cypress-io/cypress/46011/workflows/b0987ad6-1ff9-450a-a309-077d521187f7/jobs/1934413)

I also added `--concurrency 2` to the `yarn lint` command so that we don't run out of memory in CI ([which has been happening recently](https://app.circleci.com/pipelines/github/cypress-io/cypress/45961/workflows/b4588aba-f74c-4cde-ae2d-eda41d1092ca/jobs/1933862) in our `linux-lint` runs)


.gitattributes
```diff
@@ -2,4 +2,6 @@
 
 *.json text eol=lf
 
+**/.eslintrc text eol=lf
```
astone123, line 5: This is the change that makes the `windows-lint` job pass again

npm/vue/cypress/component/test-utils-api/TestUtilsApi.vue
```diff
@@ -6,7 +6,10 @@
 <script>
 export default {
   props: {
-    msg: String,
+    msg: {
+      type: String,
```
lmiller1990, line 10: Was this intended?

lmiller1990 (approved):

lmiller1990:
image


Nice

npm/vue/cypress/component/test-utils-api/TestUtilsApi.vue
astone123: Yeah, this was to resolve one of the warnings that was showing up during linting just to clean some things up. It looks like ESLint requires a default for Vue props the way we have it configured, or else it warns.

packages/graphql/.eslintignore
```diff
@@ -0,0 +1 @@
+**/tsconfig.json
```
emilyrohrbough, line 1: why don't we want to lint the tsconfig?

astone123: I added these ignore files because there's a plugin that is already skipping these files and when it does it logs a warning. Wanted to clean those up

```
/root/cypress/packages/graphql/tsconfig.json
  0:0  warning  [@cypress/eslint-plugin-json]: Skipping file due to "json/json-with-comments-files" setting: 
  ["**/tsconfig.json", ".vscode/**"]
To remove this warning add "tsconfig.json" to your `.eslintignore` file
```

emilyrohrbough (approved):

PR #24761, opened 2022-11-21
chore: run all specs tweaks

lmiller1990:


packages/app/src/specs/InlineRunAllSpecs.vue
```diff
@@ -4,7 +4,7 @@
     class="h-full truncate"
     data-cy="tooltip"
   >
-    <button>
+    <button @click.stop="emits('runAllSpecs')">
```
ZachJW34, line 7: Makes sense here, but should we have a follow up issue created for it not working when hitting `{enter}` on the keyboard? @marktnoonan isn't this what you observed?

lmiller1990: Yep https://github.com/cypress-io/cypress/issues/24762

ZachJW34 (approved):

packages/app/src/specs/InlineRunAllSpecs.vue
marktnoonan, line 7: This doesn't fix the a11y issue (probably still due to the parent's `click.capture`? Would need to look in more detail), but since there is a separate issue for that, we can continue there. It's still a better place for the click handler.

marktnoonan (approved):
Nice!

packages/app/src/specs/InlineRunAllSpecs.vue
lmiller1990: https://github.com/cypress-io/cypress/issues/24762

PR #24763, opened 2022-11-21
fix: run-all-specs duplicate directory matching

ZachJW34:
Needed to verify that, once we matched a file to a directory that the match ended with a separator, denoting it is the end of the dir path.

https://user-images.githubusercontent.com/25158820/203178798-b152faa8-0e9f-47a3-a195-0f34ef508443.mov


packages/app/src/composables/useRunAllSpecs.ts
```diff
@@ -48,5 +50,20 @@ export function useRunAllSpecs () {
 
       return Boolean(isE2E && hasExperiment)
     }),
+    directoryChildren: computed(() => {
```
lmiller1990, line 53: Ugh, lol. 

Nice job!

lmiller1990 (approved):

PR #24771, opened 2022-11-22
fix: A docblock pointing to a non-existent online tool

DCzajkowski:
Fixed a URL in a docblock pointing to a non-existent tool. It is problematic, as the domain is for sale now and could point to some NSFW content any time.

We could also say "We suggest using an online glob-testing tool to verify what files would match."

Disclaimer: I have no association with the newly-linked online tool.

### PR Tasks

### DIFF
--- a/cli/types/cypress.d.ts
+++ b/cli/types/cypress.d.ts
@@ -2695,7 +2695,7 @@ declare namespace Cypress {
      */
     env: { [key: string]: any }
     /**
-     * A String or Array of glob patterns used to ignore test files that would otherwise be shown in your list of tests. Cypress uses minimatch with the options: {dot: true, matchBase: true}. We suggest using https://toools.cloud/miscellaneous/glob-tester to test what files would match.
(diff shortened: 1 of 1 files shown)


cli/types/cypress.d.ts
```diff
@@ -2695,7 +2695,7 @@ declare namespace Cypress {
      */
     env: { [key: string]: any }
     /**
-     * A String or Array of glob patterns used to ignore test files that would otherwise be shown in your list of tests. Cypress uses minimatch with the options: {dot: true, matchBase: true}. We suggest using http://globtester.com to test what files would match.
+     * A String or Array of glob patterns used to ignore test files that would otherwise be shown in your list of tests. Cypress uses minimatch with the options: {dot: true, matchBase: true}. We suggest using https://toools.cloud/miscellaneous/glob-tester to test what files would match.
```
nagash77:
```suggestion
     * A String or Array of glob patterns used to ignore test files that would otherwise be shown in your list of tests. Cypress uses minimatch with the options: {dot: true, matchBase: true}. We suggest using a tool to test what files would match.
```

PR #24799, opened 2022-11-23
feat: Improve CLOUD_PARALLEL_GROUP_PARAMS_MISMATCH error message

### LINKED ISSUE #24734: Update PARALLEL_GROUP_PARAMS_MISMATCH error
### What would you like?

Update the error that is thrown when the cloud responds with `PARALLEL_GROUP_PARAMS_MISMATCH` - the cloud is gonna include in the payload the difference in the specs (in case there are) and also the difference in the params (in case there are)

Payload:
```
{
  runId: string,
  runUrl: string,
  differentSpecs: string[],
  differentParams: {
    [string]: {
      previous: string,
      current: string
    }
  },
}
```

This change only adds a new field in the returned payload (`differentParams`).

Also, we can reshape the `differentParams` type in the case is needed

### Why is this needed?

To simplify the process of identifying machine mismatches

### Other

_No response_

estrada9166:
- Closes  https://github.com/cypress-io/cypress/issues/24734

### User facing changelog

Using the returned payload by the cloud, we should be able to show an error message displaying the mismatch information when `CLOUD_PARALLEL_GROUP_PARAMS_MISMATCH` is thrown

image

image

### PR Tasks


packages/errors/test/unit/visualSnapshotErrors_spec.ts
```diff
@@ -485,6 +485,23 @@ describe('visual error templates', () => {
               browserVersion: '59.1.2.3',
               specs: [
                 'cypress/integration/app_spec.js',
+                'cypress/integration/foo_spec.js',
+                'cypress/integration/bar_spec.js',
+              ],
+            },
+            payload: {
```
emilyrohrbough, line 509: Can you update this to test both the default scenario & this test case?

packages/server/__snapshots__/cypress_spec.js
```diff
@@ -203,10 +203,10 @@ This machine sent the following parameters:
...
   "osVersion": "v1",
-  "browserName": "Electron",
-  "browserVersion": "59.1.2.3",
-  "specs": [
-    "cypress/e2e/app.cy.js"
+  "browserName": "Electron.... (Expected: Electron)",
+  "browserVersion": "59.1.2.3.... (Expected: 64)",
+  "differentSpecs": [
+    "cypress/integration/foo_spec.js"
   ]
```
emilyrohrbough: Curious how different specs come into play here. If this is starting the same parallel run, shouldn't this be a different set of specs ?

packages/errors/src/errors.ts
```diff
@@ -279,7 +279,23 @@ export const AllCypressErrors = {
...
         https://on.cypress.io/parallel-disallowed`
   },
-  CLOUD_PARALLEL_GROUP_PARAMS_MISMATCH: (arg1: {runUrl: string, parameters: any}) => {
+  CLOUD_PARALLEL_GROUP_PARAMS_MISMATCH: (arg1: {runUrl: string, parameters: any, payload: any }) => {
+    let params: any = arg1.parameters
+
+    if (arg1.payload?.differentParams) {
+      params = {}
+
+      _.map(arg1.parameters, (value, key) => {
```
emilyrohrbough, line 288:
```suggestion
      _.map(params, (value, key) => {
```

packages/errors/src/errors.ts
```diff
@@ -279,7 +279,23 @@ export const AllCypressErrors = {
 
         https://on.cypress.io/parallel-disallowed`
   },
-  CLOUD_PARALLEL_GROUP_PARAMS_MISMATCH: (arg1: {runUrl: string, parameters: any}) => {
+  CLOUD_PARALLEL_GROUP_PARAMS_MISMATCH: (arg1: {runUrl: string, parameters: any, payload: any }) => {
+    let params: any = arg1.parameters
+
+    if (arg1.payload?.differentParams) {
```
emilyrohrbough, line 285: when would `differentParams` not exist if there is a mismatch?

packages/errors/src/errors.ts
```diff
@@ -279,7 +279,23 @@ export const AllCypressErrors = {
...
+      params = {}
+
+      _.map(arg1.parameters, (value, key) => {
+        if (key === 'specs') {
```
emilyrohrbough:
```suggestion
        if (key === 'specs'  && arg1.payload.differentSpecs) {
```
do we need to include `differentSpecs` at all if there is nothing to show the user?

packages/errors/src/errors.ts
```diff
@@ -279,7 +279,23 @@ export const AllCypressErrors = {
...
+          params['differentSpecs'] = arg1.payload.differentSpecs ?? []
```
emilyrohrbough: why is show this attribute as the negative where as the? Are we able to differentiate if a new spec was sent or a specific spec was missing from the payload?

packages/server/__snapshots__/cypress_spec.js
estrada9166: Is possible that in the same way that the params can change, the `--spec` flag can be set in one machine, running a different set of specs?

packages/errors/src/errors.ts
estrada9166: do we want to show different error messages depending on the payload? If there are no `differentSpecs`  we show `specs` with all the specs ran but if there're different specs, we are showing it

packages/errors/src/errors.ts
estrada9166: We're able to get the different values in the cloud, keeping track of the specs sent by the TR

packages/errors/src/errors.ts
estrada9166: This is more a safeguard in case the cloud changes were released after the app

packages/errors/src/errors.ts
estrada9166: If we map `params` we may map an empty object https://github.com/cypress-io/cypress/pull/24799/files/5637f13fd9c6b90823fddbc4e82f7e9ca7c30355#diff-18a8227d9456573d2bf0ae8d3cc1a2d6f3552e85f8c2345d946f7189160149e9R286

packages/errors/src/errors.ts
emilyrohrbough: If the payload is meant to show the differences on why this errored, I think providing: `differentSpecs: []` is a little confusing / doesn't provide much but can go either way.

packages/errors/src/errors.ts
emilyrohrbough: so diff specs is only tied to the current run that errored right?

How does this look if it sent 1 additional spec vs sent 1 spec less?

packages/errors/src/errors.ts
estrada9166: Going to remove that field in case it is empty! you're right! it can be confusing

packages/errors/src/errors.ts
estrada9166: When there's one spec extra, its gonna show that spec extra, and the same when there's one less, it shows the spec missing

emilyrohrbough (approved):

emilyrohrbough:
@estrada9166 there's a failing system-test.

ryanthemanuel (approved):

PR #24832, opened 2022-11-28
feat: [BREAKING] make cypress/react default to React 18 support

### LINKED ISSUE #24160: Update react/mount to point to react18 mount
### Current behavior

Currently react/mount points to the component mount that works for react <=17.

### Desired behavior

As part as our [evergreen strategy](https://github.com/cypress-io/prod-eng-docs/blob/main/team-component-testing/ecosytemsupport.md#how-we-will-handle-updates) with Cy 11 we want react/mount to point the the react18/mount. We should leave react18/mount so that there are no dead ends.

lmiller1990:
- Closes #24160

### User facing changelog

`cypress/react` now defaults to React 18. For React <= 17 support, use `cypress/react17`.

### Additional details

*Note*: this PR targets `release/12.0.0`.

As part as our [evergreen strategy](https://github.com/cypress-io/prod-eng-docs/blob/main/team-component-testing/ecosytemsupport.md#how-we-will-handle-updates) with Cy 11 we want `cypress/react` to work with the latest React version, which is now React 18. We should leave `cypress/react18` so that there are no dead ends, as well as add a `cypress/react17` for users who are on React 16 and 17 (like the Cypress monorepo).

### Steps to test

Note: you probably want to use the pre-release binaries. Grab them from the bot comment on the latest commit.

- Try making a new React project (CRA or Vite) that uses React 18. Install the binary - it should work out of the box. It will scaffold `cypress/react` in the `support/component.js` - not `cypress/react18`, like we currently do.
- Try making a new React project (CRA or Vite) that uses React 17. Install the binary - it should work out of the box. It will scaffold `cypress/react17` in the `support/component.js` - not `cypress/react`, like we currently do.
- Try an existing project using React 17 working with Cypress 11. Upgrade to Cypress 12 by installing the binary - it should NOT work out of the box. You get VERY confusing error. I do not have any way around this - `cypress/react` imports `react-dom/client`, which does NOT exist in React <= 17. Compilation fails. I am open to suggestions to improve this. For now, we need to have good messaging and point users to the migration guide. This will be discussed at the product sync meeting later this week.

Note: I highlighted the Vite error in the terminal - the launchpad error is basically impossible to decipher. 

image

I am not particularly happy with this. The alternatives are

- migration guide (text based). Cheap and easy, but not as discoverable.
- migration UI (expensive, time consuming to build)
- change our strategy around evergreen versioning 

### PR Tasks

- [ TODO ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)?  *will do pending product review*


npm/react-mount-utils/src/index.ts
```diff
@@ -1,12 +1,64 @@
 import * as React from 'react'
-import getDisplayName from './getDisplayName'
+import type { ReactNode } from 'react'
```
lmiller1990, line 2: All shared react utils are here now, all the react adapters are thin layers on top of this. This proved the easiest and cleanest way to solve this.

npm/react/cypress/component/basic/rerender/effects.cy.jsx
```diff
@@ -35,7 +36,8 @@ it('should not run unmount effect cleanup when rerendering', () => {
   })
 })
 
-it('should run unmount effect cleanup when unmounting', () => {
+// TODO: is this still valid in React 18? Does re-rendering cause previous effects to be cleaned up?
+it.skip('should run unmount effect cleanup when unmounting', () => {
```
lmiller1990, line 40: I am not even sure if this test makes sense, I'm pretty sure re-rendering should not cause the return value from `useEffect` to be triggered - only when unmounting.

npm/react18/src/index.ts
```diff
@@ -1,85 +1 @@
...
-export function unmount (options: UnmountArgs = { log: true }) {
-  // @ts-expect-error - undocumented API
-  Cypress.utils.throwErrByPath('mount.unmount')
-}
-
-// Re-export this to help with migrating away from `unmount`
-export {
-  getContainerEl,
-}
+export * from '@cypress/react'
```
lmiller1990, line 1: No dead ends.

Thought this was pretty neat :sunglasses:

npm/react-mount-utils/package.json
```diff
@@ -0,0 +1,37 @@
+{
```
ZachJW34, line 1: Should we mark this package as private and not publish it?

lmiller1990: Yes, good idea.

Thanks for the early review, I'll get this one green today.

system-tests/projects/react17/src/UsingLegacyMount.cy.jsx
```diff
@@ -1,22 +1,24 @@
-import React from 'react'
-import { mount } from 'cypress/react'
+// import React from 'react'
```
lmiller1990, line 1: I do not have a good solution here. This test cannot execute - in fact, it doesn't even compile,. since `cypress/react` assumes `react-dom/client` exists, which is a React 18 only API. This causes Vite to throw a compilation error.

npm/react-mount-utils/package.json
lmiller1990: Done.

npm/react-mount-utils/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "watch": "tsc -w",
+    "lint": "eslint --ext .js,.ts,.json, ."
+  },
+  "devDependencies": {
+    "@cypress/mount-utils": "0.0.0-development",
+    "@rollup/plugin-commonjs": "^17.1.0",
+    "@rollup/plugin-node-resolve": "^11.1.1",
+    "rollup": "^2.38.5",
+    "rollup-plugin-dts": "^4.2.3",
+    "rollup-plugin-typescript2": "^0.29.0",
```
ZachJW34: Can remove the rollup deps

.circleci/workflows.yml
```diff
@@ -1847,6 +1848,15 @@ jobs:
           command: yarn workspace @cypress/mount-utils build
       - store-npm-logs
 
+  npm-react-mount-utils:
```
ZachJW34, line 1851: We don't have to do this now but what do you think of having a shared job that builds all of the npm packages before they get their own separate job? Cloning and restoring the workspace is expensive for just this build.

npm/react17/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "@rollup/plugin-commonjs": "^17.1.0",
+    "@rollup/plugin-node-resolve": "^11.1.1",
+    "@types/react": "^17",
+    "@types/react-dom": "^17",
+    "cypress": "0.0.0-development",
+    "react": "^17",
+    "react-dom": "^17",
+    "rollup": "^2.38.5",
+    "rollup-plugin-typescript2": "^0.29.0",
+    "semver": "^7.3.2",
```
ZachJW34:
```suggestion
```

No longer using semver in this package

ZachJW34 (commented):
For the error, could we catch it when the error bubbles up in `packages/data-context/src/data/ProjectConfigIpc.ts`? We could search for `Could not resolve "react-dom/client", then tweak the error message to display something that makes more sense?

Just a thought, not sure how Webpack displays this error or if it's even something that can be caught at runtime

.circleci/workflows.yml
lmiller1990: This is probably a good idea, there's way too many separate jobs on CI now.

npm/react-mount-utils/package.json
lmiller1990: done

lmiller1990:
> For the error, could we catch it when the error bubbles up in `packages/data-context/src/data/ProjectConfigIpc.ts`? We could search for `Could not resolve "react-dom/client", then tweak the error message to display something that makes more sense?
> 
> Just a thought, not sure how Webpack displays this error or if it's even something that can be caught at runtime

I took a stab at this, I couldn't find a way to catch this error with webpack. There might be some way if we join everything up, but I think it's likely to increase the scope and complexity fairly significantly.

I feel like the best option might be a proper migration UI - we'd just find all references to `"cypress/react"` and update them. This, too, increases the scope. We'd need designs, which means we'd be unlikely to ship this with Cypress 12, since Cypress 12 is due out soon.

Let's see what Brian thinks at the product sync meeting tomorrow. In the meantime, I've responded to all your other comments (mostly removing obsolete code).

npm/react-mount-utils/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@cypress/react-mount-utils",
+  "version": "0.0.0-development",
+  "description": "Shared utilities for the various component testing adapters",
```
mike-plummer, line 4: nit: "_React_ component testing adapters"

npm/react-mount-utils/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "files": [
+    "dist"
+  ],
+  "types": "dist/index.d.ts",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/cypress-io/cypress.git"
+  },
+  "homepage": "https://github.com/cypress-io/cypress/tree/develop/npm/mount-utils#readme",
```
mike-plummer, line 28: Tweak url for this package

npm/react-mount-utils/src/index.ts
```diff
@@ -1,12 +1,64 @@
...
+  component: React.ReactNode
+  /**
+   * Rerenders the specified component with new props. This allows testing of components that store state (`setState`)
+   * or have asynchronous updates (`useEffect`, `useLayoutEffect`).
+   */
+  rerender: (component: React.ReactNode) => globalThis.Cypress.Chainable<MountReturn>
+  /**
+   * Removes the mounted component.
+   *
+   * Removed as of Cypress 11.0.0.
```
mike-plummer, line 57: Mark as `@deprecated`?

system-tests/projects/react18/src/UsingLegacyMount.cy.jsx
```diff
@@ -1,23 +1,22 @@
 import React from 'react'
-import { mount } from 'cypress/react'
+import { mount } from 'cypress/react18'
 
 function App () {
   return <h1>Hello world</h1>
 }
 
 describe('using legacy mount', () => {
```
mike-plummer, line 8: To validate the "legacy" mount behavior would we want this React18 test to attempt using the React17 mount and validate the user gets the helpful message?

npm/react17/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "files": [
+    "dist"
+  ],
+  "types": "dist/index.d.ts",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/cypress-io/cypress.git"
+  },
+  "homepage": "https://github.com/cypress-io/cypress/blob/develop/npm/react18/#readme",
```
mike-plummer, line 44: Should be `react17`

npm/react17/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@cypress/react17",
+  "version": "0.0.0-development",
+  "description": "Test React components using Cypress",
```
mike-plummer, line 4: Reference that this is only for React <=17?

npm/react17/README.md
```diff
@@ -0,0 +1,7 @@
+# @cypress/react18
```
mike-plummer, line 1: Multiple 18's that should be "17" in this file

npm/react17/CHANGELOG.md
```diff
@@ -0,0 +1,40 @@
+# [@cypress/react18-v2.0.0](https://github.com/cypress-io/cypress/compare/@cypress/react18-v1.1.1...@cypress/react18-v2.0.0) (2022-11-07)
```
mike-plummer, line 1: This file feels odd - copy of the React18 changelog?

npm/react/src/index.ts
```diff
@@ -1,7 +1,94 @@
...
+
+  return makeMountFn('mount', jsx, { ReactDom: ReactDOM, ...options }, rerenderKey, internalOptions)
+}
+
+function internalUnmount (options = { log: true }) {
+  return makeUnmountFn(options)
+}
+/**
+ * Removed as of Cypress 11.0.0.
+ * @see https://on.cypress.io/migration-11-0-0-component-testing-updates
```
mike-plummer, line 84: Deprecate?

system-tests/projects/react17/src/UsingLegacyMount.cy.jsx
mike-plummer: Could we use `mock-require` or `proxyquire` to stub out the `react-dom/client` module just to make Vite happy while it resolves dependencies?

mike-plummer (commented):
Awesome work - seems to work well for 17 & 18 cases locally. Will be curious to see what the group thinks about the 18 module resolution issue in launchpad, can't say I have any bright ideas on that.

Couple comments, mostly just docs and typos.

system-tests/projects/react17/src/UsingLegacyMount.cy.jsx
lmiller1990: Interesting idea... I haven't `proxyquire` - It might be possible, we'd need to make sure it works for both bundlers... at this point, I feel like a clean migration UI might even be more simple and cleaner, without adding too much technical complexity. As we also discussed earlier, we won't rush this for Cy 12, so maybe we can spend a bit more time thinking about it.

PR #24846, opened 2022-11-28
fix: add top-level run-all button to allow all specs to be run at once

### LINKED ISSUE #24759: Allow running all specs without root directory
### What would you like?

The new Run All Specs works on a directory level. If you haven't got a single root directory, you cannot truly "Run All Specs.

The fix is an additional button at the top, like this. The text changes to "Run N Specs" when using the text search.

Screen Shot 2022-11-22 at 5 50 37 am

### Why is this needed?

_No response_

### Other

_No response_

ZachJW34:
- Closes #24759 

### User facing changelog
experimentalRunAllSpecs: Allow running all specs regardless of folder structure by adding a top-level run-all button.

### Additional details
We only allowed running "all specs" if the user happened to have one root folder e.g. storing all specs under `cypress/e2e`. If they had two root folders containing specs, this wasn't possible.

There is now a "top-level" run all button in both the Specs Page SpecsList and the Runner SpecsList.

DRAFT: Need to verify InlineSpecList button with design.

### Steps to test
Checkout this branch, choose a project and add `experimentalRunAllSpecs` to the `e2e` config. Run `yarn watch` and `yarn cypress:open` in two separate terminals and then choose your project. Verify the Specs page and the Runner has the top-level "Run All Specs" button and that it functions correctly.

https://user-images.githubusercontent.com/25158820/204351144-6a502ef1-8d24-4b1f-9864-56eb2665f34e.mov

### PR Tasks


packages/app/src/store/run-all-specs-store.ts
```diff
@@ -0,0 +1,96 @@
...
+gql`
+mutation RunAllSpecs ($specPath: String!, $runAllSpecs: [String!]!) {
+  setRunAllSpecs(runAllSpecs: $runAllSpecs)
+  launchOpenProject(specPath: $specPath) {
+    id
+  }
+}
+`
+
+const isRunMode = window.__CYPRESS_MODE__ === 'run' && window.top === window
```
lmiller1990: We've got something like

```js
const isRunMode = window.__CYPRESS_MODE__ === 'run'
```

In quite a few places - any chance we can move this to a single function?

packages/app/src/store/run-all-specs-store.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export const useRunAllSpecsStore = defineStore('runAllSpecs', () => {
```
lmiller1990, line 33: Should we be consistent and use the "options" style store that we use for our other stores? Or is there a limitation preventing us from using that style here?

packages/app/src/store/run-all-specs-store.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    const config: ResolvedConfig = query.data.value?.currentProject?.config || []
+    const hasExperiment = config.find(({ field, value }) => field === 'experimentalRunAllSpecs' && value === true)
+
+    return Boolean(isE2E && hasExperiment)
+  })
+
+  return {
+    isRunAllSpecsAllowed,
+    allSpecs: allSpecsRef,
```
lmiller1990: Seems `allSpecs` is not used outside the store; a private ref - should we not return it here (thus not exposing it)?

lmiller1990 (commented):
Did not test it out -- still in draft -- just left a few comments.

packages/app/src/store/run-all-specs-store.ts
ZachJW34: Only way to execute the gql query without having to call an explicit action. I didn't want to convert to using the store but I would have had to refactor the InlineSpecsList to get the communication to work. I'll think on this a bit more as I'm not too happy with the implementation.

packages/app/src/store/run-all-specs-store.ts
ZachJW34: You're right! It became unused after a refactor

Edit: Addressed in [b86fb64](https://github.com/cypress-io/cypress/pull/24846/commits/b86fb641a9ed191aaf128f295a1224b8929d12e1)

packages/app/src/store/run-all-specs-store.ts
ZachJW34: Can do, I think the `window.top === window` is for cy-in-cy? Makes sense to consolidate that as well, going to include that and see if anything breaks.

Addressed in [b86fb64](https://github.com/cypress-io/cypress/pull/24846/commits/b86fb641a9ed191aaf128f295a1224b8929d12e1)

packages/app/cypress/e2e/run-all-specs.cy.ts
```diff
@@ -6,9 +6,15 @@ describe('run-all-specs', () => {
     spec2: { relative: 'cypress/e2e/folder-a/spec-b.cy.js', name: 'runs folder-a/spec-b' },
     spec3: { relative: 'cypress/e2e/folder-b/spec-a.cy.js', name: 'runs folder-b/spec-a' },
     spec4: { relative: 'cypress/e2e/folder-b/spec-b.cy.js', name: 'runs folder-b/spec-b' },
+    spec5: { relative: 'folder-c/spec-a.cy.js', name: 'runs folder-c/spec-a' },
+    spec6: { relative: 'folder-c/spec-b.cy.js', name: 'runs folder-c/spec-b' },
   }
 
   const clickRunAllSpecs = (directory: string) => {
+    if (directory === 'all') {
+      return cy.get('[data-cy=run-all-specs-for-all]').click()
```
lmiller1990: Nit, could do `cy.getByTestId`

packages/app/src/pages/Specs/Runner.vue
```diff
@@ -31,6 +31,7 @@ import SpecRunnerContainerOpenMode from '../../runner/SpecRunnerContainerOpenMod
 import SpecRunnerContainerRunMode from '../../runner/SpecRunnerContainerRunMode.vue'
 import { useEventManager } from '../../runner/useEventManager'
 import { useSpecStore } from '../../store'
+import { isRunMode } from '@packages/frontend-shared/src/utils/isRunMode'
```
lmiller1990, line 34: I don't think this belongs in `frontend-shared` - the concept of open/run mode is exclusively to the App. Any particular reason to put in in `frontend-shared`?

packages/app/src/pages/Specs/Runner.vue
```diff
@@ -59,14 +60,12 @@ subscription Runner_ConfigChange {
...
 
-const isRunMode = window.__CYPRESS_MODE__ === 'run'
-
 // subscriptions are used to trigger live updates without
 // reloading the page.
 // this is only useful in open mode - in run mode, we don't
 // use GraphQL, so we pause the
 // subscriptions so they never execute.
-const shouldPauseSubscriptions = isRunMode && window.top === window
+const shouldPauseSubscriptions = isRunMode
```
lmiller1990: Seems like we don't need this temporary variable. You could update line 82:

```js
  pause: isRunMode
```

packages/app/src/specs/InlineSpecList.cy.tsx
```diff
@@ -207,18 +190,43 @@ describe('InlineSpecList', () => {
   })
 
   describe('Run all Specs', () => {
+    const hoverRunAllSpecs = (directory?: string, specNumber?: number) => {
```
lmiller1990: This is only called in two places and both `directory` and `specNumber` are always defined. We can remove the `?`. It also means line 199 is never executed.

packages/app/src/specs/InlineSpecListHeader.vue
```diff
@@ -60,26 +50,36 @@
...
+        </span>
+      </template>
+    </Tooltip>
+    <InlineRunAllSpecs
+      v-if="isRunAllSpecsAllowed"
+      spec-number="all"
+      directory="all"
+      grayscale
+      class="rounded-md flex outline-none border-1 border-gray-900 h-24px w-24px duration-300 hocus-default items-center justify-center hocus:ring-0 hocus:border-indigo-300"
+      @runAllSpecs="emit('runAllSpecs')"
```
lmiller1990, line 81: It looks like we just emit up to the top then do `store.runAllSpecs` - and particular reason to emit up then call the method (as opposed to just importing the store and calling the method inside of the `InlineRunAllSpecs` component)? Or was that to make it more testable?

packages/app/src/store/run-all-specs-store.ts
lmiller1990: I'm still not entirely clear why this isn't possible using the options style store, but happy to move forward as is. I also feel this implementation isn't ideal, but I can't really nail down why - other than the fact we had to touch so many files to implement it.

Anyway, we can add this to our mental list of "things to improve" if/when we get a chance to attack some of the technical debt. It would be nice if the stores could be all written in a consistent style, though - what's the blocker on the options style store?

lmiller1990 (commented):
I left some comments, no obvious blocker, mostly style/architectural things.

lmiller1990:
According to the issue linked, the label should be "Run N Specs" when using the filter. Eg:

image

This should be "Run 2 specs". Did that requirement change? I might have missed it.

The actual feature is working as expected. :100

astone123:
Worked well for me, just waiting on the answer to [Lachlan's comment](https://github.com/cypress-io/cypress/pull/24846#issuecomment-1331645364) - seems like a missing requirement

packages/app/src/specs/InlineRunAllSpecs.vue
```diff
@@ -22,7 +24,7 @@
         class="font-normal text-sm inline-flex"
         data-cy="tooltip-content"
       >
-        {{ t('specPage.runAllSpecs', specNumber) }}
+        {{ specNumber === 'all' ? t('specPage.runAllSpecs') : t('specPage.runSelectedSpecs', specNumber) }}
```
mike-plummer: Could we just use `RUN_ALL_SPECS_KEY` instead of the "all" magic string?

packages/app/src/specs/InlineSpecList.vue
```diff
@@ -80,4 +83,6 @@ const specs = computed<FuzzyFoundSpec[]>(() => {
   return fuzzySortSpecs(specs, debouncedSpecFilterModel.value)
 })
 
+const runAllSpecsStore = useRunAllSpecsStore()
```
mike-plummer, line 86: nit: Thoughts on just destructuring here to simplify the template expressions above?

`const { isRunAllSpecsAllowed, runAllSpecs } = useRunAllSpecsStore()`

packages/app/src/store/run-all-specs-store.ts
```diff
@@ -0,0 +1,94 @@
+import { RUN_ALL_SPECS_KEY } from '@packages/types/src'
+import { gql, useMutation, useQuery } from '@urql/vue'
+import { defineStore } from 'pinia'
+import { computed, ref } from 'vue'
+import { useRouter } from 'vue-router'
+import { RunAllSpecsDataDocument, RunAllSpecsDocument } from '../generated/graphql-test'
```
mike-plummer: Should probably be pulling from `generated/graphql`, not `graphql-test`

packages/app/src/store/run-all-specs-store.ts
```diff
@@ -0,0 +1,94 @@
...
+import { getSeparator, SpecTreeNode, UseCollapsibleTreeNode } from '../specs/tree/useCollapsibleTree'
+import { isRunMode } from '@packages/frontend-shared/src/utils/isRunMode'
+
+type ResolvedConfig = { value: any, from: 'string', field: string }[]
```
mike-plummer: Are the quotes around `'string'` intentional?

packages/app/src/store/run-all-specs-store.ts
```diff
@@ -0,0 +1,94 @@
...
+    const directoryChildren: Record<string, string[]> = {}
+
+    for (const { id, isLeaf } of tree) {
+      if (!isLeaf) {
+        directoryChildren[id] = []
+      } else {
+        allSpecs.push(id)
+
+        Object.keys(directoryChildren).forEach((dir) => {
+          if (id.startsWith(dir) && id.replace(dir, '').startsWith(separator)) {
```
mike-plummer, line 69: nit: Could simplify

```
if (id.startsWith(`${dir}${separator}`) {
```

packages/app/src/store/run-all-specs-store.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    allSpecsRef.value = allSpecs
+    directoryChildrenRef.value = directoryChildren
+  }
+
+  const isRunAllSpecsAllowed = computed(() => {
+    const isE2E = query.data.value?.currentProject?.currentTestingType === 'e2e'
+
+    const config: ResolvedConfig = query.data.value?.currentProject?.config || []
+    const hasExperiment = config.find(({ field, value }) => field === 'experimentalRunAllSpecs' && value === true)
```
mike-plummer: If you use `some` instead of `find` then you'll end up with a simple `boolean` and can remove the `Boolean` constructor below

mike-plummer (changes requested):
Works great! Just one question about expected behavior:

In Inline, if you search for a spec that doesn't exist you can still attempt to do a "Run All" on an empty set of specs. It's gracefully handled (gives the "no tests found" message), but I wonder if it makes more sense to disable the "Run All" control if the inline list is empty

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -218,7 +218,8 @@
         "content": "Record a run to see your test results in Cypress Cloud. You can then optimize your test suite, debug failing and flaky tests, and integrate with your favorite tools."
       }
     },
-    "runAllSpecs": "Run {n} spec | Run {n} specs"
+    "runSelectedSpecs": "Run {n} spec | Run {n} specs",
+    "runAllSpecs": "Run all specs"
```
marktnoonan: I'd suggest we do like 9.x, and when there's a filter active, make the wording be "run filtered specs" for clarity about what to expect.

packages/app/src/specs/InlineSpecList.vue
marktnoonan: Reactive get stale when destructured from the pinia store, which is why I've generally avoided doing it. There's a `storeToRefs` helper we can use if we want. But I ended up kind of liking that when we use reactive values from the store, it's clear where they are coming from, even if it's a bit verbose.

Docs for reference: https://pinia.vuejs.org/core-concepts/index.html#using-the-store

Could be swayed towards using `storeToRefs` if the rest of us like it.

packages/app/src/specs/InlineRunAllSpecs.vue
lmiller1990: I'll go ahead and make this change - I think it makes sense, and Zach W looks to be out for a day.

packages/frontend-shared/src/locales/en-US.json
lmiller1990: I did this here https://github.com/cypress-io/cypress/pull/24846/commits/8e12cbbfc1a70eafa619aae6dff967be8bbb3136

packages/app/src/store/run-all-specs-store.ts
lmiller1990: done

packages/app/src/store/run-all-specs-store.ts
lmiller1990: fixed this

packages/app/src/specs/InlineRunAllSpecs.vue
lmiller1990: Turns out we don't need this anymore, instead we just show the number as recommended by design: "Run N specs".

lmiller1990:
@mike-plummer @ZachJW34 @astone123 @marktnoonan I pushed a commit addressing the final comments to make sure we finish this in a timely fashion. Please check https://github.com/cypress-io/cypress/pull/24846/commits/8e12cbbfc1a70eafa619aae6dff967be8bbb3136

Here's a preview. I also disabled the button on the inline spec list if there are 0 specs after filtering as recommended by @mike-plummer.

https://user-images.githubusercontent.com/19196536/204967993-cefc32dc-0dbc-4db7-b308-1f428d0a8046.mp4

packages/app/src/specs/InlineSpecList.vue
mike-plummer: Ah, didn't know that. Thanks for the info 👍

marktnoonan:
There were a couple of failures in CI, I've pushed [f263211](https://github.com/cypress-io/cypress/pull/24846/commits/f263211654f9be50b74db9b23e7d08bb8c6b21a7) and [3e0a325](https://github.com/cypress-io/cypress/pull/24846/commits/3e0a32595778f2f035f8312532e3d00e252a7fa0) to update the tests. Both pass for me locally now so hopefully CI will be all good now.

mike-plummer (approved):
LGTM 🚀

marktnoonan (approved):

marktnoonan:
I've noticed one of my changes is coincidentally reverting one of Lachlan's changes, seems harmless but [commented here](https://github.com/cypress-io/cypress/commit/8e12cbbfc1a70eafa619aae6dff967be8bbb3136#r91686962) just in case.

PR #24859, opened 2022-11-28
feat: Add new local and session storage commands

### LINKED ISSUE #24276: Add Get/ClearAllLocalStorage
Add new apis for getting All local storage in the browser regardless of domain

this should return an object of domain to local storage mapping.

Also add a new API for clearing all local storage in the browser regardless of domain.

### LINKED ISSUE #24277: Add Get/ClearAllSessionStorage
Add new apis for getting All session storage in the browser regardless of domain

this should return an object of domain to session storage mapping.

Also add a new API for clearing all session storage in the browser regardless of domain.

chrisbreiding:
- Closes #24276
- Closes #24277

### User facing changelog

- (feature) Added `cy.getAllLocalStorage()`, `cy.clearAllLocalStorage()`, `cy.getAllSessionStorage()`, and `cy.clearAllSessionStorage()` commands

### PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4876
- [x] on links? https://github.com/cypress-io/cypress-services/pull/5026


packages/driver/cypress/e2e/commands/local_storage.cy.ts
```diff
@@ -0,0 +1,325 @@
...
+
+    it('does not log when log: false', () => {
+      cy.clearAllSessionStorage({ log: false }).then(() => {
+        assertLogLength(logs, 1)
+        expect(logs[0].get('name')).to.eq('visit')
+      })
+    })
+  })
+
+  context('#clearLocalStorage', () => {
```
chrisbreiding: I converted this file from `.js` to `.ts`. The below tests were the only existing ones in the `.js` file and I left them unchanged outside of a couple type annotations.

packages/driver/src/cy/commands/sessions/manager.ts
```diff
@@ -283,41 +294,34 @@ export default class SessionsManager {
...
-        sessionStorage: originOptions,
-      })
+      if (!type || type === 'sessionStorage') {
+        options.sessionStorage = originOptions
+      }
+
+      await this.sessions.setStorage(options)
     },
 
-    setStorage: async (options: any, clearAll = false) => {
```
chrisbreiding, line 297: As far as I can tell, we don't use the `clearAll` option anywhere, so I removed it and its associated logic.

packages/driver/src/cy/commands/sessions/manager.ts
```diff
@@ -227,7 +238,7 @@ export default class SessionsManager {
       *      send a message telling our proxy server to intercept the next req to the https domain,
       *      then follow 2)
       */
-    getStorage: async (options = {}) => {
```
emilyrohrbough, line 230: Now that we are exposing these as commands,  do we need to include the global session API? I don't think we need to since they are currently undocumented.

packages/driver/src/cy/commands/local_storage.ts
```diff
@@ -3,6 +3,8 @@ import _ from 'lodash'
 import $errUtils from '../../cypress/error_utils'
 import $LocalStorage from '../../cypress/local_storage'
```
emilyrohrbough, line 4: I would have expected `../../cypress/local_storage` to no longer be needed after this change.

packages/driver/src/cy/commands/sessions/manager.ts
```diff
@@ -283,41 +294,34 @@ export default class SessionsManager {
       return results
     },
 
-    clearStorage: async () => {
+    clearStorage: async (type?: InternalCypress.StorageType) => {
       const origins = await this.getAllHtmlOrigins()
```
emilyrohrbough: We can expand this to accept `currentOrigin` and use `mapOrigins()` instead of `await this.getAllHtmlOrigins()` to allows specific origins to be specified as an option to cy.clearLocalStorage()

packages/driver/src/cy/commands/local_storage.ts
```diff
@@ -38,7 +88,13 @@ export default (Commands, Cypress, cy, state) => {
...
 
   Commands.addAll({
-    clearLocalStorage (keys, options: Partial<Cypress.Loggable> = {}) {
+    getAllLocalStorage: getAllStorage.bind(null, 'localStorage', Cypress),
+    getAllSessionStorage: getAllStorage.bind(null, 'sessionStorage', Cypress),
+
+    clearAllLocalStorage: clearAllStorage.bind(null, 'localStorage', Cypress),
+    clearAllSessionStorage: clearAllStorage.bind(null, 'sessionStorage', Cypress),
+
+    clearLocalStorage (keys, options: Options = {}) {
```
emilyrohrbough, line 98: I would expect clearLocalStorage to use the same API as clearAllLocalStorage with the current origin specified.

packages/driver/src/cy/commands/local_storage.ts
```diff
@@ -3,6 +3,8 @@ import _ from 'lodash'
 import $errUtils from '../../cypress/error_utils'
```
emilyrohrbough, line 3: Can we rename this file from local_storage since this handles both local & session storage?

packages/driver/src/cy/commands/sessions/manager.ts
chrisbreiding: Not sure I understand. What would including the global session API mean?

packages/driver/src/cy/commands/local_storage.ts
chrisbreiding: This PR is meant to be purely additive. Any refactoring is out of scope and should be separate.

packages/driver/src/cy/commands/sessions/manager.ts
chrisbreiding: I think that's out of scope for adding these commands.

packages/driver/src/cy/commands/local_storage.ts
chrisbreiding: I'd rather do that in a separate PR.

packages/driver/src/cy/commands/local_storage.ts
chrisbreiding: Addressed in https://github.com/cypress-io/cypress/pull/24859/commits/8437b8eca6a8ecfa0a3ebe447306c484a476f63e

cli/types/cypress.d.ts
```diff
@@ -792,7 +819,35 @@ declare namespace Cypress {
     clearCookies(options?: CookieOptions): Chainable<null>
 
     /**
-     * Clear data in local storage.
+     * Get local storage for all origins.
+     *
+     * @see https://on.cypress.io/getalllocalstorage
+     */
+    getAllLocalStorage(options?: Partial<Loggable>): Chainable<StorageByOrigin>
```
emilyrohrbough, line 826: Shouldn't this return an array?
```suggestion
    getAllLocalStorage(options?: Partial<Loggable>): Chainable<Array<StorageByOrigin>>
```

cli/types/cypress.d.ts
```diff
@@ -792,7 +819,35 @@ declare namespace Cypress {
...
+     * @see https://on.cypress.io/clearalllocalstorage
+     */
+    clearAllLocalStorage(options?: Partial<Loggable>): Chainable<null>
+
+    /**
+     * Get session storage for all origins.
+     *
+     * @see https://on.cypress.io/getallsessionstorage
+     */
+    getAllSessionStorage(options?: Partial<Loggable>): Chainable<StorageByOrigin>
```
emilyrohrbough, line 840:
```suggestion
    getAllSessionStorage(options?: Partial<Loggable>): Chainable<Array<StorageByOrigin>>
```

cli/types/cypress.d.ts
```diff
@@ -792,7 +819,35 @@ declare namespace Cypress {
...
+     * @see https://on.cypress.io/getalllocalstorage
+     */
+    getAllLocalStorage(options?: Partial<Loggable>): Chainable<StorageByOrigin>
+
+    /**
+     * Clear local storage for all origins.
+     *
+     * @see https://on.cypress.io/clearalllocalstorage
+     */
+    clearAllLocalStorage(options?: Partial<Loggable>): Chainable<null>
```
emilyrohrbough, line 839: If this shouldn't be chained further, isn't the correct return type just null?
https://github.com/cypress-io/cypress-documentation/pull/4876/files#diff-cfb67e9dbdc66c73104dbe980789350ec2bc60eadaf0471868ab1c7937c43902R44

cli/types/cypress.d.ts
```diff
@@ -142,6 +142,33 @@ declare namespace Cypress {
...
+    | boolean
+    | null
+    | { [property: string]: any }
+    | any[]
+
+  type StorableRecord = Record<string, Storable>
+
+  interface OriginStorage {
+    origin: string
+    value: StorableRecord
```
emilyrohrbough, line 166: Shouldn't storable record be an object of Key value pairs where both key & value are type string? https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem

cli/types/cypress.d.ts
chrisbreiding: Even commands that _shouldn't_ be chained further still return chainers and technically _can_ be chained off of. They just yield `null` instead of a subject.

cli/types/cypress.d.ts
chrisbreiding: That's how it's stored by the browser, but we parse it into JavaScript before yielding it to the user, so the value can be anything JSON-stringifiable.

cli/types/cypress.d.ts
emilyrohrbough: We should prob drop the statement that it cant' be chained further then. It's misleading if the types don't reflect the docs.

chrisbreiding: Other commands that yield null say that, so I added it for consistency, but I agree it's misleading and doesn't really add any benefit, so I'll remove it from the docs.

packages/driver/src/cy/commands/sessions/index.ts
```diff
@@ -515,4 +515,8 @@ export default function (Commands, Cypress, cy) {
   })
 
   Cypress.session = sessions
+  // Cypress.session is currently unsupported within the cy.origin() callback.
+  // This also stops us from using it internally, because we replace it with
+  // a proxy that throws an error. We can use this reference internally instead.
+  Cypress._session = sessions
```
emilyrohrbough: This change shouldn't attach more to the Cypress global. Seems like a good time to figure out the best structure for the get/set/clear cookies  & storage methods that live in this file. They don't need to be exposed globally on the sessions API anymore with the new commands.

emilyrohrbough (changes requested):

packages/driver/src/cy/commands/sessions/index.ts
chrisbreiding: I moved the storage-related methods out, so there's no longer a need for `Cypress._session`. I'd rather not blow out the scope of this PR by touching any of the other methods. I'll extract the cookie-related ones when I work on the getAll/clearAll cookie commands.

system-tests/projects/session-and-origin-e2e-specs/cypress/e2e/session/session.cy.js
```diff
@@ -38,102 +38,6 @@ const sessionUser = (name = 'user0', cacheAcrossSpecs = false) => {
   })
 }
 
-describe('cross origin automations', function () {
```
emilyrohrbough, line 41: Are you planning to replace these tests that were deleted?

packages/driver/src/cypress/error_messages.ts
```diff
@@ -1304,7 +1304,7 @@ export default {
         docsUrl: 'https://on.cypress.io/github-issue/20721',
       },
       Cypress_session: {
-        message: `\`Cypress.session.*\` methods are not supported in the ${cmd('switchToDomain')} callback. Consider using them outside of the callback instead.`,
```
emilyrohrbough, line 1307: 🎉

packages/driver/cypress/e2e/e2e/origin/commands/misc.cy.ts
```diff
@@ -219,7 +219,7 @@ it('verifies number of cy commands', () => {
     'focused', 'get', 'contains', 'root', 'shadow', 'within', 'request', 'session', 'screenshot', 'task', 'find', 'filter', 'not',
     'children', 'eq', 'closest', 'first', 'last', 'next', 'nextAll', 'nextUntil', 'parent', 'parents', 'parentsUntil', 'prev',
     'prevAll', 'prevUntil', 'siblings', 'wait', 'title', 'window', 'document', 'viewport', 'server', 'route', 'intercept', 'origin',
-    'mount',
+    'mount', 'getAllLocalStorage', 'clearAllLocalStorage', 'getAllSessionStorage', 'clearAllSessionStorage',
```
emilyrohrbough, line 222: With @BlueWinds changes, several of these are longer added as `commandFns` but as `queryFns` and `ng` was removed. How is this working??

packages/driver/cypress/e2e/e2e/origin/commands/storage.cy.ts
```diff
@@ -0,0 +1,120 @@
+import { findCrossOriginLogs } from '../../../../support/utils'
+
+context('cy.origin storage', { browser: '!webkit' }, () => {
```
emilyrohrbough, line 3: With Origin going GA soon do we need a distinct `origin/commands` spec for this or can we include origin-specific tests in commands/storage moving forward?

packages/driver/cypress/e2e/e2e/origin/commands/storage.cy.ts
```diff
@@ -0,0 +1,120 @@
...
+      })
+
+      cy.clearLocalStorage().should((localStorage) => {
+        expect(localStorage.length).to.equal(0)
+        expect(localStorage.getItem('foo')).to.be.null
+      })
+    })
+  })
+
+  context('#consoleProps', () => {
```
emilyrohrbough, line 91: Should we have tests validating the console props of the new commands we added?

system-tests/projects/session-and-origin-e2e-specs/cypress/e2e/session/session.cy.js
chrisbreiding: The `clearStorage` and `getStorage` functions are well covered by the tests for the new commands. I moved the `setStorage` test into the driver.

packages/driver/cypress/e2e/e2e/origin/commands/misc.cy.ts
chrisbreiding: I think Blue's changes are only in the v12 branch, right? This is targeting `develop`.

packages/driver/cypress/e2e/e2e/origin/commands/storage.cy.ts
chrisbreiding: Yeah, once the flag is gone, I think we should integrate the cy.origin() tests alongside the other related tests, especially where commands are concerned.

packages/driver/cypress/e2e/e2e/origin/commands/storage.cy.ts
chrisbreiding: It's [tested in the non-origin tests](https://github.com/cypress-io/cypress/pull/24859/files/76df9c9617b0991b9cd0d0ab88368715d2c91116#diff-7bb3e31f1185bad7da26bf1a92fea8d7592ed9a5c41d2dfa5300fa1e963a5332R55-R82). I think it's overkill to test it again here.

emilyrohrbough (approved):

packages/driver/src/cy/commands/sessions/origins.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export async function mapOrigins (Cypress: Cypress.Cypress, origins: string | string[]): Promise<string[]> {
+  const getOrigins = Bluebird.map(
+    ([] as string[]).concat(origins), async (origin) => {
+      if (origin === '*') {
+        return await getAllHtmlOrigins(Cypress)
+      }
+
+      if (origin === 'currentOrigin') {
+        return $Location.create(window.location.href).origin
```
mjhenkes: Could this be:
```suggestion
        return window.location.origin
```

chrisbreiding: I believe so. Addressed in https://github.com/cypress-io/cypress/pull/24859/commits/37fee092b371df7221edd63a4d6def7f1957f8e3

mjhenkes (approved):

AtofStryker (approved):

PR #24860, opened 2022-11-28
fix: add v8 snapshot usage to cypress in cypress testing

ryanthemanuel:
### User facing changelog

n/a

### Additional details

This adds support for utilizing the v8 snapshot in cypress in cypress testing. In order to do this, we need to ensure that when the child process is started it is done via the `spawn` command since electron has [special logic](https://github.com/electron/electron/blob/main/patches/node/refactor_alter_child_process_fork_to_use_execute_script_with.patch) in its `fork` command which effectively disables the snapshot. This has the implication of having an additional Cypress icon in the dock on Macs, but that will only happen during development and is a small inconvenience given that we will be able to more accurately simulate prod during cypress in cypress tests.

In addition, as just general cleanup, I cleaned up the `bytenode` logic and was able to make it a dev dependency by bundling it into the electron entry point

### Steps to test

Running the app and launchpad tests is the best way to see this in action

n/a

### PR Tasks


packages/data-context/src/data/ProjectConfigIpc.ts
```diff
@@ -310,6 +319,17 @@ export class ProjectConfigIpc extends EventEmitter {
...
+      configProcessArgs.push('--no-sandbox')
+    }
+
+    if (process.env.CYPRESS_INTERNAL_E2E_TESTING_SELF_PARENT_PROJECT) {
+      return spawn(process.execPath, ['--entryPoint', CHILD_PROCESS_FILE_PATH, ...configProcessArgs], {
+        ...childOptions,
+        stdio: ['pipe', 'pipe', 'pipe', 'ipc'],
+      })
+    }
+
```
ryanthemanuel:
```suggestion
    if (process.env.CYPRESS_INTERNAL_E2E_TESTING_SELF_PARENT_PROJECT) {
      if (isSandboxNeeded()) {
        configProcessArgs.push('--no-sandbox')
      }

      return spawn(process.execPath, ['--entryPoint', CHILD_PROCESS_FILE_PATH, ...configProcessArgs], {
        ...childOptions,
        stdio: ['pipe', 'pipe', 'pipe', 'ipc'],
      })
    }

```

packages/config/package.json
```diff
@@ -21,7 +21,7 @@
     "@babel/parser": "^7",
     "@babel/plugin-syntax-typescript": "^7",
     "@babel/plugin-transform-typescript": "^7",
-    "@babel/traverse": "^7",
+    "@babel/traverse": "7.15.4",
```
ryanthemanuel, line 24: Updated this to match the rest of the repo.

packages/server/index.js
```diff
@@ -10,10 +15,16 @@ const startCypress = async () => {
 
     await require('./start-cypress')
   } catch (error) {
-    // eslint-disable-next-line no-console
+  // eslint-disable-next-line no-console
```
ryanthemanuel:
```suggestion
    // eslint-disable-next-line no-console
```

AtofStryker (approved):

scripts/after-pack-hook.js
```diff
@@ -10,67 +10,74 @@ const { buildEntryPointAndCleanup } = require('./binary/binary-cleanup')
...
-  })
-
-  const buildSubfoldersPerPlatform = {
-    darwin: join('Cypress.app', 'Contents', 'Resources', 'app'),
-    linux: join('resources', 'app'),
-    win32: join('resources', 'app'), // TODO check this path
-  }
-  const buildSubfolder = buildSubfoldersPerPlatform[os.platform()]
-  const outputFolder = join(params.appOutDir, buildSubfolder)
+  try {
```
flotwig, line 13: I wouldn't have expected test changes to affect build scripts, why are these changes needed?

ryanthemanuel: Oops forgot to comment on the description for why this was changed. I added a couple of changes to be able to move `bytenode` from a main dependency to a dev dependency and just generally cleaned some of this up. No functional changes, but it was something that was nagging at me from my last PR that I wanted to clean up while I had time.

flotwig (approved):

scripts/after-pack-hook.js
ryanthemanuel: Added this to the description.

packages/data-context/src/data/ProjectConfigIpc.ts
```diff
@@ -310,6 +319,17 @@ export class ProjectConfigIpc extends EventEmitter {
       debug(`no typescript found, just use regular Node.js`)
     }
 
+    if (process.env.CYPRESS_INTERNAL_E2E_TESTING_SELF_PARENT_PROJECT) {
```
lmiller1990, line 322: Bit of a shame we need to ship test code to prod :'(

This is "the change" that lets v8 snapshots run in cy-in-cy?

lmiller1990: Is there anyway to verify it now uses snapshots in the tests? How do we know it works (and catch it if stops?)

lmiller1990 (approved):
Great, I have a question, but no blocker.

packages/data-context/src/data/ProjectConfigIpc.ts
ryanthemanuel: Yeah it's unfortunate, but I'm not sure of a clean way to do it otherwise with how everything is structured.

And, I like your idea. I added a check to ensure that we fail in cypress in cypress if the snapshot is unavailable.

PR #24863, opened 2022-11-28
fix: Clean up command attributes after no longer needed

BlueWinds:
### User facing changelog
No user-facing changes.

### Additional details
This PR adds a "cleanSubjects" step to the command_queue, which, every time a command resolves, does two things:
1. Looks through `cy.state('subjects')` and removes any subject chains that we know aren't needed by future commands. This should let us clean up the closures for executed queries, and drop unnecessary references to DOM elements that may no longer be attached to the page for commands.
2. Cleans up the attributes of passed commands, removing a bunch of data that we know we'll no longer display to the user or reference in the future.

### Steps to test
Existing tests should cover this fairly thoroughly; if it broke any features of the command log, they'll let us know.

If you wish to verify manually, the best place would be visually in the runner - examine the command log, verify that console props and snapshots continue to work for commands and tests that have already resolved.

No change in user experience. Possibly less memory used in long tests with many commands executing.

### PR Tasks


packages/driver/src/cypress/command_queue.ts
```diff
@@ -557,4 +559,27 @@ export class CommandQueue extends Queue<$Command> {
...
+
+  // This function iterates through all upcoming commands in the queue, then
+  // discards the subject chain for every chainer that can't be referenced
+  // in the future (eg, no upcoming commands belong to the same chain).
+
+  // This is safe because aliases (which might be referenced later) are stored
+  // separately, in state('aliases'), and any subjects that "flow upwards" (eg.
+  // the subject of a chain inside a .then() command) have already replaced
+  // the subject of their parent chainer by the time this is called.
+  cleanSubjects () {
```
emilyrohrbough, line 571: Is this safe for commands that execute callbacks?

packages/driver/src/cypress/command_queue.ts
```diff
@@ -557,4 +559,27 @@ export class CommandQueue extends Queue<$Command> {
...
+    const stillNeeded = this.queueables.slice(this.index).map((c) => c.get('chainerId'))
+
+    this.queueables.slice(0, this.index).forEach((command) => {
+      // Once a command has resolved, and its chainer is no longer referenced
+      // by future commands, we can throw away the reference to the function
+      // and its subject to free memory.
+      if (command.get('subject') && stillNeeded.indexOf(command.get('chainerId')) === -1) {
+        command.set({ fn: null, subject: null, queryFn: null })
```
emilyrohrbough, line 579: why does the command store a reference to the subject at all? Is this being used?

emilyrohrbough: Are we able to drop userInvocationStack as well? At this point we don't necessarily need the stack details since the command has finished.

packages/driver/src/cypress/command_queue.ts
BlueWinds: Yep. Consider a test like this.

```
cy
  .get('button') // chainer-1
  .then(btn => { // chainer-1
    cy
      .find('form') // chainer-2
      .click() // chainer-2
  })
  .dblclick() // chainer-1
```

So it ends up playing out like this:

1. Outer test function executes, resulting in command queue `[get, then, dblclick]` - the .then() callback hasn't been called yet, so .find and .click aren't in the queue.
2. .get resolves
  a. cleanSubjects() leaves the chainer-1 subject intact (.then and .dblclick are in the same chainer).
3. .then() executes the callback.
  a. Its commands are inserted into the queue, leaving it looking like `[get, then, find, click, dblclick]`.
  b. A link is created: the subject for chainer-2 is linked with the subject for chainer-1 (because this is how .then() works).
  c. .then resolves, cleanSubjects again leaves chainer-1 intact.
4. .find() resolves.
  a. Because of the link, chainer-1's subject (the button) is overwritten with the result of .find() (the form).
  b. cleanSubjects leaves chainer-1 (.dblclick references it) and chainer-2 intact (.click() references it)
5. .click() resolves.
  a. Because of the link, chainer-1's subject is again overwritten.
  b. cleanSubject leaves chainer-1 intact, but discards chainer-2 (no future commands reference it, the only future command is .dblclick)
6. .dblclick() resolves.
  a. cleanSubject discards chainer-2 (there are no future commands remaining)

A touch convoluted, because "a flat command queue" isn't the right abstraction for Cypress anymore - it really needs to be a nested structure. But that's work for another year.

emilyrohrbough: what about `.within()`?

packages/driver/src/cypress/command_queue.ts
BlueWinds: The subject is used in three places:
- `src/cypress/log.ts` - The default `consoleProps` function reads the subject from command.get('subject') to display it as 'Yielded'. I think this could probably be redone. How often is the default consoleProps not overridden? And when it isn't, is there a better place to get this? Not sure.
- `src/cy/logGroup.ts` - in forwardYieldedSubject. I think this could probably be rewritten / removed, but haven't spent time looking at logGroups to know how they working.
- `src/cypress/command_queue.ts` - The command queue itself. In `commandRunningFailed`, we look at `current.get('prev').get('subject')` to set the consoleProps. This could be read from cy.state('subjects') rather than the command's individual subject.
- One of our tests, `cypress/e2e/commands/assertions.cy.js`. Could be rewritten easily, 100% sure it's not needed here.

So quick census says "we probably don't need to anymore, but I'm uncomfortable trying to remove it just before release, would need more investigation."

packages/driver/src/cypress/command_queue.ts
BlueWinds: The `within` subject is stored in two places: cy.state('withinSubject') and in the closure of each individual command that uses it. Within doesn't really interact with cy.state('subjects') at all, and it already has its own cleanup mechanism inside the .within implementation.

(these are all good questions, I'm double-checking my work as I write these answers to be sure)

packages/driver/src/cypress/command_queue.ts
BlueWinds: I was looking at `userInvocationStack` as well, but got scared off by the number of places it's used and my unfamiliarity with them. It doesn't look bad, but our stack traces are pretty fragile and not well tested. Any changes there I'd want to spend time digging through to understand more carefully.

The subject, fn and queryFn I understand fairly well, so feel confident messing with on short notice.

packages/driver/src/cypress/command_queue.ts
emilyrohrbough: my other thought --  if this runs after each command, this might be 'heavy' for tests with several tests.  is it possible when we break the subject chain if this can be cleaned up there? Maybe that's leaking implementation?

BlueWinds:
Did some very unscientific benchmarking with Real World App, and this doesn't appear to have much effect on memory usage - at the very least, any change is much smaller than the variability between runs. So this is definitely no golden bullet, but I can't imagine it *hurts* memory usage to clean up unneeded references after commands resolve.

lmiller1990 (commented):
Is there any way to actually verify if this makes a noticeable difference? Or the amount of work to measure the memory footprint is more effort than it's worth?

Regarding 

> discards the subject chain for every chainer that can't be referenced

What kind of things are discarded? What kind of things are discarded that might take a lot of memory?

chrisbreiding (approved):

emilyrohrbough:
@BlueWinds If we didn't find significant memory reduction, could we rename this title to something like "clean up command attributes" so when committed, it'll prevent users from believing we made an improvement?

BlueWinds:
> @BlueWinds If we didn't find significant memory reduction, could we rename this title to something like "clean up command attributes" so when committed, it'll prevent users from believing we made an improvement?

Good call, done.

> Is there any way to actually verify if this makes a noticeable difference? Or the amount of work to measure the memory footprint is more effort than it's worth?

I spent some time doing benchmarking this morning, and found that there's basically no effect after all. My benchmark was the following:

```
it.only('long test', () => {

  for (let j = 0; j < 10; j++) {
    let a = cy.get('button').first()

    for (let i = 0; i < 30; i++) {
      a = a.parent().children()
    }
  }

  cy.wait(100000)
})
```

This creates 10 chains, each with 62 commands in them, simulating a very long test (that does little actual work). The 100s wait at the end is so that I could execute this spec in `run` mode without having it exit, giving me time to run:

`ps aux | grep Cypress | grep -v grep | awk '{s+=$6} END {print s}'`

This command sums up the RSS - Resident Set Size. This isn't *exactly* "memory used", but it's close enough of an approximation for this case. https://linuxconfig.org/ps-output-difference-between-vsz-vs-rss-memory-usage explains RSS if you're interested in the details.

I ran it a couple of times, and watched the RSS as it fluctuated (it goes up and down over time with garbage collection):

```
| RSS     | Low     | High    |
| Without | 2930084 | 3168148 |
| With    | 2926904 | 3115992 |
```

These values are in kb, but don't take them literally - summing RSS over multiple processes means we're counting things multiple times. Only the *difference* in numbers is meaningful here, not the absolute numbers.

So it saves something like 4mb of memory on the low end, and 9mb on the high end. Not a massive change, but it is consistently present!

> What kind of things are discarded? What kind of things are discarded that might take a lot of memory?

The "subject chain" is the thing discarded - this is an array shaped like `[initialValue, queryFn, queryFn...]`. The hope - which hasn't borne out super strongly in practice - was that discarding the query functions would let them get cleaned up. The query functions maintain some state in their closures - sometimes references to the DOM, always a reference to a log message, various other smaller pieces of state.

In retrospect, it's clear that discarding these won't have a huge impact - all of this is still referenced elsewhere in Cypress (eg, each $Command keeps a `logs` array, so they'll remain in memory even if the queryFn isn't pointing to them).

I do think it's a step towards potential future wins though - keeping fewer references to things around makes fewer references we need to remove in the future to GC things.

lmiller1990:
>  sometimes references to the DOM

This was the main one I was expecting - good to clear references to DOM elements where we can. Even a few mb reduction is a solid win imo, nice! I haven't done much benchmarking but my feeling is we suffer death from a thousand cuts - we just need to keep chipping away at things, bit by bit.

lmiller1990 (approved):

packages/driver/src/cypress/command_queue.ts
emilyrohrbough: @BlueWinds Can we log issues to clean these up? I would have though these would have all been removed with the detached dom work & would be nice to remove these stale reference if we can! heavy bloat for large doms.

emilyrohrbough (approved):

emilyrohrbough:
Still wondering about the clean up impact when lopping over all commands in an extremely large test but seem like an okay change.

packages/driver/src/cypress/command_queue.ts
BlueWinds: Logged https://github.com/cypress-io/cypress/issues/24903 to cover that, good idea. Also created https://github.com/cypress-io/cypress/issues/24904 to cover some other work in the area I'd like to do.

BlueWinds:
> Still wondering about the clean up impact when lopping over all commands in an extremely large test but seem like an okay change.

I didn't record the ms for running tests (should have, didn't think of it), but the benchmark I ran above was consistently 28-29s across runs, with and without the change. That's a test that has ~600 commands, and I didn't notice an impact - it's probably there if we measure carefully in milliseconds.

Ultimately this is just a bandaid over the work I'd like to do in https://github.com/cypress-io/cypress/issues/24903.

PR #24864, opened 2022-11-28
feat: Handle special characters in Create from Component filenames

### LINKED ISSUE #23492: "Create from Component" with special characters in component filename can scaffold incorrect spec name and non-working import/mount
### Current behavior

Currently if I choose the catchall route component named `[...all].vue` we scaffold the spec as normal, but with the name `l.spec.ts`, and our import statements are (pretty naturally) non-functional because the filename happens to include a `[`.

Later I noticed this also happens with filenames like `new-component.vue` where we generate `import new-component from './components/new-component.vue'` and `cy.mount(new-component)`, which are invalid JavaScript. Hyphens in filenames seem very common so we should handle them gracefully.

Screen Shot 2022-08-22 at 10 47 38 AM
Screen Shot 2022-08-22 at 10 31 31 AM
Screen Shot 2022-08-22 at 10 33 03 AM

### Desired behavior

For the internal import name, Cypress should

- strip all characters from the file name that are not valid in JS variable names before generating `import ${filename}`, or
- name the import `component` or something predictable so that the filename will never be used as JS, and we don't have to think it about it.

This version of the spec is valid and runs:

```ts
import component from './pages/[...all].vue'

describe('<[...all] />', () => {
  it('renders', () => {
    // see: https://test-utils.vuejs.org/guide/
    cy.mount(component)
  })
})
```
Screen Shot 2022-08-22 at 10 49 05 AM

For the filename, `[...all].spec.ts` is a valid spec name, and we should preserve the literal name of the component. Seems possible the extra periods are running afoul of some regex we use for the extensions maybe. 

We have lots of tests around spec filenames displaying correctly in the Specs List with various characters and languages (e.g. `d~e(f)g.spec.js`), we could use some component files with those same non-English word names to test generating specs.

In general I think there are filename conventions using special characters beyond this specific case, so it would be good to handle them predictably, even if it's not possible to do it elegantly right away.

### Test code to reproduce

In the app package, open Cypress component testing and choose "New Spec" then "Create from Component", and attempt to create a spec from `[...all].vue`.

### Cypress Version

develop

### Node version

16.14.2

### Operating System

MacOS

### Debug Logs

_No response_

### Other

_No response_

mike-plummer:
- Closes #23492 

### User facing changelog
* Fixes "Create Spec from Component" to better handle special characters in filenames

### Additional details
In "Generate from Component", handle situations where source filename contains special characters

| Condition | Source File | Spec Filename | Component Name |
|---|---|---|---|
| Braces | `[...MyComponent].vue` | `[...MyComponent].cy.js` | `MyComponent` |
| Parens | `My(Component).js` | `My(Component).cy.js` | `MyComponent` |
| Periods | `my.component.js` | `my.component.cy.js` | `my` |
| Hypens | `my-component.js` | `my-component.cy.js` | `mycomponent` |
| Dollar | `$MyComponent.ts` | `$MyComponent.cy.ts` | `$MyComponent` |

### Steps to test
1. Scaffold a basic Vue project
2. Create the files in the table above in the project (can be left empty)
3. Open project in Cypress, use "Create Spec from Component" feature to create spec for each file
4. Validate generated spec file uses expected filename
5. Validate content of spec file uses expected component name

Currently, these patterns either result in malformed partial names or try to use illegal characters in the generated spec file

For example,
`[...MyComponent].vue` => `[.cy.ts` with an invalid component name of `[...MyComponent]`
`my-component.js` => `my-component.cy.js` with an invalid component name of `my-component`

### PR Tasks

### DIFF
--- a/packages/data-context/test/unit/codegen/spec-options.spec.ts
+++ b/packages/data-context/test/unit/codegen/spec-options.spec.ts
@@ -246,10 +246,12 @@ describe('spec-options', () => {
       context('file name contains special characters', async () => {
         [
           { condition: 'braces', fileName: '[...MyComponent].vue', expectedFileName: '[...MyComponent].cy.js', expectedComponentName: 'MyComponent' },
-          { condition: 'hyphens', fileName: 'my-component.vue', expectedFileName: 'my-component.cy.js', expectedComponentName: 'mycomponent' },
+          { condition: 'hyphens', fileName: 'my-component.vue', expectedFileName: 'my-component.cy.js', expectedComponentName: 'MyComponent' },
           { condition: 'parentheses', fileName: 'My(Component).js', expectedFileName: 'My(Component).cy.js', expectedComponentName: 'MyComponent' },
-          { condition: 'period-separated', fileName: 'my.component.js', expectedFileName: 'my.component.cy.js', expectedComponentName: 'my' },
+          { condition: 'period-separated', fileName: 'my.component.js', expectedFileName: 'my.component.cy.js', expectedComponentName: 'MyComponent' },
           { condition: 'dollar', fileName: '$MyComponent.js', expectedFileName: '$MyComponent.cy.js', expectedComponentName: '$MyComponent' },
+          { condition: 'underscores', fileName: 'My_Component.js', expectedFileName: 'My_Component.cy.js', expectedComponentName: 'My_Component' },
+          { condition: 'mixed period- and hypen-delimited', fileName: 'about-us.component.js', expectedFileName: 'about-us.component.cy.js', expectedComponentName: 'AboutUsComponent' },
         ].forEach(({ condition, fileName, expectedFileName, expectedComponentName }) => {
           it(`generates options for ${condition}`, async () => {
             const testSpecOptions = new SpecOptions({
--- a/packages/data-context/test/unit/codegen/spec-options.spec.ts
+++ b/packages/data-context/test/unit/codegen/spec-options.spec.ts
@@ -242,6 +242,35 @@ describe('spec-options', () => {
         expect(result.codeGenType).to.eq('e2e')
         expect(result.fileName).to.eq(`TestName.foo.bar-copy-2.js`)
       })
+
+      context('file name contains special characters', async () => {
+        [
+          { condition: 'braces', fileName: '[...MyComponent].vue', expectedFileName: '[...MyComponent].cy.js', expectedComponentName: 'MyComponent' },
+          { condition: 'hyphens', fileName: 'my-component.vue', expectedFileName: 'my-component.cy.js', expectedComponentName: 'mycomponent' },
+          { condition: 'parentheses', fileName: 'My(Component).js', expectedFileName: 'My(Component).cy.js', expectedComponentName: 'MyComponent' },
+          { condition: 'period-separated', fileName: 'my.component.js', expectedFileName: 'my.component.cy.js', expectedComponentName: 'my' },
+          { condition: 'dollar', fileName: '$MyComponent.js', expectedFileName: '$MyComponent.cy.js', expectedComponentName: '$MyComponent' },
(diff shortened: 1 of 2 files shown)


packages/data-context/test/unit/codegen/spec-options.spec.ts
```diff
@@ -242,6 +242,35 @@ describe('spec-options', () => {
         expect(result.codeGenType).to.eq('e2e')
         expect(result.fileName).to.eq(`TestName.foo.bar-copy-2.js`)
       })
+
+      context('file name contains special characters', async () => {
+        [
+          { condition: 'braces', fileName: '[...MyComponent].vue', expectedFileName: '[...MyComponent].cy.js', expectedComponentName: 'MyComponent' },
+          { condition: 'hyphens', fileName: 'my-component.vue', expectedFileName: 'my-component.cy.js', expectedComponentName: 'mycomponent' },
+          { condition: 'parentheses', fileName: 'My(Component).js', expectedFileName: 'My(Component).cy.js', expectedComponentName: 'MyComponent' },
+          { condition: 'period-separated', fileName: 'my.component.js', expectedFileName: 'my.component.cy.js', expectedComponentName: 'my' },
```
ZachJW34: Angular has the `my.component.ts` convention, I'd expect the output to be `MyComponent` rather than 	my`. Same for the `my-component.vue` where I'd expect it to be pascal case.

ZachJW34 (commented):
Code works great, just have a comment on the output of some of the component names.

packages/data-context/test/unit/codegen/spec-options.spec.ts
mike-plummer: @ZachJW34 This may just be my React background talking - I would want a generated spec for `ContactUs.component.tsx` to result in code like `<ContactUs />` rather than `<ContactUsComponent />`. Makes more sense in a JSX context (to me at least), but I can definitely see where it would make more sense to tack on the `...Component` for Angular and other non-JSX situations.

We don't support this flow for Angular yet so this is technically a non-issue and could be tackled when we get to implementing it, but I hesitate to do too much language-specific logic in here since it would become a mess pretty quickly. I could see just going with including the `Component` suffix and calling it good since it would be a simple rename for the React devs out there who are opinionated like me

mike-plummer: I'll also look into PascalCase for period- and hyphen-delimited names

ZachJW34 (approved):
👍

astone123 (approved):
Looks good 👍🏻 this shouldn't be as much of a problem for create from React since we'll be explicitly naming the component the way that it's named in the file, so it will be following React naming conventions

packages/data-context/test/unit/codegen/spec-options.spec.ts
```diff
@@ -242,6 +242,37 @@ describe('spec-options', () => {
         expect(result.codeGenType).to.eq('e2e')
         expect(result.fileName).to.eq(`TestName.foo.bar-copy-2.js`)
       })
+
+      context('file name contains special characters', async () => {
+        [
+          { condition: 'braces', fileName: '[...MyComponent].vue', expectedFileName: '[...MyComponent].cy.js', expectedComponentName: 'MyComponent' },
```
lmiller1990, line 248: Just to clarify, this PR only fixes the <name_of_component> part?

```js
import <name_of_component> from "'./[...blah].vue"
```

Looks good!

lmiller1990 (approved):

packages/data-context/test/unit/codegen/spec-options.spec.ts
mike-plummer: Correct, just improving the conversion of the filename into a logical, legal JS var name. Expected filename conditions in this test are just there to ensure no regression was/is introduced

Issue #24444, opened 2022-10-28
[IATR](M0) Failures page test component


warrensplayer:
Create a component to represent the test within a spec that failed during a recorded run. 

**NOTE:** M0 issues will be committed against a feature branch and not released 

### UI Design
* Row that contains the test title fragments separated by `>` signs 

<img width="790" alt="image" src="https://user-images.githubusercontent.com/1702361/198610414-771b039c-ed81-4986-bd40-40f38a43e705.png">

_Wireframe from Whimsical. Final design is in progress._

### Inputs
* Test
  *   Title - with fragments that represent the parent context and describe descriptions for the test within the spec

### Interactivity
None.  

#### Resources
* [Clickup](https://app.clickup.com/t/18033298/PM-2546)
* [Whimsical](https://whimsical.com/q3-2022-solutions-CB1g7VnZeHsrVdz4gurF2o@AhRp651J2s3PQev2xg2uwbceF) - See V0 section

warrensplayer:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Component-Testing-606e43bbf1b3b20013c142c1/issues/cypress-io/cypress/24444?planning-poker) with Zenhub @rockindahizzy @lmiller1990 @ZachJW34 @marktnoonan @amehta265 @astone123

PR #24877, opened 2022-11-29
feat: IATR-M0 Spec and Test Component

### LINKED ISSUE #24443: [IATR](M0) Failures page spec component
Create a component to represent a spec file that has test failures during a cloud recorded run.

**NOTE:** M0 issues will be committed against a feature branch and not released 

### UI Design

* Container that contain a header and a body that will contain a list of [Test](https://github.com/cypress-io/cypress/issues/24444) components
* Header will contain spec file name, relative path to test within the project, and a "Run Spec" button
* "Run Spec" button
  * disabled if the spec file is not found in the local project, this will be indicated with a prop for `foundInLocalProject` being set to true
  * if the button is enabled and clicked by the user, the user will be taken to the runner for that spec and the entire spec will be run 

image

_Wireframe from Whimsical. Final design is in progress. This component is just the container and header._

### Inputs
* Spec {object}
  * file name {string}
  * relative path to file within the project {string}
  * Collection of failed [tests](https://github.com/cypress-io/cypress/issues/24444) {array{object}}
  * _The cloud may return more data in a spec that is not being currently used._
* foundInLocalProject {boolean} - used to enable the 'Run Spec' button only if

#### Resources
* [Clickup](https://app.clickup.com/t/18033298/PM-2546)
* [Whimsical](https://whimsical.com/q3-2022-solutions-CB1g7VnZeHsrVdz4gurF2o@AhRp651J2s3PQev2xg2uwbceF) - See V0 section

amehta265:
- Closes #24443 and #24444 

### User facing changelog

This is the spec and test component for the new [debug page](https://github.com/cypress-io/cypress/issues/24441)
The design for this page can be found [here](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/QkCp2F69X0qyLo7qnYQ67e/Cypress-App?node-id=21413%3A202471&t=jaeAWR6n9srpSsKo-0)

### Additional details

The spec component refers to the top section of the `debug spec item` (information about a single failed spec file and the tests that have failed within it) which contains the relative path of the spec and a "Run Failures" button. The test component is the bottom section which contains a list of failed tests in the form of a title and description.
- `DebugSpec.vue` renders an entire `debug spec item`. There is no use of `gql` props here as the tentative implementation would involve a `gql` query in a top level component which would pass props down to the `DebugSpec` component. 
- `DebugSpec.vue` uses `DebugFailedTest.vue` to render the test component section of the `debug spec item`

### Steps to test

This issue has 2 component tests. Checkout to this branch to run the tests
Both DebugSpec.vue and DebugFailedTest.vue exist in the packages/app/ folder. To run these tests:

1. Change directory to packages/app
2. Run `yarn dev`
3. Navigate to Component Testing and run `DebugSpec.cy.tsx` and `DebugFailedTest.cy.tsx`

Screenshot 2022-11-29 at 10 27 17 AM

Screenshot 2022-11-29 at 10 27 50 AM

### PR Tasks

### DIFF
--- a/packages/app/src/debug/DebugFailedTest.cy.tsx
+++ b/packages/app/src/debug/DebugFailedTest.cy.tsx
@@ -16,7 +16,7 @@ describe('<DebugFailedTest/>', () => {
     })
 
     testResult.titleParts.forEach((title, index) => {
-      cy.findByTestId(`titleParts-${index}`).should('have.text', `${title } `)
+      cy.findByTestId(`titleParts-${index}`).should('have.text', ` ${title }`)
     })
   })
 
@@ -32,7 +32,9 @@ describe('<DebugFailedTest/>', () => {
 
     cy.findByTestId('test-row').children().should('have.length', 5).should('be.visible')
     multipleTitleParts.titleParts.forEach((title, index) => {
-      cy.findByTestId(`titleParts-${index}`).should('have.text', `${title } `)
+      cy.findByTestId(`titleParts-${index}`).should('have.text', ` ${title }`)
     })
+
+    cy.percySnapshot()
   })
 })
--- a/packages/app/src/debug/DebugFailedTest.cy.tsx
+++ b/packages/app/src/debug/DebugFailedTest.cy.tsx
@@ -1,0 +1,9 @@
+import DebugFailedTest from './DebugFailedTest.vue'
+
+describe('<DebugSpec/>', () => {
+  it('mounts correctly', () => {
+    cy.mount(() => (
+      <DebugFailedTest failedTest="Login should redirect unauthenticated user to signin page" />
+    ))
+  })
+})
--- a/packages/app/src/debug/DebugFailedTest.cy.tsx
+++ b/packages/app/src/debug/DebugFailedTest.cy.tsx
@@ -11,12 +11,12 @@ describe('<DebugFailedTest/>', () => {
       <DebugFailedTest failedTestResult={testResult} />
     ))
 
-    cy.findByTestId('test-row').children().should('have.length', 2).then((ele) => {
-      expect(ele).to.have.css('color', 'rgb(90, 95, 122)')
-    })
+    cy.findByTestId('test-row').children()
+    .should('have.length', 2)
+    .and('have.css', 'color', 'rgb(90, 95, 122)')
 
     testResult.titleParts.forEach((title, index) => {
-      cy.findByTestId(`titleParts-${index}`).should('have.text', ` ${title }`)
+      cy.findByTestId(`titleParts-${index}`).should('have.text', `${title}`)
     })
   })
 
@@ -32,7 +32,7 @@ describe('<DebugFailedTest/>', () => {
 
     cy.findByTestId('test-row').children().should('have.length', 5).should('be.visible')
     multipleTitleParts.titleParts.forEach((title, index) => {
-      cy.findByTestId(`titleParts-${index}`).should('have.text', ` ${title }`)
+      cy.findByTestId(`titleParts-${index}`).should('have.text', `${title}`)
     })
 
     cy.percySnapshot()
--- a/packages/app/src/debug/DebugFailedTest.cy.tsx
+++ b/packages/app/src/debug/DebugFailedTest.cy.tsx
@@ -2,8 +2,19 @@ import DebugFailedTest from './DebugFailedTest.vue'
 
 describe('<DebugSpec/>', () => {
   it('mounts correctly', () => {
+    const failedTest: string[] = ['App: Runs', 'should fetch newer runs and maintain them when navigating']
+
     cy.mount(() => (
-      <DebugFailedTest failedTest="Login should redirect unauthenticated user to signin page" />
+      <DebugFailedTest failedTest={failedTest} />
     ))
+
+    cy.findByTestId('test-row').children().should('have.length', 3)
+    cy.findByTestId('test-title')
+    .should('have.text', 'App: Runs')
+    .should('have.css', 'color', 'rgb(90, 95, 122)')
+
+    cy.findByTestId('test-description')
+    .should('have.text', 'should fetch newer runs and maintain them when navigating')
+    .should('have.css', 'color', 'rgb(90, 95, 122)')
   })
 })
--- a/packages/app/src/debug/DebugFailedTest.cy.tsx
+++ b/packages/app/src/debug/DebugFailedTest.cy.tsx
@@ -1,20 +1,38 @@
 import DebugFailedTest from './DebugFailedTest.vue'
 
-describe('<DebugSpec/>', () => {
+describe('<DebugFailedTest/>', () => {
   it('mounts correctly', () => {
-    const failedTest: string[] = ['App: Runs', 'should fetch newer runs and maintain them when navigating']
+    const testResult = {
+      id: '676df87878',
+      titleParts: ['Login', 'Should redirect unauthenticated user to signin page'],
+    }
 
     cy.mount(() => (
-      <DebugFailedTest failedTest={failedTest} />
+      <DebugFailedTest failedTestResult={testResult} />
     ))
 
-    cy.findByTestId('test-row').children().should('have.length', 3)
-    cy.findByTestId('test-title')
-    .should('have.text', 'App: Runs')
-    .should('have.css', 'color', 'rgb(90, 95, 122)')
+    cy.findByTestId('test-row').children().should('have.length', 2).then((ele) => {
+      expect(ele).to.have.css('color', 'rgb(90, 95, 122)')
+    })
 
-    cy.findByTestId('test-description')
-    .should('have.text', 'should fetch newer runs and maintain them when navigating')
-    .should('have.css', 'color', 'rgb(90, 95, 122)')
+    testResult.titleParts.forEach((title, index) => {
+      cy.findByTestId(`titleParts-${index}`).should('have.text', `${title } `)
+    })
+  })
+
+  it('contains multiple titleParts segments', () => {
+    const multipleTitleParts = {
+      id: '676df87878',
+      titleParts: ['Login', 'Describe', 'it', 'context', 'Should redirect unauthenticated user to signin page'],
+    }
+
+    cy.mount(() => (
+      <DebugFailedTest failedTestResult={multipleTitleParts} />
+    ))
+
+    cy.findByTestId('test-row').children().should('have.length', 5).should('be.visible')
+    multipleTitleParts.titleParts.forEach((title, index) => {
+      cy.findByTestId(`titleParts-${index}`).should('have.text', `${title } `)
+    })
   })
 })
--- a/packages/app/src/debug/DebugFailedTest.cy.tsx
+++ b/packages/app/src/debug/DebugFailedTest.cy.tsx
@@ -11,9 +11,7 @@ describe('<DebugFailedTest/>', () => {
       <DebugFailedTest failedTestResult={testResult} />
     ))
 
-    cy.findByTestId('test-row').children()
-    .should('have.length', 2)
-    .and('have.css', 'color', 'rgb(90, 95, 122)')
+    cy.findByTestId('test-row').children().should('have.length', 2)
 
     testResult.titleParts.forEach((title, index) => {
       cy.findByTestId(`titleParts-${index}`).should('have.text', `${title}`)
--- a/packages/app/src/debug/DebugFailedTest.vue
+++ b/packages/app/src/debug/DebugFailedTest.vue
@@ -1,0 +1,29 @@
+<template>
+  <li
+    data-cy="test-row"
+    class="flex flex-row items-center py-12px pr-12px pl-16px gap-x-2.5 non-italic text-base text-gray-700 font-normal border-b-gray-100 border-b-1px"
+  >
+    <span
+      data-cy="test-title"
+    >
+      {{ props.failedTest.split(' ')[0] }}
+    </span>
+    <IconChevronRightSmall
+      size="8"
+      stroke-color="gray-200"
+      fill-color="gray-200"
+    />
+    <span>
+      {{ props.failedTest.split(' ').slice(1).join(' ') }}
+    </span>
+  </li>
+</template>
+<script lang="ts" setup>
+import { IconChevronRightSmall } from '@cypress-design/vue-icon'
+
+const props = defineProps<{
+  failedTest: string
+}>()
+
+// figure out proper caret sign placement
+</script>
--- a/packages/app/src/debug/DebugFailedTest.vue
+++ b/packages/app/src/debug/DebugFailedTest.vue
@@ -6,24 +6,28 @@
     <span
       data-cy="test-title"
     >
-      {{ props.failedTest.split(' ')[0] }}
+      {{ testData[0] }}
     </span>
     <IconChevronRightSmall
       size="8"
       stroke-color="gray-200"
       fill-color="gray-200"
     />
-    <span>
-      {{ props.failedTest.split(' ').slice(1).join(' ') }}
+    <span
+      data-cy="test-description"
+    >
+      {{ testData[1] }}
     </span>
   </li>
 </template>
 <script lang="ts" setup>
+import { computed } from 'vue'
 import { IconChevronRightSmall } from '@cypress-design/vue-icon'
 
 const props = defineProps<{
-  failedTest: string
+  failedTest: string[]
 }>()
 
-// figure out proper caret sign placement
+const testData = computed(() => props.failedTest)
+
 </script>
--- a/packages/app/src/debug/DebugFailedTest.vue
+++ b/packages/app/src/debug/DebugFailedTest.vue
@@ -3,21 +3,21 @@
     data-cy="test-row"
     class="flex flex-row items-center gap-x-2.5 non-italic text-base text-gray-700 font-normal"
   >
-    <li
+    <span
       v-for="(titlePart, index) in props.failedTestResult.titleParts"
       :key="`${titlePart}-${index}`"
-      class="flex flew-row items-center gap-x-2.5"
+      class="flex flex-row items-center gap-x-2.5"
       :data-cy="`titleParts-${index}`"
     >
-      {{ titlePart }}
       <IconChevronRightSmall
-        v-if="(index != props.failedTestResult.titleParts.length - 1)"
+        v-if="index != 0"
         data-cy="right-chevron"
         size="8"
         stroke-color="gray-200"
         fill-color="gray-200"
       />
-    </li>
+      {{ titlePart }}
+    </span>
   </div>
 </template>
 <script lang="ts" setup>
--- a/packages/app/src/debug/DebugFailedTest.vue
+++ b/packages/app/src/debug/DebugFailedTest.vue
@@ -1,33 +1,31 @@
 <template>
-  <li
+  <div
     data-cy="test-row"
-    class="flex flex-row items-center py-12px pr-12px pl-16px gap-x-2.5 non-italic text-base text-gray-700 font-normal border-b-gray-100 border-b-1px"
+    class="flex flex-row items-center gap-x-2.5 non-italic text-base text-gray-700 font-normal"
   >
-    <span
-      data-cy="test-title"
+    <li
+      v-for="(titlePart, index) in props.failedTestResult.titleParts"
+      :key="`${titlePart}-${index}`"
+      class="flex flew-row items-center gap-x-2.5"
+      :data-cy="`titleParts-${index}`"
     >
-      {{ testData[0] }}
-    </span>
-    <IconChevronRightSmall
-      size="8"
-      stroke-color="gray-200"
-      fill-color="gray-200"
-    />
-    <span
-      data-cy="test-description"
-    >
-      {{ testData[1] }}
-    </span>
-  </li>
+      {{ titlePart }}
+      <IconChevronRightSmall
+        v-if="(index != props.failedTestResult.titleParts.length - 1)"
+        data-cy="right-chevron"
+        size="8"
+        stroke-color="gray-200"
+        fill-color="gray-200"
+      />
+    </li>
+  </div>
 </template>
 <script lang="ts" setup>
-import { computed } from 'vue'
 import { IconChevronRightSmall } from '@cypress-design/vue-icon'
+import type { TestResults } from './DebugSpec.vue'
 
 const props = defineProps<{
-  failedTest: string[]
+  failedTestResult: TestResults
 }>()
-
-const testData = computed(() => props.failedTest)
 
 </script>
--- a/packages/app/src/debug/DebugFailedTest.vue
+++ b/packages/app/src/debug/DebugFailedTest.vue
@@ -3,21 +3,23 @@
     data-cy="test-row"
     class="flex flex-row items-center gap-x-2.5 non-italic text-base text-gray-700 font-normal"
   >
-    <span
+    <div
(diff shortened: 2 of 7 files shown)


packages/app/src/debug/DebugSpec.vue
```diff
@@ -0,0 +1,115 @@
...
+            <span
+              class="font-medium text-gray-900"
+            >
+              {{ specData.specName.split('.')[0] }}
+            </span>
+            <span
+              class="font-normal text-gray-600"
+            >
+              {{ "." + specData.specName.split('.').slice(1, specData.specName.length).join('.') }}
+            </span>
```
warrensplayer: Move the additional string manipulation from the template to your `splitHeader` method below.  It could output three parts (path, fileName, extension).  That way the logic is consolidated in one place. Did you look to see if the code does this anywhere else?

packages/app/src/debug/DebugSpec.vue
```diff
@@ -0,0 +1,115 @@
...
+        </li>
+      </ul>
+      <ul
+        v-for="test in specData.failedTests"
+        :key="`test-${test.id}`"
+        data-cy="test-group"
+        class="w-full flex flex-col flex-start h-12"
+      >
+        <DebugFailedTest
+          :failed-test="test.titleParts"
```
warrensplayer: Pass the entire test as a prop, not just the `titleParts`.  This will allow for more flexibility in the future without having to change the prop.

packages/app/src/debug/DebugSpec.vue
```diff
@@ -0,0 +1,115 @@
...
+import { IconActionRefresh } from '@cypress-design/vue-icon'
+import DebugFailedTest from './DebugFailedTest.vue'
+
+const props = defineProps<{
+  spec: Spec
+  testResults: TestResults[]
+  disabled?: boolean
+}>()
+
+const splitHeader = (relativePath: string) => {
```
warrensplayer: Add a method comment describing what this function is doing. Examples would be nice as well.

packages/app/src/debug/DebugFailedTest.vue
```diff
@@ -0,0 +1,33 @@
...
+    </span>
+    <IconChevronRightSmall
+      size="8"
+      stroke-color="gray-200"
+      fill-color="gray-200"
+    />
+    <span
+      data-cy="test-description"
+    >
+      {{ testData[1] }}
```
mike-plummer: What about tests that have more than two title segments? For example, a spec might be within two or more nested `describe` blocks, or `describe` > `context` > `it`

packages/app/src/debug/DebugFailedTest.vue
```diff
@@ -0,0 +1,33 @@
...
+    </span>
+  </li>
+</template>
+<script lang="ts" setup>
+import { computed } from 'vue'
+import { IconChevronRightSmall } from '@cypress-design/vue-icon'
+
+const props = defineProps<{
+  failedTest: string[]
```
mike-plummer: I feel like this could be more descriptively named - the intent is that this prop contains the test title segments, right?

packages/app/src/debug/DebugFailedTest.vue
```diff
@@ -0,0 +1,33 @@
...
+}>()
+
+const testData = computed(() => props.failedTest)
```
mike-plummer: Why does this need to be computed?

packages/app/src/debug/DebugFailedTest.vue
mike-plummer: Or, conversely, a test that's at the top level (just an `it` without a wrapping `describe` would just have one title segment, wouldn't it?

packages/app/src/debug/DebugSpec.vue
```diff
@@ -0,0 +1,115 @@
...
+          <button
+            v-else
+            data-cy="run-failures-disabled"
+            class="inline-flex gap-x-10px whitespace-nowrap h-8 justify-center items-center
+            isolate rounded border border-gray-100 bg-white non-italic text-sm text-gray-500 font-medium px-12px"
+            :disabled="disabled"
+          >
+            <IconActionRefresh stroke-color="gray-500" />
+            Run Failures
+          </button>
```
mike-plummer: This looks duplicative - you should be able to dynamically build the `data-cy`, `class` and `disabled` props based on the value of `disabled` without duplicating this block

packages/app/src/debug/DebugSpec.vue
```diff
@@ -0,0 +1,115 @@
...
+}>()
+
+const splitHeader = (relativePath: string) => {
+  const before = relativePath.split('/')
+
+  return {
+    initialPath: `${before.slice(0, before.length - 1).join('/') }/`,
+    specName: before.slice(-1)[0],
+  }
+}
```
mike-plummer: I feel like this could just be inlined into the `computed` block below, the separate function declaration makes this component a bit more complex

packages/app/src/debug/DebugSpec.vue
```diff
@@ -0,0 +1,115 @@
...
+            >
+              {{ "." + specData.specName.split('.').slice(1, specData.specName.length).join('.') }}
+            </span>
+          </span>
+          <button
+            v-if="!disabled"
+            data-cy="run-failures"
+            class="inline-flex gap-x-10px whitespace-nowrap h-8 justify-center items-center
+            isolate rounded border border-gray-100 bg-white non-italic text-sm text-indigo-500 font-medium px-12px"
+          >
```
mike-plummer: Most of our buttons have hover behaviors that either add the border or a background color, but I don't see that here

mike-plummer (changes requested):
Left a couple comments. Also wondering about responsiveness in this UI - seems like the button is hardcoded to flow off the screen and the text wrapping can break row height, especially on tests with very long names

Screen Shot 2022-11-29 at 12 02 37 PM

packages/app/src/debug/DebugSpec.vue
warrensplayer: @amehta265 You can use the `Button` component from `frontend-shared` which should take care of the styling.  It should also handle the disabled styling, so you will not need to duplicate it below just for being disabled.

packages/app/src/debug/DebugSpec.vue
warrensplayer: I could go either way with inlining this or not. I am fine with it being a separate function to encapsulate the string parsing logic from the creation of the return value from the `specData` method.

packages/app/src/debug/DebugSpec.vue
warrensplayer: Also, this button should route to the runner to run the selected spec just like clicking the spec in the SpecList page.  Happy to pair on this if that would he helpful.

packages/app/src/debug/DebugFailedTest.vue
warrensplayer: Yes, this should handle `n` numbers of title segments.

packages/app/src/debug/DebugFailedTest.vue
warrensplayer: The prop should be the entire test with the `titleParts` just being a field on the test.

warrensplayer (changes requested):

packages/app/src/debug/DebugSpec.vue
amehta265: Consolidating all the string manipulation in `splitHeader` makes sense
The only other place this is somewhat seen is in the specsList. When searching for spec files the `getHighlightedIndexes` method in the `useCollapsibleTree.ts` file will return certain indexes that need to be highlighted based on the `fuzzySortSpecs`. But I think the logic there is much more complicated for our use case here.

packages/app/src/debug/DebugSpec.vue
amehta265: This is a good point. I have used the `Button` component from 'frontend-shared` which would help avoid this duplication and dynamically build the component attributes

packages/app/src/debug/DebugFailedTest.vue
amehta265: This was an oversight on my end. There is no need for this reactivity layer in this case. If there is any update, a new value will be passed into this component through the failedTest prop and a component will be created based on the new value. No reactivity is needed here. The reactivity will be needed in the parent component that has to update and re-render `DebugFailedTest` if any new tests have failed.
I have removed the computed property on the latest commit

packages/app/src/debug/DebugFailedTest.vue
amehta265: I have made this update on the latest commit

packages/app/src/debug/DebugSpec.vue
amehta265: The use of the `Button` component from `frontend-shared` as well as the "wiring" of the button to run the spec has been added on the latest commit

packages/app/src/debug/DebugFailedTest.vue
BIGTECH1982: THANK YOU GUYS! IPHONE 12 Pro max

packages/app/src/debug/DebugFailedTest.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <div
+    data-cy="test-row"
+    class="flex flex-row items-center gap-x-2.5 non-italic text-base text-gray-700 font-normal"
+  >
+    <li
```
warrensplayer: You should not have an `li` inside a `div`, only as a direct child of a `ul` element.  Here, this `li` can just be a span.

packages/app/src/debug/DebugFailedTest.vue
```diff
@@ -0,0 +1,31 @@
...
+      :data-cy="`titleParts-${index}`"
+    >
+      {{ titlePart }}
+      <IconChevronRightSmall
+        v-if="(index != props.failedTestResult.titleParts.length - 1)"
+        data-cy="right-chevron"
+        size="8"
+        stroke-color="gray-200"
+        fill-color="gray-200"
+      />
```
warrensplayer: You can simplify the `v-if` logic here by moving this outside the element containing the `titlePart`, putting it ahead of the `titlePart` and changing the `v-if` to be `index != 0`

packages/frontend-shared/src/components/Button.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <button
-    v-if="!props.href && !props.to"
+    v-if="((!props.href && !props.to) || (props.disabled && props.to))"
```
warrensplayer: Why add this extra conditional to render a button when the element should be disabled?  Was the disabled styling for the `a` tag not looking correct?  It would be better to always render an `a` tag when it is disabled.

packages/frontend-shared/src/components/Button.vue
```diff
@@ -88,6 +88,7 @@ const VariantClassesTable = {
   linkBold: 'border-transparent text-indigo-500 font-medium',
   text: 'border-0',
   secondary: 'bg-jade-500 text-white hocus-secondary',
+  white: 'bg-white text-indigo-500 font-medium hocus-default',
```
warrensplayer, line 91: Update the CT test for the Button to add a case for this new variant.  There is a `playground` test in the spec that showcases all the variants.

warrensplayer (changes requested):
Just a few minor things and a question in my comments.  Additionally, add appropriate `cy.percySnapshot()` calls to the CT tests to capture the UI of the new components that cannot be checked with just normal assertions.

packages/frontend-shared/src/components/Button.vue
amehta265: The reason for this extra conditional is because of a `typeError: Cannot use 'in' operator to search for 'path' in undefined` would surface when passing a `:to` and a `:disabled` prop to the `Button` from front-end shared. 

The reason for this error is when these 2 props are passed to the Button, a component with an `:is` attribute of value `RouterLink` is rendered because of the `:to` prop, however, the component also binds a null `href` (`v-bind:linkProps`) Hence, in order to make the button disabled I added this extra conditional to render the disabled button if both the disabled prop and to prop is passed in

packages/app/src/debug/DebugFailedTest.cy.tsx
```diff
@@ -0,0 +1,40 @@
...
+      titleParts: ['Login', 'Should redirect unauthenticated user to signin page'],
+    }
+
+    cy.mount(() => (
+      <DebugFailedTest failedTestResult={testResult} />
+    ))
+
+    cy.findByTestId('test-row').children().should('have.length', 2).then((ele) => {
+      expect(ele).to.have.css('color', 'rgb(90, 95, 122)')
+    })
```
mike-plummer: The `then` could be simplified to an `and` clause:

```
cy.findByTestId('test-row').children()
    .should('have.length', 2)
    .and('have.css', 'color', 'rgb(90, 95, 122)')
```

packages/app/src/debug/DebugFailedTest.cy.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+    testResult.titleParts.forEach((title, index) => {
+      cy.findByTestId(`titleParts-${index}`).should('have.text', ` ${title }`)
```
mike-plummer: Looks like extra spaces are getting added to the output of the title parts, is this for spacing/padding? Should use CSS for that

packages/app/src/debug/DebugFailedTest.vue
```diff
@@ -0,0 +1,31 @@
...
+    class="flex flex-row items-center gap-x-2.5 non-italic text-base text-gray-700 font-normal"
+  >
+    <span
+      v-for="(titlePart, index) in props.failedTestResult.titleParts"
+      :key="`${titlePart}-${index}`"
+      class="flex flex-row items-center gap-x-2.5"
+      :data-cy="`titleParts-${index}`"
+    >
+      <IconChevronRightSmall
+        v-if="index != 0"
```
mike-plummer: `!==`

packages/app/src/debug/DebugSpec.cy.tsx
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    cy.mount(() => (
+      <DebugSpec spec={spec} testResults={testResults} />
+    ))
+
+    cy.findByTestId('spec-path').should('have.css', 'text-overflow', 'ellipsis')
+    cy.findByTestId('run-failures').should('be.visible')
+
+    cy.percySnapshot()
```
mike-plummer, line 100: Nice job on the snapshots 👍

packages/app/src/debug/DebugSpec.vue
```diff
@@ -0,0 +1,122 @@
...
+            <span
+              class="font-medium text-gray-900"
+            >
+              {{ specData.fileName }}
+            </span>
+            <span
+              class="font-normal text-gray-600"
+            >
+              {{ specData.extension }}
+            </span>
```
mike-plummer: FYI - there is a `SpecNameDisplay` component that I believe does something very similar (bold fileName, non-bold extension). Might see if that could be reused

packages/app/src/debug/DebugSpec.vue
```diff
@@ -0,0 +1,122 @@
...
+          </div>
+          <div
+            class="ml-10px"
+          >
+            <Button
+              data-cy="run-failures"
+              variant="white"
+              class="inline-flex gap-x-10px whitespace-nowrap justify-center items-center isolate"
+              :disabled="isDisabled"
+              :to="{ path: '/specs/runner', query: { file: (specData.path + specData.fileName + specData.extension).replace(/\\/g, '/') } }"
```
mike-plummer: Couldn't you just use the value from the input prop `spec.path`?

packages/app/src/debug/DebugSpec.vue
```diff
@@ -0,0 +1,122 @@
...
+            <span
+              class="font-normal text-gray-600"
+            >
+              {{ specData.extension }}
+            </span>
+          </div>
+          <div
+            class="ml-10px"
+          >
+            <Button
```
mike-plummer, line 39: FYI - always a good practice to specify `type="button"` so it's obvious this isn't intended as a `type="submit"` form submission button. That way you don't accidentally end up submitting a form if this button happens to nested within one

packages/app/src/debug/DebugSpec.vue
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * @param relativePath: The relative path of a failed test e.g. cypress/test/auth.spec.ts
+ * We split the relative path into 3 pieces and return an object containing this split sections
+ * so we can correctly style the relative path in the debug spec header. For example:
+ * path: cypress/tests/ ; fileName: auth ; extension: .spec.ts
+ */
+const splitHeader = (relativePath: string) => {
+  const before = relativePath.split('/')
+  const specName = before.slice(-1)[0]
+  const splitSpecName = specName.split('.')
```
mike-plummer: What about a spec filename like `my.component.spec.cy.js`? I think this logic would only end up highlighting `my` when it should cover `my.component.spec`. Our GQL schema exposes `Spec.fileName` and `Spec.specFileExtension` fields that are already pre-parsed, may want to see if those can be reused and just accept those prepared values as props

packages/app/src/debug/DebugSpec.vue
warrensplayer: @mike-plummer The Button component ends up rendering an `a` tag when the `to` prop is used.  So, the `type="button"` would not be needed here.  I am starting to think the `Button` component needs to get split up tough because it tries to do a lot in one component.

packages/app/src/debug/DebugFailedTest.cy.tsx
amehta265: Great catch! Turns out that the `<span>` containing the chevron and title part was adding a whitespace to the text inside it. I changed this component from a `span` to a `div` which alleviates the issue

packages/app/src/debug/DebugSpec.vue
amehta265: Thank you for pointing this component out. I would agree that it would be more consistent to use `SpecNameDisplay` on this component.

packages/app/src/debug/DebugSpec.vue
amehta265: @warrensplayer Would it be possible to pass in `Spec.fileName` and `Spec.specFileExtension` instead of  `path` to this component when using it in the parent component? 
My understanding is that the GQL query will be in the parent component, therefore if this is something that could be done, that it would overcome several of these edges cases and we could simplify the `splitHeader` method

packages/app/src/debug/DebugSpec.cy.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <DebugSpec spec={spec} testResults={testResults} />
+    ))
+
+    cy.findByTestId('debug-spec-item').children().should('have.length', 3)
+    cy.findByTestId('spec-contents').children().should('have.length', 2)
+    cy.findByTestId('spec-path').should('have.text', 'cypress/tests/auth.spec.ts')
+    cy.contains('auth').should('have.css', 'color', 'rgb(67, 72, 97)')
+    cy.findByTestId('run-failures').should('not.be.disabled')
+    .should('have.text', ' Run Failures ')
+    .should('have.css', 'color', 'rgb(73, 86, 227)')
```
lmiller1990: I am not sure we should assert CSS like this - I think the snapshot with percy should cover this.

packages/app/src/debug/DebugSpec.vue
```diff
@@ -0,0 +1,122 @@
...
+                  data-cy="icon-refresh"
+                  stroke-color="indigo-500"
+                />
+              </template>
+              Run Failures
+            </Button>
+          </div>
+        </div>
+      </div>
+      <ul
```
lmiller1990: Is this supposed to be `<ul>`? It does not contain any `<li>` - it seems odd to have a list without any list items.

packages/app/src/debug/DebugSpec.vue
lmiller1990: I am a bit out of the loop for this project - does this data come straight from the Cloud or going via our data-context?

We've got logic to get all the parts of a file here: https://github.com/cypress-io/cypress/blob/develop/packages/data-context/src/sources/ProjectDataSource.ts#L94-L104 we use this for the Specs List and some other places. I think, if possible, we should do all this preprocessing on the server and then serve it up using the same interface we use for the rest of App https://github.com/cypress-io/cypress/blob/develop/packages/graphql/src/schemaTypes/objectTypes/gql-Spec.ts

Not sure if we need to do this here or later on, just some ideas. Good to reuse battle tested code where we can.

lmiller1990 (commented):
Left a few quick comments, I'm guessing once the backend is in place we might need to slightly change things up, so happy to let this get through without too much resistance, as long as everything is generally looking correct.

packages/app/src/debug/DebugSpec.vue
warrensplayer: @amehta265 Yes, I see the logic that Lachlan is pointing to above and we can make sure the new GraphQL fields take advantage of that same parsing logic.  You can remove that parsing logic and update the props of the component to take something like what Mike is talking about on the Specs List page: 
```
{
  path: '/src/main/components',
  fileName: 'MyComponet',
  fileExtension: '.cy.ts'
}
```

packages/frontend-shared/src/components/Button.vue
warrensplayer: I still do not like adding this conditional here.  The more I look at the Button component, I think it is doing too much, but do not want to refactor too much for this story.  I did debug this and it is not because of the null `href`, but because the `RouterLink` is expecting a `to` prop which is not passed to it when disabled.  You can look at the `linkProps` computed around line 150.  

There are two solutions to try:
* Remove this conditional and go to the `linkProps` and ensure the `to` parameter is passed to the `RouterLink` even when disabled.  In this case, an additional class of `pointer-events-none` needs to get added along with `cursor-default` around line 139 when disabled.  This will prevent the link from being clickable.
* Look at the `linkVersion` variable and have it render a `BaseLink` when disabled instead of a `RouterLink`.  This will keep it rendering an `a` tag instead of switching to a `button`, but should render it without an href which means clicking it would do nothing.  

I am struggling with which option is the right solution for accessibility purposes.  @marktnoonan Do you have any suggestions here?  

Also, please add a test in Button.cy.tsx to test a disabled `Button` with a `to` prop.

packages/app/src/debug/DebugSpec.vue
```diff
@@ -0,0 +1,122 @@
...
+              :disabled="isDisabled"
+              :to="{ path: '/specs/runner', query: { file: (spec.path).replace(/\\/g, '/') } }"
+            >
+              <template #prefix>
+                <IconActionRefresh
+                  data-cy="icon-refresh"
+                  stroke-color="indigo-500"
+                />
+              </template>
+              Run Failures
```
warrensplayer: Use the `@cy/i18n` library for this text.

packages/app/src/debug/DebugSpec.vue
warrensplayer: The `DebugFailedTest` used to render `li` elements, but no longer does.  This can just be a `div` now.

warrensplayer (changes requested):

packages/app/src/debug/DebugSpec.cy.tsx
amehta265: I agree. The CSS assertions have been removed on the latest commit

warrensplayer (approved):
Looks good!

mike-plummer (approved):

PR #24881, opened 2022-11-29
feat: server logic for create from React component

### LINKED ISSUE #24774: Implement 'create from React component' logic
Implement all of the server-side logic that we need for Create from React Component feature. This should include

- React component spec file template (see `packages/data-context/src/codegen/templates/vue-component/vue-component.ejs` for example)
- Query that takes a path to a file and returns a list of React components in that file, using [react-docgen](https://github.com/reactjs/react-docgen) to parse the file

See the [Create from React Component tech brief](https://github.com/cypress-io/tech-briefs/tree/main/briefs/ct-create-from-react-component) for guidance

[Figma designs](https://www.figma.com/file/gq6zQUQ1bx2uqfnF64lwX9/Create-from-React-Component?node-id=0%3A1&t=4zY13RAbZqIikzLH-0) for reference

astone123:
- Closes #24774

NOTE: This targeting a feature branch [`feature/create-from-react-component`](https://github.com/cypress-io/cypress/tree/feature/create-from-react-component).

### User facing changelog

n/a

### Additional details

NOTE: The work in this PR assumes that the component passed to `spec-options` is a named export. In #24929 we will add support for components that are exported by default.

This PR does the following:
- Add a new GraphQL mutation `getReactComponentsFromFile` that the front end can call to parse a JS/TS file and get info about the exported React components that are defined in the file
- Add a new `react-component` spec template
- Add logic in `spec-options` to accept a `componentName` option and to generate a spec using the `react-component` template when the project framework is React

### Steps to test

There aren't many ways to manually test this without a front end. Take a look at the automated tests that I've written so far.

n/a

### PR Tasks


packages/data-context/package.json
```diff
@@ -75,6 +75,7 @@
     "@types/sinon-chai": "3.2.8",
     "@types/stringify-object": "^3.0.0",
     "mocha": "7.0.1",
+    "react-docgen": "5.4.3",
```
lmiller1990: One thing that might be nice to do is see how much the binary increases if we add a dependency. Just something I thought of - I'm not sure what this uses under the hood, but I'm guessing it uses babel, which might bring in a bunch of extra node modules.

I think we could check it by building the binary, and just comparing the size vs the current production version.

packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -345,7 +355,23 @@ export class ProjectActions {
...
   }
 
-  async codeGenSpec (codeGenCandidate: string, codeGenType: CodeGenType): Promise<NexusGenUnions['GeneratedSpecResult']> {
+  async getReactComponentsFromFile (filePath: string): Promise<ReactComponentDescriptor[]> {
+    const project = this.ctx.currentProject
+
+    assert(project, 'Cannot create spec without currentProject.')
+
+    try {
+      const src = await fs.readFile(this.ctx.path.join(project, filePath), 'utf8')
```
lmiller1990: I think we can use this method: https://github.com/cypress-io/cypress/blob/develop/packages/data-context/src/actions/FileActions.ts#L11

packages/data-context/src/codegen/spec-options.ts
```diff
@@ -53,18 +55,19 @@ export class SpecOptions {
...
+
+    switch (this.options.framework.codeGenFramework) {
+      case 'react':
+        frameworkOptions = await this.getReactSpecOptions()
+        break
+      case 'vue':
+        frameworkOptions = await this.getVueSpecOptions()
+        break
+      default:
+        throw new Error(`Unable to generate spec for ${this.options.framework.codeGenFramework}`)
```
lmiller1990:
```suggestion
    switch (this.options.framework.codeGenFramework) {
      case 'react':
        return await this.getReactSpecOptions()
      case 'vue':
        return await this.getVueSpecOptions()
      default:
        throw new Error(`Unable to generate spec for ${this.options.framework.codeGenFramework}`)
```

I think this can be simplified, unless I'm missing something.

packages/data-context/src/codegen/templates/react-component/react-component.ejs
```diff
@@ -0,0 +1,13 @@
+---
+fileName: <%= fileName %>
+---
+
+import React from 'react'
+import { <%- componentName %> } from '<%- componentPath %>'
+
+describe('<<%- componentName %> />', () => {
+  it('renders', () => {
+    // see: https://docs.cypress.io/guides/component-testing/react/api
```
lmiller1990: Should this be an `on` link?

packages/data-context/test/unit/actions/ProjectActions.spec.ts
```diff
@@ -17,6 +18,55 @@ describe('ProjectActions', () => {
     actions = new ProjectActions(ctx)
   })
 
+  context('getReactComponentsFromFile', () => {
+    beforeEach(async () => {
```
lmiller1990:
```suggestion
    beforeEach(() => {
```

packages/data-context/test/unit/actions/ProjectActions.spec.ts
```diff
@@ -3,6 +3,7 @@ import { ProjectActions } from '../../../src/actions/ProjectActions'
 import { createTestDataContext } from '../helper'
 import { expect } from 'chai'
 import sinon from 'sinon'
+import path from 'path'
```
lmiller1990: I wonder if we need `CodegenActions` soon... WDYT?

packages/data-context/src/codegen/templates/react-component/react-component.ejs
```diff
@@ -0,0 +1,13 @@
+---
+fileName: <%= fileName %>
+---
+
+import React from 'react'
+import { <%- componentName %> } from '<%- componentPath %>'
```
lmiller1990, line 6: What happens for default export components, like `CounterDefault`? Would this become `import { CounterDefault } from './counter-default'` (which would be undefined). Did we find a solution for this, or are we moving forward with this as a known limitation?

It should be possible to write a quick babel parser to find the default export and compare that to the components `react-docgen` found. Not sure how expensive this would be, probably not that much.

lmiller1990 (commented):
Left some comments, I will wait for the UI before I test it more thoroughly. Great to see this coming together, this is pretty much exactly what I was expecting to see when I was writing the technical brief.

packages/data-context/src/codegen/templates/react-component/react-component.ejs
astone123: I think we'll need a separate spec template for default export components. I'm looking into how we can figure out if a component is exported by using a custom [handler](https://github.com/reactjs/react-docgen#handlers-1) with React Docgen.

Otherwise yeah we can probably do it after we parse with React Docgen

ZachJW34: I looked at the result the `componentDefinition` which is what the handler receives and it doesn't contain enough info to determine how the component was exported. For example:
```tsx
const Button = (props: {text: string}) => <button>{props.text}</button>

export { Button }
export default Button
```

The componentDef only contains info on `const Button = (props: {text: string}) => <button>{props.text}</button>`. Rather than a custom handler, I believe a custom resolver makes sense here so we can scrape all of the exported declarations and then compare them to the returned componentDefinitions. We don't need to write our own parser since a resolver directly receives the parsed ast.

packages/data-context/src/actions/ProjectActions.ts
astone123: I actually think that this should be an absolute path to the file, so we won't need to do this anyway

packages/data-context/src/codegen/templates/react-component/react-component.ejs
astone123: Going to handle this as a separate task (to be completed before releasing this feature) https://github.com/cypress-io/cypress/issues/24929

packages/data-context/package.json
astone123: yeah... when I run `yarn`, it adds a ton of lines to `snapshot-meta.cache.json` which leads me to believe that this brings in a bunch of other modules (although I'm not sure exactly what that file does)

packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -76,6 +80,11 @@ type SetForceReconfigureProjectByTestingType = {
   testingType?: TestingType
 }
 
+interface ReactComponentDescriptor {
+  description: string
```
lmiller1990: Is this unused? Seems that way.

astone123: we're using this to type cast the result from `parseReactComponent` since the types are messed up
https://github.com/cypress-io/cypress/pull/24881/files#diff-d36f6626bf805d1bbbfba4786ca3a518f0083613a9814fd037c5592808949606R365

astone123:
The diff is massive because of some issues that we had with V8 snapshots. The main changes are in the following directories:

- `packages/data-context/`
- `packages/frontend-shared/`

You should be able to collapse the rest of the changes and trust CI for those

npm/webpack-preprocessor/__snapshots__/compilation.spec.js
```diff
@@ -1,6 +1,6 @@
...
+//# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJmaWxlIjoiZXhhbXBsZV9zcGVjX291dHB1dC5qcyIsIm1hcHBpbmdzIjoiQUFBQUEsR0FBRyxhQUFhLFdBR2RDLE9BRmdCLEdBRU5DLEdBQUdDLE1BQU0sR0FDbkJGLE9BSG1CLEdBR1RDLEdBQUdDLE1BQU0sR0FDbkJGLE9BQU9HLEtBQUtDLElBQUcsTUFBUkQsS0FBWSxDQUFDLEVBQUcsS0FBS0YsR0FBR0MsTUFBTSIsInNvdXJjZXMiOlsid2VicGFjazovL0BjeXByZXNzL3dlYnBhY2stcHJlcHJvY2Vzc29yLy4vdGVzdC9fdGVzdC1vdXRwdXQvZXhhbXBsZV9zcGVjLmpzIl0sInNvdXJjZXNDb250ZW50IjpbIml0KCdpcyBhIHRlc3QnLCAoKSA9PiB7XG4gIGNvbnN0IFthLCBiXSA9IFsxLCAyXVxuXG4gIGV4cGVjdChhKS50by5lcXVhbCgxKVxuICBleHBlY3QoYikudG8uZXF1YWwoMilcbiAgZXhwZWN0KE1hdGgubWluKC4uLlszLCA0XSkpLnRvLmVxdWFsKDMpXG59KVxuIl0sIm5hbWVzIjpbIml0IiwiZXhwZWN0IiwidG8iLCJlcXVhbCIsIk1hdGgiLCJtaW4iXSwic291cmNlUm9vdCI6IiJ9
```
ryanthemanuel, line 3: Do we know why this changed?

astone123: I'm not sure about this one... looks like Lachlan updated it earlier. I can try to revert it and see what happens

ryanthemanuel: No concerns about it necessarily. I was just curious. Maybe some source mapping dependency updates?

astone123: Maybe. Looks like CI is happy 🤷🏻‍♂️

packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -345,7 +354,24 @@ export class ProjectActions {
     this.api.insertProjectPreferencesToCache(this.ctx.lifecycleManager.projectTitle, args)
   }
 
-  async codeGenSpec (codeGenCandidate: string, codeGenType: CodeGenType): Promise<NexusGenUnions['GeneratedSpecResult']> {
+  async getReactComponentsFromFile (filePath: string): Promise<ReactComponentDescriptor[]> {
+    try {
+      const src = await fs.readFile(filePath, 'utf8')
+
+      let result = parseReactComponent(src, reactDocgenResolvers.findAllExportedComponentDefinitions)
```
mike-plummer: `const`

packages/data-context/src/codegen/spec-options.ts
```diff
@@ -54,19 +55,14 @@ export class SpecOptions {
...
-        templateKey: 'componentEmpty' as TemplateKey,
-        overrideCodeGenDir: '',
-      }
+    switch (this.options.framework.codeGenFramework) {
+      case 'react':
+        return await this.getReactSpecOptions()
+      case 'vue':
+        return await this.getVueSpecOptions()
+      default:
+        throw new Error(`Unable to generate spec for ${this.options.framework.codeGenFramework}`)
```
mike-plummer, line 64: Since this now errors rather than falling back to an empty spec do we need the `componentEmpty` entry in `TemplateKey` above?

packages/data-context/test/unit/actions/ProjectActions.spec.ts
mike-plummer: I like the idea of splitting these out, just to keep things organized

packages/data-context/test/unit/actions/ProjectActions.spec.ts
```diff
@@ -17,6 +18,51 @@ describe('ProjectActions', () => {
...
+
+      expect(reactComponents).to.have.length(1)
+      expect(reactComponents[0].displayName).to.equal('CounterDefault')
+    })
+
+    it('does not throw while parsing empty file', async () => {
+      const reactComponents = await actions.getReactComponentsFromFile(`${absolutePathPrefix}/empty.jsx`)
+
+      expect(reactComponents).to.have.length(0)
+    })
```
mike-plummer: As we discussed @astone123 , there are a few conditions that aren't covered here that I would like to see tests added for:

1. Multiple named exports are parsed
2. Mix of named and default exports are parsed
3. Functional component that returns a non-JSX value is parsed (namely, a string) - appears to be a limitation of the docgen library
4. Functional component that returns null is parsed - appears to be a limitation of the docgen library

packages/data-context/test/unit/actions/project/counter-multiple-components.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+export function CounterContainer () {
```
mike-plummer, line 3: I would like to see some different syntaxes used just to make sure our babel parsing/libraries can handle them:

1. const arrow function:
```
const Counter = () => <p>Counter</p>
```
2. Separate export statements:
```
function Counter () { return <p>Counter</p> }
function OtherCounter() { return <p>OtherCounter</p> }

export {
  Counter
}
export default OtherCounter
```

mike-plummer (changes requested):
Very cool stuff 😎.  Couple edge cases identified that will need to be addressed.

packages/data-context/src/codegen/spec-options.ts
astone123: It shouldn't be possible to reach this line since the UI will only show the Create from Component generator card for React and Vue apps.

`componentEmpty` is still a valid template key because that generator card will still appear in the UI for all applications

packages/data-context/test/unit/actions/ProjectActions.spec.ts
astone123: I'd like to try to handle 3 and 4 in https://github.com/cypress-io/cypress/issues/24929 since I think we'll need to use a custom resolver for both of these issues. As of right now I'm considering 3 and 4 as edge cases.

Will add test cases for 1 and 2

mike-plummer (approved):
Thanks for adding the new test cases. Looks like we're leaving the "non-function return type" resolver for future work, along with potentially splitting the codegen actions into a separate `*Actions` container. Works for me

ZachJW34 (approved):
There is another edge case with regards to `export { Foo as Bar }` but will add that to #24929

npm/webpack-preprocessor/__snapshots__/compilation.spec.js
lmiller1990: @ryanthemanuel @astone123  Only 2 characters changed - I did a diff, char by char. Characters 271 and 277 changed. These are the boundaries between one token.

```
$ diff old new

271c271
< 271 w
---
> 271 c
277,278c277,278
< 277 Q
< 278 U
---
> 277 U
> 278 k
```

This diff isn't useful, but shows how the change was only a few chars. Due to the implicit babel deps update, the source map is now slightly more granular. Specifically, here's what those two characters correspond to:

https://user-images.githubusercontent.com/19196536/205534804-13b12ca0-fc9a-4a5c-8f70-1a25c99639f8.mp4

In the update, `Math.min(` separates `min` and `(`. This is not the case before - `Math.min(...` is a single token.

From the user's point of view, there isn't really a difference.

packages/data-context/src/actions/ProjectActions.ts
lmiller1990: Sorry, I should be more specific - I'm saying the `description` property is unused. We only use `displayName`. We can keep both if you think it's useful.

packages/data-context/src/actions/ProjectActions.ts
```diff
@@ -345,7 +354,24 @@ export class ProjectActions {
     this.api.insertProjectPreferencesToCache(this.ctx.lifecycleManager.projectTitle, args)
   }
 
-  async codeGenSpec (codeGenCandidate: string, codeGenType: CodeGenType): Promise<NexusGenUnions['GeneratedSpecResult']> {
+  async getReactComponentsFromFile (filePath: string): Promise<ReactComponentDescriptor[]> {
+    try {
+      const src = await fs.readFile(filePath, 'utf8')
```
lmiller1990:
```suggestion
      const src = await this.ctx.fs.readFile(filePath, 'utf8')
```

I see a convention where do use `ctx.fs` to access this. WDYT?

packages/data-context/test/unit/actions/ProjectActions.spec.ts
lmiller1990: @astone123 thoughts? Not sure how this fits into the timeline, happy to help out if you think this is a good idea -- as much as I want to say "we can do this later", I really doubt we actually will. Might be good to make this refactor now - I don't think it's a lot of work.

packages/graphql/schemas/schema.graphql
```diff
@@ -1576,6 +1581,15 @@ type Query {
   wizard: Wizard!
 }
 
+"""Properties describing a React component"""
+type ReactComponentDescriptor {
+  """A description for the component"""
+  description: String!
```
lmiller1990: Just to revisit - I don't think we ever consume `description` in the UI - this seems like an unused field, and doesn't seem to contain anything useful in practice.

lmiller1990 (commented):
Nice, couple of quality-of-life changes I think are worth doing. LMK what you think.

packages/data-context/test/unit/actions/ProjectActions.spec.ts
astone123: Created `GodegenActions` in https://github.com/cypress-io/cypress/pull/24881/commits/9449d66fbfadd25c70dad78bd243f6fde64b999d

packages/data-context/src/actions/ProjectActions.ts
astone123: Misunderstood - yeah, we don't need that property. Removed in https://github.com/cypress-io/cypress/pull/24881/commits/9449d66fbfadd25c70dad78bd243f6fde64b999d

lmiller1990 (approved):
Looking good to me. Merge when :heavy_check_mark: on CI!

PR #24890, opened 2022-11-29
fix: Fix incompletely cleared subjects when linking chainers

BlueWinds:
### User facing changelog
No user facing changes.

### Additional details
This fixes a bug in the implementation of query subject chains, where linked chainers (eg: the inside and outside of cy.then()) would result in the outer chainer having an incorrect subject.

This never went in front of users, was discovered in the pre-release binary during internal testing.

No change, it now works the same as Cy11.

### PR Tasks


chrisbreiding (approved):

marktnoonan (approved):
I was able to test this enough to see that the chained `then()` is passing in this branch and not in the main release branch, so approving this. Will still plan on testing against the original component testing example I had, but it's not blocking to this PR.

PR #24892, opened 2022-11-30
fix: fix browser focus issue on MacOS

### LINKED ISSUE #21743: App window does not respond to mouse events in window focus race condition
https://user-images.githubusercontent.com/1271364/171277456-f5201f5a-0703-4111-8668-8f198cae8a5d.mov

### Browsers
All browsers, **EXCEPT for Electron** because Electron is launched in the same docked application as Cypress

### Operating Systems and Hardware
* Reproduced on OSX Monterey w/ M1 Chip
* Could not reproduce on OSX Catalina w/ Intel Chip
* Reproduced intermitently on Monterey with intel Chip

### Steps to reproduce
1. Click "start component testing" with Chrome browser
1. IMMEDIATELY move your cursor (don’t click/change focus) onto where the window will open.
1. Again, do not click anything

**Observe:** hover states don’t work, pointer states aren’t working, you have to click 2-3x for any clicks to register

**Workaround:** Once you Cmd + Tab/lose focus to another docked application, everything will work again

This does not happen in Electron because it shares the same Docked Application with Cypress.

┆Issue is synchronized with this [Jira Bug](https://cypress-io.atlassian.net/browse/UNIFY-1172) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/11503">Screen Recording 2022-03-11 at 6.07.14 PM.mov</a> | <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/11849">Screen Recording 2022-04-21 at 1.34.59 PM.mov</a>
┆author: Shawn Taylor
┆epic: Edge Cases
┆friendlyId: UNIFY-1172
┆priority: Low
┆sprint: Unification Hardening
┆taskType: Bug

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/21743
### User facing changelog

Fix a MacOS specific bug where the browser would not be correctly focused when launched.

### Additional details

For reasons unknown, if you quickly jiggle your mouse back and forth when launching Chrome or Firefox browsers on MacOS, the browser would not be correctly focused. You can see this below in the first video - I jiggle my mouse, and then I need to click the search bar a bunch of times to get the browser to properly focus.

**Note** this also happens in Cypress 9 on the latest MacOS version. I reproduced it here: https://github.com/cypress-io/cypress/issues/21743#issuecomment-1331598390

There was a work around which was to switch to any other window, then switch back, and it would work as expected. I don't know why this fixes the issue, but i does. I was able to do the same thing programmatically, but simply showing and immediately hiding the electron app in code.

## Bug

https://user-images.githubusercontent.com/19196536/204693368-d3e96843-8ae1-45fb-bc82-02f9513d1f14.mov

## Fix

https://user-images.githubusercontent.com/19196536/204693495-ff657962-15f4-4530-a648-a5723ffc3fca.mov

### Steps to test

Note: You need to use MacOS to test this. 

1. Reproduce the issue (explained above).
2. Get this branch.
3. Try to reproduce it - you shouldn't be able to.

I see no practical way to write an automated test around this behavior. 🤔 

The launched browser is now correctly focused.

### PR Tasks

### DIFF
--- a/packages/data-context/src/actions/BrowserActions.ts
+++ b/packages/data-context/src/actions/BrowserActions.ts
@@ -48,7 +48,7 @@ export class BrowserActions {
     if (this.ctx.util.isMac) {
       this.ctx.actions.electron.showBrowserWindow()
       this.ctx.actions.electron.hideBrowserWindow()
-      this.focusActiveBrowserWindow()
+      this.focusActiveBrowserWindow().catch(this.ctx.onError)
     }
   }
 
--- a/packages/data-context/src/actions/BrowserActions.ts
+++ b/packages/data-context/src/actions/BrowserActions.ts
@@ -44,6 +44,14 @@ export class BrowserActions {
     await this.browserApi.focusActiveBrowserWindow()
   }
 
+  resetFocusIfMacOS () {
+    if (this.ctx.util.isMac) {
+      this.ctx.actions.electron.showBrowserWindow()
+      this.ctx.actions.electron.hideBrowserWindow()
+      this.focusActiveBrowserWindow()
+    }
+  }
+
   async relaunchBrowser () {
     await this.browserApi.relaunchBrowser()
   }
--- a/packages/server/lib/browsers/index.ts
+++ b/packages/server/lib/browsers/index.ts
@@ -10,6 +10,7 @@ import os from 'os'
 import { BROWSER_FAMILY, BrowserLaunchOpts, BrowserNewTabOpts, FoundBrowser } from '@packages/types'
 import type { Browser, BrowserInstance, BrowserLauncher } from './types'
 import type { Automation } from '../automation'
+import type { DataContext } from '@packages/data-context'
 
 const debug = Debug('cypress:server:browsers')
 const isBrowserFamily = check.oneOf(BROWSER_FAMILY)
@@ -130,7 +131,7 @@ export = {
     return this.getBrowserInstance()
   },
 
-  async open (browser: Browser, options: BrowserLaunchOpts, automation: Automation, ctx): Promise<BrowserInstance | null> {
+  async open (browser: Browser, options: BrowserLaunchOpts, automation: Automation, ctx?: DataContext): Promise<BrowserInstance | null> {
     await kill(true)
 
     _.defaults(options, {
@@ -138,7 +139,7 @@ export = {
       onBrowserClose () {},
     })
 
-    ctx.browser.setBrowserStatus('opening')
+    ctx?.browser.setBrowserStatus('opening')
 
     const browserLauncher = await getBrowserLauncher(browser, options.browsers)
 
@@ -157,7 +158,7 @@ export = {
     // so that there is a default for each browser but
     // enable the browser to configure the interface
     instance.once('exit', async (code, signal) => {
-      ctx.browser.setBrowserStatus('closed')
+      ctx?.browser.setBrowserStatus('closed')
       // TODO: make this a required property
       if (!options.onBrowserClose) throw new Error('onBrowserClose did not exist in interactive mode')
 
@@ -209,7 +210,8 @@ export = {
     if (!options.onBrowserOpen) throw new Error('onBrowserOpen did not exist in interactive mode')
 
     options.onBrowserOpen()
-    ctx.browser.setBrowserStatus('open')
+    ctx?.browser.setBrowserStatus('open')
+    ctx?.actions.browser.resetFocusIfMacOS()
 
     return instance
   },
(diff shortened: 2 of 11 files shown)


packages/server/lib/browsers/index.ts
```diff
@@ -130,15 +131,15 @@ export = {
     return this.getBrowserInstance()
   },
 
-  async open (browser: Browser, options: BrowserLaunchOpts, automation: Automation, ctx): Promise<BrowserInstance | null> {
+  async open (browser: Browser, options: BrowserLaunchOpts, automation: Automation, ctx?: DataContext): Promise<BrowserInstance | null> {
```
lmiller1990, line 134: This is probably NOT allowed to be nullable, but I'm not really sure, so I just made it nullable for now (previously it was `any`, so at least this is more tightly typed).

packages/server/lib/browsers/index.ts
```diff
@@ -209,7 +210,12 @@ export = {
     if (!options.onBrowserOpen) throw new Error('onBrowserOpen did not exist in interactive mode')
 
     options.onBrowserOpen()
-    ctx.browser.setBrowserStatus('open')
+    ctx?.browser.setBrowserStatus('open')
+
+    // work around for a focus MacOS specific issue where the
+    // browser is not correctly focused when launched.
+    // https://github.com/cypress-io/cypress/issues/21743
+    ctx?.actions.browser.resetFocusIfMacOS()
```
lmiller1990, line 218: This is the """fix""".

packages/data-context/src/actions/BrowserActions.ts
```diff
@@ -44,6 +44,14 @@ export class BrowserActions {
     await this.browserApi.focusActiveBrowserWindow()
   }
 
+  resetFocusIfMacOS () {
```
mike-plummer, line 47: /nit: Personal preference, I would put the "this is a workaround" and GH issue link on the function as JSDoc

mike-plummer (approved):
Works well for me., just one small nitpick Nice job tracking down something that works for this, and thanks for adding the additional typings throughout 💯

astone123:
I wasn't able to reproduce the issue on this branch 👍🏻 

One thing that I'm seeing now though, is that once I launch Chrome from the launchpad, I can't find the launchpad window anymore... it is nowhere to be found. I see the Cy icon in my dock, but when I ask it to show me where the launchpad window is, it says that there aren't any windows.

@mike-plummer are you also seeing this? https://www.loom.com/share/f411c08995134eeea27e6e1ba7ed0dec

I think I see the same issue happen in the videos in the PR description. If you look closely you can see the launchpad window disappear once Chrome launches in the fixed video

mike-plummer:
> I wasn't able to reproduce the issue on this branch 👍🏻
> 
> One thing that I'm seeing now though, is that once I launch Chrome from the launchpad, I can't find the launchpad window anymore... it is nowhere to be found. I see the Cy icon in my dock, but when I ask it to show me where the launchpad window is, it says that there aren't any windows.
> 
> @mike-plummer are you also seeing this? [loom.com/share/f411c08995134eeea27e6e1ba7ed0dec](https://www.loom.com/share/f411c08995134eeea27e6e1ba7ed0dec)
> 
> I think I see the same issue happen in the videos in the PR description. If you look closely you can see the launchpad window disappear once Chrome launches in the fixed video

@astone123 Confirmed, I see this too. Didn't notice it in my earlier testing. If I update `BrowserActions` to do a show/hide/show in the `resetFocusIfMacOS` then everything seems to work as expected. Do you see the same?

```
resetFocusIfMacOS () {
    if (this.ctx.util.isMac) {
      this.ctx.actions.electron.showBrowserWindow()
      this.ctx.actions.electron.hideBrowserWindow()
      this.ctx.actions.electron.showBrowserWindow()
      this.focusActiveBrowserWindow().catch(this.ctx.onError)
    }
  }
```

astone123:
@mike-plummer can confirm that this works, only thing is that the Launchpad window flashes over top of the browser for a split second after it launches... I wonder if that's something that we're willing to live with to fix this issue on MacOS? 🤔

lmiller1990:
> @mike-plummer can confirm that this works, only thing is that the Launchpad window flashes over top of the browser for a split second after it launches... I wonder if that's something that we're willing to live with to fix this issue on MacOS? 🤔

Yeah I had considered this but I don't think it feels great... I didn't realize `hide()` prevents the electron app from *re*focusing. Hm.

lmiller1990:
This is actually a bug in macos - I don't think we can "properly" fix it here, without the "flicker".

I logged an issue in electron: https://github.com/electron/electron/issues/36506

lmiller1990:
I am going to close this for now - we can revisit once we decide what we'd like to do, but none of the proposed solutions seems appropriate.

PR #24909, opened 2022-11-30
feat: various v8 snapshot improvements

ryanthemanuel:
### User facing changelog

Further improved the startup time and decreased the size of the binary by using `bytenode` to pre-compile the entry point.

### Additional details

* Bundle and precompile the entry point for the binary. This will improve start up time a bit and shrink the binary size even further. We use [`bytenode`](https://github.com/bytenode/bytenode) as the mechanism to precompile the entry point.
* Establish integrity checks for code within the binary. `bytenode` and `v8` already provide some level of integrity checking on the compiled code it generates, but we also added additional checks to protect the binary even further.

### Steps to test

The best way to test this is with the downloaded binary attached to the last commit of this PR.

n/a

### PR Tasks


.circleci/workflows.yml
```diff
@@ -1985,7 +1985,7 @@ jobs:
       <<: *defaultsParameters
       resource_class:
         type: string
-        default: large
+        default: xlarge
```
ryanthemanuel, line 1932: We're doing a lot in the binary build now. `xlarge` is needed to not run out of memory.

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Cypress is a next generation front end testing tool built for the modern web",
   "private": true,
   "scripts": {
-    "binary-build": "node ./scripts/binary.js build",
+    "binary-build": "cross-env DEBUG=cypress:*snap*info NODE_OPTIONS=--max_old_space_size=16384 node ./scripts/binary.js build",
```
ryanthemanuel: Try and change this to 8192

scripts/binary/binary-cleanup.js
```diff
@@ -114,13 +113,51 @@ const getDependencyPathsToKeep = async (buildAppDir) => {
...
+  return [...Object.keys(esbuildResult.metafile.inputs)].map((input) => `./${input}`)
+}
+
 const cleanup = async (buildAppDir) => {
   // 1. Retrieve all dependencies that still need to be kept in the binary. In theory, we could use the bundles generated here as single files within the binary,
   // but for now, we just track on the dependencies that get pulled in
-  const keptDependencies = [...await getDependencyPathsToKeep(buildAppDir), 'package.json', 'packages/server/server-entry.js']
+  const keptDependencies = [...await getDependencyPathsToKeep(buildAppDir), 'package.json']
 
   // 2. Gather the dependencies that could potentially be removed from the binary due to being in the snapshot
```
ryanthemanuel:
```suggestion
  // 2. Create the entry point bundle and then gather the dependencies that could potentially be removed from the binary due to being in the snapshot or in the entry point bundle
```

scripts/binary/binary-cleanup.js
```diff
@@ -114,13 +113,51 @@ const getDependencyPathsToKeep = async (buildAppDir) => {
   return [...Object.keys(esbuildResult.metafile.inputs), ...entryPoints]
 }
 
+const createServerEntryPointBundle = async (buildAppDir) => {
+  const unixBuildAppDir = buildAppDir.split(path.sep).join(path.posix.sep)
+  const entryPoints = [path.join(unixBuildAppDir, 'packages/server/index.js')]
+  const esbuildResult = await esbuild.build({
```
ryanthemanuel, line 120:
```suggestion
  // Build the binary entry point ignoring anything that happens in the server-entry since that will be in the v8 snapshot
  const esbuildResult = await esbuild.build({
```

scripts/binary/binary-cleanup.js
```diff
@@ -114,13 +113,52 @@ const getDependencyPathsToKeep = async (buildAppDir) => {
...
+    ],
+  })
+
+  console.log(`copying server entry point bundle from ${path.join(workingDir, 'index.js')} to ${path.join(buildAppDir, 'packages', 'server', 'index.js')}`)
+
+  await fs.copy(path.join(workingDir, 'index.js'), path.join(buildAppDir, 'packages', 'server', 'index.js'))
+
+  console.log(`compiling server entry point bundle to ${path.join(buildAppDir, 'packages', 'server', 'index.jsc')}`)
+
+  await bytenode.compileFile({
```
ryanthemanuel, line 140:
```suggestion
  // Use bytenode to compile the entry point bundle. This will save time on the v8 compile step and ensure the integrity of the entry point
  await bytenode.compileFile({
```

scripts/after-pack-hook.js
```diff
@@ -63,7 +66,7 @@ module.exports = async function (params) {
       },
     )
 
-    await setupV8Snapshots(params.appOutDir)
     await cleanup(outputFolder)
+    await setupV8Snapshots({ cypressAppPath: params.appOutDir, integrityCheckSource: getIntegrityCheckSource(outputFolder) })
```
BlueWinds: Why did this move after `cleanup()`?

scripts/binary/binary-cleanup.js
```diff
@@ -114,13 +113,53 @@ const getDependencyPathsToKeep = async (buildAppDir) => {
   return [...Object.keys(esbuildResult.metafile.inputs), ...entryPoints]
 }
 
+const createServerEntryPointBundle = async (buildAppDir) => {
+  const unixBuildAppDir = buildAppDir.split(path.sep).join(path.posix.sep)
+  const entryPoints = [path.join(unixBuildAppDir, 'packages/server/index.js')]
+  // Build the binary entry point ignoring anything that happens in the server-entry since that will be in the v8 snapshot
```
BlueWinds: What is the difference between `server/index.js` and `server-entry.js`? Two different entrypoints for the same package?

scripts/binary/binary-entry-point-source.js
```diff
@@ -0,0 +1,21 @@
...
+
+  Module._extensions['.jsc']({
+    require: module.require,
+    id: filename,
+    filename,
+    loaded: false,
+    path: dirname,
+    paths: Module._nodeModulePaths(dirname),
+  }, filename)
+} catch (error) {
```
BlueWinds, line 18: Isn't logging the error and exiting the process with a non-0 return code the default behavior when an uncaught error is thrown?

scripts/binary/binary-integrity-check-source.js
```diff
@@ -0,0 +1,151 @@
...
+  }
+}
+
+function validateFs (fs) {
+  if (toString.call(fs.readFileSync) !== `function(t,r){const n=splitPath(t);if(!n.isAsar)return g.apply(this,arguments);const{asarPath:i,filePath:a}=n,o=getOrCreateArchive(i);if(!o)throw createError("INVALID_ARCHIVE",{asarPath:i});const c=o.getFileInfo(a);if(!c)throw createError("NOT_FOUND",{asarPath:i,filePath:a});if(0===c.size)return r?"":s.Buffer.alloc(0);if(c.unpacked){const t=o.copyFileOut(a);return e.readFileSync(t,r)}if(r){if("string"==typeof r)r={encoding:r};else if("object"!=typeof r)throw new TypeError("Bad arguments")}else r={encoding:null};const{encoding:f}=r,l=s.Buffer.alloc(c.size),u=o.getFdAndValidateIntegrityLater();if(!(u>=0))throw createError("NOT_FOUND",{asarPath:i,filePath:a});return logASARAccess(i,a,c.offset),e.readSync(u,l,0,c.size,c.offset),validateBufferIntegrity(l,c.integrity),f?l.toString(f):l}`) {
+    throw new Error(`Integrity check failed for fs.readFileSync.toString()`)
+  }
+}
+
+function validateCrypto (crypto) {
```
BlueWinds, line 75: All crypto is invalid. :computer:  :coin: 

(this is a joke because bitcoin and NFTs are a scam, not a serious review comment)

scripts/binary/binary-integrity-check-source.js
```diff
@@ -0,0 +1,151 @@
...
+  const stack = tempError.stack.filter((frame) => !frame.getFileName().startsWith('node:internal') && !frame.getFileName().startsWith('node:electron'))
+
+  OrigError.prepareStackTrace = originalStackTrace
+  Error.stackTraceLimit = originalStackTraceLimit
+
+  if (stack.length !== options.stackToMatch.length) {
+    throw new Error(`Integrity check failed with expected stack length ${options.stackToMatch.length} but got ${stack.length}`)
+  }
+
+  for (let index = 0; index < options.stackToMatch.length; index++) {
```
BlueWinds, line 35: `options.stackToMatch.forEach((expected, index) => {...})`

scripts/binary/binary-sources.js
```diff
@@ -0,0 +1,38 @@
+const fs = require('fs')
+const crypto = require('crypto')
+const path = require('path')
+
+const escapeString = (string) => string.replaceAll(`\``, `\\\``).replaceAll(`$`, `\\$`)
+
+function read ({ file }) {
```
BlueWinds:
```suggestion
function read (file) {
```

scripts/binary/binary-sources.js
```diff
@@ -0,0 +1,38 @@
...
+  const pathToFile = require.resolve(`./${file}`)
+
+  return fs.readFileSync(pathToFile, 'utf8')
+}
+
+const getBinaryEntryPointSource = () => {
+  return read({ file: 'binary-entry-point-source.js' })
```
BlueWinds:
```suggestion
  return read('binary-entry-point-source.js')
```

scripts/binary/binary-sources.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const getIntegrityCheckSource = (baseDirectory) => {
+  const fileSource = read({ file: 'binary-integrity-check-source.js' })
```
BlueWinds:
```suggestion
  const fileSource = read('binary-integrity-check-source.js')
```

scripts/binary/binary-sources.js
```diff
@@ -0,0 +1,38 @@
...
+
+  const mainIndexHash = crypto.createHmac('md5', secret).update(fs.readFileSync(path.join(baseDirectory, './index.js'), 'utf8')).digest('hex')
+  const bytenodeHash = crypto.createHmac('md5', secret).update(fs.readFileSync(path.join(baseDirectory, './node_modules/bytenode/lib/index.js'), 'utf8')).digest('hex')
+  const indexJscHash = crypto.createHmac('md5', secret).update(fs.readFileSync(path.join(baseDirectory, './packages/server/index.jsc'), 'utf8')).digest('hex')
+
+  return fileSource.split('\n').join(`\n  `)
+  .replaceAll('MAIN_INDEX_HASH', mainIndexHash)
+  .replaceAll('BYTENODE_HASH', bytenodeHash)
+  .replaceAll('INDEX_JSC_HASH', indexJscHash)
+  .replaceAll('HMAC_SECRET', secret)
```
BlueWinds, line 29: The step where we begin signing our sources and actively working against people trying to hack (in the original sense of "play around with") our open source project is a step too far, IMO.

I assume there was a product meeting and decisions around this and it's too late to change anyone's mind, but it does leave a sour taste in my mouth regardless.

tooling/v8-snapshot/cache/prod-win32/snapshot-meta.cache.json
```diff
@@ -3931,5 +3936,5 @@
     "./tooling/v8-snapshot/cache/prod-win32/snapshot-entry.js"
   ],
   "deferredHashFile": "yarn.lock",
-  "deferredHash": "5a515f98fe67a8a56a035563ff9c8b9e4a9edc4d035907c3fa5fef1bb60f1dfc"
```
BlueWinds, line 3934: I've been seeing merge conflicts in these snapshot caches fairly frequently recently, as we move towards 12.0.0 release - I know you were looking at having them auto-update in develop, so we wouldn't need to do it manually. I take it that's not quite ready yet?

tooling/v8-snapshot/src/doctor/snapshot-doctor.ts
```diff
@@ -266,6 +266,7 @@ export class SnapshotDoctor {
   private readonly nodeEnv: string
   private readonly _scriptProcessor: AsyncScriptProcessor
   private readonly _warningsProcessor: WarningsProcessor
+  private readonly integrityCheckSource: string | undefined
```
BlueWinds, line 269: There seems to be a fair bit of duplication, adding this same change across multiple files and times in each file.

scripts/binary/binary-integrity-check-source.js
ryanthemanuel: I'm actually purposefully doing a for loop here, to avoid using node internals where possible.

ryanthemanuel: I can add a comment to clarify.

tooling/v8-snapshot/cache/prod-win32/snapshot-meta.cache.json
ryanthemanuel: Yeah not ready yet unfortunately. It'll be one of my next tasks.

scripts/binary/binary-entry-point-source.js
ryanthemanuel: This is actually interesting. The way that this is included by electron, if you don't do this and throw an exception on startup, you get a dialog like this:

image

It seems to be a better experience (you can actually see the whole error) to log the error. Added benefit is we can test the console output fairly easily.

scripts/after-pack-hook.js
ryanthemanuel: I'll comment in this file to clarify.

scripts/binary/binary-entry-point-source.js
BlueWinds: Logging and exiting does seem a lot better than a dialog, yeah. Was just surprised that it doesn't end up in the console regardless.

ryanthemanuel: Yeah I was surprised by it too. My guess is it's not a common occurrence. If your electron app crashes on startup, it's a bad time.

scripts/binary/binary-sources.js
ryanthemanuel: Good call addressed here: [`fe8b6cd` (#24909)](https://github.com/cypress-io/cypress/pull/24909/commits/fe8b6cd8f5df54f9db4b8b87c6fa387217b0bf73)

scripts/binary/binary-sources.js
ryanthemanuel: Good call addressed here: [`fe8b6cd` (#24909)](https://github.com/cypress-io/cypress/pull/24909/commits/fe8b6cd8f5df54f9db4b8b87c6fa387217b0bf73)

scripts/binary/binary-integrity-check-source.js
ryanthemanuel: Added: [`fe8b6cd` (#24909)](https://github.com/cypress-io/cypress/pull/24909/commits/fe8b6cd8f5df54f9db4b8b87c6fa387217b0bf73)

scripts/binary/binary-sources.js
ryanthemanuel: Good call addressed here: [`fe8b6cd` (#24909)](https://github.com/cypress-io/cypress/pull/24909/commits/fe8b6cd8f5df54f9db4b8b87c6fa387217b0bf73)

scripts/binary/binary-cleanup.js
ryanthemanuel: I updated the naming a little bit. I agree it's confusing as is:
Addressed here: [`fe8b6cd` (#24909)](https://github.com/cypress-io/cypress/pull/24909/commits/fe8b6cd8f5df54f9db4b8b87c6fa387217b0bf73)

scripts/after-pack-hook.js
ryanthemanuel: Commented: [`fe8b6cd` (#24909)](https://github.com/cypress-io/cypress/pull/24909/commits/fe8b6cd8f5df54f9db4b8b87c6fa387217b0bf73)

scripts/binary/binary-integrity-check-source.js
```diff
@@ -0,0 +1,168 @@
+const OrigError = Error
+const captureStackTrace = Error.captureStackTrace
+const toString = Function.prototype.toString
+const callFn = Function.call
+
+// eslint-disable-next-line no-unused-vars
+const stackIntegrityCheck = function stackIntegrityCheck (options) {
+  const originalStackTraceLimit = Error.stackTraceLimit
```
ryanthemanuel:
```suggestion
  const originalStackTraceLimit = OrigError.stackTraceLimit
```

scripts/binary/binary-integrity-check-source.js
```diff
@@ -0,0 +1,168 @@
...
+    return stack
+  }
+
+  const tempError = new OrigError
+
+  captureStackTrace(tempError, arguments.callee)
+  const stack = tempError.stack.filter((frame) => !frame.getFileName().startsWith('node:internal') && !frame.getFileName().startsWith('node:electron'))
+
+  OrigError.prepareStackTrace = originalStackTrace
+  Error.stackTraceLimit = originalStackTraceLimit
```
ryanthemanuel:
```suggestion
  OrigError.stackTraceLimit = originalStackTraceLimit
```

scripts/binary/binary-integrity-check-source.js
```diff
@@ -0,0 +1,168 @@
...
+const captureStackTrace = Error.captureStackTrace
+const toString = Function.prototype.toString
+const callFn = Function.call
+
+// eslint-disable-next-line no-unused-vars
+const stackIntegrityCheck = function stackIntegrityCheck (options) {
+  const originalStackTraceLimit = OrigError.stackTraceLimit
+  const originalStackTrace = OrigError.prepareStackTrace
+
+  Error.stackTraceLimit = Infinity
```
ryanthemanuel:
```suggestion
  OrigError.stackTraceLimit = Infinity
```

packages/rewriter/script/worker-shim.js
```diff
@@ -3,13 +3,8 @@ if (process.env.CYPRESS_INTERNAL_ENV === 'production') {
...
-if (require.name !== 'customRequire') {
-  // Purposefully make this a dynamic require so that it doesn't have the potential to get picked up by snapshotting mechanism
-  const hook = './hook'
+const { hookRequire } = require('@packages/server/hook-require')
 
-  const { hookRequire } = require(`@packages/server/${hook}-require`)
-
-  hookRequire(true)
-}
+hookRequire({ forceTypeScript: true })
```
emilyrohrbough, line 8: how do we know to force typescript?

ryanthemanuel: This file never runs in prod (see line 2). It only runs in development where we still need to force typescript compilation on the fly.

scripts/after-pack-hook.js
```diff
@@ -63,7 +66,8 @@ module.exports = async function (params) {
       },
     )
 
-    await setupV8Snapshots(params.appOutDir)
-    await cleanup(outputFolder)
+    // Build out the entry point and clean up prior to setting up v8 snapshots so that the state of the binary is correct
+    await buildEntryPointAndCleanup(outputFolder)
+    await setupV8Snapshots({ cypressAppPath: params.appOutDir, integrityCheckSource: getIntegrityCheckSource(outputFolder) })
```
emilyrohrbough:
```suggestion
    await setupV8Snapshots({
      cypressAppPath: params.appOutDir, integrityCheckSource: 
      getIntegrityCheckSource(outputFolder),
     })
```

scripts/binary/binary-cleanup.js
```diff
@@ -114,13 +113,53 @@ const getDependencyPathsToKeep = async (buildAppDir) => {
   return [...Object.keys(esbuildResult.metafile.inputs), ...entryPoints]
 }
 
-const cleanup = async (buildAppDir) => {
+const createServerEntryPointBundle = async (buildAppDir) => {
+  const unixBuildAppDir = buildAppDir.split(path.sep).join(path.posix.sep)
+  const entryPoints = [path.join(unixBuildAppDir, 'packages/server/index.js')]
```
emilyrohrbough, line 118: Should this be:
```suggestion
  const entryPoints = [unixBuildAppDir, 'packages', 'server' , 'index.js'].join(path.posix.sep)
```

scripts/binary/binary-cleanup.js
```diff
@@ -114,13 +113,53 @@ const getDependencyPathsToKeep = async (buildAppDir) => {
...
+  console.log(`compiling server entry point bundle to ${path.join(buildAppDir, 'packages', 'server', 'index.jsc')}`)
+
+  // Use bytenode to compile the entry point bundle. This will save time on the v8 compile step and ensure the integrity of the entry point
+  await bytenode.compileFile({
+    filename: path.join(buildAppDir, 'packages', 'server', 'index.js'),
+    output: path.join(buildAppDir, 'packages', 'server', 'index.jsc'),
+    electron: true,
+  })
+
+  return [...Object.keys(esbuildResult.metafile.inputs)].map((input) => `./${input}`)
```
emilyrohrbough, line 147: Should this be a hard-coded separator?

scripts/binary/binary-cleanup.js
ryanthemanuel: I am trying to limit node internals references as much as I possibly can. `path.posix.sep` is something we already handle uniquely in the snapshot (since when we're generating the snapshot we don't have access to `path` or any other node internals), so I just decided to reuse what we had available.

See: 

https://github.com/cypress-io/cypress/blob/develop/tooling/v8-snapshot/src/generator/blueprint.ts#L56

and

https://github.com/cypress-io/cypress/blob/develop/tooling/v8-snapshot/src/generator/blueprint.ts#L110

scripts/binary/binary-cleanup.js
ryanthemanuel: Yes, the way these files are read in assumes "require" syntax not platform specific syntax.

scripts/binary/binary-cleanup.js
```diff
@@ -114,13 +113,53 @@ const getDependencyPathsToKeep = async (buildAppDir) => {
...
+}
+
+const buildEntryPointAndCleanup = async (buildAppDir) => {
   // 1. Retrieve all dependencies that still need to be kept in the binary. In theory, we could use the bundles generated here as single files within the binary,
   // but for now, we just track on the dependencies that get pulled in
-  const keptDependencies = [...await getDependencyPathsToKeep(buildAppDir), 'package.json', 'packages/server/server-entry.js']
+  const keptDependencies = [...await getDependencyPathsToKeep(buildAppDir), 'package.json']
```
emilyrohrbough: Can we use promise.all for getting getDependencyPathsToKeep & createServerEntryPointBundle?

```js
const [
  keptDependencies,
  serverPointEntries,
] = Promise.all(getDependencyPathsToKeep(buildAppDir), createServerEntryPointBundle(buildAppDir)]
```

ryanthemanuel: I like this idea. Will work on it.

scripts/binary/binary-entry-point-source.js
```diff
@@ -0,0 +1,21 @@
+const Module = require('module')
+const path = require('path')
+
+process.env.CYPRESS_INTERNAL_ENV = process.env.CYPRESS_INTERNAL_ENV || 'production'
+try {
+  require('./node_modules/bytenode/lib/index.js')
```
emilyrohrbough, line 6: is index.js not the default export?
```suggestion
  require('bytenode/lib/index.js')
```

ryanthemanuel: I'm actually trying to purposefully reference the exact file here. I don't anyone to be able to use any kind of module loading weirdness to reference a different file (may not be possible, but just in case).

scripts/binary/binary-integrity-check-source.js
```diff
@@ -0,0 +1,168 @@
+const OrigError = Error
+const captureStackTrace = Error.captureStackTrace
+const toString = Function.prototype.toString
+const callFn = Function.call
+
+// eslint-disable-next-line no-unused-vars
+const stackIntegrityCheck = function stackIntegrityCheck (options) {
+  const originalStackTraceLimit = OrigError.stackTraceLimit
+  const originalStackTrace = OrigError.prepareStackTrace
```
emilyrohrbough:
```suggestion
  const originalPrepareStackTrace = OrigError.prepareStackTrace
```

emilyrohrbough: Do we need to create each time or can we make these global?

scripts/binary/binary-integrity-check-source.js
```diff
@@ -0,0 +1,168 @@
...
+
+  OrigError.stackTraceLimit = Infinity
+
+  OrigError.prepareStackTrace = function (_, stack) {
+    return stack
+  }
+
+  const tempError = new OrigError
```
emilyrohrbough, line 22: Is is possible clone the class instance so we don't need to reset the overrides?

scripts/binary/binary-integrity-check-source.js
```diff
@@ -0,0 +1,168 @@
+const OrigError = Error
+const captureStackTrace = Error.captureStackTrace
+const toString = Function.prototype.toString
+const callFn = Function.call
+
+// eslint-disable-next-line no-unused-vars
```
emilyrohrbough: is this disable check needed? we are using it below without conditional checks

ryanthemanuel: Ah yeah. This can be removed.

scripts/binary/binary-integrity-check-source.js
```diff
@@ -0,0 +1,168 @@
...
+
+  OrigError.prepareStackTrace = originalStackTrace
+  OrigError.stackTraceLimit = originalStackTraceLimit
+
+  if (stack.length !== options.stackToMatch.length) {
+    throw new Error(`Integrity check failed with expected stack length ${options.stackToMatch.length} but got ${stack.length}`)
+  }
+
+  for (let index = 0; index < options.stackToMatch.length; index++) {
+    const expectedFunctionName = options.stackToMatch[index].functionName
```
emilyrohrbough:
```suggestion
    const { expectedFunctionName, expectedFileName}  = options.stackToMatch[index]
```

scripts/binary/binary-integrity-check-source.js
```diff
@@ -0,0 +1,168 @@
...
+    const actualFunctionName = stack[index].getFunctionName()
+    const expectedFileName = options.stackToMatch[index].fileName
+    const actualFileName = stack[index].getFileName()
+
+    if (expectedFunctionName && actualFunctionName !== expectedFunctionName) {
```
emilyrohrbough, line 40: should we throw if expectedFunctionName or expectedFunctionName are undefined? can we get in this situation?

scripts/binary/binary-integrity-check-source.js
```diff
@@ -0,0 +1,168 @@
...
+
+function validateToString () {
+  if (toString.call !== callFn) {
+    throw new Error('Integrity check failed for toString.call')
+  }
+}
+
+function validateElectron (electron) {
+  if (toString.call(electron.app.getAppPath) !== 'function getAppPath() { [native code] }') {
+    throw new Error(`Integrity check failed for electron.app.getAppPath.toString()`)
```
emilyrohrbough:
```suggestion
    throw new Error(`Integrity check failed for toString.call(electron.app.getAppPath)`)
```

scripts/binary/binary-integrity-check-source.js
```diff
@@ -0,0 +1,168 @@
...
+  }
+}
+
+function validateFs (fs) {
+  if (toString.call(fs.readFileSync) !== `function(t,r){const n=splitPath(t);if(!n.isAsar)return g.apply(this,arguments);const{asarPath:i,filePath:a}=n,o=getOrCreateArchive(i);if(!o)throw createError("INVALID_ARCHIVE",{asarPath:i});const c=o.getFileInfo(a);if(!c)throw createError("NOT_FOUND",{asarPath:i,filePath:a});if(0===c.size)return r?"":s.Buffer.alloc(0);if(c.unpacked){const t=o.copyFileOut(a);return e.readFileSync(t,r)}if(r){if("string"==typeof r)r={encoding:r};else if("object"!=typeof r)throw new TypeError("Bad arguments")}else r={encoding:null};const{encoding:f}=r,l=s.Buffer.alloc(c.size),u=o.getFdAndValidateIntegrityLater();if(!(u>=0))throw createError("NOT_FOUND",{asarPath:i,filePath:a});return logASARAccess(i,a,c.offset),e.readSync(u,l,0,c.size,c.offset),validateBufferIntegrity(l,c.integrity),f?l.toString(f):l}`) {
```
emilyrohrbough, line 69: risk this will ever be different given it is internal node code?

scripts/binary/binary-integrity-check-source.js
```diff
@@ -0,0 +1,168 @@
...
+    throw new Error(`Integrity check failed for fs.readFileSync.toString()`)
```
emilyrohrbough:
```suggestion
    throw new Error(`Integrity check failed for toString.call(fs.readFileSync)`)
```

scripts/binary/binary-integrity-check-source.js
emilyrohrbough: looking below looks like expectedFunctionName depends on the file

scripts/binary/binary-integrity-check-source.js
ryanthemanuel: This is actually electron code not node code and there's not really a good way for me to get at it at package time. I think we are safe to update this as we update electron (if we need to). I'll add a comment for posterity though.

emilyrohrbough (approved):
solid changes 👍🏻 most comments are minor.

scripts/binary/binary-integrity-check-source.js
ryanthemanuel: This function is really only called once. I like the idea of storing this off just in time since we are replacing them immediately after. The other global concepts (Error, captureStackTrace, etc.) are ones that we don't do a replacement dance with.

scripts/binary/binary-integrity-check-source.js
ryanthemanuel: We control what's passed in here so we shouldn't be in that situation.

scripts/binary/binary-integrity-check-source.js
ryanthemanuel: Clarified: [`5c1f981` (#24909)](https://github.com/cypress-io/cypress/pull/24909/commits/5c1f98154e643a23ccddc01e3c02d7700059fcbe)

scripts/binary/binary-integrity-check-source.js
ryanthemanuel: Updated: [`5c1f981` (#24909)](https://github.com/cypress-io/cypress/pull/24909/commits/5c1f98154e643a23ccddc01e3c02d7700059fcbe)

scripts/binary/binary-integrity-check-source.js
ryanthemanuel: Removed: [`5c1f981` (#24909)](https://github.com/cypress-io/cypress/pull/24909/commits/5c1f98154e643a23ccddc01e3c02d7700059fcbe)

scripts/binary/binary-integrity-check-source.js
ryanthemanuel: I think there's some alternative things we can do here, but I think I like this pattern best.

scripts/binary/binary-integrity-check-source.js
ryanthemanuel: Updated: [`5c1f981` (#24909)](https://github.com/cypress-io/cypress/pull/24909/commits/5c1f98154e643a23ccddc01e3c02d7700059fcbe)

scripts/binary/binary-cleanup.js
ryanthemanuel: Updated: [`5c1f981` (#24909)](https://github.com/cypress-io/cypress/pull/24909/commits/5c1f98154e643a23ccddc01e3c02d7700059fcbe)

scripts/after-pack-hook.js
ryanthemanuel: Updated: [`5c1f981` (#24909)](https://github.com/cypress-io/cypress/pull/24909/commits/5c1f98154e643a23ccddc01e3c02d7700059fcbe)

mjhenkes (approved):

scripts/binary/smoke.js
```diff
@@ -201,7 +201,96 @@ const runV8SnapshotProjectTest = function (buildAppExecutable, e2e) {
...
+
+    await fs.writeFile(path.join(buildAppDir, 'index2.js'), `${additionalCode}\nrequire("./index.js")`)
+    await runErroringProjectTest(buildAppExecutable, e2e, 'altering entry point', errorMessage)
+
+    // Restore original state
+    await fs.move(path.join(buildAppDir, 'package.json.bak'), path.join(buildAppDir, 'package.json'), { overwrite: true })
+    await fs.remove(path.join(buildAppDir, 'index2.js'))
+  }
+
+  await testAlteringEntryPoint('console.log("simple alteration")', 'Error: Integrity check failed with expected stack length 9 but got 10')
```
chrisbreiding: Is `Error: Integrity check failed with expected stack length 9 but got 10` the exact error the user will see? I feel like we could add some extra information/guidance, something like "We detected an issue with the integrity of the Cypress binary. It may have been compromised and is not safe to run. We recommend re-installing the Cypress binary with `cypress cache clear && cypress install`."

ryanthemanuel: That's a good call. I can make them a little more user friendly.

ryanthemanuel: Updated: [`5f3a735` (#24909)](https://github.com/cypress-io/cypress/pull/24909/commits/5f3a7351474e241ae20cac5ec488f7bd8a146c19)

scripts/binary/binary-sources.js
ryanthemanuel: I see you have concerns here Blue. Let's continue this discussion offline with the product team.

chrisbreiding (approved):

Issue #24900, opened 2022-11-30
Detached DOM issues in Firefox headless not headed


j-maddo:
### Current behavior

In Cypress versions 10-12, Firefox appears to execute differently between headed and headless versions. Using Firefox 107, all Cypress tests pass in open mode but fail with an assertion error: `"AssertionError: Timed out retrying after 4000ms: Expected to find element: `[data-testid=video-overflow-share]`, but never found it."`  in run mode.  

### Desired behavior

In this situation, where all environmental factors are the same, I would expect the headless version to mirror the headed version and pass all the tests that pass in headed mode.

### Test code to reproduce

Here is a [link](https://github.com/j-maddo/Client-Repro/tree/removeCreds) to my repo where you can reproduce the error. The failing test is in the file: testing-dom__page-reloads/Client-Repro/cypress/e2e/1-getting-started/vidyard.cy.js. 

Upon request, I can give the email and password that I used to sign in to vidyard.com. 

### Cypress Version

Cypress 12 from binary

### Node version

v18.12.0

### Operating System

macOS Ventura 13.0.1

### Debug Logs

```shell
[...]

PR #24924, opened 2022-12-01
fix: Remove unneeded Firefox headless hack (run ci)

BlueWinds:
- Closes https://github.com/cypress-io/cypress/issues/24900

### User facing changelog
Fixes an issue where recent versions of Firefox in headless mode would behave inconsistently with headed.

### Additional details
Back in 2021, Firefox had a quirk where in headless mode it would never gain focus; a workaround was added in https://github.com/cypress-io/cypress/commit/8a4848837aea88f79446c77eeac097b713a0d0a6, forcing the active tab into focus whenever an element in the document gained focus.

However, this results in inconsistent behavior between headed and headless mode in some circumstances. For example, a user might click on a button to open a dropdown, but then Cypress forces focus onto the tab as a whole, causing the dropdown to close again before it can be interacted with.

The workaround no longer appears to be necessary in modern versions of Firefox; both locally and in CI, removing it results in consistent and working behavior.

### Steps to test
See https://github.com/cypress-io/cypress/issues/24900 for a repro. Contact @BlueWinds if you want to run this example locally, she has the login information from the reporter.

No change; behavior is now consistent between headless and headed mode.

### PR Tasks


emilyrohrbough (dismissed):
@BlueWinds are we able to add a driver or system test to verify this is fixed?

mjhenkes (approved):

emilyrohrbough:
Even if we don't get a test in, have you verified this doesn't break older firefox versions? We say we support: Firefox 86 and above.

PR #24925, opened 2022-12-01
feat: v12.0.0 Welcome Page Splashscreen

### LINKED ISSUE #24644: v12 Major Version Splashscreen
Work with Jess to determine V12 welcome / splash screen.

emilyrohrbough:
- Closes #24644
- Related to https://github.com/cypress-io/cypress-services/pull/5034

https://user-images.githubusercontent.com/14099737/205126057-cfdd1f80-539a-4df5-9439-f32330ce6713.mov

### User facing changelog

- Not needed. Updated v12.0.0 Welcome Page content.

### PR Tasks


packages/launchpad/src/migration/MajorVersionWelcome.vue
```diff
@@ -9,64 +9,117 @@
...
+                Detaching Ourselves From Detached Dom Errors
+              </h2>
+
               <p>
-                For a complete list of updates in v11, please review our <ExternalLink
-                  href="https://on.cypress.io/changelog#11-0-0"
-                >
-                  <!--eslint-disable-next-line vue/multiline-html-element-content-newline-->
-                  changelog</ExternalLink>.
+                We have made enhancements to how Cypress manages DOM element resolution to reduce the likelihood of hitting the dreaded detached DOM errors due to maintaining stale DOM references we have all experienced from time-to-time. We've updated our
```
jennifer-shehane: Just a bit of a run on sentence here. 

```suggestion
                We have made enhancements to how Cypress manages DOM element resolution to reduce the likelihood of hitting the dreaded detached DOM errors due to maintaining stale DOM references. We've updated our
```

packages/launchpad/src/migration/MajorVersionWelcome.vue
```diff
@@ -9,64 +9,117 @@
...
+                Testing Multi-Origin WorkFlows
               </h2>
 
               <p>
-                Component Testing is now generally available for projects using React, Next.js, Angular, and Vue!
+                Cypress now has full support for testing multiple domains of different origins in a single test with the <ExternalLink href="https://on.cypress.io/origin">
+                  <code>cy.origin()</code>
+                </ExternalLink> command! To take a deep-dive into how this works, read our
+
+                <ExternalLink href="https://www.cypress.io/blog/2022/12/02/the-journey-behind-cy-origin/">
```
jennifer-shehane: This should be an on link.

jennifer-shehane (changes requested):
Looks good to me other than a couple of updates. Really great content for the first page of v12!

packages/launchpad/src/migration/MajorVersionWelcome.vue
emilyrohrbough: I'll get that on link created.

marktnoonan: You may find this naturally but just in case, there's a with a few of these: https://github.com/cypress-io/cypress-services/blob/develop/packages/on/data/links.yml#L598

packages/launchpad/src/migration/MajorVersionWelcome.vue
```diff
@@ -9,64 +9,117 @@
...
-                href="https://on.cypress.io/changelog#11-0-0"
+                href="https://on.cypress.io/changelog#12-0-0"
                 class="font-bold text-indigo-500"
               >
-                11.0.0
+                12.0.0
               </ExternalLink>
               <span class="font-light pl-10px text-gray-500 text-14px">
-                Released {{ versionReleaseDates['11'] }}
+                Released {{ versionReleaseDates['12'] }}
```
marktnoonan, line 27: We need to manually add the release date, I should put this in the docs. `versionReleaseDates` is a computed value in this same file.

emilyrohrbough: OH I was assuming this was coming from NPM. Good catch.

packages/launchpad/src/migration/MajorVersionWelcome.vue
```diff
@@ -9,64 +9,117 @@
...
               </span>
             </div>
             <div class="children:mb-16px">
+              <p>
+                For a complete list of updates and breaking changes in v12.0.0, please review our
+                <ExternalLink href="https://on.cypress.io/changelog#12-0-0">
+                  changelog
+                </ExternalLink>.
```
marktnoonan: There's a space between `changelog` and the period that's caused by the linebreak before the closing tag for `ExternalLink`

Screen Shot 2022-12-01 at 1 38 37 PM

We have to disable a lint rule to get around it

```html
 <!--eslint-disable-next-line vue/multiline-html-element-content-newline-->
                  changelog</ExternalLink>.
```

Maybe this is a rule that we actually don't want at all and should remove.

packages/launchpad/src/migration/MajorVersionWelcome.vue
marktnoonan: 👍  Mainly doing it this way so it works offline.

packages/launchpad/src/migration/MajorVersionWelcome.vue
emilyrohrbough: @marktnoonan huh, this must have auto-updated because it wasn't the case testing locally. How do I produce that error? I didn't have an error when running lint in this package. Is this a VS code plugin?

emilyrohrbough: fixed: https://github.com/cypress-io/cypress/pull/24925/commits/cabb14cb3cbd558f5ebb8d9910e2e3df9fc8fdca

packages/launchpad/src/migration/MajorVersionWelcome.vue
emilyrohrbough: [`c4975a8` (#24925)](https://github.com/cypress-io/cypress/pull/24925/commits/c4975a881898d30e15fb9d934e5de62ed6ded47c) 

//cc @mjhenkes FYI when we go to release - we will need to verify this date is correct.

packages/launchpad/src/migration/MajorVersionWelcome.vue
marktnoonan: oh that's interesting, I don't show any diffs locally compared to what I pulled down, but I see that the extra space is not present in your video 🤔 

Messing around with window size I was able to get the period on the line below, where you had it on the same line.

Screen Shot 2022-12-01 at 1 46 45 PM

I have no idea why it would not be the same way for you if the line break is the same in your local code, that's unexpected.

Not sure if it's a plugin, I think it's just a setting, I'll check. I get this warning if I try to change it, and then it fixes on save:

Screen Shot 2022-12-01 at 1 52 58 PM

emilyrohrbough: Yeah, I do think theres a plugin at play. but then, I should have seen it locally?? who knows 😄

marktnoonan: I guess it's both, maybe you have something locally that overrides this setting of the eslint extension:

https://github.com/cypress-io/cypress/blob/v12-splashscreen/.vscode/settings.json#L22

Still, it got fixed somewhere before you pushed the commit 🤷 

There's a similar situation at the end of the next paragraph, this doesn't happen there because the period is a part of the link, which I guess is another solution (though I like the period outside the link better).

emilyrohrbough: ewww I'll fix that.

packages/launchpad/src/migration/MajorVersionWelcome.vue
```diff
@@ -9,64 +9,119 @@
...
+                For a complete list of updates and breaking changes in v12.0.0, please review our
+                <!--eslint-disable-next-line vue/multiline-html-element-content-newline-->
+                <ExternalLink href="https://on.cypress.io/changelog#12-0-0">
+                  changelog
+                </ExternalLink>.
+              </p>
+
               <h2 class="font-bold text-18px text-jade-1000">
-                Component Testing Released
+                Testing Multi-Origin WorkFlows
```
marktnoonan:
```suggestion
                Testing Multi-Origin Workflows
```

Capital F looked odd to me?

packages/launchpad/src/migration/MajorVersionWelcome.vue
emilyrohrbough: [`9102eee` (#24925)](https://github.com/cypress-io/cypress/pull/24925/commits/9102eee00867ae367e9e3ab93fc359b3fa67f845)

packages/launchpad/src/migration/MajorVersionWelcome.vue
emilyrohrbough: [`9102eee` (#24925)](https://github.com/cypress-io/cypress/pull/24925/commits/9102eee00867ae367e9e3ab93fc359b3fa67f845)

packages/launchpad/src/migration/MajorVersionWelcome.vue
```diff
@@ -9,64 +9,119 @@
...
+
               <h2 class="font-bold mt-24px mb-16px text-18px text-jade-1000">
-                Faster Startup Time
+                We Now Store The Browser Context, So You Don’t Have To
               </h2>
 
               <p>
-                We have also massively improved our startup performance by shipping a snapshot of our binary instead of the source files. Results will vary based on your situation, but we saw up to 84% faster startup times!
+                The <ExternalLink href="https://on.cypress.io/session">
+                  <code>cy.session()</code>
```
marktnoonan, line 75: possibly the InlineCodeFragment component is better here? It would match the way we display code fragments in the rest of the app

jennifer-shehane (approved):

marktnoonan (approved):
Approving as is, not blocking: think it would be better with `InlineCodeFragment` instead of `code`, but that might be something to get design feedback on.

nagash77 (approved):

PR #24932, opened 2022-12-01
fix: improve keyboard behavior of the inline specs list "run all" buttons

### LINKED ISSUE #24762: Make inline spec list more a11y with Run All Specs experiment
### What would you like?

When using Run All Specs, on the right hand side of the directories in the inline spec list, we should a "Play" button. Right now, you ca tab there, but hitting enter doesn't do anything. It's not a11y.

image

You should be able to tab there and hit enter to Run All Specs in that directory.

### Why is this needed?

_No response_

### Other

_No response_

marktnoonan:
- Closes #24762

### User facing changelog

The "Run N specs" directory-level buttons in the Inline Specs List now function correctly when activated by the keyboard.

### Additional details

This is an incremental improvement in the accessibility of this list so that keyboard behavior is better, even though it doesn't get us everything we'd like for accessibility here and there are further improvements that could be made (for example, nesting the lists to that the tree structure is better represented).

Nesting controls inside other controls should be avoided, but that is something we will need to come back to again as we revisit this experiment and the code around the specs lists more generally. I want to keep the diff small and easy to review.

The reason that keyboard events weren't working correctly previously is that the "directory item" had custom keyboard behavior to open/close a particular folder. Since it was not a button element, keydown event listeners had been added to it to simulate button behavior, and those listeners were capturing keyboard events firing on the internal button, before they had a chance to trigger the internal button's own `click` event.

This PR makes the toggle control for folders an actual button, which means keyboard events for space and enter are "free" since they trigger `click` events, so we can remove those listeners. Then the internal button functions correctly and things are overall a bit simpler.

### Steps to test

With `experimentalRunAllSpecs` turned on, open a spec and use the inline specs list. Arrow keys should move you up and down the list and close/open sections, tab key should let you reach the `run n specs` buttons. Enter and space keys have the following behaviors:

- Focused on a "directory" row, toggle it open and closed
- Focused on a "spec" row, run that spec
- Focused on a "run n specs" button, run the specs

On a project without `experimentalRunAllSpecs`, the list should still function as before.

The one change is that tab key will now also allow moving up and down items. Previously we had disabled this and required arrows for navigation, I think we could consider that again to make it easier to "skip" the list with the keyboard and access other content, but since these are actually interactive elements lets give that some more thought and find a way to communicate that arrow keys are the only way to navigate. As it stands I think it's more correct and predictable to let the browser handle tab order.

Before: space and enter keys when used on the "Run N Specs" button in the Inline Specs List would just toggle the row

https://user-images.githubusercontent.com/8340719/205347610-d0d993b6-8aaa-4579-8ff5-147f91047dd7.mov

After: specs can be run with the keyboard

https://user-images.githubusercontent.com/8340719/205347660-7cfe0834-2617-4d17-8dbf-6e73d97fe47f.mov

### PR Tasks


packages/app/src/specs/InlineSpecListTree.vue
```diff
@@ -124,7 +124,16 @@ const toggle = (row: UseCollapsibleTreeNode<SpecTreeNode<FuzzyFoundSpec>>, idx:
   row.toggle()
 }
 
-const submit = (row: UseCollapsibleTreeNode<SpecTreeNode<FuzzyFoundSpec>>, idx: number) => {
+const submitOrToggle = (row: UseCollapsibleTreeNode<SpecTreeNode<FuzzyFoundSpec>>, idx: number, event?: KeyboardEvent) => {
```
marktnoonan: Renamed this since it has condition logic that doesn't always "submit" the spec.

packages/app/src/specs/InlineRunAllSpecs.vue
```diff
@@ -6,6 +6,7 @@
     <button
       class="flex h-full w-full items-center justify-center"
       data-cy="run-all-specs-button"
+      data-run-all-specs-button
```
mike-plummer: In most places we use the `data-cy="..."` pattern - any reason you're just using a data attribute here?

packages/app/src/specs/InlineSpecList.cy.tsx
```diff
@@ -222,5 +222,30 @@ describe('InlineSpecList', () => {
...
+
+      // make sure typing didn't change displayed items
+      cy.findAllByTestId('spec-file-item').should('have.length', 4)
+      cy.findAllByTestId('run-all-specs-button').eq(1)
+      .focus()
+      .type('{enter}')
+
+      // make sure typing didn't change displayed items
+      cy.findAllByTestId('spec-file-item').should('have.length', 4)
+      cy.get('@mutationStub').should('have.been.calledThrice')
```
mike-plummer, line 250: I must be missing something - I see the `space` and `enter` invocations of Run All, what's the third?

mike-plummer (approved):
Everything works as described for me, just a couple questions

packages/app/src/specs/InlineRunAllSpecs.vue
marktnoonan: Oh I can pull this out now, removed. For a second I was using this in one of the higher level listeners to avoid reacting to events fired on this button element, before I realized we didn't need those listeners themselves.

packages/app/src/specs/InlineSpecList.cy.tsx
marktnoonan: Before the space, there's a click which does the same thing. Then it already has focus for the type. Then for `{enter}` we artificially focus, to avoid Cypress doing it's own click.

Not sure I really need to test the click here as such, but it's sorta nice to assert all 3 in one test.

packages/app/src/specs/InlineRunAllSpecs.cy.tsx
```diff
@@ -23,6 +25,15 @@ describe('<InlineRunAllSpecs/>', () => {
...
       })
     })
+
+    it('emits expected event on click and with keyboard', () => {
+      cy.get('button')
+      .click()
+      .type(' ')
+      .type('{enter}')
+
+      cy.get('@runAllStub').should('have.been.calledThrice')
```
lmiller1990: Can we just create a fresh stub? This test relies on other tests to pass - deleting or changing the order of those would break this test, which isn't good - I think the tests should be able to run in any order, independently of each other.

packages/app/src/specs/InlineSpecList.cy.tsx
lmiller1990, line 250: Same here.

I think we should delete one of the two tests you've added? Maybe `InlineRunAllSpecs.cy.tsx` - they test *exactly* the same thing, but this one covers a bit more.

lmiller1990 (commented):
Two comments, feature looks good, tests can be better (I know you didn't originally write the tests, but good to fix them up while we are here.

packages/app/src/specs/InlineRunAllSpecs.cy.tsx
marktnoonan: It relies on the stub from the `beforeEach`, since that's where the `mount` happens, so the stub is freshly created each time. The three calls are all generated within this block before the assertion:

```js
cy.get('button')
.click()
.type(' ')
.type('{enter}')
```

I don't think there's a reliance on any other test? Unless you mean that deleting/rearranging the click and typing steps could cause the test to fail, which is true, but I think in-line with what we already do.

packages/app/src/specs/InlineSpecList.cy.tsx
marktnoonan: The `InlineRunAllSpecs` component is responsible for managing its own keyboard behavior, so ideally it should be tested there. The component was already working correctly, so in this PR I'm just adding a case that I thought was missing from the original.

`InlineRunAllSpecs` is used by 2 other components, one of which happened to break keyboard behavior because it's nesting controls in a non-standard way. So the test here is more about adding coverage for the bug we're fixing - something that fails on the old code and passes on this branch. 

But I think it's confusing if this is the only coverage of the behavior itself, we should be able to delete this spec without losing coverage on other components that have their own component tests.

I'll add a note to this test that we can remove it at a future point when we are no longer nesting a button within a button. Actually I'll make the issue for this first and then comment with the issue number - we can improve both specs lists by using a similar pattern to our card component, where the target that receives keyboard focus is a real button inside the larger element, even though the focus style and click listeners apply to the whole element. Then we can have internal controls in a row that aren't nested inside other controls, and it fixes a big problem where the whole row is announced as a label.

A bigger change than I wanted to make in this PR (keyboard behavior is still a win), but would solve the problems with the current implementation.

marktnoonan: That said, I did simplify the `InlineRunAllSpecs` to not repeat the keyboard stuff specifically. Since we already confirm it's a button element, the behavior on space and enter keys is free since a click is emitted by them, so we keep the keyboard specific stuff in this test, where it was broken, and avoid duplicating in the other component test. We still want to test the event emitted by that component.

Also made https://github.com/cypress-io/cypress/issues/24966 and added a TODO comment for future cleanup.

packages/app/src/specs/InlineSpecListTree.vue
```diff
@@ -18,22 +18,22 @@
...
         :data-selected-spec="isCurrentSpec(row.data)"
-        @click.self="submit(row.data, row.index)"
+        @click.self="submitOrToggle(row.data, row.index)"
       >
-        <RouterLink
+        <component
+          :is="row.data.isLeaf ? RouterLink : 'button'"
           :ref="el => setItemRef(el, row.index)"
           :key="row.data.data?.absolute"
-          :tabindex="isTabbable(row, row.index) ? '0' : '-1'"
```
ZachJW34, line 26: You mention: 

> Previously we had disabled this and required arrows for navigation

but the row items were specifically disabled except for the active spec, allowing the user to tab and land straight onto the active spec. With the addition of the inline button on the directories this behavior admittedly broke down but your PR is removing it all together. I don't feel strongly about keeping this behavior, but wanted to make it clear why it was wired up like that.

Short video of what the original behavior was intended for


https://user-images.githubusercontent.com/25158820/205675066-8a0e7fda-f355-4f4d-a5ff-1d85b507d68d.mov

ZachJW34 (approved):

packages/app/src/specs/InlineSpecListTree.vue
marktnoonan: Thanks for pointing this out - I remembered the convenience of that but to keep it, what we'd really want to do is give the treeview the proper semantics to where the behavior would make sense when announced in a screenreader. Something like https://www.w3.org/WAI/ARIA/apg/example-index/treeview/treeview-navigation looks correct, but doesn't have good support info here https://a11ysupport.io/tech/aria/tree_role so a bit more research would be a good idea.

I would love to consider some high quality keyboard shortcuts to, for example, move focus between panels. It would help everybody to deal with the sheer number of controls on the screen. Created https://github.com/cypress-io/cypress/issues/24973

In the meantime I think it's slightly better to do the default browser behavior for these elements.

marktnoonan: Oh dang it took a second for this to click, I see that was a whole extra behavior on top of what I was thinking - where we still had a single spec focused.

When we make this a proper treeview I think we should put back that behavior, it seems useful for the cheap cost of 1 extra tab stop.

PR #24935, opened 2022-12-02
feat(12): change testIsolation values from on/off to true/false

emilyrohrbough:
After, sync with @brian-mann we decided `true` and `false` was more appropriate values for `testIsolation` instead of `on` and `off`.

### User facing changelog
Changes the testIsolation configuration values from `on` or `off` to `true` or `false`.

- [TODO] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 


mjhenkes (approved):

cli/types/cypress.d.ts
```diff
@@ -2913,23 +2913,23 @@ declare namespace Cypress {
...
      * Tradeoffs:
      *      Turning test isolation off may improve performance of end-to-end tests, however, previous tests could impact the
      *      browser state of the next test and cause inconsistency when using .only(). Be mindful to write isolated tests when
-     *      test isolation is off. If a test in the suite impacts the state of other tests and it were to fail, you could see
+     *      test isolation is false. If a test in the suite impacts the state of other tests and it were to fail, you could see
      *      misleading errors in later tests which makes debugging clunky. See the [documentation](https://on.cypress.io/test-isolation)
      *      for more information.
      *
-     * @default 'on'
+     * @default 'true'
```
chrisbreiding:
```suggestion
     * @default true
```

AtofStryker (approved):

packages/config/src/options.ts
```diff
@@ -381,11 +381,11 @@ const driverConfigOptions: Array<DriverConfigOption> = [
     overrideLevel: 'any',
   }, {
     name: 'testIsolation',
-    defaultValue: 'on',
+    defaultValue: true,
     validation: (key: string, value: any, opts: ValidationOptions) => {
       const { testingType } = opts
 
-      let configOpts = ['on', 'off']
+      let configOpts = [true, false]
```
chrisbreiding, line 388: I feel like this should use `validate.isBoolean`, but for component testing, would need add either `validate.isTrue` or `validate.isExactValue`.

chrisbreiding (changes requested):

chrisbreiding (approved):

packages/config/src/options.ts
emilyrohrbough: Looking at this, `validate.isBoolean` accepts a null value, and we only want true & false. Not sure it's the best option for this?

PR #24952, opened 2022-12-02
fix(breaking): exclude node_modules from cross-origin-callback-loader

mschile:
- Closes n/a

### User facing changelog

* `cy.origin` dependencies within a third-party plugin is no longer supported.

### Additional details

Previously we were including `node_modules` when scanning for `cy.origin` dependencies to ensure any third party plugins/imports would be properly loaded. However, this caused large memory spikes and performance degradation when parsing files with large dependency graphs. Thus, we are going to exclude `node_modules` until we can come up with a more optimized solution.

### Steps to test

The added system test uses a `cy.origin` dependency within a `node_modules` plugin and now fails.

* `cy.origin` dependencies within a third-party plugin will no longer work.

Screen Shot 2022-12-02 at 12 11 54 PM

### PR Tasks

- [Not Yet] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 


system-tests/test/cy_origin_error_spec.ts
```diff
@@ -72,4 +72,18 @@ describe('e2e cy.origin errors', () => {
       expect(res.stdout).to.contain('Using `require()` or `import()` to include dependencies requires enabling the `experimentalOriginDependencies` flag and using the latest version of `@cypress/webpack-preprocessor`')
     },
   })
+
+  systemTests.it.only('errors when using a plugin that has a custom command that uses cy.origin with a dependency', {
```
chrisbreiding:
```suggestion
  systemTests.it('errors when using a plugin that has a custom command that uses cy.origin with a dependency', {
```

chrisbreiding (approved):

mjhenkes (approved):

system-tests/projects/origin-dependencies/.gitignore
```diff
@@ -0,0 +1 @@
+!node_modules
```
ryanthemanuel:
```suggestion
!node_modules

```

PR #24954, opened 2022-12-02
fix: add default export detection

### LINKED ISSUE #24929: Add support for creating specs for React components that are exported as default
For the Create from React Component feature, we're using [react-docgen](https://github.com/reactjs/react-docgen) to parse files and return any React components that are defined and exported. Unfortunately, this tool does not tell us whether a component is a named export or the default export.

This is a problem because when we create the spec file, we need to either import the component by name (`import { MyComponent } from 'file'`) or by default (`import DefaultComponent from 'file'`). Right now, the spec template assumes that the component is a named export, so the spec will fail to run if it is exported by default.

There is also an edge-case with `export { Foo as Bar }` aliasing where `react-docgen` returns information about `Foo` but `import {Foo} from '...'` would not work correctly.

We should be able to use a custom resolver with `react-docgen` to add information to the response about whether each component is a named or default export. We should then be able to use this information to craft the correct import statement in the generated spec file.

ZachJW34:
- Closes #24929

### User facing changelog
na

### Additional details
Allows the detection of default exports, as well as exports that have been aliased e.g. `export { Foo as bar }`. There are some edge-cases that aren't covered like HOCs (`export default connect(Counter)`) but adding support for those would be complex.

### Steps to test
Run the system-tests for data-context `yarn workspace @packages/data-context test`

### PR Tasks


lmiller1990:
> HOCs (export default connect(Counter)

I think this is fine to skip for the initial release, talked to @astone123 but we should just ignore non capital exports - so HOCs are off the table, at least for now.

packages/data-context/src/actions/CodegenActions.ts
```diff
@@ -19,12 +22,20 @@ export class CodegenActions {
...
+
       // types appear to be incorrect in react-docgen@6.0.0-alpha.3
       // TODO: update when 6.0.0 stable is out for fixed types.
       const defs = (Array.isArray(result) ? result : [result]) as ReactComponentDescriptor[]
 
-      return defs
+      const mappedDefs = defs.map((meta) => {
+        const displayName = meta.displayName || ''
+
+        return { ...meta, displayName, ...linker[displayName] }
```
lmiller1990: Do we want to spread over `...meta`? It contains additional fields we don't care about:

```js
mappedDefs: [
    {
      description: '',
      displayName: 'hw',
      methods: [],
      exportName: 'HelloWorld',
      isDefault: false
    }
  ]
```

What about just returning the relevant data?

packages/data-context/src/actions/CodegenActions.ts
```diff
@@ -19,12 +22,20 @@ export class CodegenActions {
...
+      const linker: Linker = {}
+      let result = parseReactComponent(src, findAllWithLink(linker))
+
       // types appear to be incorrect in react-docgen@6.0.0-alpha.3
       // TODO: update when 6.0.0 stable is out for fixed types.
       const defs = (Array.isArray(result) ? result : [result]) as ReactComponentDescriptor[]
 
-      return defs
+      const mappedDefs = defs.map((meta) => {
+        const displayName = meta.displayName || ''
```
lmiller1990: We now have `exportName` **and** `displayName`. It looks like we never actually want `displayName` - we only ever care about `exportName`, since that's the thing we import and pass to `mount`.

I understand it's a bit of extra work, but should we just drop `displayName` entirely? It doesn't serve any purpose here. We should just use `exportName` in all cases, shouldn't we?

packages/data-context/src/actions/CodegenActions.ts
```diff
@@ -139,3 +150,70 @@ export class CodegenActions {
     return WIZARD_FRAMEWORKS.find((framework) => framework.configFramework === config?.component?.devServer.framework)
   }
 }
+
+type Linker = Record<string, { exportName: string, isDefault: boolean }>
```
lmiller1990: I haven't seen the term "linker" before - should this be called `resolver`? That's why the types call it.

packages/data-context/src/actions/CodegenActions.ts
```diff
@@ -19,12 +22,20 @@ export class CodegenActions {
     try {
       const src = await this.ctx.fs.readFile(filePath, 'utf8')
 
-      const result = parseReactComponent(src, reactDocgenResolvers.findAllExportedComponentDefinitions)
+      const linker: Linker = {}
```
lmiller1990: Shame about mutating the arg here :( I guess this is the babel way of doing things.

Just a style thing, but I wonder if `Map` would be more descriptive. `const linker: Map<string, ...>()`.

lmiller1990 (commented):
Some thoughts... I think we should eliminate `displayName`, superseded by `exportName`.

packages/data-context/src/actions/CodegenActions.ts
astone123: yeah, we only need `exportName`

packages/data-context/src/actions/CodegenActions.ts
ZachJW34: Narrowed it down to:
```ts
interface ReactComponentDescriptor {
  exportName: string
  isDefault: boolean
}
```

packages/data-context/src/actions/CodegenActions.ts
ZachJW34: Map looks nicer!

packages/data-context/src/actions/CodegenActions.ts
lmiller1990: Yes, great!

lmiller1990 (approved):

packages/data-context/src/actions/CodegenActions.ts
ZachJW34: Addressed in edca01e60f1a06af1981f0956fe550f58644ec12!

packages/data-context/src/actions/CodegenActions.ts
ZachJW34: Addressed in edca01e60f1a06af1981f0956fe550f58644ec12!

packages/data-context/src/actions/CodegenActions.ts
ZachJW34: Addressed in edca01e60f1a06af1981f0956fe550f58644ec12!

packages/data-context/src/actions/CodegenActions.ts
ZachJW34: Addressed in edca01e60f1a06af1981f0956fe550f58644ec12!

packages/data-context/test/unit/codegen/spec-options.spec.ts
```diff
@@ -130,13 +130,15 @@ describe('spec-options', () => {
...
             componentName: 'Counter',
+            isDefault: true,
           })
 
           const result = await testSpecOptions.getCodeGenOptions()
 
           expect(result.codeGenType).to.eq('component')
           expect(result.templateKey).to.eq('reactComponent')
           expect(result.fileName).to.eq('Counter.cy.tsx')
+          // expect(result.importStatement).to.eq()
```
ZachJW34:
```suggestion
```

remove

astone123 (approved):
Nice 👍🏻 I'm going to add the arguments to the GraphQL mutation in my [UI PR](https://github.com/cypress-io/cypress/pull/24982)

packages/data-context/src/actions/CodegenActions.ts
```diff
@@ -151,8 +158,9 @@ export class CodegenActions {
   }
 }
 
-type Linker = Record<string, { exportName: string, isDefault: boolean }>
-function findAllWithLink (linker: Linker) {
+type ExportResolver = Map<string, { exportName: string, isDefault: boolean }>
```
lmiller1990: Just pass `ReactComponentDescriptor`, same interface!

```ts
type ExportResolver = Map<string, ReactComponentDescriptor>
```

lmiller1990 (approved):
Sorry, one more nit - I'll approve, though :heavy_check_mark:

packages/data-context/src/actions/CodegenActions.ts
astone123: Nice catch 👍🏻

PR #24965, opened 2022-12-05
fix(vite-dev-server): ensure assets are correctly reloaded

### LINKED ISSUE #24874: the style update is failing after In the Vue3 component test
### Current behavior

I used tailwind CSS for my project

image

The first time you enter the cypress component test it's fine

image

When I change the class, the cypress component test does not update the view

image

Changed bg-red-500 to bg-blue-500

When you go back to cypress and click the refresh button, you will see that the style is invalid
image

### Desired behavior

The correct behavior should be after updating the component's class
you should see the correct result in the cypress component test, where the background turns blue

### Test code to reproduce

repo: https://github.com/cuixiaorui/cypress-issue-for-vue3-compoment-test-hot-update

### Cypress Version

11.2.0

### Node version

v17.5.0

### Operating System

macOS 13.0.1 (22A400)

### Debug Logs

_No response_

### Other

_No response_

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/24874

### User facing changelog

Fix a bug where updating a component would not trigger the `supportFile` to reload imported stylesheets. Now we do a `full-reload` for Vite, to ensure we re-run the spec correctly with the latest styles.

### Additional details

I don't think we even want a partial hot reload - it's more consistent with our testing philosophy to always get a refresh slate, so a full reload makes sense. The perf diff is negligible, and only in interactive mode - it won't impact CI runs at all.

### Steps to test

You can grab a binary and test it with the reproduction provided in the linked issue. Here's the binaries: https://github.com/cypress-io/cypress/commit/e1d766ce9fa8ce9d3803f634f91e5104d3e17140#comments

Correctly reloads latest stylesheets, even if it's import in `support/component.js`. Before this wouldn't work - the latest styles wouldn't load, and you would be stuck with an unstyled component.

https://user-images.githubusercontent.com/19196536/205777402-9f255c91-f121-4e3f-8ae8-f30a47dc58fd.mp4

### PR Tasks


npm/vite-dev-server/src/plugins/cypress.ts
```diff
@@ -133,13 +133,15 @@ export const Cypress = (
         // as soon as we find one of the specs, we trigger the re-run of tests
         for (const mod of moduleImporters.values()) {
           debug('handleHotUpdate - mod.file', mod.file)
+          // always re-run on supportFile changed
           if (mod.file === supportFilePath) {
-            debug('handleHotUpdate - support compile success')
-            devServerEvents.emit('dev-server:compile:success')
```
ZachJW34: I was under the impression that `devServerEvents.emit('dev-server:compile:success')` is similar to `server.ws.send( type: 'full-reload' })` as it destroys the AUT and creates a new one.

Seems like we have two mechanisms here for causing a reload, the `full-reload` and the `dev-server:compile:success` event. Do we know why the `dev-server:compile:success` wasn't working in this case?

lmiller1990: I _think_ `dev-server:compile:success` just causes the spec to re-run (using any updated code via HMR). `full-reload` is a Vite client API which tells Vite to actually go and get the latest resources in their entirety (since HMR isn't working for the stylesheets).

So this change basically means we aren't using HMR for Vite anymore, but doing a full reload. This seems more correct imo - fresh slate every test WDYT?

HMR will be useful if we have a "workbench" mode one day like Storybook... but not so much for tests.

ZachJW34: This change does fix the linked issue but I realized that we are going to have this same issue inside of specs. To reproduce, move the `import '../../src/styles.css'` from the support file into a spec file (`App.cy.ts` from the reproduction).

The update of a spec relies on the `devServerEvents.emit('dev-server:compile:success')` rather than the ws reload so it causes the same problem.

I was messing around with it and I was able to fix this by changing `return []` to `return`, not sure the implications of this. It just feels strange that we have a system for rerunning a spec when something changes `dev-server:compile:success` which should work how we want it to since it destroys the AUT and should trigger a refresh but it doesn't work for some reason..

lmiller1990: Hmmm I think the problem might be Vite caching, then? :thinking: 

`dev-server:compile:success` triggers a *re-run* but has nothing to do with fetching the latest data.

https://github.com/cypress-io/cypress/blob/develop/packages/app/src/runner/event-manager.ts#L153-L156

Right now it's 

- HMR -> get latest data (spec, component) -> `dev-server:compile:success` -> `rerun()`.

So the problem is Vite - I think it's cache is not getting updated, so it uses the stale one. I think for reliability we should probably just be using `full-reload` for Vite exclusively? It's more in line with the "fresh slate" and "test isolation" philosophy, for sure.

What do you think?

lmiller1990: I think you are right @ZachJW34, I will double down and find out why the HMR trigger doesn't work.

lmiller1990: I investigated. I think we are NOT using HMR what-so-ever right now. We refetch all assets every time the spec re-runs.

According to the [Vite HMR docs](https://vitejs.dev/guide/api-plugin.html#handlehotupdate).

> Return an empty array and perform complete custom HMR handling by sending custom events to the client

We currently return an empty array, but we don't do any custom HMR (like `import.meta.hot...`). The assets *are* rebuilt (the JS modules in memory) but we don't do any HMR updates. We just emit the `dev-server:compile:success`, which casuse Cypress to re-run the spec, which re-requests the assets.

We have a few options to fix this. 

- Just be to change `return []` to `return`. This seems to work just fine.
- use the new `reloadModule` API.

```
 /**
   * Triggers HMR for a module in the module graph. You can use the `server.moduleGraph`
   * API to retrieve the module to be reloaded. If `hmr` is false, this is a no-op.
   */
  reloadModule(module: ModuleNode): Promise<void>
```

I tried this the new `reloadModule`, I got it working with `if (mod.file.endsWith('css')) { reloadModule(mod) }`. Although this probably isn't what we want - it's more for forcing a virtual module to reload. It's also not available until Vite 3.2.0 - which seems to NOT be a straight forward upgrade :(

lmiller1990: I think we probably haven't implemented the Vite Dev Server ideally - we might be able to better leverage HMR. For now, WDYT of my fix? It seems in line with how Vite recommends using `handleHotModule`, and it seems to work great.

ZachJW34: I did some poking around of [Vite's HMR](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/hmr.ts#L100) and I can definitely see how `return []` could cause some cache issues. Looks like returning an empty array essentially tells Vite that there is no new module and the previous module isn't invalidated. Maybe this is only a problem with virtual modules which is why we are seeing this with Tailwind.

I'm cool with the fix you have here! Not sure why the vite-dev-server tests crapped out but I'll rerun them.

npm/vite-dev-server/cypress/e2e/vite-dev-server.cy.ts
```diff
@@ -3,6 +3,58 @@
...
+        import React from 'react'
+        export const App = () => {
+          return (
+            <div className='bg-blue-100'>
+              Hello
+            </div>
+          )
+        }
+        `,
+      )
```
ZachJW34: Why is this needed if you're only using a component defined within the the spec?

ZachJW34 (approved):

npm/vite-dev-server/src/plugins/cypress.ts
lmiller1990: Maybe we move to `reloadModule` once we update to Vite 3.2. That's supposed to be for virtual modules, like Tailwind (I assume this is how it's implemented, anyway).

rockindahizzy (approved):

npm/vite-dev-server/cypress/e2e/vite-dev-server.cy.ts
lmiller1990: Oops, it's not. I'll remove it.

lmiller1990: Done

PR #24975, opened 2022-12-05
fix: .within() now throws an error if given more than one subject.

BlueWinds:
### User facing changelog
BREAKING: `.within()` now throws an error if given more than one subject.

This brings the behavior in line with the documentation, and adds consistency around how `.within()` behaves across commands. Some commands inside a `within` block would silently select the first element, while others would use all of them, and still others would throw an error.

This ambiguity is now resolved - all commands use the first subject element, because only one element is allowed.

### Steps to test
See the new test added in this PR for an example of the change in action.

A visual example of the error thrown.

image

### PR Tasks
- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4898

### DIFF
--- a/packages/app/cypress/e2e/runner/sessions.ui.cy.ts
+++ b/packages/app/cypress/e2e/runner/sessions.ui.cy.ts
@@ -106,6 +106,7 @@ describe('runner/cypress sessions.ui.spec', {
     validateSessionsInstrumentPanel(['blank_session'])
 
     cy.get('.command-name-session')
+    .first()
     .within(() => {
       cy.contains('blank_session')
       cy.contains('failed')
--- a/packages/app/cypress/e2e/runner/sessions.ui.cy.ts
+++ b/packages/app/cypress/e2e/runner/sessions.ui.cy.ts
@@ -302,6 +302,7 @@ describe('runner/cypress sessions.ui.spec', {
       validateSessionsInstrumentPanel(['user1'])
 
       cy.get('.command-name-session')
+      .first()
       .within(() => {
         cy.contains('failed')
 
--- a/packages/app/cypress/e2e/runs.cy.ts
+++ b/packages/app/cypress/e2e/runs.cy.ts
@@ -941,7 +941,7 @@ describe('App: Runs', { viewportWidth: 1200 }, () => {
       cy.get('[data-cy="run-card-icon-RUNNING"]').should('have.length', 3).should('be.visible')
       cy.wrap(obj).invoke('toCall')
 
-      cy.get('[data-cy="run-card-icon-PASSED"]').should('have.length', 3).should('be.visible').within(() => {
+      cy.get('[data-cy="run-card-icon-PASSED"]').should('have.length', 3).should('be.visible').first().within(() => {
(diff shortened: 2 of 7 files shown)


packages/app/cypress/e2e/runner/sessions.ui.cy.ts
```diff
@@ -106,6 +106,7 @@ describe('runner/cypress sessions.ui.spec', {
     validateSessionsInstrumentPanel(['blank_session'])
 
     cy.get('.command-name-session')
+    .first()
```
BlueWinds, line 109: There are a couple of examples where our tests used to work, but required a change because of this PR. For example, this test had multiple commands - we now are explicitly (rather than implicitly) asserting on the first of them.

packages/launchpad/cypress/e2e/project-setup.cy.ts
```diff
@@ -87,12 +87,10 @@ describe('Launchpad: Setup Project', () => {
     cy.contains('h1', 'Configuration files')
     cy.findByText('We added the following files to your project:')
 
-    cy.get('[data-cy=valid]').within(() => {
-      cy.contains('cypress.config.js')
-      cy.containsPath('cypress/support/e2e.js')
-      cy.containsPath('cypress/support/commands.js')
-      cy.containsPath('cypress/fixtures/example.json')
-    })
+    cy.get('[data-cy=valid]').as('valid').contains('cypress.config.js')
```
BlueWinds, line 90: Here's an example of where we were genuinely relying on multiple subjects from `.within()`, and it required something more than just `.first()` to fix the test.

packages/app/cypress/e2e/runs.cy.ts
```diff
@@ -941,7 +941,7 @@ describe('App: Runs', { viewportWidth: 1200 }, () => {
       cy.get('[data-cy="run-card-icon-RUNNING"]').should('have.length', 3).should('be.visible')
       cy.wrap(obj).invoke('toCall')
 
-      cy.get('[data-cy="run-card-icon-PASSED"]').should('have.length', 3).should('be.visible').within(() => {
+      cy.get('[data-cy="run-card-icon-PASSED"]').should('have.length', 3).should('be.visible').first().within(() => {
```
emilyrohrbough:
```suggestion
      cy.get('[data-cy="run-card-icon-PASSED"]')
      .should('have.length', 3)
      .should('be.visible')
      .first().within(() => {
```

marktnoonan (approved):
Works well, I get the error when there's more than 1 subject.

chrisbreiding (approved):
Looks good! I like the error message.

mjhenkes (approved):

mschile (approved):
Should we have a test that explicitly tests our recommendation of use `.each`?

BlueWinds:
> Should we have a test that explicitly tests our recommendation of use `.each`?

I don't see the value. Each command is tested individually - we know they all work. :woman_shrugging:

PR #24982, opened 2022-12-05
feat: create from React component UI

### LINKED ISSUE #24773: Build 'create from React component' UI
Create from React Component requires a slightly different user flow from Create from Vue Component. This is because we need to parse the selected file to determine whether there are any React components defined in the file that we can create a spec for.

Create all of the UI necessary to match the designs [in Figma](https://www.figma.com/file/gq6zQUQ1bx2uqfnF64lwX9/Create-from-React-Component?node-id=0%3A1&t=Psd6euLyYiR2aCKd-0) and mock the API interactions if it isn't completed yet.

Most of the UI is the same as Create from Vue except for the "Select a Component" view. This view should be created with a loading (for when we are parsing the file) and empty state (for when there are no React components defined in the file).

See the [Create from React Component tech brief](https://github.com/cypress-io/tech-briefs/tree/main/briefs/ct-create-from-react-component) for guidance

astone123:
- Closes #24773

### User facing changelog

n/a (this can go in the feature branch PR)

### Additional details

This PR adds the base UI for Create from React component.

### Steps to test

- Open CT with a React project (that doesn't use a custom dev server - we don't support this yet)
- Click New Spec button in the top-right corner
- Select create from component
- See the list of files that matches the file glob (excluding config files and Cypress-related files)
- Expand some of the rows
- See that if there are no React components exported from a given file that you see "No components found" in the expanded content
- See that if there are React components exported from a file that you see those components listed in the expanded content
- Verify that clicking on the component generates a spec file that correctly mounts the component

Verify that the design matches [Figma](https://www.figma.com/file/gq6zQUQ1bx2uqfnF64lwX9/Create-from-React-Component?node-id=84%3A9323&t=9HnmHXvRVAef5cxm-0) (minus the "choose a file name for this spec" which will be added with #25005)

https://www.loom.com/share/bfe89f3314f5441f85d5ff97b9667140

### PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4907

### Outstanding Items

- [x] Exclude spec and config files from file chooser results
~- [ ] Add screen to allow the user to choose spec path and name~ https://github.com/cypress-io/cypress/issues/25005
- [x] The list of files is getting cut off at the bottom
- [x] Tests


packages/app/src/specs/generators/ExpandableFileChooser.vue
```diff
@@ -0,0 +1,166 @@
...
+
+import NoResults from '@cy/components/NoResults.vue'
+import CreateSpecModalBody from './CreateSpecModalBody.vue'
+import ExpandableFileList from './ExpandableFileList.vue'
+import FileMatch from '../../components/FileMatch.vue'
+import ReactComponentList from './component/ReactComponentList.vue'
+import { gql } from '@urql/core'
+
+const props = withDefaults(defineProps<{
+  files: any[]
```
lmiller1990: Do we have a type for this?

packages/app/src/specs/generators/ExpandableFileChooser.vue
```diff
@@ -0,0 +1,166 @@
...
+gql`
+fragment FileChooser on FileParts {
+  relative
+  id
+  ...FileListItem
+}
+`
+
+const emits = defineEmits<{
+  (eventName: 'selectItem', { file: File, item: string })
```
lmiller1990: I think this is using the global `File` type: https://developer.mozilla.org/en-US/docs/Web/API/File

I would highly recommend picking a type that doesn't collide with `File`, I have been burnt by this before :sweat_smile:

packages/app/src/specs/generators/ExpandableFileChooser.vue
```diff
@@ -0,0 +1,166 @@
...
+  }
+})
+
+///*------- No Results Options -------*/
+
+const noResults = computed(() => {
+  return {
+    search: filePathSearch.value || debouncedExtensionPattern.value,
+    message: filePathSearch.value ? t('noResults.defaultMessage') : t('components.fileSearch.noMatchesForExtension'),
+    clear: filePathSearch.value ?
```
lmiller1990: Not sure you wrote this originally but should this be a single fn with a conditional?

```js
clear: () => {
  if (filePathSearch.value) {
    filePathSearch.value = ''"
  } else {
    localExtensionPattern.value = initialExtensionPattern
  }
}
```

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
```diff
@@ -0,0 +1,198 @@
...
+        />
+      </div>
+      <div>
+        <StandardModalFooter
+          v-if="result"
+          class="flex gap-16px items-center"
+        >
+          <router-link
+            class="outline-none"
+            :to="{ name: 'SpecRunner', query: { file: result.file.relative?.replace(/\\/g, '/') }, params: { shouldShowTroubleRenderingAlert: true } }
```
lmiller1990: We have this regexp a lot, I wonder if we can take this chance to refactor it? It could be `toPosix` or `posixify` (sounds cooler).

image

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
```diff
@@ -0,0 +1,198 @@
...
+import TestResultsIcon from '~icons/cy/test-results_x24.svg'
+import EmptyGenerator from '../EmptyGenerator.vue'
+
+const props = defineProps<{
+  title: string
+  gql: ReactComponentGeneratorStepOne_CodeGenGlobFragment
+}>()
+const { t } = useI18n()
+const emits = defineEmits<{
+  (event: 'update:title', value: string): void
```
lmiller1990, line 99: Random q, why do we ever want to change the title/description? Did you find out? I have a feeling this is working around a flaw in the generator design, but I could be mistaken.

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
```diff
@@ -0,0 +1,198 @@
...
+  generateSpecFromSource.value = data?.generateSpecFromSource
+  result.value = data?.generateSpecFromSource?.generatedSpecResult?.__typename === 'ScaffoldedFile' ? data?.generateSpecFromSource?.generatedSpecResult : null
+  generatedSpecError.value = data?.generateSpecFromSource?.generatedSpecResult?.__typename === 'GeneratedSpecError' ? data?.generateSpecFromSource?.generatedSpecResult : null
+}
+const cancelSpecNameCreation = () => {
+  generatedSpecError.value = null
+}
+</script>
+<style scoped>
+.ghost-div {
```
lmiller1990: Another good reason to move from Windi to Tailwind... we can use arbitrary values https://v2.tailwindcss.com/docs/just-in-time-mode#arbitrary-value-support. You can just do `class=[calc(100% - 24px)]` using their arbitrary value feature.

I don't think this works here right now, but I hope we can move to "real" Tailwind soon. The more I use it over Windi, the more I want to migrate.

lmiller1990 (commented):
Just some early bird comments - I know this is still in draft, so ignore them if they are not useful. I will hold off testing until the UI is complete and this is out of draft, unless you really want some early testing.

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
```diff
@@ -0,0 +1,198 @@
...
+          v-else
+          :file="result.file"
+        />
+      </div>
+      <div>
+        <StandardModalFooter
+          v-if="result"
+          class="flex gap-16px items-center"
+        >
+          <router-link
```
marktnoonan: The Button component knows how to render a router link if you give it the `to` prop, so this might be able to be simplified to just use Button: https://github.com/cypress-io/cypress/blob/develop/packages/frontend-shared/src/components/Button.vue#L121

(Our designs treat "button" as a purely visual concept, that sometimes acts like a link or a button, so the code treats it the same way... wonder if there's a way to make this more discoverable/obvious, possibly with the component name?)

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
marktnoonan: We can do this kind of thing already, see the `w-[calc(100%-2px)]` [here](https://github.com/cypress-io/cypress/blob/develop/packages/app/src/specs/InlineSpecList.vue#L21).

But still I agree with switching to real tailwind soon for the other reasons you've mentioned.

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
astone123: It looks like if there's an error generating the spec file, we fall back to `EmptyGenerator` which updates the title once a spec is successfully generated 🤔 seems like there's probably a better way to do this

from `EmptyGenerator`
```typescript
whenever(result, () => {
  title.value = t('createSpec.successPage.header')
  emits('updateTitle', t('createSpec.successPage.header'))
})
```

packages/app/src/specs/generators/ExpandableFileChooser.vue
astone123: We do, `FileParts[]`

packages/app/src/specs/generators/ExpandableFileChooser.vue
astone123: Yeah the syntax here was incorrect anyway which is probably why I didn't notice 👍🏻 good catch

packages/app/src/paths.ts
```diff
@@ -18,3 +18,7 @@ export function getPathForPlatform (posixPath?: string) {
 
   return posixPath
 }
+
+export function posixify (path?: string): string {
```
lmiller1990: Can this be undefined? That seems weird.

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
lmiller1990: Looks like we duplicate some logic. Probably too much of a lift to work on that refactor here, but good to know and maybe something we can revisit...

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
lmiller1990: Oh, neat - we can make the refactor here, then!

Still look forward to mainline Tailwind. Windi has some neat features, and there's a time and place for trying new libraries, but a production app like Cypress isn't one of them.

packages/app/src/paths.ts
astone123: some of the places where we call this function pass things in from GraphQL query results where they might not be defined. I'll update this to pass `''` at the call sights instead

astone123:
Marked this as ready for review. There are still a couple things I'm working on 
- Testing
- Fixing an issue where the files in the file list are getting cut off at the bottom

but I want to get some feedback on the functionality of the feature

packages/app/src/specs/generators/ExpandableFileChooser.vue
```diff
@@ -0,0 +1,172 @@
...
+    </div>
+    <ExpandableFileList
+      v-show="!loading"
+      :style="{ paddingTop: `${fileMatchHeight + 36}px` }"
+      class="right-24px left-24px absolute"
+      :files="filteredFiles"
+      :search="filePathSearch"
+    >
+      <template #expanded-content="{file}">
+        <ReactComponentList
```
astone123: I want to re-work this so that the expandable section can be passed as a prop into `ExpandableFileChooser` but I couldn't figure out how to get the slots to work that way. 

That way this can be generic if we want to use it again for something else

lmiller1990:
I tried it with `npm/react`: 

image

- I think we can silently ignore `cypress` UNLESS it is in their `component.specPattern`.
- Need to set `network-policy: network-only`. I found a bug where my component doesn't show up.

1. hit create spec
2. go to "create from component'
3. close modal
4. create a component (say `blah.jsx`)
5. re-open modal and try make a spec
6. `blah.jsx` is not an option
7. need to restart cypress


https://user-images.githubusercontent.com/19196536/206093590-326e8375-bc2d-44c0-94e7-ba1682216116.mp4

I'm not sure if I'm doing it wrong, but I cannot seem to find any components :thinking: 

image

Any tips? I'm in `npm/react` with `yarn cy:open`.

packages/app/src/specs/generators/ExpandableFileChooser.vue
lmiller1990: We can sync on this, but this seems fine to me? What would you like to change?

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
```diff
@@ -0,0 +1,180 @@
...
+
+const makeSpec = async ({ file, item }) => {
+  const { data } = await generateSpecMutation.executeMutation({
+    codeGenCandidate: file.absolute,
+    type: 'component',
+    componentName: item.exportName,
+    isDefault: item.isDefault,
+  })
+
+  generateSpecFromSource.value = data?.generateSpecFromSource
```
lmiller1990, line 181: Woah, this is quite hard to read. Should we have a field on the mutation response that tells us what we want to know, instead of having to infer it like this?

lmiller1990 (commented):
Code is looking pretty good. Definitely need some tests (you mentioned you are working on this).

I couldn't get the actual feature working! I'm not sure if I'm doing it wrong, or you forgot to push a commit, or maybe it's just not fully wired up?

astone123:
@lmiller1990 the issue with not finding any component definitions in `npm/react` looks like it has to do with a dependency resolution problem. If you run with `DEBUG="cypress:data-context"` you'll see this error logged

```
[cy:open:dev:68912]:   cypress:data-context Error: Cannot find module 'babel-plugin-module-resolver'
[cy:open:dev:68912]:   cypress:data-context Require stack:
[cy:open:dev:68912]:   cypress:data-context - /Users/adams/projects/cypress/packages/v8-snapshot-require/dist/snapshot-require.js
[cy:open:dev:68912]:   cypress:data-context - /Users/adams/projects/cypress/packages/server/hook-require.js
[cy:open:dev:68912]:   cypress:data-context - /Users/adams/projects/cypress/packages/server/index.js
[cy:open:dev:68912]:   cypress:data-context - 
[cy:open:dev:68912]:   cypress:data-context     at Module._resolveFilename (node:internal/modules/cjs/loader:940:16)
[cy:open:dev:68912]:   cypress:data-context     at Function.n._resolveFilename (node:electron/js2c/browser_init:249:1106)
[cy:open:dev:68912]:   cypress:data-context     at resolve (node:internal/modules/cjs/helpers:108:20)
[cy:open:dev:68912]:   cypress:data-context     at Function.require.resolve (/Users/adams/projects/cypress/packages/v8-snapshot-require/dist/snapshot-require.js:216:25)
[cy:open:dev:68912]:   cypress:data-context     at customRequire.resolve (<embedded>:836977:25)
[cy:open:dev:68912]:   cypress:data-context     at tryRequireResolve (<embedded>:68329:12)
[cy:open:dev:68912]:   cypress:data-context     at resolveStandardizedNameForRequire (<embedded>:68363:22)
[cy:open:dev:68912]:   cypress:data-context     at sync (<embedded>:68386:15)
[cy:open:dev:68912]:   cypress:data-context     at sync (<embedded>:4811:22)
[cy:open:dev:68912]:   cypress:data-context     at <embedded>:4838:27
[cy:open:dev:68912]:   cypress:data-context     at Generator.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at resolvePlugin (<embedded>:68260:20)
[cy:open:dev:68912]:   cypress:data-context     at resolvePlugin.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at loadPlugin (<embedded>:68266:30)
[cy:open:dev:68912]:   cypress:data-context     at loadPlugin.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at createDescriptor (<embedded>:62138:19)
[cy:open:dev:68912]:   cypress:data-context     at createDescriptor.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at evaluateSync (<embedded>:4867:37)
[cy:open:dev:68912]:   cypress:data-context     at <embedded>:4697:37
[cy:open:dev:68912]:   cypress:data-context     at Array.map (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at Function.sync (<embedded>:4697:23)
[cy:open:dev:68912]:   cypress:data-context     at Function.all (<embedded>:4838:27)
[cy:open:dev:68912]:   cypress:data-context     at Generator.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at createDescriptors (<embedded>:62100:44)
[cy:open:dev:68912]:   cypress:data-context     at createDescriptors.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at createPluginDescriptors (<embedded>:62097:20)
[cy:open:dev:68912]:   cypress:data-context     at createPluginDescriptors.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at <embedded>:61913:33
[cy:open:dev:68912]:   cypress:data-context     at Generator.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at mergeChainOpts (<embedded>:68898:36)
[cy:open:dev:68912]:   cypress:data-context     at mergeChainOpts.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at chainWalker (<embedded>:68861:17)
[cy:open:dev:68912]:   cypress:data-context     at chainWalker.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at loadFileChain (<embedded>:68733:27)
[cy:open:dev:68912]:   cypress:data-context     at loadFileChain.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at buildRootChain (<embedded>:68616:30)
[cy:open:dev:68912]:   cypress:data-context     at buildRootChain.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at loadPrivatePartialConfig (<embedded>:62424:77)
[cy:open:dev:68912]:   cypress:data-context     at loadPrivatePartialConfig.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at Function.<anonymous> (<embedded>:62468:28)
[cy:open:dev:68912]:   cypress:data-context     at Generator.next (<anonymous>)
[cy:open:dev:68912]:   cypress:data-context     at evaluateSync (<embedded>:4867:37)
[cy:open:dev:68912]:   cypress:data-context     at Function.sync (<embedded>:4741:17)
[cy:open:dev:68912]:   cypress:data-context     at <embedded>:62912:63
[cy:open:dev:68912]:   cypress:data-context     at buildPluginList (<embedded>:659832:57)
[cy:open:dev:68912]:   cypress:data-context     at buildOptions (<embedded>:659840:22)
[cy:open:dev:68912]:   cypress:data-context     at buildParse (<embedded>:659847:25)
[cy:open:dev:68912]:   cypress:data-context     at parse2 (<embedded>:667633:47)
[cy:open:dev:68912]:   cypress:data-context     at defaultParse (<embedded>:668514:33)
[cy:open:dev:68912]:   cypress:data-context     at CodegenActions.getReactComponentsFromFile (/Users/adams/projects/cypress/packages/data-context/src/actions/CodegenActions.ts:26:40)
[cy:open:dev:68912]:   cypress:data-context     at Socket.handleGraphQLSocketRequest (/Users/adams/projects/cypress/packages/graphql/src/makeGraphQLServer.ts:156:20)
[cy:open:dev:68912]:   cypress:data-context  +0ms
```

To everyone else reviewing this - you should be able to use any React app from [`cypress-component-testing-apps`](https://github.com/cypress-io/cypress-component-testing-apps), or [Real World App](https://github.com/cypress-io/cypress-realworld-app) to test this while I figure out the dependency issue with `npm/react`

packages/app/src/specs/generators/ExpandableFileChooser.vue
astone123: I wanted to make it more generic. Right now, if you use `ExpandableFileChooser`, the expanded content when you click a file is always `ReactComponentList`. I wanted the caller of `ExpandableFileChooser` to be able to pass in the component that it would like to be rendered when a row in `ExpandableFileList` is expanded

packages/app/src/specs/generators/component/VueComponentGenerator.tsx
```diff
@@ -0,0 +1,12 @@
+import { filters } from '../GeneratorsCommon'
+import VueComponentGeneratorStepOne from './VueComponentGeneratorStepOne.vue'
+import type { SpecGenerator } from '../types'
+import ComponentGeneratorCard from './ComponentGeneratorCard.vue'
+
+export const VueComponentGenerator: SpecGenerator = {
+  card: ComponentGeneratorCard,
+  entry: VueComponentGeneratorStepOne,
+  show: (currentProject) => !!(currentProject?.codeGenFramework && currentProject?.codeGenFramework === 'vue'),
```
ZachJW34:
```suggestion
  show: (currentProject) => currentProject?.codeGenFramework === 'vue',
```

Can simplify

packages/app/src/specs/generators/component/ReactComponentGenerator.tsx
```diff
@@ -0,0 +1,12 @@
+import { filters } from '../GeneratorsCommon'
+import ReactComponentGeneratorStepOne from './ReactComponentGeneratorStepOne.vue'
+import type { SpecGenerator } from '../types'
+import ComponentGeneratorCard from './ComponentGeneratorCard.vue'
+
+export const ReactComponentGenerator: SpecGenerator = {
+  card: ComponentGeneratorCard,
+  entry: ReactComponentGeneratorStepOne,
+  show: (currentProject) => !!(currentProject?.codeGenFramework && currentProject?.codeGenFramework === 'react'),
```
ZachJW34:
```suggestion
  show: (currentProject) => currentProject?.codeGenFramework === 'react',
```

can simplify

packages/app/src/specs/generators/component/ReactComponentList.vue
```diff
@@ -0,0 +1,81 @@
...
+  getReactComponentsFromFile(filePath: $filePath) {
+    exportName
+    isDefault
+  }
+}`
+
+const getReactComponentsMutation = useMutation(ComponentList_GetReactComponentsFromFileDocument)
+
+const getComponents = async (file) => {
+  loading.value = true
```
ZachJW34: Can we use `getReactComponentsMutation.fetching` property instead of manually tracking it?

packages/app/src/specs/generators/component/ReactComponentList.vue
```diff
@@ -0,0 +1,81 @@
...
+      >
+        <div
+          class="cursor-pointer flex border-b-1 border-b-gray-50 leading-normal pl-56px text-16px
+    gap-8px group items-center last last:py-0 last:items-start children:h-40px children:py-8px"
+          @click="$emit('selectItem', {file, item: {exportName, isDefault}})"
+        >
+          <div
+            class="h-full inline-flex whitespace-nowrap items-center overflow-hidden"
+          >
+            <i-cy-puzzle-piece_x16
```
ZachJW34, line 28: I feel like this could use a hover effect. The Files have a noticeable effect on hover (dropdown icon color change, relative path) but then it feels like goes stale hovering over these elements

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
```diff
@@ -0,0 +1,181 @@
...
+    id
+    codeGenCandidates(glob: $glob) {
+      id
+      fileName
+      fileExtension
+      absolute
+      relative
+      baseName
+    }
+    # Add the specs, so we can keep the list up to date with the cache
```
ZachJW34, line 120: Do you know if we need this anymore now that we have subscriptions?

packages/app/src/specs/generators/component/ReactComponentList.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+  <div class="h-full">
+    <div
+      v-if="!loading && !components.length"
```
ZachJW34: We chatted a bit about this: We should have an error result here `Failed to parse component` with that red info icon.

ZachJW34 (commented):
I gave it a pretty thorough test. Worked great for projects outside the monorepo but not for `npm/react` which you're looking into. Once that is fixed and there are some tests added I'll approve!

ZachJW34:
`react-docgen` had trouble parsing

```ts
interface LoginFormProps {
  onLogin: ({
    username,
    password,
  }: {
    username: string;
    password: string;
  }) => void;
  title?: string;
  errorMessage?: string;
}
```

inside a component, as well as a component using a non-null assertion e.g. `document.getElementById('app')!`.

We should follow up with an issue to document these edge-cases and see if what we can do about them.

lmiller1990:
> `react-docgen` had trouble parsing
> 
> ```ts
> interface LoginFormProps {
>   onLogin: ({
>     username,
>     password,
>   }: {
>     username: string;
>     password: string;
>   }) => void;
>   title?: string;
>   errorMessage?: string;
> }
> ```
> 
> inside a component, as well as a component using a non-null assertion e.g. `document.getElementById('app')!`.
> 
> We should follow up with an issue to document these edge-cases and see if what we can do about them.

Weird this doesn't work... it's just basic TypeScript. I guess the version of babel that is used somehow doesn't understand this. Agreed we should show something like "failed to parse component".

lmiller1990 (commented):
Tested it out, works well in projects other than `npm/react`. I wonder if this is a problem specific to our monorepo?

I've left a few comments; let me know when you'd like me to re-review.

packages/app/src/specs/generators/component/ReactComponentList.vue
astone123: I added an error state that says

> Unable to parse file

But I didn't add an icon. After talking with @vikramrojo, we decided to keep the visual presentation the same as "No components found" because there's nothing actionable for the user to do if we fail to parse the file.

packages/app/src/specs/generators/component/ReactComponentList.vue
vikramrojo: good catch. i missed this when we switched directions. i would use the pattern we have for a runner even though a file here is not a folder it acts as a container.
expanded
list

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
astone123: I'm not sure... this was in the original implementation and I'm not sure what it does

packages/app/src/specs/generators/component/ReactComponentList.vue
astone123: Added a hover effect. You can also now see it in the mocks that are linked in the PR description

marktnoonan:
A few observations so far:

Keyboard behavior seems not working, focus moves to the area surrounding the accordion, not the accordion itself:

Screen Shot 2022-12-08 at 4 54 25 PM

Maybe not an issue with your PR, since it is using the Collapsible component, it's possible these weaknesses are there. I might suggest some chagnes so that our Collapsible component uses [Headless UI's Disclosure](https://headlessui.com/react/disclosure) internally, instead of us managing this ourselves, then I think we'd get a lot of the keyboard behavior and semantics OOTB and only have to do minor adjustments for this accordion. Ideally we might be able to the nice arrow key behavior from the inline specs list to happen here.

I was able to get to a full screen error state accidentally. I haven't been able to reproduce this yet, I'll keep trying:

Screen Shot 2022-12-08 at 5 12 03 PM

packages/app/src/specs/generators/ExpandableFileList.vue
```diff
@@ -0,0 +1,90 @@
+<template>
+  <div class="h-full">
+    <div
+      v-if="files.length"
+      class="h-full relative"
+      tabindex="0"
+    >
+      <ul class="h-full">
```
ZachJW34: Screen Shot 2022-12-08 at 4 48 58 PM

I was getting some cutoff of the component list, I fixed it with `padding-bottom: 24px; overflow: auto` on this ul. Looks like a parent has overflow which is weird cause it makes the scroll start at the top of the component rather than the actual scrollable content.

packages/app/src/specs/generators/ExpandableFileList.vue
```diff
@@ -0,0 +1,90 @@
...
+            data-cy="file-list-row"
+            @click="() => toggleExpandRow(file.id)"
+          >
+            <i-cy-chevron-down-small_x16
+              class="mr-8px text-sm icon-dark-gray-300 group-hocus:icon-dark-gray-700"
+              :class="{'transform rotate-270': !isRowExpanded(file.id)}"
+            />
+            <i-cy-document-blank_x16 class="min-w-16px min-h-16px icon-light-gray-50 icon-dark-gray-300" />
+            <div class="h-full inline-flex whitespace-nowrap items-center overflow-hidden">
+              <span class="font-medium text-gray-600">{{ name(file) }}</span>
```
ZachJW34: Screen Shot 2022-12-08 at 4 54 06 PM

Looks like the selected file has gray font but non-selected files have blue font.

packages/data-context/src/actions/CodegenActions.ts
```diff
@@ -46,11 +46,17 @@ export class CodegenActions {
...
       this.ctx.debug(err)
 
-      return []
+      // react-docgen throws an error if it doesn't find any components in a file.
+      // This is okay for our purposes, so if this is the error, catch it and return [].
+      if (err.message === 'No suitable component definition found.') {
+        return []
+      }
+
+      throw new Error('Error parsing React component file')
```
ZachJW34: We shouldn't be throwing an error here but rather returning an object with the error information.

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
```diff
@@ -0,0 +1,181 @@
+<template>
+  <div class="flex flex-col flex-grow justify-between">
+    <template v-if="generatedSpecError">
+      <EmptyGenerator
```
ZachJW34, line 4: Do we know if this flow works for CFR? Wouldn't the mutation need to know the `isDefault` for it to scaffold properly after the renaming?

lmiller1990:
I got the same problem as @marktnoonan but I _think_ it was due to usage with watch mode and refreshing too fast.

ZachJW34 (changes requested):
Left a few comments, biggest is the `throw error` which we shouldn't be doing but rather returning the error info in the response. Not sure how the error bubbles up but that could be what Mark is experiencing.

packages/app/src/specs/generators/ExpandableFileChooser.vue
lmiller1990: What about this: https://github.com/cypress-io/cypress/pull/25079

packages/data-context/src/actions/CodegenActions.ts
lmiller1990: I think this manifests in an error in GraphQL we can display (could be wrong)?

lmiller1990: Oh, maybe not 
image

I agree, we should have an inline "sorry, we could not parse this component". We don't have designs for this, though. We can probably make a smart decision and use the existing patterns we've got, though - we have an `<Alert type="warning">` component. Or, we can see what design thinks. I don't think the big overlay makes sense, though - that's more for major, runner-crashing errors.

lmiller1990: Actually wait, I think my initial instinct was correct - we this already:

image

The "big" error seems like a dev-only thing. Let me test the binary out.

packages/app/src/specs/generators/ExpandableFileList.vue
astone123: I think that's a mistake in the mock up maybe? In a different one, all of the file rows are gray

packages/app/src/specs/generators/ExpandableFileList.vue
astone123: thanks, this is one of the things I've been trying to figure out with the UI

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
astone123: I haven't tried this yet, but I don't think it would need `isDefault` since it will just create an empty spec. It doesn't try to import anything

.eslintignore
```diff
@@ -0,0 +1 @@
+system-tests/projects/no-specs/src/Invalid.jsx
```
astone123: I added this because ESLint wouldn't let me commit with this file because it has a syntax error, but the error is intentional for testing

packages/app/src/specs/generators/ExpandableFileList.vue
astone123: @vikramrojo could you confirm what the styling should be here? 
- All file names gray
or
- only selected files gray

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
astone123: yeah this is pretty complex. My understanding is that the type that this mutation returns is a union type, so it either returns a `ScaffoldedFile` response or a `GeneratedSpecError` response. In order to know which response we got, we need to check the `__typename` field like this.

There's probably a better way to refactor this... I wonder if we could make an issue to go back and refactor these types later?

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
astone123: Okay I tried this... it doesn't seem to work for either of the create from component flows, but I think it's for a different reason. Going to make an issue to look into this because we should have a fallback if something goes wrong.

ZachJW34 (approved):
My comments have been addressed. Works great!!!

Looks like you're building the binaries. Going to approve but would like to test those out as well.

.eslintignore
lmiller1990: Does `/*eslint-disable */` not work?

packages/app/src/specs/generators/ExpandableFileChooser.vue
```diff
@@ -0,0 +1,165 @@
...
+      <template
+        v-if="loading"
+        #matches
+      >
+        <i-cy-loading_x16 class="h-24px mr-10px animate-spin w-24px" />
+      </template>
+    </FileMatch>
+
+    <div
+      v-show="loading"
```
lmiller1990: Must have missed this before but was there a specific reason to use `v-show` over `v-if` and `v-else` here?

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
lmiller1990: Refactor later means refactor never - happy to live with this for now, but I hope we can start to think about how to get a handle of GraphQL. Things are getting quite messy, I think the core problem is we should be handling this on the back-end and giving the client a more well defined response in the first place. It would eliminate the need for the conditionals in the frontend.

packages/scaffold-config/src/frameworks.ts
```diff
@@ -149,7 +149,7 @@ export const WIZARD_FRAMEWORKS = [
       ])
     },
     codeGenFramework: 'react',
-    glob: '*.{js,jsx,tsx}',
+    glob: '*.{js,jsx,ts,tsx}',
```
lmiller1990: I am not sure if this makes sense - can you write React in a `ts` file? TSX doesn't work there - I guess you could non-JSX React :thinking:

lmiller1990 (approved):
Couple of tiny things, but I think this looks good. Mainly curious about  https://github.com/cypress-io/cypress/pull/24982#discussion_r1045319093.

.eslintignore
astone123: It doesn't seem to, no... when I delete the .eslintignore file and update `Invalid.jsx` to look like this

```js
import React from 'react'

/* eslint-disable */
export function MyComponent (({
```

The commit hook still fails
```
✔ Preparing...
⚠ Running tasks...
  ↓ No staged files match *.coffee [SKIPPED]
  ❯ Running tasks for *.{js,jsx,ts,tsx,json,eslintrc,vue}
    ✖ eslint --fix [FAILED]
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up...

✖ eslint --fix:

/Users/adams/projects/cypress/system-tests/projects/no-specs/src/Invalid.jsx
  4:29  error  Parsing error: Parameter declaration expected

✖ 1 problem (1 error, 0 warnings)

```

I guess I could just commit with `--no-verify` since this is intentional

astone123:
Having an issue with dependencies (looks like Babel) after merging in changes from develop. See the latest pipeline runs. Working on this now

packages/scaffold-config/src/frameworks.ts
astone123: yeah you wouldn't be able to write any JSX but theoretically you could write a component that returns a string or returns `null`. Although, react-docgen doesn't recognize those as components by default anyway 🤔 

yeah I'm thinking I should revert this change

packages/app/src/specs/generators/ExpandableFileChooser.vue
astone123: There was not - in fact, this doesn't even need to be here

packages/scaffold-config/src/frameworks.ts
astone123: reverted these changes in https://github.com/cypress-io/cypress/pull/24982/commits/d24360867c0febf7ef3aec96328f65a5bad61813

.eslintignore
astone123: Updated this in https://github.com/cypress-io/cypress/pull/24982/commits/1670c94cfc0816bec0a99da012788589222a704a

packages/graphql/schemas/schema.graphql
```diff
@@ -1596,6 +1599,15 @@ type ReactComponentDescriptor {
   isDefault: Boolean!
 }
 
+"""Response from getReactComponentsFromFile"""
+type ReactComponentResponse {
+  """Components that are exported from the parsed file"""
+  components: [ReactComponentDescriptor!]!
+
+  """Whether or not there was there an error when parsing the file"""
+  errored: Boolean
```
rockindahizzy, line 1608: Will we care about the type of error/message?

astone123: we don't display the actual error in the UI, so no. The full error will be logged via debug https://github.com/cypress-io/cypress/pull/24982/files#diff-b24c7f86b6dcda7a4a5727db3592d6a026e3334843cf65aba4313189885497bfR47

packages/app/src/specs/generators/component/ReactComponentGeneratorStepOne.vue
rockindahizzy: Thoughts on creating an issue to capture this?

PR #24987, opened 2022-12-06
fix: restore spec filter in client side store

### LINKED ISSUE #24936: Searched word reverted after re-opening the chrome window
### Current behavior

1. Open the chrome browser from the cypress window.
2. Search the feature file name
3. Run the same
4. Close the browser
5. Now, again open the chrome browser from the cypress window.
6. It's not showing the recently searched word. It's showing the last searched word

### Desired behavior

It should show the recently searched word

### Test code to reproduce

https://user-images.githubusercontent.com/102032328/205221110-0abf318b-d91a-4563-9e13-ddf4ee064428.mp4

### Cypress Version

11.2.0

### Node version

16.17.0

### Operating System

Windows 11

### Debug Logs

_No response_

### Other

_No response_

marktnoonan:
- Closes #24936

### User facing changelog

The spec search term will no longer become stale in the UI.

### Additional details

This restores some old code to keep this value up to date on the front end, and unskips a previously skipped test for this behavior.

### Steps to test

- Open Cypress at the specs list and enter a search term. 
- Open a spec from your search results and see that your latest search term is populated in the inline specs list on the spec runner page. 
- Edit the search term then navigate back to the specs list, see that the edited term is displayed.
- Quit and re-open the browser via the launchpad (as described in #24936) and see that the latest search term is correct when the browser re-opens.

### PR Tasks


marktnoonan:
@lmiller1990 I ran across this issue looking for something else today and got curious, I made a draft PR restoring this part of the client side store. This fixes the demonstrated issue for me, but definitely needs more tests and likely refactoring for clarity, need to look at it a bit more closely.

I remember reviewing these changes and didn't catch the one place this was being used in `useSpecFilter`. Now that I see the problem I think one option is "the old behavior with heavier regression testing" but we can talk about using gql fully for this instead. 

I can see now why we have the debouncing and caching all happening client side & how not having a value in the store, it solves a few problems & might still be the simplest way to do so.

packages/app/src/composables/useSpecFilter.ts
```diff
@@ -25,6 +25,7 @@ export function useSpecFilter (savedFilter?: string) {
 
   function setSpecFilter (specFilter: string) {
     if (specStore.specFilter !== specFilter) {
+      specStore.setSpecFilter(specFilter)
```
lmiller1990, line 28: I deleted this, my b... not the worst, now we know we were missing test coverage.

marktnoonan: Yes, I'll work on a test today, will try to get an e2e test that catches it and allows us to swap out the implementation.

emilyrohrbough:
Does this fix persisting the spec filter to the spec panel once a spec as been clicked? I've noticed that no longer seems to stay in sync.

marktnoonan:
> Does this fix persisting the spec filter to the spec panel once a spec as been clicked? I've noticed that no longer seems to stay in sync.

Yes, this is the same underlying cause - it brings back the code that was present previously, I guess our tests didn't do enough to navigate back and forth and detect a stale value.

BlueWinds (approved):

rockindahizzy (approved):

lmiller1990 (approved):
Test would be nice but LGTM

PR #24992, opened 2022-12-06
fix(grep): update cypress-grep references & docs & add e2e tests

### LINKED ISSUE #24512: @cypress/grep cannot find module cypress-grep or it's corresponding type declarations
### Current behavior

Support file, Config file, and the tsconfig file installation instructions are incorrect.

Module names and paths are incorrect.

Support file: I am getting a _could not find a declaration file for module @cypress/grep/src/support_ as that's what I have at the moment. Following the readme, the error is _Cannot find module 'cypress-grep'_. Additionally, the function appears to be cypressGrep and not registerCypressGrep. 

Config file registration: `require('@cypress/grep/src/plugin')(config)`

tsconfig.json: the types array addition should be `"@cypress/grep"`

I am probably missing something.

### Desired behavior

I am able to install the package, follow the configuration instructions, grepping specs works.

### Test code to reproduce

n/a

### Cypress Version

10.10.0

### Node version

16.17.1

### Operating System

Windows 10

### Debug Logs

_No response_

### Other

_No response_

### LINKED ISSUE #24696: @cypress/grep readme references previous cypress-grep package
### Subject

readme

### Description

The documentation at https://github.com/cypress-io/cypress/tree/develop/npm/grep has mostly been copied from the previous `cypress-grep` repo. Apart from the initial installation instructions a lot of references to the older package still exist so are confusing. For example:

- Support File installation instructions are wrong, commands should be `import cypressGrep from '@cypress/grep'`
  - Also worth noting here is that types are incorrect in this case as no declaration is provided for this export
- Similar issue with the Config File section
- Typescript support section again mentions the wrong package `cypress-grep` instead of `@cypress/grep`
- Debugging section mentions wrong package name when searching for the package
- The example repos use the older package which is also confusing, perhaps they can be marked with a warning about that until updating to work with the newer package

Hopefully these are all simple changes to make to avoid any confusion during the migration like we had.

### LINKED ISSUE #24967: @cypress/grep internal read.me is correct - NPM references deprecated cypress-grep package
### Current behavior

When searching for cypress grep online you're correctly pointed to [@cypress/grep npm package](https://www.npmjs.com/package/@cypress/grep), but the readme is referencing the deprecated cypress-grep plugin that is no longer maintained specifically how to import in the support file & setup in the config file.  

### Desired behavior

The current npm package should be updated to reflect what is in the Cypress @cypress/grep readme. 

### Test code to reproduce

none

### Cypress Version

11.2.0

### Node version

16

### Operating System

windows 11

### Debug Logs

```shell
none
```

### Other

none

jordanpowell88:
- Closes #24967
- Closes #24696
- Closes #24512 

### User facing changelog
fix references still pointing to `cypress-grep`

### Additional details
This PR also adds e2e tests as part of this lib

### Steps to test
`yarn workspace @cypress/grep test` // unit tests
`yarn workspace @cypress/grep cy:run` // e2e tests

### PR Tasks


npm/grep/cypress/support/commands.js
```diff
@@ -0,0 +1,107 @@
...
+  .contains('li', todo.trim(), { log: false })
+  .then(function ($li) {
+    // set the $el for the command so
+    // it highlights when we hover over
+    // our command
+    cmd.set({ $el: $li }).snapshot().end()
+  })
+})
+
+Cypress.Commands.add('addAxeCode', () => {
```
jaffrepaul: Do we need this? Didn't see where it was used.

jaffrepaul (commented):
Left one comment about `addAxeCode`.

Didn't have the bandwidth to test run this, but otherwise looks straight forward.

npm/grep/scripts.js
```diff
@@ -0,0 +1,170 @@
...
+
+function runCommand (command) {
+  if (exec(command).code !== 0) {
+    echo(`Error: ${command}`)
+    exit(1)
+  }
+}
+
+// Run e2e tests 🧪
+runCommand('npx cypress-expect --min-passing 5 --pending 0')
```
emilyrohrbough: How come these aren't leveraging the same system-test setup as the other `run` mode tests in the repo?

npm/grep/package.json
```diff
@@ -13,6 +14,7 @@
     "globby": "^11.0.4"
   },
   "devDependencies": {
+    "cypress": "^11.2.0",
```
jennifer-shehane, line 18: Do we want this using the dev version of Cypress - current within the repo? `0.0.0-development` I can't imagine someones going to maintain updating this in the future.

jordanpowell88: The issue is that we are using `cypress-expect` which doesn't work with our binary 👎 . And we need to be able to validate the output with the expected results objected in the `/expected/my-expected-results.json`

npm/grep/scripts.js
jordanpowell88: We need to use `cypress-expect` to validate the results because we aren't just checking if tests pass or not

emilyrohrbough: @jordanpowell88 what else is being verified? The expect json seem to be verifying the pass/fail results. The name of each test that ran?

npm/grep/package.json
lmiller1990: How come that doesn't work with our binary, curious?

npm/grep/scripts.js
jordanpowell88: We need essentially a way to test the results of each test ie: this test passed, this test is pending, this test skipped etc

emilyrohrbough: We have examples where do we this in the system tests: https://github.com/cypress-io/cypress/blob/develop/system-tests/test/testConfigOverrides_spec.ts#L42

jordanpowell88:
I updated to using `system-tests` using the strategy @ZachJW34 implemented on `@cypress/schematic`

system-tests/projects/grep/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "grep",
+  "version": "0.0.0",
+  "devDependencies": {
+    "cypress": "latest",
```
emilyrohrbough: We shouldn't be installing Cypress but testing against the binary.

npm/grep/test.spec.js
```diff
@@ -0,0 +1,288 @@
...
+
+  before(async () => {
+    projectPath = await scaffoldGrepProject(cypressGrepPackagePath)
+
+    await runCommandInProject(`yarn add @cypress/grep@file:${cypressGrepPackagePath}`, projectPath)
+  })
+
+  it('run Unit Tests', async () => {
+    await runCommandInProject('npx cypress run --spec cypress/e2e/unit.js --config specPattern="**/unit.js"', projectPath)
+  })
```
emilyrohrbough: there should not be unit tests in system-tests. These should live in `npm/grep`

npm/grep/test.spec.js
```diff
@@ -0,0 +1,288 @@
...
+
+  it('Run e2e tests 🧪', async () => {
+    await runCommandInProject(`npx cypress-expect --min-passing 5 --pending 0`, projectPath)
+  })
+
+  it('Run e2e skip tests without grep 🧪', async () => {
```
emilyrohrbough: I would expect these system tests to be modeled like the other system-tests in Cypress for maintainability purposes. If we need to make enhancements to support checking the report output, I think that'd be better than adding dozens of json files with the test results don't live with the test. At a min, this need to be running against the binary and not using `npx` in these tests.

emilyrohrbough (changes requested):

npm/grep/test.spec.js
jordanpowell88: The issue is that the plugin needs the `cypress` module, and it can't resolve without  it. It is the same issue we have with `@cypress/schematics`. This is why we have to move the system tests to a temp dir to run

emilyrohrbough: @jordanpowell88 What do you mean exactly? Because it's listed as a peerDependency?

jordanpowell88: Yes it needs cypress as a peerDep but it also needs cypress as a devDep

emilyrohrbough: Looking how this is setup though, that's because of `cypress-expect` right? which I why I am wondering if that's the right tool if we are unable to setup the system-tests to use the just-built binary. It looks like this is just a wrapper around running cypress programmatically to return the test results. We can pull the cypress output from `systemTests.it` to determine the tests that ran and how many passed & failed.

npm/grep/scripts.js
jordanpowell88: @emilyrohrbough I can work on changing the structure of these, but it will be a heavy lift. stay tuned

npm/grep/test.spec.js
jordanpowell88: Correct, we can do this but this would be a bigger lift than we are wanting to do at this moment

npm/grep/test.spec.js
jordanpowell88: Resolved

npm/grep/package.json
jordanpowell88: It's because `cypress-expect` wraps around the module API

lmiller1990:
@jordanpowell88 and you re-ping when this is ready to review? :pray:

Narretz:
Are there any updates for this @jordanpowell88 ?

Narretz:
Hi @jordanpowell88 I hate to be that person, but is there any news here?

Narretz:
Hi @jordanpowell88 I go the notification that this issue is closed, but the code hasn't been merged. Will the work continue in another PR?

lmiller1990:
Let me find out for you. We should decide what's happening with cypress-grep moving forward.

jordanpowell88:
@Narretz re-opened this PR (without the tests) https://github.com/cypress-io/cypress/pull/26108

Narretz:
Thanks @jordanpowell88 

What's with the tests, though? Will they be added in a separate PR?

jordanpowell88:
This will be up the @cypress-io/end-to-end team to create a follow issue around this as they are now responsible for handling the `@cypress/grep` package. Feel free to create an issue as well

PR #24996, opened 2022-12-06
fix: only show major browser version in app settings

### LINKED ISSUE #21730: Browser version in App showing full version numbers, should match launchpad
## Summary
The browser version dropdown is showing minor version numbers instead of the major ones.

## See screenshot

Screen Shot 2022-02-10 at 8 29 09 PM

┆Issue is synchronized with this [Jira Bug](https://cypress-io.atlassian.net/browse/UNIFY-1117) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://cypress-io.atlassian.net/rest/api/2/attachment/content/11296">Screen Shot 2022-02-10 at 8.29.09 PM.png</a>
┆author: Jess Sachs
┆epic: UI/UX Improvements
┆friendlyId: UNIFY-1117
┆priority: Medium
┆taskType: Bug

melkstam:
- Closes #21730 

### User facing changelog

Show major instead of minor versions for settings in browser app

### Steps to test

Open the settings for a project and open the browser version dropdown

Before:
image

After:
after

### PR Tasks


BlueWinds (approved):
I'm a fan of the change. :+1:

mike-plummer (commented):
I like this change too, helps align the browser menu with what we display in the Launchpad. Just one nitpick, curious what you think @BlueWinds - technically the changes to the spec don't validate that we *only* display the major version since `contains` does a partial match, so they would still pass even if we display the full version number. Only way I could think of to do this check feels pretty clunky - have to use a function with assertion in order to use a RegExp with an "end of line" validation to ensure there isn't more text beyond the major version value

```
.and(($el) => {
  expect($el.text()).to.match(/Version 1$/)
})
```

BlueWinds:
> I like this change too, helps align the browser menu with what we display in the Launchpad. Just one nitpick, curious what you think @BlueWinds - technically the changes to the spec don't validate that we _only_ display the major version since `contains` does a partial match, so they would still pass even if we display the full version number.

I'm not terribly fussed over "Version 1.x" matching, tending to prefer minimalistic tests over strictly correct / complete ones. Fine with me how it stands.

BlueWinds:
The test failure looks like it might be legit though, rather than flake - I haven't dug into it, but looks related to the changes here.

melkstam:
> The test failure looks like it might be legit though, rather than flake - I haven't dug into it, but looks related to the changes here.

Yes, I think it is as well. Seems to be related to showing unsupported versions, will look into it!

mike-plummer (approved):

PR #25014, opened 2022-12-06
fix(launcher): set ARCHPREFERENCE on macOS to prefer arm64 

flotwig:
- Closes 

### User facing changelog

Fixed an issue where browsers distributed as universal binaries (Chrome, Firefox) on M1 Macs could be launched in the wrong architecture, resulting in poor performance in-browser.

### Additional details

- Browsers launched via processes that are grandchildren of an Intel process will be launched as Intel processes, if they are universal.
- So if you launch arm64 Cypress via amd64 Node.js, macOS will launch Chrome/FF as their amd64 versions (slow)
- We can use `arch` and `ARCHPREFERENCE` to work around this: https://stackoverflow.com/a/74709144/3474615
- This PR also splits the `browsers` list out of `browsers.ts` and into a `knownBrowsers` object in `known-browsers.ts`, to avoid an issue where it is imported in the app's supportFile, which now causes an error due to `execa` being incompatible with the browser: https://cypressio.slack.com/archives/C01TQQCQLJY/p1671475149587209

### Steps to test

Validate that it launches Arm browsers on M1 with x64 Node:

1. Set up x64 Node.js on an M1 Mac.
2. Using x64 Node.js, install this branch.
3. Launch Cypress and validate it is launched as "Apple" in Activity Monitor.
4. Launch a browser (Chrome, FF) from Cypress and validate it is launched as "Apple" in Activity Monitor, not "Intel".

Validate that x64 Macs can still launch browsers:

1. Launch Cypress on an x64 Mac and launch a browser packaged as universal (Chrome, FF), validate it launches.

### PR Tasks


packages/launcher/test/unit/darwin_spec.ts
```diff
@@ -61,4 +86,79 @@ describe('darwin browser detection', () => {
...
+      getOutput.restore()
+
+      return sinon.stub(cp, 'spawn').returns(spawnStub())
+    }
+
+    context('in version detection', () => {
+      it('uses arch and ARCHPREFERENCE on arm64', async () => {
+        const cpSpawn = stubForArch('arm64')
+
+        // this will error since we aronlyen't setting stdout
```
flotwig:
```suggestion
        // this will error since we aren't setting stdout
```

lmiller1990:
Looks like it's working. Here's a video.

I did see cypress config manager as 'intel' - is that correct? The actual app and browser both launched as Apple.

https://user-images.githubusercontent.com/19196536/208824972-de627ab3-e742-4af3-b65b-44f7a0a81065.mov

lmiller1990 (approved):
Code looks good, I also posted a video of my testing.

ryanthemanuel (approved):
Looks good!

flotwig:
> I did see cypress config manager as 'intel' - is that correct? The actual app and browser both launched as Apple.

@lmiller1990 thanks for testing. This would be expected, since we are using the system Node to launch the config process. Node.js isn't packaged as a universal app so we have to use whatever arch of Node Cy was launched with.

packages/launcher/lib/utils.ts
```diff
@@ -1,12 +1,33 @@
 import execa from 'execa'
 import cp from 'child_process'
+import os from 'os'
 import Bluebird from 'bluebird'
 
 // export an object for easy method stubbing
 export const utils = {
   execa,
+  spawnWithArch: <T extends cp.SpawnOptionsWithStdioTuple<cp.StdioNull, cp.StdioPipe, cp.StdioPipe>>(cmd: string, args: string[], opts: T) => {
+    if (!(os.platform() === 'darwin' && os.arch() === 'arm64')) {
```
brian-mann: can you add a comment here...

> If we're not on OSX or if the binary was not built in arm64 then we can safely spawn the browser which will get launched as x86_64

packages/launcher/lib/utils.ts
```diff
@@ -1,12 +1,33 @@
...
+      return cp.spawn(cmd, args, opts)
+    }
+
+    // when Cypress spawns processes on Apple macOS with a Intel process as Cypress's parent,
+    // macOS will launch the rosetta-interpreted version (slow) by default. this wraps the
+    // spawn options to prefer arm64 over x86_64 (rosetta)
+    return cp.spawn(
```
brian-mann: can you add a comment here...

> Otherwise if we're on OSX and the binary was built in arm64 then we need to signal to OSX to prefer launching browsers with arm64. There is a situation where even though the binary was built in arm64, if the user upgraded their mac and currently has a x86_64 built version of node, then the parent CLI process that launched the binary will be x86_64 and OSX will not launch the browser using arm64. We use `arch` to tell the OS to prefer the arm64 version. For users who are already using arm64 versions of node, this won't have any effect, so it's safe to do in both situations

packages/launcher/lib/utils.ts
brian-mann: I also might change the logic to be this...

```js
if (os.platform() === 'darwin' && os.arch() === 'arm64') {
  return cp.spawn('arch', [...])
}

return cp.spawn(cmd, args, opts)
```

that'll make the conditional be affirmative instead of negative affirmative and colocates the conditional logic with the `if(...)` statement

packages/launcher/lib/utils.ts
flotwig: 3d1badc4b6fe422b0e60d07f743a24d69b6aed78

packages/launcher/lib/utils.ts
flotwig: 3d1badc4b6fe422b0e60d07f743a24d69b6aed78

Issue #25019, opened 2022-12-07
Cypress 12.0.1 - contains(element,text) inside within fails


swannypp:
### Current behavior

I find I often target content via an element and it's text inside a container with this pattern. but it fails in cypress 12.0.1
```
cy.get('form').within(() => {
    cy.contains('label', text).parent().find('input').type('some text')
});
```

<img width="1595" alt="image" src="https://user-images.githubusercontent.com/92280460/206088110-d043e064-1567-4248-a87d-8e1db0ea4931.png">

### Desired behavior

Targeting content with a `cy.contains(element, 'text')` should work inside a within, as it has done with previous Cypress versions and as indicated in the documentation.

### Test code to reproduce

```
describe('Within Test', () => {
    it('Should be able to target elements within a form', () => {
        cy.visit('https://www.demoblaze.com/');
        cy.contains('a', 'Log in').click();

        cy.get('#logInModal').within(() => {
            cy.get('form').should('be.visible');
            cy.get('label').should('be.visible');

            cy.get('form').within(() => {
                // these work
                cy.contains('Username:').should('be.visible');
                cy.contains('Username').parent().should('be.visible');
                cy.get('input').first().should('be.visible');

                // this doesn't
                cy.contains('label', 'Username').parent().find('input').type('aUsername')
            })
        })
    });
})
```

### Cypress Version

12.0.1

### Node version

v14.18.0

### Operating System

macOS 12.6

### Other

This fault is able to be replicated easily with the given test code supplied.

BlueWinds:
We have some regressions around `.contains()` in Cy12 - looking into this now. Thank you for the clear repro!

BlueWinds:
Looks like we have a regression around `.contains()`  and `<form>` elements in particular (in addition to the other regression over at https://github.com/cypress-io/cypress/issues/25025). Fix in the linked PR; we're planning to get 12.0.2 out in the next day or two containing several quick fixes.

PR #25035, opened 2022-12-07
fix: .contains() properly respects multiple incoming subjects

### LINKED ISSUE #25025: cy.contains with selector and cy.get as a subject stopped working on v12
### Current behavior

```cy.get(<parent_locator>).contains(<child_locator>, <text>)``` was finding elements, but with Cypress v12.0.1 is not finding anything:

image

The changelog was not mentioning anything about cy.contains updates.

### Desired behavior

should work like before changes

### Test code to reproduce

```js
describe("cy.contains bug Cypress v12.0.1", () => {
  beforeEach(() => {
    cy.visit("https://example.cypress.io");
  });

  it("passes", () => {
    cy.get(".container").eq(2).contains("p", "Commands ");
  });

  it("passes", () => {
    cy.get(".container").contains("Commands ");
  });

  it("fails", () => {
    cy.get(".container").contains("p", "Commands ");
  });
});
```

### Cypress Version

12.0.1

### Node version

v16.18.0

### Operating System

macOS 13.0.1

### Debug Logs

_No response_

### Other

_No response_

BlueWinds:
- Closes https://github.com/cypress-io/cypress/issues/25025

### User facing changelog
In Cypress 12.0.0, we introduced a regression where when `.contains()` receives multiple elements as a subject, it only searched inside the first one. This PR fixes that issue.

### Additional details

This PR includes a new driver test covering the case. Another reproduction that doesn't rely on any of our fixtures:
```
it("fails", () => {
  cy.visit("https://example.cypress.io");
  cy.get(".container").eq(2).contains("p", "Commands "); // passes
  cy.get(".container").contains("Commands "); // passes
  cy.get(".container").contains("p", "Commands "); // fails in 12.0.1, passes in this branch
})
```

### PR Tasks

### DIFF
--- a/packages/driver/cypress/e2e/commands/querying/querying.cy.js
+++ b/packages/driver/cypress/e2e/commands/querying/querying.cy.js
@@ -985,6 +985,11 @@ describe('src/cy/commands/querying', () => {
       })
     })
 
+    // https://github.com/cypress-io/cypress/issues/25025
+    it('searches multiple subject elements', () => {
+      cy.get('ul').contains('li', 'asdf 3')
+    })
+
     it('resets the subject between chain invocations', () => {
       const span = cy.$$('.k-in:contains(Quality Control):last')
       const label = cy.$$('#complex-contains label')
@@ -1005,7 +1010,6 @@ describe('src/cy/commands/querying', () => {
 
       cy.get('#click-me a').contains('click').then(($span) => {
         expect($span.length).to.eq(1)
-
         expect($span.get(0)).to.eq(span.get(0))
       })
     })
@@ -1013,7 +1017,6 @@ describe('src/cy/commands/querying', () => {
     it('can find input type=submits by value', () => {
       cy.contains('input contains submit').then(($el) => {
         expect($el.length).to.eq(1)
-
         expect($el).to.match('input[type=submit]')
       })
     })
@@ -1022,7 +1025,6 @@ describe('src/cy/commands/querying', () => {
     it('can find input type=submits by Regex', () => {
       cy.contains(/input contains submit/).then(($el) => {
         expect($el.length).to.eq(1)
-
         expect($el).to.match('input[type=submit]')
       })
     })
@@ -1030,7 +1032,6 @@ describe('src/cy/commands/querying', () => {
     it('has an optional filter argument', () => {
       cy.contains('ul', 'li 0').then(($el) => {
         expect($el.length).to.eq(1)
-
         expect($el).to.match('ul')
       })
     })
@@ -1046,7 +1047,6 @@ describe('src/cy/commands/querying', () => {
     it('searches all els in comma separated filter', () => {
       cy.contains('a,button', 'Naruto').then(($el) => {
         expect($el.length).to.eq(1)
-
         expect($el).to.match('a')
       })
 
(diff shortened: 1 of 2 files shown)


packages/driver/cypress/e2e/commands/querying/querying.cy.js
```diff
@@ -985,6 +985,11 @@ describe('src/cy/commands/querying', () => {
       })
     })
 
+    // https://github.com/cypress-io/cypress/issues/25025
+    it('searches multiple subject elements', () => {
+      cy.get('ul').contains('li', 'asdf 3')
+    })
```
emilyrohrbough, line 991: to ensure this change touched within subject, can we we add something like this as well?
```js
    it('searches multiple subject elements in .within', () => {
      cy.get('body').within(() => {
         cy.get('ul').contains('li', 'asdf 3')
       })
    })
```

chrisbreiding (approved):

AtofStryker (approved):

packages/driver/cypress/e2e/commands/querying/querying.cy.js
BlueWinds: I believe this sort of case is already well covered inside the `.within()` tests.

```
it('scopes additional CONTAINS finders to the subject', () => {
      const span = cy.$$('#nested-div span:contains(foo)')

      cy.contains('foo').then(($span) => {
        expect($span.get(0)).not.to.eq(span.get(0))
      })

      cy.get('#nested-div').within(() => {
        cy.contains('foo').then(($span) => {
          expect($span.get(0)).to.eq(span.get(0))
        })
      })
    })
```

and a couple of other similar tests in within.cy.js I don't think it's valuable to make this test more complicated.

emilyrohrbough: @BlueWinds wasn't my suggestion for a test this issue? https://github.com/cypress-io/cypress/issues/25019

BlueWinds: That issue has  to do with a `<form>` as the subject or withinSubject - it's incredibly specific to forms + contains, which took me quite a while to figure out (I did not expect HTML elements to have different properties).

```
let form = $("<form><input /><input /></form>").get(0) // Raw html element
let ul = $('<ul><li /><li /></ul>').get(0) // Raw html element

form[0] // An <input> element
ul[0] // undefined
```

BlueWinds: I hadn't figured that out when I made this PR, was taking them one at a time. But that other PR has its own test covering that case. `.within()` isn't necessary to hit either of these two bugs - they're entirely within .contains(), users just happened to run into them using `.within()` blocks.

BlueWinds: I had a lazy check that was looking for `subject[0]` as equivelent for "is a jquery object", which worked for every element... except `<form>`. Today I learned.

BlueWinds: The fact that .contains() hijacks the "withinSubject" property of .get() was pre-existing behavior that I really should have cleaned up / renamed when I moved them to be queries, but didn't think of at the time. It's really more a "container element to query inside" - which we call "withinSubject" because .within() is our "scope commands to a container element" command.

cypress-testing-library calls this a "container", and we probably should too. Some future variable renaming to consider.

PR #25038, opened 2022-12-07
fix: .contains() selects proper elements when inside a <form>

### LINKED ISSUE #25019: Cypress 12.0.1 - contains(element,text) inside within fails

BlueWinds:
- Closes https://github.com/cypress-io/cypress/issues/25019

### User facing changelog
In 12.0.0, we had a regression around `.contains()` where if the subject was a `<form>` element, it would only consider the first `<input>` element in the form, rather than every child. This PR fixes the regression.

### Additional details
In addition to the added driver test, this test demonstrates the issue "in the wild":

```
    it('Should be able to target elements within a form', () => {
        cy.visit('https://www.demoblaze.com/');
        cy.contains('a', 'Log in').click();

        cy.get('#logInModal').within(() => {
            cy.get('form').within(() => {
                // these work
                cy.contains('Username:').should('be.visible');
                cy.contains('Username').parent().should('be.visible');
                cy.get('input').first().should('be.visible');

                // this doesn't
                cy.contains('label', 'Username')
            })
        })
    });
```

### PR Tasks

### DIFF
--- a/packages/driver/cypress/e2e/commands/querying/querying.cy.js
+++ b/packages/driver/cypress/e2e/commands/querying/querying.cy.js
@@ -1010,6 +1010,11 @@ describe('src/cy/commands/querying', () => {
       })
     })
 
+    // https://github.com/cypress-io/cypress/issues/25019
+    it('can locate elements contained inside <form> containers', () => {
+      cy.get('#focus').contains('button', 'focusable')
+    })
+
     it('can find input type=submits by value', () => {
       cy.contains('input contains submit').then(($el) => {
         expect($el.length).to.eq(1)
--- a/packages/driver/src/cy/commands/querying/querying.ts
+++ b/packages/driver/src/cy/commands/querying/querying.ts
@@ -175,7 +175,7 @@ export default (Commands, Cypress, cy, state) => {
         let scope = userOptions.withinSubject || cy.getSubjectFromChain(withinSubject)
 
         if (includeShadowDom) {
-          const root = scope || cy.state('document')
+          const root = scope.get(0) || cy.state('document')
           const elementsWithShadow = $dom.findAllShadowRoots(root)
 
           scope = elementsWithShadow.concat(root)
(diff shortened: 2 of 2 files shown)


lmiller1990 (approved):

BlueWinds:
I tested this in conjunction with https://github.com/cypress-io/cypress/pull/25035 before pulling it into its own branch / PR, so they definitely work cleanly together.

AtofStryker (approved):

chrisbreiding (approved):

emilyrohrbough (commented):
are there issues with scoping form inputs to the withinsubject and then running commands within that?

BlueWinds:
> are there issues with scoping form inputs to the withinsubject and then running commands within that?

Scoping .contains() to a `<form>` with `.within()` hits the exact same logic - and runs into the exact same issue - as having a `<form>` as the subject.

I tend not to like writing multiple tests where there are multiple ways a user could encounter the same regression.

PR #25051, opened 2022-12-08
fix: limit the number of globals defined due to the v8 snapshot

ryanthemanuel:
### User facing changelog

Limit the number of globals defined in memory due to the v8 snapshot

### Additional details

We inadvertently created a bunch of globals when setting things up in the v8 snapshot. Specifically:

```
snapshotAuxiliaryData
validateToString
validateElectron
validateFs
validateCrypto
validateFile
integrityCheck
generateSnapshot
supportTypeScript
getSnapshotResult
```

In the spirit of only exposing what is necessary, we are going to ensure that we are only exposing `getSnapshotResult` and `supportTypeScript`. We do this by wrapping the entire snapshot creation inside of an IIFE.

### Steps to test

Ultimately, this can be tested by validate that the tests are passing on the PR and that the new smoke test is passing.

n/a

### PR Tasks


tooling/v8-snapshot/src/generator/blueprint.ts
```diff
@@ -109,101 +109,110 @@ export function scriptFromBlueprint (config: BlueprintConfig): {
```
ryanthemanuel, line 109: This file is more easily reviewed by ignoring whitespace changes.

image

mjhenkes (approved):

AtofStryker (approved):

mschile (approved):

PR #25052, opened 2022-12-08
feat: set up auto prs for snapshot metafile changes

### LINKED ISSUE #24339: Automate the regeneration of V8 snapshot cache files

ryanthemanuel:
- Closes #24339

### User facing changelog

n/a

### Additional details

This PR offers several quality of life improvements for local development with V8 snapshots

* A GitHub action has been created that will automatically update the snapshot cache file
  * This action runs by default on `develop` and `release/**` branches and will issue a PR against the respective branches
  * This action can be manually triggered against any branch providing several options
    * The branch name to run against
    * Whether or not to generate the dependencies from scratch (this takes some time)
    * Whether or not to commit directly to the branch or issue a PR to the branch
  * This action will be scheduled to run weekly on wednesday at midnight UTC. It will generate the cache file from scratch. 
* The cache files for dev and prod have been merged down into one file. `dev` was always a subset of `prod` anyway, so I updated logic to read only `node_modules` from the file if we're in `dev` mode and the whole file if we're in `prod` mode
* The cache file will now only be updated if `V8_UPDATE_METAFILE` is set to true and we're running in `prod` mode. This happens automatically on the GitHub action. This should limit the day to day concerns that people have with respect to the cache file (i.e. you won't ever have to check this file in unless you are explicitly wanting to)

### Steps to test

You can trigger the workflow and run it against a branch to test it:

image

n/a

### PR Tasks


.github/workflows/update_v8_snapshot_cache.yml
```diff
@@ -1,10 +1,134 @@
 name: Update V8 Snapshot Cache
-on: [workflow_dispatch]
+on:
+  schedule:
+    # Run every Wednesday at 00:00 UTC
```
ryanthemanuel, line 4: I chose this so it'll be after the release. I'm ensuring to run from scratch then, so any changes will be caught early rather than right before the release.

BlueWinds:
> This action will be scheduled to run weekly on wednesday at midnight UTC. It will generate the cache file from scratch.

What's the advantage in generating this from scratch in the job, rather than updating the existing one?

.github/workflows/update_v8_snapshot_cache.yml
```diff
@@ -1,10 +1,136 @@
...
+      branch:
+        description: 'Branch to update'
+        required: true
+        default: 'develop'
+      generate_from_scratch:
+        description: 'Generate from scratch'
+        type: boolean
+        default: false
+      commit_directly_to_branch:
+        description: 'Commit directly to branch'
```
BlueWinds, line 22: When would we want to set this true?

ryanthemanuel:
> > This action will be scheduled to run weekly on wednesday at midnight UTC. It will generate the cache file from scratch.
> 
> What's the advantage in generating this from scratch in the job, rather than updating the existing one?

Mainly performance improvements. Generally the way this works is that dependencies can go to a "worse" category, but can't go back the other way. Thus, over time, there may end up being more and more dependencies that could be optimized better but are not. This is a challenging problem to do iteratively, and I have not been able to figure out a way to solve it other than regenerating from scratch periodically.

tooling/v8-snapshot/cache/darwin/snapshot-meta.json
```diff
@@ -3930,8 +3930,8 @@
     "./packages/types/src/util.ts",
     "./packages/types/src/video.ts",
     "./packages/types/src/warning.ts",
-    "./tooling/v8-snapshot/cache/prod-darwin/snapshot-entry.js"
+    "./tooling/v8-snapshot/cache/darwin/snapshot-entry.js"
```
BlueWinds, line 3933: Love to see the reduced duplication. :+1:

BlueWinds:
Looks like this adds several empty yarn.lock files? (because they're empty, I can't comment on them directly from the files tab).

tooling/v8-snapshot/src/generator/snapshot-generator.ts
```diff
@@ -248,16 +218,13 @@ export class SnapshotGenerator {
...
-      previousHealthy: this.previousHealthy.size,
-      previousNoRewrite: this.previousNoRewrite.size,
       forceNoRewrite: this.forceNoRewrite.size,
       auxiliaryData: auxiliaryDataKeys,
     })
   }
 
   private _addGitignore () {
-    const gitignore = 'snapshot.js\nbase.snapshot.js.map\nprocessed.snapshot.js.map\nesbuild-meta.json\nsnapshot-meta.json\nsnapshot-entry.js\n'
+    const gitignore = 'snapshot.js\nbase.snapshot.js.map\nprocessed.snapshot.js.map\nesbuild-meta.json\nsnapshot-entry.js\n'
```
BlueWinds, line 227: > This should limit the day to day concerns that people have with respect to the cache file (i.e. you won't ever have to check this file in unless you are explicitly wanting to)

I'm not sure I follow this. When you say won't have to check it in - will the cache file still show up in the `git diff` output? I haven't been clear on when I should be checking in changes or not, and therefor also not sure how this PR changes it.

BlueWinds (approved):

.github/workflows/update_v8_snapshot_cache.yml
```diff
@@ -1,10 +1,136 @@
 name: Update V8 Snapshot Cache
-on: [workflow_dispatch]
+on:
+  schedule:
+    # Run every Wednesday at 00:00 UTC
+    - cron: '0 0 * * 3'
+  push:
+    branches:
+      - ryanm/feature/v8-snapshots-auto-pr
```
lmiller1990, line 8: Just to check is this a scheduled PIPELINE or a scheduled WORKFLOW? The reason I ask https://circleci.com/docs/configuration-reference/#schedule

> The scheduled workflows feature is set to be deprecated. Using scheduled pipeline. 

This is in a `workflows` file and called `schedule` - just want to verify we are not using a soon-to-be-deprecated feature.

.github/workflows/update_v8_snapshot_cache.yml
```diff
@@ -1,10 +1,136 @@
...
+            const { createPullRequest } = require('./scripts/github-actions/create-pull-request.js')
+
+            await createPullRequest({
+              context,
+              github,
+              baseBranch: '${{ env.BASE_BRANCH }}',
+              branchName: '${{ env.BRANCH_NAME }}',
+              description: 'Update v8 snapshot cache',
+              body: 'This PR was automatically generated by the [update-v8-snapshot-cache](https://github.com/cypress-io/cypress/actions/workflows/update_v8_snapshot_cache.yml) github action.',
+              reviewers: ['ryanthemanuel']
```
lmiller1990, line 135: :raised_hands:

scripts/unit/github-actions/create-pull-request-spec.js
```diff
@@ -0,0 +1,86 @@
+const { expect } = require('chai')
```
lmiller1990, line 1: Random q, I wonder if a better way to test if this is working would be to just have a local check in the monorepo. Eg, when you run `yarn install`, we would (somehow) check the last time the snapshot metadata was updated. If it's been more than <time period>, we could either error or warn "snapshot hasn't been updated in a week; is the automation working?"

Just an idea - if this GHA stopped working, or we changed to use another VCS service, we'd need to remember to do this. Not a requirement for this PR, specifically - just floating an idea to make the monorepo more self maintaining.

tooling/v8-snapshot/src/doctor/determine-deferred.ts
```diff
@@ -14,49 +13,65 @@ export async function determineDeferred (
...
+  let projectHealthy: string[] = []
+  let currentHealthy = opts.nodeModulesOnly ? nodeModulesHealthy : healthy
 
-  if (opts.useHashBasedCache) {
-    hashFilePath = await findHashFile(projectBaseDir)
-    assert(
-      hashFilePath != null,
-      `Unable to find hash file inside ${projectBaseDir}`,
-    )
+  healthy.forEach((dependency) => {
```
lmiller1990, line 32: Going to admit I don't fully grok the finer details of some parts of this PR (like this). I get the general idea though, and it looks like this makes sense.

lmiller1990 (commented):
Looks good, please verify that we are not using a soon-to-be-deprecated CircleCI API. My other comments are just ideas/questions, not blockers.

.github/workflows/update_v8_snapshot_cache.yml
mschile, line 8: Remove before merging?

mschile (approved):

.github/workflows/update_v8_snapshot_cache.yml
ryanthemanuel: If you're wanting to commit something to your feature branch but don't want to go through the rigmarole of a PR to your feature branch.

.github/workflows/update_v8_snapshot_cache.yml
ryanthemanuel: Ah yeah. That's circle ci. I'm using it in a github action where it is not deprecated.

tooling/v8-snapshot/src/generator/snapshot-generator.ts
ryanthemanuel: The key bit happens [here](https://github.com/cypress-io/cypress/pull/25052/files#diff-bba179ad8689b297573f933343816fd5b53a3f39965109f1d139f22f5c6b22d5R110). Basically when you're developing locally, `V8_UPDATE_METAFILE` won't be set and so the cache files won't be written. The only time they will be written is with the github action (or if you explicitly set that environment variable locally for some reason).

.github/workflows/update_v8_snapshot_cache.yml
ryanthemanuel: I could. But I also sort of see this as similar to what we do in the circle ci file. If you're working on it on your branch, update it with your branch, and so on and so on.

lmiller1990 (approved):

PR #25059, opened 2022-12-08
fix: log error on reject with string content

### LINKED ISSUE #24915: Cypress prints `(uncaught exception) undefined: undefined` when promise rejects with a string
### Current behavior

When the app code rejects with a string:
- developer console displays the error properly
- cypress displays `(uncaught exception) undefined: undefined` and fails the test

### Desired behavior

- cypress displays the exact error cause (reject's string content) and fails the test

### Test code to reproduce

https://github.com/geritol/cypress-bug-repro

### Cypress Version

11.2.0

### Node version

v16.18.1

### Operating System

macOS 12.6

### Debug Logs

_No response_

### Other

_No response_

geritol:
- Closes #24915 

### User facing changelog

Fixes issue where unhadled promise rejectection was being logged as `(uncaught exception) undefined: undefined`

### Additional details

Change was needed to help debugging failures where user has only access to cy console output.

-

### PR Tasks

### DIFF
--- a/packages/app/cypress/e2e/runner/reporter.errors.cy.ts
+++ b/packages/app/cypress/e2e/runner/reporter.errors.cy.ts
@@ -180,6 +180,16 @@ describe('errors ui', {
       ],
     })
 
+    verify('spec unhandled rejection with string content', {
+      uncaught: true,
+      column: 20,
+      originalMessage: 'Unhandled promise rejection with string content from the spec',
+      message: [
+        'The following error originated from your test code',
+        'It was caused by an unhandled promise rejection',
+      ],
+    })
+
     verify('spec unhandled rejection with done', {
       uncaught: true,
       column: 20,
--- a/packages/app/cypress/e2e/runner/reporter.errors.cy.ts
+++ b/packages/app/cypress/e2e/runner/reporter.errors.cy.ts
@@ -188,6 +188,8 @@ describe('errors ui', {
         'The following error originated from your test code',
         'It was caused by an unhandled promise rejection',
       ],
+      stackRegex: /.*/,
+      hasCodeFrame: false,
     })
 
     verify('spec unhandled rejection with done', {
--- a/packages/driver/cypress/e2e/cypress/error_utils.cy.ts
+++ b/packages/driver/cypress/e2e/cypress/error_utils.cy.ts
@@ -671,14 +671,24 @@ describe('driver/src/cypress/error_utils', () => {
       }
     })
 
-    it('calls Cypress.log with error type and message when error is instance of Error', () => {
+    it('calls Cypress.log with error name and message when error is instance of Error', () => {
       $errUtils.logError(cypressMock, 'error', new Error('Some error'))
       expect(cypressMock.log).to.have.been.calledWithMatch(sinon.match.has('message', `Error: Some error`))
     })
 
-    it('calls Cypress.log with error type and message when error a string', () => {
-      $errUtils.logError(cypressMock, 'error', 'Some error')
-      expect(cypressMock.log).to.have.been.calledWithMatch(sinon.match.has('message', `Error: \"Some error\"`))
+    it('calls Cypress.log with error name and message when error a string', () => {
+      $errUtils.logError(cypressMock, 'error', 'Some string error')
+      expect(cypressMock.log).to.have.been.calledWithMatch(sinon.match.has('message', `Error: \"Some string error\"`))
+    })
+
+    it('calls Cypress.log with default error name and provided message message when error is an object with a message', () => {
+      $errUtils.logError(cypressMock, 'error', { message: 'Some object error with message' })
+      expect(cypressMock.log).to.have.been.calledWithMatch(sinon.match.has('message', `Error: Some object error with message`))
+    })
+
+    it('calls Cypress.log with error name and message when error is an object', () => {
+      $errUtils.logError(cypressMock, 'error', { err: 'Error details' })
+      expect(cypressMock.log).to.have.been.calledWithMatch(sinon.match.has('message', `Error: {"err":"Error details"}`))
     })
   })
 })
--- a/packages/driver/cypress/e2e/cypress/error_utils.cy.ts
+++ b/packages/driver/cypress/e2e/cypress/error_utils.cy.ts
@@ -5,6 +5,7 @@ allowTsModuleStubbing()
 import $stackUtils from '@packages/driver/src/cypress/stack_utils'
 import $errUtils, { CypressError } from '@packages/driver/src/cypress/error_utils'
 import $errorMessages from '@packages/driver/src/cypress/error_messages'
+import sinon from 'sinon'
 
 describe('driver/src/cypress/error_utils', () => {
   context('.modifyErrMsg', () => {
@@ -90,7 +91,7 @@ describe('driver/src/cypress/error_utils', () => {
     })
 
     it('attaches onFail to the error when it is a function', () => {
-      const onFail = function () {}
+      const onFail = function () { }
       const fn = () => $errUtils.throwErr(new Error('foo'), { onFail })
 
       expect(fn).throw().and.satisfy((err) => {
@@ -561,7 +562,7 @@ describe('driver/src/cypress/error_utils', () => {
 
     it('does not error if no last log', () => {
       state.returns({
-        getLastLog: () => {},
+        getLastLog: () => { },
       })
 
       const result = $errUtils.createUncaughtException({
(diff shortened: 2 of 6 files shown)


packages/driver/src/cypress/error_utils.ts
```diff
@@ -572,6 +574,25 @@ const logError = (Cypress, handlerType: HandlerType, err, handled = false) => {
   })
 }
 
+const isLogabbleError = (error: unknown): error is Error => {
```
chrisbreiding:
```suggestion
const isLoggableError = (error: unknown): error is Error => {
```

packages/driver/src/cypress/error_utils.ts
```diff
@@ -572,6 +574,25 @@ const logError = (Cypress, handlerType: HandlerType, err, handled = false) => {
...
+  return (
+    typeof error === 'object' &&
+    error !== null &&
+    'message' in error &&
+    typeof (error as Record<string, unknown>).message === 'string'
+  )
+}
+
+const toLoggableError = (maybeError: unknown): Error => {
+  if (isLogabbleError(maybeError)) return maybeError
```
chrisbreiding:
```suggestion
  if (isLoggableError(maybeError)) return maybeError
```

packages/driver/src/cypress/error_utils.ts
```diff
@@ -549,9 +549,11 @@ const errorFromUncaughtEvent = (handlerType: HandlerType, event) => {
     errorFromProjectRejectionEvent(event)
 }
 
-const logError = (Cypress, handlerType: HandlerType, err, handled = false) => {
+const logError = (Cypress, handlerType: HandlerType, err: unknown, handled = false) => {
+  const error = toLoggableError(err)
+
   Cypress.log({
-    message: `${err.name}: ${err.message}`,
+    message: `${error.name}: ${error.message}`,
```
chrisbreiding:
```suggestion
    message: `${error.name || 'Error'}: ${error.message}`,
```

It's an unlikely edge case, but since we're only checking that the error has a message to be loggable (which I think is fine), this will keep `reject({ message: 'error message' })` from logging `undefined: error message`.

packages/driver/cypress/e2e/cypress/error_utils.cy.ts
```diff
@@ -5,6 +5,7 @@ allowTsModuleStubbing()
 import $stackUtils from '@packages/driver/src/cypress/stack_utils'
 import $errUtils, { CypressError } from '@packages/driver/src/cypress/error_utils'
 import $errorMessages from '@packages/driver/src/cypress/error_messages'
+import sinon from 'sinon'
```
chrisbreiding:
```suggestion

const { sinon } = Cypress
```

`Cypress` already includes `sinon`, so we can slim down the spec file by not importing it again.

chrisbreiding (changes requested):
Thanks for the contribution! The tests you added look good, but could you also add a test [in this file](https://github.com/cypress-io/cypress/blob/7826cab137ef5bbaed5051e1fa7a55eb2647f6b9/system-tests/project-fixtures/runner-specs/cypress/e2e/errors/uncaught.cy.js#L58)? That will give us better e2e coverage of this. 

Those tests are a little unusual because they run Cypress within Cypress. The [tests in here](https://github.com/cypress-io/cypress/blob/7826cab137ef5bbaed5051e1fa7a55eb2647f6b9/packages/app/cypress/e2e/runner/reporter.errors.cy.ts#L173) are the outer tests that verify the results of the inner tests. [This file](https://github.com/cypress-io/cypress/blob/7826cab137ef5bbaed5051e1fa7a55eb2647f6b9/system-tests/project-fixtures/runner-specs/cypress/e2e/errors/uncaught.cy.js#L58) has the inner tests and looks more like a standard Cypress spec. You can follow the pattern of the  `spec unhandled rejection` test. The tests can be run locally by running `yarn cypress:open` within `packages/app`.

The only trick to writing the test is that the first argument passed to verify in the outer test  (e.g. `verify('spec unhandled rejection', { ... })`) needs to match the name of the test in the inner test (e.g. `it('spec unhandled rejection', ...)`).

packages/driver/src/cypress/error_utils.ts
emilyrohrbough: We likely can can simplify this check to either use the Error or the value provided:

```suggestion
message: isLogabbleError(err) ? `${err.name || 'Error'}: ${err.message}` : err
```

packages/driver/src/cypress/error_utils.ts
```diff
@@ -572,6 +574,25 @@ const logError = (Cypress, handlerType: HandlerType, err, handled = false) => {
   })
 }
 
+const isLogabbleError = (error: unknown): error is Error => {
+  return (
+    typeof error === 'object' &&
+    error !== null &&
+    'message' in error &&
+    typeof (error as Record<string, unknown>).message === 'string'
```
emilyrohrbough: Do we need such an aggressive check? Would `_.isObject(error)` be sufficient?

packages/driver/src/cypress/error_utils.ts
geritol: @emilyrohrbough I think your suggestion makes things more complicated, it would also not pass the `err name: error cause` format to Cypress.log

packages/driver/src/cypress/error_utils.ts
geritol: we also need to verify if message is a string

geritol:
@chrisbreiding tried to add the test you mentioned, bumped into two issues which I need help on:
1) I cannot commit `unner-specs/cypress/e2e/errors/uncaught.cy.js` eslint precommit hook fails:
```
✖ eslint --fix:

/Users/geritol/work/cypress/system-tests/project-fixtures/runner-specs/cypress/e2e/errors/uncaught.cy.js
   5:5  error  'cy' is not defined  no-undef
   6:5  error  'cy' is not defined  no-undef
  10:5  error  'cy' is not defined  no-undef
  11:5  error  'cy' is not defined  no-undef
  15:5  error  'cy' is not defined  no-undef
  16:5  error  'cy' is not defined  no-undef
  20:5  error  'cy' is not defined  no-undef
  21:5  error  'cy' is not defined  no-undef
  22:5  error  'cy' is not defined  no-undef
  26:5  error  'cy' is not defined  no-undef
  27:5  error  'cy' is not defined  no-undef
  28:5  error  'cy' is not defined  no-undef
  35:5  error  'cy' is not defined  no-undef
  39:5  error  'cy' is not defined  no-undef
  40:5  error  'cy' is not defined  no-undef
  48:5  error  'cy' is not defined  no-undef
  61:5  error  'cy' is not defined  no-undef
  67:5  error  'cy' is not defined  no-undef
  78:5  error  'cy' is not defined  no-undef

✖ 19 problems (19 errors, 0 warnings)
```

2) Tests fails because it seems It is expected for the error to be printed in an other place too aside from being loggeg (after the `>`)
kép

Not sure where the logic of that is, maybe we need to add this transformation to a higher level in the code?

chrisbreiding:
> 1. I cannot commit unner-specs/cypress/e2e/errors/uncaught.cy.js eslint precommit hook fails:

Sorry about that. Seems to be an oversight with our eslint configuration. I should have that fixed in https://github.com/cypress-io/cypress/pull/25089, so once it's merged, you can update from origin/develop and it will unblock you.

> 2. Tests fails because it seems It is expected for the error to be printed in an other place too aside from being loggeg (after the >)
> 
> Not sure where the logic of that is, maybe we need to add this transformation to a higher level in the code?

Great. Looks like the test is doing its job because we do want the error to show up there as well. 

You're right about needing to move the transformation higher. I think [this is where we should do it](https://github.com/cypress-io/cypress/blob/89c013fcedc2509850ec820f938d33f08f9cbb42/packages/driver/src/cypress/error_utils.ts#L530-L544). If it ensures that the `originalErr` and `err` properties returned are errors or at least error-like objects, the rest should work, including the logging.

chrisbreiding:
The eslint issue is fixed in `origin/develop`, so you should be able to modify and commit changes to that spec file now.

geritol:
@chrisbreiding e2e test up, now failing because of the lack of stack strace on the new error log (if I get it right)
In this case we cant really show a stack trace, since we did not reject with an instace of `Error`. What should we do?
kép

packages/app/cypress/e2e/runner/reporter.errors.cy.ts
```diff
@@ -180,6 +180,16 @@ describe('errors ui', {
...
     })
 
+    verify('spec unhandled rejection with string content', {
+      uncaught: true,
+      column: 20,
+      originalMessage: 'Unhandled promise rejection with string content from the spec',
+      message: [
+        'The following error originated from your test code',
+        'It was caused by an unhandled promise rejection',
+      ],
```
chrisbreiding, line 190:
```suggestion
      ],
      stackRegex: /.*/,
      hasCodeFrame: false,
```

@geritol Adding this configuration should fix the failure.

packages/driver/src/cypress/error_utils.ts
```diff
@@ -572,6 +578,25 @@ const logError = (Cypress, handlerType: HandlerType, err, handled = false) => {
...
+
+const isLoggabbleError = (error: unknown): error is LoggableError => {
+  return (
+    typeof error === 'object' &&
+    error !== null &&
+    'message' in error)
+}
+
+const toLoggableError = (maybeError: unknown): LoggableError => {
+  if (isLoggabbleError(maybeError)) return maybeError
```
chrisbreiding:
```suggestion
const isLoggableError = (error: unknown): error is LoggableError => {
  return (
    typeof error === 'object' &&
    error !== null &&
    'message' in error)
}

const toLoggableError = (maybeError: unknown): LoggableError => {
  if (isLoggableError(maybeError)) return maybeError
```

packages/app/cypress/e2e/runner/reporter.errors.cy.ts
geritol: @chrisbreiding works 🙌

chrisbreiding (approved):

emilyrohrbough (approved):

PR #25067, opened 2022-12-08
feat: IATR-M0 Debug Page

warrensplayer:
- Closes https://github.com/cypress-io/cypress/issues/24441

### User facing changelog

**NOT APPLICABLE** This feature is being merged into a feature branch to be released with milestone 1.0 for the In App Test Results epic

### Additional details

_Background_
This PR adds a new page to the Cypress called "Debug". This page will contain the features for the project known as In App Test Results [IATR].  This project will add the ability for a developer that is connected to the Cypress Cloud to view tests that have been recorded as failed to a Cloud run in the app. The goal of the milestone 0 of this feature will allow the developer to easily run the test locally to reproduce the error. Milestone 0 is not going to be released to users, but will remain held in a feature branch.

**NOTE** The code introduces some new GraphQL fields from the Cypress Cloud that are currently being held in a feature branch on the Cloud side.  When running this branch for testing, you can still run the component tests that are introduced or updated in this branch.  You will not be able to test the full functionality in the app itself yet without running a local Cypress Cloud instance.  The Cloud team is working on setting up a "preview" environment that can be used for testing the app while we work through M1 (milestone 1) work and to use for acceptance testing before the Cloud code gets released to production.

_What is new_
* A new page in the App called "Debug" that appears in the sidebar between "Runs" and "Settings"
* The page has text only empty states that will be enhanced in M1 for:
  * Not logged in
  * Not connected
  * No runs  
* Uses the components for the run information and failed spec with tests added in these PRs: #24722 and #24877 
* New components added
  * `packages/app/src/debug/DebugContainer.vue`
  * `packages/app/src/debug/DebugSpecList.vue`
  * `packages/app/src/pages/Debug.vue`
* Changes in the cloud graphql schema caused the need to update a few types and fill in new required fields.  The main need for this was in `RunStatusDots.vue` and the e2e test `packages/app/cypress/e2e/specs_list_latest_runs.cy.ts`

### Steps to test

The easiest way to test will just be to run the tests that were added or changed for this issue.
* `packages/app/src/debug/DebugContainer.cy.tsx`
* `packages/app/cypress/e2e/sidebar_navigation.cy.ts`
* `packages/app/src/debug/DebugPageHeader.cy.tsx`
* `packages/app/src/debug/DebugSpec.cy.tsx`
* `packages/app/src/navigation/SidebarNavigationRow.cy.tsx`
* `packages/app/cypress/e2e/specs_list_latest_runs.cy.ts`

_New sidebar nav for "Debug"_
Screen Shot 2022-12-08 at 11 28 24 AM

_Basic empty states that will be enhanced later_
* Not logged in
Screen Shot 2022-12-08 at 11 28 56 AM
* Not connected
Screen Shot 2022-12-08 at 11 29 03 AM
* No runs
Screen Shot 2022-12-08 at 11 29 12 AM

_Use new components to show runs and failed tests_
Screen Shot 2022-12-09 at 2 06 52 PM

### PR Tasks


packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -137,14 +138,13 @@ fragment DebugPage on CloudRun {
 
 const props = defineProps<{
   gql: DebugPageFragment
-  commitsAhead: string
-  commitHash: string
-  runNumber: number
+  commitsAhead: number
```
mike-plummer, line 142: I only see "commits ahead" references in the mockups, but I imagine a user could be commits behind if they haven't done a pull recently. Are we not handling/displaying that scenario?

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -137,14 +138,13 @@ fragment DebugPage on CloudRun {
...
 }>()
 
 const debug = computed(() => props.gql)
 
 const relativeCreatedAt = computed(() => dayjs(new Date(debug.value.createdAt!)).fromNow())
+const totalDuration = computed(() => dayjs.duration(debug.value.totalDuration!).format('HH:mm:ss').replace(/^0+:/, ''))
```
mike-plummer: A comment would be helpful here to describe the regex behavior. Also, may just be better to use a `?? 0` rather than the non-null assertion

mike-plummer: Also, we're confident cloud will never report back a duration >= 23:59:59? Obviously would be a hung test or something else broken, but this format would roll over if it happened to

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -112,23 +112,24 @@ import CommitIcon from '~icons/cy/commit_x14'
...
+
+const { t } = useI18n()
 
-// runNumber and commitHash dont currently exist in the query and therefore are being obtained as props instead
 gql`
 fragment DebugPage on CloudRun {
 	id
+  runNumber
 	createdAt
 	status
```
mike-plummer: nitpick, but somehow parts of this query got committed using tabs for indentation

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -19,22 +19,22 @@
...
-          <div class="w-px h-3 my-6px bg-gray-100" />
+          <div class="bg-gray-100 h-3 my-6px w-px" />
           <span
-            class="font-normal text-orange-500 px-2 mx-px items-center"
+            v-if="props.commitsAhead"
+            class="font-normal mx-px px-2 text-orange-500 items-center"
             data-cy="debug-commitsAhead"
           >
-            {{ props.commitsAhead }}
+            {{ t('debugPage.header.commitsAhead', {commits: props.commitsAhead}) }}
```
mike-plummer: `vue-i18n` has built-in support for auto-pluralization that would be ideal here so we don't display a message like "You are 1 commits ahead". See the `runner.selectorPlayground.matches` in `en-US.json` for an example

packages/app/src/debug/DebugContainer.vue
```diff
@@ -0,0 +1,100 @@
...
+      <DebugPageHeader
+        :gql="run"
+        :commits-ahead="0"
+      />
+      <DebugSpecList
+        :specs="debugSpecsArray"
+      />
+    </div>
+    <div
+      data-cy="debug-empty"
```
mike-plummer, line 16: Should this be behind a `v-else`?

packages/app/src/debug/DebugContainer.vue
```diff
@@ -0,0 +1,100 @@
+<template>
+  <div>
+    <div
+      v-if="loginConnectStore.user.isLoggedIn && loginConnectStore.project.isProjectConnected && run"
+    >
+      <DebugPageHeader
+        :gql="run"
+        :commits-ahead="0"
```
mike-plummer, line 8: Temporary placeholder value?

packages/app/src/debug/DebugContainer.vue
```diff
@@ -0,0 +1,100 @@
...
+    <div
+      data-cy="debug-empty"
+    >
+      <div
+        v-if="!loginConnectStore.user.isLoggedIn"
+      >
+        {{ t('debugPage.notLoggedIn') }}
+      </div>
+      <div
+        v-if="loginConnectStore.user.isLoggedIn && !loginConnectStore.project.isProjectConnected"
```
mike-plummer: Could probably simplify these subsequent conditionals by using `v-else-if`

packages/app/src/debug/DebugContainer.vue
```diff
@@ -0,0 +1,100 @@
...
+      >
+        {{ t('debugPage.notConnected' ) }}
+      </div>
+      <div
+        v-if="loginConnectStore.user.isLoggedIn && loginConnectStore.project.isProjectConnected && !run"
+      >
+        {{ t('debugPage.noRuns') }}
+      </div>
+    </div>
+    <!-- <pre>{{ JSON.stringify(props.gql, null, 2) }}</pre> -->
```
mike-plummer: Remove?

packages/app/src/debug/DebugContainer.cy.tsx
```diff
@@ -0,0 +1,161 @@
...
+      })
+
+      // Only asserting that it is rendering the components for failed specs
+      cy.findByTestId('debug-header').should('be.visible')
+      cy.findByTestId('debug-spec-item').should('be.visible')
+    })
+  })
+
+  describe('testing util function: debugMapping', () => {
+    const mappingList = (specs: {id: string}[], tests: {specId: string, id: string}[]) => {
```
lmiller1990: This function looks like it returns a function with the exact same arguments; what is the purpose of having `mappingList`? Can we just call `specsList` instead?

packages/app/src/debug/DebugContainer.cy.tsx
```diff
@@ -0,0 +1,161 @@
...
+        { specId: '123', id: 'random1' },
+        { specId: '456', id: 'random2' },
+        { specId: '456', id: 'random3' },
+        { specId: '789', id: 'random4' },
+        { specId: '123', id: 'random6' },
+      ]
+
+      const debugMappingArray = mappingList(specs, tests)
+
+      expect(debugMappingArray).to.have.length(3)
```
lmiller1990: Unit testing with Cypress is quite a painful - we should definitely look to get [Vitest](https://vitest.dev/) in (using Vite, like for the frontend - so using the same config - but a very fast, terminal based runner, similar to Jest - which is perfect for these kind of tests).

packages/app/src/debug/DebugContainer.vue
```diff
@@ -0,0 +1,100 @@
...
+}>()
+
+const loginConnectStore = useLoginConnectStore()
+
+const run = computed(() => {
+  return props.gql.currentProject?.cloudProject?.__typename === 'CloudProject' && props.gql.currentProject.cloudProject.runByNumber
+})
+
+const debugSpecsArray = computed(() => {
+  if (props.gql.currentProject?.cloudProject?.__typename === 'CloudProject') {
```
lmiller1990: It sometimes feels like we are building an additional layer in the client, instead of just serving up the data we need. The example I see here is two derived properties based on the fact we *might* have cloud project. Instead, I wonder if we should be doing this on the server, and move the [`specsList`] function there, too. Then client code would become:

```js
const debugSpecs = `computed(() => props.gql.currentProject.cloudProject?.debugSpecs
```

Eg - one of the main things that sold me on GraphQL was the ability to easily add new fields/properties derived from existing data based on changing business requirements (unlike modifying a REST endpoint, which often means needs to modify a lot of other things, too).

Might be more of a philosophical discussion, but the biggest gripe I'm having with the current state of App is the immense complexity of all the computed values based on GraphQL. I wonder if we are doing something wrong? WDYT?

packages/app/src/debug/DebugPageHeader.cy.tsx
```diff
@@ -20,12 +20,13 @@ describe('<DebugPageHeader />', {
...
             result.commitInfo.branch = 'feature/DESIGN-183'
             result.commitInfo.authorName = 'cypressDTest'
+            result.commitInfo.sha = 'b5e6fde'
           }
         }
       },
       render: (gqlVal) => {
         return (
-          <DebugPageHeader gql={gqlVal} runNumber={468} commitsAhead='You are 2 commits ahead' commitHash='b5e6fde'/>
+          <DebugPageHeader gql={gqlVal} commitsAhead={2}/>
```
lmiller1990, line 29: Can you be commits behind? Eg, you look at a test result but you haven't done `git pull` in a while. In which case this prop might be `commitDelta` or something more representative of this state.

packages/app/src/debug/DebugSpec.vue
```diff
@@ -80,8 +75,8 @@ export interface Spec {
 }
 
 export interface TestResults {
-  id: string
-  titleParts: string[]
+  readonly id: string
```
lmiller1990, line 78: Are these readonly for the purpose of making GraphQL types happy?

I've run into this a lot, I recently found out you can turn off the readonly config, which seems to make things much nicer without any real downsides (just be careful not to mutate props, pretty easy).

Here's how I turned it off in my other project (also using Nexus and GraphQL Codegen) https://github.com/lmiller1990/eleutheria/blob/main/packages/frontend/codegen.yml#L5

Might be good to do this at some point, either here or another PR.

packages/app/src/debug/DebugSpecList.vue
```diff
@@ -0,0 +1,93 @@
...
+
+const specs = computed(() => {
+  return props.specs.map((specItem) => {
+    const fileName = specItem.spec.basename
+    const fileExtension = specItem.spec.extension
+
+    return {
+      spec: {
+        id: specItem.spec.id,
+        path: specItem.spec.path.replace(fileName + fileExtension, ''),
```
lmiller1990: Didn't check but I am kind of surprised we don't have a `folder` or `directory` properly already - we do show specs based on folder on the main specs page, is that derived in the same way as you are doing it here?

packages/app/src/debug/utils/DebugMapping.ts
```diff
@@ -0,0 +1,27 @@
+export const specsList = <S extends {id: string}, T extends {specId: string}>(specs: readonly S[], tests: readonly T[]): {spec: S, tests: T[]}[] => {
```
lmiller1990, line 1: See other comment - might not be practical, if so ignore, but it sure would be nice to move this to the backend and have GraphQL serve it up in the format we want.

packages/app/src/debug/DebugSpecList.vue
lmiller1990: Maybe `basename`?

lmiller1990:
It sure does look nice! Excited for this feature.

packages/app/src/debug/DebugContainer.cy.tsx
warrensplayer: Yes, I agree. Having a pattern to do unit testing in our front end code would be most helpful. I got to thinking if there would be a use case for having some sort of Cypress plugin to render some sort of visual for unit tests in cypress, simiar to what Filip Hiric did with the [cypress-plugin-api](https://github.com/filiphric/cypress-plugin-api).  I know I have been part of conversations where the idea of pulling more types of testing into Cypress has been discussed to have fewer places for developers to go when doing testing, but not sure where those ideas stand.

packages/app/src/debug/DebugContainer.cy.tsx
warrensplayer: Yes!  Good catch. Will update.

packages/app/src/debug/utils/DebugMapping.ts
warrensplayer: Yes, I think this makes sense as well.  Looking at doing that in the final IATR M0 issue.

packages/app/src/debug/DebugContainer.vue
warrensplayer: Yes.  Will be replaced with the actual lookup in the final IATR M0 issue.

packages/app/src/debug/DebugContainer.vue
warrensplayer: Yes.

packages/app/src/debug/DebugContainer.vue
warrensplayer: Much better.

packages/app/src/debug/DebugContainer.vue
warrensplayer: Gone.

packages/app/src/debug/DebugContainer.vue
warrensplayer: I have been having the same feeling as well in a lot of situations, especially around the patterns with the cloud data.  In this instance, the `cloudProject` field is trying to represent four states with the different types.  I am still working with the last remaining IATR M0 issue and still playing around with what to push down to the server layer. I am going to think more about this pattern as it is used elsewhere in the code.  If you have any patterns in mind about these different states, I would love to compare notes.

packages/app/src/debug/DebugPageHeader.cy.tsx
warrensplayer: The use cases for this feature do not cover trying to look forward in commits that you have not pulled into your current local branch.  The scenario this is covering is only that the query to the cloud can only find a "relevant run" to your local environment that is earlier than commit you are currently on.  The normal case for this would be that you made some local changes, committed them locally, and either have not pushed them or have and CI has not run on them.  Then you would be `X` commits ahead of the "relevant run" you are currently looking at on the "Debug" page.  There are no curent plans for this feature to show you a run that is on a commit ahead of what you have locally.  There is work in a future milestone to have a different message in a banner that tells the developer that there are newer runs that they could see if they do `git pull`.

packages/app/src/debug/DebugPageHeader.vue
warrensplayer: Not sure how that happened, but fixed.

packages/app/src/debug/DebugPageHeader.vue
warrensplayer: I will do a better job of writing up the support scenarios in future PRs for the IATR project, but that is not a scenario we are solving with the data in this page header.  This is meant to indicate that the current commit the developer is looking at locally is `X` commits ahead of the "relevant run" being shown on the "Debug" page.  There is not currently a use case where the "Debug" page will show a run that is ahead of the commit that the developer currently has.  There is work in the next milestone to show a banner on this page to let the developer know there are more recent runs that they would be able to see if they perform a `git pull` to pull down the newer commits first.

packages/app/src/debug/DebugPageHeader.vue
warrensplayer: Good catch on this one as well.

packages/app/src/debug/DebugPageHeader.vue
warrensplayer: Comment added.

packages/app/src/debug/DebugSpecList.vue
warrensplayer: There is not a `folder` or `directory` property in the cloud field, nor do I see this in the data the SpecsList uses.  The `basename` field is the filename.  In a quick look, I was not able to see where the SpecsList page parses the data it expects to pull out the directories it shows in the tree strucure, but will keep looking as I am working on the other backend code for IATR

packages/app/src/debug/DebugSpec.vue
warrensplayer: Yes, just to make types happy.  I would definitely like to have a conversation to determine if that configuration change would make sense or if we should handle that some other way.

packages/app/src/debug/DebugContainer.cy.tsx
```diff
@@ -0,0 +1,157 @@
...
+        { specId: '456', id: 'random2' },
+        { specId: '456', id: 'random3' },
+        { specId: '789', id: 'random4' },
+        { specId: '123', id: 'random6' },
+      ]
+
+      const debugMappingArray = specsList(specs, tests)
+
+      expect(debugMappingArray).to.have.length(3)
+      debugMappingArray.forEach((mapping) => {
```
ZachJW34: Would be more clear to `expect(debugMappingArray).to.deep.equal({ ... })` rather than have the conditionals.

packages/app/src/debug/DebugContainer.vue
```diff
@@ -0,0 +1,100 @@
...
+    const specs = props.gql.currentProject.cloudProject.runByNumber?.specs || []
+    const tests = props.gql.currentProject.cloudProject.runByNumber?.testsForReview || []
+
+    return specsList(specs, tests)
+  }
+
+  return []
+})
```
ZachJW34:
```suggestion
const run = computed(() => {
  return props.gql.currentProject?.cloudProject?.__typename === 'CloudProject' ? props.gql.currentProject.cloudProject.runByNumber : null
})

const debugSpecsArray = computed(() => {
  if (run.value) {
    const specs = run.value.specs || []
    const tests = run.value.testsForReview || []

    return specsList(specs, tests)
  }

  return []
})
```

Can use the computed result above.

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -112,40 +112,49 @@ import CommitIcon from '~icons/cy/commit_x14'
...
 
+/*
+  Format duration to in HH:mm:ss format. The `totalDuration` field is milliseconds. Remove the leading "00:" if the value is less
+  than an hour. Currently, there is no expectation that a run duration will be greater 24 hours or greater, so it is okay that
+  this format would "roll-over" in that scenario.
+  Ex: 1 second which is 1000ms = 00:01
+  Ex: 1 hour and 1 second which is 3601000ms = 01:00:01
+*/
+
+const totalDuration = computed(() => dayjs.duration(debug.value.totalDuration!).format('HH:mm:ss').replace(/^0+:/, ''))
```
ZachJW34: We have the same function [here](https://github.com/cypress-io/cypress/blob/eb93a7963bad41c8c59595eebe363fb2e16edc94/packages/app/src/runs/RunCard.vue#L123). Would be nice to create a utility method.

ZachJW34 (approved):
Few comments, nothing blocking. LGTM!

packages/app/src/debug/DebugContainer.cy.tsx
warrensplayer: I like that too

mike-plummer (approved):
Thanks for addressing my comments, no issues here

packages/app/src/debug/DebugPageHeader.vue
warrensplayer: Made a composable to encapsulate the logic

packages/app/src/composables/useDurationFormat.ts
```diff
@@ -0,0 +1,13 @@
...
+import { dayjs } from '../runs/utils/day.js'
+
+/*
+  Format duration to in HH:mm:ss format. The `totalDuration` field is milliseconds. Remove the leading "00:" if the value is less
+  than an hour. Currently, there is no expectation that a run duration will be greater 24 hours or greater, so it is okay that
+  this format would "roll-over" in that scenario.
+  Ex: 1 second which is 1000ms = 00:01
+  Ex: 1 hour and 1 second which is 3601000ms = 01:00:01
+*/
+export function useDurationFormat (value: number | Ref<number>) {
```
ZachJW34, line 11: I think this would be nicer as a simple utility method rather than a composable. We can simply call the function inside a composable.

lmiller1990 (approved):
This is looking good to me.

packages/app/src/composables/useDurationFormat.ts
warrensplayer: Just going to revert that composable option at this time.  I was trying to encapsulate some reactive functionality but realized that it was not providing what the current pattern is doing.  I do agree that we need a place for this type of logic to be generally shared, plus a way to unit test it as Lachlan has suggested with Vitest.  Will do that in a future iteration.

PR #25072, opened 2022-12-08
feat: IATR-M1.0 Debug page Header Update

### LINKED ISSUE #24851: [IATR](M1.0) Debug page run header updates
### Debug page header should match new M1 designs 

* Updates include
  * The run status on the second row has the status name removed and followed by the run number to be colored the same
    * Make sure all run statuses are accounted for 
  * Run number is removed from the first row following the commit description
  * Test counts no longer have the status icon and name
  * Flaky count is shown after result counts if value > 0

### Before
Screen Shot 2022-12-04 at 4 31 35 PM

### After
Screen Shot 2022-12-04 at 4 26 27 PM

[Figma designs](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/QkCp2F69X0qyLo7qnYQ67e/Cypress-App?node-id=21180%3A195482&t=mgpzpKNCFo6r0Y3a-1)

amehta265:
- Closes #24851 

### User facing changelog

This PR aims to update the Debug Page Header to provide more information. It accomplishes the following:
1. Removal of the `run Number` from the first row so that the "commits Ahead" message is the only element visible in the top section. 
2. Adding a `run status` component in the bottom section. This section contains a `status icon` and the `run number` with matching colors depending on the status returned from the cloud. This takes into account all current run statuses (Passed, Failed, Errored, Cancelled, Running)
3. Removing the `status icon` and `status name` from the test counts
4. Adding a `flaky badge` which comprises of the total flaky tests number and the "Flaky" text separated by a vertical border. This is only visible when the total number of flaky tests is > 0 and the value is not null

- Refer to the images below. Each numbered point above has a corresponding visual queue in the pictures attached below

### Additional details

- A color mapping has been added to `debugPageHeader.vue`. The design for the run status requires matching colors for the text (run number) and icon. In order to render a specific color depending on the status I have created a simple mapping for each possible status and the respective color

### Steps to test

This issue has 2 component tests. Checkout to this branch to run the tests
Both `DebugPageHeader.vue` and `DebugResults.vue` exist in the `packages/app/` folder. To run these tests:
1. Change directory to `packages/app`
2. Run `yarn dev`
3. Navigate to Component Testing and run `DebugPageHeader.cy.tsx` and `DebugResults.cy.tsx`

Updated Page Header:
Screenshot 2022-12-08 at 12 15 47 PM

Old Page Header:
205516772-cb9ec9c3-950e-4a92-988e-99bcfb4d98a4

### PR Tasks


packages/app/src/debug/DebugPageHeader.cy.tsx
```diff
@@ -32,27 +34,70 @@ describe('<DebugPageHeader />', {
...
+    })
+
+    cy.findByTestId('debug-flaky-badge')
+    .contains(defaultMessages.specPage.flaky.badgeLabel)
+
+    cy.findByTestId('total-flaky-tests')
+    .contains(flakyCount)
+  })
+
+  it('checks if all run status are accounted for', () => {
```
astone123: could we update the name of this test case to say something like
> displays each run status

?

packages/app/src/debug/DebugPageHeader.cy.tsx
```diff
@@ -32,27 +34,70 @@ describe('<DebugPageHeader />', {
...
+          }
+        },
+        render: (gqlVal) => {
+          return (
+            <DebugPageHeader gql={gqlVal} commitsAhead={0}/>
+          )
+        },
+      })
+
+      cy.findByTestId(`debug-runNumber-${status}`).should('be.visible')
```
astone123, line 97: Do you think we should capture Percy snapshots here so that we have a snapshot of the header with each status? Might be overkill. Curious what others think

packages/app/src/debug/DebugPageHeader.vue
```diff
@@ -144,6 +152,16 @@ const props = defineProps<{
...
+const ICON_MAP = {
+  PASSED: { textColor: 'text-jade-400' },
+  FAILED: { textColor: 'text-red-400' },
+  CANCELLED: { textColor: 'text-gray-500' },
+  ERRORED: { textColor: 'text-orange-400' },
+  RUNNING: { textColor: 'text-indigo-500' },
+}
+
+const runNumberColor = computed(() => ICON_MAP[String(props.gql.status)]['textColor'])
+
```
astone123: Suggestion - to make this a little more typed you could do something like this

```suggestion
const ICON_MAP: { [key in CloudRunStatus]?: {textColor: string} } = {
  PASSED: { textColor: 'text-jade-400' },
  FAILED: { textColor: 'text-red-400' },
  CANCELLED: { textColor: 'text-gray-500' },
  ERRORED: { textColor: 'text-orange-400' },
  RUNNING: { textColor: 'text-indigo-500' },
}

const runNumberColor = computed(() => {
  if (props.gql.status) {
    const icon = ICON_MAP[props.gql.status]

    return icon?.textColor
  }

  return ''
})
```

The only issue is that you'd then have to cast it to `any` when passing it to `SolidStatusIcon` because it won't accept `string`.

```jsx
<SolidStatusIcon
    size="16"
    :status="debug.status.toLowerCase() as any"
/>
```

up to you

packages/app/src/debug/DebugResults.vue
```diff
@@ -9,31 +9,34 @@
       :total-skipped="results.totalSkipped"
       :total-pending="results.totalPending"
       :order="['PASSED', 'FAILED', 'SKIPPED', 'PENDING']"
+    />
+    <div
+      v-if="(results.totalFlakyTests && results.totalFlakyTests > 0)"
```
astone123: Minor: I think we can just say

```suggestion
      v-if="results?.totalFlakyTests"
```

here. If `results.totalFlakyTests` is `null`, `undefined`, or `0` then this will evaluate to `false`. If it's greater than 0 then it will be `true`. No worries if you want to be more explicit

astone123 (approved):
Looks good 👍🏻 I left some minor comments

packages/app/src/debug/DebugPageHeader.vue
lmiller1990: I tried doing

```
const ICON_MAP: Record<CloudRunStatus, { textColor: string }> = {
  PASSED: { textColor: 'text-jade-400' },
  FAILED: { textColor: 'text-red-400' },
  CANCELLED: { textColor: 'text-gray-500' },
  ERRORED: { textColor: 'text-orange-400' },
  RUNNING: { textColor: 'text-indigo-500' },
} as const
```

Error...

```
Type '{ readonly PASSED: { readonly textColor: "text-jade-400"; }; readonly FAILED: { readonly textColor: "text-red-400"; }; readonly CANCELLED: { readonly textColor: "text-gray-500"; }; readonly ERRORED: { readonly textColor: "text-orange-400"; }; readonly RUNNING: { readonly textColor: "text-indigo-500"; }; }' is missing the following properties from type 'Record<CloudRunStatus, { textColor: string; }>': NOTESTS, OVERLIMIT, TIMEDOUTts(2739)
```

Are our types telling us something? What do we do for `NOTESTS, OVERLIMIT, TIMEDOUT`? Should it be an empty string? In this case would we just add those definitions?

If we do, this becomes

```ts
import type { CloudRunStatus  } from '../generated/graphql'

const ICON_MAP: Record<CloudRunStatus, { textColor: string }> = {
  PASSED: { textColor: 'text-jade-400' },
  FAILED: { textColor: 'text-red-400' },
  CANCELLED: { textColor: 'text-gray-500' },
  ERRORED: { textColor: 'text-orange-400' },
  RUNNING: { textColor: 'text-indigo-500' },
  NOTESTS: { textColor: 'text-indigo-500' },
  OVERLIMIT: { textColor: 'text-indigo-500' },
  TIMEDOUT: { textColor: 'text-indigo-500' },
} as const

const runNumberColor = computed(() => {
  if (!props.gql.status) {
    return ''
  }
  return ICON_MAP[props.gql.status].textColor
})
```

Which seems to be nice and  type safe, and represent all states.

packages/app/package.json
```diff
@@ -21,6 +21,7 @@
   "dependencies": {},
   "devDependencies": {
     "@cypress-design/vue-icon": "^0.12.1",
+    "@cypress-design/vue-statusicon": "0.1.4",
```
lmiller1990: Not really your wheelhouse, but I wonder why they don't just publish a single package, `cypress-design/core`, with everything we need.

lmiller1990 (approved):
This is looking good. I'm definitely interested in making things as type safe as possible, but not obvious issues with the implementation.

packages/app/src/debug/DebugPageHeader.cy.tsx
amehta265: I definitely agree that this could be more descriptive

packages/app/src/debug/DebugPageHeader.cy.tsx
amehta265: My reason for adding the percy snapshots was because each `status` has a unique color and icon hence in order to visually compare all the possible combinations of `status` pairings without explicitly testing the colors through `rgb` values I added the percy snapshots.
I could see this either way (with or without the snapshots) so I'm happy to remove it if that is the consensus

packages/app/src/debug/DebugPageHeader.vue
amehta265: I believe the suggestions above are definitely the better approach here. Type safety is important and adding this would improve that!

packages/app/src/debug/DebugResults.vue
amehta265: Good catch! I believe the extra conditional is unnecessary and verbose as you've pointed out. Will make the change

packages/app/src/debug/DebugPageHeader.cy.tsx
astone123: @amehta265 I'm confused - I don't see Percy snapshots being taken in this test currently. I'm suggesting that we add them. Am I missing something?

packages/app/package.json
elevatebart: @lmiller1990 That is because of versioning.
Upgrading a package should not risk breaking the rest of the platform.
It is a way of minimizing the surface risk while still mutualizing the components.

We might create a `@cypress-design/core` at some point once it stabilizes.

packages/app/src/debug/DebugPageHeader.cy.tsx
ZachJW34: If there is a unique UI state, then we should try and capture it with a percySnapshot! (In response to your first question)

amehta265: I've added the percySnaphot

PR #25074, opened 2022-12-08
fix: run-all-specs opens in new tab rather than new browser

### LINKED ISSUE #24919: experimentalRunAllSpecs: Replace with new tab when selecting specs rather than closing browser
### Current behavior

When you select a new subset of specs to run with the "Run All Specs" feature, the browser closes and reopens. Not only is this not performant, it also exacerbates #21743 where the Chrome browser is unable to focus properly.

### Desired behavior

Similar to run-mode, the browser should not close but rather the tab should be replaced with a new tab.

ZachJW34:
- Closes #24919
### User facing changelog
Run All Specs will open in a new tab rather than close and reopen the browser

### Additional details
Following the same [run.ts](https://github.com/cypress-io/cypress/blob/294e915b9f0f84b95eb683df111773a1220283a1/packages/server/lib/modes/run.ts#L642) logic for retaining the browser and switching out the tab with a fresh one.

### Steps to test
Choose a project that has Cypress setup and add `e2e.experimentalRunAllSpecs = true` to the config. Check out this branch, run `yarn` and the `yarn cypress:open`. Open up a browser and verify that, after clicking "Run all specs", the browser isn't closed and the current tab is replaced with a new one.

Electron still spawns a new window since there is no "New Tab" in Electron.

Not sure the best way to test this change, since any tests around `launchProject` require opening and closing browsers.

https://user-images.githubusercontent.com/25158820/206529057-a489559c-9538-4887-a2e9-207176fa7dcc.mov

### PR Tasks


packages/graphql/schemas/schema.graphql
```diff
@@ -1255,7 +1255,7 @@ type Mutation {
   internal_clearProjectPreferencesCache(projectTitle: String!): Boolean
 
   """Launches project from open_project global singleton"""
-  launchOpenProject(specPath: String): CurrentProject
+  launchOpenProject(shouldLaunchNewTab: Boolean, specPath: String): CurrentProject
```
lmiller1990, line 1258: Something is strange here - the order of the arguments is the opposite on the client mutation? https://github.com/cypress-io/cypress/pull/25074/files#diff-4efefe4d3a02c250b8acd5d08fb87caf32cf883ccff71ea7965d0e6c0448fe68R25

Also, I think it makes sense for the specPath to be first arg, and shouldLaunchNewTab to be second - it follows the usual convention of putting the options as the second argument.

tooling/v8-snapshot/cache/dev-darwin/snapshot-meta.cache.json
```diff
@@ -632,6 +621,15 @@
     "./node_modules/xml2js/lib/xml2js.js",
     "./node_modules/yauzl/index.js",
     "./node_modules/zip-stream/index.js",
+    "./packages/config/node_modules/@babel/traverse/lib/context.js",
```
lmiller1990: I see these are not deterministic - at least, the _order_ isn't. I wonder if we can sort them out. Let's ask @ryanthemanuel - this might help with not cluttering up diffs in the future.

packages/graphql/schemas/schema.graphql
lmiller1990: I am surprised this works at all - GraphQL should be erroring, the types don't match.

ZachJW34: I think this is just sorted alphabetically (this is generated). Look at:

```
cloudProjectCreate(campaign: String, ciProviders: [String!], cohort: String, medium: String, name: String!, orgId: ID!, public: Boolean!, source: String): CloudProject
```

Order doesn't matter since all the args resolve to an `args` object.

lmiller1990 (commented):
I tested it out, works great. Curious about the order of the GraphQL mutation arguments, though.

tooling/v8-snapshot/cache/dev-darwin/snapshot-meta.cache.json
ryanthemanuel: These should be alphabetical? These are actually dependencies than the ones below. Ones below are hoisted up. These are isolated to the config package. I do have a PR that will help with the maintenance of these files incoming though.

packages/graphql/schemas/schema.graphql
lmiller1990: Right, I see. I didn't realize that (and it's quite confusing - the order generally DOES matter in GraphQL, I guess not if you are using Nexus).

tooling/v8-snapshot/cache/dev-darwin/snapshot-meta.cache.json
lmiller1990: Oh, I see. I wonder why these changed here, then? This PR does not touch and dependencies. Or, it is left over from a previous PR?

lmiller1990 (approved):

tooling/v8-snapshot/cache/dev-darwin/snapshot-meta.cache.json
ryanthemanuel: Probably a previous one.

packages/server/lib/makeDataContext.ts
```diff
@@ -155,6 +155,12 @@ export function makeDataContext (options: MakeDataContextOptions): DataContext {
         return devServer
       },
       isListening,
+      resetBrowserTabsForNextTest (shouldKeepTabOpen: boolean) {
+        return openProject.resetBrowserTabsForNextTest(shouldKeepTabOpen)
+      },
+      resetServer () {
+        return openProject.getProject()?.reset()
```
rockindahizzy:
```suggestion
        return openProject.getProject()?.server?.reset()
```

ZachJW34: 💯

rockindahizzy (approved):
LGTM

lmiller1990:
Oh, one last thing we should check @ZachJW34 - are videos getting recorded correctly? How does this manifest in Cypress Cloud?

Edit: actually doesn't matter for Cloud, this is open mode only - still relevant locally though, since you *can* record videos locally (although not as common/useful, worth knowing if/how this works).

astone123 (approved):
I tested this and it works well 👍🏻

ZachJW34:
@lmiller1990 you cannot record videos when in open mode. Since this feature is only related to open-mode, this should have no affect on video recording.

PR #25079, opened 2022-12-08
chore: make `ExpandableFileChooser` more generic 

lmiller1990:
A refactor to use slots make the `ExpandableFileChooser` component more generic. Also, this allievates the need for `ExpandableFileChooser` to emit a `selectItem` event - that's the job of the user passing a component (via `<template #item>`) to the file chooser.


packages/app/src/specs/generators/ExpandableFileChooser.vue
```diff
@@ -54,9 +54,9 @@
       :search="filePathSearch"
     >
       <template #expanded-content="{file}">
-        <ReactComponentList
+        <slot
```
lmiller1990, line 57: I think we might lose some type safety here, but I'm not completely sure. This does make it more composable and generic, though.

astone123 (approved):
Nice 👍🏻

PR #25094, opened 2022-12-10
fix: add column, line, and method check to integrity check

ryanthemanuel:
### User facing changelog

Integrity check will now fail if the column and line numbers are altered in the binary entry point.

### Additional details

Previously, we would only check for column and line numbers in the entry point, but we can increase the strength of the integrity check by also checking for the specific column and line numbers of the executing code.

### Steps to test

Ensure the tests pass on CI (especially the binary tests)

n/a

### PR Tasks


scripts/binary/binary-integrity-check-source.js
```diff
@@ -2,6 +2,8 @@ const OrigError = Error
 const captureStackTrace = Error.captureStackTrace
 const toString = Function.prototype.toString
 const callFn = Function.call
+const filter = Array.prototype.filter
+const startsWith = String.prototype.startsWith
```
lmiller1990, line 6: Do you know why we create all these aliases in the first place instead of just doing `tempError.filter(...`) etc? It looks like something to do with the integrity check - how does comparing `if (filter.call !== callFn) {` verify the integrity?

patches/bytenode+1.3.7.dev.patch
```diff
@@ -0,0 +1,381 @@
+diff --git a/node_modules/bytenode/lib/compileFile.js b/node_modules/bytenode/lib/compileFile.js
+new file mode 100644
+index 0000000..314eeee
+--- /dev/null
++++ b/node_modules/bytenode/lib/compileFile.js
```
lmiller1990: This looks hard to maintain - is there a bug in the library or we just needed a specific feature?

It is possible to write our own `compileElectronCode`, wrapping the `bytenode` library functionality, and put it in the monorepo (eg under `scripts`)? That seems like it'd be easier to maintain, long term, than patching the package.

chrisbreiding (approved):

scripts/binary/binary-integrity-check-source.js
```diff
@@ -143,13 +179,18 @@ function integrityCheck (options) {
         fileName: 'evalmachine.<anonymous>',
       },
       {
-        functionName: 'Module2._extensions.<computed>',
+        functionName: 'c._extensions.<computed>',
+        methodName: '.jsc',
         // eslint-disable-next-line no-undef
         fileName: [appPath, 'index.js'].join(PATH_SEP),
+        line: 1,
+        column: 2076,
```
emilyrohrbough: what are the scenarios where this column number would change? new imports into the root file?

scripts/binary/binary-integrity-check-source.js
ryanthemanuel: We are protecting against the prototype call method on each of those functions being tampered with. So we want to make sure that they are still equal to the original function call prior to using them

patches/bytenode+1.3.7.dev.patch
ryanthemanuel: The goal here is minimize the code that is in our entry point bundle in prod to prevent as much “dead code” that could be injected into as possible. Thus, I patched bytenode to split out the compilation (package time) from the loading (run time). It is definitely a little involved but I don’t think we are going to be updating that dependency very often.

scripts/binary/binary-integrity-check-source.js
ryanthemanuel: This would change if our entry point file bundle changed at all. I don’t see that happening very frequently though and if it does, we will just update these numbers. The situations where we would update the entry point bundle is if this file changes: https://github.com/cypress-io/cypress/blob/develop/scripts/binary/binary-entry-point-source.js or if bytenode gets updated.

patches/bytenode+1.3.7.dev.patch
```diff
@@ -0,0 +1,381 @@
...
++ * @param   {string}          args.filename The JavaScript source file that will be compiled
++ * @param   {boolean}         [args.compileAsModule=true] If true, the output will be a commonjs module
++ * @param   {string}          [args.output=filename.jsc] The output filename. Defaults to the same path and name of the original file, but with `.jsc` extension.
++ * @param   {boolean}         [args.electron=false] If true, compile code for Electron (which needs to be installed)
++ * @param   {boolean}         [args.createLoader=false] If true, create a loader file.
++ * @param   {boolean}         [args.loaderFilename='%.loader.js'] Filename or pattern for generated loader files. Defaults to originalFilename.loader.js. Use % as a substitute for originalFilename.
++ * @param   {string}        [output] The output filename. (Deprecated: use args.output instead)
++ * @returns {Promise<string>}        A Promise which returns the compiled filename
++ */
++ const compileFile = async function (args, output) {
```
emilyrohrbough: Looks like compileFile & compileElectronCode are 1:1 with the "removed" code in this patch. Can we re-generate to only remove the extra methods? Also why does it matter if we remove the other methods? Any other impact beyond making it a scoped import vs a global?

scripts/binary/binary-integrity-check-source.js
emilyrohrbough: will this error provide insights into that column number or do we just fail and we'd have to manually verify?

patches/bytenode+1.3.7.dev.patch
ryanthemanuel: I’ll regenerate this to make sure it’s cleaner. The main reason I want to remove as much as possible is to prevent “dead code” from being in the bundle that could be deleted and new code added in its place.

emilyrohrbough: I didn't realize that the binary-cleanup script was being bundled and shipped with the binary. Thought it was only used for building the binary content.

ryanthemanuel: It's used for bundling but then reading the content that was bundled later as well.

scripts/binary/binary-integrity-check-source.js
ryanthemanuel: What sort of insights would you expect? The error will let us know what that there was a difference and what the difference is (similar to what we were doing with file and function names previously).

emilyrohrbough: sorry - i didn't remember what we were prev doing. I was just curious if this error outputs on any download or just during the build process

ryanthemanuel: This error will output when the binary is corrupted. So if a user tries to open Cypress with a bad binary, they will see "your code executed from column x instead of y"

patches/bytenode+1.3.7.dev.patch
emilyrohrbough: Just so I understand, could you share where? Also do we need to maintain this require if it's only exporting the compileCode fn? https://github.com/cypress-io/cypress/blob/emily/delayed-ci/scripts/binary/binary-entry-point-source.js#L6

ryanthemanuel: The require actually includes this code as well, but as I'm thinking through this, since I'm patching anyway I can make this a little clearer: https://github.com/bytenode/bytenode/blob/master/lib/index.js#L252

mjhenkes (approved):

emilyrohrbough (approved):

PR #25101, opened 2022-12-11
chore: update to Vite 4 and Rollup 3

lmiller1990:
Vite 4 migration: https://vitejs.dev/guide/migration.html
Rollup 3 migration: https://rollupjs.org/guide/en/#migration

None of the breaking changes seems to impact us. We only use a very simple Rollup configuration to bundle things in `npm/*`, and we don't do anything too strange with Vite, either. Great!


packages/driver/types/cypress/log.d.ts
```diff
@@ -11,14 +11,14 @@ declare namespace Cypress {
   }
 
   type ReferenceAlias = {
```
lmiller1990, line 13: Not sure why these changed - all I did was run `yarn`. The changes are more in line with the style of the rest of the file, though, so I think it's okay.

marktnoonan:
Looks good, though I'd like to check the binary before we merge, could you run those jobs for this branch? I'd want to check live reloading in open mode to make sure dependency changes didn't introduce any issues.

ZachJW34 (approved):
Looks good!

lmiller1990:
@marktnoonan this is just for our internal dependencies, we use the user's Vite version - we don't ship Vite or Rollup in production, neither of these dependencies is present in the binary.

rockindahizzy (approved):

mike-plummer (approved):
Looks good, but just to cross the t's and dot the i's there have been a couple point releases in the last day:
`vite@4.0.1`
`rollup@3.7.3`
Worth bumping?

lmiller1990:
@marktnoonan neither rollup or vite is in the binary, no need - can verify by doing `cd /Users/<users>/Library/Caches/Cypress/11.0.1/Cypress.app/Contents/Resources/app
` and observing neither is there.

@mike-plummer we can, I'm not sure why renovate isn't doing this for us 🤔 

Doing that now.

marktnoonan (approved):

PR #25133, opened 2022-12-13
fix: .get() respects null withinSubject

### LINKED ISSUE #25104: cy.get()'s withinSubject value is ignored when used in a within() block
### Current behavior

Setting the `cy.get`'s `withinSubject` option seems to not work anymore since v12 (it was working in v11). Setting it to `null` is particularly valuable to temporarily escape scoping in a `cy.within` block.

### Desired behavior

`withinSubject` option should be taken into account

### Test code to reproduce

The following test will fail. The page available under `/app` simpli cointains:

```html
<html>
 <body></body>
</html>
```

```ts
it.only("Reproduce", () => {
  cy.visit("/app");
  cy.get("body").within(() => {
    cy.get("html", { withinSubject: null });
  });
});
```

### Cypress Version

12.0.2

### Node version

18

### Operating System

macOS

### Debug Logs

_No response_

### Other

_No response_

BlueWinds:
- Closes https://github.com/cypress-io/cypress/issues/25104

### User facing changelog
Fixes a regression introduced in Cypress 12 where `cy.get()` would ignore a `null` withinSubject.

### Additional details
We didn't have a test around this, so a regression crept in. `withinSubject` is exposed to users, and sometimes used as an 'escape hatch' - allowing them to run commands that ignore a `.within()` block they're inside.

### Steps to test
See included unit test.

Return to 11.x behavior.

### PR Tasks

### DIFF
--- a/packages/driver/cypress/e2e/commands/querying/querying.cy.js
+++ b/packages/driver/cypress/e2e/commands/querying/querying.cy.js
@@ -82,6 +82,12 @@ describe('src/cy/commands/querying', () => {
       cy.get('doesNotExist')
     })
 
+    it('respects null withinSubject', () => {
+      cy.get('#list').within(() => {
+        cy.get('#upper', { withinSubject: null })
+      })
+    })
+
     describe('custom elements', () => {
       // <foobarbazquux>custom element</foobarbazquux>
 
--- a/packages/driver/src/cy/commands/querying/querying.ts
+++ b/packages/driver/src/cy/commands/querying/querying.ts
@@ -172,7 +172,7 @@ export default (Commands, Cypress, cy, state) => {
       let $el
 
       try {
-        let scope = userOptions.withinSubject || cy.getSubjectFromChain(withinSubject)
+        let scope = userOptions.withinSubject !== undefined ? userOptions.withinSubject : cy.getSubjectFromChain(withinSubject)
 
         if (includeShadowDom) {
           const root = scope?.get(0) || cy.state('document')


AtofStryker (approved):

packages/driver/cypress/e2e/commands/querying/querying.cy.js
```diff
@@ -82,6 +82,12 @@ describe('src/cy/commands/querying', () => {
       cy.get('doesNotExist')
     })
 
+    it('respects null withinSubject', () => {
+      cy.get('#list').within(() => {
+        cy.get('#upper', { withinSubject: null })
```
emilyrohrbough, line 87: I know this is fixing a v11 -> v12 regression ,but is this a valid use-case of get in a within callback? Seems a little odd 😅 I had always assumed this was intended to narrow down, not remove scope

emilyrohrbough (approved):

packages/driver/cypress/e2e/commands/querying/querying.cy.js
BlueWinds: I can imagine a situation like this:

```
Cypress.Commands.add('loginAs', (name) => {
  cy.get('#login-form', { withinSubject: null })
    .find('.username')
    .type(name)
  // ...do rest of login...
})

cy.get('#user-profile').within(() => {
  cy.get('.username').should('eq', '')
  cy.loginAs('fred')
  cy.get('.username').should('eq', 'fred')
})
```

where the custom command wouldn't want to respect any 'parent' `.within()` blocks. Doesn't seem super common, but also not obviously incorrect.

PR #25145, opened 2022-12-14
feat: pass parser options to allow parsing of tsx files

astone123:
**NOTE**: All of our cypress-in-cypress tests will fail on this branch because of an issue with the way that Babel dependencies are being resolved in the v8 snapshot. This is currently being investigated. I'm going to merge this into the feature branch anyway once I get approvals and I see that `CodegenActions.spec.ts` has passed in CI.

### User facing changelog

n/a, merging to feature branch

### Additional details

Our current implementation of create from React component will fail to parse TSX files that contain type annotations. This PR adds some parser plugins to allow for the parsing of TSX files.

### Steps to test

1. Check out this branch
2. Revert the change to `CodegenActions.ts`
3. Run `env DEBUG="cypress:data-context" yarn workspace @packages/data-context test`
4. Notice that the `correctly parses typescript files` test case fails
5. See the debug error in the console indicating a syntax error occurred while parsing the file
6. Re-introduce the changes to `CodegenActions.ts`
7. Run the tests again and verify that they all pass and that `LoginForm.tsx` was parsed correctly

n/a

### PR Tasks


packages/data-context/test/unit/actions/project/LoginForm.tsx
```diff
@@ -0,0 +1,63 @@
+import React, { useState } from 'react'
```
ZachJW34: Could trim this component down to just what was causing the parsing error, if we add more of these edge-case tests it would be nice to not bloat the files.

ZachJW34 (approved):
Glad we have a fix!

lmiller1990 (approved):
Just one change from Zach W and we can merge, nice job 👍

packages/data-context/test/unit/actions/project/LoginForm.tsx
astone123: Yeah good point. I trimmed basically everything except the interface definition since that's what was causing the syntax error originally

PR #25146, opened 2022-12-14
fix: Display correct major version in migration wizard

### LINKED ISSUE #25138: Upgrade wizard shows wrong version
### Current behavior

See the screenshot
Screen Shot 2022-12-13 at 17 05 53

### Desired behavior

Upgrade message should be correct and use the current Cypress major version

### Test code to reproduce

Project with Cypress v8, installed v12, opened Cypress

### Cypress Version

12.1.0

### Node version

v14

### Operating System

mac

### Debug Logs

_No response_

### Other

_No response_

mike-plummer:
- Closes #25138 

### User facing changelog
* Fixes issue where incorrect Cypress version could be shown in the migration wizard
* Fixes issue where the Cypress migration wizard would fail to run in global mode on newer versions of Cypress

### Additional details
This resolves a couple hardcoded version references in messaging and also updates an edge-case compatibility check when a user is attempting to use Cypress in global mode to migrate a project with a different version of Cypress. The compatibility check only supported `^10.0.0` which would cause failures when trying to migrate v11/12 even if there wasn't really an incompatibility.

### Steps to test
1. Initialize a mock old Cypress project to migrate
```
mkdir myProject
cd myProject
echo '{}' > cypress.json
npm init
npm i -D cypress@11.2.0
```
3. Launch this branch in global mode and attempt to open the project. You should receive an error message warning of a version mismatch (global v12.1.0 is greater than local project 11.2.0)
4. Close and update mock project Cypress version `npm i -D cypress@12.1.0`
5. Re-launch this branch, you should no longer receive error message
6. Validate that Migration Wizard displays "Migrate to Cypress 12" title

**Error Condition**
_Before_
Screen Shot 2022-12-14 at 9 26 37 AM

_After_
Screen Shot 2022-12-14 at 9 26 58 AM

**Migration Wizard**
_Before_
Screen Shot 2022-12-14 at 9 25 07 AM

_After_
Screen Shot 2022-12-14 at 9 25 29 AM

### PR Tasks


packages/data-context/src/actions/MigrationActions.ts
```diff
@@ -189,8 +189,17 @@ export class MigrationActions {
...
-        return this.ctx.onError(getError('MIGRATION_MISMATCHED_CYPRESS_VERSIONS', version))
+      const currentVersion = (await this.ctx.versions.versionData()).current.version
+
+      // Validate that the project being migrated has a version of Cypress compatible with the version being executed.
+      // This handles situations where Cypress is launched in global mode to migrate a project with an older version of
+      // Cypress as a dependency which could break the project when launched directly.
+      // For example:
+      //    Local: 9.6.0     Global: 10.0.0     FAIL
+      //    Local: 10.0.1    Global: 10.0.0     PASS
+      //    Local: 12.0.0    Global: 12.0.1     FAIL
```
ZachJW34, line 200: It should be fine to have a global version greater than local as long as they are the same major version right?

ZachJW34 (approved):
One small comment, works great!

packages/data-context/src/actions/MigrationActions.ts
mike-plummer: My thought was that a point release may add new config values that an older point release would fail to parse. For example, the Svelte support added in 10.7.0 introduced a `framework` value that a 10.6.0 version of Cypress wouldn't know how to handle. Not necessarily something the migration wizard would currently introduce, but potentially could in the future. Theoretically a newer local version should handle of a config built by an older global version so long as they're the same major version

ZachJW34: More strict = more safety, I'm cool with that. Generally, if you're trying to migrate a project it should have the most up to date version of cypress available.

warrensplayer (approved):
Everything worked great for me!

PR #25152, opened 2022-12-14
feat: debug status badge

### LINKED ISSUE #24852: [IATR](M1.0) Debug sidebar status count
Update the sidebar nav to show a count of failed tests from the cloud

### New icon
* Update the icon for the Debug page to be the new "bug" icon

_Before_
Screen Shot 2022-12-06 at 12 15 19 PM
 _After_
Screen Shot 2022-12-06 at 12 15 39 PM

### Badges

Note: The presence of a run, the failed test count and run status will be available to the badge through either a store or an existing GQL query.

#### Add "New" badge
* The "New" badge is shown when there is no Cloud project connected or no runs (states: not logged in, no project, project has no runs at all) and the feature was released <= 2 months ago (hard coded is fine; e.g. current time <= release date + 2 months)
* The "New" badge should not be shown if there a relevant run has been determined and a failed count badge is being shown
Screen Shot 2022-12-06 at 12 25 18 PM

#### Add failed test count badge

* Green badge 	
	* If the run shown in/used by Debug has status passed, then show the number 0 in that run in a green badge
Screen Shot 2022-12-06 at 12 26 06 PM

* Red badge	
	* If the run shown in/used by Debug has status failed, then show the number of failed tests in that run in a red badge
	* If the run shown in/used by Debug has status has error statuses and at least 1 failure, then show the number of failed tests in that run in a red badge
Screen Shot 2022-12-06 at 12 26 16 PM

* Orange badge
	*	If the run shown in/used by Debug has status has error statuses and 0 failures, then show the number 0 in an orange badge
Screen Shot 2022-12-06 at 12 26 27 PM
 
**Collapsed side nav**

* When the side nav is collapsed, badges should still appear.
* If the count badge has a number > 10, then change the text to say `9+`
Screen Shot 2022-12-06 at 12 59 03 PM

[Figma designs](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/QkCp2F69X0qyLo7qnYQ67e/Cypress-App?node-id=21998%3A208933&t=0OHOefQTzf0GCexU-1)

ZachJW34:
DRAFT: Need to add spider icon to design system and incorporate

- Closes #24852

### User facing changelog
Adds status badge to Debug Sidebar, indicating the results of the latest run.

### Additional details
Did some refactoring of the `SidebarNavigation.vue` component, lifting the queries up to a container component to make the component easier to test.

### Steps to test
Driven mostly off of the CT tests, if you have `cypress-services` configured locally you can:

1. Check out the `feature/CYCLOUD-665-IATR` and run `yarn`
2. Run `yarn docker-compose && yarn dev`
3. Visit `localhost:3000` and create a project
4. In the Cypress monorepo, checkout this branch and run `CYPRESS_INTERNAL_ENV=development yarn cypress:open`
5. Select a project and add your newly created `projectId`
6. Tweak tests to cause a failure and run `npx cypress run --record --key <RECORD_KEY>`
7. If you've recorded different runs, you can change the run by changing `runByNumber(runNumber: 6) {` to whatever run you want to target (not dynamic yet)

Screen Shot 2022-12-14 at 12 14 31 PM
Screen Shot 2022-12-14 at 12 15 28 PM
Screen Shot 2022-12-14 at 12 16 22 PM

### PR Tasks


packages/app/src/navigation/SidebarNavigationRow.vue
```diff
@@ -56,17 +63,44 @@
...
+}
+
+const transitioning = ref(false)
+
+// Badge is either absolutely positioned or relative. Since the navbar expands with an animation,
+// the badge needs to animate as well, otherwise it pops into place before the sidebar is finished animating
+watch(() => props.isNavBarExpanded, async () => {
+  if (props.badge) {
+    transitioning.value = true
+    await promiseTimeout(125)
```
lmiller1990, line 101: Is there a variable we can share? How did you arrive at 125?

packages/app/src/navigation/SidebarNavigationContainer.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <SidebarNavigation
+    :gql="query.data.value"
+    :is-loading="query.fetching.value"
```
lmiller1990, line 4: Nice refactor :+1:

tooling/v8-snapshot/cache/dev-darwin/snapshot-meta.cache.json
```diff
@@ -44,6 +40,10 @@
     "./node_modules/tcp-port-used/node_modules/debug/src/node.js",
     "./node_modules/trash/node_modules/make-dir/index.js",
     "./node_modules/utif/UTIF.js",
```
lmiller1990: Not sure we need to commit these now we've got the weekly task to update them automatically.

packages/app/src/navigation/SidebarNavigation.cy.tsx
```diff
@@ -86,4 +116,74 @@ describe('SidebarNavigation', () => {
...
+
+    it('renders error badge when no tests and abnormal status', () => {
+      for (const status of ['CANCELLED', 'ERRORED', 'OVERLIMIT', 'TIMEDOUT'] as CloudRunStatus[]) {
+        mountComponent({ cloudProject: { status, numFailedTests: 0 } })
+        cy.findByLabelText('Last recorded run had an error').should('be.visible').contains('0')
+      }
+
+      cy.percySnapshot('Debug Badge:errored')
+    })
+
```
lmiller1990, line 185: Now we've got Vite 4, we should definitely get Vitest in - testing this complex logic via components makes reviewing a bit challenging. It looks like you've covered all the cases, though, which is great.

lmiller1990 (approved):
This mostly looks good to me, I don't see any obvious issues. The new UI designs sure do look good.

packages/app/src/navigation/SidebarNavigation.cy.tsx
```diff
@@ -86,4 +116,74 @@ describe('SidebarNavigation', () => {
...
+        mountComponent({ cloudProject: { status, numFailedTests: 0 } })
+        cy.findByLabelText('New Debug feature').should('be.visible').contains('New')
+      }
+    })
+
+    it('renders no badge if no cloudProject and released > 2 months ago', () => {
+      // Set to February 15, 2023 to see this fail
+      cy.clock(new Date('February 20, 2023'))
+      mountComponent()
+      cy.findByLabelText('New Debug feature').should('not.exist')
```
warrensplayer, line 151: Use messages imported from`defaultMessages` instead of hardcoded strings in these tests.  I am proposing a change to these strings in the locale file and this would keep future changes in sync

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -237,6 +237,13 @@
     },
     "nav": {
       "ariaLabel": "Pages"
+    },
+    "debug": {
+      "new": "New",
+      "debugFeature": "New Debug feature",
+      "passed": "Last recorded run passed",
+      "failed": "Last recorded run had {n} test failure | Last recorded run had {n} test failures",
+      "errored": "Last recorded run had an error"
```
warrensplayer:
```suggestion
      "passed": "Relevant run passed",
      "failed": "Relevant run had {n} test failure | Relevant run had {n} test failures",
      "errored": "Relevant run had an error"
```
Nice addition to include these labels for the badge.  I do believe we have settled on on official term to call these runs, but "relevant" has been used in general conversation about the issue.  This term is meant to convey that the run is "relevant" to the current state of the code the developer currently has checked out from git.

packages/app/src/navigation/SidebarNavigationRow.vue
ZachJW34: The animations are driven off of tailwind classes like `duration-300`. 125 felt right, I didn't want it to disappear for too long, just enough for it not to thrash

tooling/v8-snapshot/cache/dev-darwin/snapshot-meta.cache.json
ZachJW34: I'll leave them in for now, but in the future I'll start reverting these.

packages/app/src/navigation/SidebarNavigation.cy.tsx
ZachJW34: We might have to steal some of the gql-testing infra if we implement vitest as this logic is heavily tied to gql

astone123:
Not sure if this is in scope but I'm seeing some weird stuff when I load the app (when I'm logged into production cloud). The debug status badge flickers to different statuses for a few seconds and then eventually lands on one. I'm also seeing the banner that says I don't have a connection to the cloud... I think I've been seeing this recently and I'm not sure if it's related at all. Not sure what's going on there

https://www.loom.com/share/cef6b0700f4f45a39ef503ce5d8f764c

packages/app/src/navigation/SidebarNavigationRow.vue
```diff
@@ -48,6 +47,14 @@
       >
         {{ name }}
       </span>
+      <span
```
warrensplayer, line 50: What about pulling this out into its own component?  Any reason that it could not stand on its own?

packages/app/src/navigation/SidebarNavigation.vue
```diff
@@ -138,16 +152,54 @@ mutation SideBarNavigation_SetPreferences ($value: String!) {
...
+    return undefined
+  }
+
+  const showNewBadge = isAllowedFeature('debugNewBadge', loginConnectStore)
+
+  if (props.gql?.currentProject?.cloudProject?.__typename === 'CloudProject') {
+    const { status, testsForReview } = props.gql.currentProject.cloudProject.runByNumber || {}
+
+    if (status === 'NOTESTS' || status === 'RUNNING') {
+      return showNewBadge ? { value: t('sidebar.debug.new'), status: 'success', label: t('sidebar.debug.debugFeature') } : undefined
```
warrensplayer: The object for the "new" badge is duplicated on line 192.  Pull this out into a variable?

packages/frontend-shared/src/utils/isAllowedFeature.ts
```diff
@@ -47,6 +51,7 @@ export const isAllowedFeature = (
     navCiPromptAutoOpened: promptsShown.ci1,
     loginModalRecordPromptShown: promptsShown.loginModalRecord,
     latestSmartBannerShown: latestBannerShownTime,
+    debugPageRelease: new Date('2022-12-20T00:00:00').getTime(),
```
warrensplayer: What about exposing this value so it can be used in the tests instead of hardcoding in two places?

packages/app/src/navigation/SidebarNavigation.cy.tsx
```diff
@@ -86,4 +116,74 @@ describe('SidebarNavigation', () => {
...
+
+      for (const status of ['NOTESTS', 'RUNNING'] as CloudRunStatus[]) {
+        mountComponent({ cloudProject: { status, numFailedTests: 0 } })
+        cy.findByLabelText('New Debug feature').should('be.visible').contains('New')
+      }
+    })
+
+    it('renders no badge if no cloudProject and released > 2 months ago', () => {
+      // Set to February 15, 2023 to see this fail
+      cy.clock(new Date('February 20, 2023'))
```
warrensplayer: Use date addition to calculate this date instead of hardcoding a date to show the relationship to the "release" date.

warrensplayer (approved):
Looks great!  Most of my comments are just questions.  My suggested change for the messages in `en-US.json` are the only mandatory changes and updating the tests to go along with that.  Going to go ahead and approve expecting that part to be done before merging to the feature branch.

astone123:
yeah, seems like whenever I'm on this feature branch I see that banner saying that there isn't a collection to cloud. Is that expected at this point?

packages/app/src/navigation/SidebarNavigation.cy.tsx
ZachJW34: I'm not the biggest fan of this, though this has been discussed in the past. For example, how do I assert that the label is `Last recorded run had 1 test failure`? Do I write `defaultMessages.sidebar.debug.failed.split(' | ')[0].replace('{n}', 1)`. Tests should be explicit, not have the misdirection of referencing some variable and then having to add logic to make sure it works. I know this is an edge case but the only pro for referencing `defaultMessages` is that I don't have to update the test when I change the text which is not worth it IMO.

ZachJW34:
@astone123 I'll look into this a bit, I've seen it as well but I've never been able to reproduce it when I'm connected to my locally running dashboard. Running `yarn cypress:open` puts you in a weird spot where the app is trying to contact the cloud but the contracts aren't the same

astone123:
@ZachJW34 yeah that's what it seems like. Probably nothing to worry about then 👍🏻

tooling/v8-snapshot/cache/dev-darwin/snapshot-meta.cache.json
ZachJW34: Reverted due to merge conflicts

packages/frontend-shared/src/locales/en-US.json
ZachJW34: Addressed in [4c21f5f](https://github.com/cypress-io/cypress/pull/25152/commits/4c21f5f4d005a19a9dbe52ffac3101f9bfeb513f)

packages/app/src/navigation/SidebarNavigationRow.vue
ZachJW34: I tried to do this but I couldn't get the animation to work properly due to the component getting destroyed. It's not a lot of logic and makes sense in the context of the row so I'm ok with how it is.

packages/app/src/navigation/SidebarNavigation.vue
ZachJW34: Addressed in [4c21f5f](https://github.com/cypress-io/cypress/pull/25152/commits/4c21f5f4d005a19a9dbe52ffac3101f9bfeb513f)

packages/frontend-shared/src/utils/isAllowedFeature.ts
ZachJW34: Addressed in [4c21f5f](https://github.com/cypress-io/cypress/pull/25152/commits/4c21f5f4d005a19a9dbe52ffac3101f9bfeb513f)

packages/app/src/navigation/SidebarNavigation.cy.tsx
ZachJW34: Addressed in [4c21f5f](https://github.com/cypress-io/cypress/pull/25152/commits/4c21f5f4d005a19a9dbe52ffac3101f9bfeb513f)

packages/app/src/navigation/SidebarNavigation.cy.tsx
lmiller1990: I agree, I find it really difficult to read and search for specs that import the i18n file. If anything, doesn't not importing it give us more coverage? If we change the text in the i18n file, we will intentionally need to update the relevant tests, etc.

packages/app/src/navigation/SidebarNavigation.cy.tsx
lmiller1990: I think we could probably decouple the logic from the UI - and just keep the component tests to make sure it's all wired up correctly. Anyway, another battle for another day.

PR #25164, opened 2022-12-14
fix: ensure cypress in cypress uses prod dependencies from the snapshot

ryanthemanuel:
### User facing changelog

n/a

### Additional details

Due to the way v8 snapshots work, we need to ensure that the dependencies are loaded via an entry point. Since the e2ePluginSetup code is sort of piecemeal the initial imports/requires are not loaded from the snapshot. By having every prod dependency go through an entry point, we will ensure that everything is loaded consistently from the snapshot. We only should do this with prod dependencies.

### Steps to test

n/a

n/a

### PR Tasks


lmiller1990:
This didn't cause any problems until @astone123 added some new deps in a feature branch. Why did adding those dependencies cause this to suddenly become an issue?

Also, any tips to debug this? How did you 

1) figure this out
2) land on this specific fix

tooling/v8-snapshot/cache/darwin/snapshot-meta.json
```diff
@@ -871,7 +871,7 @@
     "./packages/server/lib/util/fs.ts",
     "./packages/server/lib/util/glob.js",
     "./packages/server/lib/util/proxy.ts",
-    "./packages/server/lib/util/resourceTypeAndCredentialManager.ts",
+    "./packages/server/lib/util/requestedWithAndCredentialManager.ts",
```
lmiller1990, line 874: How did this get added? Was it automatic? If not, how did you decide to add it?

lmiller1990 (approved):
Thanks for this - questions about how you debug this!

astone123 (approved):

ryanthemanuel:
> This didn't cause any problems until @astone123 added some new deps in a feature branch. Why did adding those dependencies cause this to suddenly become an issue?
> 
> Also, any tips to debug this? How did you
> 
> 1. figure this out
> 2. land on this specific fix

I think it was ultimately a timing problem. The problem was still occurring but we were never noticing it. This problem manifested on @astone123's branch after I fixed the babel dependency problem.

For the babel dependency problem:

* It didn't have a chance to manifest itself until after we added react-docgen which I believe was the first point where a fair amount of babel dependencies were being added to the snapshot. Previously, babel was only used in the child process (which doesn't use the snapshot and shouldn't in prod)
* I think this actually is an example of cypress in cypress and snapshots exposing problems. The ultimate root cause was that we were deferring some code that properly populates the [`TYPES`](https://github.com/babel/babel/blob/v7.17.7/packages/babel-traverse/src/path/index.ts#L253) object in babel. Thus, the TYPES object was not properly populated.
* By norewriting [like we did](https://github.com/cypress-io/cypress/commit/f054bdcb14a0cc3c368a7debdbb05f9b86529d5e) this allows for this code to be loaded in time.
* To troubleshoot this problem, based on the error I was seeing I reasoned it had to be with some of the babel dependencies. So I tried moving all of them to norewrite and it fixed the problem, and then I whittled it down to where the problem was and reasoned about what was going on.

For the problem fixed in this PR:

* This one was trickier. This problem was manifesting itself as srv being undefined in [this](https://github.com/cypress-io/cypress/blob/4726ea90a9acdb6dec75616be29f5b761f17ecb8/packages/graphql/src/makeGraphQLServer.ts#L94) line of code. 
* I added some debug logging and determined that `makeGraphQLServer` was being imported twice which clued me into it wasn't actually properly being loaded in the snapshot (and then being loaded again from another dependency within the snapshot).

mjhenkes (approved):
Better comment plz

tooling/v8-snapshot/cache/darwin/snapshot-meta.json
ryanthemanuel: This would normally be generated by my auto generator script so normally we shouldn't see these on individual pull requests, but I explicitly made this change locally just to make sure things were working in "prod mode" on my machine.

mschile (approved):

lmiller1990:
Thanks for the explanation @ryanthemanuel :+1:

PR #25185, opened 2022-12-15
fix: reduce Percy flake

astone123:
### User facing changelog

n/a

### Additional details

There are some animations in the UI that make visual snapshots change when no related code has actually changed, the snapshot is just taken at a slightly different time so things look different. This PR fixes Percy flake in the following two cases

- The migration screen sometimes gets snapshotted when it hasn't finished loading
  - I fixed this by asserting that the loading spinner is not in the DOM before we take a snapshot
- The reporter snapshots sometimes show up as changed because of a difference in where the command progress bar is
  - I fixed this by taking the animation class off of the command "running" icon and hiding the command progress bar completely when taking snapshots of the reporter

This will hopefully reduce noise in Percy diffs going forward and make it more useful.

### Steps to test

Check out the Percy run down in the PR checks list. Once these snapshots update, they shouldn't show up in the Percy diff anymore unless something has legitimately changed.

### PR Tasks

### DIFF
--- a/packages/app/cypress/e2e/runner/support/snapshot-reporter.ts
+++ b/packages/app/cypress/e2e/runner/support/snapshot-reporter.ts
@@ -13,6 +13,14 @@ export const snapshotReporter = () => {
       '[data-cy=reporter-panel]': ($el) => {
         $el.attr('style', 'width: 450px !important')
       },
+      '[data-cy=reporter-running-icon]': ($el) => {
+        // remove 'fa-spin' class so that the icon is not animated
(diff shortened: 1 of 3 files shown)


chrisbreiding (approved):

lmiller1990:
image

:(

Still an improvement, though, so thanks for this.

lmiller1990 (approved):

astone123:
@lmiller1990 those should only fail on this PR. They just needed to be updated this one time so they don't flake anymore so this is expected

PR #25189, opened 2022-12-15
fix: electron video performance

### LINKED ISSUE #23830: Cypress 10.8.0: Much slower running in CI and randomly failing tests
### Current behavior

A successful test run with version 10.7.0: ~ 5 min
A successful test run with version 10.8.0: ~ 12 min

Also, with 10.7.0 we had our tests very stable failing almost never.

With Version 10.8.0 the tests started to run into several random timeouts, but this could be related to the very slow execution times.

Browser: Electron 102 (headless)

### Desired behavior

I would expected that Cypress would become faster with the recent update, because of the memory footprint optimizations.

### Test code to reproduce

// code should be not of relevance

### Cypress Version

10.8.0

### Node version

v16.14.0

### Operating System

Linux Ubuntu Latest GitHub Workflow

### Debug Logs

_No response_

### Other

Possibly related: https://github.com/cypress-io/cypress/issues/23824

mjhenkes:
- Closes #23830 

### User facing changelog

Fixes a regression in 10.8.0 where the CYPRESS_EVERY_NTH_FRAME env was not being honored, nor was it being set at all causing all frames to be captured which slowed down tests. Fixed in [#25189](https://github.com/cypress-io/cypress/pull/25189)

### Steps to test

To test this, i will build the binary and pull it into a realworld app branch where the electron windows github action was failing.

With the fix: https://github.com/cypress-io/cypress-realworld-app/actions/runs/3708192998/jobs/6285651500

Without the fix: https://github.com/cypress-io/cypress-realworld-app/actions/runs/3687534552/jobs/6241596846

### PR Tasks

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -28,7 +28,7 @@ mainBuildFilters: &mainBuildFilters
       only:
         - develop
         - /^release\/\d+\.\d+\.\d+$/
-        - 'ryanm/fix/column-line'
+        - 'matth/fix/electron-video-performance'
 
 # usually we don't build Mac app - it takes a long time
 # but sometimes we want to really confirm we are doing the right thing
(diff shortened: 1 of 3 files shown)


BlueWinds (approved):

emilyrohrbough (approved):
Didn't a contributor fix this same issue a few months ago? 🤔

PR #25120, opened 2022-12-13
fix: proxy CT assets to dev server

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/24272

### User facing changelog

WIP

### PR Tasks

- [ ] Have tests been added/updated?
- [ ] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [ ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [ ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?

PR #25191, opened 2022-12-16
chore: correctly resolve url sass bundle in Angular CT

### LINKED ISSUE #24272: Error with assets loading from scss in Angular
### Current behavior

Loading assets from scss with something like this :
```html
<div class="testClass"></div>
```
```scss
.testClass {
  background-image: url("/assets/test.png");
  background-size: contain;
  width: 100px;
  height: 100px;
}
```
Image is showing using `ng serve` but not when using cypress component testing

If I remove the '/' in the url, image load in component testing but angular won't compile with the following error :
```
./src/app/app.component.scss?ngResource - Error: Module Error (from ./node_modules/postcss-loader/dist/cjs.js):
[...]/theTester/src/app/app.component.scss:2:2: Can't resolve './assets/test.png' in '[...]/theTester/src/app'
```

Another loading problem comes from loading assets in scss "main" files, I have a `main.scss` containing :
```scss
@import "components/icon";
```
For simplicity purposes I didn't use abstracts variables and mixins.

`components/icons` :
```scss
$imagesTest: test "/assets/test.png", testSecond "/assets/test.png";

.icon {
  top: 0px;
  position: relative;

  &.yolo {
    background-image: url("/assets/test.png");
    background-size: contain;
    background-repeat: no-repeat;
  }

  @each $class, $url in $imagesTest {
    &.#{$class} {
      background-image: url($url);
      background-size: contain;
      background-repeat: no-repeat;
    }
  }
}
```
`app.component.html`:
```html
  ICON TEST:
  <div
    class="icon test"
    style="display: block; width: 50px; height: 50px"
  ></div>
  YOLO:
  <div
    class="icon yolo"
    style="display: block; width: 50px; height: 50px"
  ></div>
```

If I remove the `/`  here, images won't load from assets since it's using relative path and it will try to load `assets/components/test.png` and `ng serve` won't compile with the same error above.

### Desired behavior

I would like to be able to load all my assets, I've given a sample app reproducing the issue below, but in my real-world app, I can't load any of my svg that comes from scss and all the fonts I use aren't loaded since I'm using `@font-face` with urls in scss and they can't be resolved.

### Test code to reproduce

Repo with code: https://github.com/theoarmengou/angular-cypress-bug-assets

When doing `ng serve`, you should see 4 images on `localhost:4200` and when going to `localhost:4200/dev`, there should be 8 images.

When using `yarn cypress:open` in component testing, not all images are showing for `app.component` and `dev.component`.

### Cypress Version

10.10.0

### Node version

16.16.0

### Operating System

macOS 12.5.1

### Debug Logs

_No response_

### Other

_No response_

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/24272

### User facing changelog

Correctly resolve `url` in `scss` (eg `background-image: url(....)`).

### Additional details

Angular uses webpack internally. It maps urls in sass via `resolve-url-loader`. We need to provide the correct root url - Angular assumes `/` by default, but we serve the assets on the dev server on `__cypress/src`.

Known edge case: `url` in `.css` files doesn't map properly. `.scss` and `.sass` do, though. We could hack it to force `.css` use `sass-loader` for css, but that's not really ideal - it's not what their app would be using in prod. In general, to work around it for now, we can just recommend using `.scss` - it's in Angular out of the box, and it's a superset of CSS .

This is also an issue for other frameworks (like React) using `url` in `scss`. Basically, anything using `url` in `css` and `scss` may exhibit this bug. The problem is more fundamental - I think we are not routing some asset requests properly for CT. I tried tweaking the routing in https://github.com/cypress-io/cypress/pull/25120 but it's too involved right now, and a bit more risky to touch routing, since that impacts the entire of Cypress. This routing issue (at least, I think it's routing) bleeds into other things, too, like this bug: https://github.com/mswjs/msw/issues/744.

I will spend more time thinking about this in the future, but for now I'd like to get Angular working out of the box. Showing images is a pretty basic thing that should work. This PR is the most simple and low risk way to do so.

### Steps to test

Grab the repro in the issue from the original user and verify the problem where the images don't show no longer is a problem.

`url` for images correctly resolves.

### PR Tasks


rockindahizzy (approved):
Nice!

ZachJW34 (approved):
💯

PR #23227, opened 2022-08-09
fix(proxy/prerequests): fix duplicate key behavior, fallthrough

flotwig:
- Closes 

### User facing changelog

- Fixed an issue introduced in 10.3.0 where network logs could be missing certain information or cause certain requests to be delayed by 500ms.

### Additional details

- Performance fixes in https://github.com/cypress-io/cypress/pull/22462 caused a couple of  unintended changes:
	 - Missing return statement on line 77 here causes a matched prerequest to be queued anyways: https://github.com/cypress-io/cypress/blob/8fb785a4664acf74ab9eacc243c09763e2730bf7/packages/proxy/lib/http/util/prerequests.ts#L69-L83
	 - More impactful, the `Record<string, ...>` data structure would clobber any duplicate `url+method` requests/pre-requests, which could manifest itself in incorrect pre-request data or unintended delays of requests by 500ms.
	 	- Replaced this with a `StackMap` which supports duplicate keys and `.pop` by key.

### PR Tasks

- [x] Have tests been added/updated?
- [x] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [na] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [na] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?

PR #25209, opened 2022-12-19
fix: change network request sweep interval from 1s -> 10s

lmiller1990:
- Related https://github.com/cypress-io/cypress/issues/22868 (contains multiple issues, addresses some)

### User facing changelog

Addresses issue where projects loading a large number of JS modules load very slowly. It might also impact other types of files, I am not sure.

### Additional details

A lot of context written here: https://github.com/cypress-io/cypress/issues/22868

In https://github.com/cypress-io/cypress/pull/22462 a refactor was introduced in the part of the code base where we handle pre-requests. It also introduced some code that would periodically clear out requests that were never matched up (matching up refers to matching a request that is intercepted by our browser extension and the same request that comes into our HTTP proxy). We shipped this in Cypress 10.3. 

The initial PR had a bug that caused some requests would be delayed for 500ms. That was fixed in https://github.com/cypress-io/cypress/pull/23227, which was in Cypress 10.5.

I believe there was *another* bug which flew under the radar. The sweep logic was supposed to be run every 10 seconds, according to https://github.com/cypress-io/cypress/pull/22462, but if you look at the code, it's a `setInterval` triggered every 1 second (500ms * 2). 

I changed this to be every 10 seconds, and it fixed the regression. I included a simple app you can use as a minimal example of the regression: https://github.com/lmiller1990/lots-of-modules (recordings below).

My understanding is the 1s sweep was too fast - the JS module request took a while to come in (more than 1s). Having said that, I don't understand why that causes a bottleneck - but you can see the weird behavior in the first recording below, where the JS modules are processed in "chunks". **I don't understand why this is happening**.

## Recording 1: Sweep Interval 1s. JS modules loading slowly, tests time out.

Notice there is a weird "loading in chunks" behavior when you look at the terminal. 

https://user-images.githubusercontent.com/19196536/208371600-5450ad2b-4f05-478f-b6ba-1fac2d96e579.mp4

## Recording 2: Sweep Interval 10s, test does not time out.

Notice there is no "loading in chunks" behavior.

https://user-images.githubusercontent.com/19196536/208377880-7ba0fa55-00bb-4043-879a-31ca55bd7c23.mp4

## Recording 3: Cypress 9

The test runs in 9s - as opposed to the 11s Cypress 11 shows. This is pretty consistent. I am not sure if this is in the network layer, server, or elsewhere, or a bit of both. It's slower, though, which isn't really ideal.

https://user-images.githubusercontent.com/19196536/208371994-a0be6fac-5487-4d9f-8143-99178c5f0df7.mp4

### Steps to test

Not sure on the best way to test this. I can look into it once we decide if we are happy with this fix, though. If you want to experience the difference, you could grab this repo https://github.com/lmiller1990/lots-of-modules and see the instructions in the README to start it up. Then, you can try playing with the sweep interval (just grab this branch and change variable).

Loads apps that import lots of JS modules more quickly.

### PR Tasks


BlueWinds (approved):

BlueWinds:
I'm very surprised that the sweep is expensive - I suppose that makes sense if there are hundreds or thousands of requests pending. I was expecting the list iterated over to have 1-5 elements.

I originally had a sweep interval of 10s, arbitrarily - I believe it was changed to 2x the timeout after there were some review concerns about having a fixed number? It was a while ago and I do not recall exactly.

packages/proxy/lib/http/util/prerequests.ts
```diff
@@ -80,7 +80,8 @@ export class PreRequests {
   requestTimeout: number
   pendingPreRequests = new StackMap<PendingPreRequest>()
   pendingRequests = new StackMap<PendingRequest>()
-  sweepInterval: ReturnType<typeof setInterval>
+  sweepIntervalId: NodeJS.Timeout
```
flotwig: NodeJS.Timeout is an object, not an ID

```suggestion
  sweepIntervalTimer: NodeJS.Timeout
```

lmiller1990: Woah, did not know this. Fixed.

lmiller1990: https://github.com/cypress-io/cypress/pull/25209/commits/ba9339f6fd8624d396897c4ab2bba37c093c7639

flotwig (approved):

lmiller1990:
> I'm very surprised that the sweep is expensive - I suppose that makes sense if there are hundreds or thousands of requests pending. I was expecting the list iterated over to have 1-5 elements. 
> I originally had a sweep interval of 10s, arbitrarily - I believe it was changed to 2x the timeout after there were some review concerns about having a fixed number? It was a while ago and I do not recall exactly.

Me too... I think there's more at work here, but this is a good enough fix for now. We need to do a deeper dive and audit of the network layer and see if we've missed something.

adam-thomas-privitar:
> I suppose that makes sense if there are hundreds or thousands of requests pending

Just as a use case, If you try and run E2E tests locally against an application in dev mode that uses an unbundled dev tool like Vite, you have thousands of requests for static files. It takes about 20 seconds to load the app instead of < 0.5 seconds without cypress in play.

lmiller1990:
@adam-thomas-privitar can you try the pre-release with this fix and see if it helps? https://github.com/cypress-io/cypress/commit/4a9bc6f80f3020f048253d36828467db750fadc8#comments

adam-thomas-privitar:
Yep! The timing is crazy since I just started looking into this issue today. Thanks a lot for the hard work on this. Give me an hour or so I'll give it a go.

adam-thomas-privitar:
So my case is extreme -- theres 3.3K requests. 

~~It's still problematic, I think if there's an improvement its too small on the scale of things to be picked up in this particular use case.~~

~~Before https://github.com/cypress-io/cypress/pull/23227, all of these requests were over 500ms. After it, they sort of float around 250ms-600ms with very high variance. Overall though it made a massive difference the overall app load time.~~

~~That range though seems to be the case still with this change. I observe later requests are slower than earlier ones, all coming in over 600ms.~~

~~The thing where it seemed to load in batches at a time isn't there, but I'm not sure if that was already gone after above fix for me.~~

Never mind! It works great. I stupidly forgot to close cypress. Everything is actually mostly around 100ms to 400ms. Nice one!" I need to do some more testing to verify this is consistent. Its still quite a high level of variance though. I guess this is fundamental to the approach.

Personally would love to be able to opt out of all this altogether for certain URLs. My URLs for these req's are static. I don't care about any cypress processing.

Even if experimental or undocumented, a flag to allow me to set the `requestTimeout` to be 0 would be incredible for me. I appreciate it does for others, but for me it all adds no value as I don't care about it -- so the slow down is a source of great frustration.

lmiller1990:
> Never mind! It works great.

Great to here - this will be in the next release (early Jan, 2023).

PR #25217, opened 2022-12-19
chore: In cli/package.json move types to devDependencies

skvale:
### User facing changelog

N/A

### Additional details

Maybe there's a reason for these types to be dependencies?

### Steps to test

Is some functionality dependent on a types dependency?

### PR Tasks


chrisbreiding:
I believe those types are included because we bundle dependencies and have them as part of the Cypress API. Here's some context on [the sizzle types](https://github.com/cypress-io/cypress/pull/5247/files). I would guess the node one is included because some [node polyfills are included](https://github.com/cypress-io/cypress/blob/0472bb9cdbd4bb7604572baeed0873a08a81b6a9/npm/webpack-batteries-included-preprocessor/index.js#L135-L148) when we bundle the test code. 

That raises the question as to why these ones are included and not others. Why`@types/sinonjs__fake-timers` and not `@types/sinonjs` itself? I don't really have an answer for that. That would require some more investigation.

skvale:
Thanks for looking into it, I'll close this PR

PR #25234, opened 2022-12-20
feat: run failures enhancements

### LINKED ISSUE #24854: [IATR](M1.0) "Run Failures" button enhancements
Updates to the behavior of the "Run Failures" button in the `DebugSpec` component

* Add a tooltip when the button is disabled that says "Spec was not found locally"
image
* Update disabled/enabled based on local polling
	* the existing logic that is being used to monitor changes in spec files will be used to drive the setting of the `isDisabled` prop for the component.  Confirm with a test that changing the prop on an already mounted component will update the button/link
	
* Show CTA in tooltip for switching testing types if needed
	* The `isDisabled` flag will also be false for a spec if the testing type for the spec does not match the current testing type for the Cypress app.  Add a new flag to the component to let it know it needs to switch to a new testing type in order to run this test locally.  Then show a different tooltip shown in the designs below.
	* If the App is in E2E mode and the test is a component test, and the local project is configured for component testing
		* Text: "There are {failedSpecCount} component tests failing in this spec. To run them locally, switch to component testing"
		* CTA: "Switch to component testing" - clicking the button should reload the app in the component testing mode
	* If the opposite situation:
		* Text: "There are {failedSpecCount} E2E tests failing in this spec. To run them locally, switch to E2E"
		* CTA: "Switch to E2E testing" - clicking the button should reload the app in the E2E  testing mode
image

[Figma Flow](https://www.figma.com/proto/PAOozuCxtfetqP8fUVvg0y/branch/QkCp2F69X0qyLo7qnYQ67e/Cypress-App?node-id=21467%3A202325&scaling=min-zoom&page-id=21010%3A176355&starting-point-node-id=21467%3A202325&show-proto-sidebar=1)

ZachJW34:
- Closes #24854
### User facing changelog
na

### Additional details
Working with Ankit revealed we needed to change up `DebugMappings.ts`. We paired on this and the results of the refactor are relevant for his upcoming work (groups).

### Steps to test
Driven mostly from CT tests, if you want to run it e2e:

1. Check out the `feature/CYCLOUD-665-IATR` and run `yarn`
2. Run `yarn docker-compose && yarn dev`
3. Visit `localhost:3000` and create a project
4. In the Cypress monorepo, checkout this branch and run `CYPRESS_INTERNAL_ENV=development yarn cypress:open`
5. Select a project and add your newly created `projectId`
6. Tweak tests to cause a failure and run `npx cypress run --record --key <RECORD_KEY>`
7. If you've recorded different runs, you can change the run by changing `runByNumber(runNumber: 6) {` to whatever run you want to target (not dynamic yet)

https://user-images.githubusercontent.com/25158820/208729758-53a8565f-cdbd-40ea-af6f-dba66ce50256.mov

### PR Tasks

### DIFF
--- a/packages/app/src/debug/DebugContainer.cy.tsx
+++ b/packages/app/src/debug/DebugContainer.cy.tsx
@@ -6,10 +6,6 @@ import { specsList } from './utils/DebugMapping'
 import { CloudRunStubs } from '@packages/graphql/test/stubCloudTypes'
 
 describe('<DebugContainer />', () => {
-  before(() => {
-    (window as any).CYPRESS_RUN_TESTING_TYPE = 'e2e'
-  })
-
   context('empty states', () => {
     const validateEmptyState = (expectedMessage: string) => {
       cy.mountFragment(DebugSpecsFragmentDoc, {
--- a/packages/app/src/debug/DebugContainer.cy.tsx
+++ b/packages/app/src/debug/DebugContainer.cy.tsx
@@ -92,14 +92,14 @@ describe('<DebugContainer />', () => {
         { specId: 'a1c', id: 'random2' },
       ] as DebugSpecListTestsFragment[]
       const groups = [
-        { id: 'a' },
-        { id: 'b' },
+        { id: 'a', testingType: 'e2e' },
+        { id: 'b', testingType: 'e2e' },
       ] as DebugSpecListGroupsFragment[]
 
       const debugMappingArray = specsList({ specs, tests, groups, localSpecs: [], currentTestingType: 'e2e' })
 
       expect(debugMappingArray).to.have.length(1)
-      expect(debugMappingArray[0]).to.deep.equal({ spec: { id: 'a1c', groupIds: ['a'] }, tests: [{ specId: 'a1c', id: 'random1' }, { specId: 'a1c', id: 'random2' }], groups: [{ id: 'a' }], foundLocally: false, testingType: 'e2e', matchesCurrentTestingType: true })
+      expect(debugMappingArray[0]).to.deep.equal({ spec: { id: 'a1c', groupIds: ['a'] }, tests: [{ specId: 'a1c', id: 'random1' }, { specId: 'a1c', id: 'random2' }], groups: [{ id: 'a', testingType: 'e2e' }], foundLocally: false, testingType: 'e2e', matchesCurrentTestingType: true })
     })
 
     it('maps correctly for multiple specs and test', () => {
@@ -116,16 +116,16 @@ describe('<DebugContainer />', () => {
         { specId: '123', id: 'random6' },
       ] as DebugSpecListTestsFragment[]
       const groups = [
-        { id: 'a' },
-        { id: 'b' },
+        { id: 'a', testingType: 'e2e' },
+        { id: 'b', testingType: 'e2e' },
       ] as DebugSpecListGroupsFragment[]
 
       const debugMappingArray = specsList({ specs, tests, localSpecs: [], currentTestingType: 'e2e', groups })
 
       const expected = [
-        { spec: { id: '123', groupIds: ['a'] }, tests: [{ specId: '123', id: 'random1' }, { specId: '123', id: 'random6' }], groups: [{ id: 'a' }], foundLocally: false, testingType: 'e2e', matchesCurrentTestingType: true },
-        { spec: { id: '456', groupIds: ['b'] }, tests: [{ specId: '456', id: 'random2' }, { specId: '456', id: 'random3' }], groups: [{ id: 'b' }], foundLocally: false, testingType: 'e2e', matchesCurrentTestingType: true },
-        { spec: { id: '789', groupIds: ['a', 'b'] }, tests: [{ specId: '789', id: 'random4' }], groups: [{ id: 'a' }, { id: 'b' }], foundLocally: false, testingType: 'e2e', matchesCurrentTestingType: true },
+        { spec: { id: '123', groupIds: ['a'] }, tests: [{ specId: '123', id: 'random1' }, { specId: '123', id: 'random6' }], groups: [{ id: 'a', testingType: 'e2e' }], foundLocally: false, testingType: 'e2e', matchesCurrentTestingType: true },
+        { spec: { id: '456', groupIds: ['b'] }, tests: [{ specId: '456', id: 'random2' }, { specId: '456', id: 'random3' }], groups: [{ id: 'b', testingType: 'e2e' }], foundLocally: false, testingType: 'e2e', matchesCurrentTestingType: true },
+        { spec: { id: '789', groupIds: ['a', 'b'] }, tests: [{ specId: '789', id: 'random4' }], groups: [{ id: 'a', testingType: 'e2e' }, { id: 'b', testingType: 'e2e' }], foundLocally: false, testingType: 'e2e', matchesCurrentTestingType: true },
       ]
 
       expect(debugMappingArray).to.deep.equal(expected)
@@ -138,12 +138,12 @@ describe('<DebugContainer />', () => {
         { id: '789', groupIds: ['a'] },
       ] as DebugSpecListSpecFragment[]
       const tests = [{ specId: '123', id: 'random1' }] as DebugSpecListTestsFragment[]
-      const groups = [{ id: 'a' }] as DebugSpecListGroupsFragment[]
+      const groups = [{ id: 'a', testingType: 'e2e' }] as DebugSpecListGroupsFragment[]
 
       const debugMappingArray = specsList({ specs, tests, localSpecs: [], currentTestingType: 'e2e', groups })
 
       expect(debugMappingArray).to.have.length(1)
-      expect(debugMappingArray).to.deep.equal([{ spec: { id: '123', groupIds: ['a'] }, tests: [{ specId: '123', id: 'random1' }], groups: [{ id: 'a' }], foundLocally: false, testingType: 'e2e', matchesCurrentTestingType: true }])
+      expect(debugMappingArray).to.deep.equal([{ spec: { id: '123', groupIds: ['a'] }, tests: [{ specId: '123', id: 'random1' }], groups: [{ id: 'a', testingType: 'e2e' }], foundLocally: false, testingType: 'e2e', matchesCurrentTestingType: true }])
     })
 
     it('throws an error when a test does not map to a spec', () => {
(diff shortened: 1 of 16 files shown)


packages/frontend-shared/src/components/Button.vue
```diff
@@ -126,7 +126,11 @@ const props = defineProps<{
 
 const attrs = useAttrs() as ButtonHTMLAttributes
 
-const variantClasses = computed(() => (VariantClassesTable[props.variant || 'primary']))
+const variantClasses = computed(() => {
+  return (VariantClassesTable[props.variant || 'primary']).split(' ').filter((css) => {
+    return css !== 'hocus-default' || !props.disabled
```
ZachJW34: A button should not have hover states if it is disabled.

packages/frontend-shared/src/components/Tooltip.vue
```diff
@@ -71,6 +71,17 @@ const actualPopperClass = computed(() => {
     result.push('no-arrow')
   }
 
+  // color takes priority, else fallback to tooltip -> dark, menu -> light
```
ZachJW34, line 74: Refactored so as to allow menu functionality styled with the dark tooltip theme (tooltips don't stay open when the mouse leaves the designated hover element, but the tooltip added in this PR has a call to action).

packages/frontend-shared/src/components/Button.vue
warrensplayer: I see how this solves the issue for the button type being used in this issue, but it is not really clear here what it is doing or why.  Your comment here in the PR was helpful.

Can you at least separate out the additional functionality and not chain it off the original part so that you can add a good comment to explain why this is being performed?

packages/app/src/debug/DebugContainer.vue
```diff
@@ -86,15 +103,25 @@ const run = computed(() => {
...
 
+const localSpecs = computed(() => {
+  return props.gql.currentProject?.specs || []
+})
+
 const debugSpecsArray = computed(() => {
   if (run.value) {
     const specs = run.value.specs || []
     const tests = run.value.testsForReview || []
+    const currentTestingType = props.gql.currentProject?.currentTestingType || 'e2e'
```
warrensplayer: Why default to `e2e`?  This should always be populated at this point on this page.

packages/app/src/debug/utils/DebugMapping.ts
```diff
@@ -1,27 +1,64 @@
...
-    spec.tests = [...(spec.tests || []), curr]
+      const groups = (foundSpec.groupIds || []).reduce<DebugSpecListGroupsFragment[]>((acc, id) => {
+        if (id) {
+          acc.push(groupsMap[id])
+        }
+
+        return acc
+      }, [])
+      // should always exist, the testingType should not differ between groups
+      const testingType = (groups[0]?.testingType || 'e2e') as TestingTypeEnum
```
warrensplayer: Why default to `e2e` here?  It should always be populated.

packages/frontend-shared/src/components/Button.vue
```diff
@@ -126,7 +126,11 @@ const props = defineProps<{
 
 const attrs = useAttrs() as ButtonHTMLAttributes
 
-const variantClasses = computed(() => (VariantClassesTable[props.variant || 'primary']))
+const variantClasses = computed(() => {
+  return (VariantClassesTable[props.variant || 'primary']).split(' ').filter((css) => {
+    return !(css === 'hocus-default' && props.disabled)
```
warrensplayer: I am not a huge fan of having to solve this with string manipulation, but I think it is fine for this use case.  My only request is separate out the new logic with the filter and either put a nice comment on it or wrap it in a function with a descriptive name.  

Also, this solution does not fix the `secondary` button since it uses a different class `hocus-secondary`.  Filter that one out as well.

packages/frontend-shared/src/components/Tooltip.vue
```diff
@@ -46,7 +46,7 @@ const props = withDefaults(defineProps<{
   instantMove?: boolean
   showGroup?: string
 }>(), {
-  color: 'dark',
+  color: undefined,
```
warrensplayer, line 49: What did the component do with the `color` prop originally.  It does not look like it was used.  I am guessing it just got passed on to the `Menu` or `Tooltip` components from `floating-vue`.  I looked at the documentation for that library but could not find any documentation for it.  Did you look?

packages/frontend-shared/src/components/Tooltip.vue
```diff
@@ -71,6 +71,17 @@ const actualPopperClass = computed(() => {
     result.push('no-arrow')
   }
 
+  // color takes priority, else fallback to tooltip -> dark, menu -> light
+  if (props.color === 'dark') {
```
warrensplayer, line 75: Can you add some scenarios to the component test for this to show off these new options?

warrensplayer (commented):
Overall is great!  I love the addition of the types in the DebugMapping.ts file.  Makes things much easier to understand.  Mostly nit-pick stuff in my comments.

packages/frontend-shared/src/components/Tooltip.vue
ZachJW34: Couldn't find any docs for it and it wasn't being referenced so I hijacked it!

packages/frontend-shared/src/components/Tooltip.vue
ZachJW34: Added tests in [52feb4b](https://github.com/cypress-io/cypress/pull/25234/commits/52feb4bb9e19e34910da6de2cc730478aa1cfb68)

packages/frontend-shared/src/components/Button.vue
ZachJW34: Addressed in [52feb4b](https://github.com/cypress-io/cypress/pull/25234/commits/52feb4bb9e19e34910da6de2cc730478aa1cfb68)

packages/app/src/debug/utils/DebugMapping.ts
ZachJW34: I was fighting TS, fixed in [52feb4b](https://github.com/cypress-io/cypress/pull/25234/commits/52feb4bb9e19e34910da6de2cc730478aa1cfb68)

packages/app/src/debug/DebugContainer.vue
ZachJW34: Addressed in [52feb4b](https://github.com/cypress-io/cypress/pull/25234/commits/52feb4bb9e19e34910da6de2cc730478aa1cfb68)

astone123:
Is the tooltip supposed to look like this? Looks like the body needs to get moved up a couple of pixels
Screen Shot 2022-12-21 at 4 18 48 PM

packages/app/src/debug/utils/DebugMapping.ts
```diff
@@ -1,27 +1,64 @@
...
+  const groupsMap = groups.reduce<{[id: string]: DebugSpecListGroupsFragment}>((acc, group) => ({ ...acc, [group.id]: group }), {})
+
+  const mappedTests = tests.reduce<{[id: string]: CloudDebugSpec}>((acc, curr) => {
+    let debugResult = acc[curr.specId]
+
+    if (!debugResult) {
       const foundSpec = specs.find((spec) => spec.id === curr.specId)
 
-      spec = { spec: foundSpec }
-      // console.log('looking for spec', spec)
```
astone123, line 10: Remove this comment?

packages/frontend-shared/src/components/Tooltip.cy.tsx
```diff
@@ -74,4 +75,50 @@ describe('<Tooltip />', () => {
...
+
+    cy.get('body').realHover()
+    cy.get('.v-popper__popper').should('not.exist', { timeout: 250 })
+  })
+
+  it('should allow dark theme for interactive Tooltip', () => {
+    cy.mount(() => {
+      return (
+        <div class="flex m-100px p-4 w-150px gap-160px">
+          {/* @ts-ignore */}
```
astone123: What is this ts-ignore for?

astone123 (approved):
Looks good for the most part - just had a couple of comments. I also left a comment on one of the Percy snapshots that changed. It looks like that might actually be related to a change that was made here. We should check that out just to make sure 👍🏻

warrensplayer (commented):
The code looks good. The only thing I saw that was strange in the UI was the little arrow on the tooltip.  It is a little bit too high above the body of the tooltip.  If you zoom in, you can see the bottom arrows of the triangle.  
image

packages/app/src/debug/utils/DebugMapping.ts
ZachJW34: I removed this in a previous commit, maybe the PR was stale when you were reviewing?

packages/frontend-shared/src/components/Tooltip.cy.tsx
ZachJW34: No idea it was a copy and paste. Let me see if I can remove it without check-ts being pissed.

ZachJW34:
@astone123 I reverted the changes I made to the Button component but the test/snapshot didn't change. Not sure why, I'll check this out on the feature branch.

packages/app/src/debug/DebugContainer.cy.tsx
```diff
@@ -83,69 +83,81 @@ describe('<DebugContainer />', () => {
...
         { specId: 'a1c', id: 'random2' },
-      ]
+      ] as DebugSpecListTestsFragment[]
+      const groups = [
+        { id: 'a', testingType: 'e2e' },
+        { id: 'b', testingType: 'e2e' },
+      ] as DebugSpecListGroupsFragment[]
 
-      const debugMappingArray = specsList(spec, tests)
+      const debugMappingArray = specsList({ specs, tests, groups, localSpecs: [], currentTestingType: 'e2e' })
```
lmiller1990, line 99: Now we are on Vite 4 you should have an easier time adding `vitest` to the code base. Not needed for this PR to merge, but I'd like to start using Vitest more for non UI tests - as much as I like dogfooding Cypress, it really doesn't feel like the right tool for this class of test. 

I know we are on the same page about this, just noting it here since we couldn't add Vitest before, but now we should be able to. Feel free to add next time you see a chance! It's fast :zap:

lmiller1990 (approved):

ZachJW34:
@astone123 looks like that button was relying on css class ordering which is not the proper way to apply overrides (there isn't an ideal way to apply overrides but `bg-gray-500 bg-indigo-500` shouldn't be what we rely on). I've reverted the color change by applying a custom class, though I feel like the disabled primary looked better 😅

warrensplayer (approved):

PR #25235, opened 2022-12-20
chore: improve linting

flotwig:
- For cypress-io/cypress-internal#513

### Additional details

- Simplifies linting a few ways:
	- Removes eslintrc config that duplicates already-extended eslintrc config
	- Removes "extends" that are implicit (eslint always extends from all eslintrcs in the parent list)
	- Removes globally-defined rules that are only applicable in a few specific places
	- Remove some redundant eslint-ignore-next-lines
	- See eslint docs for info on why this is possible: https://eslint.org/docs/latest/user-guide/configuring/configuration-files#cascading-and-hierarchy
- Corrects a few common linting issues:
	- "no-unused-vars: off" is present in a lot of our rules. I've removed most of these and fixed the errors that arose.
	- Some packages had `.eslintrc` but no `yarn lint` command, meaning linting only happened in the IDE, not in CI. Fixed this where needed.
- Also removes an unused jscodeshift dep/script
- Future PRs will further unify the eslintrcs to extend from a few common base scenarios.

### PR Tasks


npm/xpath/src/index.d.ts
```diff
@@ -1,6 +1,7 @@
 /// <reference types="cypress" />
 
 declare namespace Cypress {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
flotwig, line 4: I'm not totally sure what the right thing to do here is, I made this change to fix the lint error from `no-unused-vars` but this doesn't look exactly right. Should `E` on line 14 extend `Subject` instead of `HTMLElement`? @jordanpowell88 any ideas?

package.json
```diff
@@ -176,9 +175,6 @@
     "inquirer": "3.3.0",
     "inquirer-confirm": "2.0.3",
     "jest": "24.9.0",
-    "js-codemod": "cpojer/js-codemod",
-    "jscodemods": "https://github.com/cypress-io/jscodemods.git#01b546e",
-    "jscodeshift": "0.7.0",
```
flotwig, line 181: Removing these 3 historical deps got rid of 249 transitive dependencies and 74 MB:

Before:

```
~ echo **/node_modules/* | wc -w
7371
~ du -c -B M **/node_modules/*
...
2792M   total
```

After:

```
~ echo **/node_modules/* | wc -w
7122
~ du -c -B M **/node_modules/*
...
2718M   total
```

lmiller1990: Nice!

I wonder if there's a tool we can use to detect unused dependencies and prune more?

packages/driver/cypress/fixtures/dangling-element.html
```diff
@@ -17,8 +17,7 @@
           return value
         },
         set (v) {
-          // eslint-disable-next-line no-console
-          console.warn(`tried to set frozen property ${property} with ${v}`)
+                    console.warn(`tried to set frozen property ${property} with ${v}`)
```
lmiller1990: This indentation seems quite large.

lmiller1990 (approved):
Code seems fine. I tried doing `yarn lint-changed` but I got a strange error:

```
$ yarn lint-changed                                            simplify-linting
yarn run v1.22.11
$ lint-changed

Oops! Something went wrong! :(

ESLint: 7.22.0

Error: 'patterns' must be a non-empty string or an array of non-empty strings
    at ESLint.lintFiles (/home/lachlan/code/work/dev/node_modules/eslint/lib/eslint/eslint.js:524:19)
    at Object.execute (/home/lachlan/code/work/dev/node_modules/eslint/lib/cli.js:296:36)
    at main (/home/lachlan/code/work/dev/node_modules/eslint/bin/eslint.js:142:52)
    at Object.<anonymous> (/home/lachlan/code/work/dev/node_modules/eslint/bin/eslint.js:146:2)
    at Module._compile (node:internal/modules/cjs/loader:1155:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1209:10)
    at Module.load (node:internal/modules/cjs/loader:1033:32)
    at Function.Module._load (node:internal/modules/cjs/loader:868:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:22:47
error Command failed with exit code 1.
```

I'm not sure if this ever worked, though, or if anyone uses it. The pre-commit lint hook still works fine, I tested it out.

npm/xpath/src/index.d.ts
lmiller1990: Doe we need `Chainable` to have `Subject`? We might be able to omit it. Should it just be

```
declare namespace Cypress {
  interface Chainable {
    /**
    * Get one or more DOM elements by an XPath selector.
    * **Note:** you can test XPath expressions from DevTools console using $x(...) function, for example $x('//div') to find all divs.
    * @see https://github.com/cypress-io/cypress-xpath
    * @example
    * cy.xpath(`//ul[@class="todo-list"]//li`)
    *   .should('have.length', 3)
    */
    xpath<E extends Node = HTMLElement>(selector: string, options?: Partial<Loggable & Timeoutable>): Chainable<JQuery<E>>
  }
}
```

lmiller1990: We have some examples in the docs that don't pass `Subject` to `Chainable`: https://docs.cypress.io/guides/tooling/typescript-support#Types-for-Custom-Commands

package.json
flotwig: Yeah, it's called depcheck: https://github.com/depcheck/depcheck I want to add this to the repo at some point for sure.

.eslintrc.js
```diff
@@ -56,6 +57,7 @@ module.exports = {
       rules: {
         'no-restricted-properties': 'off',
         'no-restricted-syntax': 'off',
+        'no-console': 'off',
```
chrisbreiding: Are we sure we want to turn this off? I often temporarily add console logs for debugging and would rather not accidentally commit them. I prefer using disable comments to allow console logs where we actually want them committed.

flotwig: This only applies to certain file glob patterns, targeting scripts and test code and other non-app code. I made this change because we do usually want to log in those situations. However, I can also see your point that it's good to have to opt-in to logging, even in scripts and tests. I think I will change this back to requiring explicit opt-in for the reasons you mention.

flotwig: Reverted: 966347aebd

packages/driver/cypress/fixtures/dangling-element.html
flotwig: Ya this was a search+replace screwup. Weirdly eslint doesn't handle indentation here, maybe cuz it's an HTML file? I reverted this in 966347aebd

chrisbreiding (approved):

flotwig:
> I'm not sure if this ever worked, though, or if anyone uses it. The pre-commit lint hook still works fine, I tested it out.

@lmiller1990 It seems broken in `develop` too. I opened a PR here to remove it, since lint-staged covers the functionality: https://github.com/cypress-io/cypress/pull/25308

PR #25237, opened 2022-12-20
feat: add --auto-cancel-after-failures flag

estrada9166:
- Closes 
- [CYCLOUD-639](https://cypress-io.atlassian.net/browse/CYCLOUD-639)
- [CYCLOUD-762](https://cypress-io.atlassian.net/browse/CYCLOUD-762)

Clickup initiative / tech brief: https://app.clickup.com/t/18033298/PM-1764

### User facing changelog

The users can set this flag and the cloud may handle it to cancel the run after the number of failures is greater than the flag value

### PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+base%3Arelease-

[CYCLOUD-639]: https://cypress-io.atlassian.net/browse/CYCLOUD-639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CYCLOUD-762]: https://cypress-io.atlassian.net/browse/CYCLOUD-762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cli/lib/cli.js
```diff
@@ -93,6 +93,7 @@ const parseVariableOpts = (fnArgs, args) => {
 }
 
 const descriptions = {
+  autoCancelAfterFailures: 'sets auto cancellation for the run. pass a number for the failed test threshold, or "false"',
```
emilyrohrbough: Should this be a configuration option as well so users don't have to pass this as a flag each time?

cli/lib/exec/run.js
```diff
@@ -45,6 +45,10 @@ const processRunOptions = (options = {}) => {
 
   const args = ['--run-project', options.project]
 
+  if (options.autoCancelAfterFailures || options.autoCancelAfterFailures === false) {
+    args.push('--auto-cancel-after-failures', options.autoCancelAfterFailures)
+  }
```
emilyrohrbough: I'd only expect this to be passed if autoCancelAfterFailures was not false

emilyrohrbough (commented):
@estrada9166 Is there an issue/jira linked to this change? I am wondering why we named is auto-cancel-after-failures instead of cancel-after-failures

cli/lib/cli.js
emilyrohrbough: We need to specify this only applies when recording to the dashboard. Also, is this number of specs that failed or total failed tests? If this run was split across multiple machines, is this total per machine or per run?

packages/server/lib/util/args.js
```diff
@@ -448,6 +450,10 @@ module.exports = {
       }
     }
 
+    if (autoCancelAfterFailures && !((typeof autoCancelAfterFailures === 'number' && autoCancelAfterFailures % 1 === 0) || autoCancelAfterFailures === false)) {
```
emilyrohrbough: why do we need `autoCancelAfterFailures % 1 === 0`? Should we only care if the number is positive?

cli/lib/cli.js
estrada9166: This should be the number of tests that failed per run

cc. @ryanpei

cli/lib/exec/run.js
estrada9166: It can be false, so the cloud can handle it and override the value set on the project settings

packages/server/lib/util/args.js
estrada9166: We only want to handle integer values - and this way we prevent float values

emilyrohrbough: ah I see, was there issues with `Number.isInteger()`?

cli/lib/exec/run.js
emilyrohrbough: 👍🏻 I think I'd expect the server code to handle this by setting it to false without needing to pass `--auto-cancel-after-failures=false` from the CLI

packages/server/lib/util/args.js
estrada9166: gonna update it!

cli/lib/cli.js
emilyrohrbough: Def want to tie it back to record mode though, otherwise I'd expect this to work locally when I wasn't recording.

cli/lib/exec/run.js
estrada9166: you can pass `--no-auto-cancel-after-failures` and it would set the value to `false`

estrada9166:
I moved it to draft - waiting for some changes on the cloud side so this can be properly handled

rockhold:
> @estrada9166 Is there an issue/jira linked to this change? I am wondering why we named is auto-cancel-after-failures instead of cancel-after-failures

@emilyrohrbough, the corresponding Jira on the Cloud side is https://cypress-io.atlassian.net/browse/CYCLOUD-639. Ryan P could probably answer this question definitively, but he's OOO until next week. The Cloud feature we're looking to provide CLI configuration around, however, is referred to as `Auto Cancellation`. My unofficial guess is that the naming is intended to keep parity with the feature.
image

cli/lib/exec/run.js
brian-mann: @emilyrohrbough no this is necessary... you could have globally **enabled** auto cancellation in the cloud, and you want to override that by turning it **off** in the CLI

cli/lib/cli.js
```diff
@@ -253,6 +256,7 @@ const addCypressRunCommand = (program) => {
   .option('-k, --key <record-key>', text('key'))
   .option('--headed', text('headed'))
   .option('--headless', text('headless'))
+  .option('--no-auto-cancel-after-failures', text('noAutoCancelAfterFailures'))
```
brian-mann: See this: https://github.com/tj/commander.js#other-option-types-negatable-boolean-and-booleanvalue

I don't think we need to define this option here

estrada9166: Without having it defined we get this error

`error: unknown option: --no-auto-cancel-after-failures`

cli/lib/exec/run.js
emilyrohrbough: Gottchya. 👍🏻 The flag definition should be updated to reflect this; it's not clear is tied to globally Cloud configuration when record mode is enabled.

cli/lib/cli.js
```diff
@@ -114,6 +115,7 @@ const descriptions = {
   headed: 'displays the browser instead of running headlessly',
   headless: 'hide the browser instead of running headed (default for cypress run)',
   key: 'your secret Record Key. you can omit this if you set a CYPRESS_RECORD_KEY environment variable.',
+  noAutoCancelAfterFailures: 'disables auto cancellation for the run',
```
emilyrohrbough:
```suggestion
  noAutoCancelAfterFailures: 'override the global Cloud configuration to disables auto cancellation for the run when recording to the Cloud',
```

cli/lib/cli.js
emilyrohrbough:
```suggestion
  autoCancelAfterFailures: 'overrides the global Cloud configuration to set the failed test threshold for auto cancellation or to disabled auto cancellation when recording to the Cloud',
```

emilyrohrbough (changes requested):
Since this is overriding global Cloud configuration for users with [Business or Enterprise tier Cloud plans](https://docs.cypress.io/guides/cloud/smart-orchestration#Activate-Auto-Cancellation-in-Cypress-Cloud), what is the feedback loop for non-paying cloud users and or Cloud uses on a plan that does not have Auto Cancellation set up? It doesn't seem like we are logging any context to users if this option isn't supported or will be ignored.

estrada9166:
@emilyrohrbough the cloud may return a warning on the run creation - @MuazOthman is working on it - he may have more context

MuazOthman:
@emilyrohrbough here's a list of error scenarios: https://docs.google.com/spreadsheets/d/1c9wmn6RgplaImt_1B2bXUJvC2IyFPa-OkXLqkeRF-O8/edit#gid=271916282

cli/lib/cli.js
brian-mann: typo on `recoding`

cli/lib/cli.js
brian-mann: Then I'd suggest we just remove it. I know I originally advocated for it, but having it show up twice (whereas no other option is defined more than once). Originally based on my reading of the docs I thought we got the `--no-*` flag for free, but since it must be defined I would recommend it's not worth the additional complexity of documentation and weight. Setting `false` already works.

brian-mann:
Yeah I don't see the error translations yet. Usually the TR enumerates through the known set of errors and helps translate the errors into more human readable forms. Other TR's that don't know how to do that fall back to a default error descriptor

estrada9166:
@emilyrohrbough @brian-mann Currently we have this warning that is being sent from the cloud, do we want to translate it to another message in the TR?
image

emilyrohrbough:
@estrada9166, we need to add these here: https://github.com/cypress-io/cypress/blob/develop/packages/server/lib/modes/record.js#L406

Based on the links @MuazOthman provided, looks like there will be two (maybe 3) errors that need added.

It looked like these were supposed to be errors too, not warnings?

packages/errors/src/errors.ts
```diff
@@ -351,6 +351,30 @@ export const AllCypressErrors = {
 
         https://on.cypress.io/run-group-name-not-unique`
   },
+  CLOUD_AUTO_CANCEL_NOT_AVAILABLE_IN_PLAN: (arg1: {link: string}) => {
+    return errTemplate`\
+      ${fmt.highlightSecondary(`Auto Cancellation`)} is not included under your current billing plan.
+      To enable this service, please visit your billing and upgrade to another plan with Auto Cancellation.
+      ${fmt.off(arg1.link)}`
```
emilyrohrbough, line 360: Why isn't this link in the error message? Is this expected to change over time?

packages/errors/src/errors.ts
```diff
@@ -351,6 +351,30 @@ export const AllCypressErrors = {
...
+  },
+  CLOUD_AUTO_CANCEL_MISMATCH: (arg1: {runUrl: string}) => {
+    return errTemplate`\
+        You passed the ${fmt.flag(`--auto-cancel-after-failures`)} flag, but this run originally started with a different value on this --auto-cancel-after-failures flag.
+      
+        The existing run is: ${fmt.url(arg1.runUrl)}
```
emilyrohrbough, line 366: Is the existing run the previously started run with a different value?

packages/errors/src/errors.ts
```diff
@@ -351,6 +351,30 @@ export const AllCypressErrors = {
...
   },
+  CLOUD_AUTO_CANCEL_NOT_AVAILABLE_IN_PLAN: (arg1: {link: string}) => {
+    return errTemplate`\
+      ${fmt.highlightSecondary(`Auto Cancellation`)} is not included under your current billing plan.
+      To enable this service, please visit your billing and upgrade to another plan with Auto Cancellation.
+      ${fmt.off(arg1.link)}`
+  },
+  CLOUD_AUTO_CANCEL_MISMATCH: (arg1: {runUrl: string}) => {
+    return errTemplate`\
+        You passed the ${fmt.flag(`--auto-cancel-after-failures`)} flag, but this run originally started with a different value on this --auto-cancel-after-failures flag.
```
emilyrohrbough: What was the original value?

packages/errors/src/errors.ts
```diff
@@ -351,6 +351,30 @@ export const AllCypressErrors = {
 
         https://on.cypress.io/run-group-name-not-unique`
   },
+  CLOUD_AUTO_CANCEL_NOT_AVAILABLE_IN_PLAN: (arg1: {link: string}) => {
+    return errTemplate`\
+      ${fmt.highlightSecondary(`Auto Cancellation`)} is not included under your current billing plan.
+      To enable this service, please visit your billing and upgrade to another plan with Auto Cancellation.
+      ${fmt.off(arg1.link)}`
+  },
+  CLOUD_AUTO_CANCEL_MISMATCH: (arg1: {runUrl: string}) => {
```
emilyrohrbough, line 362: Would this error trigger before the `CLOUD_PARALLEL_GROUP_PARAMS_MISMATCH` error on the cloud side?

emilyrohrbough:
Before this can merge, we will need to add a changelog entry & open the corresponding documentation PR with details about this new CLI flag.

packages/errors/src/errors.ts
estrada9166: This is the link that is being passed https://github.com/cypress-io/cypress/blob/a2b58f0e1875efa30c410aad2fb8939407a65af5/packages/server/lib/modes/record.js#L431 - which is dynamically build using the orgId

packages/errors/src/errors.ts
estrada9166: This is the current run - if you start one group with one value and the next group with another one

packages/errors/src/errors.ts
estrada9166: Yes! this would happen before `CLOUD_PARALLEL_GROUP_PARAMS_MISMATCH `

cli/lib/exec/run.js
```diff
@@ -45,6 +45,10 @@ const processRunOptions = (options = {}) => {
 
   const args = ['--run-project', options.project]
 
+  if (options.autoCancelAfterFailures || options.autoCancelAfterFailures === false) {
```
emilyrohrbough:
```suggestion
  // when false, send this to the server to ensure globally enabled auto cancellation is turned off
  if (options.autoCancelAfterFailures || options.autoCancelAfterFailures === false) {
```

packages/errors/src/errors.ts
```diff
@@ -351,6 +351,32 @@ export const AllCypressErrors = {
...
+    return errTemplate`\
+      ${fmt.highlightSecondary(`Auto Cancellation`)} is not included under your current billing plan.
+
+      To enable this service, please visit your billing and upgrade to another plan with Auto Cancellation.
+      
+      ${fmt.off(arg1.link)}`
+  },
+  CLOUD_AUTO_CANCEL_MISMATCH: (arg1: {runUrl: string}) => {
+    return errTemplate`\
+        You passed the ${fmt.flag(`--auto-cancel-after-failures`)} flag, but this run originally started with a different value on this --auto-cancel-after-failures flag.
```
emilyrohrbough:
```suggestion
        You passed the ${fmt.flag(`--auto-cancel-after-failures`)} flag, but this run originally started with a different value on this  ${fmt.flag(--auto-cancel-after-failures`)}  flag.
```

packages/server/__snapshots__/cypress_spec.js
```diff
@@ -325,7 +325,7 @@ exports['Long Cypress Cloud URL'] = `
   │ Browser:    FooBrowser 88                                                                      │
   │ Specs:      1 found (app.cy.js)                                                                │
   │ Searched:   cypress/e2e/**/*.cy.{js,jsx,ts,tsx}                                                │
-  │ Params:     Tag: false, Group: electron-smoke-tests, Parallel: false                           │
+  │ Params:     Tag: false, Group: electron-smoke-tests, Parallel: false, Auto Cancel After Failu… │
```
emilyrohrbough: Should this be cut off?

packages/server/lib/modes/record.js
```diff
@@ -77,13 +77,14 @@ const throwIfIndeterminateCiBuildId = (ciBuildId, parallel, group) => {
   }
 }
 
-const throwIfRecordParamsWithoutRecording = (record, ciBuildId, parallel, group, tag) => {
-  if (!record && _.some([ciBuildId, parallel, group, tag])) {
+const throwIfRecordParamsWithoutRecording = (record, ciBuildId, parallel, group, tag, autoCancelAfterFailures) => {
+  if (!record && _.some([ciBuildId, parallel, group, tag, autoCancelAfterFailures || autoCancelAfterFailures === false])) {
```
emilyrohrbough:
```suggestion
  if (!record && _.some([ciBuildId, parallel, group, tag, autoCancelAfterFailures !== undefined])) {
```

packages/server/lib/modes/run.ts
```diff
@@ -742,6 +743,7 @@ async function runSpecs (options: { config: Cfg, browser: Browser, sys: any, hea
     specPattern,
     system: _.pick(sys, 'osName', 'osVersion'),
     tag,
+    autoCancelAfterFailures,
```
emilyrohrbough, line 746: The https://docs.cypress.io/api/plugins/before-run-api docs need updated to include information on `autoCancelAfterFailures`

emilyrohrbough: The BeforeRun types also need updated: https://github.com/cypress-io/cypress/blob/alejandro/feat/add-autoCancelAfterFailures-flag/cli/types/cypress.d.ts#L5768

packages/server/lib/util/print-run.ts
```diff
@@ -168,16 +168,16 @@ function formatNodeVersion ({ resolvedNodeVersion, resolvedNodePath }: Pick<Cfg,
   return
 }
 
-function formatRecordParams (runUrl?: string, parallel?: boolean, group?: string, tag?: string) {
+function formatRecordParams (runUrl?: string, parallel?: boolean, group?: string, tag?: string, autoCancelAfterFailures?: number | false) {
   if (runUrl) {
-    return `Tag: ${tag || 'false'}, Group: ${group || 'false'}, Parallel: ${Boolean(parallel)}`
+    return `Tag: ${tag || 'false'}, Group: ${group || 'false'}, Parallel: ${Boolean(parallel)}, Auto Cancel After Failures: ${autoCancelAfterFailures || 'false'}`
```
emilyrohrbough:
```suggestion
    return `Tag: ${tag || 'false'}, Group: ${group || 'false'}, Parallel: ${Boolean(parallel)}, Auto Cancel After Failures: ${autoCancelAfterFailures}`
```

packages/server/test/unit/util/args_spec.js
```diff
@@ -229,6 +229,14 @@ describe('lib/util/args', () => {
     })
   })
 
+  context('--auto-cancel-after-failures', () => {
+    it('converts to integer', function () {
+      const options = this.setup('--auto-cancel-after-failures', '4')
+
+      expect(options.autoCancelAfterFailures).to.eq(4)
+    })
```
emilyrohrbough, line 237: Can we test that this ensure number or false value?

packages/server/lib/util/print-run.ts
emilyrohrbough: Should this fallback to false here? It would indicate this value is false when really it would be the global configuration right?

packages/server/lib/modes/record.js
estrada9166: If we want to throw always that the value exits, also when is not valid I think `_.some([ciBuildId, parallel, group, tag, autoCancelAfterFailures])`  may do the trick

estrada9166: If we want to throw always that the value exits, also when is not valid I think `_.some([ciBuildId, parallel, group, tag, autoCancelAfterFailures])`  may do the trick

packages/server/lib/util/print-run.ts
emilyrohrbough: Also thinking about this, if only some plans support auto cancellation, should we only conditionally show this in the run console?

cli/__snapshots__/cli_spec.js
```diff
@@ -67,29 +67,30 @@ exports['shows help for run --foo 1'] = `
...
-    -P, --project <project-path>               path to the project
-    -q, --quiet                                run quietly, using only the configured reporter
-    --record [bool]                            records the run. sends test results, screenshots and videos to Cypress Cloud.
-    -r, --reporter <reporter>                  runs a specific mocha reporter. pass a path to use a custom reporter. defaults to "spec"
-    -o, --reporter-options <reporter-options>  options for the mocha reporter. defaults to "null"
-    -s, --spec <spec>                          runs specific spec file(s). defaults to "all"
-    -t, --tag <tag>                            named tag(s) for recorded runs in Cypress Cloud
-    --dev                                      runs cypress in development and bypasses binary check
-    -h, --help                                 display help for command
+    --auto-cancel-after-failures <autoCancelAfterFailures>  overrides the global Cloud configuration to set the failed test threshold for auto cancellation or to disabled auto cancellation when recording to the Cloud
```
mschile:
```suggestion
    --auto-cancel-after-failures <test-failure-count || false>  overrides the global Cloud configuration to set the failed test threshold for auto cancellation or to disable auto cancellation when recording to the Cloud
```

cli/lib/cli.js
```diff
@@ -93,6 +93,7 @@ const parseVariableOpts = (fnArgs, args) => {
 }
 
 const descriptions = {
+  autoCancelAfterFailures: 'overrides the global Cloud configuration to set the failed test threshold for auto cancellation or to disabled auto cancellation when recording to the Cloud',
```
mschile:
```suggestion
  autoCancelAfterFailures: 'overrides the global Cloud configuration to set the failed test threshold for auto cancellation or to disable auto cancellation when recording to the Cloud',
```

cli/lib/cli.js
```diff
@@ -242,6 +243,7 @@ const addCypressRunCommand = (program) => {
   .command('run')
   .usage('[options]')
   .description('Runs Cypress tests from the CLI without the GUI')
+  .option('--auto-cancel-after-failures <autoCancelAfterFailures>', text('autoCancelAfterFailures'))
```
mschile:
```suggestion
  .option('--auto-cancel-after-failures <test-failure-count || false>', text('autoCancelAfterFailures'))
```

cli/lib/exec/run.js
```diff
@@ -45,6 +45,11 @@ const processRunOptions = (options = {}) => {
 
   const args = ['--run-project', options.project]
 
+  // when false, send this to the server to ensure globally enabled auto cancellation is turned off
```
mschile: > // when false, send this to the server to ensure globally enabled auto cancellation is turned off

This comment is a little misleading, the arg is sent if `autoCancelAfterFailures` is set to a value or if it is false, not only if it is `false` as the comment suggests.

cli/lib/exec/run.js
```diff
@@ -45,6 +45,11 @@ const processRunOptions = (options = {}) => {
...
+  if (options.autoCancelAfterFailures || options.autoCancelAfterFailures === false) {
```
mschile: Can the user disable auto cancellation by setting `autoCancelAfterFailures` to `0`?

cli/types/cypress-npm-api.d.ts
```diff
@@ -95,6 +95,10 @@ declare namespace CypressCommandLine {
      * Specify the specs to run
      */
     spec: string
+    /**
+     * Specify the number of failures to cancel a run being recorded to the cloud
```
mschile:
```suggestion
     * Specify the number of failures to cancel a run being recorded to the cloud or false to disable auto-cancellation.
```

packages/server/__snapshots__/cypress_spec.js
emilyrohrbough: existing issue: https://github.com/cypress-io/cypress/issues/25736

cli/lib/exec/run.js
estrada9166: Yes! if the value is set to 0 it will override the project settings making it disabled - for the current run

packages/errors/src/errors.ts
```diff
@@ -351,6 +351,32 @@ export const AllCypressErrors = {
...
+    return errTemplate`\
+      ${fmt.highlightSecondary(`Auto Cancellation`)} is not included under your current billing plan.
+
+      To enable this service, please visit your billing and upgrade to another plan with Auto Cancellation.
+      
+      ${fmt.off(arg1.link)}`
+  },
+  CLOUD_AUTO_CANCEL_MISMATCH: (arg1: {runUrl: string}) => {
+    return errTemplate`\
+        You passed the ${fmt.flag(`--auto-cancel-after-failures`)} flag, but this run originally started with a different value on this  ${fmt.flag(`--auto-cancel-after-failures`)}  flag.
```
mschile:
```suggestion
        You passed the ${fmt.flag(`--auto-cancel-after-failures`)} flag, but this run originally started with a different value for the ${fmt.flag(`--auto-cancel-after-failures`)} flag.
```

packages/server/lib/util/args.js
```diff
@@ -448,6 +450,10 @@ module.exports = {
       }
     }
 
+    if (autoCancelAfterFailures && !((typeof autoCancelAfterFailures === 'number' && Number.isInteger(autoCancelAfterFailures)) || autoCancelAfterFailures === false)) {
+      return errors.throwErr('COULD_NOT_PARSE_ARGUMENTS', 'auto-cancel-after-failures', autoCancelAfterFailures, 'auto-cancel-after-failures must be a integer or false')
```
mschile:
```suggestion
      return errors.throwErr('COULD_NOT_PARSE_ARGUMENTS', 'auto-cancel-after-failures', autoCancelAfterFailures, 'auto-cancel-after-failures must be an integer or false')
```

cli/lib/exec/run.js
mschile: Make sure the above `if` statement gets updated since it will currently not pass. Probably want to add a test for it as well.

packages/server/test/unit/util/args_spec.js
```diff
@@ -229,6 +229,36 @@ describe('lib/util/args', () => {
     })
   })
 
+  context('--auto-cancel-after-failures', () => {
```
mschile, line 232: Suggest adding a test for a negative number, a decimal number, and zero.

mschile:
@estrada9166, could you please respond to all of the comments and resolve the conversations.

packages/server/test/unit/util/args_spec.js
estrada9166: the tests were added and the `0` value pass

cli/lib/exec/run.js
estrada9166: @emilyrohrbough suggested it

estrada9166:
@mschile done!

cli/lib/exec/run.js
emilyrohrbough: yeah Matt's right, def feel free to tweak. I was originally confused why we'd send a false value down.

cli/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ _Released 02/14/2023 (PENDING)_
 **Features:**
 
 - Added the "Open in IDE" feature for failed tests reported from the Debug page. Addressed in [#25691](https://github.com/cypress-io/cypress/pull/25691).
+- Added "--auto-cancel-after-failures" flag that would override the "Auto Cancellation" feature when recording to the cloud. Addressed in [#25237](https://github.com/cypress-io/cypress/pull/25237)
```
mschile:
```suggestion
- Added "--auto-cancel-after-failures" flag that would override the "Auto Cancellation" feature when recording to the Cloud. Addressed in [#25237](https://github.com/cypress-io/cypress/pull/25237)
```

cli/lib/exec/run.js
mschile: @estrada9166, just wanted to make sure you saw this.

cli/types/cypress-npm-api.d.ts
```diff
@@ -95,6 +95,10 @@ declare namespace CypressCommandLine {
      * Specify the specs to run
      */
     spec: string
+    /**
+     * Specify the number of failures to cancel a run being recorded to the cloud or false to disable auto-cancellation.
```
mschile:
```suggestion
     * Specify the number of failures to cancel a run being recorded to the Cloud or false to disable auto-cancellation.
```

cli/__snapshots__/cli_spec.js
```diff
@@ -67,29 +67,30 @@ exports['shows help for run --foo 1'] = `
...
-    -P, --project <project-path>               path to the project
-    -q, --quiet                                run quietly, using only the configured reporter
-    --record [bool]                            records the run. sends test results, screenshots and videos to Cypress Cloud.
-    -r, --reporter <reporter>                  runs a specific mocha reporter. pass a path to use a custom reporter. defaults to "spec"
-    -o, --reporter-options <reporter-options>  options for the mocha reporter. defaults to "null"
-    -s, --spec <spec>                          runs specific spec file(s). defaults to "all"
-    -t, --tag <tag>                            named tag(s) for recorded runs in Cypress Cloud
-    --dev                                      runs cypress in development and bypasses binary check
-    -h, --help                                 display help for command
+    --auto-cancel-after-failures <test-failure-count || false>  overrides the global Cloud configuration to set the failed test threshold for auto cancellation or to disable auto cancellation when recording to the Cloud
```
ryanpei:
```suggestion
    --auto-cancel-after-failures <test-failure-count || false>  overrides the project-level Cloud configuration to set the failed test threshold for auto cancellation or to disable auto cancellation when recording to the Cloud
```

cli/lib/cli.js
```diff
@@ -93,6 +93,7 @@ const parseVariableOpts = (fnArgs, args) => {
 }
 
 const descriptions = {
+  autoCancelAfterFailures: 'overrides the global Cloud configuration to set the failed test threshold for auto cancellation or to disable auto cancellation when recording to the Cloud',
```
ryanpei:
```suggestion
  autoCancelAfterFailures: 'overrides the project-level Cloud configuration to set the failed test threshold for auto cancellation or to disable auto cancellation when recording to the Cloud',
```

packages/errors/src/errors.ts
estrada9166: Updated it here [63a7744](https://github.com/cypress-io/cypress/pull/25237/commits/63a77440084a5d0b210fa410950b259bdf6e3007)

mschile (approved):

cli/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ _Released 02/14/2023 (PENDING)_
 **Features:**
 
 - Added the "Open in IDE" feature for failed tests reported from the Debug page. Addressed in [#25691](https://github.com/cypress-io/cypress/pull/25691).
+- Added "--auto-cancel-after-failures" flag that would override the "Auto Cancellation" feature when recording to the Cloud. Addressed in [#25237](https://github.com/cypress-io/cypress/pull/25237)
```
emilyrohrbough:
```suggestion
- Added a new CLI flag, called [`--auto-cancel-after-failures`](https://docs.cypress.io/guides/guides/command-line#Options), that overrides the project-level CI ["Auto Cancellation"](https://docs.cypress.io/guides/cloud/smart-orchestration#Auto-Cancellation) value when recording to the Cloud. This gives Cloud users on Business and Enterprise plans the flexibility to alter the auto-cancellation value per run. Addressed in [#25237](https://github.com/cypress-io/cypress/pull/25237)
```

emilyrohrbough (approved):

cli/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ _Released 02/14/2023 (PENDING)_
 **Features:**
 
 - Added the "Open in IDE" feature for failed tests reported from the Debug page. Addressed in [#25691](https://github.com/cypress-io/cypress/pull/25691).
+- Added a new CLI flag, called [`--auto-cancel-after-failures`](https://docs.cypress.io/guides/guides/command-line#Options), that overrides the project-level CI ["Auto Cancellation"](https://docs.cypress.io/guides/cloud/smart-orchestration#Auto-Cancellation) value when recording to the Cloud. This gives Cloud users on Business and Enterprise plans the flexibility to alter the auto-cancellation value per run. Addressed in [#25237](https://github.com/cypress-io/cypress/pull/25237)
```
emilyrohrbough:
```suggestion
- Added a new CLI flag, called [`--auto-cancel-after-failures`](https://docs.cypress.io/guides/guides/command-line#Options), that overrides the project-level CI ["Auto Cancellation"](https://docs.cypress.io/guides/cloud/smart-orchestration#Auto-Cancellation) value when recording to the Cloud. This gives Cloud users on Business and Enterprise plans the flexibility to alter the auto-cancellation value per run. Addressed in [#25237](https://github.com/cypress-io/cypress/pull/25237).
```

PR #25250, opened 2022-12-21
fix: .contains() should only return one element at all times

### LINKED ISSUE #25225: Problem with cy.contains(). It find two elements. It should find one
### Current behavior

After update to Cypress 12.1.0 some tests stop working. The problem is with cy.contains(). 

Before this code: `cy.dataCy("OpenAddFormsButton.label").contains("Zadanie").click();` found only one element. Now it finds two elements: 
Screenshot_154

### Desired behavior

In this situation it should find only element which contains only word "Zadanie" not "Zadanie Cykliczne" also. 

### Test code to reproduce

Screenshot_154

  it("adds new task", () => {
    cy.dataCy("OpenAddFormsButton.add").contains("Dodaj").click();
    cy.dataCy("OpenAddFormsButton.label").contains("Zadanie").click();
  });
Screenshot_155

### Cypress Version

12.1.0

### Node version

16.14.2

### Operating System

Windows 11

### Debug Logs

_No response_

### Other

_No response_

BlueWinds:
- Closes https://github.com/cypress-io/cypress/issues/25225

### User facing changelog
Fix for regression introduced in 12.1.0, where `.contains()` could return multiple elements when it was matching directly on the subject, rather than on the subject's children.

### Additional details
Basically, we were *either* filtering down one element *or* checking to see if the subject matched the selector - so if there were multiple subjects that matched the contains filter (but none of their child elements matched), we'd return them all.

Fixed by moving the filtering to be the last step.

### Steps to test

In addition to the included driver test, you can reproduce the issue like so:
```
    it('issue ', () => {
      cy.visit('https://example.cypress.io')
      cy.get('a').contains('next').click()
    })
```

### PR Tasks


packages/driver/cypress/e2e/commands/querying/querying.cy.js
```diff
@@ -1020,6 +1020,11 @@ describe('src/cy/commands/querying', () => {
       })
     })
 
+    // https://github.com/cypress-io/cypress/issues/25225
+    it('returns one element when given multiple subjects with no children', () => {
+      cy.get('button').contains('submit').should('have.length', 1)
```
emilyrohrbough, line 1026: I'd like to see test coverage from returning one element when multiple subjects regardless if the subjects have children or not. I glanced through these tests and not seeing an equivalent. When debugging this I didn't expect the issue to related to the subject's dom tree. Would always expect the first matching element

This might not be the right get selector - but I expect both scenarios to pass.
```suggestion
      cy.get('button').contains('submit').should('have.length', 1)
      cy.get('div').contains('submit').should('have.length', 1)
```

emilyrohrbough (changes requested):

lmiller1990 (approved):
Definitely a fan of more thorough tests - I can't believe we  didn't have a test around this, I write code like this all the time.

I couldn't find any other broken use cases with `contains`. I tested the fix and it works.

packages/driver/cypress/e2e/commands/querying/querying.cy.js
BlueWinds: I'll edit the test name to be a bit more descriptive - "has no children" isn't quite accurate. It's "when the subjects themselves match the selector, and none of their children do."

I think we already have lots of coverage around the case where .contains() simply matches multiple elements. Other tests in this file we have covering similar cases:

```
it('searches multiple subject elements', () => {
  cy.get('ul').contains('li', 'asdf 3')
})

it('resets the subject between chain invocations', () => {
  ...snip...
  cy.get('#complex-contains').contains('nested contains').then(($label) => {

it('reduces right by priority element', () => {
  ...snip...
  // it should find label because label is the first priority element
  // out of the collection of contains elements
  cy.get('#complex-contains').contains('nested contains')
```

emilyrohrbough: This returns the first matched element correct? This would align with our documentation

emilyrohrbough: > It's "when the subjects themselves match the selector, and none of their children do."

And do we have cases for these scenarios?
- the children match the selector and the subjects don't
- both the children and the subjects match the selector

emilyrohrbough: Its seems we've had a few holes when 12 went out with our test coverage, so Im surprised each of fixes has only had one test verifying its behavior. If we have existing tests for the above scenarios, mind linking them?

BlueWinds: > the children match the selector and the subjects don't

This is super common. For example,
```
<div>
  <span>Foo</span>
</div>

cy.get('div').contains('span', 'Foo')
```

The child matches the selector, but the parent does not. Some examples from the file:
```
it('searches multiple subject elements'
it('finds the furthest descendent when filter matches more than 1 element'
it('retries until it finds a filtered contains has the matching text node'
```

> both the children and the subjects match the selector

Every time a child matches, the parent also does.

```
<div>
  <div>Foo</div>
</div>

cy.get('div').contains('Foo')
```

The div contains the text 'Foo', and so does the span. Basically every test in the file covers this case.

BlueWinds: We are missing one in the form `.get('div').contains('div', 'text')`. I'll add that in a moment here.

emilyrohrbough (approved):

PR #25251, opened 2022-12-21
feat: Add 'type' option to `.as` to store aliases by value

### LINKED ISSUE #25173: Alias seems to not store it's value
### Current behavior

During the test I'm comparing two values 

First one is stored as 'backgroundImageDefault' with following code

```
    cy.get(selectors.pageAddress)
      .prev()
      .invoke('css', 'backgroundImage')
      .as('backgroundImageDefault');
```
At this point the `backgroundImageDefault` has value `https://host.com/smth/defaultImage.jpg`.

Then I'm changing the image, the new image obviously has a new address, `https://host.com/smth/changedImage.jpg`.

```
    cy.get('@backgroundImageDefault').then((background) => {
      cy.get(selectors.pageAddress)
        .prev()
        .should('not.have.css', 'background-image', background);
    });
```
    
It fails becouse when comparing it seems that alias 'backgroundImageDefault' now has value of `https://host.com/smth/changedImage.jpg` instead of it's assigned value of `https://host.com/smth/defaultImage.jpg`.

### Desired behavior

Until I upgraded to 12.0.0/12.1.0 the test simply passed and the value stored in the `backgroundImageDefault` was correct.

### Test code to reproduce

It's the code mentioned in current behaviour

### Cypress Version

12.1.0

### Node version

v16.15.1

### Operating System

macOS 13.1

### Debug Logs

_No response_

### Other

_No response_

BlueWinds:
- Closes https://github.com/cypress-io/cypress/issues/25173

### User facing changelog
`.as()` now accepts an `options` argument, allowing test writers to store an alias as a "query" or "static".

A "query" alias (the default) re-quries the application for the current state whenever retrieved - it's a 'live' value.
A "static" alias is stored once, when the alias is created - it's a static value, and will never update.

Don't store DOM references as 'value's.

### Additional details
```
it('allows users to store a static value', () => {
  const obj = { foo: 'bar' }

  cy.wrap(obj).its('foo').as('value', { type: 'static' })
  cy.wrap(obj).its('foo').as('query', { type: 'query' })

  cy.then(() => {
    obj.foo = 'baz'
  })

  cy.get('@value').should('eq', 'bar')
  cy.get('@query').should('eq', 'baz')
})
```

Before:
image

After:

Aliases are `query` by default. If the user sets the `type` to `static`, then it is included in the command log.

image

### PR Tasks

- [x] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? https://github.com/cypress-io/cypress-documentation/pull/4943


cli/types/cypress.d.ts
```diff
@@ -2661,15 +2660,19 @@ declare namespace Cypress {
...
-     * `drag-drop` matches dragging files from the operating system into the
-     * document.
+     * If an alias should be a 'query' (re-fetched every time so it's always live) or a
+     * 'value' (read once when the alias is saved, never updated).
      *
-     * @default 'select'
+     * @default 'query'
      */
-    action: 'select' | 'drag-drop'
+    type: 'query' | 'value'
```
```suggestion
    type?: 'query' | 'value'
```

packages/driver/cypress/e2e/commands/window.cy.js
```diff
@@ -145,7 +145,7 @@ describe('src/cy/commands/window', () => {
 
           expect(win).to.eq(this.win)
 
-          expect(this.logs[0].get('alias')).to.eq('win')
+          expect(this.logs[0].get('alias')).to.eq('@win')
```
emilyrohrbough, line 148: why did these change?

packages/driver/src/cy/commands/aliasing.ts
```diff
@@ -42,7 +58,7 @@ export default function (Commands, Cypress, cy) {
 
       if (!alreadyAliasedLog && log) {
         log.set({
-          alias,
+          alias: `@${alias}${options.type ? ` (${ options.type })` : ''}`,
```
emilyrohrbough:
```suggestion
          alias: `@${alias}${options.type ? ` (${ options.type })` : '(query)'}`,
```
the default is type query, so why not set this as the fallback?

packages/driver/src/cypress/error_messages.ts
```diff
@@ -141,6 +141,8 @@ export default {
     reserved_word: {
       message: `${cmd('as')} cannot be aliased as: \`{{alias}}\`. This word is reserved.`,
     },
+    invalid_options: `${cmd('as')} only accepts an \`options\` object for its second argument. You passed: \`{{arg}}\``,
```
emilyrohrbough: Other error messages for options object  doesn't user back-ticks. should we remove to be consistent?

packages/driver/cypress/e2e/commands/window.cy.js
BlueWinds: I made the change so that we always display the '@' symbol in the command log, for both storing and retrieving - the consistency seemed more intuitive to me. It's not a necessary change and can be reverted without affecting the functionality if people disagree / discussion around it is going to delay the main features here.

It used to look like

```
// storing the alias
get     #button                            btn

// reading the alias
get     @btn
```

and now it's consistent

```
// storing the alias
get     #button                            @btn

// reading the alias
get     @btn
```

See the before/after screenshot in PR description (the 'after' shows the @ when both saving and reading the alias).

packages/driver/src/cy/commands/aliasing.ts
BlueWinds: Take a look at the "After" screenshot it the PR description, showing how it looks with `(value)` tagged onto the end - it's a bit visually busy, so I was trying to avoid it in the default case. If the user sets it explicitly - whether it be query or value - then it'll show up in the command log.

Can definitely make it always display if we like, just 'display less by default' was my intention here.

packages/driver/src/cypress/error_messages.ts
BlueWinds: Done in upcoming commit.

cli/types/cypress.d.ts
BlueWinds: Done in upcoming commit.

packages/driver/src/cy/commands/aliasing.ts
```diff
@@ -1,18 +1,34 @@
 import _ from 'lodash'
 import $dom from '../../dom'
+import $errUtils from '../../cypress/error_utils'
 
 export default function (Commands, Cypress, cy) {
-  Commands.addQuery('as', function asFn (alias) {
+  Commands.addQuery('as', function asFn (alias, options = {} as Partial<Cypress.AsOptions>) {
```
emilyrohrbough, line 6: I'd expect to use Cypress.AsOptions directly . Should be able to if the keys are optional.

packages/driver/src/cy/commands/aliasing.ts
emilyrohrbough: @BlueWinds Because these wouldn't align
```
.as('name')
.as('name', { type: 'query' }
```
Could we only prepend when the non-default is used (i.e. only when type === value)?

emilyrohrbough: It'd be helpful to also add this type to the dev tools log details when clicking a log that is referencing an alias to provide the in-depth context to how the alias value is resolved.

cli/types/cypress.d.ts
```diff
@@ -2661,15 +2660,19 @@ declare namespace Cypress {
...
-     * `drag-drop` matches dragging files from the operating system into the
-     * document.
+     * If an alias should be a 'query' (re-fetched every time so it's always live) or a
+     * 'value' (read once when the alias is saved, never updated).
      *
-     * @default 'select'
+     * @default 'query'
      */
-    action: 'select' | 'drag-drop'
+    type?: 'query' | 'value'
```
emilyrohrbough: Pulling [comment](https://github.com/cypress-io/cypress-documentation/pull/4943#discussion_r1072295359) forward from the docs  - worth getting other opinions on I think.

> These types seem very specific to element queries and context sharing. The `query` type seems like an odd name for aliasing requests or intercepts since they aren't query commands, but these types of aliases would likely need to default to query to work as expected.

Have we considered alternatives like `dynamic` and `static` or other names that can encompass/reflect non-query command aliases as well?

packages/driver/src/cy/commands/aliasing.ts
BlueWinds: Changed it to be consistent, always displaying `value` and never displaying `query`; see updated screenshot in the PR description.

I'm going to push back against adding it to the console props for .get() when an alias is retrieved - the "type" of an alias is a property of how it's stored, not how it's retrieved, so we should display it where it's stored (which we do).

packages/driver/src/cy/commands/aliasing.ts
BlueWinds: I opted here for consistency. Pre-existing commands all use `Partial<>` in this context, and do not have optional properties.

```
as(alias: string, options?: Partial<AsOptions>): Chainable<Subject>
blur(options?: Partial<BlurOptions>): Chainable<Subject>
check(options?: Partial<CheckOptions>): Chainable<Subject>

...snip...

  interface CheckOptions extends Loggable, Timeoutable, ActionableOptions {
    interval: number
  }
```

Removed the optional-ness of `AsOptions.type` it , since we have Partial<> everywhere AsOptions is used.

cli/types/cypress.d.ts
BlueWinds: `static` does read nicer than `value`, switching over to that. Will update docs PR shortly. I do prefer `query` for the dynamic / default version though - "query" is a cypress concept we reference in the docs, and we should be consistent about its usage.

packages/driver/src/cy/commands/aliasing.ts
```diff
@@ -1,18 +1,34 @@
...
     Cypress.ensure.isChildCommand(this, [alias], cy)
     cy.validateAlias(alias)
 
+    if (!_.isObject(options)) {
```
chrisbreiding:
```suggestion
    if (!_.isPlainObject(options)) {
```

`_.isObject()` returns true for an array

packages/driver/src/cy/commands/aliasing.ts
```diff
@@ -1,18 +1,34 @@
...
+      $errUtils.throwErrByPath('as.invalid_options', { args: { arg: options } })
+    }
+
+    if (options.type && !options.type.match(/query|static/)) {
```
chrisbreiding:
```suggestion
    if (options.type && !['query', 'static'].includes(options.type)) {
```

```
'fooqueryfoo'.match(/query|static/) === true
```

Could also change the regex, but this seems like a simpler alternative.

chrisbreiding (changes requested):

chrisbreiding (approved):

cli/types/cypress.d.ts
```diff
@@ -2661,15 +2660,19 @@ declare namespace Cypress {
...
+   * Options to affect how an alias is stored
+   *
+   * @see https://on.cypress.io/as
+   */
+  interface AsOptions {
     /**
-     * Which user action to perform. `select` matches selecting a file while
-     * `drag-drop` matches dragging files from the operating system into the
-     * document.
+     * If an alias should be a 'query' (re-fetched every time so it's always live) or a
```
```suggestion
     * The type of alias to store, which impacts how the value is retrieved later in the test. 
     * If an alias should be a 'query' (re-runs all queries leading up to the resulting value so it's alway up-to-date) or a
```

cli/types/cypress.d.ts
```diff
@@ -2661,15 +2660,19 @@ declare namespace Cypress {
...
+     * 'value' (read once when the alias is saved, never updated).
```
```suggestion
     * 'static' (read once when the alias is saved and is never updated). `type` has no effect when aliasing intercepts, spies, and stubs.
```

packages/driver/src/cy/commands/aliasing.ts
emilyrohrbough: The documentation PR says the `type` impacts how the value is retrieved later in the test. 😅

emilyrohrbough (approved):
It'd be nice if the types description matched the documentation content.

cli/types/cypress.d.ts
filipedanunes: static as default please

BlueWinds: Defaulting to `static` would be a breaking change for everyone aliasing DOM nodes, which is more common than aliasing static values. Reducing or eliminating detached DOM errors was the motivation for making this change to alias behavior in the first place.

filipedanunes: Humm ok, got it. Thank you!

Issue #25166, opened 2022-12-14
Make `experimentalSingleTabRunMode` not experimental anymore


lmiller1990:
### What would you like?

Works great - we won't make it the default yet, but we will remove the experimental prefix. Two potential issues:

- https://github.com/cypress-io/cypress/issues/24146 claims some issues - are they still issues? Try to reproduce.
- https://github.com/cypress-io/cypress/issues/23815 - since WK is experimental, it's not expected work perfectly, but it would be nice to fix this. We definitely need to do something about it before moving singleTabRunMode to be the default.

### Why is this needed?

`singleTabRunMode` is faster and making it the default is a vastly better experience

### Acceptance Criteria

- Make a proposal to get this feature to GA. What are the gaps?
- Some ideas relating to the WK issue, can we either:
  - 1. If both experiments active AND the browser is webkit - just default to non-single-tab-run-mode (and print a terminal warning?)
  - 2. If both experiments active AND the browser is webkit - error and explain why?
- Estimate how much time you think this will take to get to GA based on your proposal, either:
-  hours? 
- days? 
- weeks?

Deliverable is to add a comment/update this issue with the outstanding work. We can make a ClickUp from that info, once we have a good idea of what's left.

Issue #23815, opened 2022-09-13
`experimentalWebKitSupport`: in single-tab mode, only the first spec's video is recorded


flotwig:
With `experimentalSingleTabRunMode` and WebKit, only the first spec is recorded with video recording enabled.

This is due to how the Playwright driver records video - it requires recording to disk, and there is one file per page/context. This is in contrast with single-tab mode, where many specs run in the same page.

Some fix is needed, either by finding a better solution on the PW side, accessing the underlying WebKit RDP and using `startScreencast/screencastFrame`, or chopping up the video at the end using timing data and ffmpeg, which is probably too slow.

flotwig:
This also seems to sometimes cause an issue where WebKit won't exit in `experimentalSingleTabRunMode`. Per @astone123:

> I'm enabling experimentalSingleTabRunMode for all of our packages and system tests that use CT, and I'm running into issues with running the run_ct_spec.js system test spec on Webkit. Looks like they're running correctly but then never actually exiting.
> When I run them locally headed, the browser opens, I see the tests run successfully, then the browser window exits (but the icon stays in my dock), and then the terminal just hangs until it hits the timeout... looks like there's something that we're not doing in single tab run mode to close the browser completely. Weird that it only happens with Webkit though.

Turning video off fixes it in this case, but it points to some other consequence of this bug I think. Here we always call videoController.restart in single-tab mode https://github.com/cypress-io/cypress/blob/3d1badc4b6fe422b0e60d07f743a24d69b6aed78/packages/server/lib/modes/run.ts#L252-L264

which is stubbed to throw an error in WK: https://github.com/cypress-io/cypress/blob/3d1badc4b6fe422b0e60d07f743a24d69b6aed78/packages/server/lib/browsers/webkit-automation.ts#L148-L150

Possibly `endVideoCapture` is never called at the end of the first spec, causing an issue?

lmiller1990:
This is going to be a problem. Single Tab Run Mode makes CT **way** faster (up to 50%). WebKit is experimental now, but once it's not experimental, this bug is going to block Single Tab Run Mode from bcoming GA. Gating a huge perf boost isn't ideal. I wonder what the complexity of fixing this is. We should spike into it.

PR #25257, opened 2022-12-22
chore: use `experimentalSingleTabRunMode` in all of our monorepo packages and system tests that are configured for CT

### LINKED ISSUE #25165: Move monorepo CT suite to use experimentalSingleTabRunMode
### What would you like?

Working towards making `experimentalSingleTabRunMode` the default. Next step is using it more internally - all CT suites should use it.

### Why is this needed?

Dogfooding our experiment

### Other

_No response_

astone123:
- Closes #25165

### User facing changelog

n/a

### Additional details

We are [moving `experimentalSingleTabRunMode` into GA](https://github.com/cypress-io/cypress/issues/25166) at some point. Before we do this, we should use it ourselves to test it out as the default. This should make it easier to come across any edge cases, perf issues, etc. before we make it the default for everyone.

### Steps to test

Not much to test, just take a look at the CI run and make sure that everything is passing. Note that I didn't make any changes to the `pristine` projects in `system-tests/` since these are used to test our scaffolding logic.

n/a

### PR Tasks

### DIFF
--- a/system-tests/projects/simple-ct/cypress.config.js
+++ b/system-tests/projects/simple-ct/cypress.config.js
@@ -1,6 +1,7 @@
 module.exports = {
   component: {
     experimentalSingleTabRunMode: true,
+    video: false,
     supportFile: false,
     devServer: {
       bundler: 'webpack',
--- a/system-tests/projects/simple-ct/cypress.config.js
+++ b/system-tests/projects/simple-ct/cypress.config.js
@@ -1,5 +1,6 @@
 module.exports = {
   component: {
+    experimentalSingleTabRunMode: true,
     supportFile: false,
     devServer: {
       bundler: 'webpack',
--- a/system-tests/projects/simple-ct/cypress.config.js
+++ b/system-tests/projects/simple-ct/cypress.config.js
@@ -1,6 +1,8 @@
 module.exports = {
   component: {
     experimentalSingleTabRunMode: true,
+    // In order for Webkit to correctly exit in single tab mode, we have to disable video
+    // https://github.com/cypress-io/cypress/issues/23815
     video: false,
     supportFile: false,
     devServer: {
--- a/npm/react/cypress.config.js
+++ b/npm/react/cypress.config.js
@@ -8,6 +8,7 @@ module.exports = {
   },
   'experimentalFetchPolyfill': true,
   'component': {
+    experimentalSingleTabRunMode: true,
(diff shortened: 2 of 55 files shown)


system-tests/projects/simple-ct/cypress.config.js
```diff
@@ -1,5 +1,7 @@
 module.exports = {
   component: {
+    experimentalSingleTabRunMode: true,
+    video: false,
```
astone123, line 6: I needed to disable video for `simple-ct` so that the `run_ct_spec.js` system test spec passes for Webkit. If we try to take video on Webkit with single tab enabled, the browser just hangs once the specs are finished.

This is related to #23815 which we can look at later down the road of bringing `experimentalSingleTabRunMode` to GA.

mike-plummer: Can we add a comment to that extent here with the bug reference?

astone123: added in https://github.com/cypress-io/cypress/pull/25257/commits/6bcc8da4310030c3bf031728bb7039188e0bd59b

lmiller1990 (approved):
Nice!

Out of curiosity, are `packages/app` and `npm/{react,vue}` are running much faster on CI now? Bonus points for some comparisons. We parallelize, so the improvements won't be *as* obvious as if you run locally w/o parallelizing. 

Also, any change in the time to run system tests? I wouldn't really expect much change on smaller projects.

Code seems :+1:

ZachJW34 (approved):

astone123:
@lmiller1990 here's a summary of what I'm seeing. On the left is the latest `linux-x64` run on this branch, on the right is the `linux-x64` run for https://github.com/cypress-io/cypress/commit/4d6034f524a7c9bf4ec16e33180b7691fa0f0d94 on `develop`.

Some significant gains with `app` component tests:
Screen Shot 2022-12-23 at 9 54 05 AM

For system tests, Chrome and Webkit look like they're completing much faster, the other ones not so much:
Screen Shot 2022-12-23 at 9 53 17 AM

`npm/react` is seeing a pretty huge gain of about ~44 seconds. `npm/vue` and `npm/angular` don't run any tests so we don't see much difference with those runs.
Screen Shot 2022-12-23 at 9 51 41 AM

lmiller1990:
Nice, that's about what I expected. Thanks for this.

PR #25263, opened 2022-12-23
feat: IATR-M1.0 Debug Spec component update

### LINKED ISSUE #24849: [IATR](M1.0) Failed Test Component Enhancements
### Updates to `DebugSpec` and `DebugFailedTest` components

These updates follow the same visual treatment and structure as "Tests for review" in the Cypress Cloud Run Overview, except only show groups that failed (do not show any other statuses). [Example](https://cloud.cypress.io/projects/ypt4pf/runs/41735/overview)

* `DebugSpec`
	* Add "text" icon to the left of the spec path
    * Add the following spec metadata 
		* Test status counts - show single values if 1 group or show range for multiple groups from least value to greatest value
		* Spec duration - show single values if 1 group or show range for multiple groups from least value to greatest value
  		* Group information (browser, OS) 
  			* for single group, show OS icon and name followed by browser icon and name
  			* for multiple groups, show group icon with count of groups, OS icon followed by count of OSes, then overlapping browser icons followed by count of browsers 
	  	* Test type (E2E or component)
	  		* Show different icons for different testing types
* `DebugFailedTest`
	* Additions to test rows
	    * Failed icon to the left
		* Link to test results in cloud if row is clicked
		* Link to artifacts in cloud (hovering on row will show icons with tooltips for artifacts) [screenshot, video, log]
	* When there are multiple groups that failed the test, then there will be rows for each group
		* Clicking on the row with the test name will do nothing for multiple groups, but the group rows will be clickable with hover buttons as described above for 1 group 
		* Each row will have
		  * Group icon followed by group name
		  * OS icon followed by OS name
		  * Browser icon followed by browser name

#### Before:
Screen Shot 2022-12-04 at 4 48 42 PM

#### After:
_Only one group -  Note: The design should not be showing ranges for the test status and duration.  They should only be single values for one group_
Screen Shot 2022-12-04 at 4 48 59 PM

_More than one group.  Still only shows failed tests_
Screen Shot 2022-12-04 at 4 49 17 PM

_Hover states - 1 group_
test-hover

_Hover states - Multiple groups_
Screen Recording 2022-12-04 at 6 41 47 PM

### Props

These components are not using GraphQL fragments to represent their data.  The responsibility to fetch the data is handled by a parent component that is responsible for converting the data from the way it is represented in the Cloud GraphQL fields to the structure needed by these two components.

Currently, the props for `DebugSpec` looks like:

```
interface Spec {
  id: string
  path: string
  fileName: string
  fileExtension: string
}

interface TestResults {
  id: string
  titleParts: string[]
}

defineProps<{
  spec: Spec
  testResults: TestResults[]
  isDisabled?: boolean // used to enable or disabled the Run Failures button
}>()
```

The components should expect data in the following form for the updates:

**TODO: Update this section after GraphQL fields from the Cloud endpoint are finalized**
_Note: this representation is a combination of typescript and GraphQL. It is presented this way to try and concisely show the data format here, but should be adjusted in code to best reuse types_
```
type Spec {
  id: string
  path: string
  fileName: string
  fileExtension: string
  testsPassed: SpecDataAggregate
  testsFailed: SpecDataAggregate
  testsPending: SpecDataAggregate
  specDuration: SpecDataAggregate 
  testingType: 'e2e' | 'component'
  groups: CloudRunGroup[]
}

type TestResults {
  id: string
  titleParts: string[]
  instance: RunInstance[]
}

// below are types from GraphQL 

type SpecDataAggregate {
  min: integer
  max: integer
}

type RunInstance {
  group: CloudRunGroup
  hasStdout: boolean
  stdoutUrl: string
  hasScreenshots: boolean
  screenshotsUrl: string
  hasVideo: boolean
  videoUrl: string
}

type CloudRunGroup {
  id: string
  groupName: string
  os: OperatingSystem
  browser: BrowserInfo
}

```

[Figma designs](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/QkCp2F69X0qyLo7qnYQ67e/Cypress-App?node-id=21180%3A195482&t=mgpzpKNCFo6r0Y3a-1)
[Figma flow that shows hover states for rows](https://www.figma.com/proto/PAOozuCxtfetqP8fUVvg0y/branch/QkCp2F69X0qyLo7qnYQ67e/Cypress-App?node-id=21467%3A202325&scaling=min-zoom&page-id=21010%3A176355&starting-point-node-id=21467%3A202325&show-proto-sidebar=1)

amehta265:
- Closes #24849 
### User facing changelog

This PR aims at updating the Debug Spec component and the Debug Failed component. It accomplishes the following:
It refurbishes the header section of the component to include a text icon and adds a `specsMetadata` section which provides more information about the spec duration, number of groups, browsers, operating systems, and testing types for the tests within the given spec.
In addition, the failed component section is now created to support multiple groupings for a given test and hover states for each test instance.

### Additional details

The most important thing to understand is the data schema passed into debugSpec (the "parent" UI component for this PR). The debugSpec component receives the schema below from `debugMapping.ts`:
```
type CloudDebugSpec = {
  spec: DebugSpecListSpecFragment
  tests: { [thumbprint: string]: DebugSpecListTestsFragment[] }
  groups: { [groupId: string]: DebugSpecListGroupsFragment }
  testingType: TestingTypeEnum
  matchesCurrentTestingType: boolean
  foundLocally: boolean
}
```
- The `spec` is the current current spec that will be rendered by the debug Spec component.
- The `tests` are a mapping of thumbprints to the test represented by the thumbprint: A `thumbprint` is a unique hash associated with each test. If the same test is run for two different `groups` (e.g. Staging and Production) both tests will share the same `thumbprint` but will each have unique `groupId` pointing to the specific group each test is associated with e.g. `{'unique1': [{...Test, grpId: '123'}, {...Test, grpId: '456'}] }`. This is an example of two tests sharing the same thumbprint but each have unique instances as the test was run for different groups. On the other hand a test with a single group would look this: `{'Unique1': [{Test}]}`

- The `groups` are a mapping of `groupID` to their corresponding group e.g. `('123': {Staging}, '456': {Production})`.

- Each test will have at least 1 group associated with it. In this case we render the single group UI. On the other hand if there is more than 1 group then we show the multiple groups UI.
- `StatsMetadata.vue` is a reusable component that helps render a list of items separated by "." It is used in the spec component header section as well as the grouped failed test section.
- `LayeredBrowserIcons.vue` is a component used to render multiple browsers in the header section of the component.
- `GroupedDebugFailedTest.vue` is a reusable component that is rendered every time there are multiple groups associated with the same test.
- `DebugArtifacts.vue` is a the component in charge of showing the hovered state icons

- For a better understanding refer to the images below:

### Steps to test

This issue has 7 component tests: `DebugArtifacts`, `DebugContainer`, `DebugFailedTest`, `DebugSpec`, `GroupedDebugFailedTest`, `LayeredBrowserIcons`, `StatsMetadata`. Checkout to this branch to run the tests from the packages/app/ folder. To run these tests:
1. Change directory to packages/app
2. Run yarn dev
Navigate to Component Testing and run the above tests

I have labeled different sections of the picture with file names to provide better understanding of which component is responsible for what UI.

Single Groups UI:
Screenshot 2022-12-23 at 12 26 41 PM

Multiple Groups UI:
Screenshot 2022-12-23 at 12 27 05 PM

### PR Tasks


packages/app/src/debug/DebugContainer.cy.tsx
```diff
@@ -111,8 +111,18 @@ describe('<DebugContainer />', () => {
 
       const debugMappingArray = specsList({ specs, tests, groups, localSpecs: [], currentTestingType: 'e2e' })
 
+      // [{ specId: 'a1c', id: 'random1' }, { specId: 'a1c', id: 'random2' }], groups: [{ id: 'a', testingType: 'e2e' }]
```
lmiller1990: Is this comment intentional?

packages/app/src/debug/DebugContainer.cy.tsx
```diff
@@ -111,8 +111,18 @@ describe('<DebugContainer />', () => {
...
       expect(debugMappingArray).to.have.length(1)
-      expect(debugMappingArray[0]).to.deep.equal({ spec: { id: 'a1c', groupIds: ['a'] }, tests: [{ specId: 'a1c', id: 'random1' }, { specId: 'a1c', id: 'random2' }], groups: [{ id: 'a', testingType: 'e2e' }], foundLocally: false, testingType: 'e2e', matchesCurrentTestingType: true })
+      expect(debugMappingArray[0]).to.deep.equal(
```
lmiller1990: Can merge these

```js
expect(debugMappingArray).to.have.eql([ .... ])
```

Might be `eq` or `eql` or `equal`, cannot quite remember.

packages/app/src/debug/DebugContainer.cy.tsx
```diff
@@ -150,13 +185,24 @@ describe('<DebugContainer />', () => {
...
       ] as DebugSpecListSpecFragment[]
-      const tests = [{ specId: '123', id: 'random1' }] as DebugSpecListTestsFragment[]
+      const tests = [{ specId: '123', id: 'random1', thumbprint: 'unique1' }] as DebugSpecListTestsFragment[]
       const groups = [{ id: 'a', testingType: 'e2e' }] as DebugSpecListGroupsFragment[]
 
       const debugMappingArray = specsList({ specs, tests, localSpecs: [], currentTestingType: 'e2e', groups })
 
       expect(debugMappingArray).to.have.length(1)
-      expect(debugMappingArray).to.deep.equal([{ spec: { id: '123', groupIds: ['a'] }, tests: [{ specId: '123', id: 'random1' }], groups: [{ id: 'a', testingType: 'e2e' }], foundLocally: false, testingType: 'e2e', matchesCurrentTestingType: true }])
+      expect(debugMappingArray).to.deep.equal(
```
lmiller1990, line 352: Not sure if line 193 (length assertion) serves any purpose here?

packages/app/src/debug/DebugFailedTest.vue
```diff
@@ -1,12 +1,18 @@
...
   >
+    <SolidStatusIcon
+      size="16"
+      status="failed"
+      data-cy="failed-icon"
+      class="isolate"
+    />
     <div
-      v-for="(titlePart, index) in props.failedTestResult.titleParts"
+      v-for="titlePart, index in props.failedTestResult[0].titleParts"
```
lmiller1990: Do we always know that `props.failedTestResult` will contain at least one entry?

packages/app/src/debug/DebugFailedTest.vue
```diff
@@ -20,14 +26,60 @@
...
+import { SolidStatusIcon } from '@cypress-design/vue-statusicon'
+import DebugArtifacts from './DebugArtifacts.vue'
+import GroupedDebugFailedTestVue from './GroupedDebugFailedTest.vue'
+import { computed } from 'vue'
 import type { TestResults } from './DebugSpec.vue'
+import type { CloudRunGroup } from '@packages/data-context/src/gen/graphcache-config.gen'
 
 const props = defineProps<{
-  failedTestResult: TestResults
+  failedTestResult: TestResults[]
```
lmiller1990: Should this be `failedTestsResults`? It's an array, after all.

packages/app/src/debug/DebugFailedTest.vue
```diff
@@ -20,14 +26,60 @@
...
+  groups: CloudRunGroup[]
+  expandable: boolean
 }>()
 
+const debugArtifacts = computed(() => {
+  const runInstance = props.failedTestResult[0].instance
```
lmiller1990: Instead of doing `v-for="titlePart, index in props.failedTestResult[0].titleParts"` above, I wonder if we should use this computed property, something like

```js
const failedTestData = computed(() => {
  const runInstance = props.failedTestResult[0].instance

  return {
    debugArtifacts: [{ icon: ..., text: ... url: ... }],
    result: failedTestResult[0].instance
  }
})
```

Then you don't need the `"titlePart, index in props.failedTestResult[0].titleParts"`, you can just do `titlePart, index in failedTestData.result.titleParts". I like the idea of keeping all the data `<template>` in a single place. WDYT?

Also notice the assumption again that `props.failedTestResult[0]` is always defined. How are we making this assumption? I am really surprised TS is not saying that the element could be undefined.

packages/app/src/debug/DebugSpec.vue
```diff
@@ -78,73 +122,140 @@
...
 })
 
+const uniqueGrpIds: Set<string> = new Set()
+
+/**
+ * Helper function that maps each test's thumprint to all the groups in it
+ * return type: {[thumbprint: string]: CloudRunGroup[]}
+ */
+const groupsPerTest = computed(() => {
+  return Object.keys(props.testResults).reduce((acc, currThumbprint) => {
```
lmiller1990: You can type it here

```suggestion
  return Object.keys(props.testResults).reduce<Record<string, CloudRunGroup[]>>((acc, currThumbprint) => {
```

packages/app/src/debug/DebugSpec.vue
```diff
@@ -78,73 +122,140 @@
...
+    testsFailed: debugResultsCalc(testsFailed.min ?? 0, testsFailed.max ?? 0),
+    testsPending: debugResultsCalc(testsPending.min ?? 0, testsPending.max ?? 0),
+    specDuration: debugResultsCalc(specDuration.min ?? 0, specDuration.max ?? 0, true),
+    groups: props.groups,
+    testingType: props.testingType,
     fullPath: props.spec.fullPath,
   }
 })
 
+const uniqueGrpIds: Set<string> = new Set()
```
lmiller1990: This won't be reactive - is this intended?

You would need `ref` to make it reactive.

lmiller1990:
```suggestion
const uniqueGrpIds = new Set<string>()
```

lmiller1990: Bit more concise.

packages/app/src/debug/DebugSpec.vue
```diff
@@ -78,73 +122,140 @@
...
+    return acc
+  }, {})
+})
+
+/**
+ * Helper function that extracts all the unique groups for a specific spec
+ * return type: CloudRunGroup[]
+ */
+const groupsT = computed(() => {
+  return Object.keys(props.testResults).reduce((acc: CloudRunGroup[], currThumbprint) => {
```
lmiller1990: We do `Object.keys(props.testResults)` twice - should we just have a computed property that does it once so we don't have to iterate the `props.testResults` twice (bit more efficient, but up to you).

packages/app/src/debug/DebugSpec.vue
```diff
@@ -78,73 +122,140 @@
...
+    props.testResults[currThumbprint].map((test) => {
+      const groupId = test.instance?.groupId!
```
lmiller1990: Is `!` necessary? 

This might introduce a bug.

image

`undefined` and `null` are valid in Set. Should we do something like

```js
if (groupId && !uniqueGrpIds.has(groupId)
```

To ensure we do not put `undefined` in the set?

packages/app/src/debug/DebugSpec.vue
```diff
@@ -78,73 +122,140 @@
...
+
+      if (!uniqueGrpIds.has(groupId)) {
+        uniqueGrpIds.add(groupId)
```
lmiller1990: This seems incorrect - `computed` properties should not have side effects, but by mutating the `Set` here, which is effectively a global variable from the point of view of the `computed ` property - we are introducing a side effect.

I think we should rethink what we are doing here. I didn't fully grok why we need the `Set`, but this doesn't seem correct. 

Maybe we can sync on this - either move the `Set` inside the computed property, or can we find a way to not need it at all?

packages/app/src/debug/DebugSpecList.vue
```diff
@@ -1,3 +1,4 @@
+clear
```
lmiller1990: Accidental commit!

packages/app/src/debug/GroupedDebugFailedTest.vue
```diff
@@ -0,0 +1,67 @@
...
+import { computed } from 'vue'
+
+const props = defineProps<{
+  failedTests: TestResults[]
+  groups: CloudRunGroup[]
+}>()
+
+// type: {[groupID: string] : [{icon: string, text: string, url: string | null | undefined}]}
+const debugArtifacts = computed(() => {
+  return props.failedTests.reduce((acc, curr) => {
```
lmiller1990: Sam here, `reduce` is generic, you can pass the type: `reduce<.....>`

packages/app/src/debug/GroupedDebugFailedTest.vue
```diff
@@ -0,0 +1,67 @@
...
+      { icon: 'PLAY', text: 'View Video', url: curr.instance!.videoUrl ?? '' },
+    ]
+
+    return acc
+  }, {})
+})
+
+</script>
+<style scoped>
+[data-cy=grouped-row]:hover .grouped-row-artifacts, [data-cy=grouped-row]:focus-within .grouped-row-artifacts {
```
lmiller1990: I think we should only use `data-cy` for testing purposes. If we need to style this, perhaps add different css class.

packages/app/src/debug/LayeredBrowserIcons.cy.tsx
```diff
@@ -0,0 +1,32 @@
+import LayeredBrowserIcons from './LayeredBrowserIcons.vue'
+
+describe('<LayeredBrowserIcons/>', () => {
+  const browsers = ['CHROME', 'CHROME-CANARY', 'FIREFOX', 'WEBKIT', 'EDGE', 'ELECTRON']
```
lmiller1990: Are these defined somewhere else? Where did you get these definitions from?

packages/app/src/debug/LayeredBrowserIcons.vue
```diff
@@ -0,0 +1,57 @@
...
+  IconBrowserChromeCanary,
+  IconBrowserSafari,
+  IconBrowserMozillaFirefox,
+  IconBrowserEdge,
+  IconBrowserWebkit,
+  IconBrowserElectronLight,
+} from '@cypress-design/vue-icon'
+
+interface LayeredProps {
+  order: string[]
```
lmiller1990: You can type this - it's just just strring, but specific strings.

Here is a neat trick:

https://www.typescriptlang.org/play?#code/MYewdgzgLgBGCGBbAphGBeGBtA5AYQAkAlAeQFkBRHAGhhwDEBJIi+kgDRwF0Z41RIUAFBQAngAdkMAHJIpmMZJAAzOHIhYwAV0QAjZACcuQoQOgwQAawBcMuVh6ZchUpW6nw5sCCgB9K7ayKA4Y2DgAQgAyAIIE3EA

You can make a typesafe array from a `const` with `[number]`. Now the only valid types are the browsers we defined.

packages/app/src/debug/StatsMetadata.vue
```diff
@@ -0,0 +1,176 @@
...
+
+const props = defineProps<MetadataProps>()
+
+interface Metadata {
+  value: number | string | null | SpecDataAggregate | TestingTypeEnum
+  icon: any
+  name: string
+}
+
+type OSType = 'LINUX' | 'APPLE' | 'WINDOWS' | 'GROUP'
```
lmiller1990, line 96: Not essential but I wonder if we can reuse some types, eg https://github.com/cypress-io/cypress/blob/develop/packages/types/src/platform.ts#L2

I think we have several lists of browsers around the code base too, maybe take a look for those?

packages/app/src/debug/StatsMetadata.vue
```diff
@@ -0,0 +1,176 @@
...
+    return props.order.map((status) => ORDER_MAP[status])
+  }
+
+  return []
+})
+
+const arrMapping = computed(() => {
+  const acc: {browser: string[], os: string[], firstBrowser: string, firstOS: string} = { browser: [], os: [], firstBrowser: '', firstOS: '' }
+  const uniqueBrowser: Set<string> = new Set()
+  const uniqueOS: Set<string> = new Set()
```
lmiller1990: If you want an array of unique values you can do it like this

```js
m = new Map()
m.set('foo', { name: 'foo' })
m.set('foo', { name: 'foo' })
m.set('bar', { name: 'bar' })

const arr = Array.from(m.values())
```

This was, you do not need a `Set` and an Array.

lmiller1990: (Or am I missing something? I may have misunderstood what's going on here).

packages/app/src/debug/StatsMetadata.vue
```diff
@@ -0,0 +1,176 @@
...
+
+  props.groups!.forEach((group: CloudRunGroup, index) => {
```
lmiller1990: I think we want to avoid `!` in general, better to have a defensive check.

packages/app/src/debug/StatsMetadata.vue
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// default props if nothing is passed in: need to be figured out
+// change all the group icons e.g. G_OS G_BROWSERS
+// metaDataItems
+// Fix this so that it does not require all the props to be passed all the time
+
+</script>
+<style scoped>
+[data-cy=stats-metadata] li:not(:first-child)::before {
```
lmiller1990: Same here, avoid using `data-cy` for anything other than testing purposes.

lmiller1990 (commented):
Various comments, mainly

1. we can be more type safe, we can pair if you want to go over any of my comments
2. I think we should avoid `!` - better to have a null check and be defensive. More verbose, but unless we are 100000% sure the property cannot be null/undefined, it's not worth the risk in production.

packages/app/src/debug/DebugSpec.vue
```diff
@@ -1,36 +1,80 @@
 <template>
   <div
-    data-cy="debug-spec-item"
-    class="rounded flex flex-col border-t-1px border-x-1px w-full overflow-hidden items-start box-border"
+    data-cy="debug-spec-col"
+    class="grid flex flex-col px-24px gap-24px self-stretch pt-24px"
```
warrensplayer: The `px-24px` is causing it to inset to far into the page
image
If you would like room around the component in the test to keep it away from the edge, then you would need to add that directly int he test.  See `packages/frontend-shared/src/components/Button.cy.tsx` as an example

packages/app/src/debug/DebugFailedTest.vue
amehta265: This is a good point. I will add a `v-if` in `debugSpec.vue` where the `debugFailedTest.vue` will only be rendered if there is at least 1 failed test.

packages/app/src/debug/DebugFailedTest.vue
amehta265: I agree with this. Having all the data consolidated in a single place makes more sense.
As for the `props.failedTestResult[0]`, having made changes to the parent component to only pass props to `debugFailedTest` if there `failedTestResults` contains at least 1 entry, we can be sure here that this will be defined

packages/app/src/debug/DebugArtifacts.cy.tsx
```diff
@@ -0,0 +1,38 @@
...
+    artifactMapping.forEach((artifact) => {
+      cy.mount(() => (
+        <DebugArtifacts icon={artifact.icon} popperText={artifact.text} url={artifact.url}/>
+      ))
+
+      cy.findByTestId(`artifact-for-${artifact.icon}`).should('have.length', 1)
+      cy.findByTestId(`${artifact.icon}-button`).should('be.visible')
+      cy.findByTestId(`artifact-for-${artifact.icon}`).realHover().then(() => {
+        cy.findByTestId('tooltip-content').should('contain.text', `${artifact.text}`)
+      })
```
ZachJW34:
```suggestion
      cy.findByTestId(`artifact-for-${artifact.icon}`).realHover()
      cy.findByTestId('tooltip-content').should('be.visible').contains(artifact.text)
      cy.percySnapshot()
```

Visibility checks are good for the tooltips, helps when grabbing a percy snapshot (which we should do). Also, the `.then` doesn't do anything as the next command is a cy command.

packages/app/src/debug/DebugSpec.vue
amehta265: I don't believe this should be reactive. The only purpose for this variable is to hold the unique group Ids for each spec. There is no UI element tied to this variable that needs to react to any changes made to `uniqueGrpIds`. It is only used in the `groupsT` method to help identify unique group Ids and not add group Ids to an array if they already exist in the `uniqueGrpIds` set.
This is at least my understanding of how reactivity works and why we don't need it here. Would be happy to make this reactive if I am missing something here

packages/app/src/debug/DebugFailedTest.cy.tsx
```diff
@@ -22,17 +52,45 @@ describe('<DebugFailedTest/>', () => {
...
+          groupId: '123',
+        },
+      },
+    ]
+
+    cy.mount(() => (
+      <DebugFailedTest failedTestResult={testResults} groups={[group1, group2]} expandable={true}/>
+    ))
+
+    cy.findAllByTestId('grouped-row').should('have.length', 2)
```
ZachJW34, line 172:
```suggestion
    cy.findAllByTestId('grouped-row').should('have.length', 2)
    cy.percySnapshot()
```

ZachJW34, line 172:
```suggestion
    cy.findAllByTestId('grouped-row').should('have.length', 2)
    cy.percySnapshot()
```

packages/app/src/debug/DebugSpec.vue
amehta265: This is correct. I overlooked `undefined` and `null` being valid for a set.

packages/app/src/debug/GroupedDebugFailedTest.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <div
+    v-for="(group, j) in props.groups"
+    :key="j"
+    class="w-full flex flex-col flex-start justify-center pl-16px border-gray-100 border-1 rounded hover:bg-gray-50"
```
warrensplayer: Should only have rounded corners on the top and bottom of the list.  Also, the lines between items should not be doubled up.

You can use the `divide` classes from Windi.  See https://windicss.org/utilities/borders/divider.html
Example: https://windicss.org/play.html?html=DwEwlgbgfAUAcgewC4AIkAswGcakigYwBsBDLLAXgCIAHAWgA4rYUU8JDTzqAjBAJxABTfinoAmFPwQBXAHbCQzAJJIhAWxQBGYAHpw0GK3acylKn0EixdSdPmKVazeL0GWbA6e4WBw0RJSsgpCSlCqGigAzG6QsLGGMAAiCGiYOCbEZtT0TB6ZXOaW-kEOoSgGYMJ0AJ7MRqye%2BFk%2BEk6ROvpxDcZeLeZt4c4orl2GjU0c-Tm27ZoxY-mLuMsJUEA&css=FBA

Currently showing:
image

Should look like:
image

packages/app/src/debug/LayeredBrowserIcons.cy.tsx
amehta265: I defined the types for these browsers [here](https://github.com/cypress-io/cypress/blob/c43b2d8d3b0d1155324ac521acd7819908e201d6/packages/app/src/debug/LayeredBrowserIcons.vue#L45). I have updated the test file to use the type I have defined above.

packages/app/src/debug/DebugFailedTest.vue
```diff
@@ -20,14 +26,60 @@
...
+      { icon: 'TERMINAL_LOG', text: 'View Log', url: runInstance?.stdoutUrl! },
+      { icon: 'IMAGE_SCREENSHOT', text: 'View Screenshot', url: runInstance?.screenshotsUrl! },
+      { icon: 'PLAY', text: 'View Video', url: runInstance?.videoUrl! },
+    ]
+  )
+})
+
 </script>
+<style scoped>
+[data-cy=test-group]:hover .test-row-artifacts, [data-cy=test-group]:focus-within .test-row-artifacts {
```
ZachJW34, line 147: I don't like how the rendering of the debug-artifacts is dependent upon:
- a data-cy property
- a property that is outside of this component

The second one is a bit harder to deal with as we are reliant on `group` that is often times found in a parent component but where applicable it's good to encapsulate this.

I was expecting to see these debug artifacts for `<DebugFailedTest/> - mounts correctly` test but was unable to. I had to wrap the component in a `<div data-cy="test-group">` to get it to actually appear on hover. I would expect this to be tested within the component and have a percySnapshot attached to it.

packages/app/src/debug/LayeredBrowserIcons.vue
amehta265: I have changed this from being of type `string[]`to be of type `BrowserType[]` which I have already defined [here](https://github.com/cypress-io/cypress/blob/c43b2d8d3b0d1155324ac521acd7819908e201d6/packages/app/src/debug/LayeredBrowserIcons.vue#L45)

packages/graphql/test/stubCloudTypes.ts
```diff
@@ -227,15 +229,59 @@ function addFailedTests (run: CloudRun) {
     specId,
     state: 'FAILED' as const,
     duration: null,
-    instance: null,
+    instance: {
+      __typename: 'CloudRunInstance',
+      "id": "instanceID",
```
ZachJW34: Remove the quotes from the object keys

packages/frontend-shared/src/components/ResultCounts.vue
```diff
@@ -41,33 +46,33 @@ const props = defineProps<ResultCountsProps>()
...
 
 interface Status {
-  value: number|string|null
+  value: number|string|null|SpecDataAggregate
   class: string
   icon: FunctionalComponent<SVGAttributes, {}>
   name: string
 }
 
 const ORDER_MAP: Record<CountType, Status> = {
```
ZachJW34: The order map should be contained within the computed if it is relying on `props`

packages/frontend-shared/src/components/ResultCounts.vue
```diff
@@ -41,33 +46,33 @@ const props = defineProps<ResultCountsProps>()
 type CountType= 'SKIPPED' | 'PENDING' | 'PASSED' | 'FAILED'
 
 interface Status {
-  value: number|string|null
+  value: number|string|null|SpecDataAggregate
```
ZachJW34: Would passing in a `SpecDataAggregate` object make sense here? Rendering `{{ props.value }}` would render out the object.

packages/app/src/debug/DebugArtifacts.vue
```diff
@@ -0,0 +1,49 @@
+<template>
```
warrensplayer: This component can be singular since it is just one thing and could be a bit more descriptive like `DebugArtifactLink` since that is what it is encapsulating

packages/frontend-shared/src/components/ResultCounts.vue
```diff
@@ -21,14 +25,15 @@
 <script lang="ts" setup>
 
 export interface ResultCountsProps {
-  totalPassed: number|string|null
-  totalFailed: number|string|null
-  totalPending: number|string|null
-  totalSkipped: number|string|null
+  totalPassed?: number|string|null|SpecDataAggregate
```
ZachJW34:
```suggestion
  totalPassed?: number | string | null | SpecDataAggregate
```

Why are these optional but then below are marked with the non-null `!` assertion? If you are rendering this component, these should be defined no?

ZachJW34:
```suggestion
  totalPassed?: number | string | null | SpecDataAggregate
```

Why are these optional but then below are marked with the non-null `!` assertion? If you are rendering this component, these should be defined no?

packages/app/src/debug/DebugSpec.vue
```diff
@@ -78,73 +122,140 @@
...
+
 const specData = computed(() => {
+  const testsPassed = props.spec.testsPassed ?? 0
+  const testsFailed = props.spec.testsFailed ?? 0
+  const testsPending = props.spec.testsPending ?? 0
+  const specDuration = props.spec.specDuration ?? 0
+
   return {
-    path: props.spec.path,
+    path: `${props.spec.path}/`,
```
warrensplayer: An extra slash is getting rendered when real data is being used:
image

packages/app/src/debug/DebugSpec.vue
```diff
@@ -1,36 +1,80 @@
...
+                class="inline-flex items-center"
+              >
+                <SpecNameDisplay
+                  :spec-file-name="specData.fileName"
+                  :spec-file-extension="specData.fileExtension"
+                />
+              </span>
+            </div>
+          </div>
+          <ul
```
ZachJW34, line 43: Is the responsiveness of https://cypressio.slack.com/archives/C042Y29PWCD/p1671819234208119?thread_ts=1671818293.872379&cid=C042Y29PWCD going to be addressed?

packages/app/src/debug/StatsMetadata.vue
amehta265: I don't think you are missing anything. This method was made to obtain unique Os and Browser values. Thank you for the `Array.from(m.values())`

warrensplayer (commented):
Still working through other parts of the code, but wanted to share these comments.

The `DebugSpecList` component is missing some items needed to be passed down to the child components.  The `DebugSpecListSpec` fragment is missing the `specDuration` field.  In the `specs` computed value, all the information needed to be mapped in the return value for the `spec` portion are not there.  I think you may be able to just spread the entire `specItem.spec`  now and then override or add any other fields needed. Still looking at what might work.

packages/app/src/debug/DebugFailedTest.vue
amehta265: Thanks for bringing this up. I've add tests and percySnapshot for both the the groups and non-groups UI in this component

packages/frontend-shared/src/components/ResultCounts.vue
amehta265: You are correct. I changed this file during the early stages of this issue when I was expecting the `SpecDataAggregate` type. This is no longer the case as I pass in a string from 'DebugSpec.vue'

packages/frontend-shared/src/components/ResultCounts.vue
amehta265: The props are optional because we do not always pass in 4 values to `Resultcounts` e.g. in `DebugSpec` we only pass in `totalPassed`, `totalFailed`, and `totalPending` but in `RunResult.vue` we pass in all 4 values. Hence it is not always thee case that these will be defined.
In order to avoid setting `undefined` props there is a non null `!` assertion below in `Order_Map`

packages/app/src/debug/DebugSpec.vue
amehta265: After going over this, turns out we don't need this function at all. The `groups` prop holds all the groups associated with a given spec which is what we want.

packages/frontend-shared/src/components/ResultCounts.vue
lmiller1990: I wonder if this component is too generic, it seems to receive too many different values. Also, if we have `?`, shouldn't we not need `null`?

I could be missing something, I feel like this indicates the component is either too broad, or we don't understand out data well enough.

packages/app/src/debug/DebugSpec.vue
marktnoonan: Not sure if the below are included in the existing responsive discussion. It looks like there is intentional truncating of metadata, and perhaps unintentional wrapping of the test title.

Screen Shot 2022-12-29 at 11 09 49 AM
Screen Shot 2022-12-29 at 11 09 37 AM

amehta265: The responsiveness for this UI has been handled in commit [acb411b](https://github.com/cypress-io/cypress/pull/25263/commits/acb411b8c0f1e987ff0e25a4b2e796e17002afc4)

packages/app/src/debug/StatsMetadata.vue
```diff
@@ -0,0 +1,192 @@
...
+      icon: arrMapping.value.browsers,
+      name: 'browser',
+    },
+    'TESTING': {
+      value: props.testing!,
+      icon: TESTING_MAP[props.testing!],
+      name: 'testing-type',
+    },
+    'GROUPS': {
+      value: props.groups!.length > 1 ? `${props.groups!.length} groups` : `${props.groups!.length} group`,
```
marktnoonan: A couple of thoughts here:

- In `arrMapping` we have a check for `if (props.groups)` before code that depends on this being defined. If it will sometimes be `undefined` we might want to handle it similarly. It looks like all uses of `StatsMetadata` we do actually pass a group array (`:groups="[group]"` and ` :groups="Object.values(specData.groups)"`). Similarly for the others, is there a reason these props need to be optional? And if there is, what does the component do in that situation?
-  We can probably handle the pluralization through vue i18n and simplify this a little.

packages/app/src/debug/DebugFailedTest.vue
```diff
@@ -1,33 +1,150 @@
...
+          titlePart.type === 'LAST-1' ? 'shrink-0 whitespace-pre' :
+          titlePart.type === 'LAST-0' ? 'pl-2.5 truncate' : 'px-2.5 truncate'"
+      >
+        {{ titlePart.title }}
       </span>
+    </template>
+    <div
+      v-if="!props.expandable"
+      data-cy="debug-artifacts"
+      class="flex flex-grow space-x-4.5 opacity-0 pr-18px justify-end test-row-artifacts"
```
warrensplayer:
```suggestion
      class="flex flex-grow space-x-4.5 opacity-0 px-18px justify-end test-row-artifacts"
```
This also needs some left padding to keep the artifact links away from test titles that take up all the space.

packages/app/src/debug/DebugArtifactLink.cy.tsx
```diff
@@ -0,0 +1,40 @@
...
+        cy.findByTestId('tooltip-content').should('be.visible').contains(artifact.text)
+      })
+
+      cy.percySnapshot()
+    })
+  })
+
+  it('mounts correctly for all icons together and has correct URLs', () => {
+    cy.mount(() => (
+      <div class="flex flex-grow space-x-4.5 pt-24px justify-center" data-cy='debug-artifacts-all'>
```
marktnoonan, line 28: Noticed an opportunity here, we could grab the very similar looking code [here](https://github.com/cypress-io/cypress/blob/ankit/IATR-M1.0-spec-component-update/packages/app/src/debug/DebugFailedTest.vue#L36-L53) and [here](https://github.com/cypress-io/cypress/blob/ankit/IATR-M1.0-spec-component-update/packages/app/src/debug/GroupedDebugFailedTest.vue#L17-L32) to make a `DebugArtifactLinkList` component or something. We seem to have the idea of this component 3 places if you include this test, and the presence in the test tells me we want to see the artifact links grouped on their own like this.

packages/app/src/debug/GroupedDebugFailedTest.vue
```diff
@@ -0,0 +1,67 @@
...
+      data-cy="test-failed-metadata"
+    >
+      <StatsMetadata
+        :order="['GROUP_NAME', 'OS', 'BROWSER']"
+        :groups="[group]"
+        :group-name="group.groupName!"
+      />
+      <div
+        data-cy="debug-artifacts"
+        class="flex flex-grow space-x-4.5 opacity-0 pr-18px justify-end grouped-row-artifacts"
```
warrensplayer:
```suggestion
        class="flex flex-grow space-x-4.5 opacity-0 px-18px justify-end grouped-row-artifacts"
```
Make sure there is also some left padding to keep the artifact links away from the stats metadata

packages/app/src/debug/DebugArtifactLink.vue
```diff
@@ -0,0 +1,49 @@
+<template>
+  <Tooltip
+    placement="bottom"
+    :data-cy="`artifact-for-${icon}`"
+  >
+    <ExternalLink
+      class="flex h-full w-full items-center justify-center"
+      :data-cy="`${icon}-button`"
+      :href="props.url || '#'"
+      :use-default-hocus="true"
```
marktnoonan, line 10:
```suggestion
      :use-default-hocus="true"
      :aria-label="popperText"
```

This will provide audio description for screenreader users and may reduce the need for data-cy attributes, we can do a `cy.findByRole('link', { name: 'View Log' })`, which is a bit more intuitive in the tests too.

packages/app/src/debug/DebugFailedTest.cy.tsx
```diff
@@ -1,37 +1,195 @@
...
+          hasStdout: false,
+          hasVideo: false,
+        },
+      },
+    ]
+
+    cy.mount(() => (
+      <DebugFailedTest failedTestsResult={testResults} groups={[group1, group2]} expandable={true}/>
+    ))
+
```
warrensplayer, line 169: Add an assertion that the artifact links are not present on the row with the test title since they are shown per group.

packages/app/src/debug/StatsMetadata.vue
warrensplayer: @marktnoonan Ankit and I talked about refactoring this component into its various parts, but decided against it due to time constraints with this issue. I do think it could be broken up and it would reduce the need for a lot of the conditional logic in the future.  

I go agree that using i18n should be used here.

packages/app/src/debug/StatsMetadata.vue
```diff
@@ -0,0 +1,192 @@
...
+      icon: IconTechnologyServer,
+      name: 'group-server',
+    },
+    'G_OS': {
+      value: arrMapping.value.oses.length > 1 ? `${arrMapping.value.oses.length} operating systems` : `${arrMapping.value.oses.length} operating system`,
+      icon: OS_MAP['GROUP'],
+      name: 'operating-system-groups',
+    },
+    'G_BROWSERS': {
+      value: arrMapping.value.browsers.length > 1 ? `${arrMapping.value.browsers.length} browsers` : `${arrMapping.value.browsers.length} browser`,
```
warrensplayer: Use i18n here as well

warrensplayer (commented):
A few small things

packages/app/src/debug/DebugFailedTest.cy.tsx
```diff
@@ -1,37 +1,195 @@
...
+  }
+
+  cy.findByTestId('test-row').children().each((ele, index) => {
+    if (assertionArr[index]) {
+      cy.wrap(ele).should('have.attr', 'data-cy', assertionArr[index].testAttr)
+      if (assertionArr[index].text !== '') {
+        cy.wrap(ele).should('contain.text', assertionArr[index].text)
+      }
+    }
+  })
```
marktnoonan: I think this can be made easier to follow, something like:

```suggestion
  cy.get('[data-cy*=titleParts]').each((ele, index) => {
    if (index >= assertionArr.length) {
      // we don't assert about all the title parts every time
      return
    }

    const { testAttr, text } = assertionArr[index]

    cy.findByTestId(testAttr).should('contain.text', text)
  })
```

This would avoid the `children` and the need to wrap elements. Only drawback is a harmless assertion that the svgs have empty string text.

packages/app/src/debug/DebugFailedTest.cy.tsx
```diff
@@ -1,37 +1,195 @@
...
+  browser: {
+    id: '123',
+    formattedName: 'Electron',
+    formattedNameWithVersion: 'Electron 106',
+  },
+  groupName: 'Production',
+  id: '456',
+}
+
+const rowElementTesting = (testResult: TestResults) => {
```
marktnoonan:
```suggestion
const assertRowContents = (testResult: TestResults) => {
```

We can name this more inline with similar helpers in other tests.

There's a bit of conditional logic in this helper function, it would be good to have some comments explaining the rationale for what is tested in each condition.

packages/app/src/debug/GroupedDebugFailedTest.vue
```diff
@@ -0,0 +1,67 @@
...
+import type { StatsMetadata_GroupsFragment } from '../generated/graphql'
+
+const props = defineProps<{
+  failedTests: TestResults[]
+  groups: StatsMetadata_GroupsFragment[]
+}>()
+
+const debugArtifacts = computed(() => {
+  return props.failedTests.reduce<{[groupID: string]: {icon: string, text: string, url: string | null | undefined }[] }>((acc, curr) => {
+    //TODO Update logic to not rely on defaulting to empty strings
```
marktnoonan: Is this TODO a style preference or is does it functionally make a difference to something? If you don't plan to make this change, we can remove the TODO and maybe leave a comment explaining. If it's an important change, we should make an issue and put the issue number in the TODO.

marktnoonan (approved):
Going to approve as there are only small comments & we are going into a feature branch, I'll leave it up to you about which you implement.

In a general sense I think we can do more to label different content clearly for accessibility purposes and also to make the tests a bit easier to read . But I want to maybe come at that in a separate pass looking at this page as a whole a little further down the line.

packages/app/src/debug/GroupedDebugFailedTest.vue
warrensplayer: @amehta265 I just realized that this is rendering all three buttons every time, but it should only render if the corresponding boolean value in the result is true and the url is present.  I filed a separate issue to resolve this: https://github.com/cypress-io/cypress/issues/25319

warrensplayer (approved):
Looks super!

Issue #4886, opened 2019-07-31
Allow to rerun only failed tests


sylvain-hamel:
### Current behavior:

In the test runner, there is no option to only rerun failed tests. After investigating and fixing a failed test, I'd like the watcher to only rerun failed tests.

### Desired behavior:

Add a rerun "Rerun failed tests only" checkbox to the tests runner UI.

jennifer-shehane:
Some similiar requests:

- https://github.com/cypress-io/cypress/issues/1313
- https://github.com/cypress-io/cypress/issues/1599

Superkunair:
There's a plugin that can execute your failed test
[cypress-plugin-retries](https://github.com/Bkucera/cypress-plugin-retries)

sylvain-hamel:
@jennifer-shehane and @Superkunair you both refer to issues and solutions to handle flaky tests by rerunning them. That's not what I'm looking for. I'm looking for a feature in the Test Runner UI. 

The use case is:

- I change some code and then run the tests using the Test Runner UI.
- Running them takes 10 minutes and 3 tests fail.
- For the next several minutes I'll focus on these 3 tests, changing code and rerunning them many times.
- While I do this, I want the Test Runner UI to only watch/rerun these 3 tests, until they pass.
- When I'm done fixing the issue, I can rerun the whole batch before I commit and push.

So this request is really for an improved developer experience while investigating issues, not as a solution to handle flaky tests.

brian-mann:
What you're referring to is what we call burn-down mode internally.

jennifer-shehane:
Yes, just tagging similar requests - this is a new kind of request which is why the issue was left open.

DavidGWheeler:
> There's a plugin that can execute your failed test
> [cypress-plugin-retries](https://github.com/Bkucera/cypress-plugin-retries)

That is for re-running during execution to account for network issues?  We want a button in the runner marked "re-run failed tests only) to allow for debugging.  Don't confuse the issue.

DavidGWheeler:
> @jennifer-shehane and @Superkunair you both refer to issues and solutions to handle flaky tests by rerunning them. That's not what I'm looking for. I'm looking for a feature in the Test Runner UI.
> 
> The use case is:
> 
> * I change some code and then run the tests using the Test Runner UI.
> * Running them takes 10 minutes and 3 tests fail.
> * For the next several minutes I'll focus on these 3 tests, changing code and rerunning them many times.
[...]

PR #25265, opened 2022-12-23
feat: run filtered debug tests in runner

### LINKED ISSUE #24855: [IATR](M1.0) Test runner filtering
Add ability to filter tests for a spec in the runner to only the specs that failed in the cloud

### Solution for filtering tests in the runner

To filter the tests in a way that will allow for the most flexibility with the developer experience, the following pattern should be followed:
* A `runId` will be passed to the spec runner that will indicate which Cloud run should be used to filter the tests
* The runner will query the cloud with the `runId` and spec path and will get back a list of test names and their status (PASSED, FAILED, PENDING, etc.)
* After the local tests are parsed by mocha, the runner will match up the cloud tests with the parsed local tests.  The matching will occur by looking at the full suite path of the test including the test name.  In case there are multiple tests that match, then include all matching tests.
* All tests except the ones from the cloud that were marked FAILED should be removed from the list of tests for mocha to run. 
* If the test is triggered to run again either by clicking the rerun button or if the file watcher detects a change, then the local spec file should be parsed again and the test filter from the cloud reapplied.  The cloud will not need to be queried again, but use the test status information from the original query.

Background of solution can be found in recorded Zoom found in internal Cypress Slack initiative channel [here](https://cypressio.slack.com/archives/C042Y29PWCD/p1668102562402059)

### Requirements
* Add new parameter to `/specs/runner` url to allow filtering tests by run
	* Only needed for open mode
	* Parameter should be `runId={runId}` to enable filter
* Show a badge in the reporter header (`packages/reporter/src/header/header.tsx`) to indicate the tests are being filtered
	* Should show bug icon
	* Should show ratio of filtered tests being run versus total tests in the spec
	* Should have "x" icon that will close and remove the filter
		* When the filter is removed, the url parameter for the filter should be removed and the page should refresh to run all tests for the given spec 
image
 

* Filter tests in runner to only run failed tests that match from a run in the cloud

ZachJW34:
- Closes #24855

### User facing changelog
na

### Additional details
Allows the filtering of tests in the runner based on the debug run results driven by the `runId` query parameter.

This PR introduces the concept of a `Cypress.testFilter` that will allow the filtering of tests based on their titlePaths. For example, `describe('s1', () => { it('t1', () => {}) })` would result in a titlePath of `s1 t1`. If `Cypress.testFilter = ['s1 t1']`, only this test fill be processed by mocha. Currently it only supports an array of titlePaths to filter by, but I could see this expanding in the future to cb/regex.

If there is an active `Cypress.testFilter`, tests that are not included behave like they never existed (pruned from the parsed suite/tests). If there is a suite/test `.only` that __does not__ contain a matching test it will not influence the tests that matched. If there is a suite/test `.only` that __is__ included, it will behave as normal.

### Steps to test
The query for what tests are filtered is stubbed for now, so you can modify these results inside `coreDataShape.ts`. To apply the filter, click on a spec then add `&runId=123` to the url and refresh the page.

https://user-images.githubusercontent.com/25158820/210434881-febeb57d-4a0d-4320-9c9b-a2a32087266f.mov

### PR Tasks


packages/app/src/runner/unifiedRunner.ts
```diff
@@ -25,26 +42,47 @@ export function useUnifiedRunner () {
...
+
+        if (route.query.runId) {
+          const queryRes = await query.executeQuery({ runId: route.query.runId })
+          const testResults = queryRes.data.value?.currentProject?.testsForRun || []
+
+          const failedTests = (queryRes.data.value?.currentProject?.testsForRun || []).reduce<string[]>((acc, test) => {
+            if (test.status === 'FAILED') acc.push(test.titlePath)
+
+            return acc
+          }, [])
```
warrensplayer: Use filter and map to make it more clear what this is doing
```suggestion
          const failedTests = (testResults)
            .filter((test) => test.status === 'FAILED')
            .map((test) => test.titlePath)
```

cli/types/cypress.d.ts
```diff
@@ -279,6 +279,10 @@ declare namespace Cypress {
     absolute: string // "/Users/janelane/app/cypress/integration/config_passing_spec.js"
     specFilter?: string // optional spec filter used by the user
     specType?: CypressSpecType
+    testFilter?: {
```
warrensplayer: Why are you adding the testFilter fields to the Spec interface?  While the filter is related to a spec, it is not part of one.  Seems like these should remain separate to me.

packages/app/src/store/specs-store.ts
```diff
@@ -1,8 +1,10 @@
 import type { SpecFile } from '@packages/types/src'
 import { defineStore } from 'pinia'
 
+export type SpecWithFilter = SpecFile & { testFilter?: Cypress.Spec['testFilter'] }
```
warrensplayer: I will pose the same question here about combining these two types together. It would seem more straight forward to leave SpecFile alone, and then also store the test filter in the store.  Then in the other places that reference the store, they could just get what they want.

packages/reporter/src/header/header.tsx
```diff
@@ -38,6 +40,7 @@ const Header = observer(({ appState, events = defaultEvents, statsStore }: Repor
       </button>
     </Tooltip>
     <div className='spacer' />
+    {spec.testFilter ? <DebugDismiss testFilter={spec.testFilter} /> : null}
```
warrensplayer: If the type was split, then the header would only need to get passed the filter and not the whole spec.

warrensplayer (commented):
Just a preliminary review.  Most of my questions are around the mashing of the test filter fields into the Spec type. I am guessing you did that in order to get them to ride along with the Spec when it is passed into the runner.  Would it be hard to pass the test filter through separately?

packages/app/src/runner/event-manager.ts
```diff
@@ -235,6 +235,10 @@ export class EventManager {
       this.saveState(state)
     })
 
+    this.reporterBus.on('debug:dismiss', () => {
```
lmiller1990: I wonder if this might be confusing since we also have `cy.debug()`: https://docs.cypress.io/api/commands/debug

packages/app/src/runner/event-manager.ts
```diff
@@ -235,6 +235,10 @@ export class EventManager {
...
+      this.localBus.emit('debug:dismiss')
+    })
```
lmiller1990, line 243: Also there is an ongoing effort to type these, might be good to add it here https://github.com/cypress-io/cypress/blob/develop/packages/app/src/runner/event-manager-types.ts#L34

packages/app/src/runner/useEventManager.ts
```diff
@@ -58,6 +60,17 @@ export function useEventManager () {
...
+
+    eventManager.on('debug:dismiss', () => {
+      const currentRoute = router.currentRoute.value
+
+      const { runId, ...query } = currentRoute.query
+
+      // Delete runId from query which will remove the test filter
+      router.replace({ ...currentRoute, query })
+
+      specStore.setActiveSpec({ ...specStore.activeSpec!, testFilter: undefined })
```
lmiller1990: Should we check if activeSpec is defined?

```js
if (!store.activeSpec) {
  return
}
```

We only want to call this if `activeSpec` is defined?

Also not sure `testFilter` should be grouped with `activeSpec`. It seems related, but separate. I could be convinced otherwise, what's your reasoning here?

As an aside, we've got `specFilter` and `testFilter`  which are named very similar by serve quite a different purpose. I wonder if we want a more specific name here (no suggestions right now, just a thought I had when reading this).

Is this feature only for use with the Cloud? If so, maybe `cloudDebugFilter` or `cloudFailedTests`?

packages/app/src/store/specs-store.ts
lmiller1990: I see the reasoning -  "this filter is only used for the spec" but I also see a case to not nest it *in* the active spec. Could be convinced either way.

I wrote it in my above comment, too, but for clarity - Is this feature only for use with the Cloud? If so, maybe cloudDebugFilter or cloudFailedTests? Perhaps a different name would help.

packages/app/src/store/specs-store.ts
```diff
@@ -1,8 +1,10 @@
...
+
 export interface SpecState {
-  activeSpec: SpecFile | null | undefined
+  activeSpec: SpecWithFilter | undefined | null
```
lmiller1990:
```suggestion
  activeSpec?: SpecWithFilter
```

Not sure if this is valid, but if it is, it would be more concise.

packages/reporter/src/header/header.tsx
```diff
@@ -12,14 +12,16 @@ import { action } from 'mobx'
...
 
 export interface ReporterHeaderProps {
   appState: AppState
   events?: Events
   statsStore: StatsStore
+  spec: Cypress.Spec
 }
 
-const Header = observer(({ appState, events = defaultEvents, statsStore }: ReporterHeaderProps) => (
+const Header = observer(({ appState, events = defaultEvents, statsStore, spec }: ReporterHeaderProps) => (
```
lmiller1990: Not needing to do this now, but we really need to solve App <-> Command Log communication. Although this works for now, I see this becoming a pretty big mess. I don't have any great ideas, but some kind of unified way to bridge the Vue <-> React gap would be great. Let's keep this is mind, definitely open to solutions to this.

packages/reporter/src/main.tsx
```diff
@@ -68,7 +68,7 @@ class Reporter extends Component<SingleReporterProps> {
       <div className={cs(className, 'reporter', {
         'studio-active': appState.studioActive,
       })}>
-        {renderReporterHeader({ appState, statsStore })}
+        {renderReporterHeader({ appState, statsStore, spec: this.props.runnerStore.spec! })}
```
lmiller1990: Is `!` a good idea here? Should we be more defensive? What is `spec` is undefined, does something bad happen?

lmiller1990 (commented):
This wasn't as complex as expected, nice job figuring this out. Just some early bird comments.

packages/reporter/src/main.tsx
ZachJW34: I don't think we ever render the reporter unless we have a spec chosen so this should always be defined. Will clean this up though, depending on how/if I change up the testFilter communication.

packages/app/src/runner/unifiedRunner.ts
ZachJW34: I yo yo between doing the map + filter and the reduce. If I did it the suggested way I'd most likely get a comment to convert it to a reduce for performance benefits which is where I tend to lean nowadays.

cli/types/cypress.d.ts
ZachJW34: I could go both ways. First, it was easy to piggyback off the spec object and a spec is a collection of tests so they are related. On the other hand, I could see a test filter applying across all specs so you could have the separation there. The draft was mostly to get something working so now I'll dig in and see where the concept of this "test filter" makes most sense in terms of where it's stored and communicated between the app and the driver.

packages/app/src/runner/unifiedRunner.ts
lmiller1990: I always use reduce over map+filter for this reason. I could go either way here, I don't really mind, but I like to err on the side of reducing un-necessary loops.

ZachJW34:
@warrensplayer @lmiller1990 haven't addressed all the comments but did a pretty large refactor. Pulled the testFilter off the `Spec` object and I'm deriving the stats for the badge based on the total number of tests we encounter while parsing and the total amount of valid runnables

ZachJW34:
@emilyrohrbough added you as a reviewer as I'm not too experienced with driver code and could use another pair of eyes outside of the CT team!

cli/types/cypress.d.ts
ZachJW34: Separated the testFilter from the spec.

packages/app/src/store/specs-store.ts
ZachJW34: Went with `testFilter` as a top level filter, I can see this being used for more than just the cloud debug feature.

packages/app/src/runner/event-manager.ts
ZachJW34: Made it more explicit in [9074f63](https://github.com/cypress-io/cypress/pull/25265/commits/9074f631a26c33d0f83b298f7ed938dfa8653e3e)

packages/reporter/src/header/header.tsx
ZachJW34: Refactored this component to only render when testFilter is defined but the ratio is now dynamically derived and no longer attached to the filter.

packages/app/src/runner/event-manager.ts
ZachJW34: Done in [9074f63](https://github.com/cypress-io/cypress/pull/25265/commits/9074f631a26c33d0f83b298f7ed938dfa8653e3e)

packages/app/src/runner/useEventManager.ts
ZachJW34: I've refactored this flow. I think `specFilter` isn't used anymore, I believe it was related to run-all-specs.

packages/app/src/store/specs-store.ts
ZachJW34: I've reverted the changes made to this file.

packages/reporter/src/header/header.tsx
ZachJW34: I looked at the M2 work and there is a lot more communication coming up. We do get reactivity from the `mobxrunnerstore`, maybe we can setup a system whereby any change made to a pinia store is relayed to the mobx store and wrapped with `runInAction`

packages/driver/src/cypress.ts
```diff
@@ -194,6 +195,7 @@ class $Cypress {
     this.browser = config.browser
     this.platform = config.platform
     this.testingType = config.testingType
+    this.testFilter = config.testFilter
```
emilyrohrbough: Setting this on the Cypress instance like this prob makes sense to run this faster, but a new run finishes on the cloud with updated results, how will this reflect here? I assume the user would need to navigate away from this spec and come back to get the updated test filter?

Also am I understanding the reason for adding this -- to filter down flakey (or is it failed) cloud test results? If that's the case can we name this accordingly? testFilter sounds like a user-provided filter value and doesn't indicate to me this value is driven from cloud data. Would be helpful for maintenance and change impact if this ever needed updated.

packages/driver/src/cypress/runner.ts
```diff
@@ -492,7 +492,57 @@ const hasOnly = (suite) => {
...
+    rootSuite._onlyTests = onlyTests
+
+    if (!rootSuite.tests.length && !rootSuite.suites.length) {
+      pruneSuite(rootSuite)
+    }
+  }
+}
+
+const normalizeAll = (suite, initialTests = {}, setTestsById, setTests, getRunnableId, getHookId, getOnlyTestId, getOnlySuiteId, createEmptyOnlyTest, incrementFoundTestsBy) => {
+  pruneEmptySuites(suite, incrementFoundTestsBy)
```
emilyrohrbough: To confirm, we will only prune in open mode correct? if so, can we add a comment to clarify and/or can we skip this all together?

packages/driver/src/cypress/runner.ts
```diff
@@ -492,7 +492,57 @@ const hasOnly = (suite) => {
...
+  if (emptySuite.parent) {
+    const parentSuites = emptySuite.parent.suites.filter((suite) => suite !== emptySuite)
+
+    emptySuite.parent.suites = parentSuites
+    emptySuite.parent._onlySuites = emptySuite.parent._onlySuites.filter((suite) => parentSuites.includes(suite))
+    emptySuite.parent = null
+  }
+}
+
+const pruneEmptySuites = (rootSuite, incrementFoundTestsBy) => {
```
emilyrohrbough: Can you add comments on this to describe what this is doing and the expected out come. The runner code is always a lot to process

packages/driver/src/cypress/runner.ts
```diff
@@ -492,7 +492,57 @@ const hasOnly = (suite) => {
...
+  for (const suite of [...rootSuite.suites]) {
```
emilyrohrbough, line 518:
```suggestion
 if (!Cypress.testFilter) {
   return // do nothing if the test filter is not set.
 }

  for (const suite of [...rootSuite.suites]) {
```

packages/driver/src/cypress/runner.ts
```diff
@@ -492,7 +492,57 @@ const hasOnly = (suite) => {
...
+  }
+
+  if (rootSuite.tests.length && Cypress.testFilter) {
+    incrementFoundTestsBy(rootSuite.tests.length)
+
+    const tests: any[] = []
+    const onlyTests: any[] = []
+
+    for (const test of rootSuite.tests) {
+      if (Cypress.testFilter.includes(test.fullTitle())) {
```
emilyrohrbough: We update the test title when a suite or test is skipped due to the browser 
```
describe('suite 1', { browser: '!edge' }, () => {})
it('t1', { browser: 'chrome' }, () => {})
```
We update the fullTitle here: https://github.com/cypress-io/cypress/blob/zachw/iatr-debug-failures-filter/packages/driver/src/cypress/mocha.ts#L97

I would expect a browser-skipped test to appear as skipped if its original title matched the test filter. It would indicate to users they need to run tests in a different browser.

packages/driver/src/cypress/runner.ts
```diff
@@ -492,7 +492,57 @@ const hasOnly = (suite) => {
   )
 }
 
-const normalizeAll = (suite, initialTests = {}, setTestsById, setTests, getRunnableId, getHookId, getOnlyTestId, getOnlySuiteId, createEmptyOnlyTest) => {
+const pruneSuite = (emptySuite) => {
```
emilyrohrbough, line 497: Is the video snippet up-to-date with the latest code? I was noticing the suite title was showing in the Command Log with no associated tests. I would not expect an empty suite to be displayed.

packages/driver/src/cypress/runner.ts
```diff
@@ -1803,6 +1860,10 @@ export default {
 
         return logs.push(attrs)
       },
+
+      getTotalTestsBeforeFilter () {
```
emilyrohrbough: a comment would be helpful here.

packages/reporter/cypress/e2e/header.cy.ts
```diff
@@ -224,4 +238,17 @@ describe('header', () => {
...
+      setupReporter({ testFilter: ['suite 1 test 2'], totalTests: 10 })
+    })
+
+    it('displays debug filter when Cypress.testFilter is defined', () => {
+      cy.spy(runner, 'emit').as('debugDismiss')
+      cy.get('.debug-dismiss').contains(`8 / 10 tests`).click()
+      cy.get('@debugDismiss').should('have.been.called')
+      cy.percySnapshot()
+    })
+  })
```
emilyrohrbough, line 252: can we add more test cases when UI is squished and when there are large test cases to ensure the layout is as expected?

packages/reporter/src/runnables/runnables-store.ts
```diff
@@ -50,6 +50,7 @@ export class RunnablesStore {
    * content: RunnableArray
    */
   @observable runnablesHistory: Record<string, RunnableArray> = {}
+  @observable totalRunnables: number = 0
```
emilyrohrbough: I'm not sure this needs to be observable. I'd expect this to be static.

cli/types/cypress.d.ts
```diff
@@ -281,6 +281,11 @@ declare namespace Cypress {
     specType?: CypressSpecType
   }
 
+  /**
+   * List of spec titlePaths (parent suite names + test name)
```
```suggestion
   * List of full title path (parent suite names + test name) of the Cloud tests that failed that need debugged.
```

cli/types/cypress.d.ts
```diff
@@ -281,6 +281,11 @@ declare namespace Cypress {
...
+   */
+  type TestFilter = string[] | null // could make this a callback/regex in the future
```
emilyrohrbough: Lets rename this to match the intent
```suggestion
  type CloudDebugTestList = string[] | null // could make this a callback/regex in the future
```

cli/types/cypress.d.ts
emilyrohrbough: How will this work with runAll mode? Assume the long-term goal will be to move that out of experimental, should/will the cloud data return the spec name as well?

packages/reporter/src/runnables/runnables-store.ts
```diff
@@ -79,6 +80,7 @@ export class RunnablesStore {
 
     this.hasTests = numTests > 0
     this.hasSingleTest = numTests === 1
+    this.totalRunnables = numTests
```
emilyrohrbough:
```suggestion
    this.totalTests = numTests
```
// nit: runnable could be a suite

packages/app/src/runner/event-manager.ts
```diff
@@ -404,6 +408,9 @@ export class EventManager {
 
           if (!hideCommandLog) {
             this.reporterBus.emit('runnables:ready', runnables)
+            this.Mobx.runInAction(() => {
+              getMobxRunnerStore().setTotalTests(Cypress.runner.getTotalTestsBeforeFilter())
```
emilyrohrbough: In the long term, will the runner store's total Test count be used anywhere else outside of the header ui? If not, why not sent the total count with the runnables on `'runnables:ready'`? The reporter can track `filteredTestCount` and `totalTestCount`.

packages/app/src/store/specs-store.ts
```diff
@@ -4,15 +4,16 @@ import { defineStore } from 'pinia'
 export interface SpecState {
   activeSpec: SpecFile | null | undefined
   specFilter?: string
+  testFilter: Cypress.TestFilter | null
```
emilyrohrbough: why default to null vs undefined?

packages/app/src/runner/event-manager.ts
```diff
@@ -404,6 +408,9 @@ export class EventManager {
 
           if (!hideCommandLog) {
             this.reporterBus.emit('runnables:ready', runnables)
```
emilyrohrbough: Looking through this review, we are splitting the total test count and testFilter on specStore, runnablesStore, etc. Have you considered returning metadata from normalizeAll on either the runnables and/or another return signature to told this information?
```js
const runData = Cypress.runner.normalizeAll(runState.tests, hideCommandLog, specStore.testFilter)

this.reporterBus.emit('runnables:ready', runData)

// reporter
const {
  runnables,
  totalTests,
  totalUnfilteredTests,
  specFilter
} = runData

// do stuff with runnables & totals & spec filter
```

packages/app/cypress/e2e/runner/cloud-debug-filter.cy.ts
```diff
@@ -0,0 +1,85 @@
...
+      ctx.coreData.cloud.testsForRunResults = [{ titlePath: 's1 t4', status: 'FAILED' }]
+    })
+
+    cy.visitApp(`specs/runner?file=cypress/e2e/test.cy.js&runId=123`)
+    cy.waitForSpecToFinish({ passCount: 0, failCount: 1 })
+
+    cy.get('.runnable-title').contains('t4')
+  })
+
+  it('works with skips and onlys', () => {
```
emilyrohrbough, line 62: can we add a test case for browser skip?

packages/driver/src/cypress/runner.ts
emilyrohrbough: Can we add some unit tests around this behavior? https://github.com/cypress-io/cypress/blob/3925ae062a9b9dca4cc9d61391ce5f6ff0525a8b/packages/driver/cypress/e2e/cypress/runner.cy.js

packages/driver/src/cypress.ts
warrensplayer: The idea was to keep the filtering somewhat generic, even though it has only one use case to start.  I would think that the driver would not care why the tests were being filtered, just that it was asked to filter them.  Our thoughts is that this could be expanded in the future to add more ways to filter tests.

packages/driver/src/cypress/runner.ts
warrensplayer: No, that video snippet is older and from an initial implementation. It should now filter out the empty suites.

packages/app/src/runner/event-manager.ts
marktnoonan: This seems tough to answer at this point but @warrensplayer probably has the most insight into future iterations around this page.

packages/driver/src/cypress.ts
emilyrohrbough: I can see that...Though it would typically think filter being a regex or list of regex patterns. Which maybe that could be the case later one. The fact it's stored on the global Cypress instance vs state make it feel like this value will be "static" for this Cypress instance. 

This request https://github.com/cypress-io/cypress/issues/4886, would likely use some of this logic if we chose to support them, which would be driven by user actions in the reporter and not necessarily create a new Cypress instance. I also feel like there was another request recently for selecting a single test in the command long to run in `.only()` mode for quick debugging - just couldn't find the issue.

ZachJW34: I think generating a new Cypress instance would be the cleanest solution as we already have all the wiring for restarting a spec (spec file changed, hitting the refresh button). Might be less performant but restarting has never been too expensive to my knowledge. When we restart, we get another pass at the mocha suite which allows the `testFilter` to get applied again. This testFilter could be modified with each cycle, enabling [cypress-io/cypress/issues/4886](https://github.com/cypress-io/cypress/issues/4886) by listening to an event capturing the run results and changing the filter based on what failed, then triggering a restart.

I could change the filter to be regex/callback rather than string array which would add more utility to it.

packages/app/src/runner/event-manager.ts
ZachJW34: I'll see what I can do here, I agree the info is a bit spread out and would be nice to have it consolidated.

Edit: Refactored in [5e431a0](https://github.com/cypress-io/cypress/pull/25265/commits/5e431a05532e2116e048ad3468ede9aa41ff6eda)!

packages/driver/src/cypress/runner.ts
ZachJW34: Didn't know this! I'll fix this and add some tests around this behavior.

Edit: Done in [5e431a0](https://github.com/cypress-io/cypress/pull/25265/commits/5e431a05532e2116e048ad3468ede9aa41ff6eda)!

packages/driver/src/cypress/runner.ts
ZachJW34: We still want to prune empty suites regardless of a `testFilter` being present or not so that the UI doesn't show the empty suite dropdowns.

ZachJW34: Revisiting this, can I get your opinion on what you think the desired behavior is regarding empty suites e.g. `describe('empty', () => {})`?

Currently they show up in the UI as an empty runnable. Looks a bit wonky but has no side effects. With the way things are wired up now, these suites are pruned (regardless of if there is a test filter or not). What could be confusing is if a developer goes to test some new functionality and starts with `describe.only('new test coverage', () => {})` and they hit save. Since the suite has no children, it will get pruned and the `only` will not apply. This behavior seems pretty confusing to me

packages/driver/src/cypress/runner.ts
ZachJW34: Done in [5e431a0](https://github.com/cypress-io/cypress/pull/25265/commits/5e431a05532e2116e048ad3468ede9aa41ff6eda)!

packages/driver/src/cypress/runner.ts
ZachJW34: I moved this to an earlier check, we won't call `pruneEmptySuites` unless there is a filter set.

packages/driver/src/cypress/runner.ts
ZachJW34: Removed in [5e431a0](https://github.com/cypress-io/cypress/pull/25265/commits/5e431a05532e2116e048ad3468ede9aa41ff6eda)!

packages/reporter/src/runnables/runnables-store.ts
ZachJW34: I was under the assumption that it needs to be `observable` so it triggers a rerender on change. Though it is static once the run is parsed, the store is updated later

packages/reporter/src/runnables/runnables-store.ts
ZachJW34: Done in [5e431a0](https://github.com/cypress-io/cypress/pull/25265/commits/5e431a05532e2116e048ad3468ede9aa41ff6eda)!

packages/app/src/runner/event-manager.ts
ZachJW34: Refactored and grouped the run data on the root runnable to consolidate all of this logic. (Done in [5e431a0](https://github.com/cypress-io/cypress/pull/25265/commits/5e431a05532e2116e048ad3468ede9aa41ff6eda)!

packages/app/cypress/e2e/runner/cloud-debug-filter.cy.ts
ZachJW34: Added in [5e431a0](https://github.com/cypress-io/cypress/pull/25265/commits/5e431a05532e2116e048ad3468ede9aa41ff6eda)!

packages/app/src/store/specs-store.ts
ZachJW34: Reverted to undefined: [5e431a0](https://github.com/cypress-io/cypress/pull/25265/commits/5e431a05532e2116e048ad3468ede9aa41ff6eda)

packages/driver/src/cypress/runner.ts
ZachJW34: Not sure there is a good place to test this outside of the `cy-in-cy` test since the testFilter needs to be set before the spec file is executed.

packages/reporter/cypress/e2e/header.cy.ts
ZachJW34: This test takes a curated `runnables` fixture. I could setup another fixture for filtered runnables but not sure what that would be testing since the filtering logic isn't being executed in this test.

ZachJW34:
@emilyrohrbough I've addressed your feedback. I refactored `testFilter` to no longer be on the `Cypress` global instance but rather explicitly passed to the mocha suite normalization. Setting `Cypress.testFilter` felt heavy handed as it's not in a place for external users to use and try to modify, and we might change the API later if we do want to open it up.

packages/app/src/runner/unifiedRunner.ts
```diff
@@ -35,17 +50,39 @@ export function useUnifiedRunner () {
...
+        if (route.query.runId) {
+          const res = await testsForRunMutation.executeMutation({ runId: route.query.runId as string })
+
+          const testResults = res.data?.testsForRun || []
+
+          const failedTests = testResults.reduce<string[]>((acc, test) => {
+            if (test.status === 'FAILED') acc.push(test.titlePath)
+
+            return acc
+          }, [])
```
mike-plummer: Total nitpick just because I find `reduce` to be harder to parse:

```
const failedTests = testResults
  .filter((test) => test.status === 'FAILED')
  .map((test) => test.titlePath)
```

packages/reporter/src/header/header.tsx
```diff
@@ -38,6 +41,7 @@ const Header = observer(({ appState, events = defaultEvents, statsStore }: Repor
       </button>
     </Tooltip>
     <div className='spacer' />
+    {runnablesStore.testFilter ? <DebugDismiss matched={runnablesStore.totalTests} total={runnablesStore.totalUnfilteredTests} /> : null}
```
mike-plummer: Syntax tip - you can use `&&` instead of the ternary:

`{runnablesStore.testFilter && <DebugDismiss matched={runnablesStore.totalTests} total={runnablesStore.totalUnfilteredTests} />}`

packages/reporter/src/header/DebugDismiss.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import DebuggerIcon from '-!react-svg-loader!@packages/frontend-shared/src/assets/icons/debugger_x16.svg'
+import DeleteIcon from '-!react-svg-loader!@packages/frontend-shared/src/assets/icons/status-failed_x12.svg'
+import events from '../lib/events'
+
+export const DebugDismiss = (props: { matched: number, total: number }) => {
+  return (
+    <button className="debug-dismiss" onClick={() => events.emit('testFilter:cloudDebug:dismiss')}>
```
mike-plummer: `type="button"` to override the default `submit`

packages/data-context/src/data/coreDataShape.ts
```diff
@@ -219,5 +224,25 @@ export function makeCoreData (modeOptions: Partial<AllModeOptions> = {}): CoreDa
...
+        {
+          'titlePath': 's1 t3',
+          'status': 'FAILED',
+        },
+        {
+          'titlePath': 's1 t4',
+          'status': 'FAILED',
+        },
+      ],
+    },
```
mike-plummer: This feels like test data - is this meant to be here?

packages/reporter/src/header/DebugDismiss.tsx
```diff
@@ -0,0 +1,14 @@
...
+      <DebuggerIcon />
+      <span>{props.matched} / {props.total} tests</span>
```
mike-plummer: Pluralization edge case: Only one test in the project and you've filtered to include it - `1 / 1 tests`

packages/reporter/src/header/header.tsx
marktnoonan: Briefly forgot this was React and wondered why we don't do `v-if` - `&&` does seem to be the more conventional react equivalent.

packages/app/src/runner/unifiedRunner.ts
ZachJW34: I've made this simpler and I'm just returning an array of strings (titlepaths). There is no point in the cloud sending tests that didn't fail. No need to loop anymore.

packages/data-context/src/data/coreDataShape.ts
ZachJW34: I've added a comment but this is just some test data as it was never hooked up to the cloud.

packages/driver/src/cypress/runner.ts
emilyrohrbough: I think this is a good start on this behavior. If we prune empty suites, we will end up with different run results in `run` mode which will remove the reporting of these empty suites to the dashboard. Even though there is nothing to "run" it does help visually indicate that a test has been stubbed out.

I do think it's confusing we have differences between empty suites vs empty test behaviors in the reporter visuals, but seems like a small fix in the reporter to show those consistently if we wanted to go that route.

packages/reporter/cypress/e2e/header.cy.ts
emilyrohrbough: I'm looking for coverage on how this wrap if there is a large number of test cases, say `24/200 tests`. We don't have a lot of space in the header today, and wanted to make sure this is flexed/wrap. I know we had issues with this at one point

Screen Shot 2023-01-09 at 8 40 03 AM
ped properly.

packages/driver/src/cypress/runner.ts
emilyrohrbough: We should be able to mock the `testFilter` value in unit tests and ensure the correct runnable results are returned.

packages/app/src/runner/unifiedRunner.ts
```diff
@@ -35,17 +47,34 @@ export function useUnifiedRunner () {
...
-        const activeSpecInSpecsList = specs.value.find((x) => x.relative === queryFile)
+        if (route.query.runId) {
+          const res = await testsForRunMutation.executeMutation({ runId: route.query.runId as string })
 
-        if (activeSpecInSpecsList && specStore.activeSpec?.relative !== activeSpecInSpecsList.relative) {
-          specStore.setActiveSpec(activeSpecInSpecsList)
-        } else if (!activeSpecInSpecsList) {
-          specStore.setActiveSpec(null)
+          // spread to get rid of readonly
+          const failedTests = [...res.data?.testsForRun || []]
```
emilyrohrbough: Any reason the TestFilter type shouldn't be read-only?

packages/app/src/runner/unifiedRunner.ts
```diff
@@ -35,17 +47,34 @@ export function useUnifiedRunner () {
...
+
+          specStore.setTestFilter(failedTests.length ? failedTests : undefined)
+        } else {
+          specStore.setTestFilter(undefined)
         }
+
+        // Either there is a new spec or the runId has changed. Returning a new object
```
emilyrohrbough: when would the runId change? when the cloud data updates? `runId` is a cloud concept correct?

packages/app/src/runner/event-manager.ts
```diff
@@ -393,7 +397,7 @@ export class EventManager {
 
           this.studioStore.initialize(config, runState)
 
-          const runnables = Cypress.runner.normalizeAll(runState.tests, hideCommandLog)
+          const runnables = Cypress.runner.normalizeAll(runState.tests, hideCommandLog, testFilter)
```
emilyrohrbough, line 405: Since this value is set in the specStore, any reason it's not directly pulled from here vs passing this around? `useSpecStore.testFilter`?

packages/driver/src/cypress/runner.ts
```diff
@@ -492,7 +493,78 @@ const hasOnly = (suite) => {
...
+    if (!rootSuite.tests.length && !rootSuite.suites.length) {
+      pruneSuite(rootSuite)
+    }
+  }
+
+  return totalUnfilteredTests
+}
+
+const normalizeAll = (suite, initialTests = {}, testFilter, setTestsById, setTests, getRunnableId, getHookId, getOnlyTestId, getOnlySuiteId, createEmptyOnlyTest) => {
+  let totalUnfilteredTests = 0
```
emilyrohrbough, line 561: should this default value be the total number of tests in the spec, not zero?

emilyrohrbough: Thought looking at the UI, is this meant to represent the totalFilteredTests?

packages/app/src/runner/event-manager.ts
emilyrohrbough: I'm not seeing anything that stands out to make me think this won't work correctly but have you by chance verified these filter changes with a spec that changes domains via `cy.visit()` mid-test? This would re-initialize Cypress mid-test and restore the reporter UI at the correct test index.

packages/driver/src/cypress/runner.ts
```diff
@@ -492,7 +493,78 @@ const hasOnly = (suite) => {
...
+    for (const test of rootSuite.tests) {
+      // Tests/suites of the shape "it('should', { browser: 'chrome' }, ...)" will have their title
+      // updated with a skipped message. Even if the test is skipped we should still show it in the reporter
+      // so we match the "fullTitle" with the skipped message removed.
+      const fullTitle = test.fullTitle().replaceAll(SKIPPED_DUE_TO_BROWSER_MESSAGE, '')
+
+      if (testFilter.includes(fullTitle)) {
+        tests.push(test)
+
+        if (rootSuite._onlyTests.includes(test)) {
```
emilyrohrbough, line 541: we have quite a bit of logic to filter the "only" suites/tests later on in the normalize func. Is this needed here or can we rely on the existing filtering logic? if it is needed, do you know whats missing in the existing logic?

packages/app/src/runner/unifiedRunner.ts
mike-plummer: The data management of the current `runId` is still in-progress, but the end goal is that a user could be looking at a given run when a new run is kicked off by CI or another user. A user can move to view the newer run using an in-UI transition which will update the `runId` and kick off new queries

packages/driver/src/cypress/runner.ts
mike-plummer: This preemptively filters out any runnables marked with `.only` that don't match the test filter before we get too far into parsing. We want the filter to take precedence, and without this we could end up including some tests that don't match the filter if they're marked with `.only`.

packages/driver/src/cypress/runner.ts
mike-plummer: I believe Zach structured it this way to simplify accounting for filtered nested runnables (eg `.only` and "skipped due to browser" conditions). Was easier to sum up matching elements than it was to deduct non-matches

The variable naming appears correct to me - it's used for the "total" value in the output of `Header.tsx` (`x/y tests`)

packages/driver/src/cypress/runner.ts
emilyrohrbough: We are deleting the runnable references that don't match, I am surprised we are seeing tests incorrectly being included later on...

packages/app/src/runner/unifiedRunner.ts
emilyrohrbough: can we document this information?

packages/reporter/cypress/e2e/header.cy.ts
mike-plummer: [Added a test ](https://github.com/cypress-io/cypress/pull/25265/commits/0fd3283ca3d99e60e603ddfdd05972baa34368ff#diff-748a31f95b1735471223b4b8b71c9b88b47952ab6061488314743f6d45268300R35)with snapshots for 4 different reporter widths to demonstrate wrapping

For reference, here's the wrapping behavior:

https://user-images.githubusercontent.com/11482842/211831300-94f2bc45-f25f-4957-af13-6d326fa946a4.mov

mike-plummer:
@warrensplayer @marktnoonan @emilyrohrbough All feedback thus far has been addressed save adding some [unit tests for `runner.ts`](https://github.com/cypress-io/cypress/pull/25265#discussion_r1059416455). I will be looking at that this morning, but in the interim let me know if I missed anything or if there's any other feedback

mike-plummer:
I'm really struggling to find a way to do more granular unit tests for the `runner.ts` stuff, but I'm not very familiar with this area so I may be overlooking something obvious. Any approach I take to mock out the testFilter from within a test doesn't apply in time or results in a loop of Cypress re-initializing. Any ideas are welcome, otherwise I may look at beefing up the cy-in-cy tests as a follow-on

warrensplayer:
Merging this to the feature branch.  Added new issue to address additional testing here: https://github.com/cypress-io/cypress/issues/25485

PR #25274, opened 2022-12-26
chore(electron): when in dev, build the arm64 electron binary when cpu arch supports it

brian-mann:
- Closes 

### Additional details
- When we're in development and building the electron binary post `yarn` we need to detect the proper `arch` to build the electron binary for
- Previously it would naively us `os.arch()`, which returns the **node arch**, which could be incorrectly `x86`
- This PR updates the code to properly detect the actual system arch to figure out if `arm64` is capable, and uses that to build the right electron binary `arch`
- This PR only updates the logic to verify when to rebuild the electron binary. If an existing `x64_64` binary arch is detected, it'll automatically rebuild it. Users should receive the correct binary the next time they pull and run `yarn`

### DIFF
--- a/packages/electron/lib/install.js
+++ b/packages/electron/lib/install.js
@@ -2,8 +2,11 @@
 const _ = require('lodash')
 const os = require('os')
 const path = require('path')
+const systeminformation = require('systeminformation')
+const execa = require('execa')
+
 const paths = require('./paths')
-const log = require('debug')('cypress:electron')
+const debug = require('debug')('cypress:electron')
 const fs = require('fs-extra')
 const crypto = require('crypto')
 const { flipFuses, FuseVersion, FuseV1Options } = require('@electron/fuses')
@@ -27,9 +30,7 @@ module.exports = {
     return require('@packages/icons')
   },
 
-  checkCurrentVersion () {
-    const pathToVersion = paths.getPathToVersion()
-
+  checkCurrentVersion (pathToVersion) {
     // read in the version file
     return fs.readFile(pathToVersion, 'utf8')
     .then((str) => {
@@ -55,6 +56,7 @@ module.exports = {
   },
 
   checkIconVersion () {
+    // TODO: this seems wrong, it's hard coding the check only for OSX and not windows or linux (!?)
     const mainIconsPath = this.icons().getPathToIcon('cypress.icns')
     const cachedIconsPath = path.join(__dirname, '../dist/Cypress/Cypress.app/Contents/Resources/electron.icns')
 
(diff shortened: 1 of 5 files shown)


packages/electron/lib/install.js
```diff
@@ -94,42 +118,59 @@ module.exports = {
     })
   },
 
+  async getRealArch (platform, arch) {
+    if (platform === 'darwin' && arch === 'x64') {
+      // see this comment for explanation of x64 -> arm64 translation
+      // https://github.com/cypress-io/cypress/pull/25014/files#diff-85c4db7620ed2731baf5669a9c9993e61e620693a008199ca7c584e621b6a1fdR11
+      return systeminformation.cpu()
```
brian-mann, line 125: @flotwig using the updated `systeminformation` npm package here instead of `systemctl.proc_translated` which is what we do in the CLI module

packages/electron/lib/install.js
```diff
@@ -66,8 +68,30 @@ module.exports = {
...
-    return fs.stat(paths.getPathToExec())
+  checkExecExistence (pathToExec) {
+    return fs.stat(pathToExec)
+  },
+
+  async checkBinaryArchCpuArch (pathToExec, platform, arch) {
+    if (platform === 'darwin' && arch === 'x64') {
+      return Promise.all([
+        // get the current arch of the binary
+        execa('lipo', ['-archs', pathToExec])
```
brian-mann, line 79: @flotwig using `lipo` here instead of `file` to more easily parse the output. I considered that `lipo` can return a multiple space separated values such as `x64_86 arm64` but considering we don't build a universal binary, and instead only build a single arch, I didn't think it was necessary since only a single value will be returned

packages/electron/lib/install.js
```diff
@@ -55,6 +56,7 @@ module.exports = {
   },
 
   checkIconVersion () {
+    // TODO: this seems wrong, it's hard coding the check only for OSX and not windows or linux (!?)
```
brian-mann, line 59: @flotwig might want to follow up with @elevatebart who was the original author of this code below. it seems wrong and not cross platform safe. I'm not sure what would even happen on `linux` or `win32`. My guess is that it would **always** rebuild the electron binary since this check would presumably fail every single time. If that's the case then we're rebuilding the electron binary unnecessarily every single time

packages/electron/lib/install.js
lmiller1990: I played with `lipo` a bit just now, seems like a good solution. Didn't know about this util.

lmiller1990 (approved):
Make sense, `lipo` seems like a good solution, verified it is on my M1 mac, and looks like it's been around for quite a while.

I wonder how long `-archs` has been around for. This is the one thing I wish we could definitively verify.

packages/electron/lib/install.js
lmiller1990: I wonder when `-archs` was introduced. `lipo` has been around for a while from what I can see, unsure on `-archs`.

packages/electron/lib/install.js
flotwig: nice, I didn't know systeminformation could do this.

packages/electron/lib/install.js
flotwig: Hm, this is definitely incorrect, but I think it's actually doing the opposite - it's not rebuilding on linux/win32 if the icons change, only on macOS. If I run `yarn build` in `packages/electron` I get no output about this error being thrown or electron reinstalling. Given that we very rarely change icons and this is a dev-only issue, I'm ok with leaving this TODO as a note without creating an issue.

flotwig (approved):

PR #25285, opened 2022-12-27
fix: navbar border appearance at various zoom levels

### LINKED ISSUE #25284: White border on nav at certain zoom levels in Chrome
### Current behavior

In Chrome at 125% and 175% zoom a distracting white border appears on the left nav when on the Test Runner page:

incorrect border

This is caused by an apparent bug in Chrome as it happens in no other browser. At these zoom levels, Chrome seems to make too much space for the 1px border. 

### Desired behavior

The border should be the usual dark gray at all zoom levels.

### Test code to reproduce

Just change zoom levels with `Cmd + +/-` on the spec runner page (where it is noticeable).

The issue seems to be related to the use of `grid-template-columns: auto` and can be "fixed" by hardcoding a value greater than the true nav sidebar value for that part of the grid. It can also be fixed by moving the border to the right of the nav bar instead, which has the same visual appearance but doesn't show the bug.

### Cypress Version

10.0.0-12.2.0

### Node version

na

### Operating System

macOs

### Debug Logs

_No response_

### Other

_No response_

marktnoonan:
- Closes #25284

This is a small visual bug, but I zoom in to Cypress a lot for demos and presentations, and it has always bothered me a little bit. Today it occurred to me to take another look at it and the fix is small.

incorrect border

### User facing changelog

Chrome users will no longer see a white border on the nav bar at the specific zoom levels.

### Additional details

The source of this really seems to be a Chrome bug with grid and borders. At these zoom levels, the equivalent of 2px is available between elements when only 1px is required.

### Steps to test

Change zoom levels with Cmd + +/- on the spec runner page (where it is noticeable due to the dark command log/specs list).

Before, with white border appearing

https://user-images.githubusercontent.com/8340719/209731555-e387029d-9ab8-439f-9397-230ca33cbce5.mov

After, zooming is fine:

https://user-images.githubusercontent.com/8340719/209731560-0eaaef94-5074-4ba0-abb3-e18220d007e6.mov

### PR Tasks

- [na] Have tests been added/updated? - not really possible to test for this, though some percy tests may reflect the change of border location

### DIFF
--- a/packages/app/src/runner/screenshot/AdjustRunnerStyleDuringScreenshot.vue
+++ b/packages/app/src/runner/screenshot/AdjustRunnerStyleDuringScreenshot.vue
@@ -1,8 +1,5 @@
 <template>
-  <div
-    :style="style"
-    :class="screenshotStore.isScreenshotting || isRunMode ? '' : 'border-l-1'"
-  >
+  <div :style="style">
     <slot />
   </div>
 </template>
--- a/packages/app/src/navigation/SidebarNavigation.vue
+++ b/packages/app/src/navigation/SidebarNavigation.vue
@@ -2,7 +2,7 @@
   <HideDuringScreenshot
     id="sidebar"
     data-cy="sidebar"
-    class="flex flex-col bg-gray-1000 transition-all duration-300 relative"
+    class="flex flex-col bg-gray-1000 border-gray-900 border-r-1 transition-all duration-300 relative"
     :class="isNavBarExpanded ? 'w-248px' : 'w-64px'"
   >
     <button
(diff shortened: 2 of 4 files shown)


packages/app/src/runner/screenshot/AdjustRunnerStyleDuringScreenshot.vue
```diff
@@ -1,8 +1,5 @@
 <template>
-  <div
-    :style="style"
-    :class="screenshotStore.isScreenshotting || isRunMode ? '' : 'border-l-1'"
```
marktnoonan, line 4: Nice to clean this up, since the navbar is already hidden during screenshots.

lmiller1990 (approved):

mike-plummer (approved):

PR #25286, opened 2022-12-27
feat: IATR non-failure run statuses

### LINKED ISSUE #24848: [IATR](M1.0) Non-failure Run Statuses
Add support for all the other run statuses other than FAILED

* Passed
    * Message: All green!
    * Show failures: N/A
Screen Shot 2022-12-04 at 7 29 59 PM

* Errored
	* Message: <Error message>
	* Show failures: Yes
Screen Shot 2022-12-04 at 7 30 54 PM

* No Tests
	* Show banner
		* Title: Incomplete
		* Description: This run has no tests
Screen Shot 2022-12-04 at 7 30 38 PM

* Timed Out
	* Show banner with error message from Cloud
	* Show failures: Yes
Screen Shot 2022-12-04 at 7 31 00 PM

* Over Limit
	* Show "Upgrade plan" CTA button if admin
	* Show "Contact admin" message if not admin
	* Show failures: No
	* Text and links for CTAs will be provided in the `runByNumber` field.  Add on UTM parameters found [here](https://docs.google.com/spreadsheets/d/1Cs933TLXcS63PmMlXA4wKLJtQSFH90dProvzI9oyEB4/edit#gid=0)

Screen Shot 2022-12-04 at 7 30 26 PM

Screen Shot 2022-12-04 at 7 30 31 PM

* Canceled
	* Message: This run was canceled
	* Show failures: Yes
Screen Shot 2022-12-04 at 7 30 18 PM

* Hidden runs (can be for any status)
	* Show the same messages as for "Overlimit" above
    * This state should be shown if the `isHiddenByUsageLimits` value is true on the CloudRun, regardless of the overall status of the run

[Figma design](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/QkCp2F69X0qyLo7qnYQ67e/Cypress-App?node-id=21180%3A190683&t=mgpzpKNCFo6r0Y3a-1)

astone123:
- Closes #24848

### User facing changelog

n/a, merging to a feature branch

### Additional details

For some run statuses on the Debug tab, we need to display specific information, sometimes in addition to displaying the failing tests for debugging. This PR adds UI for the following run states:

- `CANCELLED`
- `PASSED`
- `ERRORED`
- `NOTESTS`
- `TIMEDOUT`
- `OVERLIMIT`

### Steps to test

Ideally, you should create a run on your local Cloud instance for each of the statuses listed above. I was only able to actually re-create a few of them end to end (I couldn't re-create `TIMEDOUT` or `OVERLIMIT`). I would suggest checking out the tests that I added [here](https://github.com/cypress-io/cypress/compare/feature/IATR-M0...24848-non-failure-run-statuses?expand=1#diff-f97e5572c31990dca87154da4f715e65acc797e63c864d5996134b33fbd5694c) and viewing the Percy snapshots that were generated from those tests and comparing those to the [Figma](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/QkCp2F69X0qyLo7qnYQ67e/Cypress-App?node-id=21180%3A190683&t=mgpzpKNCFo6r0Y3a-1).

There's an issue with the tests for `PASSED` and `OVERLIMIT` where in the test they appear towards the top of the viewport instead of further down where they should be. However, if you actually view these states on the Debug tab, they show up correctly. There should be a `flex-grow` class on these that create the correct height for the elements but it's not working in the tests.

See Percy diff

### PR Tasks


packages/app/src/debug/DebugPageDetails.vue
```diff
@@ -0,0 +1,67 @@
...
+
+const props = defineProps<{
+  status: CloudRunStatus
+  canceledAt: string | null
+  canceledByFullName?: string | null
+  canceledByEmail?: string | null
+  isHiddenByUsageLimits: boolean
+  overLimitActionType: OverLimitActionTypeEnum
+  overLimitActionUrl: string
+  specs: readonly any[]
```
astone123: I couldn't get the types to work for this. I wanted it to be of type `CloudSpecRun[]` but I wasn't sure where to import this type from

packages/app/src/debug/DebugPageDetails.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <div
+    class="flex flex-col w-full px-24px justify-center items-center align-middle"
+    :class="{'flex-grow': ['PASSED', 'OVERLIMIT'].includes(status)}"
```
astone123, line 4: This class doesn't seem to do anything in the tests, but it does in the actual debug tab view 🤷🏻‍♂️

packages/app/src/debug/DebugPageDetails.vue
```diff
@@ -0,0 +1,67 @@
...
+import DebugErrored from './DebugErrored.vue'
+import DebugNoTests from './DebugNoTests.vue'
+import DebugTimedout from './DebugTimedout.vue'
+import DebugOverLimit from './DebugOverLimit.vue'
+import type { CloudCiBuildInfo } from '@packages/data-context/src/gen/graphcache-config.gen'
+
+const props = defineProps<{
+  status: CloudRunStatus
+  canceledAt: string | null
+  canceledByFullName?: string | null
```
astone123: If anyone has tips on how to deal with this, I'd be happy to refactor. I guess when something is nullable in GQL, it can be either `TheType | undefined | null`? This leads to weird stuff with TypeScript where it's not enough to just make the prop optional, you also have to specify that it could be `null`. It's ugly

packages/app/src/debug/DebugCancelledAlert.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+  <Alert
+    :title="t('debugPage.manuallyCanceled')"
```
astone123: Just realized I need to decide between `cancelled`/`canceled`

packages/app/src/debug/DebugCancelledAlert.vue
```diff
@@ -0,0 +1,56 @@
...
+          data-cy="canceled-by-user-avatar"
+        />
+        <div>{{ canceledByFullName }}</div>
+      </div>
+      <div
+        v-if="canceledAt"
+        class="rounded-full font-semibold bg-orange-500 h-3px mx-10px w-3px"
+      />
+      <div v-if="canceledAt">
+        {{ dayjs(canceledAt).utc().format('MMM D, YYYY h:mm A UTC') }}
```
lmiller1990: Just checking, what is the format of `canceledAt`? Do we really want to show the time this was canceled in UTC? Shouldn't we show it in their locale?

packages/app/src/debug/DebugCancelledAlert.vue
```diff
@@ -0,0 +1,56 @@
...
+import { useI18n } from '@cy/i18n'
+
+const { t } = useI18n()
+
+dayjs.extend(utc)
+
+defineProps<{
+  totalSpecs: number
+  totalSkippedSpecs: number
+  canceledAt?: string | null
```
lmiller1990: Do we need `null` if we also have `?`?

packages/app/src/debug/DebugContainer.vue
```diff
@@ -1,16 +1,30 @@
 <template>
-  <div>
+  <div class="h-full">
     <DebugError
       v-if="showError"
     />
     <div
-      v-else-if="loginConnectStore.user.isLoggedIn && loginConnectStore.project.isProjectConnected && run"
+      v-else-if="loginConnectStore.user.isLoggedIn && loginConnectStore.project.isProjectConnected && run && run.status"
```
lmiller1990: Should we name this, something like `hasCloudRun`?

packages/app/src/debug/DebugPageDetails.vue
lmiller1990: Should we just make this a GraphQL component that has a fragment? It seems we are passing quite a bit of data to it, anyway.

packages/app/src/debug/DebugPageDetails.vue
```diff
@@ -0,0 +1,67 @@
...
+<script lang="ts" setup>
+import { computed } from '@vue/reactivity'
+import type { CloudRunStatus, OverLimitActionTypeEnum } from '../generated/graphql'
+import DebugCancelledAlert from './DebugCancelledAlert.vue'
+import DebugPassed from './DebugPassed.vue'
+import DebugErrored from './DebugErrored.vue'
+import DebugNoTests from './DebugNoTests.vue'
+import DebugTimedout from './DebugTimedout.vue'
+import DebugOverLimit from './DebugOverLimit.vue'
+import type { CloudCiBuildInfo } from '@packages/data-context/src/gen/graphcache-config.gen'
```
lmiller1990: I wonder if `packages/graphql/src/gen/cloud-source-types.gen.ts` is a more correct place to import this from (note it's from `graphql`, not `data-context` - I feel like the frontend code should **not** import anything from `data-context`, even just a type, since that's really a server package).

packages/app/src/debug/DebugContainer.vue
```diff
@@ -1,16 +1,30 @@
...
+      class="flex flex-col h-full"
     >
       <DebugPageHeader
         :gql="run"
         :commits-ahead="0"
       />
+      <DebugPageDetails
```
lmiller1990, line 14: Should `DebugPageDetails` declare a `fragment` and just receive a single prop (`gql`)? It would save passing a bunch of individual properties, and you wouldn't need to deal with the `:? null | undefined` issue around types.

packages/app/src/debug/DebugPageDetails.vue
lmiller1990: I spiked into fixing this but it's going to be a chunk of work. There's a bunch of options we can configure: https://the-guild.dev/graphql/codegen/plugins/typescript/typescript-operations

The ones we need to figure out and standardize:

- `maybeValue` (default is `T | null`, maybe we want `undefined`?)
- we should do `immutableTypes: false`. This will stop generating `readonly` which is a bit annoying, even if it's more technically correct

An alternative in the short term might be  `type Maybe<T> = T | undefined | null`. :man_shrugging:

packages/app/src/debug/DebugPageDetails.vue
lmiller1990: I think it would be

```ts
import type { CloudRunStatus, DebugPageFragment, OverLimitActionTypeEnum } from '../generated/graphql'

specs: DebugPageFragment[]
```

Tried it, looks like `CloudRunStatus` is 

```ts
export type CloudRunStatus =
  | 'CANCELLED'
  | 'ERRORED'
  | 'FAILED'
  | 'NOTESTS'
  | 'OVERLIMIT'
  | 'PASSED'
  | 'RUNNING'
  | 'TIMEDOUT';
```

Does not include `UNCLAIMED` - maybe my Cloud repo is out of date, or is this type incorrect?

lmiller1990: Maybe we make this have a GraphQL fragment and then get the types that way.

packages/app/src/debug/DebugPageDetails.vue
```diff
@@ -0,0 +1,67 @@
...
+  isHiddenByUsageLimits: boolean
+  overLimitActionType: OverLimitActionTypeEnum
+  overLimitActionUrl: string
+  specs: readonly any[]
+  ci?: Partial<CloudCiBuildInfo>
+  errors: readonly string[]
+}>()
+
+const totalSkippedSpecs = computed(() => {
+  return props.specs.filter((spec) => spec.status === 'UNCLAIMED' || spec.status === 'RUNNING').length
```
lmiller1990, line 71: Unrelated but what does `UNCLAIMED` mean in this context?

packages/app/src/debug/DebugPageDetails.vue
```diff
@@ -0,0 +1,67 @@
...
+    <DebugErrored
+      v-else-if="status === 'ERRORED'"
+      :errors="errors"
+      :total-specs="totalSpecs"
+      :total-skipped-specs="totalSkippedSpecs"
+    />
+    <DebugNoTests v-else-if="status === 'NOTESTS'" />
+    <DebugTimedout
+      v-else-if="status === 'TIMEDOUT'"
+      :total-specs="totalSpecs"
```
lmiller1990: I think we can just pass `props.specs.length`, no need for the computed, since we are not deriving anything. `props` are reactive out of the box iirc (like React).

packages/graphql/test/stubCloudTypes.ts
```diff
@@ -280,6 +280,30 @@ export function createCloudProjectSpecResult (config: Partial<CloudProjectSpec>)
   return indexNode(specResult)
 }
 
+const skippedSpecs = [{
```
lmiller1990: Is there a type definition we can add for this?

lmiller1990 (commented):
Some comments/questions.

packages/app/src/debug/DebugCancelledAlert.vue
astone123: `canceledAt` is a date string like this `2019-01-25T02:00:00.000Z`. Not sure, I was just going off of the Figma design which specifies UTC at the end of the timestamp [here](https://www.figma.com/file/PAOozuCxtfetqP8fUVvg0y/branch/QkCp2F69X0qyLo7qnYQ67e/Cypress-App?node-id=21215%3A209340&t=ZV0S7Uun2UNc4K3g-0)

packages/app/src/debug/DebugCancelledAlert.vue
astone123: Yeah, we do because `canceledAt` can **also** be `null` as well as undefined 😕

packages/app/src/debug/DebugPageDetails.vue
astone123: So there are two different status types:

`CloudRunStatus` and `CloudSpecStatus`

`CloudRunStatus` is the status of the overall run, while `CloudSpecStatus` is the status of an individual spec that was in the run. They overlap quite a bit so it can be confusing. `spec.status` is of type `CloudSpecStatus`, which **does** have `UNCLAIMED`.

packages/app/src/debug/DebugPageDetails.vue
astone123: I believe it's related to parallelization, where specs are "claimed" by processes. If a spec hasn't been claimed yet then it is `UNCLAIMED`. 

So, if the run has finished, any specs that are still `UNCLAIMED` were skipped, likewise for any that are still `RUNNING`.

packages/app/src/debug/DebugContainer.vue
astone123: I'd like to keep it this way because when I move this into a computed value, the types in the template get all messed up. 

I guess because it doesn't realize that we're checking for the existence of `run` and `run.status` in the computed function? Because it still thinks that `run` could be null inside of the `v-else-if` block even though at that point we've determined that `run` and `run.status` are not null 😕

packages/app/src/debug/DebugPageDetails.vue
astone123: I can't import from `@packages/graphql` because it's restricted by ESLint. I ended up declaring a fragment for it and using that type instead.

packages/app/src/debug/DebugContainer.vue
astone123: I don't have a firm stance on this but I'm used to being explicit about what we're passing down to the component. This would solve the `null | undefined` issue but only for `DebugPageDetails`. Same issue would apply with all of the individual components that we render inside of `DebugPageDetails`. 

Not crazy about the idea of passing one giant `gql` prop down rather than breaking them out into what info we actually need for the component

packages/app/src/debug/DebugCancelledAlert.vue
astone123: Decided on `Cancelled` based on this [styleguide](https://docs.google.com/document/d/1pOZ0Ntm4b1Eev_iY5vLBPhOxx66Z8Njrvvg6WMCvWRI/edit#heading=h.daxa2xpyrsop)

packages/app/src/debug/DebugCancelledAlert.vue
```diff
@@ -0,0 +1,56 @@
...
+import UserAvatar from '@packages/frontend-shared/src/gql-components/topnav/UserAvatar.vue'
+import ErrorOutlineIcon from '~icons/cy/status-errored-outline_x16.svg'
+import Alert from '@cy/components/Alert.vue'
+import dayjs from 'dayjs'
+import utc from 'dayjs/plugin/utc'
+import { useI18n } from '@cy/i18n'
+
+const { t } = useI18n()
+
+dayjs.extend(utc)
```
mike-plummer: This feels odd to have inside a component - there's similar logic in a util under `packages/app/src/runs/utils/day.js`, could we expand and reuse that?

packages/app/src/debug/DebugCancelledAlert.vue
```diff
@@ -0,0 +1,56 @@
...
+        v-if="cancelledByEmail && cancelledByFullName"
+        class="flex items-center"
+      >
+        <UserAvatar
+          :email="cancelledByEmail"
+          class="h-20px mr-7px w-20px"
+          data-cy="cancelled-by-user-avatar"
+        />
+        <div>{{ cancelledByFullName }}</div>
+      </div>
```
mike-plummer: Rather than doing the two side-by-side `v-if` with the same conditional could we wrap with a single conditional `template`?

```
<template v-if="cancelledByEmail && cancelledByFullName">
   <div ... />
   <div ... />
</template>
```

packages/app/src/debug/DebugContainer.vue
mike-plummer: We could at least combine the last two conditions using a null-safe traversal `&& run?.status`, right?

packages/app/src/debug/DebugContainer.vue
mike-plummer: I also like decomposing `gql` into individual props, makes it easier to test (IMHO), but this is a pretty extensive set of props. Could some of these be consolidated, like `cancelledAt`, `cancelledByFullName`, and `cancelledByEmail` combined into a single `cancellation` object prop?

packages/app/src/debug/DebugContainer.vue
```diff
@@ -54,13 +69,26 @@ fragment DebugSpecs on Query {
       __typename
       ... on CloudProject {
         id
-        runByNumber(runNumber: 11) {
+        runByNumber(runNumber: 3) {
```
mike-plummer: Any particular significance to `3`, or just making a bit easier to test locally pending a "real" value here?

packages/app/src/debug/DebugContainer.vue
```diff
@@ -1,16 +1,30 @@
...
+        :cancelled-by-full-name="run.cancelledBy?.fullName"
+        :cancelled-by-email="run.cancelledBy?.email"
+        :is-hidden-by-usage-limits="run.isHiddenByUsageLimits"
+        :over-limit-action-type="run.overLimitActionType"
+        :over-limit-action-url="run.overLimitActionUrl"
+        :ci="run.ci"
+        :errors="run.errors"
+      />
       <DebugSpecList
+        v-if="run.totalFailed && ['ERRORED','CANCELLED','TIMEDOUT', 'FAILED'].includes(run.status)"
```
mike-plummer: I would extract this into a guard function to keep the template simpler, but just a personal preference

```
function shouldDisplaySpecsList() {
  ...
}
```

packages/app/src/debug/DebugErrored.vue
```diff
@@ -0,0 +1,40 @@
...
+    :icon="ErrorOutlineIcon"
+    class="flex flex-col mb-24px w-full"
+  >
+    <div class="ml-5px">
+      <ul
+        v-for="(error, index) in errors"
+        :key="index"
+        class="list-disc ml-25px"
+      >
+        <li><pre>{{ error }}</pre></li>
```
mike-plummer, line 14: Probably outside the scope of this PR, but if you end up with a very long error the alert body ends up with an overflow, and when you scroll the background color of the alert box is messed up. Add a `{viewportWidth: 600}` to one of the `DebugRunStates` test cases for this component to see what I mean

packages/app/src/debug/DebugErrored.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+  <Alert
+    :title="t('debugPage.incomplete')"
+    status="warning"
+    :icon="ErrorOutlineIcon"
+    class="flex flex-col mb-24px w-full"
+  >
+    <div class="ml-5px">
+      <ul
```
mike-plummer, line 9: I'm seeing the list item bullet cut off on the left side of the alert box when I run the tests for this component locally

packages/app/src/debug/DebugOverLimit.vue
```diff
@@ -0,0 +1,36 @@
...
+    <LockedProject class="icon-dark-gray-500" />
+    <span class="font-medium mt-24px text-gray-900">
+      {{ t('debugPage.overPlanLimit') }}
+    </span>
+    <span class="mt-10px text-gray-600">{{ t('debugPage.reachedMonthlyUsage') }}</span>
+    <span class="text-gray-600">{{ t('debugPage.upgradeYourPlan') }}</span>
+    <Button
+      size="lg"
+      class="mt-25px"
+      :href="getUrlWithParams({url: overLimitActionUrl, params: { utmMedium: 'Debug Tab', utmSource: getUtmSource() }})"
```
mike-plummer: Worth pulling this into a `computed`?

packages/app/src/debug/DebugCancelledAlert.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+  <Alert
+    :title="t('debugPage.manuallyCancelled')"
+    status="warning"
+    :icon="ErrorOutlineIcon"
+    class="flex flex-col mb-24px w-full"
+  >
+    <div class="flex items-center">
```
mike-plummer, line 8: Styling of this feels very squished, is there meant to be space between these items?
Screen Shot 2022-12-28 at 1 31 17 PM

packages/app/src/debug/DebugPageDetails.vue
```diff
@@ -0,0 +1,72 @@
+<template>
+  <div
+    class="flex flex-col w-full px-24px justify-center items-center align-middle"
```
mike-plummer, line 3: Should this render anything if the status doesn't fall into one of the error conditions below. For example, `FAILED` or `RUNNING` status would just end up reserving blank space

packages/app/src/debug/DebugPageDetails.vue
```diff
@@ -0,0 +1,72 @@
...
+import type { CloudRunStatus, OverLimitActionTypeEnum, DebugSpecListSpecFragment, CloudCiBuildInfoFragment } from '../generated/graphql'
+import DebugCancelledAlert from './DebugCancelledAlert.vue'
+import DebugPassed from './DebugPassed.vue'
+import DebugErrored from './DebugErrored.vue'
+import DebugNoTests from './DebugNoTests.vue'
+import DebugTimedout from './DebugTimedout.vue'
+import DebugOverLimit from './DebugOverLimit.vue'
+
+gql`
+fragment CloudCiBuildInfo on CloudCiBuildInfo {
```
mike-plummer: Typically the fragment is named after the component it supports. Any reason not to in this case?

packages/app/src/debug/DebugTimedout.vue
```diff
@@ -0,0 +1,46 @@
...
+    :icon="ErrorOutlineIcon"
+    class="flex flex-col mb-24px w-full"
+  >
+    {{ t('debugPage.theRunStartedButNeverCompleted') }}
+    <span v-if="ci && hasCiInfo">
+      <i18n-t
+        scope="global"
+        keypath="debugPage.checkYourCiLogs"
+      >
+        <a
```
mike-plummer: Should this use `ExternalLink`?

packages/app/src/debug/DebugTimedout.vue
```diff
@@ -0,0 +1,46 @@
...
+
+const { t } = useI18n()
+
+const props = defineProps<{
+  ci?: CloudCiBuildInfoFragment
+  totalSpecs: number
+  totalSkippedSpecs: number
+}>()
+
+const hasCiInfo: Ref<boolean> = computed(() => {
```
mike-plummer: Is this `Ref<boolean>` type necessary?

packages/app/src/debug/DebugTimedout.vue
```diff
@@ -0,0 +1,46 @@
...
+  return !!(props.ci && props.ci.url && props.ci.formattedProvider && props.ci.ciBuildNumberFormatted)
```
mike-plummer: /nitpick: Could remove the first condition by using null-safe traversal on the other three:

```
return !!(props.ci?.url && props.ci?.formattedProvider && props.ci?.ciBuildNumberFormatted)
```

packages/app/src/runs/utils/day.js
```diff
@@ -1,8 +1,12 @@
 import dayjs from 'dayjs'
 import relativeTime from 'dayjs/plugin/relativeTime'
 import duration from 'dayjs/plugin/duration'
+import customParseFormat from 'dayjs/plugin/customParseFormat'
+import utc from 'dayjs/plugin/utc'
 
 dayjs.extend(relativeTime)
 dayjs.extend(duration)
+dayjs.extend(customParseFormat)
+dayjs.extend(utc)
```
mike-plummer, line 10: Ah, my earlier comment regarded this file. Could we remove the other `extend` calls to `dayjs` in favor of using this instance?

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -650,9 +650,20 @@
...
-    "notLoggedIn": "Not logged in",
-    "notConnected": "Not connected",
-    "noRuns": "No runs",
+    "overPlanLimit": "Over plan limit",
+    "reachedMonthlyUsage": "You've reached the monthly usage limit for your billing plan.",
+    "upgradeYourPlan": "Upgrade your plan to continue recording runs.",
+    "upgradePlan": "Upgrade plan",
+    "contactAdmin": "Contact admin",
+    "wellDone": "Well Done!",
+    "allYourTestsPassed": "All your tests passed.",
```
mike-plummer, line 659: I feel like this should be "All _of_ your tests passed.", but I could be overthinking it

packages/frontend-shared/src/locales/en-US.json
```diff
@@ -650,9 +650,20 @@
...
+    "incomplete": "Incomplete",
+    "specsSkipped": "{totalSkippedSpecs} of {n} spec skipped | {totalSkippedSpecs} of {n} specs skipped",
+    "runHasNoTests": "Run has no tests",
+    "theRunStartedButNeverCompleted": "The run started, but never completed. This can happen when the run is cancelled from CI or when Cypress crashes during running tests.",
```
mike-plummer: /nitpick: I think the comma is unnecessary here.

Also, should it be "crashes _while_ running tests."?

mike-plummer (commented):
Lots of little comments/questions. Looks good for the most part, just small cleanup items

packages/app/src/debug/DebugCancelledAlert.vue
astone123: yeah, it's supposed to look like this

Screen Shot 2022-12-28 at 3 17 52 PM

where are you seeing that? ^

packages/frontend-shared/src/locales/en-US.json
astone123: yeah I had this thought as well... 🤔 just went based off of what the mockup said. According to [some blog](https://proofed.com/writing-tips/word-choice-all-of/), we could use either for this sentence

packages/frontend-shared/src/locales/en-US.json
astone123: yeah I agree with both of these

packages/app/src/debug/DebugCancelledAlert.vue
mike-plummer: Hmm, was seeing it in the `DebugRunStates.cy.tsx` tests but looks good now. Must have been cached/borked local state, disregard.

packages/app/src/debug/DebugErrored.vue
astone123: yeah I think I'll leave this for now - we're still figuring out what this error state will look like in practice. They shouldn't be very long

packages/app/src/debug/DebugCancelledAlert.vue
astone123: Yep, used that instance instead

packages/app/src/debug/DebugCancelledAlert.vue
astone123: Nice, I didn't think of this!

packages/app/src/debug/DebugContainer.vue
astone123: I didn't mean to include this in the changes - was just switching back and forth between different runs when doing local testing. Reverted

packages/app/src/debug/DebugTimedout.vue
astone123: Nope!

packages/app/src/debug/DebugTimedout.vue
astone123: I actually think I just need it on the first condition since they're chained with `&&`. Nice catch

packages/frontend-shared/src/locales/en-US.json
astone123: leaving this for now

packages/app/src/debug/DebugCancelledAlert.vue
lmiller1990: UTC is for machines, no humans - I don't think we actually should show the UTC time. The idea is given UTC, we can show the correct date time to the user based on their location. The opposite isn't always true - there's ambiguity introduced, which is why we store in UTC, but display in locale time.

**TL;DR** you need to do `dayjs.utc(...).local().format('MMM D, YYYY h:mm A')`. This will show the correct time based on the user's local time.

_________

Let me write this out, just to make sure I'm understanding it correctly. I always seem to get confused around timezones. Let me know if any of this looks incorrect.

`2019-01-25T:02:00:00.000Z` would mean the time 2AM on the 25th of January, 2019, in **Greenwich**. The `Z` means UTC, no offset means +0. 

Running your code:

```js
dayjs.utc('2019-01-25T02:00:00.000Z').format('MMM D, YYYY h:mm A UTC')
"Jan 25, 2019 2:00 AM UTC" 
```

This would only be correct for users in **Greenwich**. Everyone else would need to realize it's in UTC, then do the conversion themselves (not a good UX). 

I think we want to localize this. We will need the [timezone](https://day.js.org/docs/en/timezone/timezone) plugin. It won't add any overhead, according to the docs:

> Day.js supports time zone via the [Internationalization API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) in [supported environments](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat#Browser_compatibility).

Example: I am writing this at 9:00 AM on the 29th of Dec in GMT+10. The current time right now, in Greenwich, is `2022-12-28T23:00:00Z` (it's 10 hours behind 9AM - makes sense, I am in GMT+10).

I tested it out, the times look correct to me. You can see using `utc(...).tz('Greenwich')` returns the same time, makes sense, since it's +0.

image

You can just do `local()` - that will get the timezone from the browser, and use that:

image

astone123: Yeah that makes sense - not sure why the decision was made to display it in UTC. I'll bring this up for discussion.

packages/app/src/debug/DebugCancelledAlert.vue
```diff
@@ -0,0 +1,52 @@
...
+            class="h-20px mr-7px w-20px"
+            data-cy="cancelled-by-user-avatar"
+          />
+          <div>{{ cancellation.cancelledBy.fullName }}</div>
+        </div>
+      </template>
+      <template v-if="cancellation.cancelledAt">
+        <div class="rounded-full font-semibold bg-orange-500 h-3px mx-10px w-3px" />
+        <div>
+          {{ dayjs(cancellation.cancelledAt).utc().format('MMM D, YYYY h:mm A UTC') }}
```
lmiller1990: I wrote a big comment about this (please double check my logic/reasoning).

Assuming `cancelledAt` is in UTC (from what you said, it is) this should be:

```suggestion
          {{ dayjs(cancellation.cancelledAt).local().format('MMM D, YYYY h:mm A') }}
```

lmiller1990 (commented):
Please see my comment on timezones, I think getting the correct is critical.

packages/app/src/debug/DebugCancelledAlert.vue
lmiller1990: Outside of developers, it's unlikely anyone else knows what UTC actually means or is used for -- it's probably just a oversight during design. You *never* want to show UTC in an interface.

packages/app/src/debug/DebugCancelledAlert.vue
astone123: updated with https://github.com/cypress-io/cypress/pull/25286/commits/0bf96fec361908af7619a5e28f61486cb4d0d0c7 👍🏻

packages/app/src/debug/DebugErrored.vue
astone123: Which test are you seeing this in? I haven't been able to see it

packages/app/src/debug/DebugPageDetails.vue
astone123: Updated this to render only for specific run statuses

packages/app/src/debug/DebugPageDetails.vue
astone123: good catch, I updated this to include the component name

lmiller1990 (approved):
My comments have been addressed

packages/app/src/debug/DebugErrored.vue
mike-plummer: More borked local state, sorry

mike-plummer (approved):
LGTM, thanks for addressing my comments 👍

PR #25292, opened 2022-12-28
fix: rely only on proxy-from-env

### LINKED ISSUE #19586: Failure to use NO_PROXY appropriately
### Current behavior

In our setup, both environment variables: `https_proxy`, `NO_PROXY` and the `.npmrc` configuration of `https_proxy` / `noproxy` are configured.

The particular Cypress download URL is being redirected Nexus via `CYPRESS_DOWNLOAD_MIRROR` and _should not_ go via the proxy.
All of this would work perfectly except:
During postinstall the `cypress:lib/tasks/download.js` task has the following function:
```
const getProxyForUrlWithNpmConfig = url => {
  return getProxyForUrl(url) || process.env.npm_config_https_proxy || process.env.npm_config_proxy || null;
};
```
Now `getProxyForUrl(url)` returns an _empty string_ if it matches `NO_PROXY` *or* if there are no variables defined - these are not the same states and should be treated separately.
The `||` treats _empty string_ as falsey and therefore picks up the `process.env.npm_config_https_proxy` but *doesn't* handle the corresponding `process.env.npm_config_noproxy` setting so it incorrectly selects the proxy.

Could you please adjust the function `getProxyForUrlWithNpmConfig` to correctly handle `process.env.npm_config_noproxy` and/or treat "nothing defined" differently from proxy explicitly disabled for this domain?

### Debug logs
URLs changed!

Environment variables:
```
    http_proxy = 'http://outbound-proxy.example:3128/'
    https_proxy = 'http://outbound-proxy.example:3128/'
    no_proxy = 'nexus.example,*.example'
    NO_PROXY = 'nexus.example,.example,*.example'
    CYPRESS_DOWNLOAD_MIRROR = 'https://nexus.example/repository/cypress-dist/'
```
npmrc settings:
```
proxy=http://outbound-proxy.example:3128/
https_proxy=http://outbound-proxy.example:3128/
noproxy=*.example
```

Debug logs:
```
.../cypress@8.7.0/node_modules/cypress postinstall: [STARTED] Task without title.
.../cypress@8.7.0/node_modules/cypress postinstall: 2022-01-06T09:52:36.892Z cypress:cli Using CYPRESS_DOWNLOAD_MIRROR from environment variable
.../cypress@8.7.0/node_modules/cypress postinstall: 2022-01-06T09:52:36.892Z cypress:cli Using CYPRESS_DOWNLOAD_MIRROR from environment variable
.../cypress@8.7.0/node_modules/cypress postinstall: 2022-01-06T09:52:36.896Z cypress:cli needed Cypress version: 8.7.0
.../cypress@8.7.0/node_modules/cypress postinstall: 2022-01-06T09:52:36.896Z cypress:cli source url https://nexus.example/repository/cypress-dist/desktop/8.7.0?platform=linux&arch=x64
.../cypress@8.7.0/node_modules/cypress postinstall: 2022-01-06T09:52:36.896Z cypress:cli downloading cypress.zip to "/tmp/cypress-18030.zip"
.../cypress@8.7.0/node_modules/cypress postinstall: 2022-01-06T09:52:36.901Z cypress:cli Downloading package {
.../cypress@8.7.0/node_modules/cypress postinstall:   url: 'https://nexus.example/repository/cypress-dist/desktop/8.7.0?platform=linux&arch=x64',
.../cypress@8.7.0/node_modules/cypress postinstall:   proxy: 'http://outbound-proxy.example:3128/',
.../cypress@8.7.0/node_modules/cypress postinstall:   downloadDestination: '/tmp/cypress-18030.zip'
.../cypress@8.7.0/node_modules/cypress postinstall: }
.../cypress@8.7.0/node_modules/cypress postinstall: 2022-01-06T09:52:36.945Z cypress:cli Using CYPRESS_DOWNLOAD_MIRROR from environment variable
.../cypress@8.7.0/node_modules/cypress postinstall: 2022-01-06T09:52:36.945Z cypress:cli Using CYPRESS_DOWNLOAD_MIRROR from environment variable
.../cypress@8.7.0/node_modules/cypress postinstall: 2022-01-06T09:52:36.946Z cypress:cli URL: https://nexus.example/repository/cypress-dist/desktop/8.7.0?platform=linux&arch=x64
.../cypress@8.7.0/node_modules/cypress postinstall: Error: tunneling socket could not be established, statusCode=403
```

### Cypress Version

8.7.0

### Package Manager

pnpm

### Operating system

Linux

### Other

_No response_

stefdev49:
- don't override result from getProxyForUrl()
- update proxy-from-env to 1.1.0 to get better .npmrc proxy support

- Closes [bug 19586](https://github.com/cypress-io/cypress/issues/19586)

### User facing changelog

- update proxy-from-env to 1.1.0 to get better .npmrc proxy configuration support

### Additional details

My company proxy doesn't work on sites that belong to my company domain, so we can't mirror cypress downloads in our network without setting noproxy properly. Current cypress download code breaks noproxy usage in npmrc and this patch addresses this problem.

proxy precedence will change and no_proxy in npmrc will now work.

### PR Tasks

- [ na] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [na ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -55,7 +55,7 @@
     "minimist": "^1.2.6",
     "ospath": "^1.2.2",
     "pretty-bytes": "^5.6.0",
-    "proxy-from-env": "1.0.0",
+    "proxy-from-env": "1.1.0",
     "request-progress": "^3.0.0",
     "semver": "^7.3.2",
     "supports-color": "^8.1.1",
--- a/cli/lib/tasks/download.js
+++ b/cli/lib/tasks/download.js
@@ -19,8 +19,6 @@ const defaultMaxRedirects = 10
 
 const getProxyForUrlWithNpmConfig = (url) => {
   return getProxyForUrl(url) ||
-    process.env.npm_config_https_proxy ||
-    process.env.npm_config_proxy ||
     null
 }
 
--- a/cli/test/lib/tasks/download_spec.js
+++ b/cli/test/lib/tasks/download_spec.js
@@ -628,6 +628,11 @@ describe('lib/tasks/download', function () {
       expect(download.getProxyForUrlWithNpmConfig(testUriHttp)).to.eq('http://foo')
     })
 
+    it('uses https_proxy on https request', () => {
+      process.env.https_proxy = 'http://foo'
+      expect(download.getProxyForUrlWithNpmConfig(testUriHttps)).to.eq('http://foo')
+    })
+
     it('ignores http_proxy on https request', () => {
       process.env.http_proxy = 'http://foo'
       expect(download.getProxyForUrlWithNpmConfig(testUriHttps)).to.eq(null)
@@ -635,13 +640,21 @@ describe('lib/tasks/download', function () {
       expect(download.getProxyForUrlWithNpmConfig(testUriHttps)).to.eq('https://bar')
     })
 
+    it('ignores https_proxy on http request', () => {
+      process.env.https_proxy = 'https://bar'
+      expect(download.getProxyForUrlWithNpmConfig(testUriHttp)).to.eq(null)
+      process.env.http_proxy = 'http://foo'
+      expect(download.getProxyForUrlWithNpmConfig(testUriHttp)).to.eq('http://foo')
+    })
+
     it('falls back to npm_config_proxy', () => {
       process.env.npm_config_proxy = 'http://foo'
-      expect(download.getProxyForUrlWithNpmConfig(testUriHttps)).to.eq('http://foo')
+      expect(download.getProxyForUrlWithNpmConfig(testUriHttp)).to.eq('http://foo')
       process.env.npm_config_https_proxy = 'https://bar'
       expect(download.getProxyForUrlWithNpmConfig(testUriHttps)).to.eq('https://bar')
       process.env.https_proxy = 'https://baz'
-      expect(download.getProxyForUrlWithNpmConfig(testUriHttps)).to.eq('https://baz')
+      // npm_config_https_proxy takes precedence with proxy-from-env 1.1.0
+      expect(download.getProxyForUrlWithNpmConfig(testUriHttps)).to.eq('https://bar')
     })
 
     it('respects no_proxy on http and https requests', () => {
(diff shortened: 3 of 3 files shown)


cli/package.json
```diff
@@ -55,7 +55,7 @@
     "minimist": "^1.2.6",
     "ospath": "^1.2.2",
     "pretty-bytes": "^5.6.0",
-    "proxy-from-env": "1.0.0",
+    "proxy-from-env": "1.1.0",
```
emilyrohrbough, line 58: @stefdev49 can you update the yarn lock file as well?

emilyrohrbough: I'm looking at 1.1.0 to try & determine what changed between 1.0.0 and 1.1.0 and this might be the reason we are still on 1.0.0....https://github.com/Rob--W/proxy-from-env/issues/13

@flotwig Do you remember the impact of this change or see any issues with updating?

flotwig: Cypress treats `x_PROXY` with higher precedence than `npm_config_x_proxy` because it's more "local" than npm config. proxy-from-env after 1.0.0 changes that precedence to prefer `npm_config_x_proxy` vars over `x_PROXY` vars. Updating proxy-from-env would be breaking for this reason.

emilyrohrbough: Since this provides better support, we likely have a solid use-case for introducing a new patch-package so we can maintain our level of prescience while also leveraging support for npmrc values as well..We already have this setup in the repo, you'd just need to follow the directions outlined here: https://github.com/ds300/patch-package#making-patches

Mind giving it a shot and verify if this still resolves the issues you are seeing experiencing with your company's proxy setup.

emilyrohrbough:
@stefdev49 are you able to take a look at the suggestions?

cli/package.json
flotwig: Yeah, the actual bug here seems to be that the npm_config_noproxy config var isn't used correctly. The best place to fix this would be in `proxy-from-env@1.0.0`, using `patch-package`. We can also avoid the breaking change to precedence that upgrading `proxy-from-env` to `1.1.0` would have.

flotwig: Oh actually, patch-package won't work here because the CLI is distributed as a node_module. So we'd either have to fork proxy-from-env or inline it in the CLI.

emilyrohrbough: @flotwig We are using the module directly, not the CLI so wouldn't patch-package work fine?

emilyrohrbough: Never mind. 🤦🏻‍♀️ . I thought you meant using the dependency’s cli not that this is being pulled into the Cypress CLI. Yeah, patch-package not an option here. I personally think pulling the code forward from `proxy-from-evn` would be the better approach here rather than forking a maintaining an entire repo and needing to release another dependency. It's pretty light weight.

emilyrohrbough:
@stefdev49 It's been a while since we have heard from you. I am going to close this and if you have time to re-visit, we'd appreciate the contribution!

PR #25296, opened 2022-12-28
breaking: `.should()` and `.and()` are now queries.

### LINKED ISSUE #25134: Chains of query commands do not retry anything before an assertion
### Current behavior

Notice in v12 (using 12.1.0 right now) that if there are assertions in the chain of queries, the query commands before an assertion are NOT retried, even if the assertion should pass.

### Desired behavior

Cypress should go all the way to the start of the chain of queries and retry it, even if there are assertions in the mix.

### Test code to reproduce

```js
const double = (n) => n * 2

// @ts-ignore
Cypress.Commands.addQuery('apply', (fn) => (n) => fn(n))

it('fails to retry the cy.its', () => {
  const list = []
  cy.wrap(list)
    .its(0) // first item
    // this assertion breaks the query chain retries
    // it never "sees" the new number 5
    // because it never retries cy.its above
    .should('be.a', 'number')
    .apply(double)
    .should('equal', 10)

  setTimeout(() => {
    list[0] = 1
  }, 1000)

  setTimeout(() => {
    list[0] = 5
  }, 2000)
})

it('retries queries with assertions', () => {
  const list = []
  cy.wrap(list)
    // several queries (without assertion)
    .its(0) // first item
    .apply(double)
    .should('equal', 10)

  setTimeout(() => {
    list[0] = 1
  }, 1000)

  setTimeout(() => {
    list[0] = 5
  }, 2000)
})
```

Screen Shot 2022-12-13 at 12 39 45

https://github.com/bahmutov/cypress-map/blob/main/cypress/e2e/chain-with-assertions.cy.js

### Cypress Version

v12

### Node version

16

### Operating System

mac

### Debug Logs

_No response_

### Other

_No response_

BlueWinds:
- Closes https://github.com/cypress-io/cypress/issues/25134
- Closes https://github.com/cypress-io/cypress/issues/25491

### User facing changelog

Not ready for review. PR description TODO. Marking as 'not a draft' so circleCI will build commits pushed to the branch.

### PR Tasks

  - Update the 12.0.0 migration docs with an addendum speaking to assertions also joining the queries category in 12.x.x.
  - Mention that aliases are 'locators' 


packages/driver/src/cypress/error_messages.ts
```diff
@@ -518,6 +518,12 @@ export default {
         docsUrl: `https://on.cypress.io/${_.toLower(obj.cmd)}`,
       }
     },
+    read_timed_out (obj) {
```
BlueWinds: This new log message is more of a necessary implementation detail than an intentional change. `writeFile()` handles its own timeout logic, so needs the `{{timeout}}ms` block, while `readFile()` now uses the default command_queue retry logic for queries.

Resulting error message looks like:
```
Timed out retrying after 4000ms: `cy.readFile(foo.txt)` timed out.
```

BlueWinds: I spent quite a while trying to get the old error message back (without the "timed out retrying" bit), but it was seriously troublesome / convoluted, so rather than drop more time down that rabbit hole, opted to create a new error message.

packages/driver/src/cypress/command_queue.ts
```diff
@@ -28,37 +34,32 @@ const commandRunningFailed = (Cypress, err, current?: $Command) => {
     return
   }
 
-  const lastLog = current?.getLastLog()
```
BlueWinds, line 31: This logic is all mostly the same, with the exception that we now always call `lastLog.error(err)` if we have a last log - this ensures that we wrap up all assertions (and other log messages) whether or not an onFail() function exists on the error.

packages/driver/src/cy/commands/files.ts
```diff
@@ -4,118 +4,129 @@ import { basename } from 'path'
...
   _log?: Log
 }
 
 export default (Commands, Cypress, cy, state) => {
-  Commands.addAll({
-    readFile (file, encoding, userOptions: Partial<Cypress.Loggable & Cypress.Timeoutable> = {}) {
-      if (_.isObject(encoding)) {
-        userOptions = encoding
-        encoding = undefined
+  Commands.addQuery('readFile', function readFile (file, encoding, options: Partial<Cypress.Loggable & Cypress.Timeoutable> = {}) {
```
BlueWinds: With the change of `.should()` to be a query, `.readFile()` needed to move over as well. This is because readFile() wants to re-read the file from disk every time follow-up assertions fail. For example:

```
cy.readFile('foo.json').should('eq', {bar: 'baz'})
```

is expected to re-read the file from disk until the assertion passes - very query-like behavior.

However, query's inner functions have to by synchronous, while readFile definitely isn't. So we work around this.
1. When the command first executes, we create a promise to read the file from the server.
2. Until the promise resolves, the inner function throws an error, and the command_queue keeps retrying it.
3. Once the promise resolves, the inner function returns the resolved file contents.
4. If our `onFail` handler is called, we know that an upcoming assertion threw an error - either the file didn't exist when it should have, should have existed when it didn't, or the contents didn't match. We throw away the file result and create a new promise.

packages/driver/src/cy/commands/files.ts
```diff
@@ -4,118 +4,132 @@ import { basename } from 'path'
...
   _log?: Log
 }
 
 export default (Commands, Cypress, cy, state) => {
-  Commands.addAll({
-    readFile (file, encoding, userOptions: Partial<Cypress.Loggable & Cypress.Timeoutable> = {}) {
-      if (_.isObject(encoding)) {
-        userOptions = encoding
-        encoding = undefined
+  Commands.addQuery('readFile', function readFile (file, encoding, options: ReadFileOptions = {}) {
```
emilyrohrbough: I would expect this work to be done in its own pull request to make this review smaller. I wouldn't expect this change to impact should() & and().

BlueWinds: I'll move this to its own PR, good call.

It ends up being a prerequisite for updating .should() / .and(). Part of `.readFile()`'s behavior is that if the following assertions fail, it needs to re-read the file from disk - which is basically the idea behind queries! With how assertions work changing, .readFile() needs to fall into the new system as well.

I'll put more details in the description of the new PR.

BlueWinds: https://github.com/cypress-io/cypress/pull/25595 contains these changes stand-alone.

packages/driver/src/cy/commands/files.ts
BlueWinds: See https://github.com/cypress-io/cypress/pull/25595 for these changes / disscussion.

packages/driver/src/cy/assertions.ts
```diff
@@ -275,27 +192,26 @@ export const create = (Cypress: ICypress, cy: $Cy) => {
 
   return {
     finishAssertions,
+    assert,
 
     // TODO: define the specific type of options
     verifyUpcomingAssertions (subject, options: Record<string, any> = {}, callbacks: VerifyUpcomingAssertionsCallbacks = {}) {
```
BlueWinds, line 198: `verifyUpcomingAssertions` is no longer an accurate name, but given that a variety of plugins and custom commands rely on it, I've opted not to rename it something more appropriate (like "verifyImplicitAssertions" or somesuch).

It may not be documented as part of our public API, but I'd consider it public-facing and am therefore hesitant to rename it without an extremely compelling reason.

packages/driver/cypress/e2e/commands/traversals.cy.js
```diff
@@ -160,58 +160,25 @@ describe('src/cy/commands/traversals', () => {
...
-          cy.get('#list')[name](arg).then(function () {
-            const { lastLog } = this
-
-            expect(lastLog.get('snapshots').length).to.eq(1)
-            expect(lastLog.get('snapshots')[0]).to.be.an('object')
-          })
-        })
-
-        it('has the $el', () => {
+        it('snapshots, has the element, and logs with console props', () => {
```
BlueWinds, line 163: No functional changes here, but combining the assertions into one test ends up saving about 60s per run on my machine (since this is executed in a loop), and likely more than that on a CI box.

No need to revisit dom.html hundreds of times. :woman_shrugging:

packages/driver/cypress/e2e/commands/assertions.cy.js
```diff
@@ -769,6 +766,22 @@ describe('src/cy/commands/assertions', () => {
 
         cy.should(() => {})
       })
+
+      it('throws a good error when aliased subject disappears', function (done) {
```
BlueWinds, line 771: Added these two tests based on https://github.com/cypress-io/cypress/issues/25491 / https://github.com/cypress-io/cypress/pull/25493

BlueWinds, line 771:
```suggestion
      // https://github.com/cypress-io/cypress/issues/25491
      it('throws a good error when aliased subject disappears', function (done) {
```

BlueWinds:
New PR for same work: https://github.com/cypress-io/cypress/pull/25738

emilyrohrbough:
Why open a new pr?

PR #20575, opened 2022-03-11
fix(webpack-dev-server): do not encodeUri in loader

lmiller1990:
- Closes https://github.com/cypress-io/cypress/issues/20573

### User facing changelog

Correctly load specs with `[` and `]` characters when using `@cypress/webpack-dev-server`.

### Additional details

Not really relevant to people outside Cypress, but when you back merge this to 10.0-release, you can use this branch for an example (same changes but updated system-tests project: https://github.com/cypress-io/cypress/compare/10.0-release...lmiller1990/20573?expand=1).

### How has the user experience changed?

N/A

### PR Tasks

- [x] Have tests been added/updated?
- [ ] Has the original issue (or this PR, if no issue exists) been tagged with a release in ZenHub? (user-facing changes only)
- [ ] Has a PR for user-facing changes been opened in [`cypress-documentation`](https://github.com/cypress-io/cypress-documentation)? 
- [ ] Have API changes been updated in the [`type definitions`](https://github.com/cypress-io/cypress/blob/develop/cli/types/cypress.d.ts)?
- [ ] Have new configuration options been added to the [`cypress.schema.json`](https://github.com/cypress-io/cypress/blob/develop/cli/schema/cypress.schema.json)?

PR #25299, opened 2022-12-28
fix: Allow component tests with special characters in filepath

### LINKED ISSUE #24588: No tests found when directory name has [] (Component Testing with Next.js)
### Current behavior

Hello, I encountered a similar issue previously reported here https://github.com/cypress-io/cypress/issues/20573 
Running tests inside a square-bracket folder lead to a "No tests found." error.
Screenshot 2022-11-08 at 16 28 23

### Desired behavior

Tests inside square-bracket folders work in a usual way.

### Test code to reproduce

Just wrap a folder name containing component tests in [].

### Cypress Version

10.11.0 and 11.0.1

### Node version

16.15.0

### Operating System

macOS 12.6

### Debug Logs

_No response_

### Other

Next – 12.3.0
dev server framework – 'next'
dev server bundler – 'webpack'

mike-plummer:
- Closes #24588 

### User facing changelog
* Fixes an issue where component test files that contained characters such as brackets (`[]`) would be ignored when running tests. This is a common pattern in Next.js and Gatsby.js projects

### Additional details
This was originally fixed in #20575, but was at some point lost. Tests were added by that PR but unfortunately a loophole in how they were established was just causing them to validate that no tests _**failed**_ - they did not validate that the expected number of tests were loaded and executed. This PR re-introduces the logic change from #20575 and attempts to add logic to the system test to validate that the expected spec files & included tests were found, loaded, and executed.

If anyone has a suggestion for a better way to validate the set of tests that were loaded & run I would love to hear it. Currently pattern matching on `stdout` but this feels a bit brittle

### Steps to test

1. Pull down this branch, build `npm/webpack-dev-server` using `yarn build`
2. Create a Next.js component testing project
3. Create component test files of the following filepath patterns. Ensure each file has a test case within it
`*/[foo]/Test.cy.tsx`
`*/[...Test].cy.tsx`
`*/My Test.cy.tsx`
`*/SpecialChar_ñ.cy.tsx`
4. Launch Cypress, validate that each file is listed in the specs list and executes the test when run

Before: Spec files were found and listed, but would report "No tests found" when executed
After: Specs should now execute test cases

### PR Tasks

### DIFF
--- a/system-tests/test/webpack_dev_server_spec.ts
+++ b/system-tests/test/webpack_dev_server_spec.ts
@@ -1,7 +1,13 @@
 import systemTests from '../lib/system-tests'
+import Fixtures from '../lib/fixtures'
+import path from 'path'
+import globby from 'globby'
+import { escapeRegExp } from 'lodash'
 
 describe('@cypress/webpack-dev-server', function () {
   systemTests.setup()
+
+  const wdsPath = Fixtures.projectPath('webpack-dev-server')
 
   systemTests.it('successfully loads and runs all specs', {
     project: 'webpack-dev-server',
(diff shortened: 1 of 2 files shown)


lmiller1990 (approved):
How did this get reverted :thinking: ???

system-tests/test/webpack_dev_server_spec.ts
```diff
@@ -1,14 +1,48 @@
...
 
+  const wdsPath = Fixtures.projectPath('webpack-dev-server')
+
   systemTests.it('successfully loads and runs all specs', {
     project: 'webpack-dev-server',
     testingType: 'component',
     spec: 'src/**/*',
     browser: 'chrome',
     expectedExitCode: 0,
+    onRun: async (exec) => {
```
ZachJW34, line 18: Did you consider writing a cy-in-cy test to verify these load? Would have reduced the complexity of the logic below (albeit for a performance cost)

ZachJW34 (approved):

system-tests/test/webpack_dev_server_spec.ts
mike-plummer: @ZachJW34 Was a bit of a toss-up. To do `cypress-in-cypress` I would have had to make more changes to write the 4 special test files this is validating into the cy-in-cy project which I feel would have obviated much of the advantage. More complex setup balanced against easier validation - decided to leave as much in place as possible and deal with the messy validation to keep the PR simpler and minor perf gain.

Issue #24251, opened 2022-10-13
Eliminate the need for an internal & public ESLint package


jaffrepaul:
### What would you like?

1. Move [Cypess ESLint Dev](https://github.com/cypress-io/cypress/tree/develop/npm/eslint-plugin-dev) rules into the Cypress root and eliminate the need to maintain an internal & public package.
2. Add scripts to repo root and lint rules to individual packages  
3. Update dependencies and stale issues

### Why is this needed?

There are a few things that can be addressed: 

1. There is an **internal** ESLint plugin with rules shared across different entities. However it's [published on npm](https://www.npmjs.com/package/@cypress/eslint-plugin-dev) (3,783 weekly DLs 🤔) in addition to the **public** [version](https://www.npmjs.com/package/eslint-plugin-cypress) (2,378,591 weekly DLs).
2. The internal version is a dependency and its rules are [extended](https://github.com/cypress-io/eslint-plugin-cypress/blob/master/.eslintrc.json) in the public version with [additional custom rules](https://github.com/cypress-io/eslint-plugin-cypress/tree/master/lib/rules). 
4. Some lint rules seem to be run at the root level across the Cypress repo, where adding scripts at the root to run lint rules in each lib could be more optimal.
5. We could move the rules from the internal dev plugin into the rules at the repo root, remove the dev package and move the public version into the monorepo's npm packages to not be maintaining two and minimize confusion about which to use.
 
Also...
1. Users have begun forking the public version to update & make improvements because the current package has gone stale.
2. ESLint and supporting tooling is very outdated in the dev version, making authoring new rules with modern .js incompatible in the public version. 
3. There are dependencies in the public version that [aren't being used](https://github.com/cypress-io/eslint-plugin-cypress/blob/9bcf51e333b14750df3e0148995df172fd17ed60/package.json#L29-L30) and could use general cleanup and maintenance by way of Dependebot or Renovate so the outdated dependency issue is automated away.

flotwig:
@jordanpowell88 @jaffrepaul Can this also entail removing `lib/scripts` from `eslint-plugin-dev` entirely? As of #25308, I don't see where any of these scripts are used in any repo in the `cypress-io` org. Also, they appear to be broken, I cannot get them to run, even though the tests are passing.

jaffrepaul:
Makes sense to me!

MikeMcC399:
@jennifer-shehane 

[...]

PR #25308, opened 2022-12-29
chore: remove lint-changed from scripts/docs

flotwig:
- Closes 

### User facing changelog

n/a

### Additional details

- @lmiller1990 noticed that this script is broken, I confirmed it was broken before this PR too: https://github.com/cypress-io/cypress/pull/25235#pullrequestreview-1227073464
- Our docs aren't totally correct on it either - `lint-staged` already runs `eslint --fix`, so there's not a situation where a functional `lint-changed` would be needed anyways.
- ~I kept the source code in `npm/eslint-plugin-dev` because tests for it do pass and it's possible it's being used outside of the monorepo.~
	- See comment here, these scripts are actually broken and can most likely be removed, but deferred to later: https://github.com/cypress-io/cypress/pull/25308#discussion_r1059079036

### PR Tasks

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "get-next-version": "node scripts/get-next-version.js",
     "postinstall": "node ./scripts/run-postInstall.js",
     "lint": "lerna run lint --no-bail --concurrency 2",
-    "lint-changed": "lint-changed",
     "prepare-release-artifacts": "node ./scripts/prepare-release-artifacts.js",
     "npm-release": "node scripts/npm-release.js",
     "prestart": "yarn ensure-deps",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -42,7 +42,6 @@
     "get-next-version": "node scripts/get-next-version.js",
     "postinstall": "node ./scripts/run-postInstall.js",
     "lint": "lerna run lint --no-bail --concurrency 2",
-    "lint-changed": "lint-changed",
```
emilyrohrbough, line 45: Did you log an issue to fix the broken script? Or remove in a MVB? Find with removing but if it's not working and its from something we ship, seems like this will get lost.

flotwig: I assumed that because the tests in `eslint-plugin-dev` pass, that the scripts aren't totally broken, but on second glance I actually cannot successfully invoke them no matter what I try. It looks like @cypress-io/tools is looking at removing `eslint-plugin-dev` entirely, so I commented to ask if we can blast these scripts into orbit as part of that change: https://github.com/cypress-io/cypress/issues/24251#issuecomment-1367487970

emilyrohrbough (approved):

ZachJW34 (approved):

PR #25318, opened 2022-12-30
feat: Use JSX/TSX in generated spec filenames

### LINKED ISSUE #24495: Scaffolds TS file for TSX project
### Current behavior

I followed the quickstart https://cypress-documentation-git-react-ct-guide-rework-cypress-io.vercel.app/guides/component-testing/react/quickstart#Install-Cypress

I made a new test with the "create new test" button. It made `cypress/component/ComponentName.cy.ts`. That's not very useful for React, I need a TSX file. I had to rename it manually.

### Desired behavior

Give me a TSX file if I am using React.

### Test code to reproduce

Follow this and do what it says
https://cypress-documentation-git-react-ct-guide-rework-cypress-io.vercel.app/guides/component-testing/react/quickstart#Install-Cypress

### Cypress Version

10.11

### Node version

All of them

### Operating System

Any

### Debug Logs

```shell
NA
```

### Other

_No response_

mike-plummer:
- Closes #24495 

### User facing changelog
- Utilize a JSX/TSX file extension when generating a new empty spec file if project contains at least one file with those extensions. This applies only to component testing and is skipped if the project `specPattern` has been configured to exclude files with those extensions.

### Additional details
Original writeup was specific to React which, while likely the largest use case for JSX, is not the only setup that can use JSX. It's also not necessary to use a JSX extension if using React (or other frameworks). For these reasons, I think the only reliable way to determine if we should use a JSX extension for the default spec filename is to check whether there are other existing files using that extension *and* it would not violate the project's configured specPattern to use it (so we don't generate "disappearing" spec files)

### Steps to test
1. Scaffold a CT project using any framework, keep default specPattern
2. Open Cypress, use "Create new empty spec" button. Verify suggested filename uses `cy.js` (not JSX)
3. Close Cypress, add a JSX file within the project (can be empty)
4. Reopen Cypress, use "Create new empty spec" button. Verify suggested filename now uses `cy.jsx` **_(CHANGED BEHAVIOR)_**
5. Close Cypress, add a `typescript` dependency to the project and rename `cypress.config.js` to `cypress.config.ts`
6. Reopen Cypress, use "Create new empty spec" button. Verify suggested filename now uses `cy.tsx` **_(CHANGED BEHAVIOR)_**
7. Close Cypress, update the CT `specPattern` to `**/*.cy.js`
8. Reopen Cypress, use "New spec" button. Verify suggested filename now uses `cy.ts` (not TSX)

### PR Tasks

### DIFF
--- a/packages/data-context/src/sources/ProjectDataSource.ts
+++ b/packages/data-context/src/sources/ProjectDataSource.ts
@@ -419,11 +419,24 @@ export class ProjectDataSource {
 
   async defaultSpecFileName (): Promise<string> {
     const { specPattern = [] } = await this.ctx.project.specPatterns()
+    let fileExtensionToUse: FileExtension = this.ctx.lifecycleManager.fileExtensionToUse
+
+    // If generating a component test then check whether there are JSX/TSX files present in the project.
+    // If project uses JSX then user likely wants to use JSX for their tests as well.
+    // JSX can be used (or not used) with a variety of frameworks depending on user preference/config, so
+    // the only reliable way to determine is whether there are files with JSX extension present
+    if (this.ctx.coreData.currentTestingType === 'component') {
+      const projectJsxFiles = await this.ctx.file.getFilesByGlob(this.ctx.currentProject ?? '', '**/*.[jt]sx')
+
+      if (projectJsxFiles.length > 0) {
+        fileExtensionToUse = `${fileExtensionToUse}x`
+      }
+    }
 
     return getDefaultSpecFileName({
       currentProject: this.ctx.currentProject,
       testingType: this.ctx.coreData.currentTestingType,
-      fileExtensionToUse: this.ctx.lifecycleManager.fileExtensionToUse,
+      fileExtensionToUse,
       specs: this.specs,
       specPattern,
     })
--- a/packages/data-context/src/sources/ProjectDataSource.ts
+++ b/packages/data-context/src/sources/ProjectDataSource.ts
@@ -426,10 +426,28 @@ export class ProjectDataSource {
     // JSX can be used (or not used) with a variety of frameworks depending on user preference/config, so
     // the only reliable way to determine is whether there are files with JSX extension present
     if (this.ctx.coreData.currentTestingType === 'component') {
+      debug('Checking for jsx/tsx files to determine file extension for default spec filename')
       const projectJsxFiles = await this.ctx.file.getFilesByGlob(this.ctx.currentProject ?? '', '**/*.[jt]sx')
 
       if (projectJsxFiles.length > 0) {
-        fileExtensionToUse = `${fileExtensionToUse}x`
+        debug('At least one jsx/tsx file found in project, utilizing for default spec filename')
+        const generatedSpecFileName = await getDefaultSpecFileName({
+          currentProject: this.ctx.currentProject,
+          testingType: this.ctx.coreData.currentTestingType,
+          fileExtensionToUse: `${fileExtensionToUse}x`,
+          specs: this.specs,
+          specPattern,
+        })
+
+        // There is the possibility that a specPattern has been configured to exclude spec files using jsx/tsx extensions
+        // In this case, fallback to default logic which will generate js/ts filename
+        if (await this.matchesSpecPattern(generatedSpecFileName)) {
+          return generatedSpecFileName
+        }
+
+        debug('jsx/tsx extension would violate configured specPattern, utilizing default spec filename')
+      } else {
+        debug('No jsx/tsx files found, utilizing default spec filename')
       }
     }
 
--- a/packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
+++ b/packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
@@ -819,15 +819,28 @@ describe('ProjectDataSource', () => {
       expect(defaultSpecFileName).to.equal('cypress/component-tests/foo/ComponentName.spec.js')
     })
 
-    it('yields correct filename from specpattern if there are jsx files in component testing', async () => {
-      ctx.coreData.currentTestingType = 'component'
-      sinon.stub(ctx.project, 'specPatterns').resolves({ specPattern: [] })
+    describe('jsx/tsx handling', () => {
+      beforeEach(async () => {
+        ctx.coreData.currentTestingType = 'component'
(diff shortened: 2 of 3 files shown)


packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -419,11 +419,42 @@ export class ProjectDataSource {
 
   async defaultSpecFileName (): Promise<string> {
     const { specPattern = [] } = await this.ctx.project.specPatterns()
+    let fileExtensionToUse: FileExtension = this.ctx.lifecycleManager.fileExtensionToUse
+
+    // If generating a component test then check whether there are JSX/TSX files present in the project.
+    // If project uses JSX then user likely wants to use JSX for their tests as well.
+    // JSX can be used (or not used) with a variety of frameworks depending on user preference/config, so
+    // the only reliable way to determine is whether there are files with JSX extension present
```
marktnoonan, line 427: Ultimately it would be nice to put the extension preference for generated tests in the component testing config or something so there is the ability to override this if needed 🤔. But I think this is a good rubric, especially for new people to get what they probably expect.

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -419,11 +419,42 @@ export class ProjectDataSource {
...
+          currentProject: this.ctx.currentProject,
+          testingType: this.ctx.coreData.currentTestingType,
+          fileExtensionToUse: `${fileExtensionToUse}x`,
+          specs: this.specs,
+          specPattern,
+        })
+
+        // There is the possibility that a specPattern has been configured to exclude spec files using jsx/tsx extensions
+        // In this case, fallback to default logic which will generate js/ts filename
+        if (await this.matchesSpecPattern(generatedSpecFileName)) {
```
marktnoonan, line 444: Good catch

packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
```diff
@@ -818,5 +818,29 @@ describe('ProjectDataSource', () => {
 
       expect(defaultSpecFileName).to.equal('cypress/component-tests/foo/ComponentName.spec.js')
     })
+
+    describe('jsx/tsx handling', () => {
```
marktnoonan, line 822: Could we have coverage for each possibility now? `.js`, `.ts`, `.jsx`, `.tsx`? Is this unwieldy to test?

marktnoonan:
Testing more, but in a Vue project that use jsx for tests, I get a `.js` file extension:

Screen Shot 2023-01-03 at 11 12 06 AM

It's this project: https://github.com/marktnoonan/cy-validators-example/tree/what-component

marktnoonan:
Works well in the React project I tried and in one of our other Vue example projects. Interestingly in the project that's not working as expected, the `defaultSpecFileName` is generated correctly and works fine for new specs that aren't coming from "Create from Component".

Screen Shot 2023-01-03 at 12 34 18 PM
Screen Shot 2023-01-03 at 12 34 28 PM

marktnoonan:
Hmm, @mike-plummer the testing instructions appear to only refer to the "Create new empty spec" button, so maybe I am wrong to expect this to work on the "Create from Component" feature. To close #24495 we would need to handle Create from Component:

> I made a new test with the "create new test" button. It made cypress/component/ComponentName.cy.ts. That's not very useful for React, I need a TSX file. I had to rename it manually.

But as I said above, in my testing for React I get the correct behavior from Create from React Component.

marktnoonan (approved):
Took me a few reads of everything to understand that the scope of this doesn't include "Create from Component" - so the Vue problems I reported are actually unrelated, I made a new issue here to track: https://github.com/cypress-io/cypress/issues/25346

packages/data-context/test/unit/sources/ProjectDataSource.spec.ts
mike-plummer: Got a bit tricker when I looked at stubbing out the TS case here. Since all this logic does is take the "js | ts" default extension and add "x" I didn't consider it necessary to cover all four permutations, but happy to look into it if you think it's of value

marktnoonan: Yeah it looks like `fileExtensionToUse` itself is asserted implicitly plenty, I think this is fine.

packages/data-context/src/sources/ProjectDataSource.ts
```diff
@@ -419,11 +419,42 @@ export class ProjectDataSource {
...
     const { specPattern = [] } = await this.ctx.project.specPatterns()
+    let fileExtensionToUse: FileExtension = this.ctx.lifecycleManager.fileExtensionToUse
+
+    // If generating a component test then check whether there are JSX/TSX files present in the project.
+    // If project uses JSX then user likely wants to use JSX for their tests as well.
+    // JSX can be used (or not used) with a variety of frameworks depending on user preference/config, so
+    // the only reliable way to determine is whether there are files with JSX extension present
+    if (this.ctx.coreData.currentTestingType === 'component') {
+      debug('Checking for jsx/tsx files to determine file extension for default spec filename')
+      const projectJsxFiles = await this.ctx.file.getFilesByGlob(this.ctx.currentProject ?? '', '**/*.[jt]sx')
```
ZachJW34, line 430: Can this check not be framework based? If it's react -> `{fileExtensionToUse}x` else `${fileExtensionToUse}`?

mike-plummer: You don't have to use JSX with React (albeit 99.9% of people do), and you can use JSX with non-React frameworks (like we do with our Vue tests). Doing a "is there already a JSX/TSX file present" seemed like the most reliable option, but I'll defer to the team

ZachJW34: I was trying to think through some edge-cases, like a user having some `jsx` file for storybook but not wanting to have that pollute their generated Cypress tests if they don't plan on using jsx to write their CT tests. I think the proper solution is to have a more deterministic output like what framework they have and then, like @marktnoonan pointed out, having a user config override.

In the end this is just a suggestion for a filename and the user can change it if they want. I'm cool with this if this gets us closer, especially for react devs as that is our majority.

ZachJW34 (approved):

PR #25345, opened 2023-01-03
fix: Reflect browser status in launchpad if failed to open

### LINKED ISSUE #23798: Closed Chrome Browser Isn't Reflected in Launchpad
### Current behavior

When opening chrome / chrome beta from the launchpad, when you close the browser to select a new browser, the launchpad is not re-focused and the lauchpad shows the following when re-focusing yourself:

Screen Shot 2022-09-13 at 8 03 12 AM

### Desired behavior

When Chrome closes, the Launchpad should not show browser as active.

### Test code to reproduce

1. Open launchpad
2 select chrome/chrome beta
3. close browser with exit button
4. view launchpad and see issue

### Cypress Version

10.7.0

### Node version

16.14.2

### Operating System

MacOS 12.5.1

### Debug Logs

_No response_

### Other

_No response_

mike-plummer:
- Closes #23798

### User facing changelog
* Fixes issue in Launchpad where a browser would be stuck in a "opening" state if it failed to open or was closed very quickly after being opened.

### Additional details
Just adding a "reset state" flow if an error is thrown on browser open and re-throwing the error to keep existing logic intact

### Steps to test
1. Open a project (e2e or ct)
2. From launchpad, select Electron and open
3. Before Electron finishes loading (immediately after window opens) close the Electron window using `command-W`
4. Observe launchpad resets to "no browser open" state
5. Re-open Electron and observe it successfully loads
6. You can repeat procedure with Chrome or Firefox but it can be tricky to reproduce since these browsers take a "double `command-Q` to kill them during the open operation

**_Before_**

https://user-images.githubusercontent.com/11482842/210415327-4b689623-b556-432b-b0a4-50ffcff6f978.mov

_**After**_

https://user-images.githubusercontent.com/11482842/210415377-e8aeac4f-999b-4b49-836a-02eb9a2e528b.mov

### PR Tasks

### DIFF
--- a/packages/server/lib/browsers/index.ts
+++ b/packages/server/lib/browsers/index.ts
@@ -146,11 +146,24 @@ export = {
 
     debug('opening browser %o', browser)
 
-    const _instance = await browserLauncher.open(browser, options.url, options, automation)
+    try {
+      const _instance = await browserLauncher.open(browser, options.url, options, automation)
 
-    debug('browser opened')
+      debug('browser instance: %o', _instance)
 
-    instance = _instance
+      debug('browser opened')
+
+      instance = _instance
+    } catch (e) {
+      /*
+      If an error happens during open (e.g. user closes window before we finish loading) then
+      clean up browser state and re-throw error
+      */
+      debug('Error while opening browser %o', e)
+      ctx.browser.setBrowserStatus('closed')
+      throw e
+    }
+
     instance.browser = browser
 
     // TODO: normalizing opening and closing / exiting
--- a/packages/server/lib/browsers/index.ts
+++ b/packages/server/lib/browsers/index.ts
@@ -10,6 +10,7 @@ import os from 'os'
 import { BROWSER_FAMILY, BrowserLaunchOpts, BrowserNewTabOpts, FoundBrowser } from '@packages/types'
 import type { Browser, BrowserInstance, BrowserLauncher } from './types'
 import type { Automation } from '../automation'
+import type { DataContext } from '@packages/data-context'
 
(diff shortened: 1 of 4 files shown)


marktnoonan:
Closing Chrome quickly with cmd+q after opening, I was able to get the launchpad to stick in the opening state

Screen Shot 2023-01-03 at 5 25 24 PM

In the terminal I see I caused this error:

Screen Shot 2023-01-03 at 5 25 48 PM

I'd guess this is happening outside the `catch` added by this PR, maybe it's not representative of errors we are hoping to catch during the browsers true "opening" state.

Edit: after taking a closer look at a where the catch is, it does seem very safe, maybe there is a crack between the browser opening and us updating the status. I think we might be running afoul of the [1 second delay here](https://github.com/cypress-io/cypress/blob/mikep/23798-closed-browser-status/packages/server/lib/browsers/index.ts#L225-L235)?

If there is an error after opening, but before that delay expires, we don't currently reset `browserStatus`.

ZachJW34:
Worked well for me but not sure if this work closes https://github.com/cypress-io/cypress/issues/23798. The issue with the original doesn't involve any error states like closing the new browser too quickly but rather how the browser is still reported as running whenever you close out of it via the "exit" application button.

Can we adjust the original issue or not link it if there is still work to be done?

mike-plummer:
@ZachJW34 Oh, I glossed over the "close the window without closing the browser" part. Oops. Hmm, I would say this is still a worthwhile fix since it does address a bug in the opening flow, but it looks like @marktnoonan is still seeing issues. I'll kick this to draft and take a deeper look as time allows

PR #25360, opened 2023-01-04
feat: IATR-M0 Page Header (#24722)

### LINKED ISSUE #25339: [IATR](M1.0) - Add tooltip for ellipsis when test title is shortened due to smaller breakpoints
_Current Behavior_

In the `DebugFailedTest` component, there is responsive behavior to help reduce the width of the test titles at smaller breakpoints.  At a certain small size, if there are more than 3 title parts, part of the test title parts are removed and hidden by an ellipsis in the UI. (See arrow screenshot)  The user does not know what tile parts are hidden. 

_Desired Behavior_
The ellipsis should have a tooltip that shows the missing parts.  The title parts should be separated by a `>`.

marktnoonan:
- Closes #25339

### User facing changelog

NA, merging to feature branch

### Additional details

Added an optional `originalTitle` property to the `MappedTitlePart` type. There are numerous other ways to handle this but this one seems fine to me. We could maybe work on the readability of this component at a future time.

### Steps to test

Visit the component test for `DebugFailedTest.vue` and let it run. Observe tooltip appears when you hover over the ellipses: 

Screen Shot 2023-01-04 at 1 05 17 PM

Sections that are truncated by ellipses, but still visible, will not have this tooltip, based on the wording of the original issue. To do that we would need a little more sophisticated approached to determine that the browser has truncated the text and conditionally enable the tooltip, the presence of the class alone is not enough. So that is out of scope for this ticket.

### PR Tasks

### DIFF
--- a/packages/app/src/debug/DebugFailedTest.cy.tsx
+++ b/packages/app/src/debug/DebugFailedTest.cy.tsx
@@ -198,7 +198,7 @@ describe('<DebugFailedTest/>', () => {
     assertRowContents(testResult)
 
     cy.contains('...').realHover()
-    cy.contains('[data-cy=tooltip-content]', 'Test content 2').should('be.visible')
+    cy.contains('[data-cy=tooltip-content]', 'Test content 2 > Test content 3 > Test content 4').should('be.visible')
 
     cy.percySnapshot()
   })
--- a/packages/app/src/debug/DebugFailedTest.cy.tsx
+++ b/packages/app/src/debug/DebugFailedTest.cy.tsx
@@ -175,7 +175,7 @@ describe('<DebugFailedTest/>', () => {
     cy.percySnapshot()
   })
 
-  it('tests responsvie UI', { viewportWidth: 700 }, () => {
+  it('tests responsive UI', { viewportWidth: 700 }, () => {
     const testResult: TestResults = {
       id: '676df87874',
       titleParts: ['Test content', 'Test content 2', 'Test content 3', 'Test content 4', 'onMount() should be called once', 'hook() should be called twice and then'],
@@ -196,6 +196,9 @@ describe('<DebugFailedTest/>', () => {
     ))
 
     assertRowContents(testResult)
+
+    cy.contains('...').realHover()
+    cy.contains('[data-cy=tooltip-content]', 'Test content 2').should('be.visible')
 
     cy.percySnapshot()
   })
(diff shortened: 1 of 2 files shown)


mike-plummer (approved):

packages/app/src/debug/DebugFailedTest.cy.tsx
```diff
@@ -197,6 +197,9 @@ describe('<DebugFailedTest/>', () => {
 
     assertRowContents(testResult)
 
+    cy.contains('...').realHover()
+    cy.contains('[data-cy=tooltip-content]', 'Test content 2').should('be.visible')
```
warrensplayer: This should contain everything that was hidden by the ellipsis separated by `>`
```suggestion
    cy.contains('[data-cy=tooltip-content]', 'Test content 2 > Test content 3 > Test content 4').should('be.visible')
```

packages/app/src/debug/DebugFailedTest.vue
```diff
@@ -30,7 +30,17 @@
           titlePart.type === 'LAST-1' ? 'shrink-0 whitespace-pre' :
           titlePart.type === 'LAST-0' ? 'pl-2.5 truncate' : 'px-2.5 truncate'"
       >
-        {{ titlePart.title }}
+        <template v-if="titlePart.title === '...'">
```
warrensplayer:
```suggestion
        <template v-if="titlePart.type === 'ELLIPSIS'">
```
Check with type instead of title

packages/app/src/debug/DebugFailedTest.vue
```diff
@@ -121,6 +133,7 @@ const failedTestData = computed(() => {
         {
           title: '...',
           type: 'ELLIPSIS',
+          originalTitle: ele,
```
warrensplayer: This will need to gather up all the "MIDDLE" parts to construct the `originalTitle` passed to the ELLIPSIS tooltip

warrensplayer (changes requested):
Need to account for multiple "MIDDLE" parts being hidden by the ellipsis

packages/app/src/debug/DebugFailedTest.vue
marktnoonan: That makes sense, thanks! I think I was missing the underlying concept here, that it's really combining the parts 👍

warrensplayer (approved):
Updates look great.  I really like the refactor for some of the enum names and variable names

PR #25364, opened 2023-01-04
fix: re-enable CYPRESS_INTERNAL_VITE_DEV development

ZachJW34:
### User facing changelog
n/a

### Additional details
`CYPRESS_INTERNAL_VITE_DEV=1 yarn dev` was not working due to Vite updates changing the stdout and causing our process spawner to not proceed with the gulp process. I also added `stripAnsi` to the matcher as the Vite stdout is colorful.

I also removed the `rimraf ./node_modules/.vite` as that is where the vite optimized deps live. @tgriesser do you remember why we had to do this? Maybe CI will tell me after I break everything...

### Steps to test
Run `CYPRESS_INTERNAL_VITE_DEV=1 yarn dev`. I had to run this, try and navigate to the app and then restart the process. Vite was optimizing deps but there was too many and it was timing out. It got closer and closer each time and then eventually started working. I'm on a slow machine though so maybe it was just a me thing.

Faster dev environment when making changes to the app/launchpad. Updates went from 20s to less than 1 second.

### PR Tasks

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "check-ts": "vue-tsc --noEmit",
     "build-prod-ui": "cross-env NODE_ENV=production vite build",
-    "clean": "rimraf dist && rimraf ./node_modules/.vite && echo 'cleaned'",
+    "clean": "rimraf dist && echo 'cleaned'",
     "clean-deps": "rimraf node_modules",
     "test": "echo 'ok'",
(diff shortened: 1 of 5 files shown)


tgriesser:
> I also removed the rimraf ./node_modules/.vite as that is where the vite optimized deps live. @tgriesser do you remember why we had to do this? Maybe CI will tell me after I break everything...

Seems fine to me, I think there were issues with cached files but it was awhile ago, maybe not an issue with more recent vite versions

tgriesser (approved):

lmiller1990 (approved):

PR #25365, opened 2023-01-04
chore: support vite v4 for component testing

### LINKED ISSUE #24969: CT: vite v4 support
### What would you like?

Official support for Vite v4 (currently in beta): https://github.com/vitejs/vite/releases

### Why is this needed?

We'll want to support Vite 4 even before it becomes stable, including scaffolding in the Launchpad as well. If there are bugs, we should make it easy for users to find them (which includes being able to scaffold a CT project using Vite v4.alpha/beta).

### Other

I did some preliminary testing and everything seems to be working besides Launchpad setup. The major change is Rollup v3.

https://github.com/vitejs/vite/blob/v4.0.0-alpha.4/packages/vite/CHANGELOG.md

We should add some system-tests to verify v4 functionality.

marktnoonan:
- Closes #24969

### User facing changelog

Vite 4+ will no longer show a warning when used for component testing projects.

### Steps to test

- Open component testing in the launchpad package and verify that there is no longer a warning
- Create an unconfigured (for CT) Vite project (I removed config on https://github.com/cypress-io/cypress-component-testing-apps/tree/main/vue3-vite-ts). Confirm that CT setup works correctly in the launchpad, Vite is detected as the bundler, and that setup is successful.

Previously there were warnings about the dependency in the launchpad, and setup was possible but only by ignoring what Cypress tells you about Vite versions it supports.

### PR Tasks


packages/scaffold-config/src/dependencies.ts
```diff
@@ -76,7 +76,7 @@ export const WIZARD_DEPENDENCY_VITE = {
   package: 'vite',
   installer: 'vite',
   description: 'Vite is dev server that serves your source files over native ES modules',
-  minVersion: '^=2.0.0 || ^=3.0.0',
+  minVersion: '^=2.0.0 || ^=3.0.0 || ^=4.0.0',
```
mike-plummer, line 79: Not specific to this PR, just curious to get some thoughts - as we add more major versions to the support baseline do we want to continue this pattern or would something like this be easier to manage?
`>=2.0.0 <5.0.0`

system-tests/__snapshots__/vite_dev_server_fresh_spec.ts.js
```diff
@@ -1,850 +1,4 @@
-exports['@cypress/vite-dev-server react executes all of the tests for vite3.0.2-react 1'] = `
```
mike-plummer, line 1: This deletion doesn't feel right - I would expect a large addition here for 4.0.4

marktnoonan: You're right, looks like I committed the wrong updated snapshot, will come back to this after some IATR stuff.

packages/scaffold-config/src/dependencies.ts
marktnoonan: I guess that's easier, unless we encounter the rare situation where a single major isn't supported, but newer and older versions are (like with Node).

ZachJW34 (approved):

system-tests/__snapshots__/vite_dev_server_fresh_spec.ts.js
marktnoonan: Updated!

packages/scaffold-config/src/dependencies.ts
mike-plummer: Good point, there isn't an "exclude" operator so we'd end up having to stitch ranges together. In any event, not an issue at the moment just brainstorming a bit.

mike-plummer (approved):

Ancient-Dragon:
@marktnoonan hey thanks for the work here, any chance of a release soon to include this fix?

marktnoonan:
@Ancient-Dragon this should be in a release later this month. In the meantime it's safe to just ignore the warning Cypress gives about Vite 4 support and use it. It warns because we don't explicitly support breaking changes to a bundler without testing ourselves. This PR adds some testing that verifies Vite 4 works with Cypress, but didn't require any changes to Cypress code. So you should be good to go.

PR #25380, opened 2023-01-06
fix: Restrict dependency versions to known supported ranges

### LINKED ISSUE #25379: Remove unbounded ranges from CT dependency version checks
CT currently uses unbounded ranges for Webpack, Typescript, and Angular when checking dependency versions. We should restrict the upper bound to ensure we do not claim to support a version that hasn't been certified

mike-plummer:
- Closes #25379 

### User facing changelog
* Fixes issue where component tests could incorrectly treat new major versions of certain dependencies as supported

### Steps to test
No functional changes to current behavior, validate existing system tests pass

### PR Tasks


ZachJW34 (approved):

astone123 (approved):
Good catch 👍🏻

PR #25384, opened 2023-01-06
fix: support custom workspaceRoot for angular CT projectConfig

barbados-clemens:
### User facing changelog

angular CT will respect a `workspaceRoot` build option if provided, which is useful for when cypress opens in a nested folder. i.e. an nx workspace

### Additional details

When using Angular component testing within an nx workspace, assets will not load as the proxied path was incorrectly set since Cypress opens within a nested folder. 

### Steps to test

> This change does requires changes to the @nrwl/angular nx component testing preset. Mainly the offset value is removed and a custom 'workspaceRoot' is passed to cypress. This can be done within node_modules if really wanted to test it works before this change is in Nx

1. make a new nx angular workspace, `npx create-nx-workspace@latest --preset=angular`
2. add an asset to the apps asset folder, `apps/<your-app-name>/src/assets`
3. load the asset in the app component, i.e. ` />`
4. add component testing to the app, `npx nx g cypress-component-configuration`
5. test the component that's loading the asset
6. Asset should load

Doesn't change the current experience, only improves for those who might need custom sets (like Nx :) )

### PR Tasks


barbados-clemens:
Related to https://github.com/nrwl/nx/issues/12987

ZachJW34:
@barbados-clemens thanks for the PR!

Could you revert the formatting changes in this PR? It is hard to grep what changed.

barbados-clemens:
> Could you revert the formatting changes in this PR? It is hard to grep what changed.

Sorry about that. This should be the changes here. basically adding an a conditional workspaceRoot property that will take precedence in certain places if present, if not then it defaults back the existing behavior of the projectRoot.

jordanpowell88 (approved):

chrisbreiding:
@ZachJW34 Would you be able to take another look at this PR now that your feedback has been addressed?

ZachJW34:
@barbados-clemens I created an `nx-angular` app following the `Steps to test` and had no issues loading assets with the version of Cypress installed (^12.2.0). I also checked out the repo from the original issue (https://github.com/lydemann/cypress-assets-bug/tree/nx), updated `cypress` and `nrwl/angular` to latest and the assets and icon loaded. I didn't need the changes introduced in this PR to get assets loading.

Do you know what I'm overlooking? We did recently fix some asset loading bugs with Angular. Do you have a repo that demonstrates the need for these proposed changes that I can look at?

barbados-clemens:
@ZachJW34 it does seem to be working all of a sudden so I'm not sure what changes happened. I was working with latest Cypress v12.3.0 when I made it. which looks to still be the cypress@latest version. I'm going to have to retry some things to make sure I'm not missing something.

ZachJW34:
@barbados-clemens lmk what you find, handling assets has been the biggest hurdle with our Angular support so I wouldn't be surprised if there is something not wired up perfectly.

barbados-clemens:
Talking with some folks who originally talked about the issue with me. it's _mostly_ fixed, but there are some places where it isn't working as intended.

They are stating there seems to be an issue with loading assets with the HttpClient and stylePreprocessorOptions don't look to be respected. 

working on getting a reproducible for you.

barbados-clemens:
So tried everything people have mentioned is a problem with latest and can't seem to find the issue. 
Here is the repo working with loading assets from an app/lib with image tag, HttpClient, and stylePreprocessorOptions. 
https://github.com/barbados-clemens/cy-asset-style-demo

I'm going to close the PR since it works, but if someone has something else that isn't working let me know so I can update the repo and try.

PR #25457, opened 2023-01-13
fix: match alternate edge version string format

### LINKED ISSUE #25455: Cypress not finding beta/dev builds of Edge on linux
### Current behavior

After installing Edge Beta and Edge Dev on my machine I do not see them listed as an option to run my E2E tests against. The Stable version does appear in the UI with no issues.

### Desired behavior

After installing Edge Beta and Edge Dev on my machine I should see them listed as an option to run my E2E tests against

### Test code to reproduce

There is no test code to run as the bug appears before any tests are run.

### Cypress Version

12.3.0

### Node version

v18.12.1

### Operating System

Manjaro Linux 6.0.15-1-MANJARO

### Debug Logs

_No response_

### Other

After looking into how Cypress knows about browsers I noticed two potential issues with the version regex in the `packages/launcher/lib/browsers.ts` file and the binary names.

For the version regex I'm not sure if this would be the cause of the issue but the regex for Edge Dev is `/Microsoft Edge Dev (\S+)/m` and for Edge Beta is `/Microsoft Edge Beta (\S+)/m` but when running `/usr/bin/microsoft-edge-dev --version` it returns `Microsoft Edge 110.0.1587.6 dev` which would not match the regex string. This is the same for the beta version where it returns `Microsoft Edge 109.0.1518.49 beta`

The second potential issue I see is that the Edge Beta is missing a listing for `microsoft-edge-beta` which is present for the stable and dev versions. This looks like it was done in #8931 but the beta and canary channels were missed.

CarbonCollins:
- Closes #25455

### Additional details
Edge was returning an alternate format for its version string which meant the browser detection version regex was not matching. This change allows the regex to match on both the original form and the new format which switches the version number and channel places arround.

Affects of this change would allow Cypress to match version strings from Edge in both `Microsoft Edge Beta 109.0.1518.49` and `Microsoft Edge 109.0.1518.49 beta` forms. This also applies to the Edge Canary and Edge Dev browsers.

### PR Tasks


CarbonCollins:
I've been looking into this one a little more and I'm not sure if the current format of these regex in the develop  branch are even correct in the first place... I get this alternate format on both my linux machine and on a MacOS machine too.

Do we know if they worked prior to me raising this PR as I would like to determine if we need to still support the existing version string format?

CarbonCollins:
To further follow up im seeing it in `Microsoft Edge 110.0.1587.6 dev` on my machine and `Microsoft Edge 110.0.1587.6 Dev` on the mac hence the case insensitivity addtion, however with the way that the code currently extracts the version from the regex the commit in this PR wont work...

CarbonCollins:
Just to extend from my previous message at least, the updated regex should allow for both forms of the version string, the original `Microsoft Edge Dev 110.0.0.0` and the one I'm seeing `Microsoft Edge 110.0.0.0 Dev` / `Microsoft Edge 110.0.0.0 dev` Without needing to change the way its decoded further down using a hardcoded capture group number...

I'm not sure if its the most appropriate way of doing this but the regex in the PR should return the version in either form as the 1st capture group regardless allowing the downstream code to remain unchanged.

emilyrohrbough (approved):

cli/CHANGELOG.md
```diff
@@ -1,10 +1,19 @@
  <!-- See the ../guides/writing-the-cypress-changelog.md for details on writing the changelog. -->
+ ## 12.4.2
+
+_Released 01/32/2023 (PENDING)_
+
+**Bugfixes:**
```
emilyrohrbough, line 6:
```suggestion
**Bugfixes:**

```

cli/CHANGELOG.md
```diff
@@ -1,10 +1,20 @@
...
+- Fixed an issue where alternative Microsoft Edge Beta, Canary, and Dev binary versions were not being discovered by Cypress. 
+  Fixes [#25455](https://github.com/cypress-io/cypress/issues/25455).
+
 ## 12.4.1
 
 _Released 01/27/2023_
 
 **Bugfixes:**
 
+
```
emilyrohrbough:
```suggestion
```

cli/CHANGELOG.md
```diff
@@ -1,4 +1,13 @@
  <!-- See the ../guides/writing-the-cypress-changelog.md for details on writing the changelog. -->
+ ## 12.4.2
```
emilyrohrbough:
```suggestion
## 12.4.2
```

AtofStryker (approved):
I'd like to get some unit tests in the future inside `browsers_spec` to validate Edge discovery seeing we don't have any, but I feel that's likely out of scope here.

